PR #124, opened 2015-12-28
Add Mooselurk

kamarcum:


kamarcum:
> I noticed that you're looting boxes (combat-setup.lic, line 457), but not training Locksmithing?
I've been running `;lock` manually. I bet that's not necessary.

> Did you get your problem with weapon descriptions fixed?
I haven't. I think I don't have the gear equipped that some scripts expect, and somehow "first" gets interpreted as a weapon name. That's why I'm not training brawling, incidentally. My stance gets messed up to match the weapon it thinks I'm using and I end up parry-stanced empty handed. 

> You're not training Scholarship?
Scholarship is training incidentally from Astrology and First Aid so far. 

> You do not have Utility as a skill to be trained
Good catch! Thanks.

rpherbig:
> I've been running ;lock manually. I bet that's not necessary.

Correct. Adding Locksmithing to your training list will call `pick` and then `sell-loot`.

> I think I don't have the gear equipped that some scripts expect, and somehow "first" gets interpreted as a weapon name. 

For weapon descriptions, wield the weapon and then `glance`. You'll see something like `You glance down to see a crude flail made from knotted rags in your right hand and nothing in your left hand.` In that case you would put `a crude flail made from knotted rags` in combat-setup.

rpherbig:
> Scholarship is training incidentally from Astrology and First Aid so far.

Makes sense. I would recommend double checking that this is enough to lock Scholarship.

rpherbig:
Closing this. You should convert to the YAML approach while you're at it.

kamarcum:
> You should convert to the YAML approach while you're at it.

Agreed. Thanks for the review along the way!

PR #137, opened 2015-12-31
Move characters to subdirectory

kamarcum:
I'm almost positive the symlinker batch file won't pick up this subdirectory properly, so that'll need to be fixed.

What do you think of moving characters to a subdirectory?

Is "character" even the right term?


rpherbig:
:+1:

I'd like to see create_symlinks.bat updated as part of this PR.

I'm not sure about 'characters'. I could go with 'setup' or 'config'. Maybe 'character-profiles'?

kamarcum:
I went with `profiles`. I *think* I have the batch file set up correctly. I can confirm that the newly added bash script works for me.

rpherbig:
Make sure you update the README with your bash script.

rpherbig:
What do you think about renaming profiles from Character-setup.yaml to just Character.yaml?

rcuhljr:
still waiting on profiles rename in dependencies and setupconvert.

kamarcum:
> I get 'access denied' even when I run it from an elevated command prompt.

Did you get any leads on that? Is there any more information in the error?



> You do not pass create_symlinks.bat an argument.

I split the readme entries on this, but I don't know what the right language is for that feature in the windows command prompt. Is it an environment variable?

PR #140, opened 2016-01-01
temp fix: added assembly to blacksmithing slack logic

hwoo79:


hwoo79:
it should skip it since there would/should be no argument, how does it handle weaponsmithing products that don't have assembly?

 will test tomorrow, hopefully.

hwoo79:
So basically, the script does the same thing that it does with weaponsmithing and no assembly argument.  It times out twice, goes to the oil sub then completes.  Probably need to write and issue on both errors.




[forge]>push tub
Gusts of hot steam billow from the tub as you submerse the scorching hot flat-bladed tongs.

The worked metal looks to be in need of some oil to preserve and protect it.
Roundtime: 25 sec.
R>
[forge]>get tongs from anvil
You get some unfinished flat-bladed bronze tongs from atop an iron anvil.
>
[forge]>get my  from my pack
Please rephrase that command.
>
[forge: *** No match was found after 15 seconds, dumping info]
[forge: messages seen length: 1]
[forge: message: Please rephrase that command.]
[forge: checked against [/^You get/i, /^You are already/i, /^What do you/i, /^What were you/i]]
[forge]>assemble my tongs with my 
Please rephrase that command.
>
[forge: *** No match was found after 15 seconds, dumping info]
[forge: messages seen length: 1]
[forge: message: Please rephrase that command.]
[forge: checked against [/affix it securely in place/i, /and tighten the pommel to secure it/i, /carefully mark where it will attach when you continue crafting/i]]
[forge]>get my oil from my pack
You get a flask of oil from inside your traveler's pack.
>
[forge]>pour my oil on my tongs
After applying some of the oil to the rag, you vigorously buff several rough spots on the flat-bladed tongs and spread the protective sheen across the metal's surface.
Roundtime: 20 sec.
Applying the final touches, you complete working on some flat-bladed bronze tongs.

rpherbig:
That's what I was afraid of - thanks for testing it!

One solution is to capture the bput('push tub', ...) result (line 134-ish) and act on that. However, we might be able to assume that if @part is not nil, we need to call the assemble logic. The latter is almost certainly easier to code.

What do you think?

hwoo79:
The easy fix is to check for @part and go to oil if it's nil. 

However, I think the whole section needs to be rewritten to react off the messaging.

rpherbig:
This was fixed in #143.

PR #143, opened 2016-01-01
added assemble_part logic to fix an argument error

hwoo79:
stripped out add_oil to use the logic in assemble_part and added a check for nil @part argument


rpherbig:
:+1: I see some opportunities for refactoring, but this looks good to me. We talked in-game and all four combinations currently supported by forge work (Blacksmithing-but-no-assemble, Blacksmithing-with-assemble, Weaponsmithing-but-no-assemble, Weaponsmithing-with-assemble). He tested with jewelry box, pickaxe, knife, and prod.

PR #171, opened 2016-01-08
Adding Thanatology/Ritual stuff for Necromancers

hwoo79:
WIP for adding Necromancer Thanatology training.

It's ALIVE!

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -150,6 +150,7 @@ class LootProcess
     @arrange_types = arrange_types
     @no_skins = []
     @too_many = []
+    @construct = []
   end
 
   def execute(game_state)
@@ -189,12 +190,43 @@ class LootProcess
   def dispose_body(game_state)
     return if DRRoom.dead_npcs.length < 1
     game_state['mob_died'] = true
-
-    check_skinning(DRRoom.dead_npcs.first) if @skin
+    
+#NECRO THANATOLOGY CHECK STUFF GOES BELOW
+    
+    if @necro {check_thanatology(DRRoom.dead_npcs.first)}
+      dispose_body_loot
+      
+#NECRO THANATOLOGY CHECK STUFF GOES ABOVE
+    
+    elseif @skin { check_skinning(DRRoom.dead_npcs.first) }
+      dispose_body_loot
 
     bput('loot', 'You search', 'I could not find what you were referring to')
     pause 0.2 while DRRoom.dead_npcs.length > 0
   end
+  
+#NECRO THANATOLOGY RITUAL METHOD GOES BELOW
+  
+  def check_thanatology(mob_noun)
+    return if @construct.include?(mob_noun)
+    perform_type = 
+    perform_message = "perform #{perform_type} on #{mob_noun}"
+      bput (perform_message, 'YOU CANNOT DIDDLE WITH CONSTRUCTS', 'corpse to make one long,', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'corpse to make a few quick,')
+      when '
+        break
+      when 'YOU CANNOT DIDDLE WITH CONSTRUCTS' #GET SOME REAL MESSAGING
+        echo("adding #{mob_noun} to construct list: #{@construct}") if $debug_mode
+        @construct.push(mob_noun)
+        return
+      
+      #perform preserve success:  You bend over the CRITTER'S corpse to make one long, continuous cut with your ritual knife that just barely pierces into it.  Unseen energies seep into the creature's fluids, suspending the corpse in unnaturla statis for a time.
+      #perform dissect success:   You carefully position the corpse and make a cut across its length, 
+      #perform consume success:   You kneel over the ADJECTIVE CRITTER's corpse and make a few quick precise cuts with your ritual knife.  Your cuts follow the vital lines of the corpse, causing its residual life force to bubble forth, ready for draining.
+      #perform harvest success:   You bend over the CRITTER'S corpse to make a few quick, precise motions with your ritual knife, focusing you thanatological insight intently in order to prepare it for harvesting.
+      #perform harvest fail:      
+      #failure due to construct:  
+
+#NECRO THANATOLOGY RITUAL METHOD GOES ABOVE
 
   def check_skinning(mob_noun)
     return if @no_skins.include?(mob_noun)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -152,7 +152,7 @@ class LootProcess
     @too_many = []
     @constructs = []
     @ritual_type = ritual_type
-    @necro = necro
+	@necro = necro
   end
 
   def execute(game_state)
@@ -192,7 +192,7 @@ class LootProcess
   def dispose_body(game_state)
     return if DRRoom.dead_npcs.length < 1
     game_state['mob_died'] = true
-    echo("necrocheck after game_state: #{@necro}") if $debug_mode
+	  echo("necrocheck after game_state: #{@necro}") if $debug_mode
       if @necro
 		echo("necrocheck in dispose_body: #{@necro}") if $debug_mode
         check_thanatology(DRRoom.dead_npcs.first)  #NECRO THANATOLOGY CHECK STUFF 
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -189,12 +190,43 @@ class LootProcess
...
+    
+#NECRO THANATOLOGY CHECK STUFF GOES BELOW
+    
+    if @necro {check_thanatology(DRRoom.dead_npcs.first)}
+      dispose_body_loot
+      
+#NECRO THANATOLOGY CHECK STUFF GOES ABOVE
+    
+    elseif @skin { check_skinning(DRRoom.dead_npcs.first) }
+      dispose_body_loot
```
rcuhljr: For clarity I'd use if end blocks, so 

if @necro
  check_thanatology(DRRoom.dead_npcs.first)
elsif @skin
  check_skinning(DRRoom.dead_npcs.first)
end
dispose_body_loot

You're calling dispose loot twice right now, so just move it outside of the if statement since you want to call it regardless.

combat-trainer.lic
```diff
@@ -189,12 +190,43 @@ class LootProcess
...
 
     bput('loot', 'You search', 'I could not find what you were referring to')
     pause 0.2 while DRRoom.dead_npcs.length > 0
   end
+  
+#NECRO THANATOLOGY RITUAL METHOD GOES BELOW
+  
+  def check_thanatology(mob_noun)
```
rcuhljr: every time you move into a new block of code (def, when, if...) you need a corresponding end to close that block.

combat-trainer.lic
```diff
@@ -189,12 +190,47 @@ class LootProcess
...
     pause 0.2 while DRRoom.dead_npcs.length > 0
   end
+  
+  def check_thanatology(mob_noun)  #NECRO THANATOLOGY RITUAL METHOD 
+    return if @construct.include?(mob_noun)
+    perform_type = @perform_types[mob_noun] || 'somethinggoeshere' #Setting this up like the skinning method because I have no idea what I'm doing.
+    perform_message = "perform #{perform_type} on #{mob_noun}"
+      bput (perform_message, 'Rituals do not work upon constructs', 'corpse to make one long,', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'corpse to make a few quick,')
+      when 'corpse to make one long,', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'corpse to make a few quick,'
+        break
```
rcuhljr: you don't need a when for each possible condition,  also when's go with a [case statement](http://www.skorks.com/2009/08/how-a-ruby-case-statement-works-and-what-you-can-do-with-it/) which you don't have in front of your bput.

rpherbig:
How's this going @hwoo79 ?

hwoo79:
Sorry, I haven't had time to work on it this weekend. Maybe tomorrow. Sorry for the delay.

rpherbig:
I wasn't trying to hurry you, I just wanted to check in and see if you needed anything from us :smile:

hwoo79:
Well, I think my frustration point Friday was trying to figure out where I'm going to pull the ritual list from. 

I'm reading and rereading the ruby primer and hoping it'll soak in.

combat-trainer.lic
hwoo79: check_skinning at line 236 has a when for each condition , is there a reason it does?

combat-trainer.lic
```diff
@@ -150,6 +150,8 @@ class LootProcess
     @arrange_types = arrange_types
     @no_skins = []
     @too_many = []
+    @construct = []
+    @ritual_type = ritual_type
```
rcuhljr: `ritual_type` missmatch versus `ritual_types`

combat-trainer.lic
```diff
@@ -150,6 +150,8 @@ class LootProcess
     @arrange_types = arrange_types
     @no_skins = []
     @too_many = []
+    @construct = []
```
rpherbig: I would recommend renaming this to '@constructs' (note the plural) - this will help you remember that it is an array.

combat-trainer.lic
```diff
@@ -189,12 +191,46 @@ class LootProcess
...
   end
+  
+  # partially completed, only works for a single ritual.
+  def check_thanatology(mob_noun)  #NECRO THANATOLOGY RITUAL METHOD 
+    return if @construct.include?(mob_noun)
+    ritual_type = @ritual_types[mob_noun] || 'preserve'
+    perform_message = "perform #{ritual_type} on #{mob_noun}"
+      case bput (perform_message, 'Rituals do not work upon constructs', 'corpse to make one long,', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'corpse to make a few quick,')
+      when 'corpse to make one long,', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'corpse to make a few quick,'
+        break
```
rcuhljr: remove this break, it doesn't make sense in this context.

combat-trainer.lic
```diff
@@ -189,12 +191,46 @@ class LootProcess
   def dispose_body(game_state)
     return if DRRoom.dead_npcs.length < 1
     game_state['mob_died'] = true
-
-    check_skinning(DRRoom.dead_npcs.first) if @skin
-
+      if @necro
```
rpherbig: Where does @necro come from? I don't see it in the parameter list to initialize.

hwoo79: is there a way to echo my variables to see if they're getting initialized?   like:

echo("ritual_type: #{ritual_type}") if $debug_mode

hwoo79: also, I think the problem might be that I'm not calling in my variables correctly.

when I def main(combat_vars)
  setup

variablename = combat_vars.skinning['variable']

that defines variablename as whatever is in the yaml under skinning then looks at the left side if the colon for variable and then sets variablename as whatever is on the right side of the colon, correct?

example, if I had this method in my yaml,

ex1:
  exF: exA

then this def in combat-training,

variablename = combart_vars.ex1['exF']

it would set variablename to exA?

rpherbig: Short answer: yes.

```combat_vars.blah``` is a shortcut to get you whatever is stored in the YAML (a top-level var with name 'blah'). It only works on the top-level settings - anything below that has to be accessed differently.

The easy case is when the top-level name in the YAML maps directly to a value. For example, ```cambrinth: armband``` would be accessed with ```combat_vars.cambrinth``` and return a string, ```armband```.

Skinning is slightly more complicated. Given a YAML of:
```
skinning:
  skin: true
  arrange_all: false
  arrange_count: 0
```
you could say ```combat_vars.skinning``` and it parses everything under the name ```skinning```. In this case, the data under name ```skinning``` are a bunch of key-value pairs. In Ruby-speak, that means a Hash: ```{ 'skin' => true, 'arrange_all' => false, 'arrange_count' => 0 }```. To get something out of the hash, access it by a key: ```combat_vars.skinning['skin']``` would give you ```true```.

Comment (rpherbig) [inline combat-trainer.lic]: As for the echo question, your code looks fine. I use that technique myself when debugging.

Comment (hwoo79) [inline combat-trainer.lic]: okay, so in my yaml I am creating a thanatology hash with two pairs as such:

    thanatology:
      necro: true
      ritual_type: dissect

I'm calling them into `combat-trainer.lic` with the following lines:

    def main(combat_vars)
      setup

      necro = combat_vars.thanatology['necro']
      ritual_type = combat_vars.thanatology['ritual_type']

then initializing them in the `LootProcess` class and making them an instance variable like this:

    def initialize(necro:false, ritual_type:preserve)
      @ritual_type = ritual_type
      @necro = necro

I've also added them to `LootProcess.new` as:

    combat_processes = [
      LootProcess.new(necro: necro, ritual_type: ritual_type)

Then, when I try the `echo("necro variable value: #{necro}") if $debug_mode` I get an undefined/initialized variable error.

I'm sorry for the trouble, but I'm not getting why it's not bringing it in.

File: combat-trainer.lic [hunk 8]
```diff
@@ -189,12 +192,53 @@ class LootProcess
   def dispose_body(game_state)
     return if DRRoom.dead_npcs.length < 1
     game_state['mob_died'] = true
-
-    check_skinning(DRRoom.dead_npcs.first) if @skin
-
+	  #echo("necrocheck after game_state: #{necro}") if $debug_mode
```
rpherbig: This needs to be @necro. ```necro``` is only in scope for the ```initialize``` method.

combat-trainer.lic
```diff
@@ -142,7 +142,7 @@ end
 class LootProcess
   include DRC
 
-  def initialize(skin:false, arrange_all:false, arrange_count:0, lootables:[], arrange_types:{})
+  def initialize(skin:false, arrange_all:false, arrange_count:0, lootables:[], arrange_types:{}, ritual_type:preserve, necro:false)
```
rcuhljr: preserve isn't a valid default value here, it's not a string or anything else. this should probably be `ritual_type:{}` since an empty hash seems like the correct default for ritual type.

rcuhljr:
Good progress so far.

PR #174, opened 2016-01-09
Wear longbows when emptying hands

kamarcum:
Since I started backtraining bows, a few different scripts have failed when they call empty_hands.

At first I thought to check the gear to see if it *could* be worn, and to wear it whenever empty_hands was called. I'm worried, though, that wearing things that can be worn might interfere with some of the repair and loot selling logic.

So here's a horrible hack instead. Feedback is appreciated.


rcuhljr:
longbows can be stowed or tied with the right containers.

kamarcum:
It looks like pretty much everywhere `empty_hands` is called, it'd be better to use `$EquipmentManager.return_held_gear`. I didn't change its usage in the favor script as that script doesn't depend on equipment manager and is only invoked manually so far as I can tell.

kamarcum:
The implementation checked in at the moment works really well for weapons and gear, but fails for things that aren't tracked in equipment manager, like compendiums.

rcuhljr:
`return_held_gear` should be easily expandable to return false if it fails in its attempts to stow items, in which case we just resort to stowing. make that swap and I'd be ok with this.

kamarcum:
Is this good to merge? I've used it for several hours without incident.

common.lic
```diff
@@ -197,9 +197,14 @@ module DRC
   module_function :get_noun
 
   def empty_hands
```
rpherbig: Where is this method used now?

kamarcum: Only in favor, I believe, which doesn't currently depend on the equipment
manager. I'm hesitant to add it since I imagine favor being an obvious
thing for someone completely new to the project to invoke.

rpherbig: Maybe just move this code there?

I'm trying to reduce the chance someone uses the wrong helper.

kamarcum: Yeah I can support that.

kamarcum: It looks like I was wrong and `forage` in common uses this too. How do you feel about merging this as it is?

rpherbig: Does forage still need it? I thought the logic in crossing training would
have handled such things.

rpherbig:
@kamarcum Have you been running this? Have you seen any problems so far?

rpherbig:
@kamarcum What's the status of this?

rpherbig:
@kamarcum How can I help with this PR?

kamarcum:
Just remerged master and tested this for about an hour. Things seem fine. Let's merge it.

PR #189, opened 2016-01-17
Cast moon-based spells

kamarcum:


kamarcum:
> Should this be set in the yaml file ("moon: true" or something)?

Maybe? My inclination was maybe to check guild in dependency and load moonwatch if you're a moon mage. What do you think?

rpherbig:
Is there a reason a moon mage would not want moonwatch running? We do have a lot of precedent for downloading scripts the user didn't *explicitly* request...

rpherbig:
As for the MOON_SPELLS var, I'm concerned that there might be setup issues if someone uses a different abbrev than what you have. A bool sidesteps that.

rpherbig:
@kamarcum What's the status of this?

crossing-training.lic
```diff
@@ -315,6 +317,21 @@ class CrossingTraining
...
     end
 
+    if MOON_SPELLS.include? data['abbrev']
+      moon = UserVars.moons['visible'].first
+      unless moon
+        echo "No moon available to cast #{data['abbrev']}"
+        unless Script.running? 'moonwatch'
+          echo "moonwatch is not running. Starting it now"
+          custom_require 'moonwatch'
+          echo "Run `;autostart add moonwatch` to avoid this in the future"
```
rcuhljr: dependency manager manages our autostarts, nothing of ours should be in ;autostart other than dependency. use ;e autostart('scriptname')

rcuhljr:
I'd prefer having this be data on the spell object as herbig mentioned.

rpherbig:
@kamarcum How can I help with this PR?

crossing-training.lic
```diff
@@ -22,6 +22,7 @@ class CrossingTraining
     @running = true
     @idling = false
     @settings = get_settings(variable.drop(1) + %w(art athletics perform stealing))
+    @settings.outdoor_room ||= 674
```
kamarcum: No idea what you think of adding an outdoor_room setting. I was using this while testing stuff in 705.

crossing-training.lic
```diff
@@ -301,6 +302,28 @@ class CrossingTraining
...
+    if data['moon']
+      moon = UserVars.moons['visible'].first
+      unless moon
+        echo "No moon available to cast #{data['abbrev']}"
+        unless Script.running? 'moonwatch'
+          echo "moonwatch is not running. Starting it now"
+          custom_require 'moonwatch'
+          echo "Run `;autostart add moonwatch` to avoid this in the future"
+        end
+        weather = bput('weather', "inside", "You glance up at the sky.")
```
kamarcum: I found out `weather` is a way to check if you're inside without suffering round time.

crossing-training.lic
```diff
@@ -314,6 +315,28 @@ class CrossingTraining
...
+        unless Script.running? 'moonwatch'
+          echo "moonwatch is not running. Starting it now"
+          custom_require 'moonwatch'
+          echo "Run `;e autostart('moonwatch')` to avoid this in the future"
+        end
+        weather = bput('weather', "inside", "You glance up at the sky.")
+        if weather =~ /inside/
+          walk_to @settings.outdoor_room
+        end
+        fput('perceive moons')
```
rcuhljr: Do we really want to do this every time?

kamarcum: Probably not, no. I didn't have a full appreciation for moonwatch when I wrote this.

kamarcum:
I'm inclined to merge this today. I've been running it for a couple weeks without incident. The `perceive moons` call in crossing-trainer isn't ideal, but it comes up pretty rarely.

None of this even runs unless you have a spell with `moon: true`, so impact is minimal even if it was literally a forkbomb.

PR #198, opened 2016-01-21
[WIP] Add alternate attack logic.

hwoo79:
This is a WIP.  

**DO NOT PULL**

Can I pull the variables in the same way as skinning does, or do I have to go through the equipment manager?  Now that I'm thinking about it, I probably need to pull the attack_type in when each weapon changes.

Can one of you point me to the weapon change method?

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -666,14 +666,14 @@ class AttackProcess
         pause 1
         waitrt?
         if is_offhand && @backstab
-          hidden_attack = @no_stab_current_mob || !is_offhand ? '#{type} back' : 'backstab left'
-          command = checkhidden ? hidden_attack : '#{type} left'
+          hidden_attack = @no_stab_current_mob || !is_offhand ? 'attack back' : 'backstab left'
+          command = checkhidden ? hidden_attack : 'attack left'
         else
-          hidden_attack = @no_stab_current_mob || !@backstab.include?(weapon_skill) ? '#{type} back' : 'backstab'
-          command = checkhidden ? hidden_attack : '#{type}'
+          hidden_attack = @no_stab_current_mob || !@backstab.include?(weapon_skill) ? 'attack back' : 'backstab'
+          command = checkhidden ? hidden_attack : 'attack'
         end
       else
-        command = is_offhand ? '#{type} left' : '#{type}'
+        command = is_offhand ? 'attack left' : 'attack'
       end
       fput(command)
     else
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -589,7 +589,7 @@ end
 class AttackProcess
   include DRC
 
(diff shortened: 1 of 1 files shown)


hwoo79:
Actually, I probably need to pull the attack_type into equipmanager when it changes weapons and then pass that into combat-trainer.  

I can't really think of a way to reliably pull that into combat_trainer from the yaml and keep it synchronized with the weapon.

rcuhljr:
I'd suggest an alternate approach. In your yaml

```
attack_types:
  Twohanded Edged: slice
  Polearms: jab
  Brawling: claw
```

and then pass attack types through like you passed thanatology, then `AttackProcess` can use `game_state['weapon_skill']` to look up the attack to use from attack_types.

rpherbig:
I don't think attack types needs to live in the yaml. If you look at the top of combat-trainer, there are a few global vars (`$thrown_skills` and so on) that define what are basically constants. This feels like another of those.

hwoo79:
So, the attack type does actually depend on the weapon.  A small edge weapon could be a slicer, like a scimitar, or a piercer (jab), like a short sword or sabre.  While blunt weapons are pretty consistent with impact (swing) being the preferred attack type; edged weapons can be pierce or slice.  

That's kind of why my preference would be to keep the attack type with the weapon.

rpherbig:
I think that convinces me the attack type needs to be part of the weapon definition.

When you swap, does the attack type change?

hwoo79:
It potentially could.  With a riste (LE/LB/2HB) it would change from slice to impact, I think there's some type of LE/LB hammer that swaps from impact to pierce.

rpherbig:
That's what I was afraid of. In that case, each weapon probably needs a mapping of possible weapon skill => attack type.

hwoo79:
I'm thinking it might be better to put it under each weapon def in gear: that way the user could define their own attack preference.  Well, it's have to be more like:  

    weapon_name: a darkstone riste 
      weapon_skill: Large Edge
        attack_type: slice
      weapon_skill: Large Blunt
        attack_type: swing
      weapon_skill: Two Handed Blunt
        attack_type: swing

then the script would have to pull both the weapon_name, then swap to the weapon_skill, then pull the attack_type.


--is this going to be worth the effort for some efficiency?

rpherbig:
We could use `attack` for swappable weapons (which we can already identify from the `swappable` setting) and `attack_type` otherwise. That would sidestep the issue, at least in the short-term.

rpherbig:
What verb does Brawling use? Since Brawling's weapon is `''` we may have trouble adding `weapon_skill` and `attack_type` settings to it.

hwoo79:
gouge, claw, or punch are equal.  Gouge is a little faster, better balance, but some things it doesn't damage.  I would say punch is the best candidate for this.

PR #236, opened 2016-02-14
Add just enough spellcasting to pick.lic

kamarcum:


pick.lic
```diff
@@ -66,6 +72,14 @@ class LockPicker
     fput("open my #{@settings.picking_box_storage}")
   end
 
+  def buff(buffs)
+    buffs.each do |name, spell|
+      fput("pre #{spell['abbrev']} #{spell['mana']}")
+      pause 15
```
rcuhljr: Can we just replace the pause with a waitfor fully prepared?

PR #268, opened 2016-02-27
Train with moonblades

kamarcum:


combat-trainer.lic
```diff
@@ -98,6 +98,10 @@ class SetupProcess
     @last_skill = game_state['weapon_skill']
     game_state['action_count'] = 0
 
+    if DRStats.guild == 'Moon Mage' && UserVars.moons['visible'].empty?
+      echo("skipping summoned weapons because no moons are visible") if $debug_mode
+      weapon_training = weapon_training - @summoned_weapons
```
rpherbig: You could probably in-line this logic with the `min_by` call.

combat-trainer.lic
```diff
@@ -149,13 +153,16 @@ class SetupProcess
   def check_weapon(weapon_skill)
     echo("checking weapons as #{@last_skill.inspect}!=#{weapon_skill}") if $debug_mode
 
-    if @elemental_weapons.include?(weapon_skill)
-      summon_elemental_weapon unless @elemental_weapons.include?(@last_skill)
-      shape_elemental_weapon(weapon_skill)
+    if @summoned_weapons.include?(weapon_skill)
```
rpherbig: We're checking `@summoned_weapons.include?(@last_skill)` in both branches of this if. Can we commonize some of this logic?

PR #315, opened 2016-03-16
added thicket viper area

hwoo79:


hwoo79:
Thanks for the pull!

It appears that vipers may need to use bescort to get there.

rpherbig:
Are these the thicket vipers by rock guardians (https://elanthipedia.play.net/mediawiki/index.php/RanikMap11) or by leucros (https://elanthipedia.play.net/mediawiki/index.php/RanikMap11a)?

Either way, I think you're right. You should be able to see how bescort gets to rock guardians or leucros and work with that, or you can file an issue for us to look at it when we have a chance.

hwoo79:
The room numbers that I added are before rock guardians, but there is a leucro/viper mix area between leucros and geni also.

PR #331, opened 2016-03-31
Adding stone carving

OdnaZvezda:
Use stone carving workorders for Engineering 276+

### DIFF
--- a/carve.lic
+++ b/carve.lic
@@ -1,0 +1,118 @@
+=begin
+  Suggestions and contributions are welcome: https://github.com/rpherbig/dr-scripts
+=end
+
+custom_require(%w(common common-crafting))
+
+class Carve
+  include DRC
+  include DRCC
+
+  def initialize(args:[])
+    settings = get_settings
+    @bag = settings.crafting_container('backpack')
+    @finish, @type, @chapter, @recipe_name, @material, @noun = args
+
+    case @type
+    when /carving/
+      carve_item
+    else
+      echo('Carve does not yet support that form of carving.')
+    end
+  end
+
+  def get_or_fail(item)
+    waitrt?
+    case bput("get my #{item}", '^You get', '^You are already', '^You pick up', '^What do you', '^What were you')
+    when 'What do you', 'What were you'
+      echo("You seem to be missing: #{item}")
+      exit
+    end
+  end
+
+  def store_thing(item)
+    waitrt?
+    bput("put my #{item} in my #{@bag}", 'You put your', 'What were you referring to')
+  end
+
+  def turn_to(section)
+    bput("turn my book to #{section}", 'You turn your', 'The book is already')
+  end
+
+  def carve_item
+    get_or_fail('carving book')
+    turn_to("chapter #{@chapter}")
+    turn_to("page #{find_recipe(@chapter, @recipe_name)}")
+    bput('study my book', 'Roundtime')
+    store_thing('book')
+    get_or_fail('chisel')
+    get_or_fail("#{@material} rock")
+    carve('cut my rock with my chisel')
+  end
+
+  def carve(command)
+    waitrt?
+    case bput(command,
+              'rough, jagged edges',
+              'determine it is no longer level',
+              'developed an uneven texture along its surface',
+              #'Applying the final touches',
+              'You cannot figure out how to do that.',
+              'you see some discolored areas',
+              'Roundtime'
+             )
+    when 'rough, jagged edges'
+      waitrt?
+      store_thing(checkright)
+      get_or_fail('rifflers')
+      command = "rub my #{@noun} with my rifflers"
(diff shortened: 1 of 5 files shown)


rpherbig:
Just a reminder to add Medellia as a sample Bard on the wiki (https://github.com/rpherbig/dr-scripts/wiki/Setup-Process) once this is merged.

OdnaZvezda:
Will do

rcuhljr:
Do we know what happens if you run out of polish during a work order? We might want to change it to order more if there's less than N uses of polish when we start/finish.

carve.lic
```diff
@@ -0,0 +1,118 @@
...
+    carve('cut my rock with my chisel')
+  end
+
+  def carve(command)
+    waitrt?
+    case bput(command,
+              'rough, jagged edges',
+              'determine it is no longer level',
+              'developed an uneven texture along its surface',
+              #'Applying the final touches',
```
rpherbig: Why is this line commented out?

carve.lic
```diff
@@ -0,0 +1,118 @@
...
+  include DRC
+  include DRCC
+
+  def initialize(args:[])
+    settings = get_settings
+    @bag = settings.crafting_container('backpack')
+    @finish, @type, @chapter, @recipe_name, @material, @noun = args
+
+    case @type
+    when /carving/
```
rpherbig: To me this indicates that we should roll this into `shape` (and maybe rename it to `engineer`?). Don't worry about it as part of this PR, but we'll want to look into it later.

carve.lic
```diff
@@ -0,0 +1,118 @@
...
+custom_require(%w(common common-crafting))
+
+class Carve
+  include DRC
+  include DRCC
+
+  def initialize(args:[])
+    settings = get_settings
+    @bag = settings.crafting_container('backpack')
+    @finish, @type, @chapter, @recipe_name, @material, @noun = args
```
rpherbig: We should do a check that the correct number of args were passed in. I'd hate to see it get partway through a craft and then fail on a missing arg. `Forge` does this, but `shape` and `sew` do not.

Don't worry about it as part of this PR, but we'll want to look into it later.

carve.lic
```diff
@@ -0,0 +1,118 @@
...
+    store_thing(checkright)
+    case @finish
+    when /log/
+      logbook_item
+    when /stow/
+      store_thing(@noun)
+    when /trash/
+      dispose_trash(@noun)
+    end
+    # NOTE: using 2 volume recipes leaves no leftover material when ordering a small rock
```
rpherbig: I'm not familiar with carving. Does this line (or the two commented out lines after it) indicate further work that needs to be done?

carve.lic
rpherbig: (And line 79?)

profiles/Medellia-setup.yaml
```diff
@@ -0,0 +1,674 @@
+---
```
rpherbig: There are a few things we might want to look at since we're going to present this as a Bard config for new users (like cooldown timers and prep messages).

Don't worry about it as part of this PR, but we'll want to look into it later.

workorders.lic
```diff
@@ -77,6 +80,57 @@ class WorkOrders
     end
   end
 
+
+  def carve_items(info, item, quantity)
```
rpherbig: A good candidate for some refactoring, either here in `workorders` or `common-crafting`.

Don't worry about it as part of this PR, but we'll want to look into it later.

carve.lic
OdnaZvezda: They're commented because I wasn't sure I was going to be able to use them and forgot to take them out.

"Applying the final touches" never gets matched because "Roundtime" happens on the line before "Applying the final touches".  As a work around, I am attempting to "cut my #{@noun} with my chisels" and matching "You cannot figure out how to do that" to note completion of the item.

With shaping, you start with the drawknife and never use it again, this is not the case with carving.  I have seen the completion message happen while using chisels and polish.  I would have otherwise duplicated the behavior of forge and it's pound logic.

carve.lic
OdnaZvezda: The note should say 3 volume recipes.  I initially commented the lines out with the intent that carve would be used with recipes that only use a small rock(3 volumes), meaning no leftover material.  When you order the rock, they give you a deed as boulders are too heavy to pick up.  Some reworking of the script could be done to stow the deed and get/tap it once you're in a workroom.  If recipes requiring other volumes are used, cleanup will be required.

carve.lic
OdnaZvezda: Regarding polish, adding a quick get/analyze on it should work.  I don't know that I've ever used it more than once on a given item, but like the rifflers/rasp, I'm sure it's possible.   Polish cannot be COMBINEd.  Should polish with less than 3-5 uses be discarded in order to combat possible hoarding due to invendory reversal at login?

PR #373, opened 2016-04-28
Fixed for Vowels

dstritzel:
When getting an ordinary lockpick it will use "an" for the vowel. I think removing "a" would be the best option and doing the greedy regex from there.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -115,7 +115,7 @@ class LockPicker
   def attempt_open(box)
     waitrt?
     echo "attempt_open(#{box})" if UserVars.lockpick_debug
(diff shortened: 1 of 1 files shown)


rcuhljr:
Good catch, although did you mean to change [line 213](https://github.com/dstritzel/dr-scripts/blob/lockpick-fixing/pick.lic#L213) which is where we are getting lockpicks?

dstritzel:
Oops, yes, I should have just grabbed the file I changed and put that in place. I decided to be fancy and redo the change in a new branch. I'll fix that.

dstritzel:
Going to redo and rebase my change to only change what I intended.

PR #374, opened 2016-04-28
Fix for lockpicks with Vowel names

dstritzel:
When getting an ordinary lockpick it will use an for the vowel(Ordinary, Iron, ect). I think removing a would be the best option.. Thanks for @rcuhljr for pointing out my mistake in the last pull request.


rcuhljr:
Thanks for contributing!

PR #379, opened 2016-05-09
Refactor combat trainer processes

rpherbig:
Combat-trainer is duplicating default values for settings when:
- Defining each setting in base.yaml
- The settings are read in CombatTrainer::make_processes
- The named params are passed to each individual process

@rcuhljr 's suggestion was to pass the settings into each process and let `initialize` unpack them. That's a good step forward.


combat-trainer.lic
```diff
@@ -1091,35 +1092,22 @@ class CombatTrainer
...
+      SetupProcess.new(settings),
+      SpellProcess.new(settings),
       AbilityProcess.new(buffs: settings.buff_nonspells({}), kneel_khri: settings.kneel_khri(false), khri_preps: settings.khri_preps),
       LootProcess.new(
-        skin: skin,
-        arrange_all: arrange_all,
-        arrange_count: arrange_count,
+        skin: settings.skinning['skin'],
+        arrange_all: settings.skinning['arrange_all'],
+        arrange_count: settings.skinning['arrange_count'],
```
rcuhljr: I think we might have problems here for people who delete the skinning settings or something? Not sure but this seemed red flag to me.

rpherbig: This should be addressed once I move LootProcess over.

rpherbig: If they delete `skinning`, then they'll get the default in base.yaml. If they override it (such as `skinning:`), then yeah, we might have a problem. But I'll change it to `settings.skinning({})`.

PR #387, opened 2016-05-12
Move some logic into game state

rpherbig:
Follow-up to #379 and #382. My goal here is to move *some* logic into GameState.

This isn't yet ready to merge, but shows an example of what I want to do: combat-trainer doesn't actually care if `last_weapon_skill == weapon_skill`, what it really wants to know is if `weapon_skill_changed?`.

Please let me know if you have any thoughts on other logic I can move.


rpherbig:
https://en.wikipedia.org/wiki/Law_of_Demeter, http://programmers.stackexchange.com/questions/212128/what-is-a-feature-envy-code-and-why-is-it-considered-a-code-smell, etc.

PR #388, opened 2016-05-14
Move settings into game state

rpherbig:
WIP, but this should serve as an example of what it would look like to move settings into game state.

The Processes would then focus on logic.

I picked `dance` at random as a first setting to move. Maybe I should focus on moving predicates first?

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -225,7 +225,7 @@ class LootProcess
 
   # partially completed, only works for a single ritual.
   def check_thanatology(mob_noun, ritual, game_state)
-    return if game_state.constructs.include?(mob_noun)
+    return if game_state.construct?(mob_noun)
 
     perform_message = "perform #{ritual} on #{mob_noun}"
     case bput(perform_message, 'Rituals do not work upon constructs', 'precise motions with your ritual knife', 'corpse to make one long', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'roundtime')
@@ -236,7 +236,7 @@ class LootProcess
       fput 'drop material'
       return
     when 'Rituals do not work upon constructs'
-      game_state.constructs.push(mob_noun)
+      game_state.construct(mob_noun)
       return
     end
     waitrt?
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -55,12 +43,12 @@ class SetupProcess
       game_state.next_clean_up_step
       return true
     end
-    game_state.dancing = should_dance?
-    game_state.now_retreating = should_retreat?
+    game_state.update_dancing(DRRoom.npcs.length)
+    game_state.update_retreating(DRRoom.npcs.length)
```
rcuhljr: Could we just call game_state.update_npcs(DRRoom.npcs)? and have it deal with these?

combat-trainer.lic
```diff
@@ -55,12 +43,12 @@ class SetupProcess
...
     if game_state.dancing
-      dance(game_state)
+      game_state.dance(@weapon_training)
```
rcuhljr: From reading this line I really don't have any good idea what this call would do.

combat-trainer.lic
```diff
@@ -75,55 +63,29 @@ class SetupProcess
...
-    game_state.action_count >= @target_action_count || DRSkill.getxp(game_state.weapon_skill).to_i >= game_state.target_weapon_skill
-  end
-
-  def ending_ranged?(game_state)
-    game_state.was_retreating && !game_state.now_retreating
-  end
-
   def determine_next_to_train(game_state, weapon_training)
-    return unless skill_done?(game_state) || !weapon_training[game_state.weapon_skill] || ending_ranged?(game_state)
+    return unless game_state.skill_done?(DRSkill.getxp(game_state.weapon_skill).to_i) || !weapon_training[game_state.weapon_skill] || game_state.ending_ranged?
```
rcuhljr: Could we move `DRSkill.getxp(game_state.weapon_skill).to_i` into game_state?

combat-trainer.lic
rpherbig: :+1:

combat-trainer.lic
rpherbig: I considered that, but I kind of want to keep GameState "pure" in the sense of not knowing that level of detail. I thought that might also make it harder to test GameState. Maybe we pass in a proc?

combat-trainer.lic
rpherbig: Thoughts on a better name? `select_dance_weapon`?

combat-trainer.lic
rcuhljr: So it knows about experience values it just doesn't know how to get them?

rpherbig: It knows that there is a number it needs to compare, but not how that number is generated. Maybe that's a meaningless distinction? @kamarcum thoughts?

Will having GameState use DRSkill make that method any harder to test?

rcuhljr: No real testing difference, for all the testing we do.

rpherbig: :+1:

PR #390, opened 2016-05-16
Game state refactor

rpherbig:
Continue moving logic into game state.

There's an unresolved review comment from #388: @rcuhljr would like a better name for `GameState::dance`. I suggested `update_dance_weapon`, but it's not great. Maybe inspiration will strike...

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -61,7 +61,7 @@ class SetupProcess
       check_stance(game_state)
     end
 
-    game_state.update_weapon_info(game_state.weapon_skill, game_state.weapon_name)
+    game_state.update_weapon_info(game_state.weapon_skill)
     false
   end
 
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -38,19 +37,19 @@ class SetupProcess
           break_summoned_weapon?(checkleft)
         end
       else
-        EquipmentManager.instance.stow_weapon(@weapon_training[game_state.weapon_skill])
+        EquipmentManager.instance.stow_weapon(game_state.weapon_training[game_state.weapon_skill])
```
rcuhljr: This is another one that feels like it should be combined into game_state. `game_state.get_weapon` or something.

rpherbig: This is the next thing I have my eye on. There's some weirdness with how `game_state.weapon_name` differs from `game_state.weapon_training[game_state.weapon_skill]` that I need to investigate first.

PR #417, opened 2016-06-03
Workorders refactor

rpherbig:
Goals:

1. Commonize logic between crafting types
1. Using 1, identify discrepancies that indicate bugs or fixes that weren't applied everywhere
1. Use these refactorings to identify more ways we can simplify logic in the future

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -121,23 +121,19 @@ class WorkOrders
         end
 
         walk_to info['stock-room']
-        fput("order #{info['stock-number']}")
-        fput("order #{info['stock-number']}")
-        fput("order #{info['stock-number']}")
+        order_item(info['stock-number'])
         fput('tap my deed')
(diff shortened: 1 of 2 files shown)


workorders.lic
```diff
@@ -207,32 +197,27 @@ class WorkOrders
...
-          order_stock(info['stock-number'])
-          combine_yarn
-        end
-      end
-    end
+    existing = if 'What were' == bput('get yarn', 'What were', 'You get')
+                 0
+               else
+                 bput('count my yarn', 'You count out \d+ yards').scan(/\d+/).first.to_i
+             end
```
rcuhljr: Really not a fan of this formatting, end doesn't seem to line up and it's hard to read.

rpherbig: Rubocop seems to have a strong opinion on that one

PR #419, opened 2016-06-04
[WIP] Streamline workorders/forge logic

rpherbig:
Goal:

* `workorders`: responsible for asking until it gets a valid recipe, calling smith, bundling, turning in
* `smith`: responsible for managing purchases for a single craft, calling forge
* `forge`: responsible for the actual forging
* Break forge up into three scripts: forge, hone, and temper
* Remove bundle/trash/stow logic from `forge`

Any feedback on these goals?


rpherbig:
Goals:
- [x] `workorders`: responsible for asking until it gets a valid recipe, calling smith, bundling, turning in
- [x] `smith`: responsible for managing purchases for a single craft, calling forge
- [ ] `forge`: responsible for the actual forging
- [ ] Break forge up into three scripts: forge, hone, and temper
- [ ] Remove bundle/trash/stow logic from `forge`
- [ ] Convert `crossing-training` to use `smith` instead of `forge`

rpherbig:
@rcuhljr @kamarcum I think this is done, please review/test. I will not be doing the two uncompleted tasks as part of this PR.

rpherbig:
Methods `empty_hands` and `dispose` don't feel right in `common-crafting`, but I'm not sure what to do with them. Thoughts?

smith.lic
```diff
@@ -12,47 +12,75 @@ class Smith
...
     material, item_name, *options = args
+    settings = get_settings(['crafting'])
+    discipline = settings['Blacksmithing']
+    recipes = settings.crafting_recipes.select { |recipe| recipe['type'] =~ /blacksmithing|weaponsmithing|armorsmithing/i }
+    recipe = recipe_lookup(recipes, item_name)
 
-    @settings = get_settings(['crafting'])
+    parts = recipe['part'] || []
+    temper = options.find { |x| x =~ /temper/i }
+    # recipe['hone'] = options.find { |x| x =~ /hone/i }
```
kamarcum: Remove these if they're not used.

common-crafting.lic
```diff
@@ -88,4 +88,26 @@ module DRCC
...
+
+  def order_item(room, item_number)
+    DRCT.walk_to(room)
+    fput("order #{item_number}")
+    fput("order #{item_number}")
+    empty_hands
+  end
+  module_function :order_item
+
+  def empty_hands
```
rpherbig: Methods `empty_hands` and `dispose` don't feel right here, but I'm not sure what to do with them. Thoughts?

crossing-training.lic
```diff
@@ -965,32 +963,10 @@ class CrossingTraining
...
-
-    EquipmentManager.instance.empty_hands
-
-    return unless trash_ingot
-    fput('get my bronze ingot')
-    dispose_trash('ingot')
+  
+  def smith(full_name, noun)
+    wait_for_script_to_complete('smith', ['bronze', full_name, 'buy'])
+    dispose_trash(noun)
```
rpherbig: Is it too risky to just `dispose_trash(checkright)` (and checkleft)? That would save us from needing to pass in the noun.

smith.lic
```diff
@@ -12,47 +12,75 @@ class Smith
...
+
     material, item_name, *options = args
+    settings = get_settings(['crafting'])
+    discipline = settings['Blacksmithing']
+    recipes = settings.crafting_recipes.select { |recipe| recipe['type'] =~ /blacksmithing|weaponsmithing|armorsmithing/i }
+    recipe = recipe_lookup(recipes, item_name)
 
-    @settings = get_settings(['crafting'])
+    parts = recipe['part'] || []
+    temper = options.find { |x| x =~ /temper/i }
```
kamarcum: It's a question of preference, but you _might_ prefer [Enumberable#grep](http://ruby-doc.org/core-2.0.0/Enumerable.html#method-i-grep) here.

`options.grep(/temper/i).first`

workorders.lic
```diff
@@ -176,29 +175,24 @@ class WorkOrders
...
-    fput("order #{item_number}")
-  end
-
   def combine_yarn
     pause
-    if checkleft =~ /yarn/ && checkright =~ /yarn/
-      fput('combine yarn with yarn')
-    end
+    return if 'What were you' == bput("get yarn from #{@bag}", 'You get', 'What were you')
+    return if 'What were you' == bput("get yarn from #{@bag}", 'You get', 'What were you')
```
kamarcum: If this dupe is intentional, you might want to comment on it.

smith.lic
rpherbig: Good call. Captured in #425.

workorders.lic
rpherbig: Done.

smith.lic
rpherbig: I like `grep` better, but this is a very common pattern. Captured in #426.

crossing-training.lic
rcuhljr: I'm not sure if we ever call dispose trash while holding something else we care about.

smith.lic
rcuhljr: I love not needing to do the `|x| x=~` crap, but I like the readability of options.find more. All we really care about is existence here so I'd rather use `.any?` instead of first. I'd be much happier with `options.grep(/temper/i).any?`

common-crafting.lic
rcuhljr: I don't have any problem with moving them into common proper, neither seems particularly crafting related.

forge.lic
```diff
@@ -75,7 +75,7 @@ class Forge
 
   def temper
     unless checkleft =~ /#{@item}/i || checkright =~ /#{@item}/i
-      echo('please hold the item to temper')
+      echo('***Please hold the item to temper.***')
```
rcuhljr: Honestly I'm fine with not needing a noun for temper/hone/balance. Just do the action on whatever you're holding. That's how sewing works for reinforcing or sealing.

rcuhljr:
I'd still like the repairing logic extracted a bit and made accessible for crossing-training.

common-crafting.lic
rpherbig: dispose can't be moved to common because it requires walk_to, which would introduce a circular dependency.

crossing-training.lic
rpherbig: Done.

smith.lic
rpherbig: Huh, `forge` does support honing. I'll update smith and the issue.

forge.lic
rpherbig: Outside the scope of this PR.

rpherbig:
> extract repair logic

Outside the scope of this PR.

forge.lic
rpherbig: There should be an effort (once this is merged) to make "smith" style scripts for sew/shape/carve and do similar refactoring there.

common.lic
```diff
@@ -102,7 +102,7 @@ module DRC
         return false unless kick_pile?
       when 'You really need to have at least one hand free to forage properly'
         echo 'WARNING: hands not emptied properly. Stowing...'
-        fput('stow right')
```
rcuhljr: This is going to explode most likely. It will stow braided vines and whatever is blocking the forage, which is bad. We only want to empty one hand.

rpherbig: Good catch. Fixed.

forge.lic
```diff
@@ -14,9 +14,9 @@
   arg7: item component/assemble noun (optional)
 
   EXAMPLES:
-  ;forge log weaponsmithing 2 sword steel sword hilt
-  ;forge trash blacksmithing 6 "shallow metal cup" bronze cup
-  ;forge stow blacksmithing 6 scratcher bronze scratcher
+  ;forge weaponsmithing 2 sword steel sword hilt
```
rcuhljr: Hilt is a lie, we don't take parts lists anymore.

rpherbig: Does that mean the arg length should only ever be 5?

forge.lic
rpherbig: Captured in #428

PR #448, opened 2016-06-17
Use gem pouch adjective when checking for gems

jandersson:
Using 'pouch' conflicts with other worn pouches


rpherbig:
Are there any other places that we use only 'pouch' that should also be changed?

rcuhljr:
Sorry this missed my radar. I can't think of any.

Issue #461, opened 2016-06-22
Attributes state does not update after training


jandersson:
After running ;train c

Try to ;train c again. We get stuck

[train]>out
You trek out.
[The Crossing, Water Sprite Way - 902]
The road curves slightly in concert with the river here, gradually merging with Bank Street.  Rising to the south you see the wood and white plaster facade of the Traders' Guildhall.  From this vantage point, it is more modest looking than you would have expected but still impeccably designed and executed.  You also see a stone cottage with a painted sign mounted on it and Brisson's Haberdashery.
Obvious paths: north, south, west.
>
--- Lich: train has exited.
>;train c
--- Lich: train active.
[train]>tdp
You have 100 TDPs.
>
[train: #<OpenStruct charisma="c">]
[train]>tdp project charisma 24
But your Charisma is already 24!
>;train c1
--- Lich: train is already running (use ;force [scriptname] if desired).
>;k train
--- Lich: train has exited.
>;train c2
--- Lich: train active.
[train]>tdp
You have 100 TDPs.
>
[train: #<OpenStruct charisma="c2">]
[train]>tdp project charisma 25
It will cost you 72 TDPs to reach 25 points in Charisma.
>
[train]>wealth

Debt:
  You owe 8 gold, 4 silver, 6 bronze, and 9 copper Kronars to the Principality of Zoluren. (8469 copper Kronars)
  You owe 1 gold, 3 silver, and 8 copper Lirums to the Province of Therengia. (1308 copper Lirums)

Wealth:
  1 silver, 6 bronze, and 2 copper Kronars (162 copper Kronars).
  No Lirums.
  No Dokoras.
>
--- Lich: go2 active.
[go2: ETA: 0:00:02 (12 rooms to move through)]
[go2]>south
You trek south.
[The Crossing, Bank Street - 901]
You stand at the entrance to the Traders' Guildhall, an understated, elegant structure with carved wooden doors, stout brick chimneys, tiled gables, pargetted decorations and leaded glass windows.  The door that leads inside is guarded by an officious concierge.  Due west along this busy street is the Crossing's Guard House, and east is the Moneylender, a financial institution for merchants great and small all over Elanthia.  
Obvious paths: north, east, west.
>
[go2]>east
You trek east.

jandersson:
I think it might be fixed by just running the info command after training

PR #462, opened 2016-06-22
Update DR infomon state after training

jandersson:
#461


rpherbig:
With your change this will happen after each attribute is trained (so if you train cha then str, it outputs `info` twice). I recommend you move this to line 50 so that it only happens once, after all training is done.

rpherbig:
Thanks for contributing!

rcuhljr:
Thanks!

PR #474, opened 2016-06-24
Stealing rework

rcuhljr:
In progress.


crossing-training.lic
```diff
@@ -745,21 +745,52 @@ class CrossingTraining
...
+      skill >= data[lower]
+    end
+  end
+
   def train_thievery
     start_script('jail-buddy') unless Script.running?('jail-buddy')
 
     EquipmentManager.instance.wear_equipment_set?('stealing')
 
+    @stealing_attempts = @settings.stealing_attempts
```
rpherbig: Needs a default value

rcuhljr: It has one, in base.yaml

rcuhljr:
Need to change data structure, I need to keep difficulty_min and difficulty_max for each range.

Issue #470, opened 2016-06-23
Pick fails at purchasing lockpicks


jandersson:
The hardcoded lockpick prices do not match the offered prices (due to charisma or price change or etc...)
Suggestion is to use the hardcoded prices as an upper bound to determine how much money to pull from the bank, but retrieve the offered value from the game instead.

```
[go2: travel time: 0:00:16]
--- Lich: go2 has exited.
[pick]>buy stout lockpick

Ragge sighs.  "Despite the rarity of this lockpick, I'm prepared to offer it to you for 237 kronars."
>
[pick]>offer 250
An offer of 250?  Are you SURE?  (OFFER it again within the next 20 seconds if you are.)
>
[pick]>put my lockpick on my lockpick ring
>
[pick]>retreat
What were you referring to?
>
You are already as far away as you can get!
>
--- Lich: go2 active.
[go2: ETA: 0:00:01 (9 rooms to move through)]
[go2]>out
```

jandersson:
Alternatively, you could just offer twice if "Are you sure" is seen, but that seems wasteful

rpherbig:
Thanks for the PR!

rpherbig:
Closed by #478

PR #478, opened 2016-06-25
Offer to buy lockpick at offered value

jandersson:
#470

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -112,8 +112,9 @@ class LockPicker
   end
 
   def buy_lockpick(type)
-    fput("buy #{type} lockpick")
-    fput("offer #{@lockpick_costs[type]}")
+    response = dothis("buy #{type} lockpick", /prepared to/)
+    /prepared to offer it to you for (?<amount>.*) kronars/ =~ response
+    fput("offer #{amount}")
     fput('put my lockpick on my lockpick ring')
   end
 
--- a/pick.lic
+++ b/pick.lic
@@ -112,7 +112,7 @@ class LockPicker
   end
 
   def buy_lockpick(type)
-    response = dothis("buy #{type} lockpick", /prepared to/)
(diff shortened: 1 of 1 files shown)


pick.lic
```diff
@@ -112,8 +112,9 @@ class LockPicker
   end
 
   def buy_lockpick(type)
-    fput("buy #{type} lockpick")
-    fput("offer #{@lockpick_costs[type]}")
+    response = dothis("buy #{type} lockpick", /prepared to/)
```
rpherbig: We've created `bput` (in `common.lic`) as a better version of `put`, but I haven't seen `dothis` before. At a glance, `bput` checks for different error cases (I assume the checks in `dothis` are for Gemstone...). I also prefer the error messaging of `bput`. Thoughts @rcuhljr or @kamarcum ?

pick.lic
```diff
@@ -112,8 +112,9 @@ class LockPicker
   end
 
   def buy_lockpick(type)
-    fput("buy #{type} lockpick")
-    fput("offer #{@lockpick_costs[type]}")
```
rpherbig: Can you delete the cost hash on line 61 since it is no longer used?

pick.lic
rcuhljr: I have a preference for bput since it's DR specific errors, has a timeout (although there's a dothistimeout), and only returns the matched substring sent, not the entire message. Switch to bput here would require expanding the match message to the regex below, however that does make it simpler to reuse the regex in both places.

pick.lic
jandersson: They're still used in refill_ring() on line 98 where we pull money out of the bank to pay for X lockpicks. I think it might be useful to have them there as an upper bound on the cost, since we don't know what the discount will be

rpherbig: We could use 500 * count as an upper bound.

jandersson: Its also used to error check if the specified lockpick type is valid on lines 88-91

PR #521, opened 2016-07-13
Create common-items

rpherbig:
My goal here is to find repeated logic and implement some helpers. I want us to be able to write scripts at a higher level of abstraction instead of reinventing the wheel.

I noticed a bunch of item related logic (ordering and buying) and cleaned up some of it previously, but it meant that scripts like `mining-buddy` were including `common-crafting`, which seems a bit odd.

That item logic really isn't crafting-specific, so I'm going to pull a few methods out of common and common-crafting.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -18,8 +18,15 @@ module DRCI
       amount = Regexp.last_match(1)
     when /for a mere (.*) kronars/
       amount = Regexp.last_match(1)
-  end
+    end
 
     fput("offer #{amount}") if amount
+  end
+
+  def order_item(room, item_number)
+    DRCT.walk_to(room)
+
+    fput("order #{item_number}")
+    fput("order #{item_number}")
   end
 end
--- a/common-items.lic
+++ b/common-items.lic
@@ -1,0 +1,25 @@
+# quiet
+=begin
+  Documentation: https://elanthipedia.play.net/mediawiki/index.php/Lich_script_development#common-items
+=end
+
+custom_require.call(%w(common common-travel))
+
+module DRCI
+  module_function
+
+  def buy_item(room, item)
+    DRCT.walk_to(room)
+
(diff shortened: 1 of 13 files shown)


common-items.lic
```diff
@@ -0,0 +1,25 @@
...
+    DRCT.walk_to(room)
+
+    case DRC.bput("buy #{item}", /prepared to offer it to you for (.*) kronars/, /Let me but ask the humble sum of (.*) coins/, /for a mere (.*) kronars/, 'You decide to purchase')
+    when /prepared to offer it to you for (.*) kronars/
+      amount = Regexp.last_match(1)
+    when /Let me but ask the humble sum of (.*) coins/
+      amount = Regexp.last_match(1)
+    when /for a mere (.*) kronars/
+      amount = Regexp.last_match(1)
+  end
```
kamarcum: Add a couple of spaces to the front of this line.

rpherbig:
I could be convinced to move a few other things over from `common`... For example: class Item, the rummage methods, stow_hands. Thoughts?

common-items.lic
```diff
@@ -0,0 +1,55 @@
...
+
+  def dispose_trash(item)
+    return if item.nil?
+    trashcan = DRRoom.room_objs
+                     .map { |long_name| get_noun(long_name) }
+                     .find { |obj| $TRASH_STORAGE.include?(obj) }
+    if trashcan == 'gloop'
+      trashcan = 'bucket' if DRRoom.room_objs.include? 'bucket of viscous gloop'
+      trashcan = 'cauldron' if DRRoom.room_objs.include? 'small bubbling cauldron of viscous gloop'
```
rcuhljr: we could `.find` the room object with gloop and use a regex to pull out NOUN of viscous gloop.

rcuhljr:
Looks good.

common-items.lic
rpherbig: Out of scope for this PR. Is it worth making a new issue for it?

PR #530, opened 2016-07-16
Create common-arcana for casting and such

rpherbig:
I think this is a good place to move duplicated cast/magic/etc related logic.

Maybe it's even a first step towards a DSL so we can focus on the what, not the how.

Do you know of other duplicate logic @kamarcum @rcuhljr ?


rpherbig:
This is a comment to test our Slack integration.

Issue #423, opened 2016-06-07
Trim down starter yamls, name them by class


rcuhljr:
Lets stop using our active character yamls as seeds. Remove thorny things like crafting, gem pouches, paired down weapon and gear lists, etc.

rpherbig:
Let's name them obvious things like `SampleBarbarian-setup.yaml`, that way they're colocated in the directory.

rpherbig:
Broke out the rest of the classes into separate issues, this was getting too big.

PR #558, opened 2016-07-30
Moon mage sample YAML

jandersson:
As requested, heres a sample file for starting moon mage. Let me know what you think. I tested it out on a F2P character and its got him circling. Also ignore my other commits. But, if they're an issue, I can revert and do another pull request.


profiles/SampleMoonMage-setup.yaml
```diff
@@ -0,0 +1,129 @@
...
+# 3 Lores: Appraisal, Tactics, Mechanical Lore (Does not include Scholarship)
+# 2 Additional: Astrology, Scholarship
+
+# You will need a runestone in your inventory to train sorcery and arcana. Grab one from any donation shelf.
+# You will need to add more magic and survival skills depending on your circle
+# However, you will probably end up training them all since they each have very useful spells
+# The other Magic skills are: Targeted Magic, Warding, Debilitation, Utility
+
+# Hunting settings
+training_manager_hunting_priority: true
```
rpherbig: Since MMs are weapon/armor tert, would it be better to prioritize noncombats?

rcuhljr: Not really, weapons will lock quick and take awhile to buffer down, I might make my stop ons and priority skills shield usage and one of my weapon skills. percept is kind of hit or miss for locking in combat, as is attunement.

jandersson: Could swap out Evasion for Outdoorsmanship and Tactics for Performance

jandersson: Need something to replace Skinning... thievery?

rpherbig: Locksmithing?

rpherbig:
Cross-linking to #423

profiles/SampleMoonMage-setup.yaml
```diff
@@ -0,0 +1,128 @@
...
+  Heavy Thrown: throwing hammer
+  Large Blunt: throwing hammer
+use_stealth_attacks: true
+
+# Non-combat settings
+exp_timers:
+  Astrology:
+    cooldown: 200
+
+crossing_training:
```
rpherbig: Attunement should be in this list (and `exp_timers`).

profiles/SampleMoonMage-setup.yaml
jandersson: I kind of like having the combats in there personally, eventually you need to train TM and/or Debil anyway to get 7 magics. I think having the weapons and shield would be pretty normal, its similar to what i do on Crann. Attunement gets locked after about three uses of perc mana, I think tactics took the longest.

profiles/SampleMoonMage-setup.yaml
jandersson: Just in exp_timers? or also in crossing_training?

profiles/SampleMoonMage-setup.yaml
jandersson: Locksmithing could be nice. zone: louts then?

jandersson:
Alright it looks like ive made a mess... bear with me while I try to remove these files i accidentally added

jandersson:
Adding attunement to crossing_train causes the character to powerwalk which isn't good for training. Going to remove it from the training list and just have it get trained in combat with perc mana (which works really well at low level). Is the attunement exp_timer still needed?

rpherbig:
That's strange, attunement has a special case for MM:
```
    if DRStats.moon_mage?
      fput('perc mana')
      pause 15
      waitrt?
      return
    end
```

Was `drinfomon` running when you tried that?

jandersson:
Now, its working as you say

rpherbig:
A few thoughts:
* I took a circle 1 MM to Kssarh and checked the circle reqs - there was no mention of a 4th magic. I'll test to see if epedia is wrong.
* A circle 1 MM can't charge an armband, so I suggest you add `cambrinth: ring` and `cambrinth_cap: 5`
* What would you think about adding `favor_god: Meraud` and `favor_goal: 6`?
* There are some one-time character setup things to be done, I created an issue to script it. I'd appreciate your thoughts on that one (see #562).

jandersson:
The 4th magic shows up at circle two, not sure why circle one gets the free pass (perhaps its something to do with the starter ranks in magic you're given). 

I'll add the cambrinth and favors.

rpherbig:
Ah, that makes more sense. In that case, would it be simpler (not necessarily optimal) to add a comment recommending Calm (Debil) or Telekenetic Throw (TM) as the spell choice at circle 2 instead of Sorcery?

jandersson:
Yea, great point. I felt the sorcery bit was kind of unusual...

profiles/SampleMoonMage-setup.yaml
```diff
@@ -0,0 +1,136 @@
+---
+# https://elanthipedia.play.net/mediawiki/index.php/Moon_mage#Circle_Requirements
+# 4 Magics: Augmentation, Attunement, Arcana, Targeted Magic/Debilitation (Does not include Lunar Magic or Astrology)
+# 4 Survivals: Perception, Athletics, Evasion, Skinning (Does not include Thievery)
+# 3 Lores: Appraisal, Tactics, Mechanical Lore (Does not include Scholarship)
+# 2 Additional: Astrology, Scholarship
+
+# It is recommended to get a target or debilitation spell for your second circle. (e.g. Dinazen Olkar or Calm)
```
rpherbig: You might consider adding the circle 2 spell in the yaml, but commented out. You can then have a note here that says to remove the #s after learning it at circle 2.

profiles/SampleMoonMage-setup.yaml
```diff
@@ -0,0 +1,136 @@
...
+- Perception
+
+safe_room: 19162
+
+training_spells:
+  Augmentation:
+    abbrev: shadows
+    symbiosis: false
+held_cambrinth: true
+cambrinth: cambrinth ring
```
rpherbig: You'll also want to set `cambrinth_cap: 5`.

profiles/SampleMoonMage-setup.yaml
```diff
@@ -0,0 +1,136 @@
...
+    cooldown: 130
+
+crossing_training:
+- Astrology
+- Augmentation
+- Appraisal
+- Athletics
+- Scholarship
+- Mechanical Lore
+- Sorcery
```
rpherbig: Should this line be removed?

profiles/SampleMoonMage-setup.yaml
jandersson: I think rings are just 4 no?

profiles/SampleMoonMage-setup.yaml
jandersson: Yea if instead we use cambrinth to train arcana. Shall I add cambrinth: 1 to the training spell?

rpherbig: I think that is a good plan. I tested on my circle 1 MM and while there is the occasional backfire, 1 mana prep + 1 mana cambrinth works most of the time.

jandersson: What does that look like?

I have this, and it is not charging my cambrinth:

```
training_spells:
  Augmentation:
    abbrev: shadows
    mana: 1
    cambrinth:
      - 1
```

rcuhljr: It's gonna require a change to crossing-training, you can't specify cambrinth charges in training spells.

rpherbig: For the time being, add cambrinth use to combat-trainer's buff section and that'll be good enough to get Arcana moving.

profiles/SampleMoonMage-setup.yaml
```diff
@@ -0,0 +1,161 @@
+---
+# https://elanthipedia.play.net/mediawiki/index.php/Moon_mage#Circle_Requirements
+# 4 Magics: Augmentation, Attunement, Arcana, Targeted Magic/Debilitation (Does not include Lunar Magic or Astrology)
+# 4 Survivals: Perception, Athletics, Evasion, Skinning (Does not include Thievery)
+# 3 Lores: Appraisal, Tactics, Mechanical Lore (Does not include Scholarship)
+# 2 Additional: Astrology, Scholarship
+
+# It is recommended to get a target or debilitation spell for your second circle. (e.g. Dinazen Olkar or Calm)
+# Please see the 'offensive_spells:' setting below for a commented (unused) example. Uncomment to begin usage (remove #'s).
+# You will need a runestone in your inventory to train arcana. Grab one from any donation shelf.
```
rpherbig: Can this line (about getting a runestone) be removed?

rpherbig:
This is fantastic, thank you for setting it up @jandersson !

PR #559, opened 2016-07-31
Updated matches for wearing a ring

UFTimmy:
I'm probably using the method wrong, but it's the best way I've found to wear my cambrinth items, given that it's part of my equipment set.

The text from the game is: You slide a twisted cambrinth ring onto your finger

### DIFF
--- a/equipmanager.lic
+++ b/equipmanager.lic
@@ -342,7 +342,7 @@ class EquipmentManager
     if weapon.wield
       stow_helper("sheath my #{weapon.short_name}", 'Sheathing')
     elsif weapon.worn
-      stow_helper("wear my #{weapon.short_name}", weapon.name, 'You sling', 'You attach', 'You .* unload', 'You place')
+      stow_helper("wear my #{weapon.short_name}", weapon.name, 'You sling', 'You attach', 'You .* unload', 'You place', 'You slide')
     elsif !weapon.tie_to.nil?
       stow_helper("tie my #{weapon.short_name} to #{weapon.tie_to}", weapon.name, 'You attach')
     elsif weapon.transforms_to


rpherbig:
It's not necessarily a wrong use of that method, but it is not something I expected. Could you give me an example of how you're using it with your cambrinth ring in a script?

UFTimmy:
Yup. I just basically want a method that returns a piece of gear from your hands to where it goes. There doesn't appear to be one for generic items. wear_item? will first take it out and then wear it.

Here's how I'm using it, the last line here is the line.

```ruby
    if Arcana.get_rank < 205
      if checkright && checkleft
        echo 'Would charge cambrinth, but your hands are full, and you need to hold the cambrinth'
        return
      end
      echo 'Getting the cambrinth' if @debug
      item = EquipmentManager.instance.item_by_desc(Profile.get('cambrinth'))
      echo "Item instance: #{item}" if @debug
      return unless EquipmentManager.instance.get_item?(item)
    end

    spell_info['cambrinth'].each do |amount|
      bput("charge my #{Profile.get('cambrinth')} #{amount}", Profile.get('charge_messages'))
      pause
      waitrt?
    end

    bput("invoke my #{Profile.get('cambrinth')} spell", Profile.get('invoke_messages'))
    pause
    waitrt?

    EquipmentManager.instance.stow_weapon("#{item.adjective} #{item.name}") unless item.nil?
```

Basically, get_item? works to get an item of any time, but I don't see the inverse. Only stow_weapon.

rpherbig:
I think `common-arcana` has most of the helpers you are looking for: `find_cambrinth`, `stow_cambrinth`, and `charge_and_invoke`.

UFTimmy:
Thanks. I didn't realize common-arcana was there. How do you feel about a feature to be able to invoke spell, rather than just plain ole invoke? It takes the dedicated cambrinth use feat, but even if you don't have it, the invoke works as normal. I hate it when a cyclic running pulls mana from my cambrinth. It could be a configuration setting in profile.

Yes, profile is just a wrapper for get_settings. I have broken my scripts up into a brazillion different files/classes, and I didn't want to call get_settings on all of them. So Profile exists as a class to enable access to the profile from anywhere. It's on my to-do list to refactor it as an OpenStruct once I saw what that was.

rpherbig:
I'm fine with that - it seems pretty useful for those using cyclics.

PR #565, opened 2016-07-31
Added balance (combat) to drinfomon

UFTimmy:
Useful if you need to regain balance in combat.

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -31,7 +31,7 @@ class DRStats
   @@favors ||= 0
   @@tdps ||= 0
   @@encumbrance ||= 'unknown'
-  @@balance ||= 9
+  @@balance ||= 8
   @@health ||= 100
   @@mana ||= 100
   @@fatigue ||= 100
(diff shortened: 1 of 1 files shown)


drinfomon.lic
```diff
@@ -699,6 +708,9 @@ while line = script.gets
         end
     elsif line =~ /^Circle: (\d+)/
       DRStats.circle = Regexp.last_match(1).to_i
+    elsif line =~ /^\[You're(.*\,)?(.*) (and|with).*\]$/
+      balance_map = ['completely imbalanced', 'hopelessly unbalanced', 'extremely imbalanced', 'very badly balanced', 'badly balanced', 'somewhat off balance', 'off balance', 'slightly off balance', 'solidly balanced', 'nimbly balanced', 'adeptly balanced', 'incredibly balanced']
+      DRStats.balance = balance_map.index(Regexp.last_match(2).strip) + 1
```
rcuhljr: Most of the appraisal scales go from 0->#, any reason to start balance at 1 instead of zero? Also I think there might be an alternate messaging form when you kill an enemy where you only get balance and no positional element, probably a pretty niche concern.

UFTimmy:
I prefer zero index, but I somehow thought everyone else liked 1 indexing.

Changed it to be zero indexed, and updated the regex to account for the balance message when not facing anything. Thanks for catching that.

PR #587, opened 2016-08-07
Go to safe_room at the end of hunting-buddy

kamarcum:
It does what you think it does.


hunting-buddy.lic
```diff
@@ -39,7 +39,7 @@ class HuntingBuddy
       fput('release hyh') if DRSpells.active_spells['Hydra Hex']
       fput('release ee') if DRSpells.active_spells['Electrostatic Eddy']
     end
-    walk_to(793)
+    walk_to(@settings.safe_room)
```
rpherbig: Line 49 should also be changed.

rpherbig:
I'd go ahead and update `pick` in this PR.

Do you think line 32 of `crossing-training` needs to change? It provides a default value for `settings.outdoor_room`.

kamarcum:
> I'd go ahead and update pick in this PR.

I find that less compelling at the moment. The part of `pick` that goes to 793 is also buying lockpicks from a shop in crossing. So that chunk of logic would need more work to become city-agnostic.

hunting-buddy.lic
kamarcum: Ah good find.

rpherbig:
> I find that less compelling at the moment. The part of pick that goes to 793 is also buying lockpicks from a shop in crossing. So that chunk of logic would need more work to become city-agnostic.

It's also one less thing for us to forget when we put in that work later.

rcuhljr:
The only thing that pick needed to do was step outside of the store when done buying. just replace it with a move('out') and be done.

PR #588, opened 2016-08-08
Added regexs for Fallen Barb

UFTimmy:
I'm not sure if it's because he's a barbarian, or because of the Fallen, but the XML sent across for my Barbarian in the Fallen is a different format than the regex here.

Here's the raw XML sent:

```
<clearStream id="percWindow"/>
<pushStream id="percWindow"/> Panther (65 roisaen)
<popStream/><pushStream id="percWindow"/> Turtle (65 roisaen)
<popStream/><pushStream id="percWindow"/> Monkey (65 roisaen)
<popStream/><pushStream id="percWindow"/> Piranha (65 roisaen)
<popStream/><prompt time="1470622380">&gt;</prompt>
```

It wouldn't surprise me if the XML sent doesn't match Prime, sometimes they forget to move changes over here. Does spellmonitor work in prime for barb abilities?


rpherbig:
I don't have Powermonger, so I've been unable to test barb abilities with spellmonitor. @rcuhljr thoughts?

PR #593, opened 2016-08-09
Use more descriptive names for built-in lich functions

rpherbig:
Lich offers quite a few aliases for its built-in functions. I think many are more descriptive than the original names we've been using.

I changed the ones I'm confident in. There are a few other options which I am undecided on (feedback welcome!):
* `checkmana` -> `mana`
* `checkhealth` -> `health`
* `checkname` -> `myname?`
* `checkright` -> `righthand`
* `checkleft` -> `lefthand`

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -13,7 +13,7 @@ loop do
   line = script.gets?
   pause 0.05 unless line
   fput(%w(tdp time age).sample) if line =~ /you have been idle too long/i
-  if checkdead
+  if dead?
     echo '*' * 30
     echo 'Afk - detected death departing in 3 minutes'
     echo '*' * 30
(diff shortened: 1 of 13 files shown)


rcuhljr:
If you're doing hands you should make the wrapper user GameObj to get us noun+adj so we can stop using glance everywhere.

rcuhljr:
Also myname? is not a predicate.

rpherbig:
The aliases are built into lich, I didn't pick `myname?`. The non-predicateness of it is why I didn't make the change - I would have if it was simply `myname`.

rcuhljr:
I misunderstood this, I thought you were making new aliases.

rpherbig:
@rcuhljr Do you know if `checkhealth` is different than `DRStats.health`?

rcuhljr:
They should be the same, I'm pretty sure they both populate off the healthbar xml data.

PR #595, opened 2016-08-10
LootProcess Timer

jandersson:
Return from `dispose_body` if a time delta is below threshold. Timer is reset upon looting

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -263,7 +263,6 @@ class LootProcess
       @loot_timer = Time.now
       waitrt?
     end
-    pause 0.2 until DRRoom.dead_npcs.empty?
   end
 
   # partially completed, only works for a single ritual.
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -183,6 +183,9 @@ class LootProcess
 
     @gem_pouch_adjective = settings.gem_pouch_adjective
     echo("  @gem_pouch_adjective: #{@gem_pouch_adjective}") if $debug_mode_ct
+
+    @loot_timer = Time.now - 5
+    echo("  @loot_timer: #{@loot_timer}") if $debug_mode_ct
   end
 
   def execute(game_state)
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -481,8 +485,6 @@ class SpellProcess
     @osrel_no_harness = settings.osrel_no_harness
     echo("  @osrel_no_harness: #{settings.osrel_no_harness}") if $debug_mode_ct
 
-    @cast_only_to_train = settings.cast_only_to_train
```
jandersson: I'm a little confused about why these changes are still showing up. I've fetched and merged with the remote 'upstream' which is your master, and then created this branch

rcuhljr: Yeah, the timer commit removes these lines.

jandersson: Thanks, should be fixed now

combat-trainer.lic
```diff
@@ -255,10 +260,10 @@ class LootProcess
       check_skinning(DRRoom.dead_npcs.first, game_state)
     end
     while 'and get ready to search it' == bput('loot', 'You search', 'I could not find what you were referring to', 'and get ready to search it')
+      @loot_timer = Time.now
```
rpherbig: I think this needs to be moved outside of the while loop. If you loot and get 'You search', the while exits without running the code inside, so `@loot_timer` never gets updated.

jandersson:
Do some mobs decay at different rates than others?

combat-trainer.lic
jandersson: The idea was to just update the timer if a looting is actually performed to act as a stopgap between the first check in `dispose_body` being `return if DRRoom.dead_npcs.empty?`

rpherbig: If a loot occurred successfully, the while loop exits and `@loot_timer` is never updated.

If loot occurs but fails, the while loop pauses and retries. In that case I do not think we would want to update `@loot_timer`.

jandersson: ah ha! OK rookie mistake

rcuhljr:
Looks good, although we could probably safely use a lower loot timer, and at some point I wanted to just make this logic smarter and work off `DRRoom.dead_npcs.size`.

rpherbig:
I ran this overnight and 5 seconds felt about right. I think we could lower it to 3-4 seconds without problem, but in practice you generally end up attacking or using a training ability between loots, which take long enough.

Issue #596, opened 2016-08-11
Replace checkleft and checkright with GameObj.right_hand and GameObj.left_hand


rpherbig:
We'll need a wrapper that converts 'Empty' to nil

We should be able to get rid of a lot of the glances we do.

PR #603, opened 2016-08-12
Account for stowing braided vines/grass/ropes

UFTimmy:
When you try to stow a braided item, you get a message like this:

The braided vines is too long
The braided heavy rope is too long

Which hung the stow_hands method. So I added conditions to bput to catch that, and then dispose of the braided thing.

I'm not sold on the implementation, so if you don't like it, I can convert this to an issue.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -192,8 +192,14 @@ module DRC
 
   def stow_hands
     pause 1
-    bput('stow right', 'You put') if checkright
-    bput('stow left', 'You put') if checkleft
+    stow_hand('right') if checkright
+    stow_hand('left') if checkleft
+  end
+
+  def stow_hand(hand)
+    braid_regex = /The braided (.+) is too long/
+    result = bput("stow #{hand}", 'You put', braid_regex)
+    DRCI.dispose_trash(DRC.get_noun(Regexp.last_match(1))) if braid_regex.match(result)
   end
 
   def bind_wound(part, person = 'my')
--- a/common-items.lic
+++ b/common-items.lic
@@ -40,7 +40,7 @@ module DRCI
   def dispose_trash(item)
     return if item.nil?
     trashcans = DRRoom.room_objs
-                      .map { |long_name| get_noun(long_name) }
+                      .map { |long_name| DRC.get_noun(long_name) }
                       .select { |obj| $TRASH_STORAGE.include?(obj) }
 
     trashcans.each do |trashcan|


common.lic
```diff
@@ -192,8 +192,14 @@ module DRC
...
-    bput('stow right', 'You put') if checkright
-    bput('stow left', 'You put') if checkleft
+    stow_hand('right') if checkright
+    stow_hand('left') if checkleft
+  end
+
+  def stow_hand(hand)
+    braid_regex = /The braided (.+) is too long/
+    result = bput("stow #{hand}", 'You put', braid_regex)
+    DRCI.dispose_trash(DRC.get_noun(Regexp.last_match(1))) if braid_regex.match(result)
```
rpherbig: Have you created a circular dependency by using `common-items` (the dispose_trash method) here in `common`?

You do not need to prefix get_noun with DRC since it's being used in `common`.

rpherbig:
Related to this, we really need to take a look at item management as a whole (see #596).

common.lic
UFTimmy: I did a test with a simple script that was just this:

```
custom_require.call(%w(dependency equipmanager drinfomon events spellmonitor common common-arcana common-items common-travel common-summoning common-money common-crafting textsubs roomnumbers))

#DRC.forage?('grass')
EquipmentManager.instance.empty_hands
```
And it errored without the DRC on the dispose_trash call.

As far as the circular dependency, I can't say. I don't know enough to fully understand dependencies in Ruby.

rpherbig: The `DRCI.dispose_trash` is still necessary. I was referring to `DRC.get_noun`. Try removing `DRC` and let me know if that works.

common.lic
```diff
@@ -192,8 +192,14 @@ module DRC
 
   def stow_hands
     pause 1
-    bput('stow right', 'You put') if checkright
-    bput('stow left', 'You put') if checkleft
+    stow_hand('right') if checkright
```
rpherbig: I wonder if a lot of this logic could be simplified if you called `stow_hand(checkright)` (and checkleft). Then add a nil check to return early to `stow_hand`. That way you pass in the noun and don't need to try to parse it.

UFTimmy: I think you still need to pass in 'left' or 'right'. Sometimes you may have two of the same thing in your hand, like two ingots if you use the cutter.

If you just pass in the noun, you may stow the wrong one. So you'd need to pass in left or right, so you could stow left|right.

rpherbig: In this case we're stowing both hands.

common.lic
rcuhljr: I'd suggest moving stow_hands to common-items to avoid potential problems.

PR #613, opened 2016-08-15
add new hunting construct hunting zone

jandersson:
Update base.yaml to include newly mapped construct hunting zone on the STR


jandersson:
I just realized `endrus_serpent` should have been pluralized as all the other zones are

rpherbig:
Fixed.

PR #623, opened 2016-08-17
Common healing

jandersson:
I've moved the wound check method from healme and changed its name to reflect what its doing (perceive health) and cherry-picked out my check wounds from necro_heal. The latter method was renamed to check_health to reflect what its doing. I tested the slightly modified healme.lic on my empath and it seems to work fine.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/mediawiki/index.php/Lich_script_development#combat-trainer
 =end
 
-custom_require.call(%w(common common-arcana common-items common-summoning common-travel drinfomon equipmanager events spellmonitor))
+custom_require.call(%w(common common-arcana common-healing common-items common-summoning common-travel drinfomon equipmanager events spellmonitor))
 
 class SetupProcess
   include DRC
@@ -152,6 +152,7 @@ class LootProcess
   include DRC
   include DRCI
   include DRCS
+  include DRCH
 
   def initialize(settings)
     echo('New LootProcess') if $debug_mode_ct
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/mediawiki/index.php/Lich_script_development#combat-trainer
 =end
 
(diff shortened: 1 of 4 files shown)


jandersson:
Some things:
* the hash severity_to_text still needs to be populated. I've been using https://elanthipedia.play.net/mediawiki/index.php/Wounds as a reference.
* check_health does not map nerve damage to skin/nerve as the word skin/nerve does not show up in some of the nerve damage text

combat-trainer.lic
```diff
@@ -152,6 +152,7 @@ class LootProcess
   include DRC
   include DRCI
   include DRCS
+  include DRCH
```
rpherbig: Is this necessary as part of this PR, or is this a line leftover from the necro healing changes?

jandersson: Ah, that is definitely a leftover.

healme.lic
```diff
@@ -133,7 +102,7 @@ class HealMe
     if @settings.empath_healing['HEAL'] && skip_parts.nil?
       spam_heal(@settings.empath_healing['HEAL'].first, @settings.empath_healing['HEAL'][1..-1])
     else
-      wounds = check_wounds
+      wounds = DRCH.check_perc_health
```
rpherbig: You should be able to get away with removing `DRCH` here, since it is included above.

healme.lic
```diff
@@ -13,23 +13,9 @@ custom_require.call(%w(common common-arcana events))
 class HealMe
   include DRC
   include DRCA
+  include DRCH
```
rpherbig: You'll want to add `common-healing` to the `custom_require` call on line 11.

rpherbig:
What do you need to get the rest of the messaging?

jandersson:
I just need to sit down and do it. All the messaging is on elanthipedia. I also need to figure out how to deal with match text that doesn't have the actual part name in it anywhere, like `minor twitching` for nerve/skin damage

rpherbig:
This looks reasonable to me. What else do you need to do before we merge it @jandersson ?

jandersson:
I can't think of anything, I think that it needs to be tested but for the higher level wound cases. That might be easier when more people are using it

PR #645, opened 2016-08-21
Combat-trainer clutter management

drboldwin:


drboldwin:
Hmm was gonna submit those seperately - base.yaml I just added a dump_junk: false - but take a look at it because I am seeing an odd waggle addition/subtraction that I didn't do.  Maybe this was updated between me forking and submitting?

rpherbig:
The `waggle_sets` line looks like a newline was added to the end of the file. Don't worry about it.

combat-trainer.lic
```diff
@@ -197,9 +197,18 @@ class LootProcess
 
     @loot_specials = settings.loot_specials
     echo("  @loot_specials: #{@loot_specials}") if $debug_mode_ct
+    @dump_junk = settings.dump_junk
+    @dump_delay = 300 # in seconds
```
rpherbig: Do we need the delay stored as a variable if it is only used in one place? @rcuhljr @kamarcum thoughts?

The same thing probably applies to `@loot_delay` on line 191.

combat-trainer.lic
```diff
@@ -197,9 +197,18 @@ class LootProcess
 
     @loot_specials = settings.loot_specials
     echo("  @loot_specials: #{@loot_specials}") if $debug_mode_ct
+    @dump_junk = settings.dump_junk
```
rpherbig: Please add an echo that outputs the state of this setting (see line 199) for an example.

combat-trainer.lic
```diff
@@ -197,17 +197,16 @@ class LootProcess
 
     @loot_specials = settings.loot_specials
     echo("  @loot_specials: #{@loot_specials}") if $debug_mode_ct
+    
     @dump_junk = settings.dump_junk
-    @dump_delay = 300 # in seconds
-    @dump_timer = Time.now - @dump_delay
+    echo("  @dump_junk: #{@dump_junk}") if $debug_mode_ct
```
drboldwin: Believe this is what you wanted.

combat-trainer.lic
```diff
@@ -197,17 +197,16 @@ class LootProcess
...
+    @dump_timer = Time.now - 300
```
drboldwin: got rid of dump_delay and just put 300 in there

combat-trainer.lic
```diff
@@ -197,17 +197,16 @@ class LootProcess
...
   end
 
   def execute(game_state)
-    if (Time.now - @dump_timer > @dump_delay) && @dump_junk
-      if DRRoom.room_objs.count >= 13
-        fput "DUMP JUNK"
-        @dump_timer = Time.now
-      end
+    if (Time.now - @dump_timer > 300) && @dump_junk && DRRoom.room_objs.count >= 13
```
drboldwin: got rid of the embedded if statement and made it one large one.  Also adjusted for the lack of dump_delay

PR #655, opened 2016-08-23
Added lichbot room as first line of healing.

ohbeanz:
Checks to see if Xaji is in room 1230 before going to Dokt.


ohbeanz:
Guess all lines were marked as new/deleted because I uploaded a new file. Really just added the "check_lichbot" function and new room numbers.

safe-room.lic
```diff
@@ -1,158 +1,174 @@
...
+    until Flags['healthy'] || Flags['moved']
+      lie_down
+      pause 1
+      check_idle
+    end
+    wait_for_healing if Flags['moved']
+  end
+
+  def check_lichbot
+	walk_to 1230
```
rpherbig: In general, I'm not a fan of hardcoding room numbers that could change (i.e. Dokt will never move rooms, but Xaji might).

safe-room.lic
```diff
@@ -1,158 +1,174 @@
...
+	unless Room.current.id == 1230
+      beep
+      beep
+      echo 'ERROR Navigating to safe room'
+      return
+    end
+
+	unless DRRoom.pcs.include? 'Xaji'
```
rpherbig: Likewise, Xaji might not be the PC empath on duty today - maybe Selanas takes over when Xaji isn't around.

rpherbig:
Is the goal of this PR is to check for a PC empath before going to Dokt?

If so, I would suggest you instead change the logic around `use_pc_empath?` to check for a PC empath, and then go to Dokt if one is not in the room.

safe-room.lic
ohbeanz: True. I can have it pull from settings. Have a list of potential PC empaths you could be healed by.

safe-room.lic
ohbeanz: Same thing here... have an option setting for lichbot_room

rpherbig: How would that new setting differ from `settings.safe_room_id`?

safe-room.lic
rpherbig: How would that new setting differ from `settings.safe_room_empath`?

(Aside from `settings.safe_room_empath` being one name. We could convert it to a list)

safe-room.lic
ohbeanz: I suppose in this case it would not differ... problem is the few definitions of 'safe_room:'
1. "Safe room" where braiding, kicking rocks, etc. happens (934, by default)
2. "Safe room" where Dokt is taking up residence (6218)
3. "Safe room" where the lich bot will be (Currently 1230, but could be anything else). 

The labeling and it having slightly different context doesn't make it very intuitive.

rpherbig: That's fair, having a setting called `safe_room` and one called `safe_room_id` is unfortunate. However, no scripts reference lichbot, so I'd be hesitant to make a setting for that until we need it.

What we really need to do is rename one of those settings to something more sensible.

ohbeanz: Yeah. Then, the script's name in general isn't very fitting. It has a singular function -- and that is to heal the PC. "Safe room" seems fitting for where the player will kick, braid, and pick boxes.

"heal-room," (and thus 'heal_room_id') or something to that effect, would seem like the more appropriate script name for this... but, that may be something to change later, if at all.

rpherbig: Agreed. But it's an unfortunate reality of the infrastructure that renaming a script is error-prone.

Would you be up for keeping the focus of this PR on the "do something if PC empath isn't around" logic, and submit a new PR/issue for the script/settings rename?

rcuhljr:
It'd be nice to just be able to ask the empath if they're working instead of having to travel there perhaps.

scroll-search.lic
```diff
@@ -0,0 +1,459 @@
...
+end
+
+class ScrollSearch
+  include DRC
+  include DRCI
+
+  def initialize
+
+    # Variables
+    $scroll_count = 0
```
rpherbig: To avoid collisions with other scripts, I recommend these be instance variables (`@`) instead of globals (`$`).

scroll-search.lic
```diff
@@ -0,0 +1,459 @@
...
+    [
+      [
+        { name: 'action', regex: /(^dis.*|^sea.*)/i, optional: false, description: 'Whether to SEARCH | DISPLAY your container for spell scrolls. ' },
+        { name: 'container', regex: /\w+/i, optional: true, description: 'Searches given container and labels any unlabeled scrolls. ' }
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+
+    if args.to_h.empty?
```
rpherbig: This should happen naturally as part of the parse_args logic. Have you been able to trigger this `if` block when running your script?

scroll-search.lic
```diff
@@ -0,0 +1,459 @@
...
+      display_args(arg_definitions)
+      return
+    end
+
+    pause
+
+    case args.action
+    when /^sea.*/i
```
rpherbig: Rather than doing regex matching, you might want to consider two optional args (search and display, similar to how you did container), then you could act on `args.search` or `args.display`. The `if args.to_h.empty?` block above could then be `unless args.search || args.display` and display your error message.

scroll-search.lic
```diff
@@ -0,0 +1,459 @@
...
+        #$container = args.container
+        search_container(args.container)
+      end
+    when /^dis.*/i
+      if $owned_scrolls.length <= 0
+        echo("*** NO SCROLLS HAVE BEEN ADDED! SEARCH YOUR CONTAINERS FIRST!***")
+        return
+      end
+
+      echo("Scroll scout is: #{$owned_scrolls.length}") if $debug_mode
```
rpherbig: Should this be 'Scroll count'?

scroll-search.lic
```diff
@@ -0,0 +1,459 @@
...
+      display_scrolls()
+    else
+      echo("***INVALID ACTION***")
+    end
+  end
+
+  def search_container(container)
+    # return if invalid container
+
+    if nil == exists?(container)
```
rpherbig: In Ruby, `nil` is also `false`, so you could change this line to `unless exists?(container)`.

scroll-search.lic
```diff
@@ -0,0 +1,459 @@
...
+    cur_scroll_type = scroll_info.get_type
+
+    # Make sure containers are opened
+    bput("open my #{container}", 'That is already open', 'You open', 'You can\'t open that')
+    bput("open my #{cur_scroll_container}", 'That is already open', 'You open')
+
+    bput("get #{cur_scroll_ordinal} #{cur_scroll_type} from my #{cur_scroll_container}", 'You get')
+
+	container_noun = container
+	container = full_name(container)
```
rpherbig: What will happen if `full_name` returns 'Undefined' here?

scroll-search.lic
```diff
@@ -0,0 +1,459 @@
...
+  def search_container(container)
+    # return if invalid container
+
+    if nil == exists?(container)
+      echo("*** INVALID CONTAINER. ***")
+      return
+    end
+
+    # Get long name of container
+    $container_full = full_name(container)
```
rpherbig: What will happen if `full_name` returns 'Undefined' here?

rpherbig:
I'd prefer each logic change be in its own PR (safe-room changes vs scroll search). Can you split these up?

scroll-search.lic
ohbeanz: This should be "Total owned scrolls count is: " in the debug message -- bad language there. Another good catch.

scroll-search.lic
ohbeanz: True. Works fine without the if check... should've taken a peek at the parse_args func. Thanks!

ohbeanz:
Reverted master and created a new PR fro scroll search.

rpherbig:
See eee68d5d80f25977bdcda9da1bea1289912ef358 which may be related

ohbeanz:
This can be safely closed with **safe_room_empath**, **safe_room_id**, and the [eee68d5 ](https://github.com/rpherbig/dr-scripts/commit/eee68d5d80f25977bdcda9da1bea1289912ef358)commit.

PR #659, opened 2016-08-23
combat-trainer (Offensive Spells - specific targeting)

drboldwin:
The idea is that you can add specific targets for spells which will trigger them to be cast - and will face the target before preparing it/targeting it.  Useful in area's where you might need a little more bang for the casual roaming baddie you can't handle.  Or if there are ever undead/evil things mixed with non-undead/evil things.

in YAML you would use:
target_enemy: noun (note: that this will not work in areas where the nouns are the same ie silverback bear and dire bear both share the noun bear.  I cannot find a way to pull more information from DRRoom.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -734,6 +734,7 @@ class SpellProcess
     return if game_state.npcs.empty?
     return if mana < 40
     ready_spells = @offensive_spells
+                   .select { |spell| spell['target_enemy'] ? DRRoom.npcs.include?(spell['target_enemy']) : true }
                    .select { |spell| spell['min_threshold'] ? game_state.npcs.length >= spell['min_threshold'] : true }
                    .select { |spell| spell['max_threshold'] ? game_state.npcs.length <= spell['max_threshold'] : true }
                    .select { |spell| spell['expire'] ? Flags["ct-#{spell['abbrev']}"] : true }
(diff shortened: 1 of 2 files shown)


rpherbig:
Am I correct that by facing the `target_enemy` you also change your engagement? I.e. this has the side effect of making the `target_enemy` a priority whenever it comes into the room?

drboldwin:
Yes by facing it you are now engaged to that target for all intents and purposes (melee attacks as well) which will focus on killing it.  The drawback of course is if another of the same target enter the room currently you will swap to that one before the initial one dies.  Though we could address this with assess likely - have it check before face and if you are already facing a target of that type then have it not FACE TARGET

drboldwin:
Although the idea behind target_enemy is that the enemy IS a priority target.

rpherbig:
You raise a good point that you won't face it until it comes time to cast an offensive spell.

Is the concept of a "priority enemy" different (or independent) of the concept of "only cast this spell if a certain enemy is in the room"?

drboldwin:
It came about because when I was detecting the enemy it would trigger the spell but then cast it at your current target.  I thought that by facing the enemy it would allow for default casting mechanics to still function unchanged and instead just change your current target.  Though I suppose if this is undesirable though I imagine we could go about it differently.

rpherbig:
Sorry for the confusion, what I mean to say is: would it be preferable to put the "face enemy" logic in, say, SetupProcess, so you will turn to face the priority target ASAP, rather than waiting for the next offensive spell cast?

drboldwin:
ahh - well I thought about this and it may be preferable but I did not know a way to do it without you needing to list target_enemy: outside of the spell.  Can we pull during SetupProcess that there is a spell with a target enemy and to search for that enemy?  and is there a way to make that a constant thing?  or would it have to be in a function that gets called frequently?

drboldwin:
Either way I think we would need to make some use of the assess feature to avoid changing targets before it dies in case another of the same type enters the room.  (and this would preferably be done after it has already detected the enemy with DRRoom.npc so as to avoid any unneeded combat assessments.  (I don't know of a way to track if a specific NPC is dead or not, if this is already a feature we could likely use that instead)

drboldwin:
Hrmm... Those were supposed to be seperate pull requests for a different issue.  Not sure why that happened.

PR #669, opened 2016-08-25
Cyclic Spells During Research

drboldwin:


drboldwin:
Clearly I don't know how to keep pulls separate.

crossing-training.lic
```diff
@@ -403,13 +403,22 @@ class CrossingTraining
   end
 
   def research
+    if @settings.cyclic_research
```
rpherbig: You'll want to add a default value for this setting to base.yaml.

I'd appreciate seeing an example of this in a character's YAML, but it is not required.

crossing-training.lic
```diff
@@ -489,7 +498,9 @@ class CrossingTraining
     fput('release care') if data['cyclic'] && DRSpells.active_spells['Caress of the Sun']
     fput('release bes') if data['cyclic'] && DRSpells.active_spells['Bear Strength']
     fput('release fg') if data['cyclic'] && DRSpells.active_spells['Faenella\'s Grace']
-
+    fput('release rezz') if data['cyclic'] && DRSpells.active_spells['Ressurection']
```
rpherbig: I think we should really refactor this section to release any active cyclics - instead of a bunch of special cases.

crossing-training.lic
```diff
@@ -403,13 +403,22 @@ class CrossingTraining
   end
 
   def research
+    if @settings.cyclic_research
+      research_spell = @settings.crossing_cyclic_research_spells([])
```
rpherbig: You'll want to add a default value for this setting to base.yaml.

I'd appreciate seeing an example of this in a character's YAML, but it is not required.

crossing-training.lic
```diff
@@ -403,13 +403,22 @@ class CrossingTraining
...
+    end
     if 'You cannot begin' == bput("research #{@researching} 300", 'You focus', 'You tentatively', 'You confidently', 'Abandoning the normal', 'You cannot begin')
       fput('research cancel')
       fput('research cancel')
       research
     end
   end
 
+  def cyclic_research(research_spell)
+    cast_spell(@settings.research_training_spells({})[research_spell], research_spell)
```
rpherbig: You'll want to add a default value for this setting to base.yaml.

I'd appreciate seeing an example of this in a character's YAML, but it is not required.

drboldwin:
My YAML looks like this for those skills:

crossing_cyclic_research_spells:
- Utility
- Warding


cyclic_research: true
research_training_spells:
  Warding:
    abbrev: GHS
    cyclic: true
  Utility:
    abbrev: REZZ
    cyclic: true

drboldwin:
So it checks that cyclic_research: is True
Then it runs through Utility and Warding to see which is lower field XP
Selects that spell and sends it to cast_spell function.  So it works as a regular cyclic spell in that way - releases it before casting a new one.

rpherbig:
Can you describe at a high level what your goal is with this code?

drboldwin:
currently crossing-training doesn't really support multiple cyclics and doesn't utilize time spent researching to train magic via cyclic spells.  The goal was to address these issues to make magic training more efficient.

rpherbig:
Would a `cyclic: true` in training_spells give you the same information?

rpherbig:
Can you separate out the code changed in these two PRs to make review and merging easier?

PR #683, opened 2016-08-28
Scroll search Modifications

ohbeanz:
New PR with recommended changes to initial pass of scroll search. 
 * Modified script so that we are using instance variables as opposed to globals ($)
 * Fixed the Transfer and Get scroll actions so that they are appropriately removed from the user's scroll listing


rpherbig:
Can you revert the changes to `safe-room` in this PR?

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+      end
+	elsif args.list
+      if @owned_scrolls.length <= 0
+        echo("*** NO SCROLLS HAVE BEEN ADDED! SEARCH YOUR CONTAINERS FIRST!***")
+        return
+      end
+
+      echo("Scroll scout is: #{@owned_scrolls.length}") if $debug_mode
+      display_scrolls()
+    else
```
rcuhljr: Why is search optional if the else condition is a failure state?

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+      echo("***INVALID ACTION***")
+    end
+  end
+
+  def search_container(container)
+    # return if invalid container
+
+	unless exists?(container)
+      echo("*** INVALID CONTAINER. ***")
```
rcuhljr: You could make a helper `echo_error` or something that takes a val and `echo("*** #{val} ***")` So that you've got uniform formatting on the messages and it's easier to add error outputs.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    end
+
+    # Get long name of container
+    $container_full = full_name(container)
+    $container = get_noun($container_full)
+
+    # iterate over all scroll types
+    scroll_types = $settings.scroll_nouns
+
+    if scroll_types.length <= 0 || scroll_types.nil?
```
rcuhljr: `scroll_types.empty?`

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    args = parse_args(arg_definitions)
+
+	unless args.search || args.list
+		display_args(arg_definitions)
+		return
+	end
+
+    pause
+
+    if args.search
```
rpherbig: If search MUST also have container, then you should consider having two different sets of args (see bescort for an example): 1) just list; and 2) search and container.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    if scroll_types.length <= 0 || scroll_types.nil?
+      echo("*** NO SCROLL TYPTES SET! SETUP YOUR YAML TO INCLUDE VALID \"scroll_types\". ***")
+      return
+    end
+
+    echo("Scroll type count is: #{scroll_types.length}.") if $debug_mode
+
+	# Clear current scroll listing from owned scrolls if
+	# user is searching an existing container
+	if @owned_scrolls.key? ($container_full)
```
rcuhljr: You can use truthy falsey values to simplify this to `if @owned_scrolls[$container_full]`. Also it would be nice to move away from global vars.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+		echo("Clearing scroll array for this container, since it was previously searched --- #{$container_full}!") if $debug_mode
+		@owned_scrolls[$container_full] = []
+	end
+	
+	# reset scroll count
+	@scroll_count = 0
+
```
rcuhljr: Indent levels need to be adjusted.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    end
+
+    # Save scrolls to user variables.
+	if UserVars.owned_scrolls.nil? && @owned_scrolls.nil?
+		echo("Creating an empty hash for vars!") if $debug_mode
+		UserVars.owned_scrolls = {}
+		return
+	end
+	
+	existing_hash = if UserVars.owned_scrolls.nil? then @owned_scrolls.clone else UserVars.owned_scrolls.clone end
```
rcuhljr: If you want to one line this use ruby's ternary operator.
`existing_hash = UserVars.owned_scrolls ? UserVars.owned_scrolls.clone : @owned_scrolls.clone`

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+	
+	new_hash = @owned_scrolls
+	existing_hash.each do |container, scroll_array|
+		if new_hash.key? (container)
```
rcuhljr: `if new_hash[container]`

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+			echo("Existing hash already has this container: #{container}") if $debug_mode
+			next
+		else
+			new_hash[container] = scroll_array
+		end
+	end
```
rcuhljr: You can replace this whole section with `new_hash[container] ||= scroll_array`

rcuhljr: Although it seems like `existing_hash.merge!(new_hash)` might just do what you want to do.

scroll-search.lic
ohbeanz: Because you can execute a scroll-search with the "list" argument without having to search.

scroll-search.lic
ohbeanz: Fair point. Will pop this in next iteration.

scroll-search.lic
ohbeanz: One faster/more efficient than the other? Or is that a style guide thing?

scroll-search.lic
ohbeanz: I'll check that out and follow that way of doing things. Thanks!

scroll-search.lic
ohbeanz: Gotcha. Thanks!

scroll-search.lic
ohbeanz: Ahh, good catch. Thanks!

scroll-search.lic
ohbeanz: Sure. The ternary operator reads better.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+      echo ("*** NO SCROLLS TO DISPLAY -- SEARCH FIRST! ***")
+      return
+    end
+
+	
+    echo ("|============================== Scroll List =============================|")
+    echo (Kernel::sprintf("|%s%s%s%s%s%s%s   %s", format_bar(3, "#"), format_bar(3, "|"), format_bar(14, "Scroll Name"), format_bar(7, "|"),
+																	  format_bar(current_scrolls[0].get_container.length - "Container".length, "Container"), format_bar("Container|".length, "|"), 
+																	  format_bar(5 + "Type".length, "Type"), format_bar("Type".length, "|")))
+    echo ("|========================================================================|")
```
rcuhljr: We can move away from sprintf without losing much here, and it would be a lot simpler in a single helper method that you could use for the header and the body of the table.

```
def echo_scroll_list_line(index, name, container)
  echo("|#{index.to_s.center(6)}|#{name.center(30)}|#{container.center(15)}|")
end
```

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+
+    # Iterate over scrolls to display
+    index = 0
+
+    current_scrolls.each do |container, scroll_array|
+	  
+	  scroll_array.each do |scroll_info|
```
rcuhljr: try `.each_with_index do |scroll_info, index|`

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+  def display_scrolls()
+    current_scrolls = UserVars.owned_scrolls
+
+    if current_scrolls.length <= 0 || current_scrolls.nil?
+      echo ("*** NO SCROLLS TO DISPLAY -- SEARCH FIRST! ***")
+      return
+    end
+
+	
+    echo ("|============================== Scroll List =============================|")
```
rcuhljr: `echo("|#{'Scroll List'.center(45,'=')}|")`  or however many spaces this is.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+		echo(the_scroll)
+		index += 1
+	  end
+    end
+
+    echo ("|                                                                        |")
+    echo ("|           (;send <get #> or ;send <transfer # to <container>>)         |")
+    echo ("|========================================================================|")
+
+    line = nil
```
rcuhljr: This isn't needed.

scroll-search.lic
ohbeanz: Ahhhh! Center! Did NOT know about that. Sooooooo helpful. Thanks a lot!

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+
+  def look_scrolls(all_scrolls)
+    # Reset scroll count
+	index = 0
+
+    all_scrolls.each do |scroll|
+      ordinal = $ORDINALS[index]
+
+      case scroll_look = bput("look #{ordinal} #{scroll} in my #{$container}", 'It is labeled "\w*\s*\w*."', '.* of the \w*\s*\w* spell.', 'three-dimensional shapes cover much of the scroll', 'You see nothing unusual.', 'I could not find what you were referring to')
+      when 'You see nothing unusual.', 'What were you referring to', 'I could not find what you were referring to' then break
```
rcuhljr: It might be easier to make a when for `/It is labeled "(\w*\s*\w*)."/i` and remove this when, letting all of these cases just be ignored.

scroll-search.lic
ohbeanz: Ahhhh, functions! I love this. Thanks.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+      when 'three-dimensional shapes cover much of the scroll' then read_scroll(scroll)
+      else
+        scroll_look =~ /It is labeled "(\w*\s*\w*)."/i
```
rcuhljr: If you move this into the when you don't need to run the regex match again.

scroll-search.lic
ohbeanz: What's "this"?

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    parameter = "C #{scroll_noun}"
+    scrolls = rummage(parameter, $container)
+    return scrolls
+  end
+
+  def look_scrolls(all_scrolls)
+    # Reset scroll count
+	index = 0
+
+    all_scrolls.each do |scroll|
```
rcuhljr: `each_with_index` again.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+			if scroll_info.get_index == index.to_i
+				echo ("Found scroll info : #{scroll_info.get_spell}") if $debug_mode
+				the_scroll = scroll_info
+			end
+		end
+	end
+	
+	echo("The scroll is: #{the_scroll.get_spell}") if $debug_mode
+	
+	if the_scroll.nil?
```
rcuhljr: unless the_scroll

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    # Instance variables
+    @container = container
+    @index = index
+	@container_index = container_index
+    @ordinal = ordinal
+    @spell = spell
+    @type = type
+  end
+
+  def get_container
```
rcuhljr: You can just make all of these accessors so people can just do scroll_info.index scroll_info.container etc. Name things after what they return not what they do.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    name_full = if Regexp.last_match(1).empty? then "Undefined" else Regexp.last_match(1) end
+	
+	return name_full
+  end
+
+  def get_scroll_info(index)
+    # Retrieve scroll info 
+    current_scrolls = UserVars.owned_scrolls
+	
+	the_scroll = nil
```
rcuhljr: `the_scroll = current_scrolls.values.flatten.find{|scroll_info| scroll_info.index == index.to_i}`

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+	end
+
+    echo("*** SCROLL WAS GRABBED! ***")
+  end
+  
+  def full_name(noun)
+	noun_name = bput("tap #{noun}", '^You tap ((\w+\s*\W*)+) that\s*.*')
+    noun_name =~ /^You tap ((\w+\s*\W*)+) that\s*.*/i
+
+    name_full = if Regexp.last_match(1).empty? then "Undefined" else Regexp.last_match(1) end
```
rcuhljr: Ternary `?` operator.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+		scroll_information.set_container_index(new_index)
+	end
+  end
+
+  def resave_container(container, scroll_info, remove)
+  
+	# if wasRemoved is true, we remove the scroll from
+	# the container. If not, we simply add it to the container specified.
+	
+	if remove
```
rcuhljr: Break some of these cases out into helper methods maybe? this thing is a beast.

scroll-search.lic
rcuhljr: There's also ljust and rjust depending on how you want them to work, and all three take an optional padding character argument.

scroll-search.lic
rcuhljr: the line = nil assignment.

rcuhljr:
Looking good so far, my mind is a little fried. I can look it over again once we've recycled the suggestions you like.

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+      else
+        search_container(args.container)
+      end
+	elsif args.list
+      if @owned_scrolls.length <= 0
+        echo("*** NO SCROLLS HAVE BEEN ADDED! SEARCH YOUR CONTAINERS FIRST!***")
+        return
+      end
+
+      echo("Scroll scout is: #{@owned_scrolls.length}") if $debug_mode
```
rpherbig: I think this message should be "Total owned scrolls count is: #{...}" if...

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+  def search_container(container)
+    # return if invalid container
+
+	unless exists?(container)
+      echo("*** INVALID CONTAINER. ***")
+      return
+    end
+
+    # Get long name of container
+    $container_full = full_name(container)
```
rpherbig: What will happen here if `full_name` returns "Undefined"?

scroll-search.lic
```diff
@@ -0,0 +1,520 @@
...
+    bput("open my #{cur_scroll_container}", 'That is already open', 'You open')
+
+    bput("get #{cur_scroll_ordinal} #{cur_scroll_type} from my #{cur_scroll_container}", 'You get')
+
+	# Reorder the origin container
+	reorder_container(scroll_info.get_container, scroll_info, true)
+	resave_container(scroll_info.get_container, scroll_info, true)
+	
+	container_noun = container
+	container = full_name(container)
```
rpherbig: What will happen here if `full_name` returns "Undefined"?

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	
+	unless the_scroll
+	  echo_message("INVALID SCROLL NUMBER")
+	  exit
+	end
+	
+	return the_scroll
+  end
+
+  def empty_hands
```
rpherbig: How does this differ from `stow_hands` in `common-items`?

scroll-search.lic
```diff
@@ -63,8 +63,8 @@ class ScrollSearch
 
     # Variables
     @scroll_count = 0
-	@global_scroll_index = if UserVars.scroll_counter.nil? then 0 else UserVars.scroll_counter end
-    @owned_scrolls = if UserVars.owned_scrolls.empty? then {} else UserVars.owned_scrolls end
+	@global_scroll_index = UserVars.scroll_counter.nil? ? 0 : UserVars.scroll_counter
```
rpherbig: You could also do this assignment with `@global_scroll_index = UserVars.scroll_counter || 0` (also applies to the next line).

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+
+class ScrollSearch
+  include DRC
+  include DRCI
+
+  def initialize
+
+	# Variables
+	@scroll_count = 0
+	@global_scroll_index = UserVars.scroll_counter.nil? ? 0 : UserVars.scroll_counter
```
rpherbig: You could also do this assignment with `@global_scroll_index = UserVars.scroll_counter || 0`.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	# Instance variables
+	@container = container
+	@index = index
+	@container_index = container_index
+	@ordinal = ordinal
+	@spell = spell
+	@type = type
+  end
+
+  def container
```
rpherbig: Ruby has something called `attr_accessor`, which basically is a shorthand for defining a getter and setter. You can see an example in combat-trainer (GameState). That might simplify the ScrollInfo class a bit.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
+=begin
+  Documentation: TBD
```
rpherbig: Don't forget to fill this out (here and the wiki) once we merge :)

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	@scroll_count = 0
+	@global_scroll_index = UserVars.scroll_counter.nil? ? 0 : UserVars.scroll_counter
+	@owned_scrolls = UserVars.owned_scrolls.empty? ? {} : UserVars.owned_scrolls
+    
+	# RESET ALL
+	# UserVars.owned_scrolls = { }
+	# @owned_scrolls = { }
+	# UserVars.scroll_counter = 0
+
+	$settings = get_settings()
```
rpherbig: There are still some global vars

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	$debug_mode = $settings.debug_mode
```
rpherbig: This needs to have a unique name to avoid a collision with other scripts. For example, the debug flag in `addroom` is `debug_mode_ar`.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+		{ name: 'container', regex: /\w+/i, description: 'The container to search. ' }
+	  ],
+	  [
+		{ name: 'list', regex: /^lis.*/i, optional: false, description: 'Lists all of the scrolls that have been categorized after a SEARCH. ' },
+	  ]
+	]
+
+	args = parse_args(arg_definitions)
+
+	unless args.search || args.list
```
rpherbig: Is this now obsolete? The user will have to provide args that match one of the two defined in arg_definitions; and search and list are not optional.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	  return
+	end
+
+	pause
+
+	if args.search
+	  if args.container.nil?
+		echo_message("SPECIFY THE CONTAINER TO SEARCH!")
+		return
+	  else
```
rpherbig: Because of the return one line 105, you don't need the else case here (i.e. if the return happens, line 107 won't happen even if the else is removed).

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	  end
+	elsif args.list
+	  if @owned_scrolls.length <= 0
+		echo_message("NO SCROLLS HAVE BEEN ADDED! SEARCH YOUR CONTAINERS FIRST!")
+	    return
+	  end
+
+	  echo("Total scroll count is: #{@owned_scrolls.length}") if $debug_mode
+	  display_scrolls()
+	else
```
rpherbig: Given the arg definitions, can this else be reached?

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	  return
+	end
+
+	# Empty your hands
+	empty_hands
+
+	# Get long name of container
+	$container_full = full_name(container)
+	if $container_full =~ /Undefined/i
+	  echo("Was unable to retrieve the full name of this container...") if $debug_mode
```
rpherbig: I recommend including the name of the container in the debug output

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	# Save scrolls to user variables.
+	if UserVars.owned_scrolls.nil? && @owned_scrolls.nil?
+	  echo("Creating an empty hash for vars!") if $debug_mode
+	  UserVars.owned_scrolls = {}
+	  return
+	end
+	
+	existing_hash = UserVars.owned_scrolls ? UserVars.owned_scrolls.clone : @owned_scrolls.clone
+	new_hash = @owned_scrolls
+	existing_hash.merge!(new_hash)
```
rpherbig: What does this `merge!` accomplish? As far as I can tell, `existing_hash` is not used (and `new_hash` is not modified).

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+
+	  transfer_scroll(scroll_index, transfer_container)
+	end
+  end
+
+  # Private methods
+  private
+
+  def get_scrolls(scroll)
+	scroll_noun = scroll #get_noun(scroll)
```
rpherbig: Is this commented out code a TODO, or a reminder of something, or dead code?

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+		  scroll_array.push(new_scroll)
+		  echo("1. Scroll size is #{scroll_array.length}") if $debug_mode
+		else
+		  scroll_array.push(new_scroll)
+		  echo("2. Scroll size is #{scroll_array.length}") if $debug_mode
+		end
+		
+		echo("Adding scroll array of size #{scroll_array.length} to hash.") if $debug_mode
+		@owned_scrolls[$container_full] = scroll_array
+      end
```
rpherbig: This (and the next few lines) need their indentation fixed.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+  def get_scrolls(scroll)
+	scroll_noun = scroll #get_noun(scroll)
+	parameter = "C #{scroll_noun}"
+	scrolls = rummage(parameter, $container)
+	return scrolls
+  end
+
+  def look_scrolls(all_scrolls)
+
+	all_scrolls.each_with_index do |scroll, index|
```
rpherbig: Doesn't DR (the game itself) only support up to "twelve" for an ordinal? Have you tested this with thirteen scrolls in a container?

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+		@owned_scrolls[$container_full] = scroll_array
+      end
+
+	  @scroll_count += 1
+	  @global_scroll_index += 1
+    end
+  end
+
+  def read_scroll(scroll)
+	bput("get $ORDINALS[#{@scroll_count}] #{scroll} from my #{$container}", 'You get')
```
rpherbig: `$ORDINALS` needs to be in `#{...}` to be evaluated.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+
+	#case bput("read #{scroll}", 'The writing is too small', )
```
rpherbig: What is this commented out code?

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	  scrolls_in_container = @owned_scrolls[scroll_info.container]
+	  removed = scrolls_in_container.delete(scroll_info)
+	  if removed
+		echo("Deleting scroll #{removed.spell} from container #{scroll_info.container}") if $debug_mode
+		# Re-save owned scrolls listing
+		@owned_scrolls[scroll_info.container] = scrolls_in_container
+			
+		# Re-save user variables
+		UserVars.owned_scrolls = @owned_scrolls
+	  else
```
rpherbig: I recommend checking this case first (e.g. `unless removed`) and then, due to the exit, you do not need to have an else.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+		  echo("Scrolls in container is #{scroll_array.spell}") if $debug_mode
+		end
+			
+		# List everything
+		@owned_scrolls.each do |container, scroll_array|
+		  scroll_array.each do |scroll_information|
+			echo("Container #{container} | #{scroll_information.spell}") if $debug_mode
+		  end
+		end
+	  else
```
rpherbig: I recommend checking this case first (e.g. `unless removed`) and then, due to the exit, you do not need to have an else.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+	# Reorder container
+	reorder_container(scroll_info.container, scroll_info, true)
+	
+	# Remove this scroll from our hash map
+	scrolls_in_container = @owned_scrolls[scroll_info.container]
+	
+	echo("Scrolls in container count is: #{scrolls_in_container.length}") if $debug_mode
+	scrolls_in_container.each do |scroll_information|
+	  echo ("Index of current scroll in container is: #{scroll_information.index}. Index is: #{index}") if $debug_mode
+	  if scroll_information.index != index.to_i
```
rpherbig: You could do something like `scrolls_in_container.select { |scroll_information| scroll_information.index == index.to_i }.each do ...` to remove the if and else from this block.

scroll-search.lic
```diff
@@ -0,0 +1,533 @@
...
+  end
+
+  def get_scroll_info(index)
+	# Retrieve scroll info 
+	current_scrolls = UserVars.owned_scrolls
+	
+	the_scroll = nil
+	
+	current_scrolls.each do |container, scroll_array|
+	  scroll_array.each do |scroll_info|
```
rpherbig: You could use `.find` instead of `.each` (see https://ruby-doc.org/core-2.2.0/Enumerable.html#method-i-find)

rpherbig:
How's this going @ohbeanz ?

scroll-search.lic
ohbeanz: Dead code. Good catch.

scroll-search.lic
ohbeanz: Turns out I don't need this at all since I'm clearing the current hash set (@owned_scrolls) if I'm re-searching a container.

scroll-search.lic
ohbeanz: Untested. Marked as a todo though.

scroll-search.lic
ohbeanz: Curious on the efficiency of this... seems "Linq-esque" and I assume some sort of extra overheard, though I'm not sure how it compares to doing a comparison check. I defer though as it is a bit more readable.

rpherbig:
This looks pretty good, but GitHub indicates a merge conflict. Can you merge master in to your branch/fork?

scroll-search.lic
```diff
@@ -0,0 +1,491 @@
...
+	@index = index
+	@container_index = container_index
+	@ordinal = ordinal
+	@spell = spell
+	@type = type
+  end
+
+  attr_accessor :container
+  attr_accessor :spell
+  attr_accessor :ordinal
```
rcuhljr: Don't feel compelled, but you can list all of these on one line for future information.
`attr_accessor :container, :spell, :ordinal, ...`

Issue #674, opened 2016-08-26
Universal Cyclic Release


drboldwin:
So Etreu has been mentioning how he wants to get away from the Hard Coded cyclic release that is used and so I thought I'd propose this idea here.  (I haven't tested this and there's things to work through still but at least we can get a dialogue going).

I thought that we could handle each cyclic spell with 1 line like:

fput('release #{data['abbrev']') if data['cyclic'] && DRSpells.active_spells[#{data['name']]"

note this would require people to add the line:
name: <spell name> 
to their yaml within the spell - example:

```
  Warding:
    abbrev: GHS
    name: Ghost Shroud
    cyclic: true
```

the issue I find is that this is being called in cast_spell and so the data that it is referring to is from the spell it is currently casting - so although this would resolve single cyclics I would really like this to support multiple cyclics (see my other pull request).

So perhaps saving those variables as some sort of last_cast and calling it that way instead?  Thoughts?

drboldwin:
Another option is for us to create an array that holds to spell name associated with the abbrev, and just compare DRSpells.active (looking for a matching spell name) and then release that abbrev.

rpherbig:
Why not compare `@settings.training_spells` to `DRSpells.active_spells` and release any that are marked as cyclic?

drboldwin:
we would still need to add spell name though would we not?  since that isn't done in training spells like it is done in combat-trainer offensive/buff spells.  Unless I am missing something (I know I've never seen it done in a YAML)

rcuhljr:
I'd rather just have spell data stored somewhere.

```
- name: Ghost Shroud
  ability: Warding
  abbrev: /^ghs$/i
  cyclic: true
```

So that people can just list the abbreviation and things we don't know ahead of time.

drboldwin:
```
- name: Ghost Shroud
  ability: Warding
  abbrev: /^GHS$/i
  cyclic: true
- name: Abandoned Heart
  ability: Targeted Magic
  abbrev: /^ABAN$/i
  cyclic: true
- name: Aesandry Darlaeth
  ability: Augmentation
  abbrev: /^AD$/i 
  cyclic: true
- name: Aether Cloak
  ability: Warding
  abbrev: /^AC$/i
  cyclic: true
- name: Aether Wolves
  ability: Debilitation
  abbrev: /^AEWO$/i
  cyclic: true
- name: Albreda's Balm
  ability: Debilitation
  abbrev: /^ALB$/i
[...]

PR #689, opened 2016-08-29
Cyclic Spells During Research

drboldwin:
Also a small update to Training_manager to handle releasing cyclic spells during transitions from crossing_training to hunting_buddy

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -413,11 +413,20 @@ class CrossingTraining
   end
 
   def research
+    if @settings.cyclic_research
+      research_spell = @settings.crossing_cyclic_research_spells([])
+                       .min_by { |s| DRSkill.getxp(s) }
+      cyclic_research(research_spell)
+    end
     if 'You cannot begin' == bput("research #{@researching} 300", 'You focus', 'You tentatively', 'You confidently', 'Abandoning the normal', 'You cannot begin')
       fput('research cancel')
       fput('research cancel')
       research
     end
+  end
+
+  def cyclic_research(research_spell)
+    cast_spell(@settings.research_training_spells({})[research_spell], research_spell)
   end
 
   def check_research
@@ -499,6 +508,8 @@ class CrossingTraining
     fput('release care') if data['cyclic'] && DRSpells.active_spells['Caress of the Sun']
     fput('release bes') if data['cyclic'] && DRSpells.active_spells['Bear Strength']
     fput('release fg') if data['cyclic'] && DRSpells.active_spells['Faenella\'s Grace']
+    fput('release rezz') if data['cyclic'] && DRSpells.active_spells['Resurrection']
+    fput('release ghs') if data['cyclic'] && DRSpells.active_spells['Ghost Shroud']
 
     bput('prep symb', 'You recall the exact details of the', 'But you\'ve already prepared') if data['symbiosis']
     scaled_prep_mana = data['mana']
--- a/training-manager.lic
+++ b/training-manager.lic
@@ -32,6 +32,7 @@ class TrainingManager
       clear
 
       check_favors
+      check_cyclics
 
       wait_for_script_to_complete('sell-loot')
       wait_for_script_to_complete('mining-buddy') if @settings.mine_while_training(false) && !skipmine
@@ -57,7 +58,8 @@ class TrainingManager
       clear
 
       check_favors
-
+      check_cyclics
+      
       wait_for_script_to_complete('sell-loot')
 
       wait_for_script_to_complete('mining-buddy') if @settings.mine_while_training(false)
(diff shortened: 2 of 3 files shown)


drboldwin:
Redid this so it was by itself.

training-manager.lic
```diff
@@ -99,6 +102,17 @@ class TrainingManager
       end
     end
   end
+  
+  def check_cyclics
+    echo DRSpells.active_spells
```
rpherbig: This should be conditional on the debug flag

drboldwin:
Oops - will add

drboldwin:
Tried to do that but would no longer let me likely because the fork I used to create this pull request has been deleted.

Although the line can be deleted if you would like since I don't believe there is a debug flag for Training-manager.  I would likely end up trying to add the crossing-trianing debug but not sure if that would work.  Like this:

`echo DRSpells.active_spells if UserVars.crossing_trainer_debug`

drboldwin:
but really I forgot I had that echo in there and I am addicted to echoing things when I code.  Got rid of most of them and missed this one.

training-manager.lic
```diff
@@ -99,6 +102,17 @@ class TrainingManager
...
+    fput('release gj') if DRSpells.active_spells['Glythtide\'s Joy']
+    fput('release care') if DRSpells.active_spells['Caress of the Sun']
+    fput('release bes') if DRSpells.active_spells['Bear Strength']
+    fput('release fg') if DRSpells.active_spells['Faenella\'s Grace']
+    fput('release rezz') if DRSpells.active_spells['Resurrection']
+    fput('release ghs') if DRSpells.active_spells['Ghost Shroud']
+    return
```
rpherbig: This return is unnecessary

training-manager.lic
```diff
@@ -99,6 +102,17 @@ class TrainingManager
       end
     end
   end
+  
+  def check_cyclics
```
rpherbig: This method might belong in common-arcana... what do you think?

crossing-training.lic
```diff
@@ -413,13 +413,22 @@ class CrossingTraining
   end
 
   def research
+    if @settings.cyclic_research
+      research_spell = @settings.crossing_cyclic_research_spells([])
```
rpherbig: This appears to be a list of skills, and I feel 'crossing' is not communicating any information. I recommend naming it `cyclic_research_skills`.

crossing-training.lic
```diff
@@ -413,13 +413,22 @@ class CrossingTraining
   end
 
   def research
+    if @settings.cyclic_research
```
rpherbig: Is this boolean necessary, or is the presence of `crossing_cyclic_research_spells` (or `cyclic_research_skills`) enough to indicate this should be done?

crossing-training.lic
```diff
@@ -413,13 +413,22 @@ class CrossingTraining
...
+      research_spell = @settings.crossing_cyclic_research_spells([])
+                       .min_by { |s| DRSkill.getxp(s) }
+      cyclic_research(research_spell)
```
rpherbig: This appears to work for non-cyclic spells too. Is that intentional? If so, should any of the naming change?

rpherbig:
This is the sample YAML excerpt posted in the previous PR:
```
crossing_cyclic_research_spells:
- Utility
- Warding


cyclic_research: true
research_training_spells:
  Warding:
    abbrev: GHS
    cyclic: true
  Utility:
    abbrev: REZZ
    cyclic: true
```

drboldwin:
Some of these will be addressed when we finish up the Universal Cyclic Release as seen here:

https://github.com/rpherbig/dr-scripts/issues/674

It will take care of the defining of cyclics in training-manager.lic as well as the hardcoded aspect in crossing-training.lic and should be pushed before this is so that we can adjust accordingly.

> This appears to work for non-cyclic spells too. Is that intentional? If so, should any of the naming change?

It does though it was not my intent.  I guess I just didn't account for non-cyclic since the only spells listed under research_training_spells should be cyclic spells anyways.  I was unsure of how to go about first picking the lower ability and then determining if it was cyclic and then going back and picking a different ability if it wasn't.  If this logic is put in then we could avoid listing the spells separately completely and base it off of training_spells as you had previously mentioned.

> This appears to be a list of skills, and I feel 'crossing' is not communicating any information. I recommend naming it cyclic_research_skills.

Agreed here - crossing has no actual bearing to it - I was simply copying the structure used for listing skills to train in crossing which was crossing_training so I called it crossing_cyclic_research_spells.  With the expanding of hometowns and whatnot this seems even more unnecessary.

drboldwin:
Closing this and reopening to reflect new changes to common-arcana

PR #690, opened 2016-08-29
Priority Spell Targeting

drboldwin:


drboldwin:
updated this to be by itself

combat-trainer.lic
```diff
@@ -787,6 +787,7 @@ class SpellProcess
     return if game_state.npcs.empty?
     return if mana < 40
     ready_spells = @offensive_spells
+                   .select { |spell| spell['target_enemy'] ? DRRoom.npcs.include?(spell['target_enemy']) : true }
```
rpherbig: Should you check `game_state.npcs` instead of `DRRoom.npcs`?

drboldwin:
Perhaps, what exactly is the difference?  (I tried both and both worked.)

rpherbig:
`DRRoom.npcs` is all NPCs in the room. `game_state.npcs` is the same, but filtering out anything from `settings.ignored_npcs`. Generally you should use the helper in game state.

rcuhljr:
I'm fine with this, but it still feels like facing should be done in a process other than spell process, like the setup process.

rpherbig:
I think "switching to a priority target" is independent of "only cast these spells at a priority target". But I think this PR is the simplest solution. What do you think about merging and changing this if we hear feedback from the userbase?

PR #704, opened 2016-09-02
Handle Shard Gemshop in Sell-loot

jandersson:
The shard gem shop changes the person appraising the gems depending on time of day. At night it is Veilex, in the morning it is Fatima. In base-town we have Fatima listed, so if we see Veilex in the shop we temporarily overwrite Fatima.

### DIFF
--- a/profiles/base-town.yaml
+++ b/profiles/base-town.yaml
@@ -2,10 +2,10 @@
 Crossing:
   leather_repair:
     id: 1544
-    name: randal
+    name: Randal
   metal_repair:
     id: 19093
-    name: catrox
+    name: Catrox
   deposit:
     id: 1900
   exchange:
@@ -16,7 +16,7 @@ Crossing:
     name: appraiser
   tannery:
     id: 8266
-    name: falken
+    name: Falken
   npc_empath:
     id: 6218
 Leth Deriel:
@@ -40,13 +40,13 @@ Shard:
     - Veilex
   tannery:
     id: 19340
-    name: tremagis
+    name: Tremagis
   leather_repair:
     id: 19270
-    name: ylono
+    name: Ylono
   metal_repair:
-    id:
-    name:
+    id: 9639
+    name: Granzer
 Therenborough:
 Riverhaven:
   deposit:
--- a/profiles/base-town.yaml
+++ b/profiles/base-town.yaml
@@ -19,13 +19,22 @@ Crossing:
     name: Falken
   npc_empath:
     id: 6218
+    action: lie
 Leth Deriel:
   gemshop:
     id: 50268
-    name:
+    name: Ghendil
   tannery:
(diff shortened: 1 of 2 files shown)


rpherbig:
I'm wondering if a better approach would be to change base-town... maybe a list of gemshop NPCs? What do you think?

jandersson:
That sounds like a better approach as there might be other towns with similar situations. However that line of code in the PR would remain similar though, right? We check against the list instead of against the string

jandersson:
I noticed I deleted some spaces in the empty fields. Are those considered good style or does that change the datatype?

rcuhljr:
They'll just be ignored, I have 'remove trailing whitespace on save' set in my editor so they wouldn't last long anyways.

sell-loot.lic
```diff
@@ -76,11 +76,18 @@ class SellLoot
...
-    fput("ask #{@hometown['gemshop']['name']} for #{adj} pouch")
+    clerk = which_clerk(@hometown['gemshop']['name'])
+    fput("ask #{clerk} for #{adj} pouch")
     fput("put my pouch in my #{container}")
   end
 
+  def which_clerk(clerks)
+    return clerks if clerks.is_a? String
+
+    clerks.select! { |clerk| DRRoom.npcs.include?(clerk) }
```
rpherbig: These two lines can be replaced with `clerks.find { |clerk| DRRoom.npcs.include?(clerk) }`

jandersson: Perfect, thank you

jandersson: Would this require turning every gemshop entry into a list? I was trying to avoid that with the return if is a String statement

jandersson: Or do you mean the two lines to be the select! and clerks.first

rcuhljr: I'd move towards making it a list instead of relying on type checking, but these aren't really user editable yaml files so it's low concern that they might get confused. However the whole method could just be `clerks.is_a?(String) ? clerks : clerks.find {...`

jandersson: OK I will test out the changes. Thanks for the advice!

jandersson: So, just to put my thoughts out there... Since this is semistructured data, we dont have to make every attribute a list because of one outlier. We have to adapt the program to handle this outlier in either case. However the data entry becomes much easier if its just a field for a string, which i think all the other gemshops are.

profiles/base-town.yaml
```diff
@@ -16,16 +16,25 @@ Crossing:
     name: appraiser
   tannery:
     id: 8266
-    name: falken
+    name: Falken
   npc_empath:
     id: 6218
+    action: lie
```
rpherbig: Out of curiosity, why did you add an action to `npc_empath`?

PR #705, opened 2016-09-02
Bead-Carving in Theurgy.lic

drboldwin:
This isn't completely done yet but wanted to get it up for initial review.  All I will likely add is an invasion check maybe - and a check against Commune Truffenyi training ability in your YAML (which I haven't written yet but this is a super simple addition I will be making later today) it will use the beads carved in crossing-training to commune during combat-trainer for additional theurgy XP.


theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
...
+    if find_project.include? "Sorcery"
+      hold_research = 'SORCERY'
+    end
+    if find_project.include? "Utility"
+      hold_research = 'UTILITY'
+    end
+    if find_project.include? "Warding"
+      hold_research =  'WARDING'
+    end
+    fput 'prep bless'
```
drboldwin: Also might make this an if/elseif/else statement (it was how I had originally done it but I wasn't getting hold_research to work properly.  This ended up being because I didn't have '' around the word but by the time I realized that I had changed it to multiple if statements.

theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
   put_in_container($settings.water_holder)
 end
 
+def prayer_bead
+  return unless DRSkill.getrank('Mechanical Lore') >= 140
```
rpherbig: Would `return if DRSkill(...) < 140` be clearer?

theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
...
+
+  fput "stow #{GameObj.right_hand.noun}" if GameObj.right_hand.name != 'Empty'
```
rpherbig: How does this differ from the helper in `common-items`?

theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
...
+  fput "stow #{GameObj.left_hand.noun}" if GameObj.left_hand.name != 'Empty'
+  get_from_container('prayer chain')
+  chain_contents = DRC.bput("look at my chain", /^Strung on to the prayer bead chain you see .*\./, 'There are currently no beads on it.')
+  bead_count = DRC.list_to_nouns(chain_contents.match(/^Strung on to the prayer bead chain you see (.*)\./).to_a[1])
+  return unless bead_count.count < 10
```
rpherbig: Would `return if bead_count.count >= 10` be clearer?

Why does the count matter?

theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
...
 end
 
+def prayer_bead
+  return unless DRSkill.getrank('Mechanical Lore') >= 140
+
+  fput "stow #{GameObj.right_hand.noun}" if GameObj.right_hand.name != 'Empty'
+  fput "stow #{GameObj.left_hand.noun}" if GameObj.left_hand.name != 'Empty'
+  get_from_container('prayer chain')
+  chain_contents = DRC.bput("look at my chain", /^Strung on to the prayer bead chain you see .*\./, 'There are currently no beads on it.')
+  bead_count = DRC.list_to_nouns(chain_contents.match(/^Strung on to the prayer bead chain you see (.*)\./).to_a[1])
```
rpherbig: Would a better name be `bead_nouns` or `beads`?

theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
...
+  return unless bead_count.count < 10 
+  put_in_container('prayer chain')
+  DRC.forage? 'limb'
+  get_from_container($settings.water_holder)
+  fput("sprinkle #{$settings.water_holder} on my limb")
+  put_in_container($settings.water_holder)
+  if 'interrupt your research' == DRC.bput('prep bless', 'Bless spell', 'interrupt your research')
```
rpherbig: Do we really want to interrupt research to do this? There's potentially quite a bit of time invested in the research...

theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
...
+  get_from_container('carving knife')
+  while checkright == 'limb'
+    fput('carve my limb with my knife')
+    pause 1
+    waitrt?
+  end
+  put_in_container('carving knife')
+  get_from_container('shaper')
+  while checkright == 'block'
+    fput("shape block to #{$settings.aspect}")
```
rpherbig: Is `aspect` a new setting? If so, it needs to have a default in `base.yaml`.

theurgy.lic
```diff
@@ -62,6 +62,69 @@ def clean_altar
   put_in_container($settings.water_holder)
 end
 
+def prayer_bead
```
rpherbig: Is there a circle requirement to do this?

theurgy.lic
drboldwin: I was thinking about checking the % and deciding to interrupt or wait it out.  I personally prefer the  interrupt in most cases.  I was thinking 60% could be the threshold

theurgy.lic
drboldwin: because they can't have more than 10 beads on them.  So it's checking if the chain is full.

theurgy.lic
drboldwin: I did it this way because it was done this way in other parts of theurgy.lic

theurgy.lic
drboldwin: yeah - I just haven't pulled the base.yaml for it yet.  It would be otherwise I would have to run a check against your favored god and hope that your last favor was from that god.  This way you have more control.

theurgy.lic
drboldwin: not exactly - you can make the beads with 140+ mech lore (with the method I'm doing it) but it's going to run a check against commune truffanyis which does have a circle 60 requirement.

theurgy.lic
drboldwin: Sure, I can change that.

theurgy.lic
drboldwin: Sure - I like beads and bead_count

theurgy.lic
```diff
@@ -80,21 +80,18 @@ def prayer_bead
     find_project = fput("research status")
     if find_project.include? "Fundamental"
       hold_research = 'FUNDAMENTAL'
-    end
-    if find_project.include? "Augmentation"
+    elseif find_project.include? "Augmentation"
```
rpherbig: Does `elseif` work? I thought Ruby only supported `elsif`...

theurgy.lic
rpherbig: In that case you'll want to add a check to the `prayer_bead` method against `DRStats.circle` (see `comm_eluned` for an example.

theurgy.lic
```diff
@@ -62,6 +62,66 @@ def clean_altar
   put_in_container($settings.water_holder)
 end
 
+def prayer_bead
+  return if DRSkill.getrank('Mechanical Lore') < 140
+
+  fput "stow #{GameObj.right_hand.noun}" if GameObj.right_hand.name != 'Empty'
+  fput "stow #{GameObj.left_hand.noun}" if GameObj.left_hand.name != 'Empty'
+  get_from_container('prayer chain')
```
rpherbig: I don't see any conditionals here - what happens if you do not have a `prayer chain` in your inventory?

theurgy.lic
```diff
@@ -62,6 +62,66 @@ def clean_altar
...
+  put_in_container('prayer chain')
+  DRC.forage? 'limb'
+  get_from_container($settings.water_holder)
+  fput("sprinkle #{$settings.water_holder} on my limb")
+  put_in_container($settings.water_holder)
+  if 'interrupt your research' == DRC.bput('prep bless', 'Bless spell', 'interrupt your research')
+    find_project = fput("research status")
+    if find_project.include? "Fundamental"
+      hold_research = 'FUNDAMENTAL'
+    elseif find_project.include? "Augmentation"
```
rpherbig: Does `elseif` work? I thought Ruby only supported `elsif`...

theurgy.lic
drboldwin: I had hoped this would be resolved because of the check against commune truff that will be in your YAML - that will only function if you have a prayer chain (and will be noted as such) so this is really just being added to theurgy so you can keep your prayer chain full with the added benefit of a large devotion boost.  I haven't added that check yet because I haven't finished commune truff for combat-trainer yet - but essentially if you don't use that in training-abilities theurgy.lic will skip right past this.

theurgy.lic
drboldwin: Oh.  That's odd - did not know that.  Haven't tested these new changes yet.  Will give it a go.

theurgy.lic
drboldwin: I actually checked this against $settings.aspect instead.  That way if you want to carve beads without having commune truffinyi you can still do it to at least fill a chain - and will likely add a similar call while at the altar later in Theurgy.lic to meditate on the chain so you get the devotion boost.

rpherbig: `training_abilities` only affects `combat-trainer`. This script is also called from `crossing-training` and can be called manually.

drboldwin: yeah - both those should be resolved I believe since I decided to check against .aspect instead.

drboldwin:
This got screwed up with other pull requests.  Github needs to make this easier.  Gonna make some edits and will resubmit shortly.

PR #714, opened 2016-09-02
Theurgy - Prayer Beads + Truffanyi Commune

drboldwin:


theurgy.lic
```diff
@@ -62,6 +62,68 @@ def clean_altar
   put_in_container($settings.water_holder)
 end
 
+def prayer_bead
+  return if DRSkill.getrank('Mechanical Lore') < 140
```
drboldwin: Changed - to make more understandable

theurgy.lic
```diff
@@ -62,6 +62,68 @@ def clean_altar
...
+  put_in_container('shaper')
+  get_from_container('prayer chain')
+  fput ("put #{$settings.aspect} bead on prayer chain")
+  put_in_container('prayer chain')
+end    
+
+DRCT.walk_to '1030'
+if $settings.aspect
+  prayer_bead
+end
```
drboldwin: this check should stop people from running through the bead carve routine if they don't have the shaper/knife/chain for it.  Since you will need to add aspect: to your YAML in order for it to initialize.

theurgy.lic
```diff
@@ -62,6 +62,68 @@ def clean_altar
...
+      hold_research = 'STREAM'
+    elsif find_project.include? "Sorcery"
+      hold_research = 'SORCERY'
+    elsif find_project.include? "Utility"
+      hold_research = 'UTILITY'
+    elsif find_project.include? "Warding"
+      hold_research =  'WARDING'
+    else
+      return
+    end
```
drboldwin: fixed these to use elsif instead of elseif (my bad) also if for some reason it doesn't recognize your research it will simple not carve (since it wouldn't be able to restart your research)

drboldwin:
So all that's left is - add a meditate on your chain while at the altar - and a check that determines whether or not it's worth interrupting research or just waiting for it to finish.

theurgy.lic
rpherbig: I recommend renaming it to `theurgy_aspect` or `prayer_bead_aspect` to make it a bit clearer where it is being used.

theurgy.lic
```diff
@@ -55,6 +55,14 @@ def buy_wine
...
 end
 
+def buy_Taffelberries
+  DRCI.order_item(8280, 14)
+  fput 'get Taffelberries'
+  put_in_container('Taffelberries')
+  fput 'get Taffelberries'
+  put_in_container('Taffelberries')
+end
+
```
drboldwin: order_item in common-items ends up buying 2 of these so I just get them both.  (will make you have to do this run less often)

theurgy.lic
```diff
@@ -109,19 +117,51 @@ def prayer_bead
...
+def meditate_bead
+  get_from_container('prayer chain')
+  beads = DRC.bput("look at my chain", /^Strung on to the prayer bead chain you see .*\./, 'There are currently no beads on it.')
+  bead_count = DRC.list_to_nouns(beads.match(/^Strung on to the prayer bead chain you see (.*)\./).to_a[1])
+  if bead_count.count >= 10
+    fput 'kneel'
+    fput 'meditate my prayer chain'
+    pause 92
```
drboldwin: will likely change this pause and rearrange some things in theurgy so that this isn't wasted time.

theurgy.lic
```diff
@@ -167,7 +211,9 @@ until ['flawless performance to those on high', 'In your condition'].include? DR
   DRC.fix_standing
 end
 
-fput "get #{$settings.flint_lighter}", 'You get a', 'You are already'
+while checkright != "#{$settings.flint_lighter}"
+  fput "get #{$settings.flint_lighter}", 'You get a', 'You are already'
+end
 get_from_container('incense')
```
drboldwin: added this while around the get flint_lighter because theurgy (not often) but sometimes bugs out on me here.  It somehow was missing grabbing this (would get ...wait twice or so and it would move on for some reason) and then spam the room forever with trying to light incense with no lighter.  I thought this would be a simple way to keep it in that spot until it is sure you have it.

rpherbig: The safer way to do this is to change `fput` to `bput`, but that requires you to capture some messaging.

drboldwin: will look into doing this instead

theurgy.lic
drboldwin: I changed to immortal_aspect - I hope this is clearer.

drboldwin:
Added checks against whether or not you have the required things (ie Prayer Chain/Knife/Shaper as well as messages indicating what you are missing.  Also added functionality to support worn and stowed prayer chain and pilgrim badge.

drboldwin:
Most current iteration - Still needs testing but possible this is complete now.

theurgy.lic
```diff
@@ -55,13 +55,129 @@ def buy_wine
...
+  fput("sprinkle #{$settings.water_holder} on my limb")
+  put_in_container($settings.water_holder)
+  if 'interrupt your research' == DRC.bput('prep bless', 'Bless spell', 'interrupt your research')
+    research_type = ["Fundamental", "Augmentation", "Stream", "Sorcery", "Utility", "Warding"]
+    find_current_project = DRC.bput("research status", / (\d+)% complete with a portion .*\./)
+    hold_research = research_type.find{ |hold_research| find_current_project =~ /#{hold_research}/i }
+    if hold_research == nil
+      return
+    end
+    find_current_project.split
```
rpherbig: This line does nothing

theurgy.lic
```diff
@@ -55,13 +55,129 @@ def buy_wine
...
+  put_in_container('prayer chain')
+  DRC.forage? 'limb'
+  get_from_container($settings.water_holder)
+  fput("sprinkle #{$settings.water_holder} on my limb")
+  put_in_container($settings.water_holder)
+  if 'interrupt your research' == DRC.bput('prep bless', 'Bless spell', 'interrupt your research')
+    research_type = ["Fundamental", "Augmentation", "Stream", "Sorcery", "Utility", "Warding"]
+    find_current_project = DRC.bput("research status", / (\d+)% complete with a portion .*\./)
+    hold_research = research_type.find{ |hold_research| find_current_project =~ /#{hold_research}/i }
+    if hold_research == nil
```
rpherbig: I recommend using `.nil?`

theurgy.lic
```diff
@@ -55,13 +55,129 @@ def buy_wine
...
+    echo 'No Prayer Chain Found - Skipping Bead Carve'
+    return
+  when 'already in your inventory'
+    fput 'remove my prayer chain'
+  end
+  beads = DRC.bput("look at my chain", /^Strung on to the prayer bead chain you see .*\./, 'There are currently no beads on it.')
+  bead_count = DRC.list_to_nouns(beads.match(/^Strung on to the prayer bead chain you see (.*)\./).to_a[1])
+  return if bead_count.count >= 10 
+  put_in_container('prayer chain')
+  DRC.forage? 'limb'
```
rpherbig: Would it be more efficient to forage after prepping Bless? The forage roundtime would then occur while you're waiting on Bless to prep, and you don't incur the forage roundtime (and have to dispose of the 'limb') if you decide not to interrupt your research.

theurgy.lic
```diff
@@ -55,13 +55,129 @@ def buy_wine
...
+  put_in_container($settings.water_holder)
+  if 'interrupt your research' == DRC.bput('prep bless', 'Bless spell', 'interrupt your research')
+    research_type = ["Fundamental", "Augmentation", "Stream", "Sorcery", "Utility", "Warding"]
+    find_current_project = DRC.bput("research status", / (\d+)% complete with a portion .*\./)
+    hold_research = research_type.find{ |hold_research| find_current_project =~ /#{hold_research}/i }
+    if hold_research == nil
+      return
+    end
+    find_current_project.split
+    if find_current_project[1,2].to_i >= 60
```
rpherbig: What if find_current_project is only a single digit?

theurgy.lic
```diff
@@ -55,13 +55,129 @@ def buy_wine
...
+    return
+  when 'already in your inventory'
+    fput 'remove my prayer chain'
+  end
+  beads = DRC.bput("look at my chain", /^Strung on to the prayer bead chain you see .*\./, 'There are currently no beads on it.')
+  bead_count = DRC.list_to_nouns(beads.match(/^Strung on to the prayer bead chain you see (.*)\./).to_a[1])
+  return if bead_count.count >= 10 
+  put_in_container('prayer chain')
+  DRC.forage? 'limb'
+  get_from_container($settings.water_holder)
```
rpherbig: You might consider checking for a carving knife, shaper, etc. before starting the process. `common-items` offers an `exists?` method that might help.

theurgy.lic
drboldwin: great idea

theurgy.lic
drboldwin: well if it's a single digit like 1% to .to_i drops the % and returns 1.
if it's 10% it picks up 10

theurgy.lic
drboldwin: Ooo that would be nicer for sure.  Will look into this more.

theurgy.lic
drboldwin: Ooops - this was leftover from something else I had tried before settling on this method and I forgot to remove it.  Good eye

drboldwin:
Had Gabellia and Jezzail run this for me a couple time and worked out some kinks that otherwise I wouldn't have known about.  In addition I changed the main altar you go to from crossing to kaerna - so with that I changed when you meditate at the refactory to the beginning after you run around shopping (since this will be the only time you are in town) also added a spot to steal a low skill item (provided you meet the required thievery ranks) - at the new altar things work as they used to with the addition of bathing in a pool and offering that stolen item to Damaris.  Will update shortly.

drboldwin:
Quick update for posterity - going through to make some changes based on comments now (I think I am done adding stuff)

Script now (in order)
- Walks to the bank
- Withdraws 10 Silver (because Ensure doesn't account for needing silvers)
- Tithes at Crossing Temple (This may or may not be off timer from crossing-training)
- Prays to 
- Goes Shopping
    - Buys Taffleberries (if circle 60+)
    - Buys Incense (if out)
    - Buys Sage (if out)
    - Buys Lavender (if out)
    - Buys Wine (if out)
    - Buys Flint (if out)
    - Buys a prayer parchment: gift of life (if you don't own it)
  Total Cost: 473(mats) + 500 (tithe) = 973
  Additional Cost: 6,000 (One time purchase)
- Walks to Urrem'tier Altar at Crossing Temple
- Cleans Altar
- Goes to secret room; STUDY's the wall (If Circle higher then 30 - skipped if researching)
- Walks to Cleric Reflectory; Meditates
- Walks to NTR (1030) crafts a prayer bead (If toggled on via YAML)
- Walks to 1032
    - Communes Eluned (If circle higher then 3)
    - Forages up Sirese Seed
    - Moves South - Plants seed
- Walks to Hidden Damaris Shrine in Kaerna Village
- Communes Tasmine (If circle higher then 2)
- Bathes in pool
- Meditates on bead (If toggled on via YAML)
- Prays to Meraud
- Prays on Prayer Badge
- Dances
- Waves some incense at altar
- Pours Wine on Altar
- Communes Truffinyi (If Circle 60+)
- Recites a Prayer
- Cleans Altar
- Walks to 1230 - Invokes Prayer Parchment

drboldwin:
Will of course still add a few things if anyone has good idea - technically that last Clean Altar is frivolous (can delete it) I had hoped you would be off timer by the time you got back to it but I have yet to be in any of my test runs.  That being said I may leave it because if you have ENOUGH trouble with finding a sirese seed it could potentially work.  Ah well.

Haven't gone through to address all your previous comments yet Sheltim but plan on doing that later today.

drboldwin:
Edited to move bless to before limb forage.  Got rid of some item checks and replaced with one large exists? check before carve routine (left prayer chain check in there because it had the added purpose of supporting a worn chain instead of a stowed one.  Also changed the pray merauds to reflect the favor_god: listed in YAML (should I make this a if/else check that uses a default god of favor_god: isnt provided?)

Also fixed an issue with the check for the prayer parchment (think I had it checking against the wrong line)

drboldwin:
considering adding commune kertigen if you are not level 60 yet but are over level 8.  This would pick one or the other.  Truffinyi or Kertigen

theurgy.lic
```diff
@@ -55,15 +65,191 @@ def buy_wine
...
+  waitrt?
+else
+  DRC.bput('pray meraud', 'meraud')
+  waitrt
+end  
+
+if DRStats.circle > 60
+  buy_Taffelberries if 'and see' != DRC.bput("rummage /C Taffelberries my #{$settings.theurgy_supply_container}", 'but there is nothing in there like that', 'and see')
+end
+buy_incense if 'and see' != DRC.bput("rummage /C incense my #{$settings.theurgy_supply_container}", 'but there is nothing in there like that', 'and see')
```
rpherbig: I think the `inside?` helper (in `common-items`) could be useful here.

theurgy.lic
rpherbig: Are you still planning on making this change?

theurgy.lic
```diff
@@ -120,10 +321,39 @@ get_from_container('wine')
 fput 'pour wine on altar'
 put_in_container('wine')
 
+if DRSkill.getxp('Theurgy') <= 24
```
rpherbig: I'd like to see this check moved into `comm_truff`

theurgy.lic
```diff
@@ -120,10 +321,39 @@ get_from_container('wine')
...
+
+clean_altar
+waitrt?
+
+if $settings.favor_god
+  DRC.bput("pray #{$settings.favor_god}", "#{$settings.favor_god}")
+  waitrt?
+else
+  DRC.bput('pray meraud', 'meraud')
+  waitrt
```
rpherbig: This should be `waitrt?`

theurgy.lic
```diff
@@ -120,10 +321,39 @@ get_from_container('wine')
...
+end  
+
+move 'go path'
+move 'go hole'
+waitrt?
+move 'east'
+move 'go hole'
+move 'out'
+
+DRCT.walk_to('1230')
```
rpherbig: Why 1230 in specific? Can you invoke your parchment in any room?

drboldwin: needs to be a populated room - more people the better.  I thought about 851 but 1230 was closer to the altar.

theurgy.lic
drboldwin: oops

theurgy.lic
rpherbig: I'd be inclined to use the safe room then, but I guess some people use isolated rooms for that... I can live with 1230.

theurgy.lic
drboldwin: this got changed to:

```
if DRSkill.getxp('Theurgy') <= 24 && DRStats.circle > 60
  comm_truff
elsif DRSkill.getxp('Theurgy') <= 31 && DRStats.circle > 8
  comm_kert
end
```

to give you an idea what I was going for.  I'm blanking on a good way to do this otherwise but I could manage them elsewhere.

drboldwin:
Personal Testing:
 Starting Theurgy XP - 0/34
 Starting Devotion - 13/16
-----------------------------
 Ending Theurgy XP - 34/34
 Ending Devotion - 13/16

There may be a fractional net loss/gain in devotion it's difficult to tell.  Likely can be made up for with standard crossing-training tithing.

theurgy.lic
rpherbig: The reason I suggest moving the checks into the individual commune methods is so that the 'main' method is more about flow - let each individual helper decide how it should achieve its goal.

Not having the elsif would also help if, say, you're still under 31/34 after comm_truff - you could then go and comm_kert.

theurgy.lic
drboldwin: Yup

drboldwin:
Changed that checkright loop to a bput
moved the checks for communes to their respective definitions

drboldwin:
Still have an issue with the research refresh.  If you happen to be on the last leg of research it will start a new project of the same type when it ends (after waiting for it)

Instead of the wait I use I will change this to match the different possible outcomes of finishing research.

theurgy.lic
```diff
@@ -40,11 +40,51 @@ def comm_tasmine
...
 end
 
+def comm_kert
+  return unless DRStats.circle > 8 && DRSkill.getxp('Theurgy') <= 31
+  case DRC.bput("get oil from #{$settings.theurgy_supply_container}", 'holy oil', 'some oil', 'I could not find', 'What were you referring to')
+  when 'I could not find', 'What were you referring to'
+    return
+  when 'holy oil'
+  when 'some oil'
+    if 'interrupt your research' == DRC.bput('prep bless', 'Bless spell', 'interrupt your research')
```
rcuhljr: I'd like to see this guy pulled out into a pre-emptive function that checks the result of a `research status` to give the all clear for casting or not. We'd probably need to improve crossing training research flags so it notices the cancel and restarts it when it gets a chance.

theurgy.lic
rcuhljr: or `return unless hold_research`

theurgy.lic
```diff
@@ -55,15 +95,197 @@ def buy_wine
...
+    echo 'No Prayer Chain Found - Skipping Bead Carve'
+    return
+  when 'already in your inventory'
+    fput 'remove my prayer chain'
+  end
+  beads = DRC.bput("look at my chain", /^Strung on to the prayer bead chain you see .*\./, 'There are currently no beads on it.')
+  bead_count = DRC.list_to_nouns(beads.match(/^Strung on to the prayer bead chain you see (.*)\./).to_a[1])
+  return if bead_count.count >= 10 
+  put_in_container('prayer chain')
+  if 'interrupt your research' == DRC.bput('prep bless', 'Bless spell', 'interrupt your research')
```
rcuhljr: ditto to earlier research extraction comment.

theurgy.lic
```diff
@@ -120,10 +355,40 @@ get_from_container('wine')
...
+move 'go hole'
+waitrt?
+move 'east'
+move 'go hole'
+move 'out'
+
+DRCT.walk_to('1230')
+if 'but there is nothing in there like that' != DRC.bput("rummage /C parchment my #{$settings.theurgy_supply_container}", 'but there is nothing in there like that', 'pale golden prayer parchment')
+  get_from_container('golden parchment')
+  fput 'invoke my parchment'
```
rcuhljr: There's some new specific item requirements for theurgy, namely the parchment and carving knife. We should see they get added to the docs on e-pedia?

rcuhljr:
I think this came up in a conversation with @rpherbig but I'd like to see a lot of the little one off rituals moved into methods, that way we can have a concise section of the script that reads like your `Script now (in order)` comment calling out methods for each ritual step. If you do that I'd be willing to on my own turn this into a class so we can stop using DR* references in front of all the methods.

drboldwin:
So testing done on a ~30th Cleric had some issues.

Eluned commune was not yet off cooldown when tasmine came around (this is off cooldown almost instantly for me so was surprised by it.)

Also this message from the kertigen commune confuses me and I'll have to look into it:
```
[theurgy]>commune kert
Your body tingles for a moment, but there are no injuries to bind.
```

rcuhljr:
It most likely failed the holy water sprinkle turning it into a commune hodierna to stop bleeding.

drboldwin:
```
>get oil
>
You get some holy oil from inside your duffel bag.
>
>sprinkle oil on week
You sprinkle yourself with some holy oil.
You have about 1 doses left.

>stow oil

You put your oil in your duffel bag.

>commune kert
The thick smell of ozone fills your nostrils as the air around you crackles and pops for a moment.
```

This being me doing one

theurgy.lic
drboldwin: Yeah - the parchment it purchases for you but I will make mention of it.  The carving knife/shaper/prayer chain are tied into the YAML addition immortal_aspect so will mention them in that description.

drboldwin:
Yeah I see what you are saying about the oil thing.  It just doesn't make sense to me how it could be confused when you call the commune specifically.  Will need to do more testing.

drboldwin:
I think that's what you were getting at

rcuhljr:
Yeah that's the direction I was thinking.

drboldwin:
need to clean up some more still but just wanted to get the ball rolling on that.  Also perhaps I could use a flag to determine when the next commune goes off?  The line that represents commune timer being off is: You feel fully prepared to seek assistance from the Immortals once again.

drboldwin:
Let's see - added a couple more outs if your theurgy is higher then 31/34 before communes.  I left kertigen/truffinyi untouched by this because at that point you are so close to being done you may as well grab the devotion.

theurgy.lic
drboldwin: thought I changed this already

rpherbig:
@drboldwin Have you see anything like what I saw in #782 ?

theurgy.lic
```diff
@@ -40,11 +47,84 @@ def comm_tasmine
   put_in_container($settings.water_holder)
 end
 
+def comm_kert
+  return unless DRStats.circle > 8 && DRSkill.getxp('Theurgy') <= 31
+  case DRC.bput("get oil from #{$settings.theurgy_supply_container}", 'holy oil', 'some oil', 'I could not find', 'What were you referring to')
+  when 'I could not find', 'What were you referring to'
+    return
+  when 'holy oil'
```
rpherbig: If this when is empty, it can be removed.

theurgy.lic
```diff
@@ -40,11 +47,84 @@ def comm_tasmine
...
+    fput 'get Taffelberries'
+    put_in_container('Taffelberries')
+  end
+end
+
+######################## SHOPPING ########################
+
+def check_supplies
+  if DRStats.circle > 60
+    buy_Taffelberries if 'and see' != DRC.bput("rummage /C Taffelberries my #{$settings.theurgy_supply_container}", 'but there is nothing in there like that', 'and see')
```
rpherbig: Did you look into using the `exists?` method (or `inside?`) to replace the rummages?

theurgy.lic
```diff
@@ -55,75 +135,296 @@ def buy_wine
...
+  fput 'read placard'
+  pause 1
+  fput 'order 4 from monk'
+  fput 'unroll my parchment'
+  put_in_container('golden parchment')
+end
+
+def buy_Taffelberries
+  DRCI.order_item(8280, 14)
+  fput 'get Taffelberries'
```
rpherbig: `order_item` leaves the items in hand. Why do you need to get them? Do they go to the ground after buying?

theurgy.lic
```diff
@@ -55,75 +135,296 @@ def buy_wine
...
+  fput 'rub sage'
+  pause 1
+  get_from_container('lavender')
+  fput 'rub lavender'
+  waitfor 'A Dark Grove, A Deep Pool'
+  DRC.fix_standing
+  move 'out'
+end
+
+def tithe
```
rpherbig: Where is this method used?

theurgy.lic
```diff
@@ -55,75 +135,296 @@ def buy_wine
...
-waitrt?
-fput 'wear badge', 'You put on a', 'You are already'
+######################## CUSTOM TRAVEL ########################
 
-until ['flawless performance to those on high', 'In your condition'].include? DRC.bput('dance', 'flawless performance to those on high', 'You begin to dance', 'Your actions grow', 'Your dance', 'but you falt', 'In your condition')
-  pause 1
+def enter_altar
+  DRCT.walk_to('1164')
+  move 'go shack'
+  fput 'look under blanket'
```
rpherbig: I don't think you need to do this (you can just 'go hole')

theurgy.lic
```diff
@@ -55,75 +135,296 @@ def buy_wine
...
+  move 'go hole'
+  move 'west'
+  move 'up'
   waitrt?
-  DRC.fix_standing
```
rpherbig: Are you sure you do not need to stand up? I thought move failed on not-standing.

theurgy.lic
```diff
@@ -55,75 +135,296 @@ def buy_wine
...
+  move 'go path'
 end
 
-fput "get #{$settings.flint_lighter}", 'You get a', 'You are already'
-get_from_container('incense')
-while 'nothing happens' == DRC.bput('light my incense with flint', 'nothing happens', 'bursts into flames')
+def exit_altar
+  move 'go path'
+  move 'go hole'
```
rpherbig: I thought this was `go stump`?

theurgy.lic
drboldwin: they go on the counter.

theurgy.lic
drboldwin: tithe gets called at the beginning when you are by the crossing temple.

theurgy.lic
drboldwin: that may work also - it's the hole in the stump I think

theurgy.lic
drboldwin: not yet - been distracted this weekend and then started working on using flags to let commune run more dynamically (since static positions are hard to guarantee the person is off the commune timer.

theurgy.lic
drboldwin: yeah I don't think you need to either - I wasn't sure when first adding it if you needed to look first but you don't.

theurgy.lic
drboldwin: probably not.  I think this assuming I'm at the right spot in my head is after the bathe - I had expected to be laying down (added the entire section while hunting) but you don't actually get put prone from this.

theurgy.lic
drboldwin: kk done deal :+1:

drboldwin:
This still needs testing as a whole but I believe it will work.  Will test further shortly.

drboldwin:
I actually think this is done in a really cool way now.

drboldwin:
Feedback from testers:

Eluned no longer creates more water then is consumed at low levels - need to add responses for being out of holy water in water_holder

Hold_research has been causing issue since changing to proc do's (not sure why since it remains in a standard method) trying to resolve this.

drboldwin:
Alright so any method that will require holy water checks to make sure you still have some in your water_holder, if you don't it skips it.

drboldwin:
Alright - I moved the DRCT.walk_to '5846' into the study_wall method (since that's where it should be anyways) and got rid of the first altar clean (you were only benefiting from 1 anyways - this way you never even enter the temple unless you are circle 30+)

theurgy.lic
```diff
@@ -6,45 +6,120 @@ $settings = get_settings
...
+def comm_eluned
   fput "stow #{GameObj.right_hand.noun}" if GameObj.right_hand.name != 'Empty'
   fput "stow #{GameObj.left_hand.noun}" if GameObj.left_hand.name != 'Empty'
 
   DRC.forage? 'dirt'
 
   get_from_container($settings.water_holder)
-  fput 'commune eluned'
+  if 'completed this commune too recently' == DRC.bput('commune eluned', 'completed this commune too recently', 'You grind some dirt in your fist')
+    Flags['theurgy-eluned'] = true
```
rcuhljr: You shouldn't be manually setting flags.

theurgy.lic
```diff
@@ -55,75 +130,357 @@ def buy_wine
...
+  echo Flags['theurgy-kertigen']
+  if Flags['theurgy-commune']
+    Flags.reset('theurgy-commune')
+    next_commune
+  end
+  method.call
+end
+
+def next_commune
+  if Flags['theurgy-eluned'] == false && DRStats.circle > 3 && DRSkill.getrank('Outdoorsmanship') >= 20 && DRSkill.getxp('Theurgy') <=31
```
rcuhljr: Unless I'm wrong we won't be running any of these communes twice, we could just determine the valid list of communes at the beginning of the file by pushing each one into the array if the requirements are met, then we can just call `@commune_list.shift.call if @commune_list.any?` in the next commune method. I think we're kind of setting ourselves up for failure by trying to make this training more modular than it needs to be. For example commune eluned has the implicit assumption that you are somewhere that you can forage dirt, but it just relies on that happening through loose structure instead of just doing commune_eluned once at the beginning when you know you're in a dirt room.

theurgy.lic
```diff
@@ -55,75 +130,357 @@ def buy_wine
...
+    else
+      echo "stopping research"
+    end
+    fput 'prep bless'
+  end
+  echo $hold_research
+  return $hold_research
+end
+
+def controller(method)
```
rcuhljr: This should probably be named something more informative like `check_communes`

rcuhljr: And change the arg named method to prevent confusion with the ruby method named method.

theurgy.lic
```diff
@@ -55,75 +130,357 @@ def buy_wine
...
+  echo Flags['theurgy-commune']
+  echo Flags['theurgy-eluned']
+  echo Flags['theurgy-tasmine']
+  echo Flags['theurgy-truffenyi']
+  echo Flags['theurgy-kertigen']
+  if Flags['theurgy-commune']
+    Flags.reset('theurgy-commune')
+    next_commune
+  end
+  method.call
```
rcuhljr: assuming you pass in a symbol, method(sym).call
```
irb(main):001:0> def test1
irb(main):002:1> puts "1"
irb(main):003:1> end
irb(main):004:0> def test2
irb(main):005:1> puts "2"
irb(main):006:1> end
irb(main):009:0> [:test1, :test2].each{|sym| method(sym).call }
1
2
=> [:test1, :test2]
```

theurgy.lic
```diff
@@ -55,75 +130,357 @@ def buy_wine
...
+    comm_truff
+  elsif Flags['theurgy-kertigen'] == false && DRStats.circle > 8 && DRSkill.getxp('Theurgy') <=31
+    comm_kert
+  end
+end
+
+######################## FLAGS ########################
+  
+Flags.add('theurgy-commune', 'fully prepared to seek assistance from the Immortals once again', 'completed this commune too recently')
+Flags['theurgy-commune'] = true
```
rcuhljr: Flags are for monitoring text coming from the game, you should be using another meaning for booleans you wish to set and track yourself.

theurgy.lic
```diff
@@ -55,75 +130,357 @@ def buy_wine
...
+controller(find_seed)
+controller(plant_seed)
+controller(enter_altar)
+controller(bathe)
+controller(meditate_bead)
+controller(pray_badge)
+controller(dance)
+controller(incense_wine)
+controller(recite_prayer)
+controller(clean_altar)
```
rcuhljr:
```
ritual_order = [:tithe, :study_wall, :reflectory, ...]
ritual_order.each do |ritual|
  check_communes
  method(ritual).call
end
```

theurgy.lic
drboldwin: Yeah - I had thought about that with eluned and so it calls it first outside the temple (where dirt is foragable) - there are only a couple possible outcomes - it completes the commune and it no longer will be called (all other communes can be done anywhere) or the commune fails because it was done to recently - at which point it turns it off anyways.  So it theoretically shouldn't ever come up anywhere but that first spot.  But I could just hard code that commune there also (I had done that before anyways)

I did the checks with the flags because I wanted it to be able to toggle between kertigen and truffenyi in a way that would skip truffenyi at times (based on field XP) and move straight to kertigen and if I kept those checks in the communes themselves they would get called but perhaps not executed.

theurgy.lic
drboldwin: will do

theurgy.lic
drboldwin: Hmm.  I did it this way because if I added the line completed this commune too recently to the flag it would trigger on all commune flags (since they share the same line).  I couldn't find any messaging that was unique between them for failed attempts.  I can try to look into a better way of doing this.

theurgy.lic
drboldwin: Yeah - I needed the functionality of the flag but I also want it to start true not false because it turns whenever it sees you are ready for a new commune (but at the start you are already ready)

theurgy.lic
drboldwin: ahh I see what you did here.  That makes sense.

theurgy.lic
drboldwin: gotcha

theurgy.lic
drboldwin: Is there a way to initiate the flag when I create it instead?

rcuhljr: Except it may not actually be true when you start, since they may have communed manually before then for some reason. If you want to set the commune flags to an initial value generally the right idea is to set up all your flags, then force the game to send the appropriate text, in this case send a `commune sense` and let all of your flags be set appropriately.

drboldwin: Oh perfect.  Thank you (I did not even think of commune sense)

drboldwin:
So this all needs running through still but just wanted to keep things up to date so you can see some of the changes.

drboldwin:
So this is working as it should - but there's an issue with it blending with crossing-training that I don't know how to solve, specifically in regards to research.

theurgy.lic will account for researching and make adjustments so as to not disrupt it, but it appears that if research for something (let's say augmentation) finishes during theurgy.lic crossing-training doesn't know it, and will start researching augmentation again.  I'm not sure how this can be remedied.

drboldwin:
alright so here's so solution I came up with:

This would be an edit to def train_theurgy in crossing-training.  It just resets your research flags if you aren't researching anything.  I haven't tested it yet with crossing-training to see if it's the desired results but in theory it should work.
```
  def train_theurgy
    wait_for_script_to_complete('theurgy')
    if 'not researching anything' == fput('research status', 'not researching anything', 'Fundamental', 'Augmentation', 'Stream', 'Sorcery', 'Utility', 'Warding')
      Flags.reset('research-partial')
      Flags.reset('research-complete')
    end
  end
```

theurgy.lic
```diff
@@ -55,75 +122,359 @@ def buy_wine
...
+  end
+  echo $hold_research
+  return $hold_research
+end
+
+def commune_check
+  echo Flags['theurgy-commune']
+  if Flags['theurgy-commune'] == false
+    next_commune
+  elsif Flags['theurgy-commune'].include? "fully prepared to seek assistance from the Immortals once again"
```
drboldwin: try as I might I can't get these to work together as one (return unless... or statement) so set it up this way for now since it works.  Willing to change to a nicer format but I just needed something stable.

theurgy.lic
drboldwin: There is more protection here now - as flags are set for communes off a commune sense done at the start of theurgy (right before communes start being called) so if Eluned can't be done due to timer/cooldown it will be flagged that way and skipped.

PR #721, opened 2016-09-03
Consume base-spells in common-arcana

rpherbig:


combat-trainer.lic
```diff
@@ -1830,11 +1814,11 @@ class GameState
   end
 
   def charge_cambrinth(next_charge)
```
rpherbig: Given this change, I'm debating removing `charge_cambrinth` from GameState.

combat-trainer.lic
```diff
@@ -1830,11 +1814,11 @@ class GameState
...
-    charge(@cambrinth, next_charge, @charge_messages)
+    charge(@cambrinth, next_charge)
   end
 
   def invoke_cambrinth
```
rpherbig: Given this change, I'm debating removing `invoke_cambrinth` from GameState.

common-arcana.lic
```diff
@@ -25,28 +25,32 @@ module DRCA
...
+      charge_and_invoke(settings.cambrinth, spell['cambrinth'])
 
       stow_cambrinth(settings)
     end
     waitcastrt?
-    DRC.bput('cast', settings.cast_messages)
+    DRC.bput('cast', get_data('spells').cast_messages)
   end
 
   def find_cambrinth(settings)
```
rpherbig: This would be my next target (i.e. to remove settings and pass in real args).

It occurred to me that if I invert the order of the conditionals (i.e. check stored first, then held), this method doesn't even need `held_cambrinth`... is that too risky?

PR #738, opened 2016-09-05
Updated Ambush Stun in combat-trainer

drsuik:
updated Ambush Stun in combat-trainer to use two new YAML settings. 

stun_weapon is the actual weapon to use: IE broadsword
stun_skill is either Debilitation or Backstab

This will allow thieves to use either an edged or blunt weapon to Ambush Stun. Edged weapons work but they don't do head damage like blunts do. This will also allow thieves to pick which skill to train with Ambush Stun. I, personally, found it better to stop on a primary skill over a tertiary.

Should probably also remove the aimed_skill check in line 1843; however I haven't had any issues.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -946,7 +946,7 @@ class TrainerProcess
 
   def initialize(settings)
     echo('New TrainerProcess') if $debug_mode_ct
-
+	@stun_skill = settings.stun_skill
     @training_abilities = settings.training_abilities({})
     echo("  @training_abilities: #{@training_abilities}") if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -946,15 +946,15 @@ class TrainerProcess
 
   def initialize(settings)
     echo('New TrainerProcess') if $debug_mode_ct
-
+	@stun_skill = settings.stun_skill
```
rpherbig: If this is only used on line 957, then it does not need to be an instance var. You could make it local to this method (`stun_skill`, no @), or inline `settings.stun_skill` on line 957. I think I'm in favor of the inlining.

combat-trainer.lic
```diff
@@ -964,7 +964,7 @@ class TrainerProcess
     @no_app = []
 
     @dont_stalk = settings.dont_stalk
-
+	@stun_weapon = settings.stun_weapon
```
rpherbig: This needs an echo after it (similar to line 951). While you're in here, could you add two more echos after lines 968 and 969?

combat-trainer.lic
```diff
@@ -1055,15 +1055,15 @@ class TrainerProcess
     dead_count = DRRoom.dead_npcs.size
 
     EquipmentManager.instance.stow_weapon(game_state.weapon_name)
-    EquipmentManager.instance.wield_weapon(game_state.ambush_stun_weapon, 'Small Blunt')
+    EquipmentManager.instance.wield_weapon(@stun_weapon)
```
rpherbig: You'll still need to pass in a weapon skill here. If the `@stun_weapon` is swappable, `equipmanager` needs to know which skill it should use.

rpherbig:
The two new settings should have default values in `base.yaml`.

combat-trainer.lic
drsuik: I will use inline. thanks.

combat-trainer.lic
drsuik: done and done

combat-trainer.lic
drsuik: I'll set this up with another instance variable and then try to think of a way to do it without adding another one.

PR #739, opened 2016-09-05
Reopened: Updated Ambush Stun in combat-trainer

drsuik:
updated Ambush Stun in combat-trainer to use new YAML settings.

stun_weapon is the actual weapon to use: IE broadsword
stun_weapon_skill is the skill this weapon trains or the skill you want to swap it to train
stun_skill is either Debilitation or Backstab

This will allow thieves to use either an edged or blunt weapon to Ambush Stun. Edged weapons work but they don't do head damage like blunts do. This will also allow thieves to pick which skill to train with Ambush Stun. I, personally, found it better to stop on a primary skill over a tertiary.

Should probably also remove the aimed_skill check in line 1843; however I haven't had any issues.

Had to close the last pull request and open this one. Sorry. This should address all issues from #738.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -946,7 +946,6 @@ class TrainerProcess
 
   def initialize(settings)
     echo('New TrainerProcess') if $debug_mode_ct
-
     @training_abilities = settings.training_abilities({})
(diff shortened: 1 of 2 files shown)


rpherbig:

PR #740, opened 2016-09-05
added Riverhaven shaping workorders via hometown setting

ericmeek:
Added Riverhaven section to Shaping section in base-crafting.yaml
Modified shape.lic to account for this using hometown setting in base.yaml
Modified workorders.lic to account for this using hometown setting in base.yaml
Added practice longbow recipe
Modified workorders.lic to buy bow string for practice longbow

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -28,7 +28,8 @@ class WorkOrders
     settings = get_settings
     crafting_data = get_data('crafting')
     @bag = settings.crafting_container('backpack')
-    info = crafting_data[discipline]
+    @bank_info = get_data('town')[settings.hometown]['deposit']['id']
+    info = crafting_data[discipline][settings.hometown]
 
     @recipes = crafting_data.crafting_recipes.select { |recipe| recipe['work_order'] && recipe['type'] =~ /#{discipline}/i }
 
@@ -163,7 +164,7 @@ class WorkOrders
   end
 
   def carve_items(info, item, quantity)
-    ensure_copper_on_hand(5000)
+    ensure_copper_on_hand(5000, @bank_info)
     recipe, items_per_stock, spare_stock, scrap = find_recipe(info, item, quantity)
 
     quantity.times do |count|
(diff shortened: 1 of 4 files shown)


workorders.lic
```diff
@@ -28,7 +28,8 @@ class WorkOrders
     settings = get_settings
     crafting_data = get_data('crafting')
     @bag = settings.crafting_container('backpack')
-    info = crafting_data[discipline]
+    @bank_info = get_data('town')[settings.hometown]['deposit']['id']
+    info = crafting_data[discipline][settings.hometown]
```
rpherbig: Will this work for Forging/Tailoring/Carving (i.e. the disciplines that aren't yet by-town)?

ericmeek: No, there would have been a problem and as I was looking at the easiest way
to correct this I decided on another approach. I'm looking at all the
crafting scripts that have room #'s and such hard coded in and fixing them
so they pull that info from the base-crafting yaml.

I'll make sure all of the scripts work from Crossing and then it'll be as
simple as updating the yaml for other locations. Please reject that pull
request.

rcuhljr:
We need to expand crafting information a bit anyways. We should have part names listed under a vendor location, because armorsmithing uses parts from the outfitting store.

ericmeek:
Good point and I've got a yaml going with a section separate for parts.
Another thing, the specific scripts to create an item are not consistent
between crafts when it comes to getting items. For instance, workorders
buys parts for the forge script but the shaping script buys parts on it's
own.

I can move the buying of parts all into either the individual scripts or
all into the workorders scripts if you think that'd be a good idea.

rpherbig:
I have a vision for crafting that looks (approximately) like:
- There is a script to handle the mechanics of the actual crafting (forge, sew, carve, etc.). This script takes very raw inputs (page number, metal type, etc.) and makes some assumptions (you already have the ingot, oil, etc.). This keeps that logic self-contained.
- There is another script which handles the mechanics of getting ready to craft (smith is an example of this). It takes more user-friendly inputs (item name, etc.) and handles buying raw materials (ingots), parts, etc. This is the one I expect users to actually execute.
- Finally workorders would only handle the mechanics of an actual workorder. Finding the order giver, repeatedly asking for a workorder until we find one we like, bundle items up after crafting, and finally turning in the logbook. But for the actual crafting, it would call into one of the scripts above.

I think forge->smith->workorders is a pretty good example of this vision, but I haven't had time to do the same with the other crafts or think of a better way to do this.

PR #746, opened 2016-09-06
modified stow_lootables function in combat-trainer

drsuik:
Modified stow_lootables function in combat-trainer to attempt to stow loot items in the picking_box_storage container set in YAML before declaring it unlootable and dropping it.  If picking_box_storage is not set or unable to store in set container, it finishes declaring it unlootable and dropping it.

I set this up mainly for boxes, as pick.lic will look in pick_box_storage after picking_box_source is empty. I don't see it unsustainably filling the container with loot. However, perhaps a function somewhere to clean up this container might be necessary. IE a fill my gempouch with my pick_box_storage?

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -191,10 +191,13 @@ class LootProcess
 
     @spare_gem_pouch_container = settings.spare_gem_pouch_container
     echo("  @spare_gem_pouch_container: #{@spare_gem_pouch_container}") if $debug_mode_ct
-
+    
 	@picking_box_storage = settings.picking_box_storage
 	echo("	@picking_box_storage: #{@picking_box_storage}") if $debug_mode_ct
-
+	
+	@store_extra_loot = settings.store_extra_loot
+	echo("	@store_extra_boxes: #{store_extra_loot}") if $debug_mode_ct
+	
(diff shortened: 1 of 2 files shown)


rcuhljr:
My biggest problem with this is that filling up both bags will cause severe problems, you'll be unable to take gear off successfully in order to lockpick the boxes filling up the entire inventory for example.

drsuik:
That makes sense. Also, people who are harvesting boxes while training may not go through them quickly enough to clear out their containers. And burden could become an issue with lower circle characters. Would the ability to set a overflow box container separate from someone's main stow container be desirable(too keep it free for armor and such)? Coming with a burden warning in the description? IDK, I'll close this if the potential for severe problems can't be sorted out as this would be a niche benefit anyways

rcuhljr:
I'm just positing that if you want to pursue this it should be an opt in feature in someones yaml.

drsuik:
I added a store_extra_loot setting to base.yaml to drive this option. Its default is false.

drsuik:
Bah, pick.lic doesn't look in pick_box_storage for unopened boxes. Will revisit

rpherbig:
Should this be closed, or are you pursuing other ideas?

PR #751, opened 2016-09-06
Craft changes

ericmeek:
Workorders for all disciplines work with hometown setting now (Crossing & Riverhaven)
Parts has it's own section (recipe_parts) in base-crafting.yaml
All room #'s taken out of crafting scripts and pulled from base-crafting.yaml
Carve script updated with get_item/stow_item same as shape script
ensure_copper_on_hand in common-money updated to check hometown

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -12,9 +12,9 @@ module DRCC
     ['nothing in there'].include?(DRC.bput('look in cruc', 'nothing in there', 'you see'))
   end
 
-  def find_empty_crucible
+  def find_empty_crucible(idle_room = nil, crucibles)
     DRCT.find_empty_room(
-      DRC.sort_destinations([8774, 19_030, 8773, 8777, 8779]), 8775, proc { DRRoom.pcs.empty? && empty_crucible? }
+      DRC.sort_destinations(crucibles), idle_room, proc { DRRoom.pcs.empty? && empty_crucible? }
     )
   end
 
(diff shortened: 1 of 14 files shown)


common-crafting.lic
```diff
@@ -12,26 +12,26 @@ module DRCC
     ['nothing in there'].include?(DRC.bput('look in cruc', 'nothing in there', 'you see'))
   end
 
-  def find_empty_crucible
+  def find_empty_crucible(idle_room = nil, crucibles)
```
rpherbig: Would it be any easier if you passed in a hometown string as an argument and had these helpers do the lookup?

ericmeek: Yes, I should have the lookups in the common- files. Changing that.

ericmeek: Actually I think I'll have the default argument lookup the hometown and it
can be overridden if in some case someone wanted to. Although I can't think
of a reason anyone would. See any issues with that?

rpherbig: I'm not sure what you mean. Something like `def foo(hometown='Crossing')`?

ericmeek: This:

def foo(hometown = get_settings["hometown"])

Could be done a couple different ways as long as it can be called with
defaults or be overridden.

rpherbig: I'm not a big fan of calling get_settings from the helpers since it won't account for args that were passed to the script which load other setting files.

ericmeek: Ahh, didn't realize that. Hometown string it is.

ericmeek:
Helpers in common-crafting.lic now take hometown string as argument
Updated ensure_copper_on_hand method
Updated workorders.lic and individual crafting scripts re: changes in common-crafting
Tested workorders for all disciplines in Crossing & Haven

rpherbig:
It looks like this cannot be automatically merged due to conflicts. Can you merge master back into your branch and fix the conflicts?

forge.lic
```diff
@@ -91,7 +92,12 @@ class Forge
     turn_to("page #{find_recipe(10, 'metal weapon honing')}")
     bput('study my book', 'Roundtime')
     stow_item('book')
-    find_grindstone
+    
+    settings = get_settings
```
rpherbig: Are lines 96, 97, 98 still necessary?

See also lines 111, 112, 113, 307, 308, 309.

makesteel.lic
```diff
@@ -1,13 +1,17 @@
-
-
 custom_require.call(%w(common common-crafting common-items common-money drinfomon))
 
 count = variable.last.to_i
 
-DRCM.ensure_copper_on_hand(2000 * count)
+get_settings["hometown"]
```
rpherbig: Can this line be removed?

workorders.lic
```diff
@@ -250,7 +261,9 @@ class WorkOrders
     stock_needed = ((quantity * recipe['volume'] - existing) / 100.0).ceil
     order_yarn(info['stock-room'], stock_needed, info['stock-number'])
 
-    find_sewing_room
+    data = get_data('crafting')['tailoring']['Riverhaven']
```
rpherbig: Why is Riverhaven hardcoded here?

workorders.lic
```diff
@@ -250,7 +261,9 @@ class WorkOrders
...
+    find_sewing_room(@hometown)
+    #find_sewing_room(@engineering_room)
```
rpherbig: Can this line be removed?

common-money.lic
```diff
@@ -33,16 +33,18 @@ module DRCM
...
 
-  def wealth
-    fput 'wealth'
-    waitfor('Wealth:')
-    wealth = matchfind '\\(? copper Kronars\\)\\.', 'No Kronars?'
-    wealth == '.' ? 0 : wealth.to_i
+  def wealth(hometown)
+    case hometown
+    when "Riverhaven"
+      wealth = DRC.bput("wealth lirum", '\(\d+ copper Lirums\)', "No Lirums").scan(/\d+/).first.to_i
```
rpherbig: You've got a map between hometown and its currency. That data is already in `base-town` (hometown['exchange']['currency']).

I think we should move currency up a level so it is a property of the town itself, but that's not necessary in this PR.

PR #752, opened 2016-09-06
Missed the alchemy belt in the tool repair section.

Dartellum:


workorders.lic
```diff
@@ -79,6 +79,7 @@ class WorkOrders
     when 'remedies'
       @alchemy_room = settings.alchemy_room
       tools = settings.alchemy_tools([])
+      @belt = settings.alchemy_belt(nil)
```
rpherbig: Where is `@belt` used?

It should probably default to `[]` instead of `nil`.

PR #754, opened 2016-09-07
Added another message found by a tester and added stow_hands in case …

Dartellum:
…something stayed in hand.


remedy.lic
```diff
@@ -56,6 +56,7 @@ class Remedy
   end
 
   def create_item
+    stow_hands
```
rpherbig: You need to put `common-items` in the `custom_require` call; and put `include DRCI` at the top.

Dartellum: Added. Thanks for the catch!

-- 
Regards,
484-300-4833 (h)
484-431-3355 (m)
484-302-7006 (f)

PR #755, opened 2016-09-07
remove 19063

jandersson:


weave-cloth.lic
```diff
@@ -3,7 +3,7 @@
 =end
 
 DRCT.walk_to(19_065)
-DRCT.walk_to(19_063) unless DRRoom.pcs.empty?
+# DRCT.walk_to(19_063) unless DRRoom.pcs.empty?
```
jandersson: This is commented instead of deleted as I do not do outfitting, and this is possibly needed in the future (with a new room number)? Tell me if i should delete it.

rcuhljr: Nah it's fine as is. Just a personal use script that's not very good.

Issue #667, opened 2016-08-25
Refactor LootProcess.dispose_body with respect to Necromancer Rituals


jandersson:
While working on #657 I found that `dispose_body` is handling necromancer rituals. I think if necromancer rituals are given their own handler then `dispose_body` becomes more readable. The method `check_thanatology` is performing rituals, so I would propose using `check_thanatology` to act as a handler for rituals, and a new method `do_ritual` for performing them.

rpherbig:
@jandersson Are you working on this?

jandersson:
Yes, I'll get started on this

PR #767, opened 2016-09-10
Refactor ritual performance in combat trainer

jandersson:
Addressing #667


combat-trainer.lic
```diff
@@ -186,12 +186,15 @@ class LootProcess
     @heal = thanatology['heal'] || false
     echo("  @heal: #{@heal}") if $debug_mode_ct
 
+<<<<<<< HEAD
```
rpherbig: These lines (and line 197) should be removed.

jandersson: Somehow missed those!

jandersson:
I'm not sure about returning a boolean from do_ritual anymore and I cant remember what my plan was with that.

combat-trainer.lic
```diff
@@ -295,73 +295,74 @@ class LootProcess
...
-    elsif @skin && !%w(consume harvest dissect).include?(@last_ritual)
-      arrange_mob(DRRoom.dead_npcs.first, game_state)
-      check_skinning(DRRoom.dead_npcs.first, game_state)
-    end
-    unless healing
-      while 'and get ready to search it' == bput('loot', 'You search', 'I could not find what you were referring to', 'and get ready to search it')
-        pause
-        waitrt?
+
+      if @ritual_type == 'harvest' && ! game_state.prepare_consume
```
jandersson: Are these compound if statements clear? Should nest all other possible rituals inside a single `unless game_state.prepare_consume` instead?

combat-trainer.lic
```diff
@@ -295,73 +295,74 @@ class LootProcess
...
-  def check_thanatology(mob_noun, ritual, game_state)
-    return false if game_state.construct?(mob_noun)
+  def do_ritual(mob_noun, ritual, game_state)
+    return unless DRStats.necromancer?
+    return if game_state.construct?(mob_noun)
 
     perform_message = "perform #{ritual} on #{mob_noun}"
-    case bput(perform_message, 'Rituals do not work upon constructs', 'Rituals do not work upon the undead', 'This ritual may only be performed on a corpse', 'precise motions with your ritual knife', 'corpse to make one long', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'roundtime')
-    when 'corpse to make one long', 'carefully position the corpse', 'corpse and make a few quick precise cuts', 'Your cuts follow the vital lines of the corpse'
+    case bput(perform_message, 'Rituals do not work upon constructs', 'Rituals do not work upon the undead', 'This ritual may only be performed on a corpse', 'precise motions with your ritual knife', 'corpse to make one long', 'carefully position the corpse', 'roundtime', 'Your cuts follow the vital lines of the corpse')
```
jandersson: I was also thinking it might be useful to build a hash map of rituals to messages because it is not exactly clear what rituals are being checked here.

rpherbig: Would `base-spells` be a good place to put this messaging?

jandersson: It makes sense from the point of view that it is a supernatural act

jandersson: I'll put the messages in there with some structure.. that might help #772 because that will also use rituals

combat-trainer.lic
```diff
@@ -177,9 +177,12 @@ class LootProcess
     echo("  @lootables: #{@lootables}") if $debug_mode_ct
 
     thanatology = settings.thanatology({})
-    @ritual_type = thanatology['ritual_type'] || {}
+    @ritual_type = thanatology['ritual_type'].downcase || nil
```
rcuhljr: Or nil doesn't do anything.

combat-trainer.lic
```diff
@@ -295,73 +298,76 @@ class LootProcess
...
+    return unless @loot_bodies
+    if DRRoom.dead_npcs.empty?
+      @last_ritual = nil
+      return
+    end
+    return if Time.now - @loot_timer < @loot_delay
+
+    game_state.mob_died = true
+    check_rituals(game_state)
+    if @skin && !%w(consume harvest dissect).include?(@last_ritual)
```
rpherbig: Would it be clearer to have `check_rituals` return a bool which indicates if skinning can occur?

jandersson: that might be useful, ill give it a try

combat-trainer.lic
```diff
@@ -295,73 +298,76 @@ class LootProcess
...
-      arrange_mob(DRRoom.dead_npcs.first, game_state)
-      check_skinning(DRRoom.dead_npcs.first, game_state)
-    end
-    unless healing
-      while 'and get ready to search it' == bput('loot', 'You search', 'I could not find what you were referring to', 'and get ready to search it')
-        pause
-        waitrt?
+
+      unless game_state.prepare_consume
+        if @ritual_type == 'harvest'
```
jandersson: That rewrite seems to make sense. The array might work as well, but ill try it in either case and see how it goes.

combat-trainer.lic
jandersson: I'm trying to think about how this would work with a list. Is there a sensible way to define do_ritual to work on both a string or a list as input? Most situations will call for a single ritual, whatever the user has defined to be done, so that would be passed as a string. However some rituals require a preserve ritual to be done first, so that could be passed as an array.

A second thought: do ritual can just check if the ritual arg is 'harvest', 'arise', or 'consume' (the rituals that make use of preserve) then apply preserve on its own

rpherbig: A list can have a single item. But I'm also ok with the second option.

jandersson: I'm a fan of the second option, i think it makes sense if do_ritual is responsible for knowing if some other ritual is a prereq for whatever its trying to do

rpherbig: I'm sold on the second option. If ritual B always requires ritual A, then that knowledge is more of a game mechanic than configuration (and should be in the script).

rpherbig:
@jandersson Looks good to me. Are you planning to make any other changes?

jandersson:
I am good with this if you guys are, its working for me - same as before.

PR #786, opened 2016-09-13
Crafting

ericmeek:
created crafting.lic. can be called for smithing and shaping. orders material/parts, finds room.
workorders.lic calls crafting.lic for smithing and shaping.
smith.lic not modified, can still be called.
ordering of mats/parts for shaping removed from workorders.lic.
readded longbow stuff to shape.lic.
added "lumber" to material for shaping to be consistent with smithing having "ingot"

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -177,8 +177,11 @@ class LootProcess
     echo("  @lootables: #{@lootables}") if $debug_mode_ct
 
     thanatology = settings.thanatology({})
-    @ritual_type = thanatology['ritual_type'] || {}
+    @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
+
+    @rituals = get_data('spells').rituals
+    echo("  @rituals: #{@rituals}") if $debug_mode_ct
 
     @last_ritual = nil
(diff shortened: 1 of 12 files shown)


ericmeek:
I completely forgot to take all my little echo's out. Will remove those and commit again.

ericmeek:
Thought before I move forward on some stuff. As I was adding carving to crafting.lic and realized I had to account for stone and bone I had the thought that maybe every material in a product should be in the recipe section of that material, like parts are. Material is just another part, no?

The upside to this is that it cuts down the number of material sections in the discipline section of a craft. The downside is that I'll have to update the parts list for every recipe. Not a problem, actually, and I think in the long run it'd be a positive. Thoughts?

ericmeek:
Disregard all that. Not too many options to choose from. This is what I'm going for. Unless I change my mind....

crafting "slender metal rod"
    fail if huge bronze ingot (with count) not on person

crafting "slender metal rod" buy
    would buy huge bronze metal ingot

crafting "slender metal rod" bronze/steel/whatever
    fail if "whatever" ingot isn't present
    start crafting, fail if mechanics says ingot is too small

ericmeek:
Back in the states! I have some merging and catching up to do I see and then I'm back to work on crafting.

PR #787, opened 2016-09-13
Shard Messaging for Safe-Room

jandersson:
Removed the `action: lie` bit, since it seems like the mechanics are the same with this autohealer


jandersson:
Also the messaging is great, heres the full bit 
```
Quentin snickers at you.  "Get up, get up, go have yourself a birthday party 'cause you don't need healing!"
You sit up.
```

PR #813, opened 2016-09-16
Minor fix for using town data

jandersson:
Ensure copper on hand has a default argument for Crossing, so if you were in Shard it would run to the crossing bank


jandersson:
Dammit, and the other commit snuck in there..

Issue #620, opened 2016-08-16
Analyze Flame


jayRyan24:
As a Barbarian main, I find it incredibly difficult (as do most people) to properly maintain the IF needed to train all four of the magics we need. I propose that the combat trainer has a setting use_analyze_flame: true/false setting that let's Barbarian Analyze Flame in combat and then use the sequence of attacks lists. The message is always the same and I can provide that if Torg isn't available to. This would actually be an amazing addition as with it we can continuously maintain 4-5 abilities to train everything needed (and effectively make us even more monsters in combat). I know for some creatures, we actively need multiple abilities maintained to fight them. For instance, I had issues where I needed a magical defense and also physical defense running. Pleae let me know if I can be of any help!

jayRyan24:
OK, I added this myself. It uses the game's existing paradigm. It requires the setting `use_analyze_flame: true` in the YAML, using `use_weak_attacks` will override it. 

`=begin
  Documentation: https://elanthipedia.play.net/Lich_script_repository#combat-trainer
=end

custom_require.call(%w(common common-arcana common-healing common-items common-summoning common-travel drinfomon equipmanager events spellmonitor))

class SetupProcess
  include DRC
  include DRCS
  include DRCT

  def initialize(settings)
    echo('New SetupProcess') if $debug_mode_ct

    @stance_override = settings.stance_override
    echo("  @stance_override: #{@stance_override}") if $debug_mode_ct
    @priority_defense = settings.priority_defense
    echo("  @priority_defense: #{@priority_defense}") if $debug_mode_ct
  end

  def execute(game_state)
    return true if game_state.done_cleaning_up?
    if game_state.stowing?
      echo('SetupProcess::clean_up') if $debug_mode_ct
      retreat
      if game_state.summoned_info(game_state.weapon_skill)
        if DRStats.moon_mage?
          bput('wear moon', 'telekinetic')
        else
          break_summoned_weapon(game_state.weapon_name)
        end
      else
        EquipmentManager.instance.stow_weapon(game_state.weapon_name)
      end
      game_state.next_clean_up_step
      return true
    end

    was_retreating = game_state.retreating?
    game_state.update_room_npcs
    if game_state.dancing?
      game_state.dance
    elsif game_state.retreating?
      determine_next_to_train(game_state, game_state.retreat_weapons, false)
    else
      determine_next_to_train(game_state, game_state.weapon_training, was_retreating)
    end

    if game_state.parrying
      check_stance(game_state)
[...]

PR #828, opened 2016-09-18
Add support for analyze flame

rpherbig:
See #620

Contributed by @jayRyan24


combat-trainer.lic
```diff
@@ -1910,14 +1919,38 @@ class GameState
...
+    elsif use_analyze_flame?
+      update_analyze_flame_array if @analyze_flame_array.empty?
+      @analyze_flame_array.shift
     else
       'attack'
     end
   end
 
+  def update_analyze_flame_array
+    result = DRC.bput('analyze flame', 'Analyze what\?', 'by landing a .*')
```
rpherbig: You could `include DRC` on line (roughly 1512) so you don't need to prefix these methods.

combat-trainer.lic
```diff
@@ -1910,14 +1919,38 @@ class GameState
...
+
+    case result
+    when 'Analyze what?'
+      fput('face next')
+      return []
+    end
+
+    text = result.match(/by landing a (.*)\.$/).to_a[1]
+    @analyze_flame_array = DRC.list_to_nouns(text)
```
rpherbig: You could `include DRC` on line (roughly 1512) so you don't need to prefix these methods.

combat-trainer.lic
```diff
@@ -1910,14 +1919,38 @@ class GameState
...
+  def stop_analyze_flame
+    @use_analyze_flame = false
+  end
+
   def melee_attack_verb
     if use_weak_attacks?
       brawling? ? 'punch' : 'jab'
+    elsif use_analyze_flame?
+      update_analyze_flame_array if @analyze_flame_array.empty?
+      @analyze_flame_array.shift
```
rpherbig: I was originally worried about balance and position, but now I think it will be ok as-is:
* You'll occasionally see "positive" moves in your combos
* You'll likely end up doing some bob/weave/circle occasionally to regen fatigue
* You can now put up a balance buff with the excess Inner Fire

The log you posted looks like you were fine.

combat-trainer.lic
```diff
@@ -1641,6 +1642,14 @@ class GameState
 
     @use_weak_attacks = settings.use_weak_attacks
     echo("  @use_weak_attacks: #{@use_weak_attacks}") if $debug_mode_ct
+
+    @use_analyze_flame = settings.use_analyze_flame
```
rpherbig: 1. This could be generalized to support any Expertise-specific analyze by making the new setting a string instead of a bool. nil means we do not analyze, anything else is used as part of the analyze command. I'd be in favor of making this change.

2. We could even make this usable by non-Barbarians by having a bool *and* a string. The bool is whether or not to use analyze, the string is the type, and non-Barbarians use nil for the string and true for the bool. I'm inclined to skip this unless someone requests it.

combat-trainer.lic
```diff
@@ -1910,14 +1919,38 @@ class GameState
...
 
+  def use_analyze_flame?
+    @use_analyze_flame
+  end
+
+  def stop_analyze_flame
+    @use_analyze_flame = false
+  end
+
   def melee_attack_verb
```
rpherbig: I'm worried about an edge case that I've encountered before.

I do not see anything in the "switch to a new skill" logic that prevents you from changing skills mid-combo. If the before and after skills are not compatible (i.e. use different moves), it's possible to still have a combo going which you cannot complete with your current weapon. For example, try `combat-trainer` using only Brawling and an Edged weapon.

Did you test this case?

combat-trainer.lic
```diff
@@ -1641,6 +1642,14 @@ class GameState
 
     @use_weak_attacks = settings.use_weak_attacks
     echo("  @use_weak_attacks: #{@use_weak_attacks}") if $debug_mode_ct
+
+    @use_analyze_flame = settings.use_analyze_flame
+    echo("  @use_analyze_flame: #{@use_analyze_flame}") if $debug_mode_ct
```
rpherbig: This new settings needs a default value in `base.yaml`.

combat-trainer.lic
jayRyan24: So, Offhand weapons don't work with this, never thought to specifically test this, but Analyze based offhand weapons as if they were brawling. I fixed this using the Brawling? check that already exists. 

The big issue is found thanks to your question. Basically, if you have a melee weapon in your hand and a set of brawling attack commands, it will still work and complete the set. If you don't, it will continue to try and chop, etc, with your fists and tell you a weapon would be a better choice. 

If someone calls combat-trainer manually, they can see this problem often, especially if they ended a combo that was melee weapon and went to brawling. It will eventually fix itself after about 90 seconds (I timed it). Analyze combos seem to be on a 90 second completion timer. 

Some ways to handle this would be to set a timer if the "Wouldn't it be better if you used a melee weapon?" was received. Set it for 90s and during that time and use the default setting. We can also check that if the analyze combo array isn't empty, that a weapon cannot be swapped. You could run into errors with this because if combat-trainer is started and gets a fresh array, the game will swap weapons to the next weapon to train, and if it happens to be brawling and the 90s timer hasn't reset, you will analyze again and get the melee weapon combo. I say, 90s of "attacking", isn't a bad thing to avoid the other headache.

Any other ideas?

combat-trainer.lic
jayRyan24: This is probably fine, and as GameState advances, it will probably use DRC more.

combat-trainer.lic
jayRyan24: Think 1 is fine, 2 I agree could be skipped.

combat-trainer.lic
jayRyan24: Actually, I think the Brawling and Offhand checks work for this without doing anything. It will just simply bypass the case until you are no longer brawling or using an offhand.

jayRyan24:
I added the recommended changes to the file. It now takes in a string and checks for nil. It uses that to analyze. I checked it using 4 weapons and letting them switch between them all. Crossbow, LE, Brawling, and Offhand. 

Code:  http://pastebin.com/8VrLnDRk

Combat Log:  http://pastebin.com/1N1WUapW

rpherbig:
I updated the PR with your changes. I did a bit of refactoring while I was in there (you might want to double check I didn't break anything).

I recommend you write a validator for this new setting, to make sure that only valid analyze strings are used.

jayRyan24:
I added support for this in validate.lic

```
  def assert_that_analyze_combo_is_valid(settings)
    return unless settings.analyze_combo_type
	
	if !['flame', 'intimidation'].include?(settings.analyze_combo_type)
      error("analyze_combo_type: '#{settings.analyze_combo_type}' is invalid, only flame and intimidation are supported at this time.")	
	end
  end
```

Also added the change to GameState Settings Initialize:
```
    if ['flame', 'intimidation'].include?(@analyze_combo_type)
	  @use_analyze_combo = true
	  @analyze_combo_array = []
	  echo("  analyze_combo_type: #{@analyze_combo_type}") if $debug_mode_ct
	else
      @use_analyze_combo = false	
	end
```

jayRyan24:
Important to add this change in as well. 	
	
Line: 1508:` @game_state.stop_analyze_combo`

PR #836, opened 2016-09-20
Workorders remedy updated with a function to order the herb(s).

Dartellum:
The function checks on hand and combines the herbs, if more than one present. Will look for three stacks. Next, it will count them and check if enough is on hand and purchase any additional required. Also added tilt to clean mortar before making a new if an old is stuck in it. Finally, added a check for two catalysts, since coal, pewter, electrum, etc are not combined by the combine command. This will make sure to use up a nugget with under 3 volume left.


remedy.lic
```diff
@@ -66,8 +66,15 @@ class Remedy
     stow_item('book')
     get_item(@container.to_s)
     get_item(@herb1.to_s)
-    fput("put #{@herb1} in my #{@container}")
-    stow_item(@herb1.to_s) # Added for when the herb is larger than 25 pieces
+    case bput("put #{@herb1} in my #{@container}", "You put", "already has something")
+      when 'You put'
+        stow_item(@herb1.to_s) # Added for when the herb is larger than 25 pieces
```
rpherbig: Are all of these `to_s` calls unnecessary?

workorders.lic
```diff
@@ -278,6 +278,52 @@ class WorkOrders
...
+            case bput('combine', 'You combine', 'That stack of herbs')
+              when 'That stack of herbs', 'You combine'
+                stow_hands
+            end
+          end  
+      end
+    end
+    if inside?("first #{herb}", @bag)
+      herb_have = bput("count #{herb}", 'You count out \d+ pieces').scan(/\d+/).first.to_i
+      if herb2 == "yes" and herb_have <= quantity
```
rpherbig: Why do you use 'yes' and 'no' rather than true and false?

workorders.lic
```diff
@@ -332,55 +389,26 @@ class WorkOrders
...
-        stow_tool(recipe['herb2'])
-      end
     else
-      herb2_needed = 1
-      order_rem(info['stock-room'], herb2_needed, recipe['herb2_stock'])
+      herb2 = 'yes'
+      count_combine_rem(info['stock-room'], recipe['herb2'], recipe['herb2_stock'], quantity, herb2)
     end
 
     quantity.times do
```
rpherbig: There appears to be a merge conflict. Can you merge master back in?

Dartellum: #837 is the fixed pull request.

PR #842, opened 2016-09-21
Update theurgy.lic

drboldwin:
Think this should address those 2 issues for now - maybe look into a better way of doing the crossing temple one still

### DIFF
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -272,7 +272,7 @@ class Theurgy
   def pray
     waitrt?
     if $settings.favor_god
-      bput("pray #{$settings.favor_god}", $settings.favor_god)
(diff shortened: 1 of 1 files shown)


theurgy.lic
```diff
@@ -272,7 +272,7 @@ class Theurgy
   def pray
     waitrt?
     if $settings.favor_god
-      bput("pray #{$settings.favor_god}", $settings.favor_god)
+      DRC.bput("pray #{$settings.favor_god}", $settings.favor_god, 'Lady of healing', 'reward of hard effort', 'sign with your hand', 'god of the Void', 'bane of accursed', 'blessing be upon your love', 'honorable and true', 'glory shine on us', 'fire across the lands', 'madness and pain', 'floods strike down', 'Lady of supreme beauty')
```
rpherbig: You don't need the DRC/DRCT prefixes now that this is in a class

Issue #857, opened 2016-09-26
Dedicated Camb Use


drboldwin:
So there is more we can expand on this (perhaps have it charge a cyclic camb similar to how it maintains OM?  but either way.  I'd like us to implement a way to have it invoke <camb> spell if we so choose.  I run a lot of cyclics and them draining from my camb often causes issues.  Things like trying to determine appropriate casting level for symbiosis spells (it drains and the spell thinks it successfully cast at a level it actually can't cast at.  which causes it to increase in power and backfire)

PR #864, opened 2016-09-27
Dedicated camb

drboldwin:
https://github.com/rpherbig/dr-scripts/issues/857

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1654,6 +1654,9 @@ class GameState
 
     @analyze_combo_type = settings.analyze_combo_type
     echo("  @analyze_combo_type: #{@analyze_combo_type}") if $debug_mode_ct
+	
+	@dedicated_camb_use = settings.dedicated_camb_use
+	echo(" @dedicated_camb_use: #{@dedicated_camb_use}") if $debug_mode_ct
   end
 
   def next_clean_up_step
@@ -1881,7 +1884,7 @@ class GameState
       echo('check_charging?: done') if $debug_mode_ct
 
       cambrinth_charges(nil)
-      invoke(@cambrinth)
+      invoke(@cambrinth, @dedicated_camb_use)
     else
       next_charge = @charges.pop
       echo("check_charging?: #{next_charge}") if $debug_mode_ct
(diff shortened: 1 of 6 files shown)


drboldwin:
So the one drawback I foresee is that this is really only effective for dedicated_camb_use: spell

drboldwin:
also does anything else call invoke or find_charge_invoke_stow other than crossing-training and combat-trainer?

combat-trainer.lic
```diff
@@ -1654,6 +1654,9 @@ class GameState
 
     @analyze_combo_type = settings.analyze_combo_type
     echo("  @analyze_combo_type: #{@analyze_combo_type}") if $debug_mode_ct
+	
+	@dedicated_camb_use = settings.dedicated_camb_use
```
rpherbig: This identation looks off.

combat-trainer.lic
```diff
@@ -1881,7 +1884,7 @@ class GameState
       echo('check_charging?: done') if $debug_mode_ct
 
       cambrinth_charges(nil)
-      invoke(@cambrinth)
+      invoke(@cambrinth, @dedicated_camb_use)
```
rpherbig: I think you also need to update line 823.

rpherbig:
`buff.lic` uses `find_charge_invoke_stow`

`healme.lic` uses `charge_and_invoke`

drboldwin:
kk

combat-trainer.lic
drboldwin: Hmm I think this is more complicated than just adding it to 823.  That is sending your ritual focus into invoke - I'm not sure what would happen if I try to invoke inkwell spell

combat-trainer.lic
drboldwin: fixed - stupid tabs

combat-trainer.lic
rpherbig: Good point. But you'll still have to pass in something or Ruby will complain about the number of arguments.

drboldwin: oh for sure

drboldwin: can I just do, invoke(data['focus'], nil)?

PR #866, opened 2016-09-27
Update theurgy.lic

drboldwin:
Having both tithe's in there really does us no good since there is a 10 minute timer (so just dropping coin for nothing) same likely with crossing-training tithe.  I kept the one closer to the bank/what comes next.

Also Gabellia reported she didn't have enough coin on hand for supplies - I couldn't find where that line went so added it in a spot somewhat arbitrarily for now - this could go in the check_supplies method if we want.

### DIFF
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -474,7 +474,7 @@ class Theurgy
     ######################## SCRIPT ########################
 
     pray
-    ensure_copper_on_hand(600, $settings.hometown)
+    ensure_copper_on_hand(600, @hometown)
     check_supplies
     walk_to '741'
     fput 'commune sense'
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -92,6 +92,7 @@ class Theurgy
   ######################## SHOPPING ########################
 
   def check_supplies
+    ensure_copper_on_hand(600, @hometown)
     if DRStats.circle > 60
       buy_taffelberries unless inside?('Taffelberries', @theurgy_supply_container)
     end
@@ -474,7 +475,6 @@ class Theurgy
     ######################## SCRIPT ########################
 
     pray
-    ensure_copper_on_hand(600, @hometown)
     check_supplies
     walk_to '741'
     fput 'commune sense'
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -267,11 +267,9 @@ class Theurgy
   end
 
   def tithe
-    return unless withdraw_exact_amount?('10 silver', @hometown)
+    return unless withdraw_exact_amount?('5 silver', @hometown)
     walk_to 741
     bput('put 5 silver kronar in almsbox', 'You drop', 'But you do not')
-    walk_to 815
-    bput('put 5 silver kronar in box', 'You drop', 'But you do not')
   end
 
   def pray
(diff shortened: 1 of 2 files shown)


rpherbig:
Why reduce the amount of tithing?

drboldwin:
Hehe you were too fast - didn't finish writing up why yet :-P

theurgy.lic
```diff
@@ -476,6 +474,7 @@ class Theurgy
     ######################## SCRIPT ########################
 
     pray
+    ensure_copper_on_hand(600, $settings.hometown)
```
rpherbig: This should be `@hometown`?

rpherbig:
Interesting. https://elanthipedia.play.net/Almsbox suggests the max is 9 silver, both https://elanthipedia.play.net/Clerical_rituals#Release_from_greed and https://elanthipedia.play.net/Soul_system#Improving_soul_state suggest the max is 5.

Please update `crossing-training` with this PR.

theurgy.lic
drboldwin: touche'

drboldwin:
yeah I had read 5 and rolled with that.  Never saw the other one

rpherbig:
The timer is different between Cleric and Paladin, that's kind of annoying...

drboldwin:
is it?  That is odd.  I suppose we should use whichever has the shorter timer then

theurgy.lic
```diff
@@ -476,6 +474,7 @@ class Theurgy
     ######################## SCRIPT ########################
 
     pray
+    ensure_copper_on_hand(600, @hometown)
```
rpherbig: I suggest moving this into check_supplies

rpherbig:
That's what the two links said (10 minutes vs 4 hours). It seems like a pretty dramatic difference to me, but I haven't tested it.

We can set the cooldown based on the class (DRStats.cleric? and/or DRStats.paladin?) if we wanted to.

theurgy.lic
drboldwin: I figured that would be the case.  Just wanted to get the pull request in first and figured could move it after.

drboldwin:
I don't know that it matters which you tithe to - as I cleric I get the same devotion from the cleric one as I do the paladin one.  So we should just force pally's to fund the cleric guild

rpherbig:
Sorry, I wasn't suggesting we use guild to determine *where* to tithe, but rather the *cooldown*.

drboldwin:
Ahh I get it - I had misunderstood and thought you meant the cleric box and paladin box had different cooldowns - not the guilds themselves had different cooldowns.

drboldwin:
```
    walk_to(shrine['start_room'])
    shrine['path_in'].each { |dir| move(dir) }
  end
```

can we change this to an each do with a fixstanding? check called between each?  Will help to solve the low athletics people falling over.

rpherbig:
We need to get those people moved over to the lich fork :)

drboldwin:
Oh bah I keep forgetting that's the issue here.

drboldwin:
What were we still trying to do here?

rpherbig:
The timer.

drboldwin:
Well you can push this then.  Timer is in crossing Training I think.  My fork is likely behind on that one, so I'll delete it and refork but will have to open up a new Pull Request at that point.

PR #869, opened 2016-09-28
Scroll Search Refactor

ohbeanz:
Removed ScrollInfo class.


scroll-search.lic
```diff
@@ -202,7 +178,15 @@ class ScrollSearch
         # If we cannot match, move on to the next scroll
         next if spell_name =~ /Undefined/i
 
-        new_scroll = ScrollInfo.new($container_full.to_s, @global_scroll_index, index, ordinal.to_s, spell_name.to_s, scroll.to_s)
+        new_scroll =
+        {
+          :container => $container_full.to_s,
```
rpherbig: I suspect none of these `to_s` calls are necessary. Can you check?

Issue #882, opened 2016-10-04
Theurgy - tap my incense in my backpack


Brachius:
**PROBLEM:** 
When theurgy does 'tap my incense in my backpack' AND you have no incense in your backpack, it taps Inoo when he's in the room. I heard Inoo was also asking that we fix this.

**POSSIBLE SOLUTION:** 
Change it to 'look my incense in my backpack'. The look verb also does not have issues with names starting with My who happen to be in the same room.

**MESSAGING:**
**Incense is present:** You see nothing unusual (Tested with Inoo and Mycena in the same room)
**Incense is not present:** I could not find what you were referring to.

rpherbig:
Unfortunately, `look` has a very large number of responses, so we may have to go with only checking the negative case.

Brachius:
Sounds good to me. He gets tapped a lot cause he sits at that safe zone a lot of people use.

PR #886, opened 2016-10-05
Update theurgy.lic

drboldwin:
#882

as well as accounting for cyclic spells before traveling through the temple.

### DIFF
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -290,6 +290,10 @@ class Theurgy
 
   def study_wall
     return if DRStats.circle < 30
+    get_data('spells')
+      .spell_data
+      .select { |_name, properties| properties['cyclic'] }
+    return unless get_data('spells').spell_data.select { |name, _properties| DRSpells.active_spells.keys.include?(name) }.empty?
 
     walk_to '5872'
     move 'go stair'
--- a/common-items.lic
+++ b/common-items.lic
@@ -62,17 +62,17 @@ module DRCI
   end
 
   def wearing?(description)
-    result = DRC.bput("tap my #{description}", 'You tap .*', 'I could not find')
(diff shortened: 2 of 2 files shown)


drboldwin:
https://github.com/rpherbig/dr-scripts/issues/882

as well as accounting for cyclic spells before traveling through the temple.

theurgy.lic
```diff
@@ -290,6 +290,10 @@ class Theurgy
 
   def study_wall
     return if DRStats.circle < 30
+    get_data('spells')
+      .spell_data
+      .select { |_name, properties| properties['cyclic'] }
+    return unless get_data('spells').spell_data.select { |name, _properties| DRSpells.active_spells.keys.include?(name) }.empty?
```
rpherbig: Is this logic used in any other scripts? I'm wondering if there would be use in moving it to common-arcana as a helper.

drboldwin: It's very similar to how common-arcana does release cyclics, but I'm only trying to check if a cyclic is running here, which doesn't have a method built into common-arcana.  Though we could push the method over there and just call it from here.  I wasn't sure how often we would need to only check if a cyclic was running.

rpherbig: If this is the only occurrence, I wouldn't bother making a helper yet.

drboldwin: Either way - it becomes fairly easy to do if we ever feel like it.

drboldwin:
Alright well assuming we are keeping this by itself for now - I think the changes made to common-items will account for tapping someone - but this wont stop it from tapping the person, just register it as a fail.  If you are cool with that we can push this.

rpherbig:
I'll have to ponder a better way to handle 'In' and 'My' people in the room. But in the meantime, you're right - this is better than what we had before.

Issue #884, opened 2016-10-05
Separate cyclic spells from training spells


drboldwin:
Alright so I've been playing with this a bit this morning because I am always looking for the most efficient way to train my magics during crossing-training.

Currently we use training_spells: and I had previously added research_training_spells - what I was thinking is I scrap out research_training_spells and replace it with cyclic_training_spells.  And perhaps a toggle something akin to cyclic_cycle: true.

The way I am using cyclic cycle currently is with it's own timer (set to 300), it is checked BEFORE choosing a skill to train, and if it is off timer it will put up a new cyclic spell depending on lowest EXP.  Then crossing-training picks a skill as per normal, if that skill is aug/util/warding it casts from the training_spells list instead. 

- my quick code (I would likely change variable names and whatnot but is what I tossed together to test with)

**During initialize**
`    @cyclic_cycle_timer = Time.now - 300`
**During Event_Loop (call to refresh cyclics)**
```
    stop_play unless @researching
    refresh_cyclic

    echo "***Attempting to train #{skill}***" if UserVars.crossing_trainer_debug

    update_timer(skill)

    case skill
    when 'Appraisal'
      train_appraisal
    when 'Athletics'
      train_athletics
    when 'Arcana', 'Life Magic', 'Holy Magic', 'Lunar Magic', 'Elemental Magic', 'Arcane Magic'
      do_research skill
```
**It's method**
```
  def refresh_cyclic
    return unless @settings.cyclic_cycle
    if @settings.cyclic_training_spells && (Time.now - @cyclic_cycle_timer > 300)
      skill_pick = @settings.cyclic_cycle_skills
                                       .min_by { |s| DRSkill.getxp(s) }
      cast_spell(@settings.cyclic_training_spells[skill_pick], skill_pick)
      @cyclic_cycle_timer = Time.now
    end
  end
```
^^ Likely need to add a check against researching here also
`    return unless @settings.cyclic_cycle && !@researching`

But we can continue to refresh cyclics between projects by basically scrapping the entire cyclic research section by just throwing a call into def research:

```
  def research
    if @settings.cyclic_cycle
      refresh_cyclic
    end
```
I think we still need that since I feel like I remember the 2nd and 3rd call to research a project is independent of finding the next skill to train (which refresh_cyclic is called just beforehand) but I think for those moments in between projects @researching is false.  I'll have to test this.  But essentially should keep all current functionality and add further cyclic support.

------------------------------------------------------------------
**Overview**

[...]

PR #887, opened 2016-10-05
Cyclic rewrite

drboldwin:
https://github.com/rpherbig/dr-scripts/issues/884

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -200,9 +200,13 @@ class CrossingTraining
     if @settings.cyclic_training_spells && (Time.now - @cyclic_cycle_timer > 300)
       cyclic_skill = @settings.cyclic_cycle_skills
                                        .min_by { |s| DRSkill.getxp(s) }
-      return if DRSkill.getxp(cyclic_skill) >= 32
-      cast_spell(@settings.cyclic_training_spells[cyclic_skill], cyclic_skill)
-      @cyclic_cycle_timer = Time.now
+      if DRSkill.getxp(cyclic_skill) >= 32
+        release_cyclics
+        return
+      else
+        cast_spell(@settings.cyclic_training_spells[cyclic_skill], cyclic_skill)
+        @cyclic_cycle_timer = Time.now
+      end
     end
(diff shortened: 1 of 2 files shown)


crossing-training.lic
```diff
@@ -192,6 +194,16 @@ class CrossingTraining
       pause 15
     end
   end
+  
+  def refresh_cyclic
+    return unless @settings.cyclic_training_spells
+    if @settings.cyclic_training_spells && (Time.now - @cyclic_cycle_timer > 300)
+      cyclic_skill = @settings.cyclic_cycle_skills
```
rpherbig: Does this mean you will always have a cyclic up, even if the applicable skills are mindlocked?

drboldwin: as a note - this won't stop you from always having a cyclic up - that's actually something crossing-training already does if you train with cyclics. There are limited times it calls to release_cyclic and typically it's done just before casting a new cyclic spell. If you stop working on magic because it is all locked, release cyclics wont be called. I can add a feature to stop all cyclics in cyclic cycle when over an XP threshold though if that's what we are interested in.

drboldwin:
What is your thought's on adding a flag to crossing-training that trips when it see's "Your spell cannot maintain itself in this magically inert area"

we can have set the cyclic-cycle timer so that it refreshes a new cyclic at the next possible moment instead of waiting the remainder of the 5 minutes.

crossing-training.lic
drboldwin: I did add this check but as I think about it further I wonder if we wouldn't prefer to have these run while locked - to passively maintain that lock as well as afford Necro's the opportunity to list only rite of grace in their cyclic cycle and therefor keep it up at all times.

drboldwin:
Otherwise I'm not sure what else we would want to do with this before we commit it.  What you thinking @rpherbig ?

rpherbig:
I don't use cyclics much, so I was hoping @rcuhljr could chime in :)

drboldwin:
Damn barbarian

Issue #804, opened 2016-09-15
Create bankbot to get around f2p coin restrictions


rpherbig:
To get around the f2p bank restrictions

Add a 'balance' command too, to check how much the bot owes you.

rpherbig:
@ohbeanz If you want me to assign this to you, you'll have to create a new issue that duplicates this. I will then close this one in favor of your new issue.

rpherbig:
The first pass of bankbot was pushed in 3490c7fab64688ff8d34e619ba31651a3d00f035

It needs to log to a file to be ready for public use. I'm thinking two files, a `.yaml` for balance and a `.log` for transactions.

PR #891, opened 2016-10-07
Adding a Bot Bank to lichbot

ohbeanz:
References Issue: https://github.com/rpherbig/dr-scripts/issues/804
-	Adding a Bot Bank to lichbot. Only handles Kronars right now.

### DIFF
--- a/lichbot.lic
+++ b/lichbot.lic
@@ -128,7 +128,7 @@ class Lichbot
     fput "whisper #{character}   (give|giveme) [item] - take an item from me"
     fput "whisper #{character}   (buffs|spells) - view the buffs I can cast"
     fput "whisper #{character}   (buff|cast) [abbreviation] - receive [abbreviation] buff from me"
-    fput "whisper #{character} I will also accept and stow any item you trade me, or accept coins to hold for you."
+    fput "whisper #{character} I will also accept and stow any item you trade me, or accept coins via TIPs to hold for you."
   end
 
   def balance(character)
@@ -145,6 +145,9 @@ class Lichbot
 
   def deposit(character, amount, denomination)
     # TODO: Handle when multiple withdraw/deposit requests are being made (queue?)
+    # TODO: Handle case when player tries to give coins directly to bot (without using TIP). OR, if they
+    #       do not use the TIP command, check against current copper in-hand to ensure you actually received the
+    #       amount given, to avoid spoofing a 'give'
     @last_action = Time.now
     if denomination =~ /(Lirum|Dokora)/i
       fput "whisper #{character} Unfortunately, I'm only accepting Kronars at the moment."
--- a/lichbot.lic
+++ b/lichbot.lic
@@ -56,8 +56,6 @@ class Lichbot
     @last_teacher = nil
     @inventory = []
     @bot_bank = UserVars.bot_bank || {}
-
-    echo
     refresh_inventory
   end
 
@@ -256,7 +254,6 @@ class Lichbot
   end
 
   def get_balance(character)
-    @last_action = Time.now
     unless has_bot_account(character)
       return 0
     else
--- a/lichbot.lic
+++ b/lichbot.lic
@@ -55,6 +55,9 @@ class Lichbot
     @last_action = Time.now
     @last_teacher = nil
     @inventory = []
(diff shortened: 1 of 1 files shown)


rpherbig:
Why not accept coins other than Kronars?

lichbot.lic
```diff
@@ -55,6 +55,7 @@ class Lichbot
     @last_action = Time.now
     @last_teacher = nil
     @inventory = []
+    @bot_bank = UserVars.bot_bank || {}
```
rpherbig: Why use an instance var over just manipulating the UserVar directly?

rpherbig:
Are you worried that someone might GIVE coins (rather than TIP) and we would lose traceability of that money?

Also, I *suspect* using TIP would be harder to integrate into `sell-loot` (vs GIVE), but probably only by a little.

lichbot.lic
```diff
@@ -122,10 +123,87 @@ class Lichbot
...
     fput "whisper #{character}   (give|giveme) [item] - take an item from me"
     fput "whisper #{character}   (buffs|spells) - view the buffs I can cast"
     fput "whisper #{character}   (buff|cast) [abbreviation] - receive [abbreviation] buff from me"
-    fput "whisper #{character} I will also accept and stow any item you trade me"
+    fput "whisper #{character} I will also accept and stow any item you trade me, or accept coins via TIPs to hold for you."
+  end
+
+  def balance(character)
+    @last_action = Time.now
+    unless has_bot_account(character)
```
rpherbig: Is there a need to manage whether or not they have an account? If you use `Hash.new(0)` you can default everyone to 0.

rcuhljr:
Maybe outside of the scope of this but UserVars is questionable in its integrity at best. I'd suggest writing to a local file with backups via dependency if you're going to track balances. losing everyones balance due to corrupted uservars would be bad.

lichbot.lic
ohbeanz: Sure, makes sense to create new accounts/empty set.  And yes, that's one of the problems with GIVE -- it's not impossible to work around, of course, but not sure how much value I'd get out of going that way. 

TIPs are more transactional, which is helpful when managing balances. Plus, if 3 people give you coins at the same time (before balances/current money on-hand is evaluated) it could quickly screw things up -- or add a lot of extra logic to properly handle these cases.

lichbot.lic
ohbeanz: Fair. Also, I agree with the point made by E... this isn't a good way to handle folk's coins reliably. It should be using a file at the very least.

lichbot.lic
```diff
@@ -122,10 +123,87 @@ class Lichbot
...
+      fput "whisper #{character} The max you can withdraw is: #{balance} Kronar."
+      return
+    end
+
+    case bput("tip #{character} #{amount} kronar", '\w+ declines your tip offer.', 'Your tip offer to \w+ has expired.', '\w+ accepts your tip and slips it away with a smile')
+    when /\w+ declines your tip offer./i
+    when /Your tip offer to \w+ has expired./i
+      fput "whisper #{character} Alright then, I will continue to hold on to it."
+    when /\w+ accepts your tip and slips it away with a smile/i
+      balance = balance - amount.to_i
```
ohbeanz: Completely agree on writing to a flat file for both redundancy and as the actual checkbook for this.

rpherbig:
What do you think about making this a separate script?

The "risk" in being banker is quite different than the equipment storage role. I could see someone wanting the ability to enable one without the other (i.e. I wouldn't want people handing money to Roobee who is in a room in which he can be stolen from).

There is quite a bit of overlap with the rest of lichbot, but... thoughts?

ohbeanz:
Could possibly be a toggle within lichbot since there is so much overlap... instead of creating a whole new script for it. Though, that could get ugly. 

Could pull out "bot/validator" common functions into their own script/class as well and have the inventory bot and bank bot utilize that common class?

rpherbig:
I think I'm still in favor of this being a separate script. The validator logic is really all that is duplicated and could be moved to a common class, as you said.

ohbeanz:
Starting a new PR (https://github.com/rpherbig/dr-scripts/pull/924) that doesn't live off the back of Lichbot.

PR #915, opened 2016-10-11
Astral travel 100th

jandersson:
Added some conditional statements for entering the astral plane if over 99th circle


jandersson:
Ran it in 1230 for testing: http://paste.ubuntu.com/23310049/

PR #916, opened 2016-10-12
Combat Trainer - Rutilor's Edge Support

drboldwin:
Added support for Rutilor's Edge spell.  Similar to ignite - releases when weapons change - lets buff spells take care of it from there with a custom cast that targets active weapon.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -599,6 +599,7 @@ class SpellProcess
     check_consume(game_state)
     check_bless(game_state)
     check_ignite(game_state)
+    check_rue(game_state)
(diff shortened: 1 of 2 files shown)


rpherbig:
Speaking of "commit messages that make good patch notes"... :)

drboldwin:
haha - alright alright

drboldwin:
oh this needs an add into common-spells for it's cast message too - will do that in a moment and bundle into this.

rpherbig:
I'm glad this approach is working, but I think for the next one (maybe Resonance?) we need to find a way to make it more scalable.

PR #918, opened 2016-10-12
Add throne city to base-town

jandersson:
Just adding some rooms. I added a blank npc_healer with the thought that if we ever do a search for 'nearest npc_healer' or 'nearest X' it would show up as nil or the empty string. Would not including it have the same effect?


rpherbig:
Blank 'npc_healer' isn't necessary, it'll be nil either way.

PR #924, opened 2016-10-13
Adding a new Bank bot

ohbeanz:
- Uses common class for validation
- Accepts all currency types
- Still needs to write to a file for redundancy and UserVar failure.


bankbot.lic
```diff
@@ -0,0 +1,229 @@
...
+
+      if account_balance <= 0 || account_balance.nil?
+        next
+      end
+
+      money_readable = minimize_coins(account_balance)
+
+      denomination_msg = ""
+      money_readable.each_with_index do |amount, index|
+        denomination_msg = denomination_msg + "#{money_readable[index]} "
```
rpherbig: Is `money_readable[index]` the same as `amount`? If so, you could use `.each` instead of `.each_with_index`.

bankbot.lic
```diff
@@ -0,0 +1,229 @@
...
+class Bankbot
+  include DRC
+  include DRCM
+
+  def initialize
+    # TODO: Write transactions to a file
+
+    @settings = get_settings
+    @last_action = Time.now
+    @bankbot_debug = UserVars.bankbot_debug
```
rpherbig: You might consider using `UserVars.bankbot_debug` everywhere in the script since this is a long-running script. Checking the UseVar directly (instead of caching it in a local var) would mean you can turn debugging on or off while the script is running, without having to restart it.

bankbot.lic
```diff
@@ -0,0 +1,229 @@
...
+
+        show_balance_sheet(character)
+      end
+    else
+      fput "whisper #{character} I was unable to process your withdrawal. Please try again later."
+      fput "cancel tip"
+    end
+  end
+
+  def ping
```
rpherbig: I don't think you need `ping` or `@last_action` in bankbot.

bankbot.lic
```diff
@@ -0,0 +1,229 @@
...
+    fput "whisper #{character} Supported commands are:"
+    fput "whisper #{character}   (help) - this list"
+    fput "whisper #{character}   (bal|balance) - view your current bot balance"
+    fput "whisper #{character}   (with|withdraw) [amount] [denomination] - withdraw [amount] [denomination] from your bot balance."
+    fput "whisper #{character} I accept coins via TIPs to hold for you."
+  end
+
+  def balance(character)
+    @last_action = Time.now
+    unless has_bot_account(character)
```
rpherbig: Are you going to keep the concept of an account, or default everyone to 0?

bankbot.lic
```diff
@@ -0,0 +1,229 @@
...
+    { name: 'start', regex: /start/i, description: 'Required: prevents accidentally starting up bankbot' },
+    { name: 'announce', regex: /announce/i, optional: true, description: 'If arg is present, bankbot will announce its presence in LNet' },
+    { name: 'greet', regex: /greet/i, optional: true, description: 'If arg is present, bankbot will greet characters after validating (but only once)' }
+  ]
+]
+
+args = parse_args(arg_definitions)
+
+bankbot = Bankbot.new
+validator = CharacterValidator.new(args.announce, args.sleep, args.greet)
```
rpherbig: Since bankbot is storing its data somewhere, unlike lichbot, there's potential to initialize CharacterValidator with a list of known good characters. In this case, I think it would be `UserVars.bot_bank.keys`.

Since that's an optimization, I don't think it needs to be part of this PR. But it might be neat to do once this is merged.

common-validation.lic
```diff
@@ -0,0 +1,39 @@
...
+
+  def confirm(character)
+    return if valid?(character)
+
+    echo "Successfully validated: #{character}"
+    @validated_characters << character
+
+    return unless @greet
+
+    put "whisper #{character} Hi! I'm your friendly neighborhood lichbot. Whisper me 'help' for more details. Don't worry, I've memorized your name so you won't see this message again."
```
rpherbig: We might consider making 'lichbot' here configurable so lichbot and bankbot have different strings.

Or maybe (and this could be out of scope of this PR) bankbot can output their balance here, at greeting time.

bankbot.lic
```diff
@@ -0,0 +1,229 @@
...
+        denomination_msg = denomination_msg + "#{money_readable[index]} " 
+      end
+
+      fput "whisper #{character} It looks like your current balance is #{denomination_msg} (#{account_balance}) #{type}."
+    end
+  end
+
+end
+
+arg_definitions = [
```
rpherbig: There's no arg for `sleep`, yet it is referenced on line 180. I recommend changing line 180 to `false` to make it clear.

Although, to be honest, I've been considering removing that functionality from lichbot since it never gets used...

bankbot.lic
```diff
@@ -0,0 +1,229 @@
+# want_script_output
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#bankbot
+=end
+
+custom_require.call(%w(common common-money common-validation drinfomon events))
```
rpherbig: I *think* you do not need drinfomon or events here.

bankbot.lic
ohbeanz: Makes sense.

bankbot.lic
ohbeanz: Agreed. Removing them.

bankbot.lic
ohbeanz: Yes, I don't see the harm. At least a harm that would really warrant the rewrite/refactor at this junction.

bankbot.lic
ohbeanz: Agreed. This is a good optimization to cut down on re-valididating. However, like you mentioned, I will probably hold off on it.

common-validation.lic
ohbeanz: Yup, agreed. Though, also a bit OOS.

bankbot.lic
ohbeanz: Agreed. Bank bot does not need or use either. Will remove.

bankbot.lic
```diff
@@ -106,25 +105,26 @@ class Bankbot
...
       return UserVars.bot_bank[character][denomination].to_i
     end
   end
 
   def has_bot_account(character)
     unless UserVars.bot_bank.has_key?(character)
-      return false
+      opened_account = (UserVars.bot_bank[character].has_key?("Kronar") || UserVars.bot_bank[character].has_key?("Lirum") || UserVars.bot_bank[character].has_key?("Dokora"))
+      
+      unless opened_account
```
rpherbig: Could this block be `return opened_account`?

bankbot.lic
```diff
@@ -0,0 +1,232 @@
...
+    bankbot.withdraw(character, amount, denomination) if validator.valid?(character)
+  when /^(.*) whispers, "(bal|balance)"$/i
+    character = Regexp.last_match(1)
+
+    bankbot.balance(character) if validator.valid?(character)
+  when /^(.*) whispers, ".*"$/i
+    character = Regexp.last_match(1)
+
+    fput "whisper #{character} I'm sorry, I did not understand that command"
+    bankbot.help(character)
```
rpherbig: This call (and the whisper above) should check if the character is validated. I think that's a bug in lichbot, too...

bankbot.lic
```diff
@@ -0,0 +1,232 @@
...
+      echo ("Trying to return: #{UserVars.bot_bank[character][denomination]}") if UserVars.bankbot_debug
+      return UserVars.bot_bank[character][denomination].to_i
+    end
+  end
+
+  def has_bot_account(character)
+    unless UserVars.bot_bank.has_key?(character)
+      opened_account = (UserVars.bot_bank[character].has_key?("Kronar") || UserVars.bot_bank[character].has_key?("Lirum") || UserVars.bot_bank[character].has_key?("Dokora"))
+      
+      unless opened_account
```
rpherbig: Is this the same as `return opened_account`?

bankbot.lic
```diff
@@ -0,0 +1,232 @@
...
+      
+      fput "whisper #{character} #{msg} #{UserVars.bot_bank[character]} #{currency_type}!" if UserVars.bankbot_debug
+
+      show_balance_sheet(character)
+    end
+  end
+
+  def withdraw(character, amount, denomination)
+    # TODO: Handle when multiple withdraw/deposit requests are being made (queue?)
+    @last_action = Time.now
```
rpherbig: `@last_action` is in quite a few methods and can be removed

bankbot.lic
rpherbig: I'll fix lichbot, just keep an eye out for a merge conflict.

rpherbig: I very slightly changed the style of these validations. Lichbot now looks like:
```
 when /^(.*) whispers, ".*"$/i
    character = Regexp.last_match(1)

    return unless validator.valid?(character)
    fput "whisper #{character} I'm sorry, I did not understand that command"
    bankbot.help(character)
```

PR #931, opened 2016-10-14
Added expanded combos with buffs for Barbs

jayRyan24:
Added the ability for Barbs to maintain the short-term buffs of Damage and Accuracy and then fill in with Flame to maintain high IF.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1314,6 +1314,11 @@ class AttackProcess
     pause
     waitrt?
 
+    if reget(20, 'increase the accuracy')
+      game_state.set_barb_accuracy_flag
+    elsif reget(20, 'increase the damage')
+      game_state.set_barb_damage_flag
+    end
     if reget(5, 'You can\'t backstab that')
       if game_state.npcs.uniq.length == 1
         game_state.unstabbable(game_state.npcs.first)
@@ -1707,6 +1712,20 @@ class GameState
     @analyze_combo_type = settings.analyze_combo_type
     echo("  @analyze_combo_type: #{@analyze_combo_type}") if $debug_mode_ct
 
+    @use_barb_combos = settings.use_barb_combos
+    echo("  @use_barb_combos: #{@use_barb_combos}") if $debug_mode_ct
+
+    # For deprecation support
+    if !@analyze_combo_type.nil?
+      @use_barb_combos = true
+    end
+
+    if @use_barb_combos
+      Flags.add('ct-accuracy-ready', 'You sense your combat accuracy decrease')
+      Flags['ct-accuracy-ready'] = true
+      Flags.add('ct-damage-ready', 'You sense your ability to land deadly blows decrease')
+      Flags['ct-damage-ready'] = true
+    end
     @dedicated_camb_use = settings.dedicated_camb_use
     echo(" @dedicated_camb_use: #{@dedicated_camb_use}") if $debug_mode_ct
   end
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -1984,31 +2011,53 @@ class GameState
...
     end
   end
 
   private
 
-  def update_analyze_array
-    result = bput("analyze #{@analyze_combo_type}", 'Analyze what\?', 'by landing a .*')
-    waitrt?
+  def perform_analyze?(combo_type, expertise_requirement)
+    return false if combo_type == "accuracy" && !Flags["ct-#{combo_type}-ready"]
```
rpherbig: You could combine these two cases by checking `combo_type != 'flame' && ...`

combat-trainer.lic
```diff
@@ -1314,6 +1314,11 @@ class AttackProcess
     pause
     waitrt?
 
+    if reget(20, 'increase the accuracy')
```
rpherbig: I generally try to avoid using reget because it's not reliable. Can we instead reset flags if we've finished the combo?

combat-trainer.lic
```diff
@@ -1707,6 +1712,20 @@ class GameState
     @analyze_combo_type = settings.analyze_combo_type
     echo("  @analyze_combo_type: #{@analyze_combo_type}") if $debug_mode_ct
 
+    @use_barb_combos = settings.use_barb_combos
+    echo("  @use_barb_combos: #{@use_barb_combos}") if $debug_mode_ct
+
+    # For deprecation support
+    if !@analyze_combo_type.nil?
```
rpherbig: You could simplify this to `if @analyze_combo_type`

combat-trainer.lic
```diff
@@ -1707,6 +1712,20 @@ class GameState
...
+      @use_barb_combos = true
```
rpherbig: I recommend doing this before echoing the value on line 1716 (or better yet, echo something else here so we know the setting got overriden).

combat-trainer.lic
jayRyan24: This is tricky. Just finishing the array doesn't mean that it was successful, sometimes it inputs actions that get triggers are completed, but don't really. This happens when something dies and you gotta switch targets, it leaves, it's out of range, etc. I thought about a flag that checked if doing the combo, but you still need to check if that text was received to know you have done it. A way to make this work without it, is just to remove the reget and let it attempt to go through, if it finds that it's been done and the combo is on cooldown, it sets the flag to false. This is actually part of the backup checks to make sure it doesn't continuously loop. Removing reget will put this as the main check. The only issue is that it might appear spammy the first time it goes through to find out if it's on cooldown.

combat-trainer.lic
```diff
@@ -1912,6 +1931,14 @@ class GameState
     @aim_queue.shift
   end
 
+ def next_combo_action
```
jayRyan24: Do you mean both the aim_queue.shift and the Analyze_combo_array? def next_combo_action would be a new function that I included to follow convention, I can just remove it from the commit.

rpherbig: I was referring to `next_combo_action` and `use_barb_combos?`. They are only used internally to GameState.

The other methods you were looking at (with aim_queue.shift) are used externally.

jayRyan24: That's fine, I'll just remove it from the commit and PR.

jayRyan24:
I had a lot of issues with my main PC and switched it over to a Linux box, etc. I deleted the old fork and made a new one. I never stopped using my old files though, and have been running them without issue for nearly a month. Please use the Pull request.

PR #958, opened 2016-10-21
Direct Message Status-Monitor Output to Slack User

jandersson:


profiles/base.yaml
```diff
@@ -436,4 +436,6 @@ quit_on_status_warning: false
 
 compost_room: 991
 
-carving_workorder_material_type: stone
\ No newline at end of file
+carving_workorder_material_type: stone
+
+slack_username:
```
jandersson: As an example, for my characters it is
`slack_username: jonas`

jandersson:
If you all feel this is too niche, let me know! I won't be offended if the consensus is that this should be a private application. I felt that this was a simple solution to do push notifications. There is potential to respond to the slackbot to have it perform a specified action, e.g. respond with 'remove armband' and 'lick armband'

jandersson:
Do you think it would be better to add the slack_username as an input somewhere in the GUI?

status-monitor.lic
```diff
@@ -154,6 +154,31 @@ class GameFilter
   end
 end
 
+require 'slack-ruby-client'
```
rcuhljr: Could you move the slackbot into a separate file? we could then just custom_require it if the slack_username setting exists that way people wanted to use status monitor don't need the gem. So all we'd have here is
```

if get_settings['slack_username']
  custom_require.call('somefile')
  bot = SlackBot.new
end
```

jandersson:
Moved it to a separate file, wondering if there should be a check for script trustedness to avoid this kind of messaging:
```
>;status
>
--- Lich: status-monitor active.
--- Lich: slackbot active.
--- Lich: error: Insecure operation - require
	/home/jonas/.rvm/rubies/ruby-2.0.0-p648/lib/ruby/site_ruby/2.0.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	/home/jonas/.rvm/rubies/ruby-2.0.0-p648/lib/ruby/site_ruby/2.0.0/rubygems/core_ext/kernel_require.rb:54:in `require'
--- Lich: review this script (slackbot) to make sure it isn't malicious, and type ;trust slackbot
--- Lich: slackbot has exited.
>
```

rcuhljr:
Yes it's going to require trust just like status monitor does.

PR #985, opened 2016-10-28
Commonize Performance

jandersson:
I tried to make this a singleton class by making all variables static... still testing it right now, putting it up here for any feedback on the idea


jandersson:
This is working:

```

[crossing-training]>collect rock
You move slightly to the left, hoping to find a better foraging spot.
Roundtime: 15 sec.
R>
You feel fully rested.
R>
[crossing-training]>kick rocks
You take a step back and run up to the pile of rocks.  Bringing your foot smashing down onto the pile, you send bits and pieces of debris scattering in every direction.  Looking around for the now defunct mountain of glory, you fail to find any signs of its presence.
>
[crossing-training]>play elegy
You begin a quiet elegy on your thin-edged zills with only the slightest hint of difficulty.
>
You continue playing on your thin-edged zills.

>
You find yourself relaxing as you continue to play a quiet elegy on your thin-edged zills.

>
You continue playing on your thin-edged zills.

>
Psionic Psilocybe just arrived.
>
Psionic Psilocybe goes east.
>
You continue to play your unoriginal elegy on your thin-edged zills.

>
You feel you have sufficiently pondered your latest observation.
>
The world seems a bit more foreboding than it did a moment ago.
You recognize these fading strands of Fate as those that influence your small blunt skill.
>
You continue playing on your thin-edged zills.
```

rcuhljr:
Since this isn't a module can we not use the common-* naming. It should probably just be performance.lic

rcuhljr:
`class_defs` in bootstrap.lic should have Performance added to it since it's a singleton like equipmanager.

jandersson:
What does adding it to bootstrap do?

rcuhljr:
It allows it to detect if the script is already running so it doesn't run it again when listed as a requirement.

jandersson:
OK, that sounds like it solves the multiple instance problem. Implemented the requested changes

crossing-training.lic
```diff
@@ -760,15 +758,15 @@ class CrossingTraining
...
+    Performance.stop_play if @settings.hide_to_steal
 
     wait_for_script_to_complete('steal')
   end
 
   def train_appraisal
     return if @researching
 
-    play_song? if DRSkill.getrank('Appraisal') >= 250
+    Performance.play_song? if DRSkill.getrank('Appraisal') >= 250
```
rpherbig: This line should be moved into `appraisal.lic` now.

performance.lic
```diff
@@ -0,0 +1,100 @@
+# quiet
```
rpherbig: Can you add a header comment block to this (the link to epedia).

performance.lic
```diff
@@ -0,0 +1,100 @@
...
+
+  def self.instance
+    return @@instance
+  end
+
+  def self.stop_play
+    return unless @@did_play
+    return if @@no_instrument
+    @@did_play = false
+    DRC.bput('stop play', 'You stop playing your song', 'In the name of', "But you're not performing")
```
rpherbig: I don't think `DRC` is necessary since you included DRC above. Can you check on that?

crossing-training.lic
```diff
@@ -760,15 +758,15 @@ class CrossingTraining
   end
 
   def train_thievery
-    stop_play if @settings.hide_to_steal
+    Performance.stop_play if @settings.hide_to_steal
```
rpherbig: This line should be moved to `steal.lic`.

performance.lic
```diff
@@ -0,0 +1,100 @@
...
+    Flags['ct-song'] = true
+  end
+
+  def self.play_song?(blocking = false)
+    return true if @researching
```
rpherbig: If `Performance.lic` isn't tracking researching, should this line be removed?

If it is removed, does it need to be put in `crossing-training` in each place `play_song?` is called?

jandersson: I was just looking at that.. yea that seems to be the consequence of not tracking researching. What would you suggest? I think tracking it would be the easiest. Trying to figure out how to set this up so that when `@researching` is updated in CT this also updates `Performance.is_researching`. If `is_researching` could be an alias for `@researching` that might be the easiest

jandersson:
I am having some doubts about whether or not this is a good implementation of a script that is used in other scripts. It seems for this to be successful, researching needs to be its own singleton script as well in order to share its state with Performance. I have made a change which has crossing-trainer set the researching state of both Performance and crossing-trainer anytime it wants to update `@researching` but this feels unsafe and kludge-like.

There might be a better solution in letting dependency hold some sort of game_state object that all scripts can access and update (if possible). If thats possible, then there is no need to implement this as a singleton, and can just be a module instead. Also, I welcome the education if im off-base in my assumptions!

If anyone feels like this is the way to move forward let me know, otherwise i think i will close this PR in a few days.

performance.lic
```diff
@@ -4,6 +4,7 @@ class Performance
   include DRC
 
   def initialize
+    @@is_researching ||= false
```
jandersson: This will be deleted, it is not used.

Issue #989, opened 2016-10-29
BESCORT edit for Astral Travel


Aurayn:
**any way to NOT drop harnessed mana if the moongate spell fails and to refocus and recast on the shard?**

```
[bescort]>prepare mg 5
Your eyes darken to black as a starless night as you prepare the Moongate spell.
>
[bescort]>focus Taniendar
You focus your magical senses on the silvery-white shard Taniendar.
The shard is generating a tiny amount of Lunar mana, linking it solidly to the microcosm.
You sense that the vast majority of the shard's energies are tied into streams of mana which connect it to some great source in the distance.
Reaching past the shard, you sense that the streams of Grazhir's mana flow steadily and are suffering from minor disruptions.
A distinct and vibrant pattern of Stellar Magic is set within the shard, the result of a ritual naming.
You recognize this as the shard Taniendar and forge a connection to it through the magical pattern.
Roundtime: 20 sec.
>
You feel fully attuned to the mana streams again.
>
You effortlessly maintain your place among the shifting streams of mana.
>
You sense a tiny amount of mana slip away from you.
>
You sense a tiny amount of mana slip away from you.
>
You feel fully prepared to cast your spell.
>
[bescort]>cast Taniendar
You gesture.
Your efforts are aided by your enhanced understanding of the Plane of Probability.
You attempt to open an astral conduit through your connection to the shard Taniendar.
The spell goes awry!
With supreme effort, you are able to bring the spell to an end without harm.
>
[bescort]>rel mana
You release all the streams you were concentrating on keeping localized around you.
>
--- Lich: bescort has exited.
```

rpherbig:
@jandersson Ping

Aurayn:
Pong

jandersson:
Yea definitely. Since "You gesture" is already matched, I think a flag for "With supreme effort, you are able to bring the spell to an end without harm." could be checked to see if we should retry. Any suggestions for how to handle catastrophic failure (voidspawn)? Quit?

Aurayn:
yes Quit.... I have only ever avoided 9(without help) Voidspawns twice and that was after a Quit

jandersson:
Sounds good to me, I can add add a flag for that once i know what the messaging looks like for that kind of failure

jandersson:
Nevermind the messaging is on epedia. 

It looks like there are two types of messaging, one with voidspawn one without.
```
Rogue Moongate
You gesture.
[...]

PR #990, opened 2016-10-30
Astral travel attempts to recast gate on non-catastrophic failure

jandersson:
Addresses #989 

Created a method for focusing on shard and casting gate, `cast_gate`. Added a flag to check for the failure. If failure is detected then call `cast_gate` again

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -575,9 +575,27 @@ class Bescort
     result != 'Roundtime'
   end
 
+  def open_gate(shard, entering_ap)
+    Flags.reset('gate-failure')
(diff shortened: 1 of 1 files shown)


jandersson:
Failure doesn't happen too often, so it may be a while before it can be verified

jandersson:
There also exists failure where you are stunned, summon voidspawn, and usually get incinerated. Not sure how to handle that yet

jandersson:
This is working as it did before. It'll be a while before Crannach fails (I do a walk about twice a week) and gets to verify that it works. Im done making changes though

PR #999, opened 2016-11-04
Slackbot Without Slack Gem

jandersson:
I removed the slack gem from this class and instead used http and json. I also obfuscated the token, Slack seems to crawl github and invalidates any tokens it finds. So this also includes the new token.

### DIFF
--- a/slackbot.lic
+++ b/slackbot.lic
@@ -7,9 +7,9 @@ class SlackBot
     some_string = 'ypyc-94136242850-c8t6YDXkOkW6evAO8yQITS6Q'
     alphabet = ('a' .. 'z').to_a.join + ('A' .. 'Z').to_a.join
     cipher = alphabet.chars.rotate(1).join
-    @token = some_string.tr(ciper, alphabet)
-    @dm_channel = get_dm_channel(get_settings.slack_username)
+    @token = some_string.tr(cipher, alphabet)
     @api_url = "https://slack.com/api/"
+    @dm_channel = get_dm_channel(get_settings.slack_username)
   end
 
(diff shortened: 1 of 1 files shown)


rpherbig:
What do you think about putting the token in a UserVar (or whatever the account-wide version of that is)?

jandersson:
Do you mean to store it there instead of anywhere in the script itself? So remove lines 7-9?

rpherbig:
Correct. The obfuscation is there to work around something that Slack has good reason to check for. Anyone could grab that token and impersonate you in the Slack team with it, right?

jandersson:
Yep, they could impersonate the bot I think is the implication

jandersson:
Im wondering how to communicate the token to the user so that it can be stored in a uservar

rcuhljr:
Except anyone you give the token to in game can also impersonate the bot. There's no way around this other then forcing people to make their own bots and store their own key in a uservar.

rpherbig:
I'm not worried about Lichers we give the token to, I'm worried about people crawling GitHub for tokens (which does happen).

rcuhljr:
Minimal obfuscation can stop that.

rpherbig:
Encryption would, but I don't think this kind of obfuscation would stop anyone that was actually looking for tokens.

jandersson:
I understand the risk that someone impersonates the bot could be disruptive, though I don't think any real harm can come from it other than spam (all it can do is act like another user). I can try to think of a way to pass the token in a secure fashion in the meantime. My thought is that a lichbot can be queried for the token if it isn't stored.

PR #1000, opened 2016-11-04
Added support for worn foci

adamgoerdt:


rpherbig:
What do you think about making `find_focus` and `stow_focus` helpers that are similar to their cambrinth counterparts but specific to rituals?

adamgoerdt:
That might makes sense from a function name clarity perspective,  but I
feel like the code will just be copy/paste. I think it makes more sense to
make the function name generic,  but am not sure where get/stow_cambrinth
might be called from to refactor it.

On Nov 4, 2016 5:23 PM, "Robert Herbig" <notifications@github.com> wrote:

What do you think about making find_focus and stow_focus helpers that are
similar to their cambrinth counterparts but specific to rituals?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rpherbig/dr-scripts/pull/1000#issuecomment-258551145>,
or mute the thread

rpherbig:
The ritual code hard codes some arguments to the cambrinth helpers

adamgoerdt:
Looking at the code I think it makes sense to create the new functions, but to still use the focus: and work_focus: settings from the individual spell, since different spells might require different foci, which is different than cambrinth, which can be global.

rpherbig:
Agreed

rpherbig:
It looks like you're using tabs. That makes it hard to review these changes since GitHub makes it look like the entire file changed.

Can you switch to spaces and push updated files?

PR #1001, opened 2016-11-05
Exchange slack token

jandersson:
Lichbots can store the tokens in their uservars. When a person runs status-mon and doesn't have a token in their uservars (or has a bad token), it will query the listed lichbots for the token. If the token auths, we set the token. We send a success message on first time setup


jandersson:
I'll generate a new token to distribute if this gets merged

rpherbig:
I'm having trouble getting the http gem installed. Did you do anything beyond `gem install http`?

jandersson:
Don't think so, is net/http the same as http?

jandersson:
No it isn't my mistake. Ill write a post method with net/http

rpherbig:
Let me know what the new token is and I'll load it into Roobeelitch.

PR #1005, opened 2016-11-07
Added support for worn foci and different textbook nouns

adamgoerdt:


rpherbig:

Issue #945, opened 2016-10-18
Failed rummage for boxes while invisible should release invisibility


jandersson:
We could detect and release invisibility the same way we do with cyclics

```
You feel fully prepared to cast your spell.
I>
[crossing-training]>cast
You reveal yourself.
Your cambrinth armband emits a loud *snap* as it discharges all its power to aid your spell.
You whisper the final word of your spell so that none may notice your effort.
Your spell injects more psychic corruption into the area.
I>
--- Lich: pick active.
[pick]>open my haversack
That is already open.
I>
[pick]>open my haversack
That is already open.
I>
[pick]>rummage /B my haversack
You feel about a utilitarian brown oilcloth haversack.
I>
You feel fully rested.
I>
[pick]>rummage /B my 
I don't know what you are referring to.
I>
--- Lich: pick has exited.
[crossing-training: ***UNABLE TO TRAIN LOCKSMITHING, REMOVING IT FROM THE TRAINING LIST***]
[crossing-training]>collect rock
You manage to collect a pile of rocks.
Roundtime: 15 sec.
IR>
```

PR #1017, opened 2016-11-12
Release invisibility spells

jandersson:
This adds support for releasing invisiblity spells, addressing the issue of attempting to use the bank while invisible, rummaging in containers, and dismantling boxes (which breaks pick.lic)

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -149,6 +149,15 @@ module DRCA
     result == 'absorbs all of the energy'
   end
 
+  def release_invisibility
+    get_data('spells')
+      .spell_data
+      .select { |name, properties| properties['invisibility'] }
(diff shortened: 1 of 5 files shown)


common-arcana.lic
```diff
@@ -149,6 +149,15 @@ module DRCA
     result == 'absorbs all of the energy'
   end
 
+  def release_invisibility
```
jandersson: This could potentially be combined with `release_cyclics` into `release_spells(property)` or `release_spells(*property)`

rpherbig:
I think this helper should be called in `hunting-buddy`, line 60.

rpherbig:
Note to self: this should fix #945

jandersson:
Closing and reopening for profiles -> data  improvement

PR #1027, opened 2016-11-14
Add setup files for Crannach and Jonas

jandersson:


jandersson:
The setup file Jonas-setup.yaml could replace ANecromancer-setup.yaml since they were the same at some point

rpherbig:
@jandersson Would you be up for contributing `Jonas-back.yaml`?

jandersson:
Of course

Issue #801, opened 2016-09-15
Implement "Release Eyes of the Blind" for Necros after hunting buddy


wscherer420:
When using Eyes of the Blind (EOTB) during combat-trainer it can linger into the crossings training.  EOTB makes the Necro invisible.  This prevents a Necro from being able to sell-loot or execute the crossings-repair scripts.  The spell can be released by entering "release EOTB".  It reveals you to the world.  This should be done as hunting buddy exits.  

Thoughts?

PR #1030, opened 2016-11-16
Support releasing invisibility spells

jandersson:
This releases invisibility spells in a very similar fashion to the cyclic release. Addresses #945 
The idea is to use the common method `release_invisibility` before doing an action which would fail on the grounds of being invisible

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-money
 =end
 
-custom_require.call(%w(common common-travel common-arcana))
+custom_require.call(%w(common common-travel))
 
 module DRCM
   module_function
--- a/common-money.lic
+++ b/common-money.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-money
 =end
 
-custom_require.call(%w(common common-travel))
+custom_require.call(%w(common common-travel common-arcana))
 
 module DRCM
   module_function
@@ -48,6 +48,7 @@ module DRCM
   def withdraw_exact_amount?(amount_as_string, hometown)
     bank_id = get_data('town')[hometown]['deposit']['id']
     DRCT.walk_to(bank_id)
+    DRCA.release_invisibility
     'The clerk counts' == DRC.bput("withdraw #{amount_as_string}", 'The clerk counts', 'The clerk tells')
   end
 
--- a/common-money.lic
+++ b/common-money.lic
@@ -48,7 +48,7 @@ module DRCM
   def withdraw_exact_amount?(amount_as_string, hometown)
     bank_id = get_data('town')[hometown]['deposit']['id']
     DRCT.walk_to(bank_id)
-    DRCA.release_invisibility
+    DRC.release_invisibility
     'The clerk counts' == DRC.bput("withdraw #{amount_as_string}", 'The clerk counts', 'The clerk tells')
   end
 
--- a/common.lic
+++ b/common.lic
@@ -260,9 +260,9 @@ module DRC
   def release_invisibility
     get_data('spells')
       .spell_data
-      .select { |name, properties| properties['invisibility'] }
-      .select { |name, properties| DRSpells.active_spells.keys.include?(name) }
-      .map { |name, properties| properties['abbrev'] }
+      .select { |_name, properties| properties['invisibility'] }
+      .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
+      .map { |_name, properties| properties['abbrev'] }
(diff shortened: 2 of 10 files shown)


common.lic
```diff
@@ -102,7 +104,10 @@ module DRC
     result = DRC.bput("rummage /#{parameter} my #{container}", 'but there is nothing in there like that\.', 'looking for .* and see .*', 'While it\'s closed', 'I don\'t know what you are referring to', 'You feel about')
 
     case result
-    when 'but there is nothing in there like that.', 'While it\'s closed', 'I don\'t know what you are referring to', 'You feel about'
+    when 'You feel about'
+      release_invisibility
```
rpherbig: @rcuhljr Will something weird happen with common and common-arcana both `custom_require`ing each other?

jandersson:
I'm looking at line 60 in hunting-buddy that was suggested to be changed. What does this do? It looks like it releases invisibility after hunting

rpherbig:
That line was added to fix #801. There seem to be two general approaches:
1. Call `release_invisibility` in each place that it could cause problems (I think you'd need to test `sell-loot` and `crossing-repair`)
2. Call `release_invisibility` in hunting-buddy or training-manager when we transition "states"

I'm in favor of the first option, because it is safer to call those scripts manually.

jandersson:
I think the first option makes the constraints clearer (ie. have to be visible for this action to succeed) but the second one is easier. I like the first one because it lets me run around invisible for longer :]

jandersson:
Ive added the release to banking. I have not tried it with repairing, ill see how that goes

common.lic
jandersson: Is the behaviour undefined?

rcuhljr: It's very defined, on a fresh boot it's going to deadlock hard. Common can't load waiting on common-arcana to load, common-arcana can't load waiting on common to load.

jandersson:
I could do a lamda with the release code

Sent with AquaMail for Android
http://www.aqua-mail.com

common.lic
jandersson: OK, I'm thinking of copying and pasting in the function body for release_invisibility in the `when` body

common.lic
```diff
@@ -105,7 +103,13 @@ module DRC
 
     case result
     when 'You feel about'
-      release_invisibility
+      # Reduntant copy of DRCA.release_invisibility to avoid deadlocking from including commona-arcana
```
jandersson: Don't mean redundant in a negative sense btw, just saying DRY is too rigid in this case :]

jandersson:
You cannot give items to the repair guy when invisible so added the release for that. I took out release EOTB like from combat-trainer because this should still cover #945

common.lic
```diff
@@ -105,7 +103,13 @@ module DRC
...
+      get_data('spells')
+        .spell_data
+        .select { |name, properties| properties['invisibility'] }
+        .select { |name, properties| DRSpells.active_spells.keys.include?(name) }
```
rcuhljr: Yes.

rpherbig:
I wonder if it is time to find a better home for the `rummage` and `get_*` (boxes, skins, etc.) methods. They aren't used in very many places... I'm always in favor of moving code out of `common` and into something more specific.

jandersson:
equipmanager? it has state, but rummage and get_* deal with equipment was my thought

rcuhljr:
I probably still would have preferred to move release_invisibility into common. No objections to this PR though.

jandersson:
I can still do that, it makes sense in removing the dupe code.

Sent with AquaMail for Android
http://www.aqua-mail.com

common-money.lic
```diff
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-money
 =end
 
-custom_require.call(%w(common common-travel))
+custom_require.call(%w(common common-travel common-arcana))
```
rpherbig: This no longer needs `common-arcana`

crossing-repair.lic
```diff
@@ -2,10 +2,11 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#crossing-repair
 =end
 
-custom_require.call(%w(common common-money common-travel drinfomon equipmanager))
+custom_require.call(%w(common common-arcana common-money common-travel drinfomon equipmanager))
```
rpherbig: This no longer needs `common-arcana`

crossing-repair.lic
```diff
@@ -2,10 +2,11 @@
...
 
 class CrossingRepair
   include DRC
+  include DRCA
```
rpherbig: This line can be removed

PR #1031, opened 2016-11-16
Removed non-workorder lines from outfitting

adamgoerdt:


rpherbig:
I seem to recall some people want to train Outfitting without using workorders. Have you polled the Lich community to make sure no one is using that?

rcuhljr:
We've discussed this in game.

PR #1052, opened 2016-11-20
Add support for Barbarian Meditate Research.

jayRyan24:
Adding support for Barbarian Meditate Research.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1057,12 +1057,6 @@ class TrainerProcess
 
     @flint_lighter = settings.flint_lighter
     echo("  @flint_lighter: #{@flint_lighter}") if $debug_mode_ct
-
-    @barb_research_skills = settings.barb_research_skills
-    echo("  @barb_research_skills: #{@barb_research_skills}") if $debug_mode_ct
-
-    @current_barb_research = 'Augmentation'
-    echo("  @current_barb_research: #{@current_barb_research}") if $debug_mode_ct
   end
 
   def execute(game_state)
@@ -1107,11 +1101,11 @@ class TrainerProcess
     when 'Recall'
       bput("recall #{game_state.npcs.first}", 'Roundtime', 'You are far too occupied') unless game_state.npcs.empty?   
     when 'Research Augmentation'
-      bput("meditate research #{@barb_research_skills['Augmentation']}", 'You clear ')
+      bput('meditate research monkey', 'You clear ') if DRStats.barbarian?
     when 'Research Warding'
-      bput("meditate research #{@barb_research_skills['Warding']}", 'You clear ')
+      bput('meditate research turtle', 'You clear ') if DRStats.barbarian?
     when 'Research Debilitation'
-      bput("meditate research #{@barb_research_skills['Debilitation']}", 'You clear ')    
+      bput('meditate research earthquake', 'You clear ') if DRStats.barbarian?  
     end
     waitrt?
     false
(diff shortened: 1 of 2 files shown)


jayRyan24:
Example YAML settings:

```
training_abilities:
  Hunt: 80
  App Quick: 60
  Tactics: 20
  Research Warding: 120
  Research Augmentation: 120
  Research Debilitation: 120
barb_research_skills:
  Warding: turtle
  Debilitation: earthquake
  Augmentation: monkey
use_barb_combos: true
dual_load: true
```

rpherbig:
Does which ability you meditate on matter? That is, does it need to be
configurable or could we always use turtle, earth quake, and monkey?

jayRyan24:
It needs to be configurable because not everyone has all or some of those. People will need to put the ability they want in there.

rpherbig:
You can meditate research on an ability you do not have. What happens
differently if you do or don't have the research ability?

jayRyan24:
Oh well, that was a waste of time. :( Well, learned a lot, so not totally. 

As far as what are the issues with it, none. You can do it from circle 1. I just made a test Barb and did it right after joining. 

`Agonar greets you, "Welcome, young Barbarian.  Let me tell you of our Guild!"
   
   
Name: Azwal   Race: Dwarf   Guild: Barbarian
Gender: Male   Age: 35   Circle: 0
You were born on the 31st day of the 3rd month of Lirisa the Archer in the year of the Bronze Wyvern, 386 years after the victory of Lanival the Redeemer.
   
Your birthday is more than 4 months away.
   
     Strength :  10              Reflex :   8  
      Agility :   8            Charisma :  10  
   Discipline :  12              Wisdom :  10  
 Intelligence :  10             Stamina :  12  
   
Concentration : 331    Max : 331
   
       Favors : 0
         TDPs : 600
  Encumbrance : None
   
Wealth:
  1 gold and 2 silver Kronars (1200 copper Kronars).
  No Lirums.
  No Dokoras.
   
Debt:
  No debt.
   

meditate research turtle
You clear your mind and begin to meditate upon the training you have received.
Roundtime: 9 sec.
You recall that Turtle Form is an Expert ability in the Path of the Flame.  Practicing these movement styles will enshroud the Barbarian with a strong potency barrier if they possess sufficient warding skill.`

rpherbig:
Do you need to retreat before using meditate research? Or do you keep standing?

jayRyan24:
It doesn't sit or anything, it just goes. I've been using it all day long.

combat-trainer.lic
```diff
@@ -1097,7 +1099,13 @@ class TrainerProcess
     when 'Meraud'
       meraud_commune(game_state)
     when 'Recall'
-      bput("recall #{game_state.npcs.first}", 'Roundtime', 'You are far too occupied') unless game_state.npcs.empty?
+      bput("recall #{game_state.npcs.first}", 'Roundtime', 'You are far too occupied') unless game_state.npcs.empty?   
+    when 'Research Augmentation'
+      bput('meditate research monkey', 'You clear ') if DRStats.barbarian?
```
rpherbig: I don't think we need the `if barbarian` check - non barbarians shouldn't attempt these.

jayRyan24:
So then do we add a validate for it? A lot of times people will just copy and paste things, and this will just spam and spit out errors if we don't.

combat-trainer.lic
jayRyan24: So then do we add a validate for it? A lot of times people will just copy and paste things, and this will just spam and spit out errors if we don't.

rpherbig: Maybe rename them to "Barbarian Research X"? But we already have quite a few guild-specific ones (Bardic Scream, Commune Meraud, ...) and it hasn't been a problem so far.

jayRyan24: I guess Barb Research X would be OK, and the reason I chose to check except for the others is because probably how I chose to name them. The others are kind of given, this is "Research Warding", etc, where people could think "I need to Research Warding because I use magic". I mean, it does make sense that people should be able to figure it out on their own to a certain point, but I was trying to keep the name of the ability shorter to match the others, and I am not really sure what the issue is with checking if the person is a Barbarian before putting it to the screen. Though I guess what could be the case is people would have bloated YAMLs because they have no real idea of what they're doing.

jayRyan24:
Ignore last commit.

PR #1053, opened 2016-11-21
Added support for sort auto head to equipment manager.

jayRyan24:
I believe this is the correct location for this.

### DIFF
--- a/equipmanager.lic
+++ b/equipmanager.lic
@@ -42,6 +42,7 @@ class EquipmentManager
 
   def wear_items(items_list)
     items_list.each { |item| wear_item?(item) }
+    fput("sort auto head")
   end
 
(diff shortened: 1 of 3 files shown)


equipmanager.lic
```diff
@@ -42,6 +42,7 @@ class EquipmentManager
 
   def wear_items(items_list)
     items_list.each { |item| wear_item?(item) }
+    fput("sort auto head")
```
rpherbig: Do you think we also need to do this in `wear_equipment_set?`?

jayRyan24: Yes, I mistook wear_item in wear_equipment_set's as wear_items. After looking more, I think this is a good place because it won't be called repeatedly, just the once.

Issue #791, opened 2016-09-14
Add Shard Walls to Athletics Training


jandersson:
I'd like to add the Shard walls to the athletics training portion of crossing-training by basically translating this https://elanthipedia.play.net/Climb_Shard_(script) to base-athletics

I notice the Crossing walls data is pulled from base-athletics. Any thoughts on how to add the Shard walls data to this and use it in `train_athletics`? I was thinking of making a new list on the same level as athletics_options, but named shard_options or shard_walls, then doing one of two things..
1. a shortest path calc to the first target in both athletics_options and shard_walls and use whichever list has shortest path to the first target.
2. if the person has `hometown: Shard` in their settings then pull the shard_walls list, else use athletics_options

The first would seem to work no matter what 'hometown' you have, the second one seems much easier to implement

rpherbig:
I think I'd be in favor of making `base-athletics` follow `base-town` and the like: have a list for Crossing, a list for Shard, etc. and pick based on your hometown.

rcuhljr:
I don't think the simple list of places is the correct looking term solution. Climbing should be outta own script and room lists may not even be in the tank since they are highly variable based on skill ranges, and we train climbing at least three different ways.

jandersson:
Just spitballing here:

What about storing the data in a database (a relational one), querying it, the resulting query contains a 'recipe' so to speak. The script just follows the recipe. This could allow feedback from the script. I think you're doing something similar with steal.lic

jandersson:
So just wanted to mess around with this on paper.. but heres a potential schema/query

rooms(id, province)
has_action(room_id, action_id)
actions(id, action, target, min_skill, max_skill)

```
select rooms.id, actions.action
from rooms
join has_action
on rooms.id = room_id
join actions
on actions.id = action_id
where rooms.province = 'zoluren'
and actions.max_skill < 100
and actions.min_skill > 50
```

rpherbig:
I think step one of any approach is to break athletics training into its own script.

jandersson:
I started by copying out the methods in crossing-trainer to an athletics script and just running that, the same way astrology is done. Noticing that there are some method calls in athletics that play zills, train outdoorsmanship, check research, and one called `check_osrel`. How should these be handled? 

Are these methods all candidates for common code or becoming their own scripts?

PR #1070, opened 2016-11-24
Support custom climb practice targets in crossing trainer

jandersson:
This slightly modifies the athletics training portion of CT to check for a user specified `climbing_target` if it exists, you go there hide and practice climb on the target. This is really similar to the `climb_branch` method below it. I've been testing this on the undergondola pit with good results (i would say it is possibly more time efficient than the crossing wall circuit at my athletics skill)

The climbing targets are listed in base-athletics and the idea was to choose these targets in a similar fashion to selecting a hunting zone. Its up to the user to decide if it meets their training requirements.

Some ideas for improvement:
1. Check if the player has gained any mind state after a climb practice attempt, if not then break and remove athletics from the training options. This'll keep the player from getting stuck there if they suddenly cross over some athletics skill threshold.
2. Check for a waggle-set called 'athletics' and cast whatever is specified in there... this might let someone cast invisibility before heading out to train


jandersson:
Handles issue #1070 and addresses the spirit of #791 (to have options in shard)

jandersson:
Wow i totally merged the master branch into this... i will close and reopen

PR #1073, opened 2016-11-24
Don't force Necromancers to always perform a ritual

ohbeanz:
Created an explicit setting for using rituals, so we don't force players with the Necromancer guild to always be performing a ritual. This allows for necros to 'turn off' the rituals and strictly perform skins/arranges if desired.


rpherbig:
I'm not a fan of this approach. I can see where you'd want to heal, but not regularly perform rituals. I think you can already do that with the existing settings:
```
thanatology:
  ritual_type:
  heal: true
```

How does this new setting differ?

ohbeanz:
No, a null ritual_type right now would not skin the critter. It would still heal, yes... but rituals would be broken.

Issue #1080, opened 2016-11-28
Vivisection spell should be treated as a stealth attack


jandersson:
As per the discussion on Slack:

The vivisection spell should be cast if the user has set `use_stealth_attacks: true`. This is reasonable because vivisection can be cast while from hiding and not hiding. So it is a snipe attack from hiding or just another TM spell. Its up to the user, and doesn't add any new settings/attributes.

PR #1083, opened 2016-11-29
hide if casting vivisection and using stealth attacks

jandersson:
addresses #1080

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -715,6 +715,8 @@ class SpellProcess
       end
     end
 
+    waitrt? while XMLData.prepared_spell == 'Vivisection' && game_state.use_stealth_attack? && ! hide?
+
     cast(@custom_cast, @before, @after)
     @custom_cast = nil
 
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -715,6 +715,8 @@ class SpellProcess
       end
     end
 
+    waitrt? while XMLData.prepared_spell == 'Vivisection' && game_state.use_stealth_attack? && ! hide?
```
rpherbig: I would not recommend using a `while` here. I prefer a single attempt to hide to avoid getting caught in a loop or spending a lot of time trying to hide if your chance of success is so-so.

rpherbig: The `hide?` helper also takes case of waitrt, so I don't think you need to do that here.

rpherbig: I think `DRSpells.prep_spell` is a better way to check the spell name, rather than using the XMLData itself.

rcuhljr: I'd use the xml over DRSpells, we made those DRSpells entries because we didn't know the game was tracking the data for us.

PR #1087, opened 2016-11-29
barb specific names - no if barb checks

jayRyan24:
Borked my previous branch and fork. Learned what to do. Hopefully this is what's been waited for to get this change through.


rpherbig:
You'll also want to update the test in `validate.lic` that checks for valid training_abilities.

Issue #1043, opened 2016-11-19
Athletics: Specify a 'climb practice' target similar to how hunting zones are done


jandersson:
Currently the athletics training utilizes two methods: running a circuit around crossing as well as climb practice around the Shard undergondola. I would argue that it would diversify training if we could have a list of climbing targets, similar to base-hunting, which we do climb practice on for athletics training. The climbing circuit is nice because it is lazy but it is difficult to design one that is efficient. 

I'd like to suggest an idea for training athletics where we specify an athletics_zone, just like hunting_zone, and we go there and do `climb practice thing`. It would be up to the user to decide if this is an appropriate spot for their skills.

jayRyan24:
Have you started groundwork on this yet? I'd let to set it up ASAP for places like Shard and Hib. Would like to spend this weekend on it actually, we can collaborate.

jayRyan24:
Could very well work like hunting does, A file Base-athletics.yaml, and the personal yaml could have something like:

athletics_training:
- :room: under_gondola
  args:
  - ladder

jandersson:
https://github.com/rpherbig/dr-scripts/pull/1090

jandersson:
Ive got something like that in place in the above pull request, let me know what you think. Would appreciate it.

PR #1090, opened 2016-11-29
Support Climb Practice in Crossing-Trainer

jandersson:
This re-opens #1070 which addresses the spirit of #791 and handles #1043 

The idea is to define targets to climb practice on. It is a nearly identical copy of the method in place to climb the branch in undergondola. I avoided changing any existing athletics data/methods because they are very specific. For example, the entire `athletics_options` list is used for the Crossing athletics circuit, so I did not want to mess with that data. The advantage of this implementation is that it is region independent and I am currently using it in Shard with great results. It will be up to the user to select targets which match their skills, just like they select a hunting zone.

I understand that an athletics rewrite is imminent but I feel like this will at least address the issue of "shard-trainer" not training athletics or any other "x-trainer".  

The user will specify a `climb_target: something_to_climb` in their setup file. When crossing-trainer trains athletics it checks if this exists, looks it up in the data, and climbs practice, else it behaves normally.


crossing-training.lic
```diff
@@ -29,8 +29,13 @@ class CrossingTraining
     args = parse_args(arg_definitions, true)
 
     @settings = get_settings(args.flex)
+
+
     @training_room = (@settings.training_rooms || [@settings.safe_room]).sample
     @athletics_options = get_data('athletics').athletics_options
+    @practice_options = get_data('athletics').practice_options
```
rpherbig: I recommend combining these two: `@practice_climbing_target = get_data('athletics').practice_options[@settings.climbing_target]`

profiles/Jonas-setup.yaml
```diff
@@ -1,15 +1,13 @@
...
-# safe_room: 2717 # Shard - Ironwood Tree
-# safe_room: 2715 # Shard - Old Mine, Bar in Blackthorn Canyon (has rocks!)
+# safe_room: 1645 # Crossing - Spooky Tree ooOoOoO
+safe_room: 2715 # Shard - Old Mine, Bar in Blackthorn Canyon (has rocks!)
 # Hunting Settings
 
-hometown: Crossing
+hometown: Shard
 slack_username: jonas
+climbing_target: undergondola_pit
```
rpherbig: There is a single `climbing_target`. Would you ever want to list multiple targets here (i.e. a list)?

crossing-training.lic
```diff
@@ -656,6 +665,19 @@ class CrossingTraining
     end
   end
 
+  def climb_practice
+    check_osrel unless @researching
```
rpherbig: Most of this logic appears to be a duplicate of `practice_branch`. How much can be extracted into a common method?

rpherbig:
Sorry for the delay in reviewing this

profiles/Jonas-setup.yaml
jandersson: I don't believe so. I'm able to train with this target, I dont know that practicing on multiple targets would help anything. I'm up for adding support for it

rpherbig: I think a single target is ok then.

crossing-training.lic
jandersson: I think it would be possible to make a common method from `practice_branch` but making the target and id variable like in `climb_practice`

jandersson:
The undergondola branch is going to be difficult for me to test with below the needed athletics, but i will give it a whirl

crossing-training.lic
```diff
@@ -684,7 +689,8 @@ class CrossingTraining
     if DRSkill.getxp('Athletics') < 29
       walk_to(2245)
       if UserVars.athletics < 550
-        practice_branch
+        @climbing_target['target'] = 'branch'
```
rpherbig: Does this work if you don't have a climbing_target set? I expected this line to throw an error since `@climbing_target` would be `nil`.

rpherbig: If this does work, won't that mean that `@climbing_target` is defined? I think that would mean the next time Athletics is trained, the `train_athletics` method would take the first branch, but `@climbing_target['id']` would not be defined.

jandersson: yea most likely it was a bad implementation. I've added the branch to the data and loaded that before going into the branch practice bit.

crossing-training.lic
```diff
@@ -679,18 +685,18 @@ class CrossingTraining
 
   def climb_branch
     start_script('skill-recorder') unless Script.running?('skill-recorder')
-    walk_to(2245)
+    walk_to @climbing_target['id']
     check_research
     if DRSkill.getxp('Athletics') < 29
-      walk_to(2245)
+      walk_to @climbing_target['id']
       if UserVars.athletics < 550
```
rpherbig: This check is against 550 athletics, but the comment in `base-athletics` says undergola_branch is for 530, and the `if` in `train_athletics` is against 400. Which one is correct?

crossing-training.lic
```diff
@@ -679,18 +685,18 @@ class CrossingTraining
...
-        practice_branch
+        climb_practice
       else
         play_song?
         walk_to(9607)
         walk_to(19_464)
         while DRSkill.getxp('Athletics') < 29
-          walk_to(2245)
+          walk_to climbing_target['id']
```
rpherbig: Should this be `@climbing_target['id']`?

crossing-training.lic
```diff
@@ -642,8 +643,13 @@ class CrossingTraining
...
 
   def train_athletics
-    if DRSkill.getrank('Athletics') >= 400
+    if @climbing_target
       return if @researching
+      walk_to(@climbing_target['id'])
+      climb_practice
+    elsif DRSkill.getrank('Athletics') >= 400
+      return if @researching
+      @climbing_target = get_data('athletics').practice_options['undergondola_branch']
```
rpherbig: Since this assigns to `@climbing_target`, does that mean the next time Athletics is trained, `train_athletics` would take the first branch (`climb_practice`) instead of the second (`climb_branch`)?

crossing-training.lic
jandersson: I'm not really sure, I havent had a character have enough athletics use this part of the script yet

crossing-training.lic
jandersson: yea that seems like a huge mistake, maybe climb practice method can just take the target as an argument and we avoid setting `@climbing_target` for the user

rcuhljr:
👍

PR #1094, opened 2016-12-01
Updates to steal.lic and base-stealing.yaml

Endomlik:
Expanded stealing beyond Crossing

example Player-setup.yaml
steal_town:
  - Crossing
  - Muspar'i

### DIFF
--- a/steal.lic
+++ b/steal.lic
@@ -15,7 +15,7 @@ class Steal
 
     UserVars.stealing_timers ||= {}
 
-    @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
+    @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) || !settings.steal_town.include?(target['steal_town'])}
     @stealing_bag = settings.stealing_bag
     @bin_stolen = settings.bin_stolen
     @hide_to_steal = settings.hide_to_steal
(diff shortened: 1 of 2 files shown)


rpherbig:
@rcuhljr Was this the same idea behind the `province` attribute that already exists in base-stealing? I noticed that `province` isn't used anywhere, and wonder if it should be removed in favor of this new attribute.

steal.lic
```diff
@@ -15,7 +15,7 @@ class Steal
 
     UserVars.stealing_timers ||= {}
 
-    @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
+    @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) || !settings.steal_town.include?(target['steal_town'])}
```
rpherbig: Style-wise, I'd prefer this to be a new clause, something like:
```
stealing_options
  .reject {...}
  .select {...}
```

rcuhljr:
I named it province just as a guess of thats how we would want to toggle stealing areas on and off. I think it should probably be renamed to something more granular like Crossing, Arthe Dale, etc and like people have a mining vein list they have a stealing area list.

steal.lic
```diff
@@ -153,7 +153,8 @@ class Steal
     {
       'new' => true, 'room' => Room.current.id, 'province' => 'zoluren',
```
Endomlik: I was wondering if we should keep province as you mentioned below since it is such a board area.  When my thief is in Muspar'i, I wouldn't want him trying to run to Riverhaven to steal.  I left it the same because it wasn't doing anything.

steal.lic
rcuhljr: Yes just add a default steal list to base.yaml

PR #1106, opened 2016-12-02
Update combat-trainer.lic

cpryor6:
Match for stun weapon in offhand until ambush stun can evaluate weapon hand

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1189,7 +1189,7 @@ class TrainerProcess
     end
 
     if hide?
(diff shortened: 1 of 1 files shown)


rpherbig:
What is the problem this is fixing?

cpryor6:
When 'offhand weapon' is configured to the same as 'stun weapon' and interval for ambush stun arrives when attacking with offhand weapon, "ambush stun" command is immediately sent.  This results in the "wouldn't it be better" message and the 15 second timeout.  Combat trainer is broken after that , trying to load ranged weapon with empty hands, etc.  This just adds the match; real fix is to evaluate and if needed, execute swap to place stun weapon in main hand before attempting an ambush stun.

rpherbig:
Could you open an issue for the remaining issue?

If you're interested in fixing it yourself, the `game_state.offhand?` helper should get you part of the way there :)

PR #1139, opened 2016-12-11
Train Astrology in the Astral Plane

jandersson:
Train astrology by walking the astral plane. A timer gets set on each round-trip from source to destination and back to source. A walk can be done once an hour

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -46,7 +46,7 @@ class Astrology
     return unless @ap_training
     return unless @ap_source
     return unless @ap_destination
-    return unless UserVars.astral_plane_exp_timer.nil? && Time.now - UserVars.astral_plane_exp_timer < 3600
+    return if UserVars.astral_plane_exp_timer.nil? && Time.now - UserVars.astral_plane_exp_timer < 3600
 
     wait_for_script_to_complete('bescort', ['ways', @ap_destination])
     UserVars.astral_plane_exp_timer = Time.now
--- a/astrology.lic
+++ b/astrology.lic
@@ -2,19 +2,25 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#astrology
 =end
 
-custom_require.call(%w(common common-arcana drinfomon spellmonitor))
+custom_require.call(%w(common common-arcana common-travel drinfomon spellmonitor))
 
 class Astrology
   include DRC
   include DRCA
+  include DRCT
 
   def initialize
     settings = get_settings
     @constellations = get_data('constellations').constellations
     @have_telescope = settings.have_telescope
+    @ap_training = settings['astral_plane']['train_in_ap']
+    @ap_source = settings['astral_plane']['train_source']
+    @ap_destination = settings['astral_plane']['train_destination']
+    @safe_room = settings.safe_room
 
     do_buffs(settings)
 
+    check_astral
     train_astrology
   end
 
@@ -33,6 +39,20 @@ class Astrology
       result << all_bodies.find { |body| /\b#{body['name']}\b/i =~ line && line !~ /below the horizon/ }
     end
     result.compact.select { |data| data['circle'] <= DRStats.circle }
+  end
+
+  def check_astral
+    return unless DRStats.circle > 99
+    return unless @ap_training
+    return unless @ap_source
+    return unless @ap_destination
+
+    if UserVars.astral_plane_timer.nil? || Time.now - UserVars.astral_plane_timer > 3600
+      wait_for_script_to_complete('bescort', ['ways', @ap_destination])
+      wait_for_script_to_complete('bescort', ['ways', @ap_source])
+      UserVars.astral_plane_timer = Time.now
+      walk_to @safe_room
+    end
   end
 
   def train_astrology
@@ -54,7 +74,7 @@ class Astrology
     Flags.add('bad-search', 'is foiled by the (daylight|darkness)', 'turns up fruitless')
 
     observed = if @have_telescope
-                 bput('get telescope', 'you get', 'You are already')
+                 bput('get telescope', 'you get', 'You are already', "That can't be picked up")
                  things_to_try = @constellations.select do |data|
                    data['telescope'] &&
                      data['circle'] <= DRStats.circle &&
--- a/astrology.lic
+++ b/astrology.lic
@@ -46,7 +46,7 @@ class Astrology
     return unless @ap_training
     return unless @ap_source
     return unless @ap_destination
-    return if UserVars.astral_plane_exp_timer.nil? && Time.now - UserVars.astral_plane_exp_timer < 3600
+    return if ! UserVars.astral_plane_exp_timer.nil? && Time.now - UserVars.astral_plane_exp_timer < 3600
 
     wait_for_script_to_complete('bescort', ['ways', @ap_destination])
     UserVars.astral_plane_exp_timer = Time.now
--- a/astrology.lic
+++ b/astrology.lic
@@ -2,12 +2,11 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#astrology
 =end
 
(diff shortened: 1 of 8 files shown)


astrology.lic
```diff
@@ -54,7 +74,7 @@ class Astrology
     Flags.add('bad-search', 'is foiled by the (daylight|darkness)', 'turns up fruitless')
 
     observed = if @have_telescope
-                 bput('get telescope', 'you get', 'You are already')
+                 bput('get telescope', 'you get', 'You are already', "That can't be picked up")
```
jandersson: This is for telescopes in the room, found in most moonmage guilds.

jandersson:
Reminder to self to document settings in astrology and bescort headings on elanthipedia

astrology.lic
```diff
@@ -2,19 +2,25 @@
...
 class Astrology
   include DRC
   include DRCA
+  include DRCT
 
   def initialize
     settings = get_settings
     @constellations = get_data('constellations').constellations
     @have_telescope = settings.have_telescope
+    @ap_training = settings['astral_plane']['train_in_ap']
```
rpherbig: These could be `settings.astral_plane['sub-setting-name']`.

astrology.lic
```diff
@@ -2,19 +2,25 @@
...
+    @ap_source = settings['astral_plane']['train_source']
```
rpherbig: It looks like some of the sub-settings of `astral_plane` are used in astrology.lic and others in bescort.lic, which could be confusing.

astrology.lic
```diff
@@ -35,6 +41,20 @@ class Astrology
...
+  def check_astral
+    return unless DRStats.circle > 99
+    return unless @ap_training
+    return unless @ap_source
+    return unless @ap_destination
+
+    if UserVars.astral_plane_timer.nil? || Time.now - UserVars.astral_plane_timer > 3600
+      wait_for_script_to_complete('bescort', ['ways', @ap_destination])
+      wait_for_script_to_complete('bescort', ['ways', @ap_source])
+      UserVars.astral_plane_timer = Time.now
```
rpherbig: The timer is set after you astral walk. Does that mean the timer never gets set if you die while walking?

astrology.lic
jandersson: I could rename it to `astral_plane_training` and move the `exit_if_killed` outside. `exit_if_killed` would have to be renamed to something more clear like `quit_on_death_in_ap`

astrology.lic
jandersson: to be clear, the timer is an exp timer. someone mentioned that it only teaches every hour, so i put that in there

rpherbig: What do you think about `astral_plane_exp_timer`?

jandersson: sounds more clear. should I add it to the crossing-trainer timers or shall it stand on its own?

jandersson: In which case it'd be `UserVars.exp_timers['astral_plane']` does that make more sense?

rpherbig:
Since this is a standalone script, I'm hesitant to reuse crossing training
vars. What do you think?

jandersson:
Yea that makes perfect sense, ill go for the rename

astrology.lic
```diff
@@ -35,6 +41,20 @@ class Astrology
...
+    return unless DRStats.circle > 99
+    return unless @ap_training
+    return unless @ap_source
+    return unless @ap_destination
+
+    if UserVars.astral_plane_exp_timer.nil? || Time.now - UserVars.astral_plane_exp_timer > 3600
+      wait_for_script_to_complete('bescort', ['ways', @ap_destination])
+      wait_for_script_to_complete('bescort', ['ways', @ap_source])
+      UserVars.astral_plane_exp_timer = Time.now
+      walk_to @safe_room
```
rpherbig: My initial thought is that this shouldn't be astrology's concern. I would expect it to be the caller's responsibility to decide where to go next, which may not be the safe_room (crossing-training has a fair bit of logic just for this case, to avoid running around when not necessary).

Are any of the start or end rooms truly dangerous? E.g. populated with mobs?

astrology.lic
```diff
@@ -35,6 +41,20 @@ class Astrology
     result.compact.select { |data| data['circle'] <= DRStats.circle }
   end
 
+  def check_astral
+    return unless DRStats.circle > 99
+    return unless @ap_training
+    return unless @ap_source
+    return unless @ap_destination
+
+    if UserVars.astral_plane_exp_timer.nil? || Time.now - UserVars.astral_plane_exp_timer > 3600
```
rpherbig: You could move these checks to be returns.

astrology.lic
```diff
@@ -35,6 +41,20 @@ class Astrology
...
+      wait_for_script_to_complete('bescort', ['ways', @ap_destination])
```
rpherbig: What does bescort do if your MM is circle 100 but f2p (i.e. can't do astral travel)? I guess in that case you just wouldn't use these settings...?

astrology.lic
jandersson: The rooms are safe. But some of them are inside so training with a telescope will fail. The script could just exit after doing a walk instead of going into the `train astrology` method

astrology.lic
jandersson: I'm not sure what happens, but bescort does a check to make sure you know the destination shard. So presumably this would call bescort twice, set the timer and move on

astrology.lic
rpherbig: Walking should mostly fill your astrology exp right?

jandersson: crossing-trainer could walk to safe room before and after executing astrology

jandersson: yea, it gets me to about 27-29/34. So the telescope training afterwards only serves to bump it up another notch or two.

astrology.lic
```diff
@@ -46,7 +46,7 @@ class Astrology
     return unless @ap_training
     return unless @ap_source
     return unless @ap_destination
-    return unless UserVars.astral_plane_exp_timer.nil? && Time.now - UserVars.astral_plane_exp_timer < 3600
+    return if UserVars.astral_plane_exp_timer.nil? && Time.now - UserVars.astral_plane_exp_timer < 3600
```
jandersson: This is wrong, fixing. should be `if !`

crossing-training.lic
```diff
@@ -771,6 +771,7 @@ class CrossingTraining
 
   def train_astrology
     wait_for_script_to_complete('astrology')
+    walk_to(@training_room) unless Room.current.id == @training_room
```
jandersson: It doesn't, gonna remove the moving bit entirely

jandersson:
Tested with the latest changes for a bit. Exiting astrology in the shard room doesn't seem to be any issue

astrology.lic
```diff
@@ -2,19 +2,25 @@
...
+  include DRCT
 
   def initialize
     settings = get_settings
     @constellations = get_data('constellations').constellations
     @have_telescope = settings.have_telescope
+    @ap_training = settings.astral_plane_training['train_in_ap']
+    @ap_source = settings.astral_plane_training['train_source']
+    @ap_destination = settings.astral_plane_training['train_destination']
+    @safe_room = settings.safe_room
```
rpherbig: This can be removed.

astrology.lic
```diff
@@ -2,19 +2,25 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#astrology
 =end
 
-custom_require.call(%w(common common-arcana drinfomon spellmonitor))
+custom_require.call(%w(common common-arcana common-travel drinfomon spellmonitor))
```
rpherbig: Is this (and line 10) still needed?

jandersson: No that was for `walk_to`, deleting

PR #1141, opened 2016-12-12
Fix for the 'You don't need to. . .' bug in crossing-repair

b-Y-d:
In case of hand-exploding backlash. For crossing-trainer.lic
I am new to this githubbery

### DIFF
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -32,9 +32,7 @@ class CrossingRepair
       fput('swap') unless right_hand
       command = "give #{repairer}"
     else
-      unless item.name.split.last == (left_hand.split.last || right_hand.split.last)
-	return unless @equipment_manager.get_item?(item)
-      end
+      return unless @equipment_manager.get_item?(item)
       command = "give my #{item.short_name} to #{repairer}"
     end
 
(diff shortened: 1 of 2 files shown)


rpherbig:
`Sorcery.lic` *should* already be doing this: https://github.com/rpherbig/dr-scripts/blob/master/sorcery.lic#L42

Is it not working?

b-Y-d:
I think it missed it at least once, found my bard logged via afk.lic sitting in my training room. (Not in the sorcery room). I'd have to check the logs to see what happened exactly.

rpherbig:
If you could check your logs, that would be great

crossing-repair.lic
```diff
@@ -32,11 +32,13 @@ class CrossingRepair
       fput('swap') unless right_hand
       command = "give #{repairer}"
     else
-      return unless @equipment_manager.get_item?(item)
+      unless item.name.split.last == left_hand.split.last || right_hand.split.last
```
rcuhljr: This won't work like you want, == has a higher operator precedence than ||.  [Precedence table](https://ruby-doc.org/core-2.2.0/doc/syntax/precedence_rdoc.html)

crossing-repair.lic
```diff
@@ -32,11 +32,13 @@ class CrossingRepair
       fput('swap') unless right_hand
       command = "give #{repairer}"
     else
-      return unless @equipment_manager.get_item?(item)
+      unless item.name.split.last == (left_hand.split.last || right_hand.split.last)
```
rpherbig: The `split` here worries me. Shouldn't they match without splitting?

b-Y-d:
**crossing-training.lic:** Removed the change/addition of running safe-room after sorcery to. Looks like I had safe-room paused for some reason when sorcery tried to run it that one time.

**crossing-repair.lic:** removed the if block with the splits, combined the two repeat cases.

PR #1153, opened 2016-12-14
multi-difficulty support for workorders.lic

b-Y-d:
Small change that does not require any changes to current yamls.
Allows for the current:
```
workorder_diff: hard
```
or
```
workorder_diff:
  shaping: hard
  blacksmithing: chall
```
etc. 
Helps when you've got techs and/or career/hobby in one disciple but struggle in another.

### DIFF
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -32,9 +32,7 @@ class CrossingRepair
       fput('swap') unless right_hand
       command = "give #{repairer}"
     else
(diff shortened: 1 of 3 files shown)


rpherbig:
This looks like a reasonable change. Would you be willing to upload your YAML too, so people can see an example?

We'll want to update epedia after merging this.

To clean up the history, you can pull our master into your fork's master, then create a new branch from there.

PR #1155, opened 2016-12-14
Multi-difficulty support for workorders.lic

b-Y-d:
Allows additional support for the current yaml setting:
```
workorder_diff: hard
```
to handle assigning individual disciples their own difficulty:
```
workorder_diff:
  shaping: hard
  blacksmithing: challenging
```
etc

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -56,8 +56,11 @@ class WorkOrders
       exit
     end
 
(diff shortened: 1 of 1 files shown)


rpherbig:
This looks like a reasonable change. Would you be willing to upload your YAML too, so people can see an example?

We'll want to update epedia after merging this.

PR #1157, opened 2016-12-15
Adding support for hib's furrier!

jayRyan24:
Adding support for Hib's furrier messaging. 

We could also remove the "carefully" from 'sorts through it carefully', and it would work for Hib. Just let me know!


rpherbig:
Removing carefully seems fine to me.

jayRyan24:
Do I need to do anything else for this?

jayRyan24:
Do I need to do anything else for this?

PR #1169, opened 2016-12-18
Seordhevor Kartais (180-250) hunting support.

b-Y-d:
https://elanthipedia.play.net/Seordhevor_kartais

These guys make a decent gap-filler for after black leucros for people stuck in P1.
I've been using this general setup with training-manager. It's been working but I do not think the format
is what you guys will want for any final addition. 

Any help getting this all fixed up would be nice! (also ;go2 can find its way out of the mansion fine.)


oshu_manor.lic
```diff
@@ -0,0 +1,148 @@
+custom_require.call(%w(common common-travel drinfomon))
```
rpherbig: You need `events` here since you are using Flags

oshu_manor.lic
```diff
@@ -0,0 +1,148 @@
...
+    fput("south")
+  end
+
+  def head_reset?
+    return unless Flags['head-reset']
+    Flags.reset('head-reset')
+    head_game
+  end
+
+  def waitandcheck
```
rpherbig: What do you think about naming this `wait_and_check`?

oshu_manor.lic
```diff
@@ -0,0 +1,148 @@
...
+    Flags.add('allready', "The gargoyle head doesn't budge.")
+    Flags.add('blasted', "The blast catches you fully and sends you careening northward through the air!")
+    Flags.add('head-reset', 'A stone gargoyle head suddenly makes a series of loud grinding')
+    Flags.add('grave_exit', 'You dig to the surface')
+    Flags.add('retreat', 'closes to pole weapon range on you', 'closes to melee range on you', 'begins to advance on you', 'advances from nearby and is closing steadily' )
+    grave_digger
+  end
+  
+  def germfree?
+    return (DRRoom.npcs - @ignored_npcs).empty?
```
rpherbig: Is this used to find an empty hunting room? If so, it also needs to check for players.

oshu_manor.lic
```diff
@@ -0,0 +1,148 @@
...
+  end
+
+  def through?
+    return unless Flags['grave_exit']
+    Flags.reset('grave_exit')
+    DRC.fix_standing
```
rpherbig: You don't need the `DRC` prefix here since it is `include`d above.

oshu_manor.lic
```diff
@@ -0,0 +1,148 @@
...
+    puzzle_begin
+  end
+
+  def retreat?
```
rpherbig: There is a helper in `DRCT` called `retreat`

oshu_manor.lic
```diff
@@ -0,0 +1,148 @@
...
+      pause 0.5
+    end
+    puzzle_begin
+  end
+    
+  def puzzle_begin
+    walk_to(2350)
+    bput('push statue', 'and it settles into its new position with a click', 'The granite statue has already been pushed')
+    walk_to(2340)
+    fput('go house')
```
rpherbig: Should these cases instead use the `move` helper?

data/base-hunting.yaml
```diff
@@ -1,599 +1,604 @@
----
```
rpherbig: GitHub seems to think the entire `base-hunting.yaml` file changed. Do you know what happened here?

bescort.lic
```diff
@@ -243,7 +249,17 @@ class Bescort
...
   end
-
+	
+	def oshu_manor(mode) 
+		case mode
+		when /exit/i
+      walk_to(2317)
+    when /enter/i
+			wait_for_script_to_complete('oshu_manor')
+			find_room_list(%w(e e w w w n n n))
```
rpherbig: Is this redundant with the 'empty room' check in `oshu_manor.lic`?

rpherbig: I was surprised to see that the `find_room_list` logic does not match other logic for finding an empty room (e.g. it doesn't use the `hunting_buddies` setting). @rcuhljr might know if that is intentional or not...

oshu_manor.lic
b-Y-d: This was to check for no germish'din (330 to 520) in the room. The player needs to dig grave (8 second rt + 4-5 seconds of wait time) to get to the next area. Its to avoid getting hurt on the way to the hunting zone.

data/base-hunting.yaml
b-Y-d: Github novice stuff.

bescort.lic
b-Y-d: oshu_manor.lic's 'empty room' check is entirely to avoid running into higher-circle mobs on the way to the manor. It's basically a long bescort.lic routine that gets the player from one relatively safe room (the **base** room outside of germish'dins ) to just inside the manor (manor's door is locked until the puzzle is solved).

b-Y-d: http://pastebin.com/raw/NbkjVw5R

That's a log of what oshu_manor.lic does (when not running into trouble).

oshu_manor.lic
b-Y-d: No problem I'll rework the script with your feedback later today.

oshu_manor.lic
rpherbig: Is there a name that would be clearer?

b-Y-d:
Added in the suggested changes to oshu_manor.lic
Will comment once I've tested to make sure I didn't break anything during the edits.

bescort.lic
```diff
@@ -99,6 +99,10 @@ class Bescort
         { name: 'wilds', regex: /wilds/i, description: 'The Leucro and Geni wilds on the NTR.' },
         { name: 'mode', options: %w(exit leucro1 leucro2 geni), description: 'Where do you need to get to?' }
       ],
+			[
+        { name: 'oshu_manor', regex: /oshu_manor/i, description: 'The Leucro and Geni wilds on the NTR.' },
```
rpherbig: Is this description correct?

oshu_manor.lic
```diff
@@ -0,0 +1,140 @@
...
+    word_list = reget(4)
+      .flatten
+    head_turner(word) unless word_list.any? {|line| line.include?("You grab the gargoyle head and wrestle it around until it points at the word \"#{word}.\"")}
+  end  
+
+  def gtfo
+    move('north')
+    move('go door')
+    pause 1
+    unless Room.current.id.to_i == 2340
```
rpherbig: `Room.current.id` already returns an int, so you can remove the `to_i` calls.

bescort.lic
b-Y-d: No, not at all.

oshu_manor.lic
b-Y-d: k

bescort.lic
```diff
@@ -100,6 +100,10 @@ class Bescort
         { name: 'mode', options: %w(exit leucro1 leucro2 geni), description: 'Where do you need to get to?' }
       ],
       [
+        { name: 'oshu_manor', regex: /oshu_manor/i, description: 'The Seordhevor kartais in Oshu'ehhrsk Manor. Near germish'din on the STR.' },
```
b-Y-d: Added description, fixed the funky tab issues.

oshu_manor.lic
```diff
@@ -0,0 +1,140 @@
...
+    word_list = reget(4)
+      .flatten
+    head_turner(word) unless word_list.any? {|line| line.include?("You grab the gargoyle head and wrestle it around until it points at the word \"#{word}.\"")}
+  end  
+
+  def gtfo
+    move('north')
+    move('go door')
+    pause 1
+    unless Room.current.id == 2340
```
b-Y-d: Changed all instances of **Room.current.id.to_i** to **Room.current.id**

b-Y-d:
Added a description to the bescort args, fixed some indentation issues.
Removed the .to_i from all Room.current.id instances in oshu_manor.lic.

Issue #1170, opened 2016-12-18
Doesn't withdraw enough money to fill lockpick ring


richardhertzog:
[pick]>offer 125
 
Ragge hands over your lockpick.
>
[pick]>put my lockpick on my lockpick ring
You put an ordinary lockpick on your lockpick ring.
>
[pick]>buy ordinary lockpick
 
Ragge sighs.  "Despite the rarity of this lockpick, I'm prepared to offer it to you for 125 kronars."
>
[pick]>offer 125
 
Ragge nudges you and winks.  "Well, I could loan you the kronars, but with the rates that I charge, you'd be better off using your own funds to pay for the purchase."
>
[pick]>put my lockpick on my lockpick ring
What were you referring to?
>
[pick: *** No match was found after 15 seconds, dumping info]
[pick: messages seen length: 4]
[pick: message: What were you referring to?]
[pick: message: Hands of Lirisa  (11 roisaen)]
[pick: message: Psychic Shield  (Fading)]
[pick: message: Clear Vision  (3 roisaen)]
[pick: checked against [/You put/i]]
[pick]>buy ordinary lockpick
 
Ragge sighs.  "Despite the rarity of this lockpick, I'm prepared to offer it to you for 156 kronars."
>
[pick]>offer 156
 
Ragge nudges you and winks.  "Well, I could loan you the kronars, but with the rates that I charge, you'd be better off using your own funds to pay for the purchase."
>
[pick]>put my lockpick on my lockpick ring
What were you referring to?
>

basically does this forever

b-Y-d:
I wrote up a quick fix for this, I'll put it in a PR.

PR #1177, opened 2016-12-20
re: issue #1170, Refilling lockpick rings

b-Y-d:
- Removed the **ensure_copper_on_hand** helper from **pick.lic** and put it inside **refill_lockpick_container**.
- Added an optional **cost** (defaults to 125) variable to **refill_lockpick_container**.
- pick.lic now passes the lockpick cost to **refill_lockpick_container**
- **refill_lockpick_container** determines how much coin to withdraw based on the appraisal regex. **count** * **cost**.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -80,13 +80,14 @@ module DRCI
     result =~ /You tap/
   end
 
-  def refill_lockpick_container(lockpick_type, hometown, container)
+  def refill_lockpick_container(lockpick_type, hometown, container, cost = 125)
     app_result = DRC.bput("app my lockpick #{container}", 'and it appears to be full', 'and it might hold an additional \d+', '\d+ lockpicks would probably fit')
     waitrt?
     return unless app_result =~ /(\d+)/
 
     room = get_data('town')[hometown]['locksmithing']['id']
     count = Regexp.last_match(1).to_i
+    ensure_copper_on_hand(cost * count, hometown)
     count.times do
       buy_item(room, "#{lockpick_type} lockpick")
       DRC.bput("put my lockpick on my lockpick #{container}", 'You put')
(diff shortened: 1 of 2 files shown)


common-items.lic
```diff
@@ -80,13 +80,14 @@ module DRCI
...
 
-  def refill_lockpick_container(lockpick_type, hometown, container)
+  def refill_lockpick_container(lockpick_type, hometown, container, cost = 125)
     app_result = DRC.bput("app my lockpick #{container}", 'and it appears to be full', 'and it might hold an additional \d+', '\d+ lockpicks would probably fit')
     waitrt?
     return unless app_result =~ /(\d+)/
 
     room = get_data('town')[hometown]['locksmithing']['id']
     count = Regexp.last_match(1).to_i
+    ensure_copper_on_hand(cost * count, hometown)
```
rpherbig: You will need to use the `DRCM` prefix here, and update the include and custom_require lists above.

rpherbig: Rather than introducing a new dependency, you could split `refill_lockpick_container` into two methods, one to count your ring and one to refill.

common-items.lic
```diff
@@ -80,13 +80,14 @@ module DRCI
     result =~ /You tap/
   end
 
-  def refill_lockpick_container(lockpick_type, hometown, container)
+  def refill_lockpick_container(lockpick_type, hometown, container, cost = 125)
```
rpherbig: I'm not sure I like having a default value here... but if we do, I suggest erring on the side of the most expensive lockpick type.

b-Y-d:
- Split **refill_lockpick_container** into itself and **count_lockpick_container***
- Removed the cost variable from **refill_lockpick_container**

Need to test, currently untested!

b-Y-d:
Tested, found and fixed one error. Looks like it works fine.

PR #1189, opened 2016-12-22
Oshu_manor.lic - Some over-zealous empty-hand checks added.

b-Y-d:
Someone discovered that not having empty hands will break the script, further testing showed that it could break in multiple areas if one or both hands are full.

I added `hands_free` to do a `stow_hands` only if anything is actually in hand. This avoids the 1 second pause from calling `stow_hands` directly.

### DIFF
--- a/oshu_manor.lic
+++ b/oshu_manor.lic
@@ -1,8 +1,9 @@
-custom_require.call(%w(common common-travel events drinfomon))
+custom_require.call(%w(common common-travel common-items events drinfomon))
 
 class Oshu
   include DRC
   include DRCT
+  include DRCI
 
   def initialize
     settings = get_settings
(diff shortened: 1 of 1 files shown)


rpherbig:
The pause in `stow_hands` is needed because it can take a moment for DR to send updated XML. But I think the pause could be reduced to 0.5 seconds. @rcuhljr thoughts?

b-Y-d:
Mind pushing this change for the moment? It'll serve better than 0 hand checks as it currently is.

rcuhljr:
I'm not a fan of the pause, since if it's really needed that just means things break during lag, otherwise we're wasting a lot of seconds pausing anytime stow_hands gets called (Which is everywhere). Why not kill the pause entirely and make stow_hand handle trying to stow an empty hand gracefully?

rpherbig:
It looks like that pause was added back in June during the forge/smith refactor (https://github.com/rpherbig/dr-scripts/commit/da8c87b03fabf899876c640212d7fc5362b57d50).

I'm inclined to remove the pause and see if anything misbehaves... I'll try that locally tonight.

@rcuhljr What did you have in mind for handling problems more gracefully?

PR #1202, opened 2016-12-25
Added support for before and after hunt scripts

adamgoerdt:
Will kick off listed scripts before or after a hunt

Example yaml snippet:

```
hunting_info:
- :zone: cougars
  args:
  - stealth
  :duration: 10
  stop_on:
  - Stealth
  before:
  - buff
  after:
  - workorders shaping
```


rpherbig:
Somehow I missed this PR. It's very similar to #1211 - do we need to make this more complicated change or is that change flexible enough?

adamgoerdt:
I used buffs as an example but it's not limited to it.  I've actually come to love the before/after sections as it gives me a huge amount of flexibility with how I train.  I've added 'attunement' runs when changing hunting areas which seems to be the only way to keep my attunement above 2/34 for the hunts.  I have a custom script to release Mantle of Flame prior to my stealth hunt since it will occasionally still be on me at the start of that hunt and prevents me from hiding.  I'm also using it on my bard so that I start and end hunts with engineering workorders, which resolves an issue I had with him in that I couldn't train crafting with hunting_priority set to false since I'd almost never get to hunting.  

Another thing with before: is that you can support different buffs for different hunting areas since parameters are passed

jayRyan24:
As a side thing to your MOF problem. You could add this to your stealth backtrain YAML and not need a separate YAML to support that:

jayRyan24:
I know Byd also does this to keep all of his skills moving. He has something incredibly close to this setup.

adamgoerdt:
It's really handy.  I just realized that one of my characters was running
in to combat with a 2/11 burden from extra lumber due to engineering so I
built a simple 'dump_lumber' script to clear his burden before the hunt
(and now he doesn't get killed!).

I've been running it on two characters for about a month now so it's well
tested.

rpherbig:
I think there was a merge error somewhere. The only change shown by the diff is adding the ice_adders hunting zone.

rpherbig:
Re-opened in #1318

Issue #167, opened 2016-01-07
Combat prep section


rcuhljr:
I'd like to stop in a room outside the hunting area and put up a few pre-buffs. Cyclic debuffs, shield of light, etc.

rpherbig:
This would be great for summoned weapons too.

rcuhljr:
combat prep and clean up, add optional waypoints towards and away from hunting spot

```
hunting_info:
- :zone: black_goblins
  entry:
  - 851
  - 4366
  - 3457
  prebuff:
    room: 4366
    actions:
    - Instinct
    - Hands of Lirisia
  exit:
  - 3457
  - 851
  args:
  - d0
  stop_on:
  - Light Armor
  - Small Edged
  - Shield Usage
  - Parry Ability
  - Bow
  - Evasion
  - Stealth

```

rpherbig:
This would alleviate a lot of the problems in #396

drboldwin:
Can't we just call waggle in hunting-buddy before it enters the zone?  Maybe start mapping out a room to each hunting area just outside it, that hunting-buddy stops at first?

jandersson:
Using a waggle set sounds pretty cool

rpherbig:
Sounds neat... one complication I can think of is that you probably want different waggle settings per hunt (not just hunting zone since you can hunt repeatedly in the same zone).

I think it would be sufficient to have one waggle setting (e.g. 'hunting-prep') and rely on the user to redefine it in each of their hunting definitions. Thoughts?

rcuhljr:
Just note that the waggle definitions all still need to live in `-setup.yaml` since `;buff` can't take a file arg, or at least not gracefully.

rpherbig:
Oh, that means we couldn't take my approach. We'd need X number of waggle sets defined in *-setup.yaml then... maybe 'hunting-prep-*', where * is the name of the YAML file?

PR #1211, opened 2016-12-28
Add basic prebuff support for hunting buddy

jayRyan24:
Tired of separate yamls to prebuff, adding initial, basic, support for this feature.

### DIFF
--- a/buff.lic
+++ b/buff.lic
@@ -21,6 +21,10 @@ class Waggle
 
     waggle_spells = @settings.waggle_sets
 
+    if args.spells['prehunt_buffs']
+      return unless waggle_spells[args.spells]
+    end
+
     spells = args.spells ? waggle_spells[args.spells] : waggle_spells['default']
 
     check_spells(spells)
--- a/buff.lic
+++ b/buff.lic
@@ -21,12 +21,10 @@ class Waggle
 
     waggle_spells = @settings.waggle_sets
 
-    if args.spells['prehunt_buffs']
-      return unless waggle_spells[args.spells]
-    end
-
     spells = args.spells ? waggle_spells[args.spells] : waggle_spells['default']
 
+    return unless spells
+    
     check_spells(spells)
   end
 
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -12,12 +12,22 @@ class CrossingRepair
   def initialize
     setup
 
-    @repair_info.each do |repairer, items|
+    @repair_info_metal.each do |repairer, items|
       echo "Repairing at #{repairer}" if UserVars.crossing_repair_debug
       repair_at(repairer['name'], repairer['id'], items)
     end
 
-    @repair_info.to_a.reverse.each do |repairer, _items|
+    @repair_info_leather.each do |repairer, items|
+      echo "Repairing at #{repairer}" if UserVars.crossing_repair_debug
+      repair_at(repairer['name'], repairer['id'], items)
+    end
+
+    @repair_info_leather.to_a.reverse.each do |repairer, _items|
+      while tickets_left?(repairer)
+      end
+    end
+
+    @repair_info_metal.to_a.reverse.each do |repairer, _items|
       while tickets_left?(repairer)
       end
     end
@@ -81,6 +91,8 @@ class CrossingRepair
     return if items.nil? || items.empty?
     walk_to target_room
 
+    echo "Repairer: #{repairer} target_room: #{target_room} items: #{items}"
+
     items.each { |item| repair(repairer, item) }
   end
 
@@ -94,9 +106,12 @@ class CrossingRepair
 
     ensure_copper_on_hand(@repair_withdrawal_amount, settings.hometown)
 
-    @repair_info = {
-      hometown['metal_repair'] => @equipment_manager.items.reject(&:leather),
+    @repair_info_leather = {
       hometown['leather_repair'] => @equipment_manager.items.select(&:leather)
+    }
+
+    @repair_info_metal = {
+      hometown['metal_repair'] => @equipment_manager.items.reject(&:leather)
     }
   end
 end
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -12,22 +12,12 @@ class CrossingRepair
   def initialize
     setup
 
-    @repair_info_metal.each do |repairer, items|
+    @repair_info.each do |repairer, items|
       echo "Repairing at #{repairer}" if UserVars.crossing_repair_debug
       repair_at(repairer['name'], repairer['id'], items)
     end
 
-    @repair_info_leather.each do |repairer, items|
-      echo "Repairing at #{repairer}" if UserVars.crossing_repair_debug
-      repair_at(repairer['name'], repairer['id'], items)
-    end
-
-    @repair_info_leather.to_a.reverse.each do |repairer, _items|
-      while tickets_left?(repairer)
-      end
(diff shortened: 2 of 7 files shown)


hunting-buddy.lic
```diff
@@ -76,6 +78,16 @@ class HuntingBuddy
     bput("stow my #{description}", 'You put')
   end
 
+  def check_prebuffs
+    return unless @settings@settings.pre_hunt_buff
```
rpherbig: You have `@settings` twice here.

hunting-buddy.lic
```diff
@@ -76,6 +78,16 @@ class HuntingBuddy
     bput("stow my #{description}", 'You put')
   end
 
+  def check_prebuffs
+    return unless @settings@settings.pre_hunt_buff 
+    return unless !@settings.waggle_sets.nil?
```
rpherbig: This could be simplified to `return if @settings.waggle_sets.nil?` or even `return unless @settings.waggle_sets`.

profiles/base.yaml
```diff
@@ -114,7 +114,8 @@ empty_hunting_room_messages:
 - Anyone here?
 hunting_buddies:
 stop_hunting_if_bleeding: true
-
+pre_hunt_buff: false
```
rpherbig: I don't think we need a boolean here. The presence of the waggle set should be enough. What do you think?

validate.lic
```diff
@@ -331,6 +331,13 @@ class DRYamlValidator
     warn('The analyze_combo_type setting is deprecated and can be removed from your YAML. The use_barb_combos setting should be used instead.')
   end
 
+  def assert_that_prebuff_has_waggle_set(settings)
+    return unless settings.pre_hunt_buff
+    return unless settings.waggle_sets.nil?
```
rpherbig: Should this also check that `settings.waggle_sets['prehunt']` exists?

crossing-repair.lic
```diff
@@ -12,12 +12,22 @@ class CrossingRepair
   def initialize
     setup
 
-    @repair_info.each do |repairer, items|
```
rpherbig: It looks like your other change got included here. Can you remove the `crossing-repair` changes from this PR?

hunting-buddy.lic
```diff
@@ -76,6 +78,16 @@ class HuntingBuddy
...
+
+    if @settings.pre_buff_room
```
rpherbig: The `if` check isn't necessary, `walk_to` will handle it.

hunting-buddy.lic
```diff
@@ -76,6 +78,16 @@ class HuntingBuddy
...
+      walk_to(@settings.pre_buff_room)
+    end
+    wait_for_script_to_complete('buff', ['prehunt'])
```
rpherbig: Do you have a character using this waggle set? If so, would you be up for including their YAML in this PR so there is an example to work from?

hunting-buddy.lic
jayRyan24: When I did this before using nil, it went through.

profiles/base.yaml
jayRyan24: I think it needs to be left here. Sometimes you have a waggle set, but you no longer suffer from mandatory pre-buff because the critter can't kill you before you are capable of buffing. This is a personal preference thing, prebuffing always could be good too.

hunting-buddy.lic
jayRyan24: Yeah, Ugsy is, sure.

profiles/base.yaml
jayRyan24: If I change and include the prebuff waggle set in the check, this doesn't need the boolean.

crossing-repair.lic
jayRyan24: UGH, I didn't commit it to this branch!

hunting-buddy.lic
jayRyan24: This removes the need for the boolean, if you have a prebuff set, it will auto buff you, where you are standing or at the room specified if there is one. 

`   return unless @settings.waggle_sets['prebuff']`

validate.lic
jayRyan24: Actually, if we don't use the boolean to determine if we're going to prebuff, then what is there to validate? If we set a prehunt waggle set, that means we're going to be using it to buff before a hunt. What do we check here?

hunting-buddy.lic
jayRyan24: Fixed, this is great!

validate.lic
rpherbig: I'm not sure if there is anything to validate in that case.

rpherbig:
Note to self: this may fix #167

hunting-buddy.lic
```diff
@@ -78,13 +78,11 @@ class HuntingBuddy
     bput("stow my #{description}", 'You put')
   end
 
-  def check_prebuffs
-    return unless @settings@settings.pre_hunt_buff 
-    return unless !@settings.waggle_sets.nil?
+  def check_prehunt_buffs
```
rpherbig: I like `prehunt_buffs` over `prebuffs`. What do you think about changing the waggle set name and room setting name to match?

hunting-buddy.lic
```diff
@@ -76,6 +78,14 @@ class HuntingBuddy
     bput("stow my #{description}", 'You put')
   end
 
+  def check_prehunt_buffs
+    return unless @settings.waggle_sets['prehunt_buffs']
```
rpherbig: We could even get rid of this check if we modify `buff.lic` to handle an empty spell set. What do you think?

hunting-buddy.lic
```diff
@@ -76,6 +78,14 @@ class HuntingBuddy
...
+
+    walk_to(@settings.pre_buff_room)
```
rpherbig: This setting wasn't renamed, and doesn't match the change to `base.yaml`.

hunting-buddy.lic
jayRyan24: I did this. One issue is that since buffs will use default if empty arguments, it doesn't natively support the prebuff_buffs without some type of boolean check. Otherwise, it will always prebuff regardless of whether or not you set it up to do so. I added logic to take care of this, let me know if this is easier done in ruby, or cleaner, than I have it now.

buff.lic
```diff
@@ -21,6 +21,10 @@ class Waggle
 
     waggle_spells = @settings.waggle_sets
 
+    if args.spells['prehunt_buffs']
+      return unless waggle_spells[args.spells]
+    end
+
```
jayRyan24: Due to us not using a flag for buffing or not buffing, and buffing if the prehunt_buffs exists, I added the logic you see here. Let me know suggestions if ruby handles this better or if my logic could just be all around better.

rpherbig: I'd rather not have a specific check here, `buff.lic` shouldn't be concerned with which specific set it was told to do.

I don't think it defaults the way you're suggesting. `Hunting-buddy` will be passing in an arg all the time, so `args.spells` will be whatever you pass in. Then `waggle_spells[args.spells]` will resolve, I assume to `nil` (that might be worth double checking). That is what we would want to check for, perhaps `return unless spells` in `check_spells(...)`?

jayRyan24: Yes, this didn't seem very good to me either. This makes sense though, checking for the spells in waggle_spells. That's better and makes way more sense.

Issue #1180, opened 2016-12-20
(Crossing-Training) - Tithing with Debt


jayRyan24:
Possibly, CT should pay debts before/after tithing, or at least handle the messaging. 

```
[crossing-training: *** No match was found after 15 seconds, dumping info]  
[crossing-training: messages seen length: 4]   
[crossing-training: message: You hear a soft voice say, "Thou should first attend to thy own woes before seeking to help others."]                                                                                                
[crossing-training: message: Bless  (16 roisaen)]                                                                  
[crossing-training: message: Minor Physical Protection  (14 roisaen)]                                              
[crossing-training: message: Centering  (15 roisaen)]                                                              
[crossing-training: checked against [/You drop/i, /But you do not/i]] 
[crossing-training: checked against [/You drop/i, /But you do not/i]]
```

PR #1231, opened 2017-01-01
Adding the payment of debt to tithing

jayRyan24:
Also I added another bput case, just in case there is a lack of funding and the debt still exists. Causes you to hit the 15s time which is annoying.


jayRyan24:
This resolves: #1180

Issue #1047, opened 2016-11-20
Afk should utilize SlackBot if it is configured


rpherbig:
E.g. when logging you out due to low health/spirit.

jandersson:
Working on this now, this means we gotta trust afk. Does that still seem acceptable?

rcuhljr:
If you delay for a little while, I'm going to roll out a new version of the lich fork that removes the trust model entirely, so that won't be an issue.

rpherbig:
How long is 'a little while'?

rcuhljr:
I need to write up `go3` and do some more proof of concept testing. Hopefully by this weekend.

jandersson:
Awesome! I'll hold off then

Sent with AquaMail for Android
http://www.aqua-mail.com

PR #1240, opened 2017-01-03
Output AFK messages to SlackBot

jandersson:
Covers #1047


jandersson:
Should be noted that this requires trusting afk if using slackbot

jandersson:
Closed pending trust issues

PR #1243, opened 2017-01-04
Increase safety around lodged items

kamarcum:
I'd like to run this for a bit before merging since it's in such a critical path, but wanted to share the code for feedback.

The `fput`s for dropping removed lodged items would fail in combat if additional roundtime was incurred, leaving the character with full hands.

Additionally, the exit call in `SafetyProcess` would never be reached if there were trouble in `fix_standing` or `tend_lodged`.  Let's pull that up. Having a call like that in a predicate was weird anyway.

### DIFF
--- a/common-healing.lic
+++ b/common-healing.lic
@@ -203,16 +203,14 @@ module DRCH
   def bind_wound(part, person = 'my')
     snap = [DRC.left_hand, DRC.right_hand]
     if 'You skillfully remove' == DRC.bput("tend #{person} #{part}", 'You work carefully at tending', 'That area has already been tended to', 'That area is not bleeding', 'You fumble', 'You \w+ remove', 'You (carelessly|foolishly)', 'You deftly remove')
-      waitrt?
-      fput("drop my #{DRC.left_hand}") if DRC.left_hand != snap.first
-      fput("drop my #{DRC.right_hand}") if DRC.right_hand != snap.last
+      DRC.bput("drop my #{DRC.left_hand}", 'You drop', 'What were you') if DRC.left_hand != snap.first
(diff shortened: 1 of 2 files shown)


common-healing.lic
```diff
@@ -203,16 +203,14 @@ module DRCH
   def bind_wound(part, person = 'my')
     snap = [DRC.left_hand, DRC.right_hand]
     if 'You skillfully remove' == DRC.bput("tend #{person} #{part}", 'You work carefully at tending', 'That area has already been tended to', 'That area is not bleeding', 'You fumble', 'You \w+ remove', 'You (carelessly|foolishly)', 'You deftly remove')
-      waitrt?
```
rpherbig: Why remove the `waitrt?` ?

common-healing.lic
```diff
@@ -203,16 +203,14 @@ module DRCH
...
-      fput("drop my #{DRC.right_hand}") if DRC.right_hand != snap.last
+      DRC.bput("drop my #{DRC.left_hand}", 'You drop', 'What were you') if DRC.left_hand != snap.first
+      DRC.bput("drop my #{DRC.right_hand}", 'You drop', 'What were you') if DRC.right_hand != snap.last
       bind_wound(part, person)
     end
     waitrt?
   end
 
   def unwrap_wound(part)
-    waitrt?
```
rpherbig: Why remove the `waitrt?` ?

common-healing.lic
kamarcum: Dealing with RT is one of the best things about bput. I don't see any reason to clutter this up with the extra call.

rpherbig:
Looks good to me 👍

PR #1255, opened 2017-01-06
Add custom prep for SLS

jandersson:
SLS is targeted but cannot be prepared with "tar"

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -272,6 +272,7 @@ spell_data:
     ability: Targeted Magic
     abbrev: SLS
     cyclic: true
+    prep: pre
   Steps of Vuan:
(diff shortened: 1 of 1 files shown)


rpherbig:
Does `combat-trainer` consume that attribute?

My understanding was that in the character YAML you would add the `prep_type` attribute to the spell definition.

jandersson:
Yea, i was looking through `prepare_spell` in combat_trainer and saw the following lines (at 970):

```
    command = 'pre'
    command = data['skill'] != 'Targeted Magic' ? 'pre' : 'tar' if data['skill']
    command = @spell_data[data['name']]['prep'] if @spell_data[data['name']]['prep']
```

PR #1262, opened 2017-01-07
Adding support for Scouting and F2P error messages.

jayRyan24:
Adding support for this:

```
--- Lich: go2 has exited.                                                                                      
[scouting]>scout trail                                                                                         
You have difficulty navigating that trail.                                                                     
[Because your account is free, you don't have access to the province this trail leads to.  To purchase access, 
  visit https://store.play.net/store/purchase/dr .]                                                            
You are unable to locate a trail in the area.                                                                  
>         
```

### DIFF
--- a/scouting.lic
+++ b/scouting.lic
@@ -49,14 +49,10 @@ class Scouting
 
   def take_trail(room_id)
     walk_to(room_id)
-    case bput('scout trail', 'You notice a trailmarker', 'Because your account is free')
-    when 'You notice a trailmarker'
-      return if 'I could not find what you were referring to' == bput('go trail', 'You set off into the wild', 'That would be difficult', 'I could not find what you were referring to')
-      waitfor('As your journey ends')
-      pause 0.5 until Room.current.id
-    when 'Because your account is free'
-      return
-    end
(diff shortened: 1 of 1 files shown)


jayRyan24:
Fek, one sec.

scouting.lic
```diff
@@ -49,10 +49,14 @@ class Scouting
...
-    bput('scout trail', 'You notice a trailmarker')
-    return if 'I could not find what you were referring to' == bput('go trail', 'You set off into the wild', 'That would be difficult', 'I could not find what you were referring to')
-    waitfor('As your journey ends')
-    pause 0.5 until Room.current.id
+    case bput('scout trail', 'You notice a trailmarker', 'Because your account is free')
+    when 'You notice a trailmarker'
+      return if 'I could not find what you were referring to' == bput('go trail', 'You set off into the wild', 'That would be difficult', 'I could not find what you were referring to')
+      waitfor('As your journey ends')
+      pause 0.5 until Room.current.id
+    when 'Because your account is free'
```
rpherbig: I don't think you need this `when` case since it's already the end of the method.

I suggest something like `return if 'Because your... == bput('scout...')'` and then something similar to the old code. What do you think?

jayRyan24: Good call, pushed it.

Issue #1245, opened 2017-01-04
Undergondola Pit Failure


jayRyan24:
Have an issue with it here. If you can't hide on the creatures, we need to handle it by stopping the training. Also, we need to handle this error message.

```
0500:The yvhh la'tami notices your attempt to hide!
2017-01-04 15:00:05 -0500:The yvhh la'tami discovers you, ruining your hiding place!
2017-01-04 15:00:05 -0500:Roundtime: 5 sec.
2017-01-04 15:00:10 -0500:[crossing-training]>hide
2017-01-04 15:00:10 -0500:* As if effort and skill were a bad thing, an yvhh la'tami springs upwards, its thin sharp teeth bared at you.  You knock aside nearly all the teeth with a blackened steel parry stick with sturdy steelsilk straps.  
2017-01-04 15:00:10 -0500:[You're solidly balanced with no advantage.]
2017-01-04 15:00:11 -0500:The yvhh la'tami notices your attempt to hide!
2017-01-04 15:00:11 -0500:The yvhh la'tami discovers you, ruining your hiding place!
2017-01-04 15:00:11 -0500:Roundtime: 5 sec.
2017-01-04 15:00:15 -0500:[crossing-training]>hide
2017-01-04 15:00:16 -0500:* Looking as if this were a bad idea, an yvhh la'tami springs upwards, its thin sharp teeth bared at you.  You dodge.  
2017-01-04 15:00:16 -0500:[You're solidly balanced with no advantage.]
2017-01-04 15:00:16 -0500:You melt into the background, convinced that your attempt to hide went unobserved.
2017-01-04 15:00:16 -0500:Roundtime: 5 sec.
2017-01-04 15:00:17 -0500: * Perune Nocarrain joins the adventure.
2017-01-04 15:00:21 -0500: * Ribbet returns home from a hard day of adventuring.
2017-01-04 15:00:21 -0500:[crossing-training]>climb practice pit
2017-01-04 15:00:21 -0500:You are engaged to an yvhh la'tami at melee range!
2017-01-04 15:00:22 -0500:The yvhh la'tami searches around for a moment.
2017-01-04 15:00:22 -0500:An yvhh la'tami has spotted you!
2017-01-04 15:00:22 -0500:The yvhh la'tami discovers you, ruining your hiding place!
2017-01-04 15:00:22 -0500: * Yanbelev returns home after a long day of fun adventures.
[...]

PR #1263, opened 2017-01-07
Adding support for a hiding limit to practice climb

jayRyan24:
Adding support for a max hiding count. This takes care of #1245.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -674,6 +674,7 @@ class CrossingTraining
       attempts_to_hide = 5
       while attempts_to_hide > 0 do
         break if hide?
+        attempts_to_hide - 1
       end
       return if attempts_to_hide == 0
       bput("climb practice #{target}", 'You begin to practice ')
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -671,9 +671,11 @@ class CrossingTraining
       check_osrel unless @researching
       check_research
       retreat
-      loop do
+      attempts_to_hide = 5
+      while attempts_to_hide > 0 do
```
rpherbig: Do you need to decrement `attempts_to_hide` in the loop?

jayRyan24: YES OF COURSE I DO... :(

jayRyan24: Sorry, my hate. Thanks for the catch.

crossing-training.lic
```diff
@@ -671,9 +671,12 @@ class CrossingTraining
...
         break if hide?
+        attempts_to_hide - 1
```
rpherbig: This doesn't assign back to `attempts_to_hide`, so I think this will still be an infinite loop... can you double check?

jayRyan24: Ah, was told it did, lol. I will check!

PR #1270, opened 2017-01-09
Need cash on hand after debt repayment.

jayRyan24:
Let me know if this fix is OK, or if we need to get out the debt amount, and then get out another 300 copper on a separate call.

### DIFF
--- a/jail-buddy.lic
+++ b/jail-buddy.lic
@@ -71,6 +71,8 @@ def get_out_of_jail
       break
     end
   end
+  DRC.wait_for_script_to_complete('sell-loot')
+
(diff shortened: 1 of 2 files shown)


jayRyan24:
I mean, instead of increasing there by 300, we just do. 

```
   unless ensure_copper_on_hand(debt_amount, hometown)
      echo '***STATUS*** Insufficient funds to pay your debt'
      beep
      exit
    end

   unless ensure_copper_on_hand(300, hometown)
      echo '***STATUS*** Insufficient funds for cash on hand!'
      beep
      exit
    end
```

rpherbig:
Another option would be to run `sell-loot` after getting out of jail. What do you think about that?

PR #1273, opened 2017-01-09
Use training_room sample preferred for hunting_buddy conclusion.

jayRyan24:
Will use training_room sample if one exists, or it uses safe_room. Return room after hunting is over.

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -60,7 +60,7 @@ class HuntingBuddy
       wait_for_script_to_complete('bescort', @exit) if @exit
       release_cyclics
     end
-    walk_to(@settings.safe_room)
+    walk_to((@settings.training_rooms || [@settings.safe_room]).sample)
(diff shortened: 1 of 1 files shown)


hunting-buddy.lic
```diff
@@ -60,7 +60,7 @@ class HuntingBuddy
       wait_for_script_to_complete('bescort', @exit) if @exit
       release_cyclics
     end
-    walk_to(@settings.safe_room)
+    walk_to((@settings.training_rooms || [@settings.safe_room]).sample)
```
rpherbig: This will pick a random room, but then crossing-training will pick another (likely different) random room. Do you think that is a problem?

rcuhljr: Yeah, make this just go straight to safe_room and I'll merge it.

jayRyan24: That's what it does now?

PR #1277, opened 2017-01-10
Melee Attack Overrides

b-Y-d:
example of yaml format:
```
attack_overrides:
  Polearms: feint
  Small Blunt: feint
  Large Edged: jab
  Twohanded Edged: jab

### CODE REVIEW

File: combat-trainer.lic [hunk 1]
```diff
@@ -1874,6 +1877,10 @@ class GameState
     @current_weapon_skill = weapon_skill
   end
 
+	def attack_override
+		@attack_overrides[weapon_skill]
+	end
+
```
rcuhljr: Tabbbbbbsss

PR #1282, opened 2017-01-11
Crossing-training release_cyclics tweak

drboldwin:
makes it conditional against rite of grace and regenerate

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -51,7 +51,7 @@ class CrossingTraining
 
     UserVars.crossing_timers ||= {}
 
-    release_cyclics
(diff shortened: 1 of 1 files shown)


jandersson:
Regarding the discussion we had about RoG in hunting-buddy: What do you all think about a 'post-hunting' waggle_set? This would recast or cast RoG after hunting or any other buffs desired while ;sell-loot and ;crossing-repair are running?

rcuhljr:
Could we have these cyclics it leaves up be listed in base_yaml? Do that and it's good to go.

rpherbig:
@drboldwin @jandersson Does the 'after' functionality take care of this adequately?

jandersson:
yes

PR #1286, opened 2017-01-12
Updated safe-room to not go for healing unless you're damaged past a certain threshhold

adamgoerdt:
Hopefully this will help with the lines at Dokt


safe-room.lic
```diff
@@ -47,6 +57,14 @@ class SafeRoom
     give_and_take(settings.safe_room_id, settings.safe_room_give, settings.safe_room_take)
   end
 
+  def get_health_score(health_hash)
```
rpherbig: You could replace this method with `health_hash.map { |k, v| k * v.count }.reduce(:+)`

profiles/base.yaml
```diff
@@ -382,6 +382,7 @@ storage_containers:
 gear:
 gear_sets:
 empath_healing:
+heal_threshhold: 25
```
rpherbig: What do you think about `safe_room_health_threshold`? I'm looking for something to indicate how or why it is used.

safe-room.lic
```diff
@@ -18,13 +18,21 @@ class SafeRoom
...
 
-    unless args.force || check_health.any?
+    unless args.force || health_hash.any?
       echo '***STATUS*** No wounds, skipping healing'
       return
     end
+    
 
-    settings = get_settings
+    if get_health_score(health_hash) < settings.heal_threshhold
```
rpherbig: There are only 2 'h's in 'threshold'.

safe-room.lic
kamarcum: Or `health_hash.reduce(0) { |sum, (key, array)| sum + key * array.size }`, if I'm reading this right. @rpherbig's version is prettier, but my version only iterates once. Pick your poison.

profiles/base.yaml
adamgoerdt: Makex sense. Fixed

safe-room.lic
adamgoerdt: I normally avoid one-liners since they reduce readability, but can get it switched out if that's how you'd like to see it

rpherbig: I think the method name makes the `score` variable unnecessary... Using something better than `k` and `v` should make it readable enough?

adamgoerdt: Updated

adamgoerdt: I realized overnight that this could delay heading even when bleeding so I updated the check method to force healing if any of the injuries are bleeders

safe-room.lic
```diff
@@ -47,6 +54,23 @@ class SafeRoom
     give_and_take(settings.safe_room_id, settings.safe_room_give, settings.safe_room_take)
   end
 
+  def need_healing(health_hash)
+    settings = get_settings
+    score = 0
+    health_hash.each do |key, array|
+      if key >= 4
```
rpherbig: Instead of hardcoding 4, which could change, you could use the `bleeding?` and `checkpoison` helper methods.

adamgoerdt: Updated.  How's that look?

safe-room.lic
```diff
@@ -47,6 +54,20 @@ class SafeRoom
...
 
+  def need_healing(health_hash)
+    if bleeding? || checkpoison
+      return true
+    end
+    
+    settings = get_settings
+    score = health_hash.map { |k, v| k * v.count }.reduce(:+)
+    
+    if score > settings.saferoom_health_threshold
```
rpherbig: Lines 65-68 could be replaced with `score > settings.saferoom_health_threshold`

adamgoerdt: I think I did what you were requesting

safe-room.lic
```diff
@@ -18,14 +18,21 @@ class SafeRoom
...
 
-    unless args.force || check_health.any?
+    unless args.force || health_hash.any?
       echo '***STATUS*** No wounds, skipping healing'
       return
     end
 
-    settings = get_settings
-
+    if not need_healing(health_hash)
```
rcuhljr: you can use `unless need_healing(health_hash)` here

profiles/base.yaml
```diff
@@ -382,6 +382,7 @@ storage_containers:
 gear:
 gear_sets:
 empath_healing:
+saferoom_health_threshold: 25
```
rcuhljr: 25 Seems way too high for this, you can have multiple bleeders and still be under 20.

adamgoerdt: The method to get 25 wasn't very scientific.  I looked at myself and said 'This is about where I would go to get healing' and checked my score as 26, but you're right in that the way the scoring works it doesn't weigh bad injuries very well.  What do you think about making the scoring method use an exponential rather than a linear scale to give more weight to the worse injuries?  Something like

score = health_hash.map { |k, v| (k ** 2) * v.count }.reduce(:+)

rcuhljr: Wouldn't be bad, I'd like if there was an override where a bleeder always goes for a heal. The two situations that worry me are nerve damage spirals not being stopped soon enough and not healing bleeders.

adamgoerdt:
Bleeding and poison are already set to go for healing

rcuhljr:
Ok, I'd leave this at default 0 in base.yaml for now then. Let people set it if they want.

adamgoerdt:
What about something like 10 as the default? I think most people won't
change from default, and that most don't want to run to the healer for a
singe scratch, either.

PR #1288, opened 2017-01-12
Update base-hunting.yaml

cpryor6:
adding red gremlins at Shard / Derbycrop farm


jandersson:
Is this a premium area?

PR #1299, opened 2017-01-16
Support Heavy TM/Debil Expiration

jandersson:
Heavy spells like mind shout have an expiration message and timer associated with them. They can't be recast until this timer expires. In the current implementation, combat-trainer resets the expiration timers for offensive spells on mob death. This adds a setting that prevents that reset. There is an example in Crannach-setup


rpherbig:
If this is true for *all* heavy spells, I'd rather see this attribute added to the spell definitions in `base-spells.yaml`. What do you think?

jandersson:
I like that idea. Should the `expire` attribute also go in there? I've only got one heavy type spell to test with, but the message seems very generic.

rpherbig:
I'm a fan of having as much constant stuff like messages in base, and only
changing things like mana in the YAML.

Issue #1287, opened 2017-01-12
training-manager: make priority skills lower limit configurable


skyleronken:
Currently the mindstate value of the priority skills is statically set to 6. I suggest making this configurable.

PR #1301, opened 2017-01-16
Custimizable mind state for priority skills

skyleronken:
Can define the mind state at which training-manager will now trigger the
hunting cycle. Issue #1287.

### DIFF
--- a/training-manager.lic
+++ b/training-manager.lic
@@ -48,8 +48,7 @@ class TrainingManager
   end
 
   def priority_skills_low?
-    lower_limit = (@settings.priority_skills_lower_limit || 6)
-    @settings.training_manager_priority_skills.any? { |skill| DRSkill.getxp(skill) <= lower_limit }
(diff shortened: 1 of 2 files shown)


skyleronken:
Example yaml:

priority_skills_lower_limit: 4

training-manager.lic
```diff
@@ -48,7 +48,8 @@ class TrainingManager
   end
 
   def priority_skills_low?
-    @settings.training_manager_priority_skills.any? { |skill| DRSkill.getxp(skill) <= 6 }
+    lower_limit = (@settings.priority_skills_lower_limit || 6)
```
rpherbig: This default value should be in `base.yaml` instead of ||ed here.

skyleronken:
Good call. Those changes should be better.

training-manager.lic
skyleronken: Good call! Change mades!

PR #1302, opened 2017-01-16
Non-Barb Tactics Combos

b-Y-d:
Yaml setting:
`use_tactics_combos: false`
off by default (don't want to confuse people)

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1813,6 +1813,9 @@ class GameState
     @use_barb_combos = settings.use_barb_combos
     echo("  @use_barb_combos: #{@use_barb_combos}") if $debug_mode_ct
     
+    @use_tactics_combos = settings.use_tactics_combos
+    echo("  @use_tactics_combos: #{@use_tactics_combos}") if $debug_mode_ct
+    
     @attack_overrides = settings.attack_overrides
     echo("  @attack_overrides: #{@attack_overrides}") if $debug_mode_ct
     
@@ -1979,6 +1982,11 @@ class GameState
     weapon_skill == 'Brawling'
   end
 
+  def brawl_tactics?
+    return true if !brawling?
+    ((weapon_skill == 'Brawling') && (weapon_name == ''))
+  end
+  
   def use_stealth_attack?
     @use_stealth_attacks && use_stealth?
   end
@@ -2113,7 +2121,9 @@ class GameState
   end
 
   def melee_attack_verb
-    if attack_override
+    if @use_tactics_combos && !offhand? && brawl_tactics? && !backstab? && tactics_analyze?
+      @tactics_combo_array.shift
+    elsif attack_override
       attack_override
     elsif @use_weak_attacks
       brawling? ? 'gouge' : 'jab'
@@ -2181,6 +2191,33 @@ class GameState
     { 'accuracy' => 50, 'damage' => 125, 'flame' => 0 }.each do |type, expertise|
       break if perform_analyze?(type, expertise)
     end
+  end
+
+  def tactics_analyze?
+    return false if DRSkill.getxp('Tactics') > 32
+    result = bput("analyze", 
+      'You fail to find any holes in' ,
+      'You must be closer to use tactical abilities', 
+      'Analyze what\?', 
+      'by landing a .*')
+    waitrt?
+
+    case result
+    when 'You fail to find any holes in'
+      tactics_analyze?
+    when 'You must be closer to use tactical abilities'
+      fput'enga'
+      pause 1
+      tactics_analyze?
+    when 'Analyze what?'
+      fput('face next')
+      tactics_analyze?
+    end
+
+    text = result.match(/by landing a (.*)\.$/).to_a[1]
+    @tactics_combo_array = DRC.list_to_nouns(text)
+
+    true
   end
 
   def dance_stealth?
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -62,6 +62,7 @@ dance_actions:
 - bob
 - weave
 use_tactics_combos: false
+use_barb_combos: false
 manipulate_threshold:
 training_abilities:
   Hunt: 80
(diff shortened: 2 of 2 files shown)


combat-trainer.lic
```diff
@@ -1979,6 +1982,11 @@ class GameState
     weapon_skill == 'Brawling'
   end
 
+  def brawl_tactics?
+    return true if !brawling?
+    ((weapon_skill == 'Brawling') && (weapon_name == ''))
```
rpherbig: Do you need to check `weapon_skill` here if you checked `brawling?` on the line above?

combat-trainer.lic
```diff
@@ -2183,6 +2193,33 @@ class GameState
     end
   end
 
+  def tactics_analyze?
```
rpherbig: There seems to be a lot of overlap with the `perform_analyze?` method. Aren't they basically doing the same thing?

profiles/base.yaml
```diff
@@ -61,6 +61,7 @@ dance_actions:
 - circle
 - bob
 - weave
+use_tactics_combos: false
```
rpherbig: I think this setting should be next to `use_barb_combos`. I don't know which one should move to be near the other. What are your thoughts?

combat-trainer.lic
b-Y-d: Sorry haven't checked the comments here. 
I often brawl while holding a SE for the small boost in parry ability from the SE's balance. This just checks to make sure the tactics combo section doesn't end up training your held weapon instead of brawling.

profiles/base.yaml
b-Y-d: Either or, I didn't see the setting while editing it directly on github. I did see it later though!

combat-trainer.lic
b-Y-d: The overlap is due to me basing the method off `perform_analyze?`, but they're sill fairly different in function.

profiles/base.yaml
b-Y-d: Moved them closer

combat-trainer.lic
rpherbig: The more I look at them, the more similar they appear to me (aside from the `return` cases at the top of the methods).

Why do they handle the `Analyze what?` case differently? They're both performing an `analyze`, just with or without an argument. Why does `perform_analyze?` not handle the `You must be closer...` case?

b-Y-d: I didn't write the `perform_analyze?` method nor do I have a barb-- so I cannot guess why it lacks that case.

The differences are between the barb version and non version are small. True. However, those differences can cause a lot of trouble if you try to make it universally applicable/usable to non-barbs:

- It expects two arguments to be passed through it.
- Has a return based on IF-related crap.
- Has a return based on expertise crap.

Additional reasons were:
- I don't have a barb to test edits to that method to ensure I didn't break it for barbarians.
- I initially expected to use cases for every magnitude of _<size> opening_ gained by analyze.
- I just wanted it to lock tactics quickly and it does!

rpherbig:
Is the goal of this only to improve tactics training? Or is completing the combo the goal of this PR? See #1336.

rpherbig:
@b-Y-d It looks like this has a merge conflict.

b-Y-d:
The goal is to improve tactics training by completing combos with your current weapon until tactics is nearly locked. It locks tactics much faster than simply spamming analyze.

combat-trainer.lic
rpherbig: If the weapon skill is anything *but* Brawling, line 2026 will return. Therefore we know the first clause of line 2027 will always be true.

combat-trainer.lic
```diff
@@ -2153,7 +2161,9 @@ class GameState
   end
 
   def melee_attack_verb
-    if attack_override
+    if @use_tactics_combos && !offhand? && brawl_tactics? && !backstab? && tactics_analyze?
+      @tactics_combo_array.shift
+    elsif attack_override
       attack_override
```
rpherbig: It feels to me like if the user has an override specified, it should take higher priority than completing a combo. What do you think?

rpherbig: I would also expect barb combos to be a higher priority than general combos. What do you think?

combat-trainer.lic
```diff
@@ -2223,6 +2233,33 @@ class GameState
     end
   end
 
+  def tactics_analyze?
+    return false if DRSkill.getxp('Tactics') > 32
```
rpherbig: Reading the code, it looks like this method will always analyze, even if `@tactics_combo_array` is already populated. Is that the case?

b-Y-d: Yes, to correctly get the next attack if your recent attack misses.

rpherbig: That makes sense... how did the barb analyze logic handle this case?

Lazasar:
hey. Any chance of this getting finished?

combat-trainer.lic
rcuhljr: I believe barbs no longer need to hit to complete combos, just make the attack.

combat-trainer.lic
```diff
@@ -2223,6 +2233,33 @@ class GameState
...
+    result = bput("analyze", 
+      'You fail to find any holes in' ,
+      'You must be closer to use tactical abilities', 
+      'Analyze what\?', 
+      'by landing a .*')
+    waitrt?
+
+    case result
+    when 'You fail to find any holes in'
+      tactics_analyze?
```
rcuhljr: this should probably be a return.

combat-trainer.lic
```diff
@@ -2153,7 +2161,9 @@ class GameState
   end
 
   def melee_attack_verb
-    if attack_override
+    if @use_tactics_combos && !offhand? && brawl_tactics? && !backstab? && tactics_analyze?
+      @tactics_combo_array.shift
```
rcuhljr: Since we're never using the fact that this is an array something like `@next_combo_attack` would make more sense.

PR #1303, opened 2017-01-16
Seperating Empathy into it's own script, adding empathy_rooms to base-town.yaml

arkrail:


arkrail:
I just tested a second time, and didn't account for some of the room numbers being shops in Shard, which close in the evening.  Not entirely sure how to address this.

jandersson:
`walk_to` returns false if it detects that the shop is closed. So one possible solution is: for each destination you can do something like `return unless walk_to ####` or `continue unless walk_to ####` which could cause it to just skip that spot.

empathy.lic
```diff
@@ -0,0 +1,51 @@
...
+  def initialize
+    settings = get_settings
+    @stationary_skills_only = settings.crossing_training_stationary_skills_only
+    @hometown = settings.hometown
+
+    train_empathy
+  end
+  
+  def check_research
+    return unless @researching
```
rpherbig: I don't think this extraction is as straightforward as it appears. `@researching` is local to this script (or rather, class) and never gets set. It will be out of sync with `crossing-training`.

I think we ran into the same problem with another extraction, I want to say... athletics?

arkrail: Would removing @researching all together work for it, as perc and perc health both will not break research?

rpherbig: The goal of `check_research` is not to avoid breaking research, but rather to avoid letting it expire (i.e. issue a new `research blah` command as needed). Power walking can take a while.

arkrail:
Going to close this out, and rework it a bit before resubmitting it.

PR #1315, opened 2017-01-23
gtk not checked unless necessary

richardhertzog:
Moved gtk check in status-monitor to beneath no window loop so that `status_monitor_no_window: true` could be accessed without gtk.


rpherbig:
Looks good to me, but @kamarcum and @rcuhljr have played with GTK more than I have...

rcuhljr:
Should be fine.

PR #1317, opened 2017-01-24
Fixed errors in outlying towns and added new Rathan hunting areas

adamgoerdt:


data/base-hunting.yaml
```diff
@@ -294,6 +294,20 @@ hunting_zones:
   - 19847
   - 19778
   - 9055
+  # https://elanthipedia.play.net/Bawdy_swain 120-150
+  # Ratha
+  bawdy_swain:
+  #- 4850
```
rpherbig: Is this line intentionally commented out?

adamgoerdt:
That room has an NPC in it that I kept trying to attack.  Not sure what the
best way is to notate something like that

rpherbig:
There is an `ignored_npcs` setting that should prevent you from attacking it. If you think the room is worth using, you might add a comment saying that. Or we can leave the room out of the list. What do you think?

adamgoerdt:
I figured I could add the ignored_npc, but anyone else who would use that
zone would need to do the same. I think it makes more sense just to leave
it out.  It's a small zone, but the population of Ratha is so small I don't
see it having an issue filling up.

PR #1318, opened 2017-01-24
Added support for before and after hunts

adamgoerdt:


hunting-buddy.lic
```diff
@@ -49,16 +49,32 @@ class HuntingBuddy
       end
 
       args = info['args']
+      before_actions = info['before']
+      after_actions = info['after']
       duration = info[:duration]
       stop_on_skills = info['stop_on']
 
+      #Execute scripts to run before the hunt
```
rpherbig: I recommend extracting this into a helper method. That keeps this main method more about flow, abstracting the details out to the helper.

hunting-buddy.lic
```diff
@@ -49,16 +49,32 @@ class HuntingBuddy
...
+      before_actions.each do |before_action|
+        action_parts = before_action.split(" ")
+        script_name = action_parts.shift
+        wait_for_script_to_complete(script_name, action_parts)
+      end
+
       next if all_skills_at_cap?(stop_on_skills)
```
rpherbig: This line (and line 65) are two cases in which after actions would not occur. Is that ok, or will that cause problems?

rpherbig:
Do you have any YAML files which show an example of how this is used that you would be willing to upload?

hunting-buddy.lic
adamgoerdt: It might be me not understanding the flow, but wouldn't release_cyclics also not be called if the stop_on_skills are hit?

adamgoerdt:
Here are a couple examples from my yamls: http://pastebin.com/UfHhhWcv

hunting-buddy.lic
adamgoerdt: Agreed.  Having a hard time thinking of what to name the function, though.

rpherbig: `execute_actions`?

hunting-buddy.lic
rpherbig: You are correct

PR #1321, opened 2017-01-26
Fixed bug which prevented preparing spells while playing an instrument

adamgoerdt:


common-arcana.lic
```diff
@@ -49,6 +49,9 @@ module DRCA
     when 'Your desire to prepare this offensive spell suddenly slips away'
       pause 1
       return prepare?(abbrev, mana, symbiosis, command)
+    when 'You should stop playing before you do that'
+      fput('stop play')
```
rpherbig: What do you think about making this a `bput` instead?

common-arcana.lic
```diff
@@ -49,6 +49,9 @@ module DRCA
...
+      return prepare?(abbrev, mana, symbiosis, command)
```
rpherbig: This could result in preparing symbiosis twice, but it looks like that case is handled. Not elegant, but not a problem (I think).

adamgoerdt:
Would hard-coding the symbiosis param (return prepare?(abbrev, mana, false,
command)) fix that?

adamgoerdt:
Good question.  Looking into the bput method I think adding a case in there
for 'You should stop playing before you do that' might be a better fix,
since it *should* fix this bug as well as some other potential bugs.  I'll
test that option now

common-arcana.lic
adamgoerdt: Fixing bput looks to be a much better solution.  I've updated the PR to be a bput change instead

PR #1325, opened 2017-01-27
minor change to unhide after the end of  climbing practice

### LINKED ISSUE #1323: Climb practice needs to unhide.
Climb practice needs to unhide after it's complete. If it doesn't, it will sneak back up to it's next location.

jayRyan24:
Fixes #1323

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -685,12 +685,13 @@ class CrossingTraining
         break if Flags['ct-climbing-finished']
         return if Flags['ct-climbing-combat']
         if DRSkill.getxp('Athletics') > 28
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -690,6 +690,8 @@ class CrossingTraining
         end
       end
     end
+
+    fput('unhide')
```
rpherbig: What do you think about making this a bput?

`bput('unhide', 'You come out of hiding', 'You slip out of hiding', 'But you are not')`

jayRyan24: Sounds good, pushed it through.

PR #1328, opened 2017-01-28
Combat-Trainer Last Rites

drboldwin:
adding last rites to combat trainer - does it instead of skinning once ever 10 min.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -242,12 +242,20 @@ class LootProcess
     @dump_junk = settings.dump_junk
     echo("  @dump_junk: #{@dump_junk}") if $debug_mode_ct
     @dump_timer = Time.now - 300
+	
+    @last_rites_toggle = false
+    @last_rites = settings.last_rites
+    echo("  @last_rites: #{@last_rites}") if $debug_mode_ct
+    @last_rites_timer = Time.now - 600
   end
 
   def execute(game_state)
     if (Time.now - @dump_timer > 300) && @dump_junk && DRRoom.room_objs.count >= 13
       fput 'DUMP JUNK'
       @dump_timer = Time.now
+    end
+    if (Time.now - @last_rites_timer > 600) && @last_rites
+      @last_rites_toggle = true
     end
     game_state.mob_died = false
     dispose_body(game_state)
@@ -384,6 +392,12 @@ class LootProcess
     return if Time.now - @loot_timer < @loot_delay
 
     game_state.mob_died = true
+    if @last_rites_toggle
+      fput "pray #{DRRoom.dead_npcs.first}"
+      @last_rites_toggle = false
+      @last_rites_timer = Time.now
+      return
+    end
     arrange_mob(DRRoom.dead_npcs.first, game_state)
     check_skinning(DRRoom.dead_npcs.first, game_state) if check_rituals?(game_state)
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -390,6 +390,9 @@ class LootProcess
     game_state.mob_died = true
     if (Time.now - @last_rites_timer > 600) && @last_rites
       bput ("pray #{DRRoom.dead_npcs.first}", 'You beseech your god for mercy', 'You pray fervently', 'You continue praying for guidance', 'Quietly touching your lips with the tips of your fingers', 'murmur a brief prayer for its safe travels beyond this life')
+      waitrt?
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -384,6 +392,12 @@ class LootProcess
     return if Time.now - @loot_timer < @loot_delay
 
     game_state.mob_died = true
+    if @last_rites_toggle
```
rpherbig: Why not use the conditional from line 257 here, so you don't need the separate `@last_rites_toggle` boolean?

drboldwin: yeah - should likely.  I had set it up differently at first (using training abilities) but liked this better.  That's a lil remnant from that.

combat-trainer.lic
```diff
@@ -384,6 +388,11 @@ class LootProcess
     return if Time.now - @loot_timer < @loot_delay
 
     game_state.mob_died = true
+    if (Time.now - @last_rites_timer > 600) && @last_rites
+      fput "pray #{DRRoom.dead_npcs.first}"
```
rpherbig: Do you have the messaging such that this could be a `bput`?

drboldwin: I have only the success message right now

drboldwin: need to grab a message if it's already skinned, if it's cursed, undead (maybe those are the same), or done too soon

rpherbig:
@drboldwin How's this going?

drboldwin:
Sorry - busy week.  Should be good to go

drboldwin:
Oops - wrong button.

combat-trainer.lic
```diff
@@ -384,6 +388,11 @@ class LootProcess
     return if Time.now - @loot_timer < @loot_delay
 
     game_state.mob_died = true
+    if (Time.now - @last_rites_timer > 600) && @last_rites
+      bput "pray #{DRRoom.dead_npcs.first}, 'You beseech your god for mercy', 'You pray fervently', 'You continue praying for guidance', 'Quietly touching your lips with the tips of your fingers', 'murmur a brief prayer for its safe travels beyond this life'"
```
rpherbig: I think you single and double quotes are out of place. I expected to see something like `bput("pray ...", 'message 1', 'message 2')`.

drboldwin: oh geeze - yeah I put them in before the close by mistake

drboldwin:
Gabellia pointed out that since they both can be done in conjunction (last rites and a recite) so did that too.  Though I should probably toss in a qualifier that you have meraud's in there also.  since it won't work without it.

combat-trainer.lic
```diff
@@ -384,6 +388,14 @@ class LootProcess
     return if Time.now - @loot_timer < @loot_delay
 
     game_state.mob_died = true
+    if (Time.now - @last_rites_timer > 600) && @last_rites
```
rpherbig: Since there is some roundtime involved, do you want to retreat if there are still mobs alive in the room?

drboldwin: don't think it needs to retreat (is only 5 second RT) nothing too worrisome

drboldwin:
also you may want to look at how I worked with game_state it kept giving me some issues so I did it a little funky figuring you could tell me a better way to do it if there is one

combat-trainer.lic
```diff
@@ -1725,6 +1726,7 @@ class GameState
     @prepare_consume = false
     @use_charged_maneuvers = false
     @wounds = {}
+    @blessed_room = false
```
rpherbig: You'll need to declare `:blessed_room` on line 1706

combat-trainer.lic
```diff
@@ -1268,6 +1268,7 @@ class TrainerProcess
     pause
     waitrt?
     fix_standing
+    game_state.blessed_room
```
rpherbig: Will `blessed_room` ever be set to false? I.e. it wears off or expires?

drboldwin: So currently - no.  It will get set to true when you do meraud commune - and it will forever be true for the remainder of the hunt.  But the intent here was to have it be ready for another update - (I could bundle them I suppose) that will add a flag for the line which indicates meraud commune has worn off ('you no longer feel connected to your consecrated area') or what-have-you.  I would need to do a lil more testing in that area but when that's done then it could turn the @blessed_room back to false.

It should be noted there are absolutely no negatives to doing last_rites/recite in a non-blessed room or against constructs/undead/cursed creatures.  I went about it this way mainly because I just wanted to avoid repetitive actions that have no 'benefit'.

PR #1330, opened 2017-02-01
Merge Spell Data with Combat-Trainer & Support heavy tm/debil

jandersson:
This will pull data from base-spells and merge it with the users offensive spell definitions. This means that we can start leaving stuff out of our setup files if its included in base-spells. I did an example in Crannach-setup where I define mind shout without mana, abbreviation, expire, or the expire_on_death flag. 

If the expire_on_death flag is set to false, then the flag will not be reset after every mob kill. This is necessary for heavy tm/debil spells which expire over time rather than mob death.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -566,7 +566,11 @@ class SpellProcess
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
     @spell_data = get_data('spells').spell_data
-
+    @offensive_spells.each.with_index do |spell, i| 
+      if @spell_data[spell['name']]
+        @offensive_spells[i] = spell.merge(@spell_data[spell['name']]) 
+      end
+    end
     @necromancer_healing = settings.necromancer_healing
     echo("  @necromancer_healing: #{@necromancer_healing}") if $debug_mode_ct
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -575,8 +575,8 @@ class SpellProcess
     
     @buff_spells
       .select { |spell| @spell_data[spell]}
-      .each do |spell|
-        @buff_spells[spell] = @spell_data[spell].merge(spell)
+      .each do |name, _data|
+        @buff_spells[name] = @spell_data[name].merge(@buff_spells[name])
       end
     echo(" @buff_spells merged with base-spells: #{@buff_spells}") if $debug_mode_ct
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -654,7 +654,7 @@ class SpellProcess
       @offensive_spells
         .select { |spell| spell['expire'] }
         .reject { |spell| spell['cyclic'] }
-        .reject { |spell| !spell['expire_on_death'] }
+        .reject { |spell| spell['heavy'] }
         .each { |spell| Flags["ct-#{spell['abbrev']}"] = true }
     end
     check_slivers(game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -565,8 +565,6 @@ class SpellProcess
     @offensive_spell_cycle = settings.offensive_spell_cycle
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
-    @spell_data = get_data('spells').spell_data
-   
     @necromancer_healing = settings.necromancer_healing
     echo("  @necromancer_healing: #{@necromancer_healing}") if $debug_mode_ct
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -566,11 +566,11 @@ class SpellProcess
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
     @spell_data = get_data('spells').spell_data
-    @offensive_spells.each.with_index do |spell, i| 
-      if @spell_data[spell['name']]
-        @offensive_spells[i] = spell.merge(@spell_data[spell['name']]) 
+    @offensive_spells
+      .select { |spell| @spell_data[spell['name']] }
+      .each_with_index do |spell, i|
+        @offensive_spells[i] = @spell_data[spell['name']].merge(spell) 
       end
-    end
     @necromancer_healing = settings.necromancer_healing
     echo("  @necromancer_healing: #{@necromancer_healing}") if $debug_mode_ct
 
@@ -975,7 +975,6 @@ class SpellProcess
     release_cyclics if data['cyclic']
     command = 'pre'
     command = data['skill'] != 'Targeted Magic' ? 'pre' : 'tar' if data['skill']
-    command = @spell_data[data['name']]['prep'] if @spell_data[data['name']]['prep']
     command = data['prep_type'] if data['prep_type']
 
     @spell_timers[data['abbrev']] = Time.now if data['recast_every']
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -650,6 +650,7 @@ class SpellProcess
       @offensive_spells
         .select { |spell| spell['expire'] }
         .reject { |spell| spell['cyclic'] }
+        .reject { |spell| !spell['expire_on_death'] }
         .each { |spell| Flags["ct-#{spell['abbrev']}"] = true }
     end
     check_slivers(game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -968,7 +968,6 @@ class SpellProcess
     end
     release_cyclics if data['cyclic']
     command = 'pre'
-    command = data['skill'] != 'Targeted Magic' ? 'pre' : 'tar' if data['skill']
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -566,20 +566,7 @@ class SpellProcess
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
     @spell_data = get_data('spells').spell_data
-    @offensive_spells
-      .select { |spell| @spell_data[spell['name']] }
-      .each_with_index do |spell, i|
-        @offensive_spells[i] = @spell_data[spell['name']].merge(spell) 
-      end
-    echo("  @offensive_spells merged with base-spells: #{@offensive_spells}") if $debug_mode_ct
-    
-    @buff_spells
-      .select { |spell| @spell_data[spell]}
-      .each do |name, _data|
-        @buff_spells[name] = @spell_data[name].merge(@buff_spells[name])
-      end
-    echo(" @buff_spells merged with base-spells: #{@buff_spells}") if $debug_mode_ct
-
+   
     @necromancer_healing = settings.necromancer_healing
     echo("  @necromancer_healing: #{@necromancer_healing}") if $debug_mode_ct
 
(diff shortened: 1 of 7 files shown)


combat-trainer.lic
```diff
@@ -566,7 +566,11 @@ class SpellProcess
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
     @spell_data = get_data('spells').spell_data
-
+    @offensive_spells.each.with_index do |spell, i| 
+      if @spell_data[spell['name']]
+        @offensive_spells[i] = spell.merge(@spell_data[spell['name']])
```
jandersson: Note to self: I need to swap the things being merged... If there are duplicate values then they are retained from y in `x.merge(y)` So we should keep the user settings

combat-trainer.lic
```diff
@@ -650,6 +654,7 @@ class SpellProcess
       @offensive_spells
         .select { |spell| spell['expire'] }
         .reject { |spell| spell['cyclic'] }
+        .reject { |spell| !spell['expire_on_death'] }
```
rpherbig: Is this the same as `.select { |spell| spell['expire_on_death'] }`?

data/base-spells.yaml
```diff
@@ -227,6 +227,12 @@ spell_data:
     ability: Targeted Magic
     abbrev: IZ
     cyclic: true
+  Mind Shout:
+    ability: Debilitation
+    abbrev: MS
+    mana: 20
```
rpherbig: I would not put `mana` in `base-spells`. The rest of these attributes are fixed, but mana can change.

combat-trainer.lic
rpherbig: Does this mean line 980 can be removed?

combat-trainer.lic
```diff
@@ -566,7 +566,11 @@ class SpellProcess
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
     @spell_data = get_data('spells').spell_data
-
+    @offensive_spells.each.with_index do |spell, i| 
+      if @spell_data[spell['name']]
```
rpherbig: I think it would be more idiomatic to do something like:
```
@offensive_spells
  .select { |spell| @spell_data[spell['name']] }
  .each.with_index do |spell, i| 
```

combat-trainer.lic
```diff
@@ -566,7 +566,11 @@ class SpellProcess
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
     @spell_data = get_data('spells').spell_data
-
+    @offensive_spells.each.with_index do |spell, i|
```
rpherbig: Shouldn't this be `.each_with_index`?

jandersson: It looks like it can be done in three ways. The current version works too (http://apidock.com/ruby/Enumerator/each_with_index)

combat-trainer.lic
jandersson: Would this require each spell to have `expire_on_death: true` ?

jandersson: I do agree it is a confusing statement

data/base-spells.yaml
jandersson: It is the minimum prep though, if merge works the way it is documented (and how I understand it) then user provided settings would override this

combat-trainer.lic
jandersson: I think maybe the variable can be changed to something more meaningful... `expires_in_time` or maybe `heavy_spell` (though that might be too constraining)

combat-trainer.lic
jandersson: I wouldn't remove it until all spell data is documented, else it might break a users settings for some undocumented spell.

rpherbig: It's pulling from base-spells. Isn't that the same as your code here?

jandersson: Oh my mistake. Yea I think that can get removed. It looks like 980 is a kind of override of the users settings? I'll take it out

jandersson:
I noticed that spell data uses `ability` and offensive_spells uses `skill` to refer to the same thing.

rpherbig:
That's a good observation. I definitely prefer `skill`.

combat-trainer.lic
jandersson: After some testing, it still needs to be there to grab the custom `prep` if there is one. I've added it back

jandersson:
Small issue occurs when changing hunting zones and the heavy spell timer is still active
```
[combat-trainer]>pre MS 20                                                                                  
You are still too fatigued from your previous efforts to manifest another major feat of offensive magic.    
```

jandersson:
Think I should merge `buff_spells` too? Right now its only merging `offensive_spells`

rpherbig:
I think we should merge wherever spells are used

combat-trainer.lic
```diff
@@ -572,7 +572,14 @@ class SpellProcess
         @offensive_spells[i] = @spell_data[spell['name']].merge(spell) 
       end
     echo("  @offensive_spells merged with base-spells: #{@offensive_spells}") if $debug_mode_ct
- 
+    
+    @buff_spells
+      .select { |spell| @spell_data[spell]}
+      .each do |spell|
+        @buff_spells[spell] = @spell_data[spell].merge(spell)
```
jandersson: This isn't working as I expected it to. will fix

rpherbig:
We should consider doing this merge in dependency. It has a transform method for this exact purpose.

jandersson:
Sounds good, I'll take a look at it

jandersson:
Its working! Will let you know if anything is off, but as of now Its working great. I've trimmed down Crannach's profile to be an example. I can add my clerics profile as a second example

combat-trainer.lic
```diff
@@ -970,7 +971,7 @@ class SpellProcess
     release_cyclics if data['cyclic']
     command = 'pre'
     command = data['skill'] != 'Targeted Magic' ? 'pre' : 'tar' if data['skill']
```
rpherbig: This might be another good candidate for the transform in dependency.

dependency.lic
```diff
@@ -770,8 +770,20 @@ class SetupFiles
     boxes = @base_files['base-items.yaml'][:box_nouns]
     gems = @base_files['base-items.yaml'][:gem_nouns]
     scrolls = @base_files['base-items.yaml'][:scroll_nouns]
-    s.lootables = (base_lootables + boxes + gems + scrolls + s.loot_additions - s.loot_subtractions).uniq
+    spells = @base_files['base-spells.yaml'][:spell_data]
 
+    s.lootables = (base_lootables + boxes + gems + scrolls + s.loot_additions - s.loot_subtractions).uniq
+    s.offensive_spells
```
rpherbig: Are there any other spell related settings?

I think a few skills had buff settings (athletics? astrology?). What about the crossing-training spell training setting?

combat-trainer.lic
jandersson: Sorry, can you be more specific? Should 973 pull the prep command from spell_data?

dependency.lic
jandersson: It looks like there are waggle, lockpicking, astrology and possibly thievery and athletics? I'm not familiar with the latter two. I kinda think these things all belong under `waggle_sets` but that may be a bit of work (im up for it though).

jandersson: I can easily test lockpick and astrology. Thievery/athletics I might need to do some more research on

jandersson:
I've merged the spell sets that I know about, let me know of any I missed. Seems to be working good so far!

combat-trainer.lic
rpherbig: This line defaults the prep type based on the defined skill. That could also be done in dependency... that might be a bit weird where Sorcery is concerned.

dependency.lic
rpherbig: I agree that those should all be waggle sets. But maybe not in this PR.

combat-trainer.lic
jandersson: The times I've done sorcery, I've listed it as `- skill: sorcery` and it seemed to work alright. I don't have any sorcery spells to test with, but in theory any user specified attributes should be kept

jandersson: OK, I understand. I'll give it a shot in dependency

combat-trainer.lic
```diff
@@ -566,7 +566,7 @@ class SpellProcess
     echo("  @offensive_spell_cycle: #{@offensive_spell_cycle}") if $debug_mode_ct
 
     @spell_data = get_data('spells').spell_data
```
rpherbig: This line can now be removed.

dependency.lic
```diff
@@ -770,8 +770,34 @@ class SetupFiles
...
+      .each_with_index do |spell, i|
+        s.offensive_spells[i] = spells[spell['name']].merge(spell)
+      end
+    s.buff_spells
+      .select { |spell| spells[spell] }
+      .each do |name, _data|
+        s.buff_spells[name] = spells[name].merge(s.buff_spells[name])
+      end
+    s.lockpick_buffs['spells']
+      .each_with_index
```
rpherbig: What is this line doing?

dependency.lic
```diff
@@ -770,8 +770,34 @@ class SetupFiles
...
+      .select do |spell, i| 
+        spells.select { |_name, data| data['abbrev'].downcase == spell['abbrev'].downcase }
+              .each { |_name, data| s.lockpick_buffs['spells'][i] = data.merge(spell) } 
+      end
+    s.astrology_buffs['spells']
+      .each_with_index
```
rpherbig: What is this line doing?

jandersson: This was a googled solution to 'select with index ruby'. I wanted to retain the index of what select is returning

dependency.lic
```diff
@@ -770,38 +770,43 @@ class SetupFiles
...
       end
+    
     s.offensive_spells
       .each_with_index
-      .select { |spell, _i| (spell['skill'].downcase == 'targeted magic' && !spell['prep']) || (spells[spell['name']]['skill'].downcase == 'targeted magic' && spell['skill'].downcase == 'sorcery')}
-      .each { |data, i| s.offensive_spells[i]['prep'] = 'target' }
-    echo s.offensive_spells  
+      .select { |spell, _i| spell['skill'].downcase == 'targeted magic' }
+      .reject { |spell, _i| spell['prep'] }
+      .select { |spell, _i| spells[spell['name']]['skill'].downcase == 'targeted magic' && spell['skill'].downcase == 'sorcery' }
```
jandersson: This is not working how i expected.. will fix

dependency.lic
```diff
@@ -770,8 +770,46 @@ class SetupFiles
...
-    s.lootables = (base_lootables + boxes + gems + scrolls + s.loot_additions - s.loot_subtractions).uniq
+    spells = @base_files['base-spells.yaml'][:spell_data] 
 
+    s.lootables = (base_lootables + boxes + gems + scrolls + s.loot_additions - s.loot_subtractions).uniq
+    
+    s.offensive_spells
+      .select { |spell| spells[spell['name']] }
+      .each_with_index do |spell, i|
+        s.offensive_spells[i] = spells[spell['name']].merge(spell)
+      end
```
rcuhljr: We could probably replace these with
```
 s.offensive_spells.map!{ |spell| spells[spell['name']] ? spells[spell['name']].merge(spell) : spell  }
```

And I'd still be tempted to pull out that inner ternary to a helper method like `merge_spell_defaults(spells, spell)`

jandersson: This great, thanks

jandersson:
Let me know if there's something on this that is missing. Testing on a few characters now, working great

PR #1334, opened 2017-02-04
Updated steal to work with other hometowns

adamgoerdt:


steal.lic
```diff
@@ -44,9 +44,21 @@ class Steal
     args = parse_args(arg_definitions)
 
     settings = get_settings
+
+    @target_province=''
+
+    if settings.hometown == 'Crossing'
```
rpherbig: I wonder if we should change the stealing data to use hometown instead of province?

Alternatively, we could put this mapping in base-town.

What do you think?

steal.lic
```diff
@@ -253,12 +264,12 @@ class Steal
             keep_item = false
           end
         end
-
         drop_item(item) unless keep_item
 
         move('go portal') if XMLData.room_title == '[[A Junk Yard]]'
       end
     waitrt?
+    start_script('dependency') unless Script.running?('dependency')
```
rpherbig: Why was this added? I would expect `dependency` to always be running...

adamgoerdt:
I would too, but it will occasionally crash in that section with the
following message:

--- Lich: error: 757: unexpected token at 'null'
    C:/Ruby200/lib/ruby/2.0.0/json/common.rb:155:in `parse'
    C:/Ruby200/lib/ruby/2.0.0/json/common.rb:155:in `parse'
--- Lich: dependency has exited.


It happens on 2 of my three characters, and I can't explain it.  It's
happening to Byd as well, so not just local to me.

adamgoerdt:
I can go either way on this.  It probably makes more sense to go based on
hometown, but I worry that a large data change would break some people's
environments (like everyone who hasn't cleared out the base-*.yamls from
their profiles folder).

Province has a couple of hiccups, though.  For example, Theren is probably
not going to work if anyone ever sets their hometown to Therenborough,
because there's a long barge ride or rope bridge climb between Theren and
Haven.  Shard is also a little weird because the stuff in Raven's point
fits better for Shard stealing than Hib because of the ice road.


I think the best way would be to not use either, but maybe call it 'region'
or 'area' and have Zoluren (or Crossing), Shard, Hibarnhvidar, Riverhaven,
etc as the possible regions

rpherbig:
I'm pretty flexible on this, I don't leave Crossing much.

If you like, you could make a PR with everything but the changes to `steal` (to get all of that merged in) and we can keep discussing the stealing changes.

rpherbig:
Some similar discussion happened over in #1094

adamgoerdt:
I've got no issue with the pay-debt and jail-buddy updates going in separately.  I'm excited to get non-Crossing stealing going.  I'll get the new PR set up

PR #1336, opened 2017-02-05
Updated combat-trainer to use analyze instead of bob/weave/circle

adamgoerdt:


rpherbig:
Does this train faster?

adamgoerdt:
Much, much faster. I'm able to lock tactics instead of only getting to 2/34
or so.

rpherbig:
Can this be commonized with the other analyze logic?

adamgoerdt:
Do you mean the barb combo section, or Byd's PR?

I don't think it fits in either place, since it doesn't actually run any
combos.  Tactics trains fine without needing to actually land the combo.  I
think this can replace PR 1302 as a tactics trainer

izawwlgood:
So long as the option to continue using a tactics routine of weave/circle/shove is still there - some guilds dont train tactics using weapons.

combat-trainer.lic
```diff
@@ -1175,7 +1175,17 @@ class TrainerProcess
     when 'App Careful'
       appraise(game_state, 'careful')
     when 'Tactics'
```
rpherbig: As @izawwlgood pointed out, some people might want to continue using weave/bob/circle. I think I'd prefer to see a new training ability added instead of replacing Tactics. Maybe `Tactics Analyze`?

That ensures that existing behavior continues for players, but then gives them the option to change if they want.

combat-trainer.lic
```diff
@@ -1175,7 +1175,17 @@ class TrainerProcess
...
-      bput(%w(weave bob circle).sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty?
+      if not game_state.npcs.empty?
```
rpherbig: I recommend moving this logic into a helper method. Then instead of an `if` around all of the logic, you can use a guard clause (see line 1295 for an example).

combat-trainer.lic
```diff
@@ -1175,7 +1175,17 @@ class TrainerProcess
...
+        case fput('analyze', 'roundtime', 'Analyze what', 'You reveal a massive','Your analysis reveals a massive', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') #unless
```
rpherbig: This should use `bput` instead of `fput`.

combat-trainer.lic
```diff
@@ -1175,7 +1175,17 @@ class TrainerProcess
...
     when 'Tactics'
-      bput(%w(weave bob circle).sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty?
+      if not game_state.npcs.empty?
+        case fput('analyze', 'roundtime', 'Analyze what', 'You reveal a massive','Your analysis reveals a massive', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') #unless 
+        when 'Analyze what'
+          fput('face next')
+          fput('analyze')
+        when 'You must be closer'
+          fput('engage')
+        when 'Your analysis reveals a massive','You reveal a massive'
```
rpherbig: Is it intentional that there is no logic in this `when`?

adamgoerdt: That's a really good question.  What this does is keeps looping 'analyze' until it sees a massive opening, then stops.  I'm not exactly sure *why* it works that way, but that's the desired effect and what I see when I run it

adamgoerdt:
Suggested changes have been implemented

combat-trainer.lic
```diff
@@ -1280,6 +1282,22 @@ class TrainerProcess
...
+    case bput('analyze', 'roundtime', 'Analyze what', 'Your analysis reveals a massive opening',/You reveal an? .* [weakness|opening]/,/Your analysis reveals an? .* [weakness|opening]/, 'You fail to find any holes', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') #unless 
+    when 'Analyze what'
+      fput('face next')
+      analyze(game_state)
+    when 'You must be closer'
+      fput('engage')
+    when 'Your analysis reveals a massive opening'
+      return
+    when /You reveal an? .* [weakness|opening]/,/Your analysis reveals an? .* [weakness|opening]/, 'You fail to find any holes'
+      analyze(game_state)
```
rpherbig: This recursive call means that you can potentially analyze many times in a row. Do you find that preferable to doing one analyze? Will this loop indefinitely if you are unable to analyze a mob (i.e. it's too hard)?

adamgoerdt:
That's right.  The loop is what I'm going for since it really trains
tactics well.  I don't have any sample messaging on what happens if you try
to analyze something beyond your ability, but as long as it doesn't match
the "You reveal an? .* [weakness|opening]/,/Your analysis reveals an? .*
[weakness|opening]" pattern it won't loop indefinitely. It will just try
the one analyze, pause for 15 seconds because it can't find a match, and
continue.

rpherbig:
@adamgoerdt Please update the training abilities documentation on elanthipedia.

adamgoerdt:
Documentation updated

PR #1342, opened 2017-02-06
Added optional argument to train to a specific mindstate

cyrn:
Allows something like 'sorcery 30' in the before or after of a hunting-info yaml. Behavior unchanged if no arguments supplied.

I can't actually cast a sorcery spell so that section is only 95% tested, but it's functionally identical to the focus rune changes.


sorcery.lic
```diff
@@ -10,22 +10,35 @@ class Sorcery
...
   end
 
-  def cast_sorcery_spell?
+  def cast_sorcery_spell?(target_mindstate)
     sorcery_spell = @settings.crossing_training_sorcery
     return false unless sorcery_spell
 
     walk_to(@settings.crossing_training_sorcery_room)
 
+    maxcasts = 3
```
rpherbig: What do you think about moving this logic into a `mind_locked?` helper?

sorcery.lic
```diff
@@ -47,22 +60,28 @@ class Sorcery
...
+      break if DRSkill.getxp('Sorcery') >= maxmindstate 
+       bput('focus my runestone', 'You focus your magical senses')
       waitrt?
     end
 
-    bput('stow runestone', 'You put your')
+   
+      bput('stow runestone', 'You put your')
+
+    return
```
rpherbig: Why is this `return` here?

cyrn: Relic of a less elegant way I tried initially, combined with not really being a Ruby person. It should go.

sorcery.lic
cyrn: Probably should do the check once before calling either cast_sorcery_spell or focus_runestone and just return if applicable to avoid running to the sorcery room needlessly. Then we'd just have "if target_mindstate" in each method.

rpherbig: You'll want to do the exp check after each cast/focus to see if we need to end early.

sorcery.lic
```diff
@@ -10,22 +10,35 @@ class Sorcery
...
-    focus_runestone
+    focus_runestone(args.target_mindstate)
   end
 
-  def cast_sorcery_spell?
+  def cast_sorcery_spell?(target_mindstate)
     sorcery_spell = @settings.crossing_training_sorcery
     return false unless sorcery_spell
 
     walk_to(@settings.crossing_training_sorcery_room)
```
rpherbig: I know it's technically unrelated to your change, but I noticed that this `walk_to` isn't in the right place. If you backfire such that you need healing, you won't run back to the sorcery room - instead you'll stay in the healing room. Can you move this into the loop below?

rcuhljr:
This might be superseded by https://github.com/rpherbig/dr-scripts/pull/1559

PR #1344, opened 2017-02-07
Added condition to deal with empty rings

adamgoerdt:


pick.lic
```diff
@@ -79,12 +79,17 @@ class LockPicker
         next
       end
 
-      until 'not even locked' == bput("pick my #{box} blind", 'not even locked', 'Roundtime')
```
rpherbig: `until` is a loop construct. By removing it, it appears that this code will no longer loop. Is that intentional?

adamgoerdt: Not intentional.  Non-looping bug has been corrected

pick.lic
```diff
@@ -79,15 +79,26 @@ class LockPicker
...
-        waitrt?
-        next unless DRSkill.getxp('Locksmithing') >= 30
+      locked=true
+      while locked
+        case bput("pick my #{box} blind", 'not even locked', 'Roundtime','Find a more appropriate tool')
+        when 'not even locked'
+          waitrt?
+          locked=false
```
rpherbig: Is `locked=false` the same as `break`?

pick.lic
```diff
@@ -79,15 +79,26 @@ class LockPicker
         next
       end
 
-      until 'not even locked' == bput("pick my #{box} blind", 'not even locked', 'Roundtime')
-        waitrt?
-        next unless DRSkill.getxp('Locksmithing') >= 30
```
rpherbig: This line appears to have checked an exp threshold. Without it, will the loop on line 83 continue even if you get above 30/34?

pick.lic
adamgoerdt: They're effectively the same.  Is there a preferred style I should use?

rpherbig: If they are the same, I would prefer `break` so there is one less variable around.

pick.lic
adamgoerdt: Whoops.  Looks like it will finish picking the current box even if locked.  Fix pushed

PR #1353, opened 2017-02-08
Fixed bug in after actions that didn't execute if hunt stopped on skills

adamgoerdt:


hunting-buddy.lic
```diff
@@ -54,31 +54,36 @@ class HuntingBuddy
...
-      before_actions.each do |before_action|
-        action_parts = before_action.split(' ')
-        script_name = action_parts.shift
-        wait_for_script_to_complete(script_name, action_parts)
-      end
+      #Execute scripts to run before the hunt
+      execute_actions(before_actions)
 
-      next if all_skills_at_cap?(stop_on_skills)
+      if all_skills_at_cap?(stop_on_skills)
```
rpherbig: This means that if all skills are at cap, we will call `before` actions and then immediately `after` actions. Do you think we should instead check for skills at cap before the `before` actions are executed?

adamgoerdt: That's a good point.  People seem to be using the before section for various things.  Some are pre-buffing (which isn't needed if the hunt won't start) but some are running athletics or workorders (which should be run), so it's a toss up.

I think that the odds of entering a hunt when all skills are capped is pretty low, which means it makes sense to leave it as it is.

rpherbig: Would you be up for taking a quick poll of LNet/Slack?

adamgoerdt: Quick poll on lnet says that if someone is entering a hunt when all skills are capped then they have an incredibly inefficient setup and wouldn't be too inconvenienced by it being made slightly more inefficient with needles pre-buffing

PR #1355, opened 2017-02-09
Added warning message when health hits 50% of threshold

adamgoerdt:


afk.lic
```diff
@@ -60,6 +62,15 @@ loop do
     fput('depart item') if depart_on_death
     exit_game
   end
+  
+  if not warned and (health < warning_threshold || spirit < warning_threshold)
```
rpherbig: I prefer `!` instead of `not` (and `&&` instead of `and`).

afk.lic
```diff
@@ -60,6 +62,15 @@ loop do
...
+    echo 'Afk - approaching low vitality/spirit threshold'
+    warned = true
+  end
+  
+  if warned and health > warning_threshold and spirit > warning_threshold
```
rpherbig: `&&`

afk.lic
adamgoerdt: Updated

Issue #1326, opened 2017-01-27
Manipulate - request for 2nd creature to be added


Aurayn:
unless I am wrong it only does 1 currently

ohbeanz:
Did a pretty rough pass of this for Truix the other day... I'll make some time and put in a PR for it.

PR #1356, opened 2017-02-09
Manipulate Multiple Enemy NPCs

ohbeanz:
Potential fix for Issue #1326


combat-trainer.lic
```diff
@@ -1102,7 +1102,19 @@ class ManipulateProcess
 
   def manipulate
     bput('manip stop all', 'You relax your will', 'But you aren')
-    bput("manipulate friendship #{@filtered_npcs.first}",  'You attempt to empathically manipulate', 'You strain')
+    @manipulate_count = 0
```
rpherbig: Should this be a local variable (i.e. drop the `@`)? I don't see it used outside of this method...

combat-trainer.lic
```diff
@@ -1102,7 +1102,19 @@ class ManipulateProcess
...
+
+    @filtered_npcs.each_with_index { |npc, index|
```
rpherbig: Our convention for multiple-line blocks is to use `do`/`end` instead of `{}`

combat-trainer.lic
```diff
@@ -1102,7 +1102,19 @@ class ManipulateProcess
...
+      if @manipulate_count >= @threshold
```
rpherbig: You can make this a one-liner: `break if @manipulate_count >= @threshold` (or `break unless @manipulate_count < @threshold` if you think that reads better).

combat-trainer.lic
```diff
@@ -1102,7 +1102,19 @@ class ManipulateProcess
...
+        break
+      end
+      
+      ordinal_string = $ORDINALS[index]
+	  
+      bput("manipulate friendship #{ordinal_string} #{npc}", 'You\'re already manipulating', 'You attempt to empathically manipulate', 'You strain')
+	  @manipulate_count += 1;
```
rpherbig: This line is indented too far.

spinner.lic
```diff
@@ -92,4 +92,4 @@ class Spinnerette
 end
 
 # Call this last to avoid the need for forward declarations
-Spinnerette.new
+Spinnerette.new
```
rpherbig: Could you revert this change?

combat-trainer.lic
ohbeanz: Agreed.

PR #1359, opened 2017-02-10
Simplified release_cyclic

adamgoerdt:


common-arcana.lic
```diff
@@ -166,11 +166,6 @@ module DRCA
   end
 
   def release_cyclics
-    get_data('spells')
-      .spell_data
-      .select { |_name, properties| properties['cyclic'] }
-      .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
-      .map { |_name, properties| properties['abbrev'] }
-      .each { |abbrev| fput("release #{abbrev}") }
+    bput('release cyclic')
```
rpherbig: I don't think `bput` will work without matches. When I tried this line of code I got a timeout error.

adamgoerdt: Switching to fput would work, right?

rpherbig: `fput` often gets tripped up by Roundtime. I feel like this is something that we need to be reliable.

adamgoerdt: I guess I've never noticed fput lose to roundtime, just retry until successful.  

Every spell has a different release message.  I wonder if bput('release cyclic',/.*/) would work.  Will give it a test

adamgoerdt: That does seem to work.  I started a cyclic, then started my simple test method to use bput and spammed a bunch of RT inducing commands.  The cyclic released correctly once I was out of RT

PR #1361, opened 2017-02-10
Added bescort routine for ice road between Shard and Hib

adamgoerdt:


bescort.lic
```diff
@@ -172,6 +178,7 @@ class Bescort
   def search
     fput('search')
     pause
```
rpherbig: Instead of calling `pause` twice, you can pass it an arg (of seconds), e.g. `pause 1`.

bescort.lic
```diff
@@ -359,6 +366,38 @@ class Bescort
...
     end
   end
+  
+  def iceroad(mode)
+    unless mode =~ /shard|hibarnhvidar/i
+      echo 'You must specify shard or hibarnhvidar for traversing the ice road'
+      exit
+    end
+    
+    hib_path = ['nw','ne','ne','e','ne','ne','ne','nw','sw','w','w','nw']
```
rpherbig: I recommend declaring these two as constants (`$` at the top of the class).

bescort.lic
adamgoerdt: Not sure how that pause got in there.  I didn't intend to edit that part.  Must have been an errant keystroke which duplicated the line

bescort.lic
adamgoerdt: Updated

PR #1362, opened 2017-02-10
Add cinder beasts to base-hunting

mwest152:
These 2 rooms are mapped and accessable.  Rooms beyond these are part of
a maze and will require special attention to navigate.


data/base-hunting.yaml
```diff
@@ -498,6 +498,10 @@ hunting_zones:
   - 7278
   - 7279
   - 7271
+  # https://elanthipedia.play.net/Maelshyvean_cinder_beast
```
rpherbig: Could you include the approximate range for this hunting zone?

rpherbig:
There appears to be a merge conflict which needs to be resolved before I can merge this.

data/base-hunting.yaml
mwest152: Ah yeah oops, 450-600

mwest152:
I will fix it!

mwest152:
I like this opportunity to try and learn to do stuff, but I'm kinda stumped as to what I did wrong.

rcuhljr:
Two of you just changed the same spot in the file at the same time.

PR #1367, opened 2017-02-12
Added new tap message for ice skates

adamgoerdt:


bescort.lic
```diff
@@ -379,15 +379,15 @@ class Bescort
     
     mode == 'shard' ? path = $ICE_PATH_SHARD : path = $ICE_PATH_HIB
 
-    case bput('tap skates','I could not find what you were referring to','You tap some')
+    case bput('tap skates','I could not find what you were referring to','You tap some','You tap a pair')
```
rpherbig: I recommend using `exists?(...)` from `common-items`.

adamgoerdt: I've updated the code to use exists, as well as fixed location of the ice road path constant

Issue #1363, opened 2017-02-11
Astrology: Perc Moons


mwest152:
Perc moons teaches astrology similarly to predict weather even at high ranks.  Should be added to astrology script and as an in combat option like predict weather is.

Issue #1349, opened 2017-02-07
Integrate Read the Ripples (RtR) with astrology script


jandersson:
RtR should be a non-lethal method for locking astrology, so It should probably be prioritized over ways walking which is lethal.

I will grab this on my Moonmage and program it into `astrology` if no one gets to it before :]

Aurayn:
RTR SF SCRIPT : TITLE RTR - http://pastebin.com/m3Wr5auH
RTR SF - POST CAST SCRIPT : TITLE RTR+ - http://pastebin.com/UUrNAuuj

This work out as I was too injured to use my telescope. I healed mid predictions but lost several minutes. Even so the messages are the same. 
RTR LOG WHEN INJURED: -  http://pastebin.com/6ZWqKunF

jandersson:
Thank you sir!

Aurayn:
Right back atcha

PR #1368, opened 2017-02-12
Update Astrology

jandersson:
This trains using Read the ripples if the user has it listed in their astrology buffs. Also adds event prediction and perc moons to the training routine. Another method is included which uses `predict state all` to gauge pool level. Gauging the levels will be useful for doing targeted predictions in a future update 

Addresses #1363 #1349

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -27,6 +27,7 @@ class Astrology
   def do_buffs(settings)
     settings.astrology_buffs['spells'].each do |spell|
       if spell['abbrev'].downcase == 'rtr'
+        echo Time.now - UserVars.rtr_timer 
         next if !UserVars.rtr_timer.nil? && (Time.now - UserVars.rtr_timer < 3600)
       end
       echo "Buffing: #{spell}" if UserVars.astrology_debug
@@ -44,7 +45,11 @@ class Astrology
     result.compact.select { |data| data['circle'] <= DRStats.circle }
   end
 
+  def check_stars
+  end
+
   def check_weather
+    echo "Checking the weather" if UserVars.astrology_debug
     fput('predict weather')
     waitrt?
   end
@@ -55,8 +60,8 @@ class Astrology
     while DRSpells.active_spells.include?('Read the Ripples')
       line = get?
 
-      @constellations.find{ |body|  /As your consciousness drifts amongst the currents of Fate .* #{body['name']}/i =~ line }                                                                     
-      
+      res = @constellations.find{ |body|  /As your consciousness drifts amongst the currents of Fate .* #{body['name']}/i =~ line }                                                                     
+      echo res unless res.nil?
     end 
     return
   end
--- a/astrology.lic
+++ b/astrology.lic
@@ -52,9 +52,9 @@ class Astrology
   end
 
   def check_moons
-    %w(xibar yavash katamba).each do |moon| 
-	    bput("pow #{moon}", 'roundtime') 
-	    waitrt? 
+    %w(xibar yavash katamba).each do |moon|
+      bput("pow #{moon}", 'roundtime')
+      waitrt?
     end
   end
 
@@ -75,7 +75,7 @@ class Astrology
     bput('predict state all', 'roundtime')
     lines = reget(50).reverse
     lines.select! { |line| line =~ /celestial influences/ }
-    @pools.select! { |name, _level| @pools[name] = lines.detect { |line| line =~ /#{name}/ } }
+    @pools.select! { |name, _level| @pools[name] = lines.find { |line| line =~ /#{name}/ } }
     @pools.each { |name, text| pools_to_size.each { |match_text, value| @pools[name] = value if match_text =~ text } }
     echo " @pools: #{@pools}" if UserVars.astrology_debug
     waitrt?
@@ -90,7 +90,7 @@ class Astrology
       result = bput('study sky', 'You feel a lingering sense', 'You feel it is too soon', 'Roundtime', 'You are unable to sense additional information', 'detect any portents')
       waitrt?
       return if ['detect any portents'].include? result
-      check_pools
+
       break if @pools['future events'] == prev_size
       prev_size = @pools['future events']
     end
--- a/astrology.lic
+++ b/astrology.lic
@@ -137,7 +137,7 @@ class Astrology
       'future events' => 'future events'
     }
 
-    @pools.reject { |_skill, size| size == 0 }
+    @pools.reject { |_skill, size| size.zero? }
           .each   { |skill, _size| align(skillset_to_pool[skill]) }
   end
 
--- a/astrology.lic
+++ b/astrology.lic
@@ -18,12 +18,17 @@ class Astrology
 
     do_buffs(settings)
 
+    check_ripples
     check_astral
+    check_weather
     train_astrology
   end
 
   def do_buffs(settings)
     settings.astrology_buffs['spells'].each do |spell|
+      if spell['abbrev'].downcase == 'rtr'
+        next if !UserVars.rtr_timer.nil? && (Time.now - UserVars.rtr_timer < 3600)
+      end
       echo "Buffing: #{spell}" if UserVars.astrology_debug
       buff(spell, settings)
     end
@@ -37,6 +42,26 @@ class Astrology
       result << all_bodies.find { |body| /\b#{body['name']}\b/i =~ line && line !~ /below the horizon/ }
     end
     result.compact.select { |data| data['circle'] <= DRStats.circle }
+  end
+
+  def check_weather
+    fput('predict weather')
+    waitrt?
+  end
+
+  def check_ripples
+    return unless DRSpells.active_spells.include?('Read the Ripples')
+    UserVars.rtr_timer = Time.now
+    while DRSpells.active_spells.include?('Read the Ripples')
+      line = get?
+
+      @constellations.find{ |body|  /As your consciousness drifts amongst the currents of Fate .* #{body['name']}/i =~ line }                                                                     
+      
+    end 
+    return
+  end
+
+  def observe(body)
   end
 
   def check_astral
@@ -53,9 +78,6 @@ class Astrology
   end
 
   def train_astrology
-    fput('predict weather')
-    waitrt?
-
     vis_bodies = visible_bodies
 
     night = vis_bodies.find { |body| body['constellation'] }
--- a/astrology.lic
+++ b/astrology.lic
@@ -16,14 +16,14 @@ class Astrology
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
     @pools = {
-	    'lore' => 0,
-              'magic' => 0,
-	      'survival' => 0,
-	      'offensive combat' => 0,
-	      'defensive combat' => 0,
-	      'future events' => 0
+      'lore' => 0,
+      'magic' => 0,
+      'survival' => 0,
+      'offensive combat' => 0,
+      'defensive combat' => 0,
+      'future events' => 0
     }
-    
+
     do_buffs(settings)
 
     train_astrology
@@ -31,9 +31,9 @@ class Astrology
 
   def do_buffs(settings)
     settings.astrology_buffs['spells'].each do |spell|
-      if spell['abbrev'].downcase == 'rtr'
-	Flags.add('rtr-expire', spell['expire'])
-	         
+      if spell['abbrev'].casecmp('rtr').zero?
+        Flags.add('rtr-expire', spell['expire'])
+
         next if !UserVars.astrology_timers['rtr'].nil? && (Time.now - UserVars.astrology_timers['rtr'] < 3600)
       end
       echo "Buffing: #{spell}" if UserVars.astrology_debug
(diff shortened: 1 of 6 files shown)


astrology.lic
```diff
@@ -39,6 +51,67 @@ class Astrology
...
+    bput('predict state all', 'roundtime')
+    lines = reget(50).reverse
+    lines.select! { |line| line =~ /celestial influences/ }
+    @pools.select! { |name, _level| @pools[name] = lines.detect { |line| line =~ /#{name}/ } }
+    @pools.each { |name, text| pools_to_size.each { |match_text, value| @pools[name] = value if match_text =~ text } }
+    echo " @pools: #{@pools}" if UserVars.astrology_debug
+  end
+
+  def check_events
+    # return if !UserVars.astrology_timers['event'].nil? && Time.now - UserVars.astrology_timers['event'] < 600
```
rpherbig: Should this line be removed?

astrology.lic
```diff
@@ -15,15 +15,27 @@ class Astrology
...
 
     do_buffs(settings)
 
-    check_astral
     train_astrology
   end
 
   def do_buffs(settings)
     settings.astrology_buffs['spells'].each do |spell|
+      if spell['abbrev'].casecmp('rtr').zero?
```
rpherbig: Does this mean RTR will only be cast once an hour, even if the buff expires before that?

astrology.lic
```diff
@@ -87,40 +168,52 @@ class Astrology
...
     pause 2
     waitrt?
 
-    bput('stow telescope', 'you put') if @have_telescope
-
     observed['pools']
       .reject { |_skill, value| value.nil? }
       .each { |skill, _value| align(skill) }
-    bput('predict anal', 'roundtime')
+    # bput('predict anal', 'roundtime')
```
rpherbig: Can this line be removed?

common-arcana.lic
```diff
@@ -60,8 +60,13 @@ module DRCA
   def buff(spell, settings)
     prepare?(spell['abbrev'], spell['mana'], spell['symbiosis'])
     find_charge_invoke_stow(settings.cambrinth, settings.stored_cambrinth, settings.cambrinth_cap, settings.dedicated_camb_use, spell['cambrinth'])
-    waitcastrt?
-    cast?(spell['cast'], spell['symbiosis'], spell['before'], spell['after'])
+
+    if spell['ritual']
+      ritual(spell)
```
rpherbig: It looks like prepare and cambrinth are done before this check and also done in `ritual`. Does that mean it will attempt to prepare, cambrinth, prepare, focus?

data/base-spells.yaml
```diff
@@ -17,6 +17,7 @@ invoke_messages:
 - Kneeling down, you draw the spell pattern's shadow with
 - You make sweeping gestures through the air
 - You lift your staff reverently
+- You lift your staff toward the .* and will the mana streams to bend into it
```
rpherbig: I recommend combining this with line 19, maybe `You lift your staff (reverently|toward)`?

data/base-spells.yaml
```diff
@@ -624,6 +642,16 @@ spell_data:
...
     recast: 1
+  Read the Ripples:
+    skill: Utility
+    abbrev: RTR
+    mana: 300
+    ritual: true
+    before: 
+    - message: sit
+      matches: 
+      - You sit|You are already sittings
```
rpherbig: Is 'sittings' a typo?

data/base-spells.yaml
```diff
@@ -624,6 +642,16 @@ spell_data:
...
     mana: 5
     recast: 1
+  Read the Ripples:
+    skill: Utility
+    abbrev: RTR
+    mana: 300
+    ritual: true
+    before: 
+    - message: sit
+      matches:
```
rpherbig: You can separate these matches into separate lines:
```
matches:
  - You sit
  - You are already sittings
```

dependency.lic
```diff
@@ -845,14 +845,14 @@ class SetupFiles
...
             .each { |_name, data| s.lockpick_buffs['spells'][i] = data.merge(spell) }
     end
-
+    
     s.astrology_buffs['spells']
      .each_with_index
      .select do |spell, i|
       spells.select { |_name, data| data['abbrev'].casecmp(spell['abbrev'].downcase).zero? }
-            .each { |_name, data| s.lockpick_buffs['spells'][i] = data.merge(spell) }
+            .each { |_name, data| s.astrology_buffs['spells'][i] = data.merge(spell) }
```
rpherbig: Nice catch!

astrology.lic
```diff
@@ -39,6 +51,67 @@ class Astrology
     result.compact.select { |data| data['circle'] <= DRStats.circle }
   end
 
+  def check_moons
+    %w(xibar yavash katamba).each { |moon| bput("pow #{moon}", 'roundtime'); waitrt? }
```
rpherbig: I'd prefer a linebreak instead of `;` (it's easy to miss the second action this way).

astrology.lic
```diff
@@ -39,6 +51,67 @@ class Astrology
...
+      /You have a powerful understanding of the celestial influences over/ => 9,
+      /You have a complete understanding of the celestial influences over/ => 10
+    }
+    bput('predict state all', 'roundtime')
+    lines = reget(50).reverse
+    lines.select! { |line| line =~ /celestial influences/ }
+    @pools.select! { |name, _level| @pools[name] = lines.detect { |line| line =~ /#{name}/ } }
+    @pools.each { |name, text| pools_to_size.each { |match_text, value| @pools[name] = value if match_text =~ text } }
+    echo " @pools: #{@pools}" if UserVars.astrology_debug
+  end
```
rpherbig: Should there be a `waitrt?` at the end of this method (to account for line 72)?

astrology.lic
```diff
@@ -39,6 +51,67 @@ class Astrology
...
+      /You have a significant understanding of the celestial influences over/ => 6,
+      /You have a potent understanding of the celestial influences over/ => 7,
+      /You have a insightful understanding of the celestial influences over/ => 8,
+      /You have a powerful understanding of the celestial influences over/ => 9,
+      /You have a complete understanding of the celestial influences over/ => 10
+    }
+    bput('predict state all', 'roundtime')
+    lines = reget(50).reverse
+    lines.select! { |line| line =~ /celestial influences/ }
+    @pools.select! { |name, _level| @pools[name] = lines.detect { |line| line =~ /#{name}/ } }
```
rpherbig: It's not a big deal, but we've been using `find` instead of `detect` throughout the codebase.

astrology.lic
```diff
@@ -39,6 +51,67 @@ class Astrology
...
+      /You have a modest understanding of the celestial influences over/ => 4,
+      /You have a decent understanding of the celestial influences over/ => 5,
+      /You have a significant understanding of the celestial influences over/ => 6,
+      /You have a potent understanding of the celestial influences over/ => 7,
+      /You have a insightful understanding of the celestial influences over/ => 8,
+      /You have a powerful understanding of the celestial influences over/ => 9,
+      /You have a complete understanding of the celestial influences over/ => 10
+    }
+    bput('predict state all', 'roundtime')
+    lines = reget(50).reverse
```
rpherbig: There's probably a more idiomatic way to do this (e.g. `map` instead of assigning to vars multiple times). I'd have to poke at it a bit to give some code. @kamarcum or @rcuhljr ?

jandersson: I bet there is. I was just happy that it worked! I noticed that I'm using a hash to map to 0-10 when a list would do the same job with the indices giving the mapped values.

astrology.lic
jandersson: I think I can get rid of the timer and just go with the expire message

astrology.lic
```diff
@@ -23,18 +23,23 @@ class Astrology
...
     settings.astrology_buffs['spells'].each do |spell|
-      if spell['abbrev'].casecmp('rtr').zero? 
-        next unless Flags['rtr-expire']
-	Flags.reset['rtr-expire']
+      if spell['abbrev'].casecmp('rtr').zero?
+	case Flags['rtr-expire']
+	when nil
+	  Flags.add('rtr-expire', spell['expire'])
+	when false
+          next
```
jandersson: This is still going to cause an issue if RtR fails to be cast, then the expire flag will never be raised

rpherbig: If RtR fails to cast, why do you care about its expiration?

jandersson: I just want the flag set on a successful cast. If it fails and still sets the expiration flag, then astrology buffs will never try to cast it again.

rpherbig:
Why do you care about the expiration at all?

jandersson:
Its only castable once every ~30-60 minutes, just trying to save time in the buffing stage

rpherbig:
Will it fail at the prep stage? It seems easier to handle there than timer
logic each place we might want to cast RtR.

jandersson:
Yea its going to fail when cast. There will be a messages for "you werent sitting down", "you're not outside", "you weren't using a focus", and "the spell hasn't expired yet."

rpherbig:
There appears to be a merge conflict which needs resolved.

jandersson:
Ah nice, it looks like bones were added! I will fix it over the weekend

jandersson:
Alright, I've got this working the way I wanted. No problems after a few hours of testing

astrology.lic
```diff
@@ -90,56 +171,72 @@ class Astrology
...
+      ['Roundtime', 'Clouds obscure'].include? bput("observe #{body} in sky", 'Your search for', 'You see nothing regarding the future', 'Clouds obscure', 'Roundtime')
+    end
+  end
+  
   def roll_bones
     if @divination_bones_tied
       bput("untie bones from my #{@divination_bones_container}", 'divination bones')
     else
-      bput("get bones from my #{@divination_bones_container}", 'divination bones')
+      bput("get bones from my #{@divination_bones_container}", 'divination bones', 'you get')
```
rpherbig: Why are there two messages here?

astrology.lic
```diff
@@ -90,56 +171,72 @@ class Astrology
...
     end
     bput('roll my bones', 'roundtime')
     waitrt?
     if @divination_bones_tied
       bput("tie bones to my #{@divination_bones_container}", 'divination bones')
     else
-      bput("put bones in my #{@divination_bones_container}", 'divination bones')
+      bput("put bones in my #{@divination_bones_container}", 'divination bones', 'you put')
```
rpherbig: Why are there two messages here?

astrology.lic
```diff
@@ -55,10 +127,21 @@ class Astrology
...
+    skillset_to_pool = {
+      'offensive combat' => 'offense',
+      'defensive combat' => 'defense',
+      'magic' => 'magic',
+      'survival' => 'survival',
+      'lore' => 'lore',
+      'future events' => 'future events'
+    }
 
+    @pools.reject { |_skill, size| size == 0 }
```
rpherbig: `.zero?`

astrology.lic
jandersson: The first one wasn't matching for me

common-arcana.lic
```diff
@@ -87,6 +91,12 @@ module DRCA
     case DRC.bput(cast_command || 'cast', get_data('spells').cast_messages)
     when /^Your target pattern dissipates/, /^You can't cast that at yourself/, /^You need to specify a body part to consume/
       fput('release spell')
+    when /^At the ritual's peak, your prophetic connection blooms a thousand-fold/
+      if Flags['rtr-expire'].nil? 
+        Flags.add('rtr-expire', get_data('spells').spell_data['Read the Ripples']['expire'])
```
rpherbig: I think it might be easier to add the Flag in `astrology`, then you can reset it here instead of conditionally adding it.

astrology.lic
```diff
@@ -18,15 +18,25 @@ class Astrology
...
     do_buffs(settings)
 
-    check_astral
     train_astrology
   end
 
   def do_buffs(settings)
     settings.astrology_buffs['spells'].each do |spell|
+      if spell['abbrev'].casecmp('rtr').zero?
+        next if !Flags['rtr-expire'].nil? || Flags['rtr-expire'] == false
```
rpherbig: If you add the Flag in `astrology`, you can remove this `nil?` check.

And you can use `!Flags['rtr-expire']`

astrology.lic
```diff
@@ -18,15 +18,25 @@ class Astrology
     @ap_training = settings.astral_plane_training['train_in_ap']
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
-
+    @pools = {
```
rpherbig: Does this need to be an instance var, or could it be passed into and out of methods?

jandersson:
After some poking, it is working as usual. The pools state is returned by `check_pools` instead of being an instance variable. Rtr-expire flag is set in astrology and reset in common-arcana when successfully cast

astrology.lic
jandersson: I'm guessing the person who added the bones usage has a different set than I do. My getting/stowing commands were not responding to 'divination bones', so I added the general text for getting or storing something.

PR #1372, opened 2017-02-15
Fixed cluttered room error

adamgoerdt:


bescort.lic
```diff
@@ -390,7 +390,13 @@ class Bescort
       path.each do |movement|
         move(movement)
         waitrt
-        fput('collect rock')
+        case bput('collect rock','You manage to collect a pile', 'The room is too cluttered')
```
rpherbig: What would you think about moving the collect logic into `common.lic`, similar to how `forage?` is done? I see some overlap here with the collect logic in `crossing-training`.

adamgoerdt: Logic has been moved and collecting in crossing-training and bescort consolidated to use it

common.lic
```diff
@@ -93,6 +93,16 @@ module DRC
     end
     true
   end
+  
+  def collect(item)
+    case bput("collect #{item}",'You manage to collect a pile', 'The room is too cluttered')
+    when 'The room is too cluttered'
+      fput('kick pile')
```
rpherbig: What do you think about `fput('dump junk')` after kicking?

common.lic
```diff
@@ -93,6 +93,16 @@ module DRC
...
+      fput('kick pile')
+      fput('kick pile')
+      fput("collect #{item}")
+    end
```
rpherbig: I recommend `waitrt?` at the end of the method (so you don't have to do it every time you call it).

common.lic
adamgoerdt: I'm thinking to err on the side of caution regarding dump junk.  We can't guarantee that there aren't people in the room who have things on the ground they don't want to lose.

common.lic
adamgoerdt: Added the wait

PR #1375, opened 2017-02-17
Updating Astrology to support using divination bones when making futu…

m-burroughs:
…re predictions.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -11,6 +11,9 @@ class Astrology
   def initialize
     settings = get_settings
     @constellations = get_data('constellations').constellations
+    @have_divination_bones = settings.have_divination_bones
+    @divination_bones_container = settings.divination_bones_storage['container']
+    @divination_bones_tied = settings.divination_bones_storage['tied']
     @have_telescope = settings.have_telescope
     @ap_training = settings.astral_plane_training['train_in_ap']
     @ap_source = settings.astral_plane_training['train_source']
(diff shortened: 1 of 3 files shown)


astrology.lic
```diff
@@ -11,6 +11,9 @@ class Astrology
   def initialize
     settings = get_settings
     @constellations = get_data('constellations').constellations
+    @have_divination_bones = settings.have_divination_bones
```
rpherbig: Please update `base.yaml` with defaults for these settings.

astrology.lic
```diff
@@ -121,10 +124,29 @@ class Astrology
     waitrt?
   end
 
+  def roll_bones
+    if @divination_bones_tied
+      bput('untie bones from my '+@divination_bones_container, 'divination bones')
```
rpherbig: When possible, we prefer string interpolation (something like `"untie bones from my #{@divination_bones_container}"`).

astrology.lic
```diff
@@ -121,10 +124,29 @@ class Astrology
...
+      bput('put bones in my '+@divination_bones_container, 'divination bones')
+    end
+  end
+
   def align(skill)
     bput("align #{skill}", 'You focus internally')
     waitrt?
-    bput('predict future', 'roundtime')
+    if @have_divination_bones
+      roll_bones
```
rpherbig: Does rolling your divination bones take the place of `predict future`, or can both be done?

rpherbig:
Would you be willing to upload a YAML file of one of your characters that has an example of these new settings?

astrology.lic
m-burroughs: Will do.

astrology.lic
m-burroughs: I'll make the appropriate update.

astrology.lic
m-burroughs: Yes, it takes the place of `predict future`.  It actually is capable of performing a much stronger prediction than just `predict future` once you are bonded with the bones sufficiently (which makes the script great, because each roll increases the bond)

m-burroughs:
Absolutely, I'll include my training YAML with my next commit.

m-burroughs:
I gave the script a few test runs, and it looks good... let me know if there are more changes you would like to review.

PR #1383, opened 2017-02-20
Two new scripts for managing contacts

lloyd-delacroix:
Reads from two YAML files to import a list of characters and attributes. base-contacts.yaml belongs in the data folder and yourname-contacts.yaml belongs in the profiles folder. Any information in yourname-contacts.yaml will take precedence over the base file.

display-guild.lic uses the same code as textsubs to append a prefix in front of all known players. For example, syntyche the warmage will be displayed as W:Syntyche in all text. Bards will be prefixed with D: and thieves will be prefixed with X: to avoid letter conflicts.

yellowpages.lic requires a single name as an argument. It will then query the two YAMLs for that character and print out any information recorded about them.

~~Currently these scripts must be trusted in order to read the contents of the YAML files. Eventually I would like to integrate these fully with dependency to have it handle all of the YAML, but I'm new to ruby and completely overwhelmed by it still.~~ A monitor has been added to dependency to handle the reading of YAML files and these scripts no longer need to be trusted.


data/base-contacts.yaml
```diff
@@ -0,0 +1,3356 @@
+contacts:
```
rpherbig: Should this file be stored in the repository? Is the intent for each player to have their own version of this file?

display-guild.lic
```diff
@@ -0,0 +1,86 @@
+Contacts = YAML.load_file('./scripts/data/base-contacts.yaml')
```
rpherbig: Our convention has been to have a documentation block at the top of each script, something like:
```
=begin
  Documentation: https://elanthipedia.play.net/Lich_script_repository#accept-sell
=end
```

And a corresponding entry on elanthipedia.

yellowpages.lic
```diff
@@ -0,0 +1,56 @@
+#Load base contacts YAML
```
rpherbig: Our convention has been to have a documentation block at the top of each script, something like:
```
=begin
  Documentation: https://elanthipedia.play.net/Lich_script_repository#accept-sell
=end
```

And a corresponding entry on elanthipedia.

profiles/sample-contacts.yaml
```diff
@@ -0,0 +1,19 @@
+contacts:
```
rpherbig: What is the purpose of this file?

yellowpages.lic
```diff
@@ -0,0 +1,56 @@
...
+Contacts = YAML.load_file('./scripts/data/base-contacts.yaml')
+Contacts = Contacts['contacts']
+
+#DEBUG INFO
+#puts "---Base Contacts---"
+#puts Contacts.inspect
+#puts Contacts.class
+
+#Check for <CharName>-contacts.yaml file and merges it if exists
+if File.exist?('./scripts/profiles/'+checkname+'-contacts.yaml')
```
rpherbig: Our convention has been to use string interpolation for cases like this (e.g. `"./scripts/profiles/#{checkname}-contacts.yaml"`).

yellowpages.lic
```diff
@@ -0,0 +1,56 @@
...
+#DEBUG INFO
+#puts "---Merged Contacts---"
+#puts Contacts.inspect
+#puts Contacts.class
+#puts "-----"
+
+
+arg_definitions = [
+    [
+    { name: 'search', regex: /[a-zA-Z]/, optional: false, description: 'Name of player to look up in contacts.' }
```
rpherbig: I recommend having a `debug` arg that will output your debug messages (instead of leaving them commented out).

yellowpages.lic
```diff
@@ -0,0 +1,56 @@
...
+#puts args.class
+
+Record = Contacts[args.search]
+if Record.nil?
+	puts "User not found in contacts"
+else
+	#puts Record.inspect
+	puts "Information for " + args.search
+	puts " "
+	puts "Guild: " + Record['guild'].inspect.gsub(/nil/,'')
```
rpherbig: Would 'unknown' be a more informative message in the nil case?

yellowpages.lic
```diff
@@ -0,0 +1,56 @@
...
+Contacts = Contacts['contacts']
+
+#DEBUG INFO
+#puts "---Base Contacts---"
+#puts Contacts.inspect
+#puts Contacts.class
+
+#Check for <CharName>-contacts.yaml file and merges it if exists
+if File.exist?('./scripts/profiles/'+checkname+'-contacts.yaml')
+	puts "Personal Contact file found! Now importing!"
```
rpherbig: I don't think `puts` will output on all Lich front-ends. Our convention has been to use `echo` instead.

display-guild.lic
```diff
@@ -0,0 +1,86 @@
...
+#DEBUG INFO
+#puts "---Merged Contacts---"
+#puts Contacts.inspect
+#puts Contacts.class
+#puts "-----"
+
+
+# quiet
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#stalker
```
rpherbig: Is this actually two scripts in the same file?

display-guild.lic
```diff
@@ -0,0 +1,86 @@
...
+
+DownstreamHook.remove('stalker')
+DownstreamHook.add('stalker', action)
+
+before_dying do
+  DownstreamHook.remove('stalker')
+end
+
+# defaults
+for index in 0 ... Names.size
```
rpherbig: You could also do `Names.each do |name|` and then replace `Names[index]` with `name`.

display-guild.lic
```diff
@@ -0,0 +1,86 @@
...
+	if (Contacts[Names[index]]['guild'] == 'Thief')
+		Contacts[Names[index]]['guild'] = 'X'
```
rpherbig: How strongly do you feel about one-character abbreviations?

Would 'Tr' and 'Th' / 'Bn' and 'Bd' work instead? Or maybe three characters? I'm not sure how much room there is on the screen...

data/base-contacts.yaml
lloyd-delacroix: This file is created using all of the information in the "of renown" charts for each of the guilds on elanthipedia. As such, I'd consider all of it "public information" and planned to have it be the base file everyone got on install to give them some benefit right away. The community could also choose to continue adding more information if desired. It might make more sense to remove the lines for relationship, notes, and alts, since I don't foresee those ever being used in the public file, but I haven't tested how the scripts respond to those being missing.

There would also be a personal file yourname-contacts.yaml in the profiles folder. This is where each player could record personal notes or add information on people that they didn't wish to share with the entire lich community.

profiles/sample-contacts.yaml
lloyd-delacroix: This file should be renamed to <yourname>-contacts.yaml and is the file where people would put information on any characters they did not want contributed to the public file or any personal notes on characters. The method to merge the two YAMLs will give precedence to this file - any fields populated here will apply over the contents of the base file.

display-guild.lic
lloyd-delacroix: I'm certainly open to other ideas! From my experience in testing, the single character with substitutions worked better than two or three characters. Tr and Th still looked too similar to distinguish at a glance, particularly with text scrolling by. Three characters (it becomes four with the colon) before every name took up too much space, and it looked weird using only one character for half the guilds but three for others.

display-guild.lic
lloyd-delacroix: Stalker was the original name I was using for this script, but decided to change it to something more descriptive. I figured there wasn't much need to go back and change the name in the code itself, but I had forgotten about the reference to the documentation. That will need to be changed.

data/base-contacts.yaml
rpherbig: A seed file makes sense to store in the repo.

The convention we've been using for personal files is to store them in the lich directory (you can browse through `dependency.lic` to see that in action).

lloyd-delacroix: Sorry, when I say profiles folder I meant /lich/scripts/profiles/ which is the same place the yourname-setup.yaml file is.

lloyd-delacroix:
I'm working on changing / fixing a number of these now. What's the best way to submit the updates? Can I do it within the context of this pull request, or do I need to make all of the changes and start a new pull request?

rpherbig:
If you make additional commits to your fork, they'll be added to this PR automatically.

lloyd-delacroix:
Made a series of changes in response to comments and commited to my fork.
-Started each script with documentation block
-Used string interpolation for YAML files
-Incorporated debug arguments into both scripts to print debug info
-Changed 'puts' to either 'echo' or 'respond'
-Fixed name of script in documentation block
-Changed for loop to 'Names.each do'

display-guild.lic
```diff
@@ -0,0 +1,91 @@
...
+end
+
+Names.each do |name|
+	if (Contacts[name]['guild'] == 'Thief')
+		Contacts[name]['guild'] = 'X'
+	end
+	if (Contacts[name]['guild'] == 'Bard')
+		Contacts[name]['guild'] = 'D'
+	end
+	Stalker.add((name.capitalize),Contacts[name]['guild'].chr+':'+name.capitalize)
```
rpherbig: This is another place string interpolation could be used (`"#{Contacts[name]['guild'].chr}:#{name.capitalize}"`)

yellowpages.lic
```diff
@@ -0,0 +1,70 @@
...
+end
+
+#DEBUG INFO
+echo "---Merged Contacts---" if $debug_mode_cc
+echo Contacts.inspect if $debug_mode_cc
+echo Contacts.class if $debug_mode_cc
+echo "-----" if $debug_mode_cc
+
+if args.search.nil?
+	echo "What user do you wish to search for?"
```
rpherbig: I think this case could be avoided by making the search arg not-optional.

yellowpages.lic
```diff
@@ -0,0 +1,70 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#yellowpages
+=end
+
+arg_definitions = [
+    [
+	{ name: 'debug', regex: /debug/i, optional: true, description: 'Very verbose, may cause lich to hang for several seconds if using large contact files' },
+    { name: 'search', regex: /[a-zA-Z]/, optional: true, description: 'Name of player to look up in contacts.' }
```
rpherbig: How about something like 'player' or 'name' instead of 'search'?

rpherbig:
If you want to remove the need for your scripts to be trusted (and follow the existing convention), you could add a manager to `dependency` to do the read/write for you. I think `ReportbotManager` is a good example for you to follow.

display-guild.lic
```diff
@@ -0,0 +1,91 @@
...
+#DEBUG INFO
+echo "---Base Contacts---" if $debug_mode_cc
+echo Contacts.inspect if $debug_mode_cc
+echo Contacts.class if $debug_mode_cc
+
+#Check for <CharName>-contacts.yaml file and merges it if exists
+if File.exist?("./scripts/profiles/#{checkname}-contacts.yaml")
+	echo "Personal contact file found - now importing." if $debug_mode_cc
+	
+	Personal_Contacts = YAML.load_file("./scripts/profiles/#{checkname}-contacts.yaml")
```
rpherbig: This means that two characters run by the same player do not share contact information. Is that the goal?

yellowpages.lic
lloyd-delacroix: I tried making search not optional first, but when I added the debug argument things wouldn't work with ;script debug name. It would only work with ;script debug OR ;script name. This was the quickest way to get it working again.

display-guild.lic
lloyd-delacroix: Hmmm, that's a great point that I hadn't really thought about (I only have one character myself). I feel like people would definitely want that information shared across characters, and duplicating the file and updates would be a bit of a pain.

Maybe the best bet would be to have it referenced as "personal-contacts.yaml" still out of the /scripts/profiles directory, but also put it in the gitignore so that other contributors don't accidentally upload theirs and overwrite it. It would still be the same amount of work as copying the sample file to yourname-contacts is now, and is a process users are already used to with the setup yamls.

I had also had the thought that contacts could just be added to the setup yaml (and tried to structure the yaml such that it would be an easy conversion) but I'm worried that for some users who have several dozen or hundreds of contacts, it would make their setup yaml too messy.

rpherbig: I agree that it should not be part of your normal YAML. The dependency manager I linked you to (as well as bankbot) has an example of saving a YAML file in the lich root directory so it can be accessed by multiple characters.

lloyd-delacroix:
So, in the case of reportbot, it seems like what they do is kick off another script, wait for it to populate UserVars.players_online, and then the original script reads UserVars.players_online - is this right?

Is UserVars global to lich? Would the trick be to have dependency open the files and write the hashtable to UserVars.contacts and then have my scripts just read out of that?

rpherbig:
UserVars.players_online is doing something different than what you're trying to do. Check out `load_reportbot_whitelist` - that's very similar to your contacts list.

lloyd-delacroix:
Alright, so the reportbot code got me going in the right direction and I got everything up and working using dependency! What I can't figure out now is why github is showing deleting everything out of dependency and adding it all back in. When I did the merge on my branch, the diff showed everything just fine. Any thoughts?

rcuhljr:
It feels like we're mixing data concerns here. Some of the data (guild, circle) feels like it belongs in the base file, the notes/alts/relationship all feels like user specific data. You could let people store their own custom information in `Yourname-contacts.yaml` and then you eliminate the need for an entirely new class in dependency.

rpherbig:
There's also a merge conflict in `dependency.lic`. You'll need to update that before we can merge the PR.

PR #1393, opened 2017-02-23
Added Will of Winter prep message

adamgoerdt:


data/base-spells.yaml
```diff
@@ -56,6 +56,7 @@ prep_messages:
 - You produce an illusory coin and begin rolling it across your knuckles
 - You softly whisper the words for the
 - Though softly humming at first
+- You hear the frigid chanting
```
rpherbig: I think it would be easier to add `frigid` to line 48.

data/base-spells.yaml
```diff
@@ -56,6 +56,7 @@ prep_messages:
 - You produce an illusory coin and begin rolling it across your knuckles
```
rpherbig: I know it isn't related to your change, but I think line 26 can be removed (it is redundant with line 32)

data/base-spells.yaml
adamgoerdt: Makes sense.  Fixed

PR #1406, opened 2017-02-25
Spell target body parts

tvalean:
Focus on a body part.


jayRyan24:
Could be simplified to just have take in a spell target in another parameter. 

` - skill: Targeted Magic
   name: Fire Ball
   abbrev: FB
   mana: 15
   min_threshold: 2
   max_threshold: 3
   cast_only_to_train: true
   target_loc: head
`
Then simplify the code to always

`if (@target_loc)
"prep @spell @mana"
"tar @loc"
else
"target @spell @mana"
`

PR #1407, opened 2017-02-26
Theurgy2.0

drboldwin:
This is not ready yet but I want it up here for editing purposes - is nearly done.

### DIFF
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -29,7 +29,7 @@ class Theurgy
     forage? 'dirt'
 
     get_from_container(@water_holder)
-    bput('commune eluned', 'completed this commune too recently', 'You grind some dirt in your fist')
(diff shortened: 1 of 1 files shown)


rpherbig:
@drboldwin Any updates on this?

drboldwin:
They actually made some changes that I was pushing for regarding wine's stackability and so I want to go in there and add capabilities for that as well.

They made it so standard wine can be blessed which allows for use of kegs (process being)
Buy wine - fill keg with wine (up to 240 parts)
then in the using wine you would either (get keg) or if you are wearing it can skip this step
POUR KEG
PREP BLESS > CAST WINE
Then use wine for devotion and pour the remainder back in the keg. 

I am unsure if I want to just have the script first look for a keg or make it a setting
Kegs currently can only be purchased during festivals and so are not readily accessible yet but I did buy a couple.

PR #1412, opened 2017-02-27
Updated combat-trainer collect to allow for user-defined items

adamgoerdt:


profiles/base.yaml
```diff
@@ -193,6 +193,7 @@ crossing_training_requires_movement:
 - Trading
 forage_item:
 braid_item:
+collect_item: rock
```
rpherbig: I don't think the name of this setting makes it clear that it is only used in `combat-trainer` (nor does its placement next to some non-combat settings).

adamgoerdt: I was thinking about updating crossing-trainer to use this change, too, especially since people seem to be congregating now in the empath guild where rocks are not available, but didn't think it fit in the same commit.

The setting is inherited by the script that call it, so in the current implementation it's possible to override the collect_item in each yaml, letting you have a different collect_item in your crossing_training, primary hunt and secondary hunts.

rpherbig: It seems reasonable to me to, as part of this PR, make use of this setting in more places.

adamgoerdt: Looks like crossing-training was already using forage_item, so I set a default value and updated combat-trainer to use the same setting

profiles/base.yaml
```diff
@@ -191,9 +191,8 @@ crossing_training_requires_movement:
 - Thievery
 - Theurgy
 - Trading
-forage_item:
+forage_item: rock
```
rpherbig: Could you also remove the default case from `crossing-training` (line 585)? (`|| 'rock'`)

adamgoerdt: All set

PR #1415, opened 2017-02-27
Astrology Update

jandersson:
I added some command line args for astrology. I think it would be cool to add more features to the script and make them callable via arguments (e.g. `;astrology predict self athletics`). There was an issue with the rtr-expire flag causing the script to never cast RtR.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -25,10 +25,10 @@ class Astrology
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
     do_buffs(settings)
-    if args.train
-      train_astrology
-    elsif args.rtr
+    if args.rtr
       check_ripples
+    else
+      train_astrology
     end
   end
 
--- a/astrology.lic
+++ b/astrology.lic
@@ -11,8 +11,8 @@ class Astrology
   def initialize
     settings = get_settings
     arg_definitions = [
-	    [],
-	    [{name: 'rtr', regex: /rtr/i, description: 'Runs Read the Ripples'}]
+      [],
+      [{ name: 'rtr', regex: /rtr/i, description: 'Runs Read the Ripples' }]
     ]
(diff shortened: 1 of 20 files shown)


common-arcana.lic
```diff
@@ -92,7 +92,7 @@ module DRCA
     when /^Your target pattern dissipates/, /^You can't cast that at yourself/, /^You need to specify a body part to consume/
       fput('release spell')
     when /^At the ritual's peak, your prophetic connection blooms a thousand-fold/
-      Flags.reset('rtr-expire')
+      Flags.add('rtr-expire', get_data('spells').spell_data['Read the Ripples']['expire'])
```
rpherbig: Why is the Flag added here and in `astrology.lic`?

astrology.lic
```diff
@@ -10,6 +10,12 @@ class Astrology
 
   def initialize
     settings = get_settings
+    arg_definitions = [
+	    [{name: 'train', regex: /train/i, description: 'Trains astrology'}],
```
rpherbig: I would expect train to be the default behavior if no args are supplied. What do you think?

jandersson: What is the way to supply a default argument?

jandersson: I thought so too, i couldn't find any way to incorporate a default argument

rpherbig:
```
arg_definitions = [
	    [],
            [{name: 'rtr', ...}]
]
```

common-arcana.lic
jandersson: I was having some issue with the flag being raised. I don't actually know what was causing it. The problem was the RtR expire flag would never be raised when the expire message came. It looks like adding it is the same as resetting it.

jandersson: I will set this back to `reset` and post some logs

jandersson:
OK Disregard the issue mentioned, using `Flags.reset` works fine in common-arcana

PR #1419, opened 2017-02-28
Adding Intercessors

izawwlgood:


data/base-hunting.yaml
```diff
@@ -758,6 +758,13 @@ hunting_zones:
   - 11468
   - 11469
   - 11474
+  # https://elanthipedia.play.net/Dragon_Priest_intercessor
```
rpherbig: Please add the approximate rank range

PR #1424, opened 2017-03-01
Added fail count to Analyze

adamgoerdt:


combat-trainer.lic
```diff
@@ -1313,19 +1316,22 @@ class TrainerProcess
...
     when 'You must be closer'
       fput('engage')
     when 'Your analysis reveals a massive opening'
       return
-    when /You reveal an? .* [weakness|opening]/, /Your analysis reveals an? .* [weakness|opening]/, 'You fail to find any holes'
-      analyze(game_state)
+    when 'You fail to find any holes'
+      analyze(game_state,fail_count+1)
+    when /You reveal an? .* [weakness|opening]/, /Your analysis reveals an? .* [weakness|opening]/
+      analyze(game_state,fail_count)
```
rpherbig: Should `fail_count` reset to 0 when analyze succeeds?

combat-trainer.lic
```diff
@@ -1177,6 +1177,9 @@ class TrainerProcess
 
     @forage_item = settings.forage_item
     echo("  @forage_item: #{@forage_item}") if $debug_mode_ct
+
+    @analyze_retry_count = settings.analyze_retry_count
```
rpherbig: The name of this setting is generic enough that I would expect it to affect all types of analyze. What do you think?

adamgoerdt: There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.  I'm open to suggestions

combat-trainer.lic
adamgoerdt: Good question.  I personally don't have a preference, since increasing the value of the setting would have a pretty similar practical effect.

rpherbig:
Is this change still needed? PR #1418 already added a case for `You fail to find any holes`.

If this is still needed, can you merge master into this PR?

combat-trainer.lic
adamgoerdt: I updated the setting name to combat_analyze_retry_count

adamgoerdt:
Still looking to get this update in.  1418 fixed an infinite loop, but there was some discussion that stopping the analyze loop after a single failure was too fast, so it's now configurable

combat-trainer.lic
```diff
@@ -1313,19 +1316,22 @@ class TrainerProcess
...
     case bput('analyze', 'roundtime', 'Analyze what', 'Your analysis reveals a massive opening', /You reveal an? .* [weakness|opening]/, /Your analysis reveals an? .* [weakness|opening]/, 'You fail to find any holes', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') # unless
     when 'Analyze what'
       fput('face next')
-      analyze(game_state)
+      analyze(game_state,fail_count)
     when 'You must be closer'
       fput('engage')
     when 'Your analysis reveals a massive opening', 'You fail to find any holes'
       return
+    when 'You fail to find any holes'
```
rpherbig: You need to remove `You fail to find any holes` from line 1329.

adamgoerdt: Good catch.  Thanks

PR #1431, opened 2017-03-02
adding assassins

izawwlgood:


data/base-hunting.yaml
```diff
@@ -758,6 +758,12 @@ hunting_zones:
   - 11468
   - 11469
   - 11474
+  # https://elanthipedia.play.net/Dragon_Priest_assassin 750-1100 
+  - 4580
```
rpherbig: You need a name for this hunting zone before the room list (see line 768).

Issue #1352, opened 2017-02-08
Request - EOTB released when hunt is done


Aurayn:
Got kicked out of hunting due to bleeding and it was still up
Hunt finished timer and it was still up

Thanks!

jandersson:
What did having invisiblity mess up? I ask because it might be more helpful to release invisibility in that script where its needed

Aurayn:
Couldn't be healed by Dokt
Couldn't get weapons fixed
Couldn't do banking

rpherbig:
Can you get a log of these?

jandersson:
To add on, I got stuck during forging training due to having invisibility up. There's a `release_invisibility` method in common that might be good to run. It should already be doing it at the bank.

Aurayn:
yessir i will try and get those today for you.

jandersson:
It looks like repair already handles release invisibility. Ill add a PR to catch the banking and some other inviso related issues i noticed

PR #1432, opened 2017-03-03
Tend at autopath

### LINKED ISSUE #1316: Dokt bug - Won't fully heal if bolts/arrows/etc
Figured out the cause of the stacked queue at Dokt. If you have bolts/arrows in your injuries you will stay laying down and he will continuously heal you without deeming you full health. This causes the backlog. We should alter the safe-room script to attempt to tend wounds with bolts/arrows in order to prevent this from clogging up.

skyleronken:
Added a call to tendme once its your turn at the autopath. This fixes #1316.

Also added a call to release_invisibility because of issues people have been ahving with SOV.


safe-room.lic
```diff
@@ -72,6 +72,9 @@ class SafeRoom
   def wait_for_healing
     Flags.reset('healthy')
     Flags.reset('moved')
+    
+    lie_down
```
rpherbig: Do we need to lie down before tending our wounds?

We're already lieing down on line 80.

jandersson:
The release invisibility will partially address #1352

skyleronken:
I put that there in case the person has multiple severe bleeders. Don't want them tending wounds for a few minutes as they bleed out at the autopath. 

We can remove the one in the inner loop, but I though to maintain it since the the lie down function does a check to see if they are standing before issuing the command.

Let me know how that sounds.

safe-room.lic
skyleronken: I put that there in case the person has multiple severe bleeders. Don't want them tending wounds for a few minutes as they bleed out at the autopath.

We can remove the one in the inner loop, but I though to maintain it since the the lie down function does a check to see if they are standing before issuing the command.

Let me know how that sounds.

skyleronken: @rpherbig Want me to change this or are we good to keep it?

safe-room.lic
```diff
@@ -72,6 +72,9 @@ class SafeRoom
...
+    wait_for_script_to_complete('tendme')
```
rpherbig: Does `tendme` ever actually end? The code looks like a loop that just repeatedly tends and never exits.

skyleronken: Yes. It will exit when there are no more bleeders. Line 35 calls "exit if bleeders.empty?"

rpherbig: Gotcha, thank you

mwest152:
Not sure if this is relevant, but noticed it at the autopath.  Maybe missing messaging? 

[tendme]>heal

The surgeon Dokt whistles between his teeth as he rather roughly rubs some ointment on your right hand.
Your right hand tingles for a moment, then suddenly feels a bit better.  Dokt nods with satisfaction.
[135 Kronars are taken from you.]
Roundtime:  2 seconds.
P!R> 
Your body feels at full strength.
You have a broken right hand with gaping holes.

Bleeding
            Area       Rate              
-----------------------------------------
      right hand       moderate

P!R> 
[tendme]>tend my right hand

Your right hand is too injured for you to do that.
P!> 
Dokt rubs the back of his neck then stretches and returns to work.
Dokt rubs a blunt thumb between his eyes and yawns.

P!> 
[tendme: *** No match was found after 15 seconds, dumping info]

[tendme: messages seen length: 13]

[tendme: message: Your right hand is too injured for you to do that.]

[tendme: message: Aura Sight  (1 roisan)]

[tendme: message: Piercing Gaze  (1 roisan)]

[tendme: message: Manifest Force  (31 roisaen)]

[tendme: message: Seer's Sense  (25 roisaen)]

[tendme: message: [DRPrime]-DR:Mitkiahn: "In terms of percentages, that's fantastic"  (14:30:33)]

[tendme: message: Dokt rubs the back of his neck then stretches and returns to work.]

[tendme: message: Dokt rubs a blunt thumb between his eyes and yawns.]

[tendme: message: [DRPrime]-DR:Mitkiahn: "That's 50%.. 30 plat to 45 plat"  (14:30:38)]

[tendme: message: Aura Sight  (1 roisan)]

[tendme: message: Piercing Gaze  (1 roisan)]

[tendme: message: Manifest Force  (31 roisaen)]

[tendme: message: Seer's Sense  (25 roisaen)]

[tendme: checked against [/You work carefully at tending/i, /That area has already been tended to/i, /That area is not bleeding/i, /You fumble/i, /You \w+ remove/i]]

rcuhljr:
Yeah tendme is hanging with skin wounds as well which causes safe-room to never finish. http://pastebin.com/63xEUGan

PR #1435, opened 2017-03-04
Astrology: closed telescopes, rtr fix, predict event setting

jandersson:
I moved the rtr flag out of astrology. If  a `common-arcana` cast detects the success message for rtr then the flag is added which also functions as a reset. This should be the only usage of the flag. Previously, common-arcana reset a flag which had never been set, so the 'match' was never given. This means the flag was never raised. The check for the RTR flag during astrology buffs was faulty, so RTR was never cast. It was really bad, and im sorry about that.

I made a setting for event prediction. Event prediction is not useful for high level moon mages and takes a lot of time to do, so turning it off is a great option for high level mages. Another argument for the setting is, it's a learned skill; a newbie mage cannot utilize it until taught by another mage. The default is off.

There was a minor bug with event prediction where predict event loops unless the prediction pool has filled since the previous observation. This is done to prevent endless observations. The check didn't consider if the pool was full already, so this resulted in a redundant observation and pool check (15+ rt). Now, a full event pool results in breaking out of the event prediction loop.

If a closed telescope was centered on a celestial body, one got an error message. Now the telescope will be opened after getting that error message and retry the observation.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -24,7 +24,7 @@ class Astrology
     @ap_training = settings.astral_plane_training['train_in_ap']
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
-		@predict_event = settings.predict_event
+    @predict_event = settings.predict_event
     do_buffs(settings)
     if args.rtr
       check_ripples
@@ -222,9 +222,9 @@ class Astrology
 
     if @have_telescope
       bput('get telescope', 'you get', 'You are already', "That can't be picked up")
-			if bput("center telescope on #{body}", 'You put your eye', 'open it to make any use of it').match('open it')
-      bput('open my telescope', 'extend your telescope')
-			return observe(body)
+      if bput("center telescope on #{body}", 'You put your eye', 'open it to make any use of it').match('open it')
+        bput('open my telescope', 'extend your telescope')
+        return observe(body)
       end
       result = bput('peer telescope', 'You peer aimlessly through your telescope', 'You see nothing regarding the future', 'Roundtime')
       bput('stow telescope', 'you put', "That can't be picked up")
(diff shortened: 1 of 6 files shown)


jandersson:
I'm really sorry about the multiple PR's to address this flag issue. I am currently running this for 2 days without requiring any tweaking,  though I have said this in the past; changes were poorly tested.

jandersson:
Give me a few days to run this! I will report back when all is gravy

jandersson:
Reporting no issues

astrology.lic
```diff
@@ -219,7 +218,10 @@ class Astrology
 
     if @have_telescope
       bput('get telescope', 'you get', 'You are already', "That can't be picked up")
-      bput("center telescope on #{body}", 'You put your eye')
+      if bput("center telescope on #{body}", 'You put your eye', 'open it to make any use of it').match('open it')
```
jandersson: Rubocop warns me about this line, should I use =~ instead? I dont see any issues, beyond style

astrology.lic
```diff
@@ -265,7 +267,7 @@ class Astrology
     check_ripples
     check_astral
     check_weather
-    check_events(check_pools)
+    check_events(check_pools) if @predict_event
```
jandersson: I considered doing the check as a guard in `check_events` but i wanted to leave open the possibility that you could add a command arg to the script for predicting events and be independent of this training setting.

astrology.lic
```diff
@@ -101,6 +102,7 @@ class Astrology
       return if ['detect any portents'].include? result
       pools = check_pools
       break if pools['future events'] == prev_size
+      break if pools['future events'] == 10
```
jandersson: It could be more useful to have a `max?` function for a Pools class. Potential refactoring :]

PR #1443, opened 2017-03-07
Fixed issue with wearing footwear on the ice road

adamgoerdt:


rpherbig:
I think there's a risk that if someone is cycling armors for training, they may not know which foot armor is on and therefore not have a reliable keyword to use there. @rcuhljr thoughts?

An alternative could be to wear a `skating` gear set which contains `ice skates`. Thoughts?

adamgoerdt:
Foot armor isn't the problem. I wear footwraps and skates with no issue.
The problem is with foot fluff like boots or sandals, and that shouldn't
change without the person knowing

PR #1447, opened 2017-03-08
Kill counter

### LINKED ISSUE #1444: Kill Counter
I envision something that when activated simply scores the number of times the script enters LOOT. It can store this as a variable that can be recalled. If it can include a time stamp, ossim. 

Something like ;kill-counter start
Let it run for 20m
;kill-counter stop
;kill-counter list

**You killed 4 billion critters! 
**Started 4:00
**Finished 4:20

If the timestamp cant be added, that's also fine.

skyleronken:
added kill-counter script. Decided to do it like a monitor instead of a command line script since it would be easier to reference and is meant to span multiple hunts possibly. resolves #1444.


kill-counter.lic
```diff
@@ -0,0 +1,99 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#kill-counter
```
rpherbig: Don't forget to add this documentation after (or before!) the PR is merged :)

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+=end
+
+unless HAVE_GTK
+  respond
+  respond 'error: ruby-gtk bindings are not installed or failed to load'
+  respond
+  exit
+end
+
+custom_require.call(%w(textsubs))
```
rpherbig: Why does this require `textsubs`?

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+
+class KillCounter
+  def initialize
+    @kill_count = 0
+  end
+
+  def process_line(line)
```
rpherbig: Convention is to name predicate functions with a `?` (e.g. `process_line?`).

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+  respond 'error: ruby-gtk bindings are not installed or failed to load'
+  respond
+  exit
+end
+
+custom_require.call(%w(textsubs))
+
+class KillCounter
+  def initialize
+    @kill_count = 0
```
rpherbig: `@kill_count` is never used. Should the loop below be tracking the count or should it be done here?

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#kill-counter
+=end
+
+unless HAVE_GTK
```
rpherbig: Why does this require GTK (instead of outputting to the game window)?

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+  respond
+  respond 'error: ruby-gtk bindings are not installed or failed to load'
```
rpherbig: Have you considered a `nowindow` option, similar to how `status-monitor` works? Some users have problems with GTK.

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+    if /^You search (?!around for a moment)/i =~ line
+      true
+    else
+      false
+    end
+  end
+
+end
+
+filter = KillCounter.new
```
rpherbig: If `KillCounter` is being used only as a filter (and only filtering on *looting*, not killling), then there might be a better name for the class. Maybe `LootDetector`?

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+
+custom_require.call(%w(textsubs))
+
+class KillCounter
+  def initialize
+    @kill_count = 0
+  end
+
+  def process_line(line)
+    if /^You search (?!around for a moment)/i =~ line
```
rpherbig: This could be a one-liner: `return /^You search (?!around for a moment)/i =~ line` (and technically `return` isn't necessary in that case).

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+    time_obj = Time.now 
+    start_time.text = "Start Time: #{time_obj.inspect}"
+    vbox.pack_start(start_time)
+    
+    stats_et = Gtk::Entry.new
+    stats_et.editable = false
+    display_font = Pango::FontDescription.new
+    display_font.weight = Pango::FontDescription::WEIGHT_BOLD
+    stats_et.modify_font(display_font)
+    k_counter_disp = stats_et
```
rpherbig: Why do this assignment, versus renaming `stats_et` to `k_counter_disp`?

kill-counter.lic
```diff
@@ -0,0 +1,99 @@
...
+      window_done = true
+    end
+    window.show_all
+  end
+
+  loop do
+    line = script.gets?
+    break if window_done
+    pause 0.05 unless line
+    next unless line
```
rpherbig: Is line 93 redundant with line 92?

kill-counter.lic
skyleronken: done

kill-counter.lic
skyleronken: done

kill-counter.lic
skyleronken: fixed

kill-counter.lic
skyleronken: i used performance-monitor as a template. However, based upon the use case it is desired to be something that could be called by command line and then send a signal to get whatever the current count is.

kill-counter.lic
skyleronken: I have not. I am not that familiar with GTK.

kill-counter.lic
skyleronken: Previously I was checking based on the regex \balanced?]\i. But that wasn't catching kills from spells. So our method of detection may change int he future, but the intent is to detect kills.

kill-counter.lic
skyleronken: done

kill-counter.lic
skyleronken: done

kill-counter.lic
skyleronken: not sure. I took this directly from performance-monitor.

PR #1453, opened 2017-03-09
Improved autopath tend

skyleronken:
wait_for_script was causing issues because of bugs in tendme where certain wounds couldn't be tended, etc. This was causing people to never exit the wait_for_healing loop even after being healed. I changed the call to 'tendme' to call the script async with start_script. Then, after being healed, call stop_script if it was still being ran.


skyleronken:
additional details on the bug: It appears that the bug only occurred when the following conditions were met:
1) They had bleeders when they entered the autopath room
2) Nobody else was waiting to be healed

If they had to wait in line to be healed, the bug did not occur (based on reporting)

skyleronken:
additional note: this fix should make it so that errors with tendme will not screw up safe-room.

rpherbig:
Would you be up for fixing `tendme` in this PR?

skyleronken:
Yeah I can do that. Might take me a day to get to it.

skyleronken:
Okay. Made two additional changes:
1) filter out skin bleeders
2) catch messaging for when your too wounded to tend

was there anything else?

Issue #1296, opened 2017-01-15
crossing-training: appraise full gem pouches if available


skyleronken:
Suggest training appraisal but appraising full gem pouches in a given container if they exist. If none are present, then fall back on equipment, or possibly just do both.

jandersson:
Can probably close :)

rpherbig:
ff088925bf531284b917d08360e60a9ef1fa929b

PR #1454, opened 2017-03-09
Appraisal with pouches

skyleronken:
train appraisal using gem pouches. requires a boolean flag in YAML as well as a defined container where the full pouches are stored. Will loop through those (to a max equivilent to the ordinals we support in common) until nearly locked. Any configuration errors should fall through to whatever the remaining appraisal training definition is but echo a warning.


skyleronken:
This relates to #1296

appraisal.lic
```diff
@@ -25,6 +26,42 @@ class Appraisal
     @equipment_manager.items.each { |item| break unless appraise_item?(item) }
     true
   end
+  
+  def train_appraisal_with_pouches?
+    settings = get_settings
+    return false unless settings.train_appraisal_with_pouches
+    full_pouch_container = settings.full_pouch_container
+    if !full_pouch_container
```
rcuhljr: Nitpicky, but you can use unless here.

appraisal.lic
```diff
@@ -25,6 +26,42 @@ class Appraisal
...
+      echo "Must set full_pouch_container in YAML"
+      return false
+    end      
+    if full_pouch_container == settings.spare_gem_pouch_container
```
rcuhljr: Honestly I'd rather have checks like these two be added to ;validate, have you looked at that script?

appraisal.lic
```diff
@@ -25,6 +26,42 @@ class Appraisal
...
+      case bput("get #{$ORDINALS[i]} #{p} from my #{full_pouch_container}", '^You get ', '^What were you referring')
+      when /^You get /
+        bput("appraise #{p} quick", 'Roundtime')
+        waitrt?
+        fput("put my #{p} in my #{full_pouch_container}")
+      else
+        return true
+      end
+      
+      if DRSkill.getxp('Appraisal') >= 32
```
rcuhljr: Since you end the method with a true, you can change all of these returns into breaks and combine them. something like
```
break if DRSkill.getxp('Appraisal') >= 32
```

appraisal.lic
```diff
@@ -25,6 +26,42 @@ class Appraisal
...
+      echo "Must set full_pouch_container in YAML"
+      return false
+    end      
+    if full_pouch_container == settings.spare_gem_pouch_container
+      echo "*** Cannot have full pouches be stored where spare pouches are ***"
+      return false
+    end
+    
+    pouches = rummage('C pouch', full_pouch_container)
+    pouches.each_with_index do |p, i|
```
rcuhljr: Since you're ending your iteration on the size of ordinals, and only use your index to look up ordinals, have you considered just looping over ordinals and trying to get/app each ordinal, and just breaking on mind lock or no pouch found for that ordinal?

profiles/base.yaml
```diff
@@ -215,6 +215,8 @@ symbiosis_learning_threshold: 2
 compost_room: 991
 mine_for_outdoorsmanship: false
 train_appraisal_with_gear: false
+train_appraisal_with_pouches: false
+full_pouch_container:
```
rcuhljr: Forward looking thought, we could have the places (lockpick and combat-trainer) that cycle pouches not use stow on full pouches and instead use this variable when available.

appraisal.lic
skyleronken: I added these checks with explanations so that if poorly configured the appraisal script would fall through to the other configured appraisal training configs. I can remove them if you think thats more appropriate.

appraisal.lic
skyleronken: I guess that the noun will always be 'pouch' so that should work

appraisal.lic
rcuhljr: Well I'd just really like to have them in validate. These are errors we can detect before the script runs, and frankly most people never notice error messages when scripts run and just say "It didn't work."

profiles/base.yaml
skyleronken: done

appraisal.lic
```diff
@@ -25,6 +26,37 @@ class Appraisal
...
+      echo "Must set full_pouch_container in YAML"
+      return false
+    end      
+    if full_pouch_container == settings.spare_gem_pouch_container
+      echo "*** Cannot have full pouches be stored where spare pouches are ***"
+      return false
+    end
+    
+    pouches = rummage('C pouch', full_pouch_container)
+    $ORDINALS.each_with_index do |p, i|
```
rcuhljr: You won't need to rummage anymore, and you can just switch this to `$ORDINALS.each{ |ordinal|` and change `p` to a more meaningful `ordinal`.

skyleronken: Makes sense! done!

skyleronken:
Everything should be good to go now.

rcuhljr:
Add a test or two to validate and I'll merge this :)

appraisal.lic
```diff
@@ -25,6 +26,36 @@ class Appraisal
...
+      return false
+    end
+    
+    $ORDINALS.each do |ordinal|
+      
+      case bput("get #{ordinal} pouch from my #{full_pouch_container}", '^You get ', '^What were you referring')
+      when /^You get /
+        bput("appraise my pouch quick", 'Roundtime')
+        waitrt?
+        fput("put my pouch in my #{full_pouch_container}")
```
rpherbig: Would you mind changing this to use `bput`?

appraisal.lic
```diff
@@ -25,6 +26,36 @@ class Appraisal
     @equipment_manager.items.each { |item| break unless appraise_item?(item) }
     true
   end
+  
+  def train_appraisal_with_pouches?
+    settings = get_settings
```
rpherbig: It might be cleaner to move `get_settings` out into `initialize`

rpherbig: And to make the same change to `train_appraisal_with_gear?`

appraisal.lic
```diff
@@ -25,6 +26,36 @@ class Appraisal
...
+        waitrt?
+        fput("put my pouch in my #{full_pouch_container}")
+      else
+        break
+      end
+      
+      break if DRSkill.getxp('Appraisal') >= 32
+      
+    end
+    true
```
rpherbig: Returning `true` here means that we will only attempt to appraise pouches, even if, we run out and are not yet mindlocked. Would it be preferable to instead continue appraising gear/etc. if Appraisal isn't yet locked?

appraisal.lic
```diff
@@ -25,6 +26,36 @@ class Appraisal
...
+      case bput("get #{ordinal} pouch from my #{full_pouch_container}", '^You get ', '^What were you referring')
+      when /^You get /
+        bput("appraise my pouch quick", 'Roundtime')
+        waitrt?
+        fput("put my pouch in my #{full_pouch_container}")
+      else
+        break
+      end
+      
+      break if DRSkill.getxp('Appraisal') >= 32
```
rpherbig: Why is the 'breakpoint' 32 here but 30 on line 65?

appraisal.lic
skyleronken: done

appraisal.lic
skyleronken: I figured if you have the gem pouches, why not attempt to train higher. I can change this to 30.

appraisal.lic
skyleronken: I was thinking about that. lets try that out and see if a different use case is needed. changes made.

Issue #938, opened 2016-10-16
Remedy/Alchemy/Workorder 


Sagadin:
Seems to still be problem with combining, marking, and tieing nemoih salve.  I cleaned my bags completely of extra odd stuff left over and this still occurred.

```
[workorders]>bun my salve with logbook
You notice the workorder calls for stacks of 5 for each remedy, and think it best to mark and cut the remedy down to the required size before bundling.
>
[workorders]>put my alchemy logbook in my toolbag
You put your logbook in your leather toolbag.
>
[workorders]>Mark my salve at 5
There is not enough remedy material present to do that.
>;chat At 250 ranks they appraised as creatures of immense power
[workorders: *** No match was found after 15 seconds, dumping info]
[workorders: messages seen length: 2]
[workorders: message: There is not enough remedy material present to do that.]
[workorders: message: [DRPrime]-You: "At 250 ranks they appraised as creatures of immense power"]
[workorders: checked against [/You measure/i]]
[workorders]>Break my salve
You can't break that.
>
[workorders: *** No match was found after 15 seconds, dumping info]
[workorders: messages seen length: 3]
[workorders: message: You can't break that.]
[workorders: message: [DRPrime]-DR:Feltzer: "wow"]
[workorders: message: [Leucius] "still a batch at the Empath Guild"]
[workorders: checked against [/You carefully/i]]
[workorders]>stow left
You just can't make that stack any larger.
>
[workorders: *** No match was found after 15 seconds, dumping info]
[workorders: messages seen length: 2]
[workorders: message: You just can't make that stack any larger.]
[workorders: message: [DRPrime]-DR:Qetu: "they didn't outright kill me, took them about 6 hits to drop me below half, but they don't miss and hit hard enough"]
[workorders: checked against [/You put/i]]
[workorders]>get my alchemy logbook
You get an alchemy work order logbook from inside your leather toolbag.
>
[workorders]>bundle my salve with my logbook
You notice the workorder calls for stacks of 5 for each remedy, and think it best to mark and cut the remedy down to the required size before bundling.
>
[workorders: *** No match was found after 15 seconds, dumping info]
[workorders: messages seen length: 2]
[workorders: message: You notice the workorder calls for stacks of 5 for each remedy, and think it best to mark and cut the remedy down to the required size before bundling.]
[workorders: message: [DRPrime]-DR:Feltzer: "guess the little ones were just a tease"]
[...]

Issue #872, opened 2016-09-30
Alchemy-Workorder Messaging Error


Sagadin:
This appears to be a different error than the other open error.  The script does not recognize certain combine messaging:

```
[workorders]>stow left
You combine the stacks of herbs together, but some was left over.
>
Your teacher is not here, so you are no longer learning anything.
>
You feel fully rested.
>
[workorders: *** No match was found after 15 seconds, dumping info]
[workorders: messages seen length: 5]
[workorders: message: You combine the stacks of herbs together, but some was left over.]
[workorders: message: Manifest Force  (10 roisaen)]
[workorders: message: Wisdom of the Pack  (11 roisaen)]
[workorders: message: Your teacher is not here, so you are no longer learning anything.]
[workorders: message: You feel fully rested.]
[workorders: checked against [/You put/i, /The braided (.+) is too long/]]
```

And also an error when trying to put large groups of herbs in the stow container (herb pouches just aren't all that big I'm afraid, and I don't want to reset my stow for herb products....)

```
[remedy]>stow left
That stack of herbs is too large to add more to.
>
[remedy: *** No match was found after 15 seconds, dumping info]
[remedy: messages seen length: 5]
[remedy: message: That stack of herbs is too large to add more to.]
[remedy: message: Manifest Force  (10 roisaen)]
[remedy: message: Wisdom of the Pack  (11 roisaen)]
[remedy: message: Manifest Force  (10 roisaen)]
[remedy: message: Wisdom of the Pack  (11 roisaen)]
[remedy: checked against [/You put/i, /The braided (.+) is too long/]]
```

Sagadin:
A couple more errors; it seems the script wasn't able to get it to properly make the correct size stack for the work order

workorders]>bun my salve with logbook
You notice the workorder calls for stacks of 5 for each remedy, and think it best to mark and cut the remedy down to the required size before bundling.
>
[workorders]>put my alchemy logbook in my toolbag
You put your logbook in your leather toolbag.
>
[workorders]>Mark my salve at 5
There is not enough remedy material present to do that.
>
[workorders: *** No match was found after 15 seconds, dumping info]
[workorders: messages seen length: 1]
[workorders: message: There is not enough remedy material present to do that.]
[workorders: checked against [/You measure/i]]
[workorders]>Break my salve
You can't break that.
>
[workorders: *** No match was found after 15 seconds, dumping info]
[workorders: messages seen length: 1]
[workorders: message: You can't break that.]
[...]

Issue #838, opened 2016-09-21
Error in alchemy workorder (remedy/remedies): 


rpherbig:
@Dartellum This is after merging in #837.

```
2016-09-21 09:49:39 -0400:[workorders]>tap my first red flowers in my backpack
2016-09-21 09:49:40 -0400:You tap yourself on the chest.
2016-09-21 09:49:40 -0400:[workorders]>tap my first red flowers in my backpack
2016-09-21 09:49:40 -0400:You tap yourself on the chest.
2016-09-21 09:49:40 -0400:[workorders]>order 13
2016-09-21 09:49:40 -0400:The attendant says, "You can purchase (25 pieces) dried red flowers for 343 Kronars.  Just order it again and we'll see it done!"
2016-09-21 09:49:40 -0400:[workorders]>order 13
2016-09-21 09:49:41 -0400:The attendant takes some coins from you and hands you (25 pieces) dried red flowers.
2016-09-21 09:49:42 -0400:[workorders]>put my red flowers in my backpack
2016-09-21 09:49:42 -0400:You put your flowers in your backpack.
2016-09-21 09:49:42 -0400:[workorders]>order 13
2016-09-21 09:49:42 -0400:[workorders]>order 13
2016-09-21 09:49:43 -0400:The attendant says, "You can purchase (25 pieces) dried red flowers for 343 Kronars.  Just order it again and we'll see it done!"
2016-09-21 09:49:43 -0400:The attendant takes some coins from you and hands you (25 pieces) dried red flowers.
2016-09-21 09:49:44 -0400:[workorders]>put my red flowers in my backpack
2016-09-21 09:49:44 -0400:You put your flowers in your backpack.
2016-09-21 09:49:44 -0400:[workorders]>order 13
2016-09-21 09:49:45 -0400:The attendant says, "You can purchase (25 pieces) dried red flowers for 343 Kronars.  Just order it again and we'll see it done!"
2016-09-21 09:49:45 -0400:[workorders]>order 13
2016-09-21 09:49:45 -0400:The attendant takes some coins from you and hands you (25 pieces) dried red flowers.
2016-09-21 09:49:46 -0400:[workorders]>put my red flowers in my backpack
2016-09-21 09:49:46 -0400:You put your flowers in your backpack.
2016-09-21 09:49:47 -0400:[workorders]>tap my first nemoih in my backpack
2016-09-21 09:49:47 -0400:You tap some dried nemoih inside your backpack.
[...]

PR #1458, opened 2017-03-10
Revamp to Workorders for Alchemy.

m-burroughs:
Revamp to Workorders for Alchemy.  New counting/combining code, along with new code to calculate volumes of catalyst and water.

Additionally, this addresses issues reported:
https://github.com/rpherbig/dr-scripts/issues/838
https://github.com/rpherbig/dr-scripts/issues/872
https://github.com/rpherbig/dr-scripts/issues/938

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -332,51 +332,80 @@ class WorkOrders
     end
   end
 
-  def count_combine_rem(stock_room, herb, herb_stock, quantity, herb2)
-    if inside?("first #{herb}", @bag) && inside?("second #{herb}", @bag)
-      bput("get #{herb} from my #{@bag}", 'You get')
-      bput("get #{herb} from my #{@bag}", 'You get')
-      case bput('combine', 'You combine', 'That stack of herbs')
-      when 'That stack of herbs'
-        stow_tool(left_hand)
-        stow_tool(right_hand)
-      when 'You combine'
-        case bput("get #{herb} from my #{@bag}", 'You get', 'What were')
-        when 'You get'
-          case bput('combine', 'You combine', 'That stack of herbs')
-          when 'That stack of herbs', 'You combine'
-            stow_tool(left_hand)
-            stow_tool(right_hand)
-          end
+  # Function which will attempt to combine any stacks of herbs in your inventory, will count them, and determine if more need to be ordered.
+  def count_combine_rem(stock_room, quantity, herb, herb_stock)
+  # Initialize variables
+    found_stack = true;
+    herb_volume_total = 0;
+    last_herb_volume = 0;
+    last_descriptor = "";
+    stack_descriptor = "";
+
+    # We have to only ever use the last word in a multi-word herb.  We will need to be careful when counting because of this.
+    # This is to resolve "red flower" vs "blue flower" because "count third red flower in my bag" fails.
+    # We must instead do a generic "tap third flower in my bag" and determine if we're tapping the right kind of flowers from the response, then actually count them.
+    herb_for_tapping = (herb.gsub(/\s+/m, ' ').strip.split(" ")).last
+
+    # Calculate the volume of the herbs that we will be searching our inventory for and/or buying.
+    # Assume that each unit we need to craft will take 25 volumes of our herb.
+    # Also assume that we also always need double the supply that the workorder would normally take in the event that some of our product is not high enough quality.
+    need_herb_volume = ((quantity * 25.0) * 2)
+
+    # Loop through all the stacks of herbs we might have in our backpack and find out how much volume we have total.
+    descriptors = [ "first", "second", "third", "fourth", "fifth", "sixth", "seventh", "eighth", "ninth", "tenth", "eleventh", "twelfth" ]
+    begin
+      stack_descriptor = descriptors.shift
+
+      # TODO: Solve the Inkin problem, which (very rarely) results in never counting any herbs, and always buying extra.
+      # The "You lightly tap" result here is what happens if you try to do this in a room with somebody having a name that start with "In" .... like... "Inkin"
+      # example:
+      #   [workorders]>tap first flower in my haversack
+      #   You lightly tap Inkin on the shoulder.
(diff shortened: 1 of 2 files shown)


workorders.lic
```diff
@@ -332,51 +332,80 @@ class WorkOrders
...
+    found_stack = true;
+    herb_volume_total = 0;
+    last_herb_volume = 0;
+    last_descriptor = "";
+    stack_descriptor = "";
+
+    # We have to only ever use the last word in a multi-word herb.  We will need to be careful when counting because of this.
+    # This is to resolve "red flower" vs "blue flower" because "count third red flower in my bag" fails.
+    # We must instead do a generic "tap third flower in my bag" and determine if we're tapping the right kind of flowers from the response, then actually count them.
+    herb_for_tapping = (herb.gsub(/\s+/m, ' ').strip.split(" ")).last
```
rpherbig: Would it be easier to, in the recipe YAML, split herb data into `adjective` and `noun`?

workorders.lic
```diff
@@ -332,51 +332,80 @@ class WorkOrders
...
+
+    # Calculate the volume of the herbs that we will be searching our inventory for and/or buying.
+    # Assume that each unit we need to craft will take 25 volumes of our herb.
+    # Also assume that we also always need double the supply that the workorder would normally take in the event that some of our product is not high enough quality.
```
rpherbig: For other crafts, quality is not something we've worried about. If you can't MC recipes at your rank, there's a setting that lets you ask for easier workorders.

Is alchemy/remedies different with respect to quality?

workorders.lic
```diff
@@ -332,51 +332,80 @@ class WorkOrders
...
+    need_herb_volume = ((quantity * 25.0) * 2)
+
+    # Loop through all the stacks of herbs we might have in our backpack and find out how much volume we have total.
+    descriptors = [ "first", "second", "third", "fourth", "fifth", "sixth", "seventh", "eighth", "ninth", "tenth", "eleventh", "twelfth" ]
```
rpherbig: I recommend using `$ORDINALS` in `common`.

workorders.lic
```diff
@@ -332,51 +332,80 @@ class WorkOrders
...
+    begin
+      stack_descriptor = descriptors.shift
+
+      # TODO: Solve the Inkin problem, which (very rarely) results in never counting any herbs, and always buying extra.
+      # The "You lightly tap" result here is what happens if you try to do this in a room with somebody having a name that start with "In" .... like... "Inkin"
+      # example:
+      #   [workorders]>tap first flower in my haversack
+      #   You lightly tap Inkin on the shoulder.
+      /You tap (.*) inside your|I could not find|You lightly tap/ =~ bput("tap #{stack_descriptor} #{herb_for_tapping} in my #{@bag}", 'You tap (.*) inside your', 'I could not find', 'You lightly tap')
```
rpherbig: I recommend using `inside?` in `common-items`.

workorders.lic
m-burroughs: I don't believe we should break out the `adjective` and `noun` in the recipe YAML for 2 reasons.
1. Because it doesn't help solve the problem that I outline in the notes here.  We would still need to perform the tapping before the counting because of how DR fails to recognize the name of the herb in the first place when it's inside a container.  There would be nothing gained in terms of reducing logic.
2. The only place we need the `adjective` and the `noun` split is inside the `count_combine_rem` function.  Looking in the `remedy_items` function, which serves as the core logic here, we can see that we actually need the combined herb name in several places when we call in to the `;remedy` script.  This means that in most places, the whole herb name is the useful keyword, and only in one small place do we need the herb split apart.  Additionally, I would argue since so few of the herbs actually require this logic to function, it seems unnecessary to make such a low level change just to accommodate.

workorders.lic
m-burroughs: For remedy workorders, the quality any item added to the bun is require to be of superior quality or better.  In my experience, even items that I can normally mastercraft on a regular basis do on occasion end up with imperfections.  This seems to be an unavoidable randomization that happens sometimes.  If you get enough bad RNGs in a row, even simple recipes can sometimes be less than the required quality.  That's why the `;workorders` script checks for verbage when attempting to bun, and will discard any product that wasn't quality enough.  It's WAY overkill to order double of all materials to account for imperfections, but that's easier than trying to go re-buy materials half way through a workorder.

workorders.lic
m-burroughs: Fixed.

workorders.lic
m-burroughs: That seems like the right choice, but unfortunately it does not work here.
If you look at the logic, we are tricking DR in to giving us information we didn't ask for.  This is not a simple "see if I have this thing in this bag"

Detail:
`> tap third flower in my haversack`

Note that we are using the generic noun here.  This is because if we try to `>tap third blue flower in my haversack` it has no idea what we are doing.
However.

The response to:
`> tap third flower in my haversack`
is:
`You tap some dried red flowers inside your spidersilk haversack.`

DR has given us additional information back about what the type of flowers we just tapped are, and we need to have that text available for regular expression logic.

This is the engine by which we are able to separate and count red flowers differently than blue flowers.  It's basically a workaround because DR is stupid.

workorders.lic
m-burroughs: I am still using a local variable though, because my strategy here is to `shift` and `unshift` off the local variable as appropriate because:
A. I need to keep track of the last ordinal word for the matching herb, which may be MORE than one behind where the loop is.  This is the case if you have 5 red flowers, and a blue flower stuck in the middle.
B. When we combine for convenience, I need to run the same ordinal twice (hence the `unshift`).  This is detailed in the notes I left in the code.

m-burroughs:
Please re-review my changes, thank you.

workorders.lic
rpherbig: > It's basically a workaround because DR is stupid.

If I had a nickel for every time I said that...

rpherbig:
Thank you for submitting this. I know it took a lot of work to get right and we really appreciate it!

PR #1460, opened 2017-03-10
Additional invisibility release usage 

jandersson:
Should take care of #1352

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -543,7 +543,7 @@ class CrossingTraining
     loop do
       waitrt?
       check_research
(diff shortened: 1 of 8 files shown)


crossing-training.lic
```diff
@@ -558,6 +558,8 @@ class CrossingTraining
         dispose_trash(item)
         wait_for_script_to_complete('safe-room')
         return
+      when 'But that would give away your hiding place'
```
jandersson: This is actually redundant i just realized. Already added this to bput itself

jandersson:
Done making changes. I'll chime in if anything else needs doing

sell-loot.lic
```diff
@@ -57,6 +57,7 @@ class SellLoot
 
   def exchange_coins
     walk_to @hometown['exchange']['id']
+    release_invisibility
```
rpherbig: Would it be clearer to do this on line 34 instead of in two places?

Issue #1425, opened 2017-03-01
Add power perceive and perceive self timers to crossing-training


brycole:
-Add timers to crossing-training to perceive every 60 seconds (or more)
-secondary: also add toggle for perceive self (which trains attunement at 60 second timer as well)

MahtraDR:
@rpherbig can close.

PR #1463, opened 2017-03-11
Ct attunement timer

skyleronken:
Train attunement by interval during crossing training. will maintain a timer based on yaml config (default 60 seconds). Will do a perc, and optionally a perc self (yaml boolean, default false). Only does this logic if attunement is not ML AND it is configured as a CT skill.

Note: will not execute exactly at interval. Depends on the event_loop logic. So will execute at 60 seconds or longer.


skyleronken:
resolves #1425

crossing-training.lic
```diff
@@ -35,6 +35,7 @@ class CrossingTraining
     @swimming_target = get_data('athletics').swimming_options[@settings.swimming_target]
     @song_list = get_data('perform').perform_options
 
+    @attunement_timer = Time.now - @settings.attunement_interval
```
rpherbig: Why does this not use the existing exp timers?

crossing-training.lic
```diff
@@ -215,6 +218,17 @@ class CrossingTraining
     @cyclic_cycle_timer = Time.now
   end
 
+  def check_attunement
+    return unless @settings.crossing_training.include?('Attunement')
+    return unless DRSkill.getxp('Attunment') <= 33
+    return unless Time.now - @attunement_timer > @settings.attunement_interval
+    fput('perc')
```
rpherbig: Is it possible to use `bput` here instead? (and line 227)

rpherbig:
Why is Attunement being treated as a special case, rather than using the existing training logic and updating the `attunement` script with any changes?

crossing-training.lic
skyleronken: I can. However, from my observations it looked like bput was used when you needed to check the output of the command whereas fput is used when checking the result was not necessary. Is that not the case?

crossing-training.lic
skyleronken: From my understanding of the feature request, this would be in addition to the power walking training that is normally done by crossing-training. Adding the 60 second timer to the normal experience timers would remove the ability to maintain attunement training while delaying the frequency of power walking.

skyleronken:
The current Attunement script is really a power walking script. However, we should be able to train attunement (albeit at a slower pace) without power walking. This provides that functionality.

rpherbig:
Lichbot does something similar, so it might be good for this to be somewhere common...

What do you think about the existing `attunement` script knowing whether it should perc your room or powerwalk?

skyleronken:
essentially have a 'stationary' argument? that could work. I can make those modifications. Any issue with adding the optional 'perc self' to attunement.lic?

rpherbig:
Sounds good to me. And if all of that is contained in `attunement.lic`, then could you also change Lichbot to call that script? (do any other scripts `perceive`?)

skyleronken:
haven't messed with lichbot but should be simple enough.

skyleronken:
@rpherbig Let me know if thats more inline with what you were envisioning.

attunement.lic
```diff
@@ -24,13 +33,19 @@ class Attunement
       return
     end
 
-    room_list = if @stationary_skills_only || DRStats.necromancer?
+    room_list = if @stationary_skills_only || DRStats.necromancer? || @stationary
```
jandersson: Does this prevent necromancers from power walking?

skyleronken:
Yup! Not sure why it's there. Assumed some necro tactic!

jandersson:
I'd like to be able to train it, but maybe thats outside of the scope of this PR

skyleronken:
Can we verify why it is there in the first place?

jandersson:
I'm guessing its a placeholder for a setting - You'll get SO if you run it and it takes you into town (though you can easily fix this by casting invisibility).

jandersson:
I'll make an issue regarding it

skyleronken:
Any other issues with this?

crossing-training.lic
```diff
@@ -215,6 +218,14 @@ class CrossingTraining
     @cyclic_cycle_timer = Time.now
   end
 
+  def check_attunement
+    return unless @settings.crossing_training.include?('Attunement')
+    return unless DRSkill.getxp('Attunment') <= 33
+    return unless Time.now - @attunement_timer > 60
```
rcuhljr: It seems like attunement timer should probably use the exp_timers setting?

crossing-training.lic
```diff
@@ -103,6 +104,8 @@ class CrossingTraining
 
     check_tithe
 
+    check_attunement
```
rcuhljr: Could there be a way to opt out of this? I'm not really sure what the advantage of it would be if you already have power walking as part of your training.

skyleronken: I could add a boolean check. However, I imagine that if you want to train Attunemenet with crossing-training than you wouldn't mind a few second RT every so often. Especially if you wanted to limit the frequency of power walking or leaving the room.

rpherbig:
Why is it preferable to make `stationary` an arg to the attunement script, rather than a YAML setting?

How does this arg/setting differ from the existing `stationary_skills_only` setting?

crossing-training.lic
skyleronken: having it be exterior to exp_timers allows users to restrict the frequency at which they power walk, yet maintain some training in attunement while training other skills.

skyleronken:
@rpherbig 
I made it an argument because we may not want an either-or functionality from the attunement script. Being able to call it as an argument means that scripts can be written to utilize attunement.lic in different ways in different situations. Oh, and it was your idea (see above)!

I suppose I could make a boolean check to only train it in this manner if stationary_skills_only is set. But again, I think if you are training Attunement at all you wouldn't mind the few seconds roundtime every so often. 

The timer is different so that users can restrict the frequency of which they power walk while still maintain some training of attunement. 

I made the change so that necros can opt-in or opt-out of that check. default to stationary for necros.

skyleronken:
was there anything else you wanted changed for sure?

intangibles:
Couldn't you piggy back off moon mage attunement scripting?   It's essentially the same thing (not tied to a room) except on a shorter timer, uses perc, and perc self instead of "perc moons, planets, etc."

skyleronken:
That's what I have done. Not sure why it hasn't been merged yet.

intangibles:
Cool. I'm pretty excited for it. My attunement hasn't been training well for whatever reason.  (And I don't power walk.) This timer piece is exactly what I need and was looking for!

crossing-training.lic
rcuhljr: Having this always enabled breaks the ability to control attunements priority in between hunts. Crossing training will skip over attunement proper in favor of a lower listed skill because it's at 1/34, so I'd leave town training with 1 mindstate of attunement and 5-6 mind states of other skills I consider less important.

PR #1466, opened 2017-03-12
adding lowbie kobolds in shard

izawwlgood:


data/base-hunting.yaml
```diff
@@ -209,7 +217,8 @@ hunting_zones:
   - 2269
   - 2270
   - 2261
-  # https://elanthipedia.play.net/Skeletal_Kobold_Savage 85-110
+  # https://elanthipedia.play.net/Skeletal_
+  kobold_savage 85-110
```
rpherbig: It looks like there is an extra line break here.

PR #1462, opened 2017-03-10
Adding support for necro harvest storage

FluxBorn:
I have updated the combat-trainer.lic file to support the addition of the option to designate a bag to keep harvested material from the necromancer ritual in.
Added the following config options in profile.
thanatology:
  store: true
  harvest_container: bag
  harvest_count: 5

PR #1471, opened 2017-03-13
stand-alone Necromancer Healing

intangibles:
This script will run the necromancer to a specified hunting area and kill creatures until fully healed.  Script uses base-hunting for user areas, hunting-buddy 'move to' logic, and combat-trainer to handle the combat portion (and consume flesh logic.)

User should setup a combat YAML such as User-heal.YAML which determines what skills/abilities to use in combat-trainer.

example: 
intangibles-heal.yaml

---
hunting_info:
- :zone: grass_eels
  args:
  - d0
  - heal
  :duration: 5
  stop_on:
  - Thievery

thanatology:
  ritual_type: dissect
  heal: true
necromancer_healing:
  Consume Flesh:
    abbrev: CF
    mana: 15
    cambrinth:
    - 5

training_abilities: {}
use_stealth_attacks: false

dance_skill:
weapon_training:
  Brawling: ''

offensive_spell_cycle:
combat_spell_training:

cycle_armors:
custom_loot_type: goods

skinning:
  skin: false


rcuhljr:
I'll look at this in some more detail tomorrow, but I think you could save a lot of work in this script by just running `start_script('hunting-buddy', ['heal'])` and then when you detect the fully healed flag call `$HUNTING_BUDDY.stop_hunting` and wait for it to exit.

necroHeal.lic
```diff
@@ -0,0 +1,171 @@
...
+## Uses base-hunting.Yaml
+$huntZone = 'goblins'
+
+## Set to your Name-heal.yaml file.
+$yamlFile = 'heal'
+
+if $huntZone == 'XXX'
+  echo "You must open the script and set your hunting zone and yamlFile"
+  exit
+end
```
rcuhljr: We should use the yaml files for these settings inside the class instead of global variables in the script.

necroHeal.lic
```diff
@@ -0,0 +1,171 @@
...
+      walk_to(escort_info['base'])
+      wait_for_script_to_complete('bescort', [escort_info['area'], escort_info['enter']])
+      @exit = [escort_info['area'], 'exit']
+    end
+
+    true
+  end
+
+  ## new method for running a script and waiting for it to finish
+  def runWaitScript(script, var1)
```
rcuhljr: DRC contains `wait_for_script_to_complete` which should do what you want.

necroHeal.lic
```diff
@@ -0,0 +1,171 @@
...
+
+  ## Start new code for necrohealing
+  def main
+
+    ## Sets flag to check for game line to determine when no wounds exist
+  
+    ## Flag uses consume logic from combat-training. If no wounds are left
+    ## then combat-training uses dissect or harvest. Regex tests for both rituals.
+    Flags.add('noWounds', '(You carefully position the corpse and make a cut|You bend over .* to make a few quick)')
+    Flags['noWounds'] = false
```
rcuhljr: This line doesn't do anything, `add()` initializes the key to falsey.

necroHeal.lic
```diff
@@ -0,0 +1,171 @@
...
+          kill_script 'combat-trainer' if running?('combat-trainer') 
+        end
+      }
+    }
+
+    ## Run YAML script via combat trainer
+    runWaitScript("combat-trainer", $yamlFile)
+    
+    ## Clean up functions once healing is done
+    bput('loot', "(You search|I could not)")
```
rcuhljr: As a general best practice we prefer to split apart the matches for bput for clarity. Something like
```
bput('loot', 'You search', 'I could not')
```

necroHeal.lic
rcuhljr: Oh, similar to bput you can also 
```
Flags.add('noWounds', 'You carefully position the corpse and make a cut', 'You bend over .* to make a few quick')
```

rpherbig: I'd rather use our existing health check logic than add a flag that relies on behavior in another script.

rpherbig:
I agree with @rcuhljr 's comments - I'll try and do a review of this tonight if it hasn't already been merged.

necroHeal.lic
```diff
@@ -0,0 +1,171 @@
...
+    ## Run to hunting zones
+    find_hunting_room?($huntZone)
+
+    ## Checks for status of wound flag in background
+    Thread.new {
+      loop {
+        sleep 3
+        if Flags['noWounds']
+          # Need to add graceful exit from combat-training
+          kill_script 'combat-trainer' if running?('combat-trainer')
```
rpherbig: The way to gracefully exit combat-trainer is `$COMBAT_TRAINER.stop`. That will tell it to wrap up what it's currently doing, then exit.

necroHeal.lic
rpherbig: The `check_health` method in `common-healing` should do what you want.

necroHeal.lic
```diff
@@ -0,0 +1,171 @@
+=begin
```
rpherbig: Our convention has been to put documentation on elanthipedia and have a script header like this:

```
=begin
  Documentation: https://elanthipedia.play.net/Lich_script_development#common-healing
=end
```

rpherbig:
I just merged PR #1462 which will store high-quality harvested material. Would it be worth having a separate script which checks for already-stored materials and uses those? If you're still wounded, then hunting would be the next step.

necroHeal.lic
intangibles: will do! thanks

necroHeal.lic
intangibles: I was initially using combat.trainers check_health but didn't want to replicate the the entire game_state variable.  Is there a way to pull status via combat.trainer?

Right now combat-trainer is doing the actual health command so that's why i was using the flag to check. Is there a way for common-health to store health data without actually doing the health command?

necroHeal.lic
intangibles: Awesome, thanks!

necroHeal.lic
rpherbig: You shouldn't need to interact with GameState in any way. Check out `common-healing`, the `check_health` method. It's independent of combat-trainer.

intangibles:
Couple thoughts:

1. 

Is there a way to use hunting-buddies methods without copying them over? This syntax is probably wrong but hopefully I'm getting the point across.

$hunting-buddy.init
$hunting-buddy.find_hunting_room?(zone_name)

2.

Harvested materials would be a good addition to the script. Is there any methods that currently take advantage of healing off material?


Thanks!

rcuhljr:
Look at my first top level comment, I don't think there's any reason to do anything other than just call hunting-buddy itself.

rpherbig:
It looks like nearly every script in the repo had a change applied. Can you take a look at this again?

intangibles:
Ah crap.  Sorry, not too familiar with git yet.  I thought i was just uploading docs with changes.

intangibles:
Closing pull request and reopening.  Slight mishap with my fork.

PR #1474, opened 2017-03-13
Adding another "container full" case.

jayRyan24:
Found another on Inoo.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -203,11 +203,8 @@ class LockPicker
     end
 
     if @make_pets
-      case bput("put my #{box} in my #{@settings.picking_pet_box_source}", 'You put your .* in your ', 'You just can\'t get', 'There isn\'t any more')
-      when 'You just can\'t get'
-        bput("put my #{box} in my #{@settings.picking_box_source}", 'You put your')
-        @pet_count = @pet_goal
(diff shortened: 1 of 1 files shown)


rcuhljr:
Should this guy be added to combat-trainer's loot?

pick.lic
```diff
@@ -203,10 +203,13 @@ class LockPicker
     end
 
     if @make_pets
-      case bput("put my #{box} in my #{@settings.picking_pet_box_source}", 'You put your .* in your ', 'You just can\'t get')
+      case bput("put my #{box} in my #{@settings.picking_pet_box_source}", 'You put your .* in your ', 'You just can\'t get', 'There isn\'t any more')
       when 'You just can\'t get'
```
rpherbig: You can combine these two cases with `when 'You just can\'t get', 'There isn\'t any more'`

rpherbig: Maybe consider double quoting those strings too

jayRyan24: Ah, was figuring you could do that, but I did when case || case, bleh.

rpherbig:
This new string was already in `combat-trainer`'s looting process.

Methinks we need to write some common helpers for get/put/stow...

Issue #1214, opened 2016-12-29
Learned.lic


drboldwin:
Could we add a bit to this that tracks each individual's skill time that is being learned?  So we can see what skills are clearing more often and adjust accordingly?

EXAMPLE

Targeted Magic   2.36 (22.9 Hours)
Tactics       0.81 (12.32 Hours)

skyleronken:
so you want to track total hours that a skill has been absorbing experience in the area, or do you want to know how many hours per rank?

rpherbig:
@drboldwin Does PR #1477 (now merged) address this?

PR #1477, opened 2017-03-14
Learned.lic - ranks per hour

skyleronken:
added a per skill column that indicates how many ranks gained per hour based upon the exp tracker. should help in optimizing training.


skyleronken:
this _might_ be sufficient for #1214

rcuhljr:
Would this be something other than the individual skill divided by the time spent learning value from the end of learned?

skyleronken:
#1214 is something different than what I have here. I asked if this sort of thing would suffice but didnt hear back from the issue creator. 

This change is something I wanted for myself (and I think others would find useful). Just thought it might suffice for that ticket as well.

rcuhljr:
Yeah I'm just not certain what you're asking for, is it literally just each value in the learned output divided by the same number?

skyleronken:
Yes. In theory it would tell you how your current training setup was performing in terms of rank gain. Modifications to your yaml could be made and you could see if your hourly rank gain improved.

Perhaps I am misunderstanding what I should be using learned.lic for?

rcuhljr:
It's certainly an option, I made learned for comparative purposes between the skills. Any increase I make in one skill likely comes at a decrease in others, so I wanted to know how much each skill was gaining compared to the others.  I've already picked my priority skills and they're kept with exp in the pool 24/7 so there's no way to improve hour/rank gain on those, I just use learned to make sure that the other skills are gaining acceptable ranks.

skyleronken:
But yes. That's all it is. The ranks learned divided by the hours you've been tracking learned ranks.

skyleronken:
Ah. For me it's a little different.

While training all weapons I don't want to neglect some other non-combat skills. So I have to leave combat where most of my weapon skills drain before I return. I keep trying to optimize how long to stay out of combat without it significantly impacting circling and such. Knowing how long it takes to gain a single rank is helpful in determining this impact.

rcuhljr:
Would a a flag be preferable or always display this output? e.g. `;learned rate` -> `Empathy      1.2/hr` or `;learned` -> ` Empathy       4.78 (1.2/hr)`

skyleronken:
I like the addition. I don't think there is any harm in having it always be present, but I know people use a lot of different FEs with different window sizes. What would you prefer I do?

rcuhljr:
And I just realized this was a PR not an issue, I hate mornings.

learned.lic
```diff
@@ -4,13 +4,15 @@
...
-  skills.each { |skill| output += skill.name.center(18) + DRSkill.gained_exp(skill.name).to_s.center(6) }
+  skills.each { |skill| output += skill.name.center(18) + DRSkill.gained_exp(skill.name).to_s.center(6) + "(#{(DRSkill.gained_exp(skill.name)/learning_time).round(2)})".center(6)}
   respond(output)
 end
 
 total = DRSkill.list.select { |item| item.current - item.baseline > 0 }.map { |skill| DRSkill.gained_exp(skill.name) }.inject(&:+).round(2)
 
 respond("   TOTAL: #{total}")
-respond("   Learning For: #{((Time.now - DRSkill.start_time) / 60.0 / 60).round(1)} hours")
+respond("   Learning For: #{learning_time} hours")
```
rcuhljr: I'd leave the round(1) call here, so that we retain the greater precision for determining ranks above.

learned.lic
```diff
@@ -4,13 +4,15 @@
 
 custom_require.call(%w(drinfomon))
 
+learning_time = ((Time.now - DRSkill.start_time) / 60.0 / 60).round(1)
+
 DRSkill.list.select { |item| item.current - item.baseline > 0 }.sort_by { |item| item.current - item.baseline }.reverse.each_slice(4) do |skills|
   output = ''
-  skills.each { |skill| output += skill.name.center(18) + DRSkill.gained_exp(skill.name).to_s.center(6) }
+  skills.each { |skill| output += skill.name.center(18) + DRSkill.gained_exp(skill.name).to_s.center(6) + "(#{(DRSkill.gained_exp(skill.name)/learning_time).round(2)})".center(6)}
```
rcuhljr: It my be time create a helper method like

```
def format_skill_data(skill, time)
  gain = DRSkill.gained_exp(skill.name)
  skill.name.center(18) + gain.to_s.center(6) + "(#{(gain/time).round(2)})".center(6)
end
```

We could then do something like
```
output = skills.map { |skill| format_skill_data(skill, learning_time)}.join
respond(output)
```

skyleronken:
should be good to go. Sample output:

`--- Lich: learned active.

    Athletics      0.02 (0.16/hr)   Light Armor     0.02 (0.16/hr)  Offhand Weapon   0.01 (0.08/hr)     Skinning      0.01 (0.08/hr)

      Arcana       0.01 (0.08/hr)     Evasion       0.01 (0.08/hr)     Warding       0.01 (0.08/hr)   Augmentation    0.01 (0.08/hr)

    Perception     0.01 (0.08/hr)    Attunement     0.01 (0.08/hr) Elemental Magic   0.01 (0.08/hr) Outdoorsmanship   0.01 (0.08/hr) 

Parry Ability    0.01 (0.08/hr)   Plate Armor     0.01 (0.08/hr)    Brigandine     0.01 (0.08/hr)   Chain Armor     0.01 (0.08/hr)

     Tactics       0.01 (0.08/hr)   Shield Usage    0.01 (0.08/hr)

   TOTAL: 0.2

   Learning For: 0.1 hours

--- Lich: learned has exited.
`

PR #1481, opened 2017-03-15
Check encumbrance helper function

skyleronken:
Noticed a few feature requests would require an encumbrance check. Added a helper function to DRC that will run encumbrace, capture the response message, and then return the numeric value from 0-11.


skyleronken:
,en echo DRC.check_encumbrance

--- Lich: exec1 active.

[exec1]>encumbrance

  Encumbrance : None (0/11)
> 
[exec1: 0]

--- Lich: exec1 has exited.

common.lic
```diff
@@ -283,4 +298,11 @@ module DRC
       .map { |_name, properties| properties['abbrev'] }
       .each { |abbrev| fput("release #{abbrev}") }
   end
+  
+  def check_encumbrance
+    enc = nil
+    enc_msg = bput('encumbrance', 'Encumbrance : .*')
```
rpherbig: You can get your current encumbrance (as a string) without sending a command via `DRStats.encumbrance`. You'll still have to map that, though.

common.lic
```diff
@@ -283,4 +298,11 @@ module DRC
       .map { |_name, properties| properties['abbrev'] }
       .each { |abbrev| fput("release #{abbrev}") }
   end
+  
+  def check_encumbrance
+    enc = nil
```
rpherbig: Where is `enc` used?

skyleronken:
so I changed it to use DRStats.encumbrance. But DRStats is only updated when the encumbrance command is issued. So I added a default true boolean flag to run the encumbrance command before fetching the value.

PR #1487, opened 2017-03-17
Self activated/stand-alone necromancer healing script

intangibles:
New version of necroheal with changes/suggestions from prior PR.  Had to delete prior PR due to some accidental git commit issues.

Current healing methods consist of turning on consume healing during training via combat-trainer via training-manager. There is currently no solution for healing outside training-manager, so this script is designed for activation either stand-alone from command bar, or anywhere else a script can be activated.

The script is far simpler in this version after using suggestions from everyone. Thank you!

I ended up not being able to use the check_health flag of common-healing to detect wounds because check_health uses the "health command".  Combat-trainer also uses the health command and having both commands spam health is not ideal.  Ideally, if combat-trainer  variable gamestate can be accessed outside of combat trainer, then I would be able to use the check_health that way.  I'm not entirely sure how i would go about doing this. 

Thanks!

See PR #1471 for original comments and script.

### DIFF
--- a/necroHeal.lic
+++ b/necroHeal.lic
@@ -1,0 +1,45 @@
+=begin
+
+  Documentation: https://elanthipedia.play.net/Lich_script_development#necroheal
+  ## ** Placeming comments here until finalized and added to pedia.
+
+  Necromancer stand-alone healing script
+  
+  Uses all settings defined in HEAL YAML FILE (default)
+  Consume logic via combat trainer
+  
+  TO DO:
+  - See pull request #1462 for harvested materials.
+  - See if there is a background way to use common-healing to determine wound status
+  - Add YAML variable to swap YAML file name
+  
+=end
+
+  if DRCH.check_health.empty?
+    echo 'Not wounded, exiting heal script'
+    exit
+  end
+
+
+  ## Sets flag to check for game line to determine when no wounds exist
+  ## Flag uses consume logic from combat-training. If no wounds are left
+  ## then harvest or dissection is used. Flags trigger off these rituals.
+  Flags.add('noWounds', 'You carefully position the corpse and make a cut', 'You bend over .* to make a few quick')
+  
+  ## Checks for status of wound flag in background
+  Thread.new {
+    loop {
+      sleep 3
+      if Flags['noWounds']
+        $HUNTING_BUDDY.stop_hunting # graceful exit
+        sleep 0.3 until !Script.running('comabt-trainer')
+      end
+    }
+  }
+
+    
+  ## Run YAML script via combat trainer
+  DRC.wait_for_script_to_complete('hunting-buddy', ['heal'])
+  
+  echo 'Healing done!'
+  
--- a/necroHeal.lic
+++ b/necroHeal.lic
@@ -1,7 +1,7 @@
 =begin
 
   Documentation: https://elanthipedia.play.net/Lich_script_development#necroheal
-  ## ** Placement comments here until finalized and added to pedia.
+  ## ** Placeming comments here until finalized and added to pedia.
 
   Necromancer stand-alone healing script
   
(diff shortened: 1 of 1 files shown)


necroHeal.lic
```diff
@@ -0,0 +1,43 @@
+=begin
+
+  Documentation: https://elanthipedia.play.net/Lich_script_development#necroheal
```
rpherbig: Our convention has been to name scripts something like `necro-heal.lic`.

necroHeal.lic
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+
+  ## Sets flag to check for game line to determine when no wounds exist
+  ## Flag uses consume logic from combat-training. If no wounds are left
+  ## then harvest or dissection is used. Flags trigger off these rituals.
+  Flags.add('noWounds', 'You carefully position the corpse and make a cut', 'You bend over .* to make a few quick')
+  
+  ## Checks for status of wound flag in background
+  Thread.new {
```
rpherbig: You don't need to put this in a new thread.

necroHeal.lic
```diff
@@ -0,0 +1,43 @@
...
+  ## ** Placement comments here until finalized and added to pedia.
+
+  Necromancer stand-alone healing script
+  
+  Uses all settings defined in HEAL YAML FILE (default)
+  Consume logic via combat trainer
+  
+  TO DO:
+  - See pull request #1462 for harvested materials.
+  - See if there is a background way to use common-healing to determine wound status
```
rpherbig: Have you looked into integrating this with `safe-room.lic`?

intangibles: Ahhhh. That's right, safe-room gives a wound status level.  That could be interesting.  Are you thinking I would connect it the safe-room script?

So part of the plan here was to make something useable on the fly, so if you are running around areas exploring, pvping, general mayhem, you could slam necroheal, run out to goblins and heal up real quick.

It's also been turning into a nice integration as an after: script if i would rather heal up in a specific area versus when hunting.  The safe-room would be a nice option for this purpose but sort of defeats the stand alone aspect.   What do you think?

necroHeal.lic
```diff
@@ -1,7 +1,7 @@
 =begin
 
   Documentation: https://elanthipedia.play.net/Lich_script_development#necroheal
-  ## ** Placement comments here until finalized and added to pedia.
+  ## ** Placeming comments here until finalized and added to pedia.
```
rpherbig: "Placing"

necroHeal.lic
rpherbig: The one goal of `safe-room` is to heal you up. This sounds like an extension of that.

There's even a `!DRStats.necromancer?` check to avoid taking you to Dokt - I imagine we could just run `hunting-buddy(heal)` in that case.

`Safe-room` already runs as part of `training-manager`, so you wouldn't need it as an 'after' script either.

rcuhljr: Generally speaking, sub scripts should always function stand alone, so people who just want to necro-heal can, but tying it into safe-room is quite useful since it's basically the necro version of automated healing.

rpherbig: I run `safe-room` by itself all the time (i.e. while I'm not running `training-manager`). I'm pretty sure others do too. It's a useful standalone script.

jandersson: Right now `;safe-room` checks if you are a necro and quits. That check could run this healing logic instead of exiting with the added benefit of being integrated into `;training-manager`. Sounds like a win-win

jayRyan24: Hey there, I think there's a disconnect between what Lich is and what Dependency is. Basically, your script is good and all, but Dependency is a scripting automation suite, it isn't the same thing as Lich. Dependency users will find value in it, so they will add it, but they also think that they have a use for it to run inside the Dependency engine. This is separate from your script being added, they just want it to be even more useful. 

Also, not sure if it's clear or not, but even though it is being used by safe-room as part of the suite, it can also be used in befores and afters as well as by using the ; invoker to run it stand-alone.

Issue #1489, opened 2017-03-18
Can we get a duskruin rat catching script?


Caledric:
https://github.com/rpherbig/dr-scripts/blob/master/samples/duskruin-labyrinth.cmd

Is a good script written for genie... we could use one for lich.

jandersson:
You can close this Caledric. Though please make new issues to make the rat catcher better :]

rpherbig:
See `dusk-labyrinth.lic`

PR #1490, opened 2017-03-18
Duskruin 422 Labyrinth

jandersson:
Working draft. Another user reported that this caught a rat

### DIFF
--- a/dusk-labyrinth.lic
+++ b/dusk-labyrinth.lic
@@ -1,0 +1,63 @@
+=begin
+
+A Duskruin Labyrinth script for March 2017 Duskruin
+
+author: Crannach
+game: dr
+=end
+custom_require.call %w(common common-items)
+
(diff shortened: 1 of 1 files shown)


jandersson:
#1490 should be what you need

jandersson:
Should work for #1489

rpherbig:
I think it would be nice to add the header block with a link to elanthipedia (and add some documentation there), but that won't hold up merging this.

jandersson:
Great! I will add this. How does it work if this is already merged? Do I create a new branch for updates or can i just push to this same branch?

jandersson:
I've pushed new changes to this same branch. I'm going to try out this 'revert' button.. let me know if you want me to do this another/a better way

jandersson:
Not what I wanted! OK, new branch it is

PR #1493, opened 2017-03-19
Perc mana helper function

skyleronken:
Added a helper function to DRCA that will do a perc/perc mana and return
the numeric value of the current mana level (relative to your attunemnt
skill level) or for MM's, it returns a hash of the different mana types
and their values.


common-arcana.lic
```diff
@@ -179,4 +186,47 @@ module DRCA
       .map { |_name, properties| properties['abbrev'] }
       .each { |abbrev| fput("release #{abbrev}") }
   end
+  
+  def parse_mana_message(mana_msg)
+      manalevels = nil
+      if mana_msg.include? 'weak'
+        manalevels = $MANA_MAP['weak']
+      elif mana_msg.include? 'developing'
```
jandersson: `elsif` no? Think `elif` is python :P

skyleronken: good catch!

common-arcana.lic
```diff
@@ -179,4 +186,47 @@ module DRCA
...
+      
+      mana_levels = Hash.new
+      mana_levels['enlightened_geometry'] = parse_mana_message(mana_msgs[0])
+      mana_levels['moonlight_manipulation'] = parse_mana_message(mana_msgs[1])
+      mana_levels['perception'] = parse_mana_message(mana_msgs[2])
+      mana_levels['psychic_projection'] = parse_mana_message(mana_msgs[3])
+      return mana_levels
+    else    
+      mana_msg = DRC.bput('perc','^You reach out with your .* and (see|hear) \w+')
+      return parse_mana_message(mana_msg)
```
rpherbig: I *suspect* we'll find it easier to use this helper if it returns the same data type for every guild (e.g. a hash with one pair for non-moon mages), but we'll know more once we use it somewhere.

PR #1503, opened 2017-03-23
Hunting room min mana

### LINKED ISSUE #1388: Hunting buddy should have an option to check for rooms with good mana
Currently, hunting buddy looks for the first empty room in the room list and begins hunting there. In many cases though, that room may be extremely low on mana and there is an empty room with high mana right next to it. Hunting buddy should have an option that users can set in their profile to have it search out for the empty room with the highest mana.

skyleronken:
Altered common-travel find_empty_room so that you can define a min mana level as well as how strictly to enforce that level. 
- if it finds an empty room, it checks the mana level
- if the mana level is not met, it moves on. 
- if there are empty rooms, but none had the right mana, it will check the strictness argument on whether or not to idle or to fall back on any empty room regardless of mana. 

Added these arguments to hunting-buddy and default false values to base.yaml

resolved #1388


profiles/base.yaml
```diff
@@ -119,7 +119,8 @@ empty_hunting_room_messages:
 hunting_buddies:
 stop_hunting_if_bleeding: true
 prehunt_buffs:
-
+hunting_room_min_mana: false
```
rpherbig: Is `false` the right default for this setting?

common-travel.lic
```diff
@@ -81,15 +81,30 @@ module DRCT
...
+      found_empty = false
       search_rooms.each do |room_id|
         walk_to(room_id)
         pause 0.1 until room_id == Room.current.id
-        empty = predicate ? predicate.call : (DRRoom.pcs - DRRoom.group_members).empty?
-        return true if empty
+        suitable_room = predicate ? predicate.call : (DRRoom.pcs - DRRoom.group_members).empty?
+        if suitable_room and check_mana and not DRStats.moon_mage?
+          found_empty = true
+          suitable_room = (DRCA.perc_mana >= min_mana)
```
rpherbig: This is the comparison that seems odd if `check_mana` (and therefore `min_mana`) is a boolean.

common-travel.lic
```diff
@@ -81,15 +81,30 @@ module DRCT
     end
   end
 
-  def find_empty_room(search_rooms, idle_room, predicate = nil)
+  def find_empty_room(search_rooms, idle_room, predicate = nil, min_mana = false, strict_mana = false)
+
+    check_mana = min_mana
```
rpherbig: Why is `check_mana` used in some places and `min_mana` in others?

rpherbig:
Do you plan to update `bescort` so that those hunting zones are also supported?

profiles/base.yaml
skyleronken: i was trying to simplify by letting me use the hunting_room_min_mana setting as a boolean flag to enter the logic at all. However, I can chagne this to 0 and have the boolean logic be > 0.

common-travel.lic
skyleronken: Same as above. I was trying to limit the number of settings and arguments. I can just change the logic to be if > 0.

skyleronken:
it looks like it could be done in bescort, but each zone has a completely different way of navigating. I am a little hesitant to take that on now. Would you be comfortable with merging without that functionality? This is a nice-to-have atm so people might not even care about those one-off cases. Sorry if im being lazy :/

The other changes have been made. Let me know what you think.

rpherbig:
Ignoring `bescort` is fine.

Why does this logic ignore moon mages? `DRCA.perc_mana` works for them (it just returns a hash instead of a number).

skyleronken:
Because the mana doesn't change room to room for them so it didn't seem to fit.

rpherbig:
Oh, TIL I learn something new, thanks!

I'll merge this, if you could please update elanthipedia's script documentation with the new settings.

PR #1519, opened 2017-04-01
Added option to smartlisten to use 'listen observe'

adamgoerdt:


profiles/base.yaml
```diff
@@ -170,6 +170,7 @@ exp_timers:
   Thievery: 600
   Attunement: 130
 listen: false
+listen_observe: false
```
rpherbig: From the name of this setting, I would expect it to be used everywhere that `DRC.listen?` is called. What are your thoughts?

adamgoerdt: Makes sense.  I updated crossing-training to use the setting as well

PR #1535, opened 2017-04-10
Ensure water_holder goes into theurgy_supply_container

tulanegator:
After adding Meraud to CT I ran into issues where my water_holder was in my default store container rather than my theurgy_supply_container. This should correct that issue.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1304,8 +1304,8 @@ class TrainerProcess
     end
     waitrt?
     @equipment_manager.stow_weapon(game_state.weapon_name)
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -1306,7 +1306,7 @@ class TrainerProcess
     @equipment_manager.stow_weapon(game_state.weapon_name)
     fput "get #{@water_holder}"
```
rpherbig: Do you also want to `get #{@water_holder} from {@theurgy_supply_container}`?

PR #1538, opened 2017-04-11
TM Focus

jandersson:
The idea is to wield the focus like a weapon and only attack with magic. To use: set the focus as a training weapon and include it as a gear item. Included my profile as an example.


jandersson:
want to test some more

PR #1545, opened 2017-04-14
Use a TM Focus in combat-trainer

jandersson:
Wield a tm focus like a weapon, invoke it after wielding it. Using targeted magic as your weapon training skill means you wont attack with weapons.

The base requirement is to have:
```
weapon_training:
  Targeted Magic: tetractys
```
Where tetractys is the name of your focus.

Alternatively, you can cast TM without attacking with weapons by listing 
```
weapon_training:
  Targeted Magic: ''
```

Crannach-tm.yaml has been uploaded as a simple example

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -174,7 +174,7 @@ class SetupProcess
     end
 
     # Invoke Focus
-    if game_state.weapon_skill == 'Targeted Magic'
+    if game_state.weapon_skill == 'Targeted Magic' && game_state.weapon_name
       bput("invoke #{game_state.weapon_name}", 'You')
       waitrt?
     end 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1077,10 +1077,13 @@ class SpellProcess
     @custom_cast = data['cast']
     @symbiosis = data['symbiosis']
     @after = data['after']
+    @skill = data['skill']
+    @prep = data['prep']
     @command = data['command']
     @reset_expire = data['expire'] ? data['abbrev'] : nil
     Flags.reset('ct-spellcast')
     @before = data['before']
+    game_state.action_taken if (@skill == 'Targeted Magic' || @prep == 'target') && game_state.weapon_skill == 'Targeted Magic'
   end
 end
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -172,6 +172,12 @@ class SetupProcess
     else
       @equipment_manager.wield_weapon(game_state.weapon_name, game_state.weapon_skill)
     end
+
+    # Invoke Focus
+    if game_state.weapon_skill == 'Targeted Magic'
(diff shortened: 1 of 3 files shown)


combat-trainer.lic
```diff
@@ -172,6 +172,12 @@ class SetupProcess
     else
       @equipment_manager.wield_weapon(game_state.weapon_name, game_state.weapon_skill)
     end
+
+    # Invoke Focus
+    if game_state.weapon_skill == 'Targeted Magic'
+      bput("invoke #{game_state.weapon_name}", 'You')
```
rpherbig: If you did something like `bput(...) if game_state.weapon_name` I think this would also support syntax like:
```
weapon_training:
  Targeted Magic:
```

I.e. Train TM without weapons, which some people have asked for.

jandersson: So in the absence of an item, or maybe something like `Targeted Magic: ''` don't invoke the focus

jandersson: Ill change it and test it out

jandersson:
Updated to work without a focus (kind of like brawling)

combat-trainer.lic
```diff
@@ -1077,10 +1083,13 @@ class SpellProcess
     @custom_cast = data['cast']
     @symbiosis = data['symbiosis']
     @after = data['after']
+    @skill = data['skill']
+    @prep = data['prep']
     @command = data['command']
     @reset_expire = data['expire'] ? data['abbrev'] : nil
     Flags.reset('ct-spellcast')
     @before = data['before']
+    game_state.action_taken if (@skill == 'Targeted Magic' || @prep == 'target') && game_state.weapon_skill == 'Targeted Magic'
```
rpherbig: Do you need the `(@skill == 'Targeted Magic' || @prep == 'target')` part of this conditional? Or, to ask my question differently, what does it add that isn't covered by `game_state.weapon_skill == 'Targeted Magic'`?

jandersson: It helps cover the gap where a person lists a spell as Sorcery, but its really a TM spell

jandersson: I believe part of the spell data merge logic also searches for spells listed as "sorcery" and inserts a "prep: target", so this covers the spells where we either have data on it or the user lists prep: target

jandersson: This will probably break if the person is using a TM spell, lists it as sorcery, and it is not in our base-spell data.

rpherbig: Even if it is a TM spell using Sorcery, won't they have to list TM in their weapon list?

jandersson: True, but how to increment the action counter?

rpherbig: Oh, I see. This prevents incrementing the action counter while we, say, cast an Aug buff during 'TM' weapon training.

PR #1559, opened 2017-04-21
Added broken-out athletics script

adamgoerdt:


jayRyan24:
This needs to include athletics for higher tiers of skills.

	
```
	if DRSkill.getrank('Athletics') >= 550
	    hlc_athletics
    elsif @settings.hometown == 'Crossing'
      crossing_athletics
    elsif @settings.hometown == 'Riverhaven'
      riverhaven_athletics
    elsif @settings.hometown == 'Shard'
      shard_athletics
    elsif @settings.hometown == 'Hibarnhvidar'
      hib_athletics
    elsif @settings.hometown == 'Ratha'
      ratha_athletics
    end
  end
  
  def hlc_athletics
    start_script('skill-recorder') unless Script.running?('skill-recorder')
    walk_to(2245)
    if DRSkill.getxp('Athletics') < 29
      walk_to(2245)
      walk_to(9607)
	  walk_to(19_464)
        while DRSkill.getxp('Athletics') < 29
          walk_to(2245)
          walk_to(9607)
          walk_to(11_126)
          walk_to(19_464)
          break if DRSkill.getxp('Athletics') >= 29
        end
    end
  end
```

rcuhljr:
So before I dig into this in a detailed review, could this be better accomplished by changing crossing-training to support athletics in those areas and giving it the option of different run patterns (run single skill, run until locked, run skill once, etc.)

adamgoerdt:
Broken out scripts allow for a lot more flexibility in training since they can be used in conjunction with before or after actions.  The section of code for the Crossing hometown is a fork-life from crossing-training, which sections added for other hometowns.

The intent of this stand-alone script is to replace methods in crossing-training.  Once this script is accepted I'll make the necessary modifications to crossing-training to do so, but didn't want to make such a large modification all at once.

rcuhljr:
Roger.

rcuhljr:
High level thoughts. This is half athletics and half performance, let break play out into it's own script. We also want this to be able to take arguments so it can support research and osrel orb maintenance.  I'd also consider adding an argument `local` that ignores hometowns and selects the closest appropriate climb. We need some way for people to do climbs in multiple locales I suspect.

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+    @no_instrument = false
+    @settings = get_settings()
+    @athletics_options = get_data('athletics').athletics_options
+    @song_list = get_data('perform').perform_options
+ 
+    Flags.add('ct-song', 'you finish playing')
+    play_song?
+    
+    @start_skill=DRSkill.getxp('Athletics')
+    @end_skill=@start_skill+20
```
rcuhljr: `@start_skill` is never used again, also why 30 instead of 32?

`@end_skill = [DRSkill.getxp('Athletics')+20, 32].min`

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+    if(@end_skill>30)
+      @end_skill=30
+    end
+
+    if @settings.hometown == 'Crossing' || @settings.hometown == 'Riverhaven'
```
rcuhljr: Since we're string comparing this might be cleaner as 
```
case @settings.hometown
when 'Crossing', 'Riverhaven'
  stuff
when 'Shard'
```

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+    elsif @settings.hometown == 'Hibarnhvidar'
+      hib_athletics
+    elsif @settings.hometown == 'Ratha'
+      ratha_athletics
+    end
+  end
+  
+  def crossing_athletics
+    if DRSkill.getrank('Athletics') >= 400
+      return if @researching
```
rcuhljr: @researching is a hold over from crossing-training.

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+      climb_branch
```
rcuhljr: `climb_branch` doesn't exist in here.

adamgoerdt:
I've got a broken out performance script as well, but athletics will sometimes stop performing, which leaves the performance script hanging.  I'm thinking that I'll remove the performance aspects from this script, but it's a nice feature to be able to hammer out some zills while climbing

athletics.lic
adamgoerdt: Hrm.  Would a UserVar make sense, then, to track state between scripts?

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+    else
+      while DRSkill.getxp('Athletics') < @end_skill
+        @athletics_options
+          .reject { |data| @settings.avoid_athletics_in_justice && data['justice'] }
+          .each do |data|
+            break unless climb?(data['room'], data['targets'])
```
rcuhljr: It feels odd that the athletics check against the target is done as part of the return value of `climb?()` if that method returned false I expect something to have gone wrong with the climb, not that we're full on exp. I'd rather have a helper method `done_climbing?` that checks exp and then we can 
```
climb(data...)
break if done_climbing?
```

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+        end
+      end
+    end
+  end
+  
+  def shard_athletics
+    while DRSkill.getxp('Athletics') <= @end_skill
```
rcuhljr: `until done_training?`

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+      walk_to(11481)
+      walk_to(11482)
+      walk_to(11483)
+    end    
+    
+    walk_to(@settings.safe_room)
```
rcuhljr: What motivated this walk back? If we allow this script more run time flexibility this seems like a problem and would be better managed by another before/after with a go2 or as part of crossing-training.

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+  end
+  
+  def hib_athletics
+    walk_to(4155)
+    move('nw')
```
rcuhljr: Can we fix the map to simplify this guy?

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+      3.times do
+        fput('south')
+        waitrt?
+      end 
+      walk_to(4155)
+      move('nw')
+    end
+    fput('climb bank')
+    waitrt?
+    walk_to(@settings.safe_room)
```
rcuhljr: safe_room ditto.

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+  end
+  
+  def climb_practice(room, target, to_hide = false)
+    return unless target
+    walk_to(room)
+    play_song?
+    while DRSkill.getxp('Athletics') <= @end_skill
+      retreat
+      if to_hide
+        attempts_to_hide = 5
```
rcuhljr: If we're counting down this should probably be `remaining_blah` to avoid confusion.

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+        break if Flags['ct-climbing-finished']
+        return if Flags['ct-climbing-combat']
+        if DRSkill.getxp('Athletics') > @end_skill
+          fput('stop climb')
+          break
+        end
+      end
+    end
+    stop_play
+    bput('unhide', 'You come out of hiding', 'You slip out of hiding', 'But you are not')
```
rcuhljr: This guy should probably be conditional on `to_hide`.

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+    walk_to(@settings.safe_room)
+  end
+  
+  def climb?(room, targets)
+    targets.each do |target|
+      walk_to(room)
+      return true if DRRoom.npcs.length >= 3
```
rcuhljr: Do we want to make this smarter and look for key npc names of invasion critters?

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+          break
+        end
+      end
+    end
+    stop_play
+    bput('unhide', 'You come out of hiding', 'You slip out of hiding', 'But you are not')
+    walk_to(@settings.safe_room)
+  end
+  
+  def climb?(room, targets)
```
rcuhljr: See other comments about not making this action method a predicate.

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+        stop_play
+        UserVars.song = @song_list.first.first
+        return play_song?
+      end
+    end
+ 
+    return true unless blocking
+ 
+    Flags.reset('song')
+    pause 1 until Flags['song']
```
rcuhljr: Probably copied from crossing-trainer, but we can swap all the `pause 1` with `pause`

athletics.lic
```diff
@@ -0,0 +1,223 @@
...
+    return if @no_instrument
+    @did_play = false
+    bput('stop play', 'You stop playing your song', 'In the name of', "But you're not performing")
+    Flags['song'] = true
+  end
+  
+end
+before_dying do
+  DRC.bput('stop climb','.*')
+end
```
rcuhljr: Is there a specific case this guards against? I try to avoid before_dying where I can.

athletics.lic
adamgoerdt: I wish.  I tried re-mapping 3 or 4 different times, but it would continue to get stuck somewhere along the way, the location of where you'd get stuck would be consistent for each person using the script, but different across characters with different typeahead settings.  

While ugly, this has been working for almost 4 months now.

athletics.lic
rcuhljr: Yeah, we need some way to do more complex interactions between scripts.

athletics.lic
jandersson: just confirming that ive been getting stuck there as well. , `;go2` usually restarts after 30 seconds or so and continues. Rooms were sending `</nav>` tags last time i checked

athletics.lic
```diff
@@ -0,0 +1,170 @@
+custom_require.call(%w(common common-travel))
```
rpherbig: Could you add a header block with a link to elanthipedia?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+  include DRCT
+  
+  def initialize
+    @settings = get_settings()
+    @athletics_options = get_data('athletics').athletics_options
+
+    @climbing_target = get_data('athletics').practice_options[@settings.climbing_target]
+    
+    @start_skill=DRSkill.getxp('Athletics')
+    @end_skill=@start_skill+15
```
rpherbig: You could use `@end_skill = [@start_skill+15, 29].max` here

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+  include DRC
+  include DRCT
+  
+  def initialize
+    @settings = get_settings()
+    @athletics_options = get_data('athletics').athletics_options
+
+    @climbing_target = get_data('athletics').practice_options[@settings.climbing_target]
+    
+    @start_skill=DRSkill.getxp('Athletics')
```
rpherbig: Maybe `start_exp`?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+    @end_skill=@start_skill+15
+    if(@end_skill>29)
+      @end_skill=29
+    end
+
+    if @climbing_target
+      walk_to(@climbing_target['id'])
+      climb_practice(@climbing_target['target'], @climbing_target['hide'])
+    elsif @settings.hometown == 'Crossing' || (@settings.hometown == 'Riverhaven' and DRSkill.getrank('Athletics') > 140)
```
rpherbig: I think this might be clearer if you moved Riverhaven to its own `elsif`

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+
+    @climbing_target = get_data('athletics').practice_options[@settings.climbing_target]
+    
+    @start_skill=DRSkill.getxp('Athletics')
+    @end_skill=@start_skill+15
+    if(@end_skill>29)
+      @end_skill=29
+    end
+
+    if @climbing_target
```
rpherbig: I'm not sure if a `case`/`when` structure would make this clearer, have you considered that?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+      walk_to(@climbing_target['id'])
+      climb_practice(@climbing_target['target'], @climbing_target['hide'])
+    elsif @settings.hometown == 'Crossing' || (@settings.hometown == 'Riverhaven' and DRSkill.getrank('Athletics') > 140)
+      crossing_athletics
```
rpherbig: I think this might be clearer if you moved Riverhaven to its own `elsif`

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+      walk_to(11483)
+    end 
+  end
+  
+  def swim_liirewsag
+    #Intermittent errors in the map will cause go2 to get stuck.  The manual moves are a workaround for that
+    walk_to(4155)
+    move('nw')
+    until done_training?
+      fput('climb bank')
```
rpherbig: Is `move` more appropriate to use here (rather than `fput`)?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+      3.times do
+        fput('south')
```
rpherbig: Is `move` more appropriate to use here (rather than `fput`)?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+        waitrt?
+      end 
+      walk_to(4155)
+      move('nw')
+    end
+    fput('climb bank')
```
rpherbig: Is `move` more appropriate to use here (rather than `fput`)?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+      Flags.add('ct-climbing-finished', 'You finish practicing your climbing')
+      Flags.add('ct-climbing-combat', 'You are engaged in combat')
+      bput("climb practice #{target}", 'You begin to practice ')
+      loop do
+        pause
+        break if Flags['ct-climbing-finished']
+        return if Flags['ct-climbing-combat']
+        break if done_training?
+      end
+      DRC.bput('stop climb','You stop practicing your climbing skills.',"You weren't practicing your climbing skills anyway.")
```
rpherbig: You don't need the `DRC` prefix here

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+    
+    bput('unhide', 'You come out of hiding', 'You slip out of hiding', 'But you are not') if to_hide
+  end
+  
+  def climb?(room, targets)
+    targets.each do |target|
+      walk_to(room)
+      return true if DRRoom.npcs.length >= 3
+
+      fput "climb #{target}"
```
rpherbig: Could this be changed to a `bput`?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+      climb_practice('branch', true)
+    else
+      walk_to(9607)
+      walk_to(19_464)
+      until done_training?
+        walk_to(2245)
+        walk_to(9607)
+        walk_to(11_126)
+        walk_to(19_464)
+        break if done_training?
```
rpherbig: Isn't this part of `until done_training?` on line 159?

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+      retreat
+      if to_hide
+        remaining_attempts = 5
+        while remaining_attempts > 0
+          break if hide?
+          remaining_attempts -= 1
+        end
+        return if remaining_attempts.zero?
+      end
+      Flags.add('ct-climbing-finished', 'You finish practicing your climbing')
```
rpherbig: You will need to add `events` to your `custom_require` list to use Flags

athletics.lic
```diff
@@ -0,0 +1,170 @@
...
+      shard_athletics
+    elsif @settings.hometown == 'Hibarnhvidar'
+      hib_athletics
+    elsif @settings.hometown == 'Ratha'
+      ratha_athletics
+    end
+  end
+  
+  def done_training?
+    return DRSkill.getxp('Athletics') >= @end_skill
```
rpherbig: You will need to add `drinfomon` to your `custom_require` list to use DRSkill and DRRoom

athletics.lic
adamgoerdt: It's been working without it.  I'm calling common-travel, which calls events and drinfomon, maybe it's picking up the libraries from there.  Either way, I can add it to my require at the top

athletics.lic
adamgoerdt: Good question.  Does move work on non-ordinal directions?

athletics.lic
adamgoerdt: Probably.  I'd need to find all of the success and fail messages for a climb, or use a wildcard. This method was copied from crossing-training, so it's been working for a while.  What's the consensus on "if it ain't broke..."

athletics.lic
adamgoerdt: Something like this? Or should I wrap all of this in a riverhaven_athletics method?

athletics.lic
rpherbig: Yes. For example: https://github.com/rpherbig/dr-scripts/blob/master/bescort.lic#L550

athletics.lic
rpherbig: This is fine

athletics.lic
rpherbig: It's not necessary, but would be nice if it's easy to do

athletics.lic
rpherbig: Hehe, oops. My bad.

athletics.lic
adamgoerdt: There isn't any messaging on a successful climb, it just moves you to the next room, so the only bput I could use matches a wildcard

athletics.lic
adamgoerdt: Re-added train_outdoorsmanship, otherwise the timer resets and you climb uselessly for 60 seconds

athletics.lic
rpherbig: This sounds like something we could do in the future but would not hold up this PR.

rpherbig:
@rcuhljr looks good to me. Do you have any more comments before this is merged?

athletics.lic
rcuhljr: Best not to rely on other libraries to pull in things you'll need, you don't want your script to die if a child script changes behavior.

athletics.lic
```diff
@@ -0,0 +1,186 @@
...
+        waitrt?
+      end 
+      walk_to(4155)
+      move('nw')
+    end
+    move('climb bank')
+    waitrt?
+  end
+
+  def climb_practice(room, target, to_hide = false)
```
rcuhljr: Do we want to play zills before climb practicing since they don't conflict?

rcuhljr:
Seems fine, it's a stand alone script that's not yet hooked into crossing training so the risk seems super low for just merging it and letting errors get sorted out in use.

PR #1562, opened 2017-04-22
Self activated/stand-alone necromancer healing script

intangibles:
EDIT: New PR, trying to fix some git issues.

New version of necroheal with changes/suggestions from prior PR. Had to delete prior PR due to some accidental git commit issues.

Current healing methods consist of turning on consume healing during training via combat-trainer via training-manager. There is currently no solution for healing outside training-manager, so this script is designed for activation either stand-alone from command bar, or anywhere else a script can be activated.

The script is far simpler in this version after using suggestions from everyone. Thank you!

I ended up not being able to use the check_health flag of common-healing to detect wounds because check_health uses the "health command". Combat-trainer also uses the health command and having both commands spam health is not ideal. Ideally, if combat-trainer variable gamestate can be accessed outside of combat trainer, then I would be able to use the check_health that way. I'm not entirely sure how i would go about doing this.

Thanks!

See PR #1487 for additional comments
See PR #1471 for original comments and script.

### DIFF
--- a/necroHeal.lic
+++ b/necroHeal.lic
@@ -1,0 +1,45 @@
+=begin
+
+  Documentation: https://elanthipedia.play.net/Lich_script_development#necroheal
+  ## ** Placeming comments here until finalized and added to pedia.
+
+  Necromancer stand-alone healing script
+  
+  Uses all settings defined in HEAL YAML FILE (default)
+  Consume logic via combat trainer
+  
+  TO DO:
+  - See pull request #1462 for harvested materials.
+  - See if there is a background way to use common-healing to determine wound status
+  - Add YAML variable to swap YAML file name
+  
+=end
+
+  if DRCH.check_health.empty?
+    echo 'Not wounded, exiting heal script'
+    exit
+  end
+
+
+  ## Sets flag to check for game line to determine when no wounds exist
+  ## Flag uses consume logic from combat-training. If no wounds are left
+  ## then harvest or dissection is used. Flags trigger off these rituals.
+  Flags.add('noWounds', 'You carefully position the corpse and make a cut', 'You bend over .* to make a few quick')
+  
+  ## Checks for status of wound flag in background
+  Thread.new {
+    loop {
+      sleep 3
+      if Flags['noWounds']
+        $HUNTING_BUDDY.stop_hunting # graceful exit
+        sleep 0.3 until !Script.running('comabt-trainer')
+      end
+    }
+  }
+
+    
+  ## Run YAML script via combat trainer
+  DRC.wait_for_script_to_complete('hunting-buddy', ['heal'])
+  
+  echo 'Healing done!'
+  
--- a/necroHeal.lic
+++ b/necroHeal.lic
@@ -1,7 +1,7 @@
 =begin
 
   Documentation: https://elanthipedia.play.net/Lich_script_development#necroheal
-  ## ** Placement comments here until finalized and added to pedia.
+  ## ** Placeming comments here until finalized and added to pedia.
 
   Necromancer stand-alone healing script
   
@@ -14,6 +14,9 @@
   - Add YAML variable to swap YAML file name
   
 =end
+
+custom_require.call(%w(common common-healing events))
+
 
   if DRCH.check_health.empty?
     echo 'Not wounded, exiting heal script'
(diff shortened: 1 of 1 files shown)


intangibles:
Can you please merge this pull request with #1487

necroHeal.lic
```diff
@@ -0,0 +1,57 @@
...
+  - See pull request #1462 for harvested materials.
+  - See if there is a background way to use common-healing to determine wound status
+  - Add YAML variable to swap YAML file name
+  
+=end
+
+custom_require.call(%w(common common-healing events))
+
+
+  if DRCH.check_health.empty?
```
rpherbig: Should this use the `saferoom_health_threshold` setting rather than healing on *any* wounds? What about poison or bleeding?

See https://github.com/rpherbig/dr-scripts/blob/master/safe-room.lic#L47 for an example of what I men.

necroHeal.lic
```diff
@@ -0,0 +1,57 @@
...
+  ## Sets flag to check for game line to determine when no wounds exist
+  ## Flag uses consume logic from combat-training. If no wounds are left
+  ## then harvest or dissection is used. Flags trigger off these rituals.
+  Flags.add('noWounds', 'You carefully position the corpse and make a cut', 'You bend over .* to make a few quick')
+  
+  ## Checks for status of wound flag in background
+  Thread.new {
+    loop {
+      sleep 3
+      $HUNTING_BUDDY.stop_hunting if Flags['noWounds'] # graceful exit
```
rpherbig: If two copies of `hunting-buddy` are running (see line 45), which copy will this 'stop'?

necroHeal.lic
```diff
@@ -0,0 +1,57 @@
...
+  
+  ## This section is designed for when running necroheal along with training manager
+  ## It allows a second copy of hunting-buddy to run on top of the current
+  ## hunting-buddy, invoke necroheal, and unpause the original hunting buddy
+
+  pause_script 'hunting-buddy' if running?('hunting-buddy') 
+  force_start_script 'hunting-buddy', [ 'heal' ]
+  
+  matchtimeout 180, "Your inventory" if running?('hunting-buddy')
+  sleep 1 until !running?('combat-trainer')
```
rpherbig: If two copies of `combat-trainer` are running (see line 45), will this loop ever end?

necroHeal.lic
```diff
@@ -0,0 +1,57 @@
...
+  Documentation: https://elanthipedia.play.net/Lich_script_development#necroheal
+  ## ** Placeming comments here until finalized and added to pedia.
+
+  Necromancer stand-alone healing script
+  
+  Uses all settings defined in HEAL YAML FILE (default)
+  Consume logic via combat trainer
+  
+  TO DO:
+  - See pull request #1462 for harvested materials.
```
rpherbig: This PR was merged

rpherbig:
Do you plan to integrate this with the `safe-room` script?

necroHeal.lic
intangibles: It's sent to the latest script running, so if there is 1 copy paused and 1 running, it goes to the running script.

intangibles:
Yeah, I'm planning on integrating it with safe-room.  I just ran into some other issues that were causing me some problems so I'm thinking about doing a full rewrite of the consume logic using common-healing.

A couple of issues I'd like on my to do list:
Less mana if nerves are damaged
Use Common healing
use a final cast on nerves or use devour if low wounds

necroHeal.lic
```diff
@@ -0,0 +1,57 @@
...
+  if DRCH.check_health.empty?
+    echo 'Not wounded, exiting heal script'
+    exit
+  end
+
+
+  ## Sets flag to check for game line to determine when no wounds exist
+  ## Flag uses consume logic from combat-training. If no wounds are left
+  ## then harvest or dissection is used. Flags trigger off these rituals.
+  Flags.add('noWounds', 'You carefully position the corpse and make a cut', 'You bend over .* to make a few quick')
```
jandersson: Is this assuming the user has harvest or dissect as their training ritual? I can see some rare case occurring where the user might not be using these.

necroHeal.lic
jandersson: would `$HUNTING_BUDDY.stop_hunting if check_health.empty?` work? that way you dont have to make a new flag

intangibles:
I'm going to take another look at this PR this week.  I've got a lot more learning under my belt since I created this PR.

Thanks!

rpherbig:
Happy anniversary!

Issue #1476, opened 2017-03-14
Rituals not being used in combat-trainer if Thanatology is locked


Morzain:
The dissect ritual trains first aid as well as thanatology.
The harvest ritual trains skinning as well as thanatology.

Ideally combat-trainer would keep using rituals until the both skills are locked (or have options for deciding this). Currently I have it setup to stop combat  when first-aid is locked but this never happens as it stops using rituals once Thanatology is locked.

intangibles:
I pushed some changes which cover this issue. See pull request #1563.

PR #1563, opened 2017-04-22
Added YAML option to allow rituals to train based on First Aid exp.

intangibles:
Added YAML option to allow rituals to train based on First Aid exp.  Current test is for thanatology exp only.

If train_first_aid is set to true, combat-trainer will continue to perform rituals until both Thanatology and First aid experience are above 30.

thanatology:
  ritual_type: dissect
  heal: false
  train_first_aid: true

Issue #1476

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -357,8 +357,7 @@ class LootProcess
 
   def should_perform_ritual(prepare_consume)
     !prepare_consume &&
-    ((@necro_train_first_aid && DRSkill.getxp('First Aid') < 32) ||
-    DRSkill.getxp('Thanatology') < 32)
+    ((@necro_train_first_aid && DRSkill.getxp('First Aid') < 32) || DRSkill.getxp('Thanatology') < 32)
   end
 
   def check_rituals?(game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -177,7 +177,7 @@ class SetupProcess
     if game_state.weapon_skill == 'Targeted Magic' && game_state.weapon_name
       bput("invoke #{game_state.weapon_name}", 'You')
       waitrt?
-    end 
+    end
   end
 end
 
@@ -355,6 +355,12 @@ class LootProcess
     end
   end
 
+  def should_perform_ritual(game_state)
+    unless game_state.prepare_consume
+      (DRSkill.getxp('Thanatology') <= 30 || (@necro_train_first_aid && DRSkill.getxp('First Aid') <= 30)) || !@skin
+    end
+  end
+
   def check_rituals?(game_state)
     return true unless DRStats.necromancer?
     mob_noun = DRRoom.dead_npcs.first
@@ -366,15 +372,7 @@ class LootProcess
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
         do_necro_ritual(mob_noun, 'consume', game_state) unless game_state.wounds.empty?
       end
-      if !game_state.prepare_consume || !@skin
-        if DRSkill.getxp('Thanatology') <= 30
-          do_necro_ritual(mob_noun, @ritual_type, game_state)
-        elsif DRSkill.getxp('First Aid') <= 30 && @necro_train_first_aid
-          do_necro_ritual(mob_noun, @ritual_type, game_state)
-        end
-      end
-      ## Old necro training block
-      #do_necro_ritual(mob_noun, @ritual_type, game_state) unless game_state.prepare_consume || (DRSkill.getxp('Thanatology') > 30 && @skin)
+      do_necro_ritual(mob_noun, @ritual_type, game_state) if should_perform_ritual(game_state)
     end
     return false if %w(consume harvest dissect).include?(@last_ritual)
     true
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -177,7 +177,7 @@ class SetupProcess
     if game_state.weapon_skill == 'Targeted Magic' && game_state.weapon_name
       bput("invoke #{game_state.weapon_name}", 'You')
       waitrt?
-    end
+    end 
   end
```
rcuhljr: You should check if your text editor has an option to remove trailing spaces, can make a mess in big change sets.

combat-trainer.lic
```diff
@@ -363,7 +366,15 @@ class LootProcess
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
         do_necro_ritual(mob_noun, 'consume', game_state) unless game_state.wounds.empty?
       end
-      do_necro_ritual(mob_noun, @ritual_type, game_state) unless game_state.prepare_consume || (DRSkill.getxp('Thanatology') > 30 && @skin)
+      if !game_state.prepare_consume || !@skin
```
rcuhljr: We can simplify this down to `unless game_state.prepare_consume && @skin`

combat-trainer.lic
```diff
@@ -363,7 +366,15 @@ class LootProcess
...
+        if DRSkill.getxp('Thanatology') <= 30
+          do_necro_ritual(mob_noun, @ritual_type, game_state)
+        elsif DRSkill.getxp('First Aid') <= 30 && @necro_train_first_aid
+          do_necro_ritual(mob_noun, @ritual_type, game_state)
+        end
```
rcuhljr: Since the content of each part of the if statement is the same we should probably condense them. for readability I'd suggest moving the checks out into a single helper method, something like

```
def should_perform_ritual
  DRSkill.getxp('Thanatology') <= 30 || (@necro_train_first_aid && DRSkill.getxp('First Aid') <= 30 )
end
```

Leaving us with 
```
do_necro_ritual(mob_noun, @ritual_type, game_state) if should_perform_ritual
```

combat-trainer.lic
```diff
@@ -363,7 +366,15 @@ class LootProcess
...
+      end
+      ## Old necro training block
+      #do_necro_ritual(mob_noun, @ritual_type, game_state) unless game_state.prepare_consume || (DRSkill.getxp('Thanatology') > 30 && @skin)
```
rcuhljr: Go ahead and kill the commented out code.

intangibles:
The old logic is to always use rituals if skinning is set to false.  I'm not sure if this is necessary but i left intact.

combat-trainer.lic
```diff
@@ -352,6 +355,12 @@ class LootProcess
     end
   end
 
+  def should_perform_ritual(game_state)
+    unless game_state.prepare_consume
+      (DRSkill.getxp('Thanatology') <= 30 || (@necro_train_first_aid && DRSkill.getxp('First Aid') <= 30)) || !@skin
+    end
```
rcuhljr: Man the logic here is becoming a trainwreck. I'm not even sure why we care if they're skinning. That will happen anyways. I vote we raise the checks to 32 instead of 30 and just ignore the whole skin requirement. That leaves us with a much saner.

```
  def should_perform_ritual(game_state)
    !game_state.prepare_consume &&
    !(@necro_train_first_aid && DRSkill.getxp('First Aid') <= 32)) &&
    DRSkill.getxp('Thanatology') <= 32
  end
```

Now it scans much better as You should perform the ritual if you're not preparing consume, and your not trying to train FA while it's not locked and you're not thanatology locked.

intangibles: Agreed, the logic is a bit of a mess. My brain was hurting there for awhile.

So the reason for the change is that it wasn't training First Aid if Thanatology was locked.  So my goal is to have it perform the ritual if EITHER are not locked.  I'd also want to leave the training threshhold at 31 since you can get 2.x mindstates from rituals.

rpherbig:
@jandersson Can you take a look at this PR too?

combat-trainer.lic
rcuhljr: I considered making it default, but I decided to bench that discussion for now. I think 31 is fine if we drop the @skin check and use the simplified logic which will keep doing rituals while either skill needs training.

combat-trainer.lic
```diff
@@ -352,6 +355,12 @@ class LootProcess
     end
   end
 
+  def should_perform_ritual(game_state)
+    unless game_state.prepare_consume
```
jandersson: I think `game_state.casting_consume` also should be considered, as when thats true youre waiting for the heal spell to become fully prepped. maybe `unless game_state.prepare_consume || game_state.casting_consume`

jandersson: Otherwise I think this will try to do a training ritual on the target for the healing spell

intangibles: Unless confuses me too.  I'm going to remove that with the simplified logic suggested above.

I think we should stick with prepare_consume as well.  That's what the old code was using and I haven't heard of any errors/issues so might be best to keep that part consistent.

combat-trainer.lic
```diff
@@ -352,6 +355,12 @@ class LootProcess
     end
   end
 
+  def should_perform_ritual(prepare_consume)
+    !prepare_consume &&
+    ((@necro_train_first_aid && DRSkill.getxp('First Aid') < 32) ||
```
rcuhljr: This should be `&&` on the end of the line, not `||`.

rcuhljr: And you can remove the outer wrap of Parens then.

PR #1564, opened 2017-04-22
Adding support for bound/hurlable offhands.

jayRyan24:
GETting a thrown/lobbed weapon will automatically put it in the hand it was thrown from, so added support for swap's when invoked!


jayRyan24:
Uh, fixing the obvious mistake...

jayRyan24:
OK, fixed and tested the obvious bug.

PR #1567, opened 2017-04-23
Steal.lic Update

Endomlik:
Selects stealing targets based on user's hometown selection in their yaml.

Changed the build_new function variable province to steal_town.  The variable was static before and did not change.  I did not find any place where the province variable was being utilized.  Due to this change in granularity and variable name change, base-stealing.yaml would need to be updated with new values.  I used find and replace to change all of the "province: zoluren" to "steal_town: Crossing" in the base-stealing.yaml.   

Thief binning changed to walk_to the bin location pulled from base-town.yaml.  Crossing is 9080 and Muspar'i is 12283 work.   I don't know if the other locations are setup to work with go2 but that is another issue that can be worked out.

Due to the base-stealing.yaml daily update, a separate pull request will be made once steal.lic this merge is deemed ready to be merged.  Here is an example:   https://pastebin.com/gWDKKvPT


rcuhljr:
Are we worried about overlap in stealing towns? namely Riverhaven/Crossing?

steal.lic
```diff
@@ -161,7 +161,7 @@ class Steal
     end
 
     {
-      'new' => true, 'room' => Room.current.id, 'province' => 'zoluren',
+      'new' => true, 'room' => Room.current.id, 'steal_town' => @steal_town,#steal_town replaced province
```
rcuhljr: I'd like to be able to upload stealing info without changing my hometown, it feels like this should be an argument.

steal.lic
```diff
@@ -18,16 +18,20 @@ class Steal
 
     UserVars.stealing_timers ||= {}
 
-    @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
+    @stealing_options = get_data('stealing').stealing_options
+	  .reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
+	  .reject { |target| !settings.hometown.include?(target['steal_town'])} #added for to specify towns
```
rcuhljr: We can move @hometown declarations above this to shorten up the line, and use `==` instead of `include?` which is somewhat awkward when we just want exact string comparison.

steal.lic
Endomlik: I like the idea of an argument.  I'll see what I can do.

Endomlik:
The more I think about the less I like the idea of the stealing being directly tied to hometown.  The stealing base for Crossing is very built up and there may be people want to raid the shelves there yet have their hometown set somewhere else.

Endomlik:
I fubar'd this shit.

Issue #1252, opened 2017-01-05
Bug on rope bridge on way to Sky Giants


drgabellia:
When traveling across the rope bridge to Sky Giants, escort (see log below) exited before reaching the destination room:
[Mistwood Forest, North of a Ravine]
A steep-sided ravine breaches the forest floor, leaving a wide gash spanned by a pair of ropes tied to trees on either side.  It's hard to dignify the arrangement by calling it a bridge, but it looks like the only way across.
Obvious paths: west, northwest.
Room Number: 8650

I've never had an issue before.....and I'm wondering if it has anything to do with prebuff, as room 8650 is my prebuff room?

When I returned to my computer, I was lying prone on the bridge.  I had gotten stunned (probably because I stood there so long) which forced me to lay down.  I was unable to stand using the stand command.  Finally figured out ONLY shuffle <direction> works on the bridge.  Log below.

You will notice the only script active was buff....which shouldn't of been, as I was not in the proper room.

~Gab

>
You climb onto a rope and begin to make your way across.
[On a Rope Bridge, Above a Ravine]
The ravine floor is far below, a painful fall if you should lose your grip.  With no other way to cross, you can only shuffle along, moving first one hand, then a foot, and hope the ropes will hold.  The south side of the ravine is almost within your reach.  You also see a rope leading north and a rope leading south.
Obvious paths: none.
Roundtime: 30 seconds
>
A runt-sized farm pig trots into the room.
>
Your body is dripping with water.  Pools of it are forming all around you.
>
[escort]>shuffle north
...wait 1 seconds.
>
[escort]>shuffle north
>
The wind whirls around you like a mad thing, making your footing decidedly unsure.  You tighten your fingers around the rope, clinging to your precarious support.
>
You shuffle along a rope leading north.
[On a Rope Bridge, Above a Ravine]
The bottom rope sags beneath your weight as you struggle to cross, halfway between the north and south edges of the ravine.  Far below, broken rocks litter the ravine floor and promise a painful landing.  You also see a rope leading north and a rope leading south.
Obvious paths: none.
Roundtime: 20 seconds
>
A runt-sized farm pig trots into the room.
>
Fitful sparks brighten and dim deep within the sky blue svelae on your forehead, swirling in a restless dance around its intaglio rainbow-hued butterfly orchid.
>
[escort]>shuffle north
You shuffle along a rope leading north.
[On a Rope Bridge, Above a Ravine]
[...]

PR #1570, opened 2017-04-23
Add Theren rope bridge to bescort (#1252)

jamesbw2001:
Will update map-db once this is added. This will prevent people people from getting stuck on the rope bridge, as the current escort.lic implementation is prone to do.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -138,7 +138,11 @@ class Bescort
       [
         { name: 'basalt', regex: /basalt/i, description: 'The ferry between Ratha and Necromancer Island' },
         { name: 'mode', options: %w(island ratha), description: 'Where do you need to get to?' }
-      ]
+      ],
+	  [
+        { name: 'therenropebridge', regex: /therenropebridge/i, description: 'The rope bridge between Theren and Rossman\'s Landing' },
+        { name: 'mode', options: %w(totheren torossman), description: 'Where do you need to get to?' }
+	  ]
     ]
 
     args = parse_args(arg_definitions)
@@ -165,6 +169,8 @@ class Bescort
       iceroad(args.mode)
     elsif args.basalt
       take_crawling_plague(args.mode)
+	elsif args.therenropebridge
+      take_theren_rope_bridge(args.mode)
     end
   end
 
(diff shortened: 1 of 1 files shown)


bescort.lic
```diff
@@ -584,6 +590,64 @@ class Bescort
...
+	end
+	
+	start_time = Time.now.to_f
+	until Room.current.id == @target_room or Time.now.to_f - start_time > 300
+		bput("shuffle #{@shuffle_direction}", "Roundtime")
+		sleep(3)
+		waitrt?
+	end
+	if Time.now.to_f - start_time > 300
+		echo("\a")
```
rpherbig: You can replace these echoes with `beep` if you think that makes the code clearer

bescort.lic
```diff
@@ -584,6 +590,64 @@ class Bescort
...
+			sleep(0.1) until Flags["rope_wait"]
+			Flags.reset("rope_wait")
+		else
+			sleep(2)
+			waitrt?
+		end
+	end
+	
+	start_time = Time.now.to_f
+	until Room.current.id == @target_room or Time.now.to_f - start_time > 300
```
rpherbig: You can compare `Time.now` directly to 300 (the `to_f` is not necessary)

bescort.lic
```diff
@@ -584,6 +590,64 @@ class Bescort
...
     end
   end
+  
+  def take_theren_rope_bridge(mode)
+	theren_side_roomid = 8650
+	rossman_side_roomid = 8637
+	@target_room = false
+	@shuffle_direction = false
+    if mode == 'totheren'
+		if not Room.current.id == rossman_side_roomid
```
rpherbig: I prefer `if Room.current.id != rossman_side_roomid` if you're ok with that

bescort.lic
```diff
@@ -584,6 +590,64 @@ class Bescort
       end
     end
   end
+  
+  def take_theren_rope_bridge(mode)
+	theren_side_roomid = 8650
+	rossman_side_roomid = 8637
+	@target_room = false
```
rpherbig: This (and line 598) don't need to be instance vars (i.e. you can remove the `@` from the name).

bescort.lic
```diff
@@ -584,6 +590,64 @@ class Bescort
...
+		exit
+	end
+	
+	start_room = Room.current.id
+	Flags.add("rope_wait","finally arriving on this side","finally reaching the far side")
+	while Room.current.id == start_room
+		put("retre") # RT irrelevant
+		put("retre")
+		Flags.reset("rope_wait") # avoiding race conditions
+		ret = bput("climb rope", "Roundtime", "is already on the rope", "You climb onto a", "You are engaged")
```
rpherbig: We typically use a `case` block when checking output from bput. It tends to result in cleaner/clearer code that is easier to modify later, but it's not a big deal.

rpherbig:
Would you be able to run `rubocop --auto-correct bescort.lic`? That will automatically apply formatting changes.

bescort.lic
rcuhljr: As a general style and safety guideline we prefer `||` over `or` since they have different operator precedence. It's probably also worth parentheses around the time difference just to make precedence explicit.

bescort.lic
jamesbw2001: I know that's present in a lot of places in dr-scripts, but I've gotten in the habit of not doing that since I often use more advanced string operations on the result than just === or nest logic on it. I also usually don't bother when I only check one thing. Granted, in this case, I may as well case it.

jamesbw2001:
@rpherbig Comments addressed, including rubucop on the whole file (including the scope creep that entails).

bescort.lic
```diff
@@ -585,6 +589,63 @@ class Bescort
...
+      target_room = rossman_side_roomid
+    else
+      echo("Unrecognized argument: #{mode}")
+      exit
+     end
+
+    start_room = Room.current.id
+    Flags.add('rope_wait', 'finally arriving on this side', 'finally reaching the far side')
+    while Room.current.id == start_room
+      put('retre') # RT irrelevant
```
rcuhljr: One of the common files has a retreat helper in it.

jamesbw2001:
@rcuhljr Done

PR #1577, opened 2017-04-25
Added LW cast messaging

izawwlgood:
Still need Arcane mana


data/base-spells.yaml
```diff
@@ -110,6 +110,10 @@ cast_messages:
 - You swipe your fingertips
 - ^Your spirit lashes out
 - You open your arms in a hieratic gesture
+- A white static
```
rpherbig: What do you think about `A (white|silver) static` and `A (glowing amber|crimson) mist`?

izawwlgood:
I didn't know the syntax. Sounds good!

PR #1583, opened 2017-04-26
Purchase Deed Packets Using Hometown Data

jandersson:
This loads the town data and looks for `deed-room` and `deed-number` in the blacksmithing section, then runs off to purchase like usual. If they dont exist for the town then the script exits.

### DIFF
--- a/data/base-crafting.yaml
+++ b/data/base-crafting.yaml
@@ -51,6 +51,8 @@ blacksmithing:
     repair-npc: clerk
     stock-volume: 5
     stock-room: 8775
+    deeds-room: 8775
+    deeds-number: 14
     stock-number: 11
     stock-name: bronze ingot
     trash-room: 8773
@@ -126,6 +128,8 @@ blacksmithing:
     finisher-full: flask of oil
     finisher-room: 4436
     finisher-number: 6
+    deeds-room: 4433
+    deeds-number: 2
     crucibles:
       - 9611
       - 9613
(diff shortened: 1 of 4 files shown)


rpherbig:
Why did you add this to the blacksmithing section and not a (new?) mining section?

jandersson:
No special reason, just that they're in the forging society. Want me to making a mining section?

jandersson:
I'm going to add this to a `deeds:` section. I think this is appropriate because deeds are decoupled from crafting disciplines and even materials. You can deed items, wood, and ore, so i think its fitting as its own section. Let me know what you think

rpherbig:
That makes sense to me

data/base-crafting.yaml
```diff
@@ -126,6 +126,8 @@ blacksmithing:
     finisher-full: flask of oil
     finisher-room: 4436
     finisher-number: 6
+    deeds-room: 4433
```
rpherbig: Should this be removed now?

jandersson: oops, yes it should

mining-buddy.lic
```diff
@@ -32,7 +32,17 @@ class MiningBuddy
   end
 
   def buy_deed_packet
-    order_item(8775, 14)
+    hometown = get_settings.hometown
```
rpherbig: What do you think about moving these few lines (just the settings part) to `setup`?

jandersson: thats fine with me, pass the room number and order number as arguments to `buy_deed_packet`?

rpherbig: Eh, I'd just make them instance vars like the others.

jandersson: Yep, on second thought that looks easier

PR #1590, opened 2017-04-30
WIP PR: Allow ;afk to use slackbot

rpherbig:
Work in progress (there's a few debug statements and some commented code still in there). Looking for feedback on the approach.

cc @kamarcum @rcuhljr @jandersson

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -8,8 +8,9 @@ no_pause_all
 no_kill_all
 
 def exit_game
-  echo "Current room: #{Room.current.id}"
-  echo "Exiting at #{Time.now}"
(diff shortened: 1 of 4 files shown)


jandersson:
Awesome, does the manager class in dependency mean we can utilize the bot without trusting the script  using it? I had this hang-up with a previous attempt at using a bot in `afk`, where including the bot meant the script needed to be trusted.

rpherbig:
Correct, afk would not need to be trusted here

rcuhljr:
Send messages internal message storage should be an array probably, this prevents race conditions from erasing messages. Also remove your debugging echos :)

Issue #637, opened 2016-08-19
Rucksack Woes part 2


drboldwin:
It looks as though tieing to rucksack was addressed in combat but not untieing?

http://pastebin.com/Z87dCBVw

Example of failed untie

PR #1592, opened 2017-04-30
Brought equipmanager.lic up to forked repo's current, reapplied fixes.

NBSL:
Change to fix issue #637


rpherbig:
:+1: Looks good to me. @rcuhljr any comments?

rcuhljr:
Assuming it was tested, not really. There's some stylistic stuff (Array for single element match in when statement) but nothing major.

PR #1593, opened 2017-05-02
Fix error

tulanegator:
mining buddy gave errors after new update. this was the issue.


rpherbig:
Nice catch. Thank you for the fix.

PR #1594, opened 2017-05-02
Allow untrusted scripts to use slack

rpherbig:
cc @rcuhljr @kamarcum @jandersson

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -700,6 +700,54 @@ class ReportbotManager
   end
 end
 
+class SlackbotManager
+  def initialize(debug)
+    @debug = debug
+    @slackbot = nil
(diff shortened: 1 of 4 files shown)


dependency.lic
```diff
@@ -700,6 +700,54 @@ class ReportbotManager
...
+    @need_to_register = false
+    @messages = []
+  end
+
+  def run_queue
+    register_slackbot
+    send_messages
+  end
+
+  def register
```
rpherbig: I considered making `register` take the username as an arg... that would mean we don't have to pass it around in the calling scripts, but could lead to a race condition if different scripts want to use a different username. Thoughts?

jandersson:
:tada:

PR #1603, opened 2017-05-08
Update to Thanatology Training

intangibles:
Added new "cycle" ritual which swaps between Harvest and Dissect based on Skinning/First Aid Training.  
Made changes to default ritual training behavior and removed first aid exp variable. 
Now checks for FA exp if using Dissect, or skinning if using harvest.


intangibles:
It looks like my combat-trainer was a bit out of date when i made the changes.  Is there a way to isolate those changes?

intangibles:
Well, I had some version errors, tried to fix and now combat-trainer.lic is updated as of 5.15.2017.

PR #1606, opened 2017-05-09
Update Remedy.lic to specify my mortar.

thejko:
Preview version would hang if there was another mortar on the ground or in a trash bin (often deposited there from lockpicking).


rpherbig:

PR #1613, opened 2017-05-11
Added an offline mode argument to allow it to be ran when the json is invalid

NBSL:
Quite a few changes.

Gives an offline mode option that pulls the json data down to get the object setup, then reads the perc moon|moons output and builds times based off of their degrees or anlas/anlaen remaining.

Also fixes the issue Myself and Krystorm experienced yesterday with the alias returning stale times.

I tried to leave as much of the original functions alone, only changing the minute_apart to use .ceil in an attempt to fix the timer jumping ahead 2 minutes instead of 1.

Timer auto updates ever 5 minutes.

The string matches aren't set to promise accuracy. My moonie is only 11, so he isn't able to get "You feel certain that moon is # degrees above the (eastern|western) horizon." messages on anything but when the moon first rises and just before it sets.

### DIFF
--- a/moonwatch.lic
+++ b/moonwatch.lic
@@ -172,7 +172,7 @@ def process_moon_message(moon, message)
     echo("should rise #{moon}:#{Regexp.last_match(1)}") if $debug_mode_mm
     offset = Regexp.last_match(1) ? Regexp.last_match(1).to_i : 1
     echo("offset: #{offset}") if $debug_mode_mm
-    timer = (Settings['set'][moon] - (offset * 30 * 60))
+    timer = offset * 30 * 60
     echo("timer: #{moon}:#{timer / 60}") if $debug_mode_mm
   elsif message =~ /sets/
     timer = (Settings['rise'][moon])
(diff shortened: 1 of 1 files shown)


rcuhljr:
This file previously had all of this logic in it, you can go back and pull it for more succinct handling of angle conversions, as well as handling zenith and exact rise situations. My only other concern is I'd rather not patch in the weird alias work around instead of trying to figure out what is going wrong with your alias setup, that's a symptom of something messed up on a deeper level for you guys.

[Here's an older version with that logic still in it.](https://github.com/NBSL/dr-scripts/blob/bb2e0318e50c3745e992b4082a38011cb735b1ef/moonwatch.lic#L71-L103)

I'm still hoping someone will take the time to just fix the dying connections to firebase so that this all becomes unnecessary.

NBSL:
This is what I get with normal moonwatch (current version in the repo)

[https://pastebin.com/gLLwySCu](https://pastebin.com/gLLwySCu)

 And my alias list (The difference in lich character is because I was running Outlander, but the output to window is so slow for me it was taking ~20 minutes to the ;killall command to register):

;alias list

Global Aliases:

   as => accept \r stow \?
   cb => ,e $temp = []
   cc => ,circlecheck
   ct => ,eq Script.running?('crossing-training') ? kill_script('crossing-training') : start_script('crossing-training')
   dm => ,repos download-mapdb
   fr => ,e echo Room.list.find{|room| room.description.find{|item| item =~ /?/i}}
   fz => ,en echo get_data('hunting').hunting_zones.find{|k,v| k =~ /?/i}
   gs => get \? \r stow \?
   hz => ,en get_data('hunting').hunting_zones.sort.each{ |x| echo x }
   ids => ,e echo "#{Room.current.id}:#{Room.current.wayto}"
   lr => ,e echo Room[Room.current.id].inspect
   map => ,narost
   moon => ;eq respond("katamba is up for 99 minutes : yavash will rise in 49 minutes : xibar will rise in 124 minutes")
   rec => ,e echo $temp << Room.current.id
   rm => ,e Map.reload
   tm => ,eq Script.running?('training-manager') ? kill_script('training-manager') : start_script('training-manager')

Forea's Aliases:

   (none)

rcuhljr:
What version of Ruby are you running? Your scripts aren't evaluating correctly and it's not creating the alias correcetly as you can see from the alias list. Your alias should display as `  moon => ;eq respond("#{UserVars.moons['katamba']['pretty']} : #{UserVars.moons['yavash']['pretty']} :  #{UserVars.moons['xibar']['pretty']}")` If this isn't running correctly on your machine you're gonna be getting systemic errors that are hard to detect.

NBSL:
2.2.5

Everything was intially setup following: https://gswiki.play.net/Mac_Installation_-_Lich_(software)

NBSL:
I just talked to Kry and he's using 2.3.3 on Fedora

rcuhljr:
Simpler solution, the alias command got broken 5 months ago when being fixed for genie mode, and no one ever reported it breaking. There's a one line fix for the alias method, it should be 
` UpstreamHook.run("<c>#{$clean_lich_char}"+ 'alias add --global moon = #{$clean_lich_char}eq respond(\"#{UserVars.moons['katamba']['pretty']} : #{UserVars.moons['yavash']['pretty']} : #{UserVars.moons['xibar']['pretty']}\")')

NBSL:
Ah. Thank you. That reminds me, I need to submit a PR for setup-alias. The HZ and FZ commands use ;en which is only in the forked version.

rcuhljr:
It's a lot more likely that we're moving dependency solely over to the forked version and removing it as a toggle.

NBSL:
I'll leave it alone then :)

NBSL:
Ok, so I had to change your alias code a bit to get it to work correctly. I incorporated the old moonwatch code in. It still mostly functions the same, besides the changing of check_for_new_moons and moon_change methods.

rcuhljr:
I pushed a fix that should solve the moon timers problem at the root level, meaning you won't lose connection over time.

PR #1629, opened 2017-05-16
Update to Thanatology Training v2 (Skinning/First Aid Ritual Cycle)

intangibles:
Added new "cycle" ritual which swaps between Harvest and Dissect based on Skinning/First Aid Training.
Made changes to default ritual training behavior and removed first aid exp variable.
Now checks for FA exp if using Dissect, or skinning if using harvest.

See #1563 for prior PR.  Calling it V1.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -213,8 +213,7 @@ class LootProcess
     @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
 
-    @cycle_rituals = false
-    @cycle_rituals = true if @ritual_type == "cycle"
+    @ritual_type == "cycle" ? @cycle_rituals = true : @cycle_rituals = false
 
     @rituals = get_data('spells').rituals
     echo("  @rituals: #{@rituals}") if $debug_mode_ct
@@ -361,19 +360,14 @@ class LootProcess
     DRSkill.getxp('Thanatology') < 32)
   end
 
-  def check_ritual_xp
-    DRSkill.getxp('First Aid') > 31 && DRSkill.getxp('Skinning') > 31
-  end
-
   def cycle_rituals
-    @ritual_type = "preserve"
-    @ritual_type = "dissect" if DRSkill.getxp('First Aid') < 32
+    return unless @cycle_rituals 
+    @ritual_type = "dissect"
     @ritual_type = "harvest" if DRSkill.getxp('Skinning') < DRSkill.getxp('First Aid')
-    @ritual_type = "preserve" if check_ritual_xp
+    @ritual_type = "preserve" if DRSkill.getxp('Skinning') > 31
   end
 
-  def check_rituals?(game_state)
-    cycle_rituals if @cycle_rituals && !game_state.prepare_consume
+  def check_rituals?(game_state) 
     return true unless DRStats.necromancer?
     mob_noun = DRRoom.dead_npcs.first
     return true if game_state.construct?(mob_noun)
@@ -384,6 +378,7 @@ class LootProcess
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
         do_necro_ritual(mob_noun, 'consume', game_state) unless game_state.wounds.empty?
       end
+      cycle_rituals
       do_necro_ritual(mob_noun, @ritual_type, game_state) if should_perform_ritual(game_state.prepare_consume)
     end
     return false if %w[consume harvest dissect].include?(@last_ritual)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -213,6 +213,9 @@ class LootProcess
     @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
 
+    @cycle_rituals = false
+    @cycle_rituals = true if @ritual_type == "cycle"
+
     @rituals = get_data('spells').rituals
     echo("  @rituals: #{@rituals}") if $debug_mode_ct
 
@@ -221,9 +224,6 @@ class LootProcess
 
     @necro_heal = thanatology['heal'] || false
     echo("  @necro_heal: #{@necro_heal}") if $debug_mode_ct
-
-    @necro_train_first_aid = thanatology['train_first_aid'] || false
-    echo("  @necro_train_first_aid: #{@necro_train_first_aid}") if $debug_mode_ct
 
     @necro_store = thanatology['store'] || false
     echo("  @necro_store: #{@necro_store}") if $debug_mode_ct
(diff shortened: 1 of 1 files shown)


Morzain:
As far as I can tell this works great with no problems

intangibles:
Great! I've been running for about a week now and haven't add any issues.

I've got a couple more items on my wishlist that I want to add in but the code is a bit more complicated (I think), that maybe you guys can chime in on.

For example, if your Thanatology is maxed, but skinning is not, and you are hunting a skinnable creature, then it would be better to arrange/skin instead of performing harvest because it teaches far better.

Right now rituals default to preserve which allows skinning but this only triggers if both skinning and thanatology are max exp.  I think I would need to pass mob status and skinnable status through when checking for ritual exp. But a lot of the skinning checks happen AFTER the ritual to make it slightly more complicated.

intangibles:
@jandersson

Can you take a poke at this? And maybe see my last comment on viability of adding a skinning check?

jandersson:
Will do

combat-trainer.lic
```diff
@@ -213,6 +213,9 @@ class LootProcess
     @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
 
+    @cycle_rituals = false
```
rpherbig: I don't think you need a separate variable for this. You would not reassign `@ritual_type`, but instead pass it around as an argument.

combat-trainer.lic
```diff
@@ -213,6 +213,9 @@ class LootProcess
...
+    @cycle_rituals = true if @ritual_type == "cycle"
```
rpherbig: You can simplify these two lines to `@cycle_rituals = @ritual_type == "cycle"`

combat-trainer.lic
```diff
@@ -355,11 +355,25 @@ class LootProcess
...
+
+  def cycle_rituals
+    @ritual_type = "preserve"
+    @ritual_type = "dissect" if DRSkill.getxp('First Aid') < 32
+    @ritual_type = "harvest" if DRSkill.getxp('Skinning') < DRSkill.getxp('First Aid')
+    @ritual_type = "preserve" if check_ritual_xp
   end
 
   def check_rituals?(game_state)
+    cycle_rituals if @cycle_rituals && !game_state.prepare_consume
```
rpherbig: I recommend putting the conditionals inside the `cycle_rituals` function. This would reduce the code in `check_rituals?` and make the flow easier to understand. What do you think?

combat-trainer.lic
```diff
@@ -355,11 +355,25 @@ class LootProcess
...
+    ((@ritual_type == "dissect" && DRSkill.getxp('First Aid') < 32) ||
+    (@ritual_type == "harvest" && DRSkill.getxp('Skinning') < 32) ||
+    DRSkill.getxp('Thanatology') < 32)
+  end
+
+  def check_ritual_xp
+    DRSkill.getxp('First Aid') > 31 && DRSkill.getxp('Skinning') > 31
+  end
+
+  def cycle_rituals
```
rpherbig: I think a lot of this logic can be merged with `should_perform_ritual`. The combined method would be something like `determine_ritual` and return whatever is appropriate based on cycle/no-cycle and experience. `nil` (or empty string) could represent 'do not do a ritual' (i.e. the `false` value right now).

combat-trainer.lic
intangibles: Yeah, I think you are right here.  I initially wrote a second variable thinking that @ritual_type would change based on the ritual and I would lose track of the cycle status. But it doesn't appear to do that.  I'll make some changes based on this and see how they work.  Thank you!

intangibles: Well, I ran through some changes and I need to keep a second variable.  @Ritual_type overrites itself to the preferred ritual so cycle is no longer available to test. It would lose the cycle function at that point.  I can either set cycle at the beginning or do it within cycle rituals itself.   Setting the variable at the beginning seems to be the cleanest way.

combat-trainer.lic
intangibles: I can clean try to clean it up a bit but I don't think its a good idea to combine them.  The consume logic is already pretty messy.  If I layered the cycle on top of that it would be even harder to manage.

Plus, not everyone will want to use the cycle ritual (unless maybe you are suggesting additional default behavior) so I don't really see that we would need to force the cycle calculations by combining everything into one piece.

rpherbig: I'm not suggesting you make everyone cycle rituals. Or, rather, another way to look at the current logic is that we're already cycling between, say, 'preserve' and 'don't do a ritual' (in the case where Thanatology is locked). I think the new logic is just an extension of what we already have.

combat-trainer.lic
```diff
@@ -213,6 +213,8 @@ class LootProcess
     @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
 
+    @ritual_type == "cycle" ? @cycle_rituals = true : @cycle_rituals = false
```
rcuhljr: Could we simplify this to `@cycle_rituals = @ritual_type == "cycle"`?

intangibles: Updated!

combat-trainer.lic
```diff
@@ -222,9 +224,6 @@ class LootProcess
     @necro_heal = thanatology['heal'] || false
     echo("  @necro_heal: #{@necro_heal}") if $debug_mode_ct
 
-    @necro_train_first_aid = thanatology['train_first_aid'] || false
```
rpherbig: If you are removing this setting, you'll probably want to add a validator that checks if it is `true` and warns the user they need to change their settings.

combat-trainer.lic
```diff
@@ -355,8 +354,17 @@ class LootProcess
   end
 
   def should_perform_ritual(prepare_consume)
-    !prepare_consume &&
-      ((@necro_train_first_aid && DRSkill.getxp('First Aid') < 32) || DRSkill.getxp('Thanatology') < 32)
+   !prepare_consume &&
```
rpherbig: Would this be clearer if you break it up into multiple `return`s? Similar to line 364.

combat-trainer.lic
```diff
@@ -213,6 +213,9 @@ class LootProcess
     @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
 
+    @cycle_rituals = false
+    @cycle_rituals = @ritual_type == "cycle"
```
rpherbig: Why is `@cycle_rituals` set to false right before re-assigning to it?

intangibles: Well, because it's not initiated before. If it's not cycle it's not initiated or nil? I'm not sure the interactions with nil so i figured false would be better.

combat-trainer.lic
intangibles: Hmmm.  So this is returning true/false.

Are you thinking something like

return false if prepare_consume

combat-trainer.lic
rpherbig: You don't need to initialize it before assigning to it. The `false` will be immediately overwritten. If the ritual type is _not_ 'cycle', then it will still do the assignment, but with a `false` value (i.e. `@cycle_rituals` will always have a value after the comparison is done).

combat-trainer.lic
intangibles: Have any examples off hand of what this would look like?

rpherbig: https://github.com/rpherbig/dr-scripts/blob/master/validate.lic#L414-L419 is probably pretty close to what you'd want

combat-trainer.lic
rpherbig: Yep, that's what I had in mind

combat-trainer.lic
intangibles: Oh. Derp. That makes sense. @ritual_type == "cycle" returns true or false which is then assigned to @cycle_rituals via =.  Does that sound accurate?

rpherbig:
You got it

PR #1630, opened 2017-05-16
Added Timer to Crossing Repair

intangibles:
Added Timer to Crossing Repair.  Timer can be set using YAML setting "repair_timer: 100". (or any number)

Updated base.yaml to have a default timer of 10 minutes (600 seconds) between repair attempts.


rpherbig:
@intangibles Please update elanthipedia with the setting information.

intangibles:
Updated epedia with the information.

PR #1643, opened 2017-05-19
YAML setting to skip Crossing-Repair

intangibles:
Added YAML setting to completely skip crossing-repair. Useful for people who don't want to repair, (and necromancers who blow their SO!)

I also figured this was a good time to plug in a setting for necromancers to avoid functions that could cause a social outrage spiral.

Setting is:
high_social_outrage: true/false
Default is set to false.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -81,6 +81,7 @@ buff_nonspells:
 dance_actions_stealth:
 thanatology:
 necromancer_healing:
+high_social_outrage: false
 weapon_training:
 combat_trainer_target_increment: 3
 combat_trainer_action_count: 10
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -81,7 +81,6 @@ buff_nonspells:
 dance_actions_stealth:
 thanatology:
 necromancer_healing:
-high_social_outrage: false
 weapon_training:
 combat_trainer_target_increment: 3
 combat_trainer_action_count: 10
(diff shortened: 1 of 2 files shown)


rpherbig:
Per #1645, I thought repairing was _NOT_ affected by SO...

intangibles:
Yeah @rpherbig.  But currently crossing-repair uses exchange/deposit at the bank!  So not sure if we would rather skip repair altogether or just skip the deposit/exchange functions.

Morzain:
All I can tell you for sure is I'm in a bit of a SO spiral from trying to use ;training after being accused yesterday lol.

it successfully repaired all my gear but later on when it went to deposit into the bank I got clipped...as well as when ;pick tried to buy lockpicks.

jayRyan24:
It seems to me that these settings do the exact same thing. What is the purpose of both of them? I can see it useful for people that don't want to repair each iteration, but two switches seems pointless. If the necro has high social outrage, they can just turn off the repairing. I could see the SO toggle being used if it got flagged automatically and wasn't a YAML setting.

intangibles:
I'll remove the SO setting. It was a placeholder for a future toggle but doesn't need to be part of this PR.

intangibles:
@rpherbig @rcuhljr 
Can either of you gentlemen push this through? I've removed the SO pieces and it is a simple repair toggle now.

Thanks!

Issue #1636, opened 2017-05-18
Add the Shard-Leth Gondola to Bescort


jandersson:
The wizard script, `escort`, which handles this crossing is notorious for breaking, leaving the user idle on the gondola.

jandersson:
The updated `StringProc` will need to be updated to handle the case where the user doesn't have dependency installed (the script `bescort` wont exist). If it doesn't exist, I think either it should call `escort` or give them an error message telling them to install `dependency`.

reithir:
Just a little more testing, it's looking like this is happening now basically every time I try to cross the gondola. <g>. And I do have dependency installed, should it still be all wonky for me?

jandersson:
No it won't help at the present, unfortunately. The crossing is handled entirely by `escort`. Once someone writes the functionality into `bescort` then having it installed will alleviate all the issues.

jandersson:
Adding a log to help whomever with writing
```
[Obsidian Pass, Chasm's Ridge]                                                                                                                                                                                   
A "room-in-progress" is rising around a large paved area as stone blocks are being mortared into place, enclosing the granite pillars anchoring the gondola ropes.  An oak door is already hung at one end of the
  room, and through a more elaborate frame at the north end, the gondola platform can be seen.                                                                                                                   
Obvious paths: none.                                                                                                                                                                                             
Room Number: 2903                                                                                                                                                                                                
[go2]>go frame                                                                                                                                                                                                   
[Obsidian Pass, Platform]                                                                                                                                                                                        
The greying wooden boards of the platform have been replaced by paving stones.  Two massive granite pillars offer support for a set of thick rope cables that allow the gondola served by this platform to cross 
  the chasm.  Stacks of building stones and mortar lie off to one side, ready to add to the structure being raised around the landing.  You also see a door frame.                                               
Obvious paths: none.                                                                                                                                                                                             
Room Number: 2904                                                                                                                                                                                                
[go2]>go gondola                                                                                                                                                                                                 
There is no wooden gondola here.  You'll have to wait for it to come back around.                                                                                                                                
>                                                                                                                                                                                                                
[go2: waiting for the gondola...]                                                                                                                                                                                
>                                                                                                                                                                                                                
The ropes begin to creak into motion.                                                                                                                                                                            
>                                                                                                                                                                                                                
The gondola arrives at the center of the chasm, and keeps heading south.                                                                                                                                         
>                                                                                                                                                                                                                
[...]

PR #1648, opened 2017-05-19
Add gondola to bescort

jandersson:
Add Shard gondola to bescort. Addresses #1636

I'll add these two lines to the mapdb if merged. Note that I'm keeping the existing logic in the StringProc, but deferring to bescort if the user has it. 
```
Room[2249].wayto["2904"] = StringProc.new("if Script.exists?('bescort'); start_script('bescort', ['gondola', 'south']); else; result = dothistimeout 'go gondola', 10, /^There is no wooden gondola here.| Gondola, Cab North/; unless result =~ /Gondola, Cab North/; echo 'waiting for the gondola...'; waitfor 'The gondola stops on the platform and the door silently swings open.'; move 'go gondola'; end;  move 'south'; waitfor 'With a soft bump, the gondola comes to a stop at its destination.'; move out; end")

Room[2904].wayto["2249"] = StringProc.new("if Script.exists?('bescort'); start_script('bescort', ['gondola', 'north']); else;result = dothistimeout 'go gondola', 10, /^There is no wooden gondola here.|Gondola, Cab South/; unless result =~ /Gondola, Cab South/; echo 'waiting for the gondola...'; waitfor 'The gondola stops on the platform and the door silently swings open.'; move 'go gondola'; end; move 'north'; waitfor 'With a soft bump, the gondola comes to a stop at its destination.'; move out; end")
```

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -140,6 +140,10 @@ class Bescort
       [
         { name: 'therenropebridge', regex: /therenropebridge/i, description: 'The rope bridge between Theren and Rossman\'s Landing' },
         { name: 'mode', options: %w[totheren torossman], description: 'Where do you need to get to?' }
+      ],
+      [
+        {name: 'gondola', regex: /gondola/i, description: 'Gondola between Shard and Leth Deriel'},
+        {name: 'mode', options: %w[north south], description: 'Which direction are you going?'}
       ]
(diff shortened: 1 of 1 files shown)


jandersson:
This will take care of #1648

Issue #1188, opened 2016-12-22
Bescort Ways wont start from Dor'na'tornal


mwest152:
```
> l
[The Spire, Chamber of Contemplation]
Four sheep-skin rugs are laid across the cold floor at the cardinal points of the room, providing places to sit with minimal discomfort.  Pale white streaks mar the otherwise pristine silvery-blue crystalline walls, clustered predominately on the northern wall.
You also see a white-streaked crystalline archway and the silvery-white shard Dor'na'torna.
Obvious exits: none.
> ,bescort ways shard

--- Lich: bescort active.

[bescort]>recall heavens grazhir

You know the names of:
  Asharshpar'i
  Dinegavren
  Mintais
  Rolagi
  Dor'na'torna
  Marendin
  Tamigen
  Taniendar
  Erekinzil
  Tabelrem
  Auilusi
  Vellano
Roundtime: 2 sec.
R> 
[bescort: You are not in a room with a Grazhir shard.]

--- Lich: bescort has exited.
```

Issue #1146, opened 2016-12-12
Bescort Ways: Store Known Shards in UserVars


jandersson:
To recap: when you do ;bescort ways X, we check if X is known. Which takes some RT

Once you know a shard... you know it. Could save some time to store the known shards as a list in UserVars. If the shard isn't in the list, recall shards, store the results, and show the error message if requested shard is still unknown. Its a minor change and makes bescort seem a little smoother I think.

Issue #1144, opened 2016-12-12
Bescort Ways: Handle Pelag ai Aldam meeting


jandersson:
Bescort should handle this meeting:

> Two beams of light collide and merge in a shower of incandescent sparks!    
> 
> Ominous mental static builds in the area.  Incomprehensible thoughts infringe on your mind, making clear thinking an increasingly difficult feat. 
> 
> Unlike most astral phenomena, the static does not go away after a few moments.  In fact, it's getting stronger.  Closer.
> 
> A skinny Human, pale-skinned and with red hair, appears in the distance.  At the sight of him, the alien thoughts invading your mind crescendo and turn into a thread of distorted Common.  The entity's mouth remains closed in a scowl, but you can hear him all the same. "Your arrogance is without limit.  Here you are, outside even your own plane, spreading the poison of your names; your ambition; your pathetic, mortal failings!"  The stream of thoughts condenses in a brain-searing point of rage before becoming comprehendible again.  "Would that I could make you suffer even a fraction of the violence you inflict on us."  
> The distortion in the entity's "voice" grows, yet you sense a tone of ritual within it.  "I was named Pelag ai Aldam.  The Defilers of Fate are mine to claim and I demand sacrifice!"  The man and the psychic barrage vanish abruptly.                          
> 
> Your vision blurs and dims.  Your body goes numb and you are only distantly aware that you have collapsed.  There is a distinct lack of pain, or any other sensation for that matter, as your heart goes quiet.  
>  * Crannach was just struck down!                 

The first three messages showed up when i was in different rooms, so its apparently a slowly building thing. It could be possible to just have bescort boot the user from the game if it gets the second message.
I mention the second message as a trigger because I'm not sure if the first one is related, it just looked unusual to me.

jandersson:
I should note that there was a time lag between me dying and seeing the old man. After I had the vision, I casted moongate at the exit shard, but promptly died after i arrived back in real world.

PR #1652, opened 2017-05-20
Bescort Ways/Astral Travel Updates

jandersson:
Addresses #1144 #1146 #1188 
* Store known shards in uservars instead of rechecking every time
* Add Hib shard room id
* Add a danger flag which checks for dangerous events in the astral plane
* If the initial gate cast fails due to spell failure, give an error message, release harnessed mana and exit

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -668,9 +668,12 @@ class Bescort
     end
   end
 
-  def know_shard?(shard_name)
-    result = bput('recall heavens grazhir', shard_name, 'Roundtime')
-    result != 'Roundtime'
+  def know_shard?(shard_name, shard_list)
+    return true if UserVars.known_shards.include?(shard_name)
+    bput('recall heavens grazhir', shard_name, 'Roundtime')
+    lines = reget(30).map { |line| line.strip }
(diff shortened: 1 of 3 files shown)


jandersson:
Also removed an unused setting `quit_on_death_in_ap`. I think its more sensible to default that behaviour (like combat-trainer does), and if someone wants to turn it off they can request it.

jandersson:
I'll update the documentation wiki with the removal of the setting if merged.

bescort.lic
```diff
@@ -680,7 +683,15 @@ class Bescort
...
       waitcastrt?
-      DRStats.circle > 99 ? cast?('cast grazhir') : cast?("cast #{shard}")
+      if DRStats.circle > 99
+        unless cast?('cast grazhir')
+          echo "You may need to move to another room"
+          fput('release mana')
+          exit
+        end
+      else
+        cast?("cast #{shard}")
```
jandersson: No return checked here because this is casting in known rooms (where shards are).

rpherbig:
Looks good to me. @rcuhljr ? @kamarcum ?

rcuhljr:
👍

PR #1662, opened 2017-05-24
Added Slip Stalk to combat_trainer & common

intangibles:
You can now supply a YAML setting of "hide_type" to change what command is performed.  For example "hide_type: stalk" will perform slip stalk.

Also updated hide? method to use optional paramaters. Since hide? is in common, i wanted to retain its functionality for other scripts that may use it.

I tested the exp unofficially and it seemed to be better, but at the very least it has 25% reduced RT. Hide is 2 sec, stalk is 2 sec. Slip stalk is 3 sec total.

Yaml Settings:
hide_type: stalk     - will perform slip stalk instead of hide
force_unhide:  true  (default false) - will always unhide after hiding via Stealth timer.  (this will not affect attacks)

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -268,13 +268,15 @@ module DRC
     end
   end
 
-  def hide?(*hide_type)
-    hide_type.any? ? hide_type = hide_type[0] : hide_type = 'hide' 
+  def hide?(hide_type = 'hide')
     unless hiding?
-      case bput(hide_type, 'Roundtime', 'too busy performing', 'can\'t see any place to hide yourself', 'Stalk what')
+      case bput(hide_type, 'Roundtime', 'too busy performing', 'can\'t see any place to hide yourself', 'Stalk what', 'You\'re already stalking')
       when 'too busy performing'
         bput('stop play', 'You stop playing', 'In the name of')
-        return hide?
+        return hide?(hide_type)
+      when "You're already stalking"
+        put 'stop stalk'
+        return hide?(hide_type)
       end
       pause
       waitrt?
--- a/common.lic
+++ b/common.lic
@@ -20,7 +20,7 @@ $ENC_MAP = {
   'It\'s amazing you aren\'t squashed!' => 11
 }
 
-custom_require.call(%w(spellmonitor drinfomon))
+custom_require.call(%w[spellmonitor drinfomon])
 
 module DRC
   module_function
@@ -174,7 +174,7 @@ module DRC
     list_to_array(list)
       .map { |long_name| get_noun(long_name) }
       .compact
-      .select { |noun| noun != '' }
+      .reject { |noun| noun == '' }
   end
 
   def get_noun(long_name)
@@ -224,17 +224,17 @@ module DRC
     end
   end
 
-  def listen?(teacher,observe_flag=false)
+  def listen?(teacher, observe_flag = false)
     return false if teacher.nil?
     return false if teacher.empty?
 
-    bad_classes = %w(Thievery Sorcery)
+    bad_classes = %w[Thievery Sorcery]
     bad_classes += ['Life Magic', 'Holy Magic', 'Lunar Magic', 'Elemental Magic', 'Arcane Magic', 'Targeted Magic', 'Arcana', 'Attunement'] if DRStats.barbarian? || DRStats.thief?
     bad_classes += ['Warding'] if DRStats.thief?
     bad_classes += ['Utility'] if DRStats.barbarian?
 
-    observe = observe_flag ? "observe" : ""
-    
+    observe = observe_flag ? 'observe' : ''
+
     case bput("listen to #{teacher} #{observe}", 'begin to listen to \w+ teach the .* skill', 'already listening', 'could not find who', 'You have no idea', 'isn\'t teaching a class', 'don\'t have the appropriate training', 'Your teacher appears to have left', 'isn\'t teaching you anymore', 'experience differs too much from your own', 'but you don\'t see any harm in listening', 'invitation if you wish to join this class')
     when /begin to listen to \w+ teach the (.*) skill/
       return true if bad_classes.grep(/#{Regexp.last_match(1)}/i).empty?
--- a/common.lic
+++ b/common.lic
@@ -20,7 +20,7 @@ $ENC_MAP = {
   'It\'s amazing you aren\'t squashed!' => 11
 }
 
-custom_require.call(%w[spellmonitor drinfomon])
+custom_require.call(%w(spellmonitor drinfomon))
 
 module DRC
   module_function
@@ -174,7 +174,7 @@ module DRC
     list_to_array(list)
       .map { |long_name| get_noun(long_name) }
       .compact
-      .reject { |noun| noun == '' }
+      .select { |noun| noun != '' }
   end
 
   def get_noun(long_name)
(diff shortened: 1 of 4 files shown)


jandersson:
What does slip stalk do?

jandersson:
Nevermind, figured it out.
```
R>slip help                                                                                              
You begin to target a mountain giant.                                                                    
>                                                                                                        
1)  SLIP <right|left> or SLIP <item> in <container>       [Silent stowing of an item into your packs]    
2)  SLIP <item>                                           [Silent getting of an item from your packs]    
 - Addition 1:    Works on WORN items.                                                                   
 - Addition 2:    Works with items ON THE GROUND.                                                        
3)  STALK <person|creature> (while unhidden)              [Hide + Stalk of target]                       
4)  SNEAK <direction>    (while unhidden)                 [Hide + Sneak in direction]                    
```
```
>stalk                                                                            
You turn to face a mountain giant.                                                
You quickly slip into hiding to prepare to stalk.                                 
You melt into the background, convinced that your attempt to hide went unobserved.
You are convinced that your stalking went unobserved.                             
You move into position to stalk a mountain giant when it moves.                   
Roundtime: 3 sec.                                                                 
```

rcuhljr:
If the only point to hiding is training stealth I think you get more exp from hide+stalk then you do from using slip. If that's not the case it would be good to know.

common.lic
```diff
@@ -268,9 +268,10 @@ module DRC
...
 
-  def hide?
+  def hide?(*hide_type)
+    hide_type.any? ? hide_type = hide_type[0] : hide_type = 'hide' 
     unless hiding?
-      case bput('hide', 'Roundtime', 'too busy performing', 'can\'t see any place to hide yourself')
+      case bput(hide_type, 'Roundtime', 'too busy performing', 'can\'t see any place to hide yourself', 'Stalk what')
       when 'too busy performing'
         bput('stop play', 'You stop playing', 'In the name of')
         return hide?
```
rcuhljr: This needs to pass hide_type through again.

common.lic
```diff
@@ -268,9 +268,10 @@ module DRC
     end
   end
 
-  def hide?
+  def hide?(*hide_type)
+    hide_type.any? ? hide_type = hide_type[0] : hide_type = 'hide'
```
rcuhljr: We can replace both of these lines with 
```
def hide?(hide_type = 'hide')
```

and add hide_type to `base.yaml` and `base-empty.yaml` with the default value.

intangibles:
I tested the exp unofficially and it seemed to be better, but its also 25% reduced RT.  Hide is 2 sec, stalk is 2 sec.  Slip stalk is 3 sec total.

However, if your stealth timer is too short you will get a "You are already stalking message." I was thinking about adding "stop stalk" but if your stealth is around 15 seconds then stalk automatically drops.

> [combat-trainer]>stalk
> 
> You quickly slip into hiding to prepare to stalk.
> You melt into the background, convinced that your attempt to hide went unobserved.
> You are convinced that your stalking went unobserved.
> You move into position to stalk a warklin mauler when it moves.
> Roundtime: 3 sec.
> 
> HR> 
> You are revealed from concealment.
> The greenish radiation around you burbles warningly before a sphere of hissing acid spews forth in all directions!
> The wave lightly pits a warklin mauler's abdomen as it blasts across it.
> The wave lightly pits an armored warklin's abdomen as it blasts across it.
> The wave lightly pits an armored warklin's neck as it blasts across it.
> The wave lightly pits an armored warklin's chest as it blasts across it.
> The wave lightly pits an armored warklin's chest as it blasts across it.
> 
> R> 
> An armored warklin carefully picks its way west.
> 
> > 
> [combat-trainer]>target cac 20
> 
> You trace a careful sigil in the air, shaping the pattern of the Carrion Call spell.
> You begin to weave mana lines into a target pattern around a warklin mauler.
> 
> > 
> [combat-trainer]>app mauler quick
> 
> You are certain that the warklin mauler is battered.
> The warklin mauler has a relatively minor wound.
> You are certain that it is about as strong as you are.
> You are certain that it is about as agile as you are.
> You wonder if it is far less disciplined than you are.
> You are certain that it is rather less quick to react than you are.
> You are confident that it is not quite as conditioned as you are.
> Taking stock of its offensive abilities, and defending with a diamond-hide shield edged with a tyrium band and an etched steel parry stick with black leather straps, you are certain that the warklin mauler is definitely less skilled.
> Taking stock of its defensive abilities, and attacking with your fists, you are certain that the warklin mauler is a worthy opponent.
> 
> [Roundtime: 3 seconds]
> R> 
> [combat-trainer]>stalk
> 
> You're already stalking a warklin mauler!

common.lic
intangibles: Great. i was trying to find a better way to do this but had no idea how!

common.lic
intangibles: Ahh.  Great, I'm going to add another piece too for "You are already stalking."

```
when 'You're already stalking'
puts 'stop stalking'  #This shouldn't have any RT associated so waiting for a bput would slow the process down
return hide?(hide_type)
```

common.lic
intangibles: Wasn't sure what to put here since these all seem to be lists

base-empty.yaml

  gear_sets: {}
  lichbot_buffs: []
  listen_skills: []
  offensive_spell_cycle: []
  ignored_npcs: []
  hunting_buddies: []
  empty_hunting_room_messages: []
  classes_to_teach: []
  hide_type:

and for base.yaml
hide_type:
or 
hide_type: hide

rcuhljr: `hide_type: hide` for both of them. The goal is just to safe guard them if they decide to overwrite their `hide_type: stalk` with `hide_type:` in a later file.

rcuhljr: Although this is drifting farther from the purpose of base-empty perhaps, maybe we should just let them error out and fix it. @rpherbig opinions?

intangibles:
Added @force_unhide setting

Also added in a quick combat fix so that if dance_action: is blank, it stops forcing line feeds.
fput(game_state.next_dance_action) if !game_state.next_dance_action.nil?

intangibles:
to do: add stop stalk to end of combat clean up

intangibles:
@rcuhljr 
@rpherbig 

I've been using this for quite awhile and everything seems to be working smoothly.  It's probably ready for a merge.

Also, I'm not really sure what the conflict is that popped up. with base-empty.yaml.

profiles/base-empty.yaml
```diff
@@ -45,4 +45,5 @@ empty_values:
   hunting_buddies: []
   empty_hunting_room_messages: []
   classes_to_teach: []
-  whirlwind_trainables: []
\ No newline at end of file
+  whirlwind_trainables: []
+  hide_type: hide
```
rpherbig: This should not be added here

rpherbig: That's what caused the merge conflict

intangibles: Hmm...

@rcuhljr suggested i add hide_type: hide  to the base-empty.   But he wasn't sure if maybe it was a better idea to just let it error out.

Or are you talking about the newline errors that I keep getting? Quite annoying!

intangibles: I found some sublime settings to fix those trailing white spaces and endlines! hopefully gone forever.

common.lic
rpherbig: Sorry, I apparently missed this conversation. We resolved it below.

intangibles:
Updated wiki.

hide_type: Value of "stalk" changes hide type to "slip stalk". This will perform the command "stalk" instead of "hide" which activates "slip stalk". "Slip stalk" is only available for a number of professions and must be learned. Default value: hide. Uses hide as normal.

force_unhide: Value of true will always unhide after hiding via Stealth timer under training_abilities:.

dont_stalk: Value of true prevents stalk usage after hiding when hiding via Stealth timer under training_abilities:. Default value is false.

PR #1673, opened 2017-05-29
Update links with What is xyz?

jayRyan24:
Adding in links to Lich and Dependency explanations on Elanthipedia.


links.lic
```diff
@@ -7,7 +7,9 @@ class Links
 
   def initialize
     links = {
+      'What is Lich?' => 'https://elanthipedia.play.net/Lich_scripting_engine'
```
rpherbig: Do you need commas at the end of these two new lines?

jayRyan24: Done... woops.

PR #1678, opened 2017-05-29
Sorcery always walk to sorcery room

jayRyan24:
I think it's important to walk to sorcery room if it exists. If the person doesn't want to do it for runestone, they just need to leave it off, rather than not attempt the walk-to at all.


jayRyan24:
This is a good change we should put in ASAP. :)

PR #1679, opened 2017-05-30
Refactor Appraisal

jandersson:
This refactors appraisal to use a single list to specify a training routine:

```yaml
appraisal_training:
- zills
- art
- pouches
- gear
```

The default is 
```yaml
appraisal_training:
- zills
- art
```
And a warning is given if appraisal is greater than 250 and `zills` is in the training list, telling the user to remove it from appraisal training

<s>This affects the appraisal training in `appraisal` and `crossing-trainer` .

Currently there is no way to turn off art-gallery appraisal.

I found myself running up from Shard to the Crossing to use the art gallery. I prefer to just train appraisal using gem pouches. `train_appraisal_with_pouches?` predicate returns false if mindstate is not above 30, and kicks off the `study-art` if this predicate did not return `true`. I added a setting to let me disable this behaviour called `train_appraisal_with_art` and defaulted it to `true`.

The default setting will have the script behave as it was before, but let the user disable running to the art-gallery if they just want to appraise pouches, zills and/or gear.</s>


rpherbig:
I think I'd rather have one setting that says how you want to train appraisal. We don't need to fall through from one type to another if you aren't mindlocked.

That would be more similar to how other scripts train.

jandersson:
OK I like that idea

```
appraisal_training:
- pouches
- art
- gear
```
Something like that?

appraisal.lic
```diff
@@ -10,25 +10,35 @@ class Appraisal
 
   def initialize
     @equipment_manager = EquipmentManager.new
-    @equipment_manager.wear_equipment_set?('standard')
     @settings = get_settings
+    train_list = get_settings.appraisal_training
```
jandersson: This needs to added to `base` with defaults 
```
appraisal_training:
- zills
- art
```
that should capture the current default behaviour

jandersson:
Looks to be working. One last consideration - how to handle the old settings? `appraise_with_pouches` etc..? Should do something in `dependency` to merge a users legacy settings into the new one?

appraisal.lic
rpherbig: This should be `train_list = @settings.appraisal_training`. Although, the only place `@settings` is even used is `train_appraisal_with_pouches`. You could make `full_pouch_container` an argument to that method and then `@settings` doesn't have to be an instance var.

appraisal.lic
```diff
@@ -54,18 +63,16 @@ class Appraisal
     DRSkill.getxp('Appraisal') < 30
   end
 
-  def assess_zills?
-    return false if DRSkill.getrank('Appraisal') >= 250
+  def assess_zills
+    return if DRSkill.getrank('Appraisal') >= 250
```
rpherbig: Maybe it should echo to inform the user that they can remove this from their config, and/or need to find a new way to train Appraisal?

appraisal.lic
```diff
@@ -10,25 +10,35 @@ class Appraisal
...
   def initialize
     @equipment_manager = EquipmentManager.new
-    @equipment_manager.wear_equipment_set?('standard')
     @settings = get_settings
+    train_list = get_settings.appraisal_training
+    
+    train_list.each do |task| 
+      break if DRSkill.getxp('Appraisal') >= 30
+      case task
+      when /zill/
```
rpherbig: I wouldn't bother with a regex or multiple options per train. I think we can document the valid options on epedia.

appraisal.lic
jandersson: Great catch, thank you

appraisal.lic
jandersson: I like that, will add

appraisal.lic
jandersson: I had the idea that the user might enter something like `gems` instead of `pouch` or `pouches`, but it seems half-baked and adds more complexity.. ill ditch the multiple options. I do like the regex though, since it lets the user write `pouch`, `pouches`, `zill`, `zills`. Is it something you feel strongly about? I don't feel strongly about it, but i do like that it covers the user unintentionally entering the plural vs singular or vice versa.

jandersson:
Made the suggested changes, also cleaned up an accident

dependency.lic
```diff
@@ -953,15 +962,12 @@ class MoonWatch
     request = Net::HTTP::Get.new(uri.request_uri)
     request['Content-Type'] = 'application/json'
     response = http.request(request)
-    if response.body.nil?
```
rpherbig: Is this a merge error?

jandersson: yes! thank you

jandersson: Wait.. this may have been something rubocop corrected. Let me look before i leap..

rpherbig: I just pushed a Rubocop commit, so this may go away if you merge in master.

jandersson: OK, looks like some experiment by me. will remove

jandersson: Nope nevermind, its gone. You were right

jandersson:
Looks like this is all set. Let me know of any concerns

rpherbig:
Looks good to me. @rcuhljr ?

rcuhljr:
I guess my question is, why zills+ art as the default? If I have zills I only want to use those, if I don't I just want art. Art might be a safer no zills option, but I can't think of when I'd want to assess zills once then do art.

jandersson:
Yes that is the current default behaviour

jandersson:
I suppose a more sensible default is just art, because the user might not have zills.

jandersson:
Updated the training default to 
```
appraisal_training
- art
```

dependency.lic
```diff
@@ -894,6 +894,15 @@ class SetupFiles
     s.bankbot_name ||= UserVars.bankbot_name
     s.bankbot_room_id ||= UserVars.bankbot_room_id
 
+    # Merge legacy appraisal settings into target appraisal settings
+    unless s.appraisal_training.include?('pouches')
+      s.appraisal_training << 'pouches' if s.train_appraisal_with_pouches
```
rcuhljr: Could we add loud obnoxious echoes here to get people on the new format?

rcuhljr: Ok I just noticed that was part of validate. Not sure if we want it both places.

rcuhljr:
I'm fine with this.

dependency.lic
jandersson: Of course

jandersson: I think it might be missed/confusing if its in dependency start up. I can make the validate warning more obnoxious though

jandersson: Done

PR #1690, opened 2017-06-06
Added Assassin's crossbow to combat trainer.

FluxBorn:
I added the assassin's crossbow to the combat-trainer, as previously it would fail to aim as it is a repeating crossbow and needs to be pushed to lock the bolt in place.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1702,18 +1702,11 @@ class AttackProcess
 
   def aim(game_state)
     case bput('aim', 'You begin to target', 'You are already', 'There is nothing else', 'Face what\?', 'You shift your target', 'Your repeating crossbow', 'Your assassin\'s crossbow','Strangely, you don\'t feel like fighting right now')
-    when 'Your assassin\'s crossbow'
-      case bput('push my cross', 'A rapid series of clicks', 'You attempt to ready your assassin\'s crossbow')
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -1701,7 +1701,14 @@ class AttackProcess
...
-    case bput('aim', 'You begin to target', 'You are already', 'There is nothing else', 'Face what\?', 'You shift your target', 'Your repeating crossbow', 'Strangely, you don\'t feel like fighting right now')
+    case bput('aim', 'You begin to target', 'You are already', 'There is nothing else', 'Face what\?', 'You shift your target', 'Your repeating crossbow', 'Your assassin\'s crossbow','Strangely, you don\'t feel like fighting right now')
+    when 'Your assassin\'s crossbow'
+      case bput('push my cross', 'A rapid series of clicks', 'You attempt to ready your assassin\'s crossbow')
+      when 'A rapid series of clicks'
+        aim(game_state)
+      when 'You attempt to ready your assassin\'s crossbow'
+        game_state.loaded = false
+      end
     when 'Your repeating crossbow'
```
rpherbig: I think it would be easier to reuse the repeating crossbow code. You can do something like: `when 'Your repeating crossbow', 'Your assassin\'s crossbow'`

PR #1700, opened 2017-06-12
Bankbot timestamp fix

NBSL:
This fixes bankbot failing to register commands/validations if you have lnet timestamps=on


rpherbig:
Thanks!

PR #1702, opened 2017-06-12
Swap loot_nouns and trash_nouns to give lootables priority over base-…

NBSL:
This swaps the if statement to check for loot_nouns before trash_nouns.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -260,10 +260,7 @@ class LockPicker
       bput("put #{item} in my #{special['bag']}", 'you put')
       return
     end
-
-    if @trash_nouns.find { |thing| item_long =~ /\b#{thing}\b/i }
-      dispose_trash(item)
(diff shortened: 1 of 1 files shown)


rcuhljr:
No concerns with this on my end, @rpherbig ?

rpherbig:
The change itself looks good to me. :+1:

I'm not sure why we check for loot and trash in two different ways, but that's not relevant to this PR.

rcuhljr:
So now people are all looting sunstone runestones because it's both a gem and a runestone.

NBSL:
I can add loot_subtractions as the first condition. Would that be sufficient enough for this issue? Cause the original issue were people in lnet wanted to keep nuggets but they had no option to stop pick from trashing them.

rcuhljr:
I pushed a special case for now with the sunstone runestones since it's the only offender while we find a better long term solution.

PR #1703, opened 2017-06-12
Sanowret Crystal script

NBSL:
Script that can be ran in the background to gaze or exhale your crystal. Checks skill state, area restrictions, hiding and concentration levels.

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -1,0 +1,39 @@
+=begin
+	Runs in the background, gazing or exhaling your sanowret crystal. Checks for concentration and current mindstate.
+
+	author: NBSL
+	game: dr
+  version: 0.1 (2017-06-12)
+
+  changelog:
+    0.1 (2017-06-12)
+     * Initial Release
+=end
+
+no_kill_all
+no_pause_all
+
+custom_require.call(%w[common drinfomon])
+invalid_room = nil
+
+while true do
+  line = script.gets?
+  response = ""
+  if DRStats.concentration == 100 && !hidden? && invalid_room != Room.current.id
+    if DRSkill.getxp('Arcana') <= 10
+      reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
+    elsif DRSkill.getxp('Arcana') <= 25
+      response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
+    end
+
+    case response
+    when /(?:A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.|you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.)/
+      invalid_room = nil
+      pause 6
+    when "This is not a good place for that."
+      invalid_room = Room.current.id
+      pause 1
+    end
+  end
+  pause 6 unless line
+end
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -48,7 +48,7 @@ class SanowretCrystal
         fput("get my sanowret crystal")
         use_crystal
         fput("stow my sanowret crystal")
-      when /(^You tap a sanowret crystal that you are wearing.$|)/
+      when /^You tap a sanowret crystal that you are wearing.$/
         use_crystal
       when /^You tap a sanowret crystal that you are holding\.$/
         use_crystal
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -27,7 +27,7 @@ class SanowretCrystal
   end
 
   def use_crystal
-    return if @invalid_rooms.include? Room.current.id || hidden?
+    return if @invalid_rooms.include?(Room.current.id) || hidden?
     if DRSkill.getxp('Arcana') <= 10
       reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
     elsif DRSkill.getxp('Arcana') <= 25
@@ -36,7 +36,7 @@ class SanowretCrystal
 
     case response
     when "This is not a good place for that."
-      @invalid_rooms.push(Room.current.id) if @invalid_rooms.include? Room.current.id
+      @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
       #move XMLData.room_exits.sample
     end
   end
@@ -61,7 +61,7 @@ class SanowretCrystal
 
   def passive
     loop do
-      next if @invalid_rooms.include? Room.current.id || hidden?
+      next if @invalid_rooms.include?(Room.current.id) || hidden?
       check_crystal
       pause 1
     end
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -35,8 +35,7 @@ class SanowretCrystal
       response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
     end
 
-    case response
-    when "This is not a good place for that."
+    if response =~ "This is not a good place for that."
       @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
       #move XMLData.room_exits.sample
     end
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -33,11 +33,8 @@ class SanowretCrystal
     elsif DRSkill.getxp('Arcana') <= 25
       response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
     end
-
-    if response =~ /This is not a good place for that\./
-      @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
-      #move XMLData.room_exits.sample
-    end
+    return if response !~ /This is not a good place for that\./
+    @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
   end
 
   def check_crystal
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -20,14 +20,14 @@ class SanowretCrystal
         { name: 'run', regex: /run/i, description: 'Single run of the script', optional: true }
       ]
     ]
-    @invalid_room = []
+    @invalid_rooms = []
     args = parse_args(arg_definitions)
     check_crystal if args.run && !hidden?
     passive unless args.run
   end
 
   def use_crystal
-    return if @invalid_room.include? Room.current.id || hidden?
+    return if @invalid_rooms.include? Room.current.id || hidden?
     if DRSkill.getxp('Arcana') <= 10
       reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
     elsif DRSkill.getxp('Arcana') <= 25
@@ -36,7 +36,7 @@ class SanowretCrystal
 
     case response
     when "This is not a good place for that."
-      @invalid_room.push(Room.current.id) if @invalid_room.include? Room.current.id
+      @invalid_rooms.push(Room.current.id) if @invalid_rooms.include? Room.current.id
       #move XMLData.room_exits.sample
     end
   end
@@ -61,7 +61,7 @@ class SanowretCrystal
 
   def passive
     loop do
-      next if @invalid_room.include? Room.current.id || hidden?
+      next if @invalid_rooms.include? Room.current.id || hidden?
       check_crystal
       pause 1
     end
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -33,22 +33,25 @@ class SanowretCrystal
     elsif DRSkill.getxp('Arcana') <= 25
       response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
     end
-    return if response !~ /This is not a good place for that\./
-    @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
+
+    if response =~ /This is not a good place for that\./
+      @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
+      #move XMLData.room_exits.sample
+    end
   end
 
   def check_crystal
     return if DRStats.concentration < 100 || DRSkill.getxp('Arcana') >= 25 || hiding?
     case DRC.bput('tap my sanowret crystal', /^You tap a sanowret crystal inside your .*.$/,/^You tap a sanowret crystal that you are wearing.$/, /^You tap a sanowret crystal that you are holding.$/, "I could not find what you were referring to.")
     when /You tap a sanowret crystal inside your/
-      fput("get my sanowret crystal")
+      DRC.bput('get my sanowret crystal', 'You get', 'What were you referring to')
       use_crystal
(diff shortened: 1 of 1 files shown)


NBSL:
This should be my last PR for today. Considering more people have crystals, it might be nice to have a script in the general package.

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+  changelog:
+    0.1 (2017-06-12)
+     * Initial Release
+=end
+
+no_kill_all
+no_pause_all
+
+custom_require.call(%w[common drinfomon])
+invalid_room = nil
```
rcuhljr: This line doesn't have any effect, you can snip it out.

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+
+while true do
```
rcuhljr: this can just be loop do

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+  line = script.gets?
+  response = ""
```
rcuhljr: We could add a line `next if invalid_room == Room.current.id` right here and simplify our later if statement.

rpherbig:
I was just about to push my own sanowret crystal script, so it's interesting to compare the two.

I'm concerned that some scripts may not handle well being paused for a long period of time (30-40s when you gaze I think). Have you been running this in the background while in combat or noncombat training?

Is this something that is advantageous to be running in the background? My initial thought was to make a standalone script (for those that want to use before/after) and then also wire it into crossing-training.

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+  if DRStats.concentration == 100 && !hidden? && invalid_room != Room.current.id
+    if DRSkill.getxp('Arcana') <= 10
+      reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
+    elsif DRSkill.getxp('Arcana') <= 25
+      response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
+    end
+
+    case response
```
rcuhljr: We could use an else statement to handle the first when here without really losing anything other than the giant match statement.
```
when "This is not...
  fail stuff
else
  success stuff
```

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+    when /(?:A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.|you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.)/
+      invalid_room = nil
```
rcuhljr: There's no reason to clear out the invalid_room id, a bad room number is still bad.

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
+=begin
+	Runs in the background, gazing or exhaling your sanowret crystal. Checks for concentration and current mindstate.
```
rpherbig: We've generally been putting a link to elanthipedia in the header block and documenting things there.

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+
+custom_require.call(%w[common drinfomon])
+invalid_room = nil
+
+while true do
+  line = script.gets?
+  response = ""
+  if DRStats.concentration == 100 && !hidden? && invalid_room != Room.current.id
+    if DRSkill.getxp('Arcana') <= 10
+      reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
```
rpherbig: What if your crystal is not worn?

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+    case response
+    when /(?:A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.|you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.)/
+      invalid_room = nil
+      pause 6
+    when "This is not a good place for that."
+      invalid_room = Room.current.id
+      pause 1
+    end
+  end
+  pause 6 unless line
```
rcuhljr: Why the six second pause here? I assume we're not waiting on a pulse.

rcuhljr:
Is there any concern about it gazing and then six seconds later gazing again or breathing?

sanowret-crystal.lic
```diff
@@ -0,0 +1,39 @@
...
+	game: dr
+  version: 0.1 (2017-06-12)
+
+  changelog:
+    0.1 (2017-06-12)
+     * Initial Release
+=end
+
+no_kill_all
+no_pause_all
```
rpherbig: I _think_ this means that if you are arrested, `jail-buddy` won't be able to pause this script. I don't see any error handling below for the case in which you've temporarily lost your crystal.

sanowret-crystal.lic
rcuhljr: There's not likely to be any support here for non worn crystals, the whole point of wearing them is they work during RT and basically no matter what you're doing, not wearing it ruins that.

NBSL:
I've been running it for a few days now, making changes when I run into issues. Only issues I've ran in to was it spamming when I was in dokts, and not being able to exhale while hidden. This has been both in combat and non combat, on both a thief and a trader. So far it handles everything just fine.

As for the not worn aspect, I actually thought about that after I pushed it to the repos. Planned on working on that today and testing it on my barb.

The 6 second wait is based off of an old memory where a tick in DR was 6 seconds. That also is there because I ran into performance issues using 0.5 and 1 second pauses.

The header stuff is because it's on ;repos. I, in truth, didn't think it'd be merged into the default repo but I hadn't seen anyone else post anything about it since that issue that was opened the other day.

And dear lord, you two respond fast.

rcuhljr:
Lunch timeeeeee. What kind of performance issues were seeing? we've got plenty of scripts running .1 or lower loops and this guy is doing almost no work.  It might be simpler to just get rid of line.gets? since it's never really used and just always pause one second each loop. If there was an option for this to run once and used stored crystals it would allow us to use the same script for crossing-training and people who like  before/after scripts for training manager.

NBSL:
It was freezing the display. Originally I had loop do and was using fputs since the crystals didn't care about most RTs. Earlier this morning I added the hidden check and changed to using bputs. I can move it to a single use if you want.

sanowret-crystal.lic
```diff
@@ -0,0 +1,73 @@
...
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'run', regex: /run/i, description: 'Single run of the script', optional: true }
+      ]
+    ]
+    @invalid_room = nil
+    args = parse_args(arg_definitions)
+    check_crystal if args.run
+    passive if args.run.nil?
```
rpherbig: You could also `passive unless args.run`

Though I wonder if this could be combined: always go into the loop, exit after the first iteration if the arg was used

sanowret-crystal.lic
```diff
@@ -0,0 +1,73 @@
...
+    return if @invalid_room == Room.current.id
+    if DRSkill.getxp('Arcana') <= 10
+      reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
+    elsif DRSkill.getxp('Arcana') <= 25
+      response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
+    end
+
+    case response
+    when "Doing that would give away your hiding place."
+      fput("unhide")
```
rpherbig: We `unhide` in a lot of places. Would you be up for adding a helper to `DRC` that does this (with a bput)? See https://github.com/rpherbig/dr-scripts/blob/1c8306610f09bbfa783d980f41b582c1f21f2b17/crossing-training.lic#L719 for an example

sanowret-crystal.lic
```diff
@@ -0,0 +1,73 @@
...
+      use_crystal
+    when "This is not a good place for that."
+      @invalid_room = Room.current.id
```
rpherbig: Why store only the last invalid room? Could this be a list instead?

sanowret-crystal.lic
NBSL: Is there any way to check if args exist at all? I have a cleanup script I wrote so that I could train alchemy and not end up with tons of trash and misplaced items, but couldn't figure out how to see if the OpenStruct was empty of args. Thats basically where that if was originally from.

sanowret-crystal.lic
rcuhljr: I'm not keen on unhiding in combat just use the crystal.

sanowret-crystal.lic
```diff
@@ -0,0 +1,73 @@
...
+      when /(^You tap a sanowret crystal that you are wearing.$|^You tap a sanowret crystal that you are holding\.$)/
+        use_crystal
+      else
+        echo "Sanowret crystal not found."
+      end
+    end
+  end
+
+  def passive
+    invalid_room = nil
```
rpherbig: Should this be `@invalid_room`? Does it need to be initialized at all?

sanowret-crystal.lic
NBSL: Sure.

sanowret-crystal.lic
```diff
@@ -0,0 +1,73 @@
...
+    @invalid_room = nil
+    args = parse_args(arg_definitions)
+    check_crystal if args.run
+    passive if args.run.nil?
+  end
+
+  def use_crystal
+    return if @invalid_room == Room.current.id
+    if DRSkill.getxp('Arcana') <= 10
+      reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
```
rpherbig: Will actions from other scripts interrupt the gaze? Do we need a pause here?

sanowret-crystal.lic
NBSL: Ah, the issues editing in 2 places lol. I have it removed in my test one, but forgot to remove it in my forked folder.

sanowret-crystal.lic
NBSL: The only issue I've seen is typeahead. I don't know of a way to handle that. 

Other than that, gaze works during RT. Exhale forces you to wait.

sanowret-crystal.lic
NBSL: If it's in passive mode, I could have it push to a list. The only time I ran into that error so far, was in with Dokt.

sanowret-crystal.lic
NBSL: @rcuhljr Would you be ok with something like this?
[Unhide?](https://pastebin.com/j0yUbzAt)

Not sure if I need the /.*/ at the end, currently testing it.

rpherbig: Why `assess`?

rcuhljr: No I'd just rather not unhide at all honestly. People hide while doing climbs, if we break that for this and then a mob wanders in and aggros them, yuck. Having to wait while hidden isn't a serious problem, you're not going to clear fully while hidden and if you do so what. Or we get someone killed/afk'd because they hid to go afk and this bumps them out.

NBSL: Outside of introducing events as a require for common, the only other thing I see is is DRRoom.npcs which would return both hostile and non-hostile NPCs. I didn't see anything in lich that would give me current combat status.

NBSL: The other solution could be Script.running?("combat-trainer") But that would stop the passive from running when you're in a crap spawn room like my trader was.

jandersson: Would it make more sense to look at hiding status and not gaze at the crystal if hidden?

NBSL: That's what it was originally @jandersson . This is more in response to adding an unhide check to common.lic

NBSL:
Ok. I added hidden checks to the spots where I think it'd run into race conditions in passive mode. I changed invalid_room to an array so that while running in passive mode, it'll automatically skip learned rooms.

sanowret-crystal.lic
```diff
@@ -0,0 +1,72 @@
...
+class SanowretCrystal
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'run', regex: /run/i, description: 'Single run of the script', optional: true }
+      ]
+    ]
+    @invalid_rooms = []
+    args = parse_args(arg_definitions)
+    check_crystal if args.run && !hiding?
```
rpherbig: You don't need to check `hiding?` here, `check_crystal` already does it

sanowret-crystal.lic
```diff
@@ -0,0 +1,72 @@
...
+
+    if response =~ "This is not a good place for that."
+      @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
+      #move XMLData.room_exits.sample
+    end
+  end
+
+  def check_crystal
+    return if hiding?
+    if DRStats.concentration == 100 && DRSkill.getxp('Arcana') <= 25
```
rpherbig: I think this function will be a bit clearer if you make this a guard clause (`return if DRStats.concentration != 100 || DRSkill.getxp('Arcana') > 25`)

sanowret-crystal.lic
```diff
@@ -0,0 +1,72 @@
...
+      case DRC.bput('tap my sanowret crystal', /^You tap a sanowret crystal inside your .*.$/,/^You tap a sanowret crystal that you are wearing.$/, /^You tap a sanowret crystal that you are holding.$/, "I could not find what you were referring to.")
+      when /You tap a sanowret crystal inside your/
+        fput("get my sanowret crystal")
+        use_crystal
+        fput("stow my sanowret crystal")
+      when /(^You tap a sanowret crystal that you are wearing.$|)/
```
rpherbig: You probably want to escape the `.` here.

Why is this captured (i.e. use of `()`)?

How does the or (`|`) affect this? Won't that only match an empty result?

sanowret-crystal.lic
```diff
@@ -0,0 +1,72 @@
...
+  def use_crystal
+    return if @invalid_rooms.include?(Room.current.id)
+    return if hiding?
+    if DRSkill.getxp('Arcana') <= 10
+      reponse = DRC.bput("gaze sanowret crystal", "A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of \w+.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
+    elsif DRSkill.getxp('Arcana') <= 25
+      response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
+    end
+
+    if response =~ "This is not a good place for that."
```
rpherbig: This could also use the guard clause style (`return if response !~ "This is not a good place for that."`)

sanowret-crystal.lic
NBSL: I removed that. It was left over from checking for 2 errors. As of right now I haven't pushed my latest changes, since I ran into issues with everything hanging when I'd try to ambush stun.

sanowret-crystal.lic
```diff
@@ -0,0 +1,66 @@
...
+      reponse = DRC.bput("gaze sanowret crystal", /A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of .*\./, "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.")
+    elsif DRSkill.getxp('Arcana') <= 25
+      response = DRC.bput("exhale sanowret crystal", "you come away from the experience with a further understanding of Arcana as the scintillating lights fade again.", "However, nothing much else happens, as you lack the concentration to focus.", "This is not a good place for that.", "Doing that would give away your hiding place.")
+    end
+    return if response !~ /This is not a good place for that\./
+    @invalid_rooms.push(Room.current.id) unless @invalid_rooms.include?(Room.current.id)
+  end
+
+  def check_crystal
+    return if DRStats.concentration < 100 || DRSkill.getxp('Arcana') >= 25 || hiding?
```
rpherbig: I prefer each guard clause to have its own `return`:
```
return if DRStats.concentration < 100
return if DRSkill.getxp('Arcana') >= 25
return if hiding?
```

rpherbig: But you don't have to change it, just noting a style thing

sanowret-crystal.lic
```diff
@@ -0,0 +1,66 @@
...
+    case DRC.bput('tap my sanowret crystal', /^You tap a sanowret crystal inside your .*.$/,/^You tap a sanowret crystal that you are wearing.$/, /^You tap a sanowret crystal that you are holding.$/, "I could not find what you were referring to.")
+    when /You tap a sanowret crystal inside your/
+      fput("get my sanowret crystal")
```
rpherbig: Can you use `bput` (with matcher messages) instead of these `fput`s?

PR #1704, opened 2017-06-12
Have the wealth beautifier ignore shop links

NBSL:
Fixes breaking shop links. Only tested on my personal build of Frostbite, but it should work on all frontends that supports links since it's coming from lich

PS I was wrong about my PRs.


rpherbig:
What 'shop links' are you referring to?

NBSL:
[XML Example](https://pastebin.com/Paviy9K4)

These

rcuhljr:
the SHOP command in shops.

PR #1708, opened 2017-06-13
Workorders.lic update

xtremenerdrage:
Adds sewing and changes sewing to knitting

Also updates 
    sew-stock-number: 8
    knit-stock-number: 13
in base-crafting

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -84,7 +84,12 @@ class WorkOrders
       if repair
         repair_items(info, tools)
       else
-        sew_items(info, item, quantity)
+	recipe = @recipes.find { |r| r['name'] == item }
+	if recipe['chapter'] == 4
+	  sew_items(info, item, quantity)
+	else	  
+          sew_items(info, item, quantity)
+	end
       end
     when 'shaping'
       @engineering_room = settings.engineering_room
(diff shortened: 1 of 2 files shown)


workorders.lic
```diff
@@ -84,7 +84,12 @@ class WorkOrders
       if repair
         repair_items(info, tools)
       else
-        sew_items(info, item, quantity)
+	recipe = @recipes.find { |r| r['name'] == item }
+	if recipe['chapter'] == 4
+	  sew_items(info, item, quantity)
```
rpherbig: What is the differences in the if and else blocks?

workorders.lic
```diff
@@ -294,6 +299,31 @@ class WorkOrders
     end
   end
 
+  def order_cloth(stock_room, stock_needed, stock_number)
```
rpherbig: This looks like it's the same as `order_yarn` except for the noun. Could that be a parameter passed into a common function?

workorders.lic
```diff
@@ -84,7 +84,12 @@ class WorkOrders
       if repair
         repair_items(info, tools)
       else
-        sew_items(info, item, quantity)
+	recipe = @recipes.find { |r| r['name'] == item }
```
rpherbig: You have to look up the recipe again in both the sew and knit cases. Can `recipe` be passed in, instead of `item`? Or is this an indicator that the lookup and the if should be inside some intermediate function which contains a lot of the shared code between knitting and sewing?

rcuhljr: I was planning that for a later refactor, work orders as a whole needs some clean up and standardization.

rpherbig: :+1:

rcuhljr:
Closing this since I've stolen the code and am pulling it into my own change. Thanks!

Issue #1710, opened 2017-06-14
Hunting Location Overflow


jayRyan24:
Add support for Overflow hunting in the base YAML. If your chosen zone is full, you should then try the overflow. A good example of this would be rock_trolls2 to rock_trolls. 

```
- :zone:
 - rock_trolls2
 - rock_trolls
```

or

```
- :zone: 
 - elder_prereni 
 - blue_prereni
 - young_prereni
```
 
At this point we just smash the roomnumbers together, appending the zones on to the previous.

rcuhljr:
Reminder for whenever someone jumps on this, update validate's tests against hunting zone names.

jayRyan24:
Done.

PR #1711, opened 2017-06-14
Hunting zone overflow

jayRyan24:
Allows the use of multiple hunting zones in the same hunting zone. This will enable someone to use an overflow if the first option is too full. Directly affects issue #1710!
```

- :zone:
  - rock_trolls2
  - rock_trolls2
  - beisswurms
```


hunting-buddy.lic
```diff
@@ -105,9 +105,11 @@ class HuntingBuddy
     wait_for_script_to_complete('buff', ['prehunt_buffs'])
   end
 
-  def find_hunting_room?(zone_name)
+ def find_hunting_room?(zone_name)
```
rcuhljr: WhyUBreakOurSpacing? :)

hunting-buddy.lic
```diff
@@ -105,9 +105,11 @@ class HuntingBuddy
...
     UserVars.friends = @settings.hunting_buddies || []
-    if rooms = @hunting_zones[zone_name]
+    rooms = []
+    rooms = Array(zone_name).each { |name| rooms.push(*@hunting_zones[name]).flatten; } 
+    if rooms
```
rcuhljr: This check is kind of meaningless now since you set rooms to an array, it should probably be `unless rooms.empty?`.

hunting-buddy.lic
```diff
@@ -107,7 +107,9 @@ class HuntingBuddy
 
   def find_hunting_room?(zone_name)
     UserVars.friends = @settings.hunting_buddies || []
-    if rooms = @hunting_zones[zone_name]
+    rooms = []
+    rooms = Array(zone_name).each { |name| rooms.push(*@hunting_zones[name]).flatten; }
```
kamarcum: I think it'll work as
`rooms = @hunting_zones.values_at(*Array(zone_name)).flatten`

rcuhljr:
Do we want to do anything about escort zones? or people who have escort and non escort zones mixed?

jayRyan24:
Won't this just go between them?

rcuhljr:
The current change won't work for escort zones, and if you have an overflow zone that's an escort zone it's going to ignore it.

jayRyan24:
I see it, my bad. I'll add support for this. Do we want mixed escort with non-escort? Seems like the best option.

jayRyan24:
How do you want to handle this. I don't think we should mix the non-escort and escorts. They work entirely different. I am looking at the idea that if the first zone is an escort zone, it will default to just that, if it isn't, it will basically skill the escort zones that get listed. 

This will hunt only the crocs. 
```
- :zone:
  - blue_crocs
  - zone2
  - zone3
```

This will combine zone2 and zone3 and skip crocs. 
```
- :zone:
  - zone2
  - blue_crocs
  - zone3
```

jayRyan24:
Can also pop out a warning message stating that it isn't supported.

jayRyan24:
Set it up to not allow mixing of escort with non-escorts. They are handled too differently in the system. So, if an escort is listed as the first zone, it will default to that only, if not, it will remove all other escorts and then condense all the hunting zones into a single array then proceed as would be expected.

rcuhljr:
Have you run this yet for awhile or should I test it locally?

jayRyan24:
I'll move it over and run it all night and let you know the results.

jayRyan24:
Been running this all day without an issue

jayRyan24:
Gonna test over tonight and into tomorrow

jayRyan24:
This is ready for merging. I've been running it all weekend and a couple days during the week. It's working as intended and works great.

Example outputs:

Condition Escort Listed First:
```
[hunting-buddy: find_hunting_room:: zones: ["blue_crocs", "thicket_vipers", "thicket_vipers"]]
[hunting-buddy: find_hunting_room:: rooms: ]
```

Condition Escort Mixed In:
```
[hunting-buddy: find_hunting_room:: zones: ["thicket_vipers", "blue_crocs", "thicket_vipers"]]
[hunting-buddy: find_hunting_room:: rooms: [6837, 6846, 6845, 6850, 6843, 6849, 6837, 6846, 6845, 6850, 6843, 6849]]
```

Condition No Escort, Multiple Normal Zones:
```
[hunting-buddy: find_hunting_room:: zones: ["thicket_vipers", "thicket_vipers"]]
[hunting-buddy: find_hunting_room:: rooms: [6837, 6846, 6845, 6850, 6843, 6849, 6837, 6846, 6845, 6850, 6843, 6849]]
```

PR #1716, opened 2017-06-16
ritual focus ties

xtremenerdrage:
updates common-arcana to add ritual tie support syntax in spells
tie_focus: pack #name of item focus is tying to


rcuhljr:
Manually merged.

PR #1717, opened 2017-06-16
Added necromancer EOTB support to attunement script

intangibles:
Updated attunement script to allow necromancers to use power walk feature of attunement.

Uses EOTB.

yaml variable:  
force_power_walk: true
default: false

### DIFF
--- a/attunement.lic
+++ b/attunement.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#attunement
 =end
 
-custom_require.call(%w[common common-travel drinfomon])
+custom_require.call(%w(common common-travel drinfomon))
 
 class Attunement
   include DRC
@@ -12,6 +12,7 @@ class Attunement
     settings = get_settings
     @stationary_skills_only = settings.crossing_training_stationary_skills_only
     @hometown = settings.hometown
+    @force_power_walk = settings.force_power_walk
 
     train_attunement
   end
@@ -24,9 +25,13 @@ class Attunement
       return
     end
 
-    room_list = if @stationary_skills_only || DRStats.necromancer?
+    room_list = if @stationary_skills_only || (DRStats.necromancer? and !@force_power_walk)
                   [Room.current.id]
                 else
+                  if DRStats.necromancer? and @force_power_walk
+                    put "prep eotb"
+                    bput "cast", "You whisper"
+                  end
                   get_data('town')[@hometown]['attunement_rooms']
                 end
 
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -81,6 +81,7 @@ buff_nonspells:
 dance_actions_stealth:
 thanatology:
 necromancer_healing:
+force_power_walk: false
 weapon_training:
 combat_trainer_target_increment: 3
 combat_trainer_action_count: 10


attunement.lic
```diff
@@ -24,9 +25,13 @@ class Attunement
       return
     end
 
-    room_list = if @stationary_skills_only || DRStats.necromancer?
+    room_list = if @stationary_skills_only || (DRStats.necromancer? and !@force_power_walk)
```
jayRyan24: Could this be better handled by using a waggle_set of ```eotb?``` instead of a new setting?

If necromancer and waggle_set !exist do

attunement.lic
```diff
@@ -24,9 +25,13 @@ class Attunement
...
                   [Room.current.id]
                 else
+                  if DRStats.necromancer? and @force_power_walk
```
jayRyan24: You've already checked for a force power walk by the time you've gotten in here, this check is redundant.

attunement.lic
jayRyan24: I do this with prehunt_buffs btw, so that's an example. hunting-buddy.lic

rpherbig: EOTB is an invis effect, so be aware that it may get dropped by other scripts. I think `release_invisibility` is the function you want to search for - it's used in several other town scripts.

rpherbig: I like the idea of a waggle_set, but might name it something more generic. Would ROC/ROG be good enough to powerwalk?

jandersson: Id prefer the waggle set as well, making it optional cast. Some attunement rooms arent in justice (e.g. Shards rooms are all outside of town) so the eotb wouldn't be necessary.

attunement.lic
```diff
@@ -24,9 +25,13 @@ class Attunement
...
+                    put "prep eotb"
+                    bput "cast", "You whisper"
```
intangibles: Negative. I viewed this as a more high level item since attunement can be learned in combat much easier and only at higher end attunements do you have to supplement with power walk.

It's a good suggestion and other commentors above have suggested using waggle_sets.  I'm planning on adding that in. 

Thank you!

attunement.lic
intangibles: A waggle set is a good idea.  I'm also a fan of a more generic name since moon mages could technically use RF/SOV as well.

ROC/ROG is not good enough for powerwalk because the main danger comes from players ACCUSING the necromancer once they have been outed and/or are known. 

There is really no downside to using EOTB as far as I know. I don't see why anyone would want to turn it off if they have it. I'd also hate to make an edge case for not having EOTB when literally every necromancer will have it.

intangibles: Also, you mentioned duration issues for younger folk, adding in a spell stance switch could be useful.

attunement.lic
jandersson: There is the existing `town-training` waggle set you could probably make use of

rpherbig:
Does the `town-training` waggle set mean we can close this PR?

intangibles:
@rpherbig 

No. Power-walking is disabled if you are playing a necromancer with original code.  Plus town-training waggle has never been reliable or useful for necromancers.  There is no way to gauge the duration of crossing-training, or what order the skills go in.  Town-training activates once as far as I can tell and buffs fall off.  Plus, invisibility can sometimes interfere with activities so it's best to be specific as possible with which activities EOTB applies to.

rpherbig:
That makes sense, thank you.

jandersson:
Town-training waggle is called on lines 74-76 in `crossing-training`. You could perhaps call it again before the attunement script is called. This would let the user train without the spell if they decided they didn't need it; it would also let them cast any other spell they like with their own settings.

rpherbig:
I think there is a difference between 'cast these spells before going into town' and 'cast these spells before powerwalking', so I'm in favor of a new waggle set instead of reusing `town-training`.

attunement.lic
jayRyan24: It does interfere at lower circles with using autopaths like Dokt. I had to control this with befores and afters until I got CF.

jandersson:
Agree on the distinction - Pulling for any waggle set though, if you think its reasonable. Just think it would empower the user to say 'i want X spell when I do Y', instead of the script deciding.

intangibles:
Yeah, new plan will be to pull a user defined waggle set!  I haven't messed with any of the coding on the waggle sets so small learning curve while I figure it out.

jandersson:
Ah nice! Might i suggest a waggle set named `attunement`? Naming waggle sets by script name feels right to me :)

jandersson:
Calling them is pretty simple too, look at how theyre called in `astrology`
https://github.com/rpherbig/dr-scripts/blob/master/astrology.lic#L40

rpherbig:
Overcome by 2a8643c6b8aa9d3ee40492e926ed64169b12a69b

PR #1719, opened 2017-06-17
Temporarily modify crossing bank id's to use Dirge due to bank closure

jandersson:
The bank in Crossing is closed due to an invasion.


jandersson:
Banks open

PR #1723, opened 2017-06-18
add match text for unrepairable item

jandersson:
Some items can't be repaired but are nice to have in a gear list. An example is a TM focus or gear used for appraisal. This is also needed to prevent the script from hanging


jandersson:
If we ever get a `dont_repair` flag for gear then seeing this match text could issue a warning to the user about setting that flag for the item.

PR #1724, opened 2017-06-18
Use Dump Junk in Crossing-Training

jandersson:
This goes off the setting used by combat-trainer, `dump_junk` and issues a call for the janitor if there are 10 or more items on the ground.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -61,13 +61,12 @@ class CrossingTraining
   end
 
   def dump_junk
-    return unless settings.@dump_junk
+    return unless @settings.dump_junk
     return unless Time.now - @junk_timer < 300
     return unless DRRoom.room_objs.size > 13
 
     fput('dump junk')
     @junk_timer = Time.now
-    end
   end
 
   def main
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -62,7 +62,7 @@ class CrossingTraining
 
   def dump_junk
     return unless @settings.dump_junk
-    return unless Time.now - @junk_timer < 300
+    return unless Time.now - @junk_timer > 300
     return unless DRRoom.room_objs.size > 13
 
     fput('dump junk')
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -89,6 +96,7 @@ class CrossingTraining
     if trash_nouns.any? { |noun| /\b#{noun}/i =~ GameObj.left_hand.noun } && !@equipment_manager.is_listed_item?(left_hand)
       dispose_trash(left_hand)
     end
+    dump_junk if DRRoom.room_objs.size > 9 && @settings.dump_junk
```
rpherbig: I'd prefer to see these two `if`s moved into guard clauses in `dump_junk`. What do you think?

crossing-training.lic
```diff
@@ -60,6 +60,13 @@ class CrossingTraining
     Flags.add('research-complete', '^Breakthrough!')
   end
 
+  def dump_junk
+    if Time.now - @junk_timer > 300
```
rpherbig: I'd prefer to see this conditional inverted and used as a guard clause (so `return if ...`). What do you think?

crossing-training.lic
```diff
@@ -89,6 +96,7 @@ class CrossingTraining
     if trash_nouns.any? { |noun| /\b#{noun}/i =~ GameObj.left_hand.noun } && !@equipment_manager.is_listed_item?(left_hand)
       dispose_trash(left_hand)
     end
```
rpherbig: Why is the threshold here 9, but 13 in `combat-trainer`?

jandersson: Plucked it out of a hat. A lower number seems good to keep the trash down though. Change it to 13?

crossing-training.lic
jandersson: Sounds good

crossing-training.lic
jandersson: Sounds good

jandersson:
Suggested changes added

jandersson:
Typos fixed!

PR #1726, opened 2017-06-18
Add repair every instead of timer

jayRyan24:
I think this is a better solution. Base yaml defaults to every iteration. This is customizable per player's yamls. I'm not really sure how effective a timer is, but this is "I go on X number hunts, then I repair my gear".


jandersson:
Would it be possible to keep the timer? I just started using it to only repair once every 24 hours

jandersson:
Thanks! The timer needs to be a UserVar so it works right, but i will PR that soon

jayRyan24:
I'm confused, was this not working right before? The timer is a YAML setting. I added it back in.

jandersson:
The snapshot time is an instance variable so it gets lost when the script ends. If the snapshot time is a uservar , we can lookup the last time we repaired something

jandersson:
Nevermind the timer isn't in crossing-repair its in training manager.. it should work just fine

training-manager.lic
```diff
@@ -80,10 +80,16 @@ class TrainingManager
 
   def check_repair
     return false if @skip_repair
-    return false if Time.now - @repair_timer_snap < @repair_timer
+    
+    if Time.now - @repair_timer_snap > @repair_timer
+      @repair_timer_snap = Time.now
```
rpherbig: Should `@repair_every_counter` be reset in this case?

training-manager.lic
```diff
@@ -80,10 +80,16 @@ class TrainingManager
...
+      return true
+    elsif @repair_every_counter+=1 >= @repair_every
+      @repair_every_counter = 0
```
rpherbig: Should 1@repair_timer_snap` be reset in this case?

training-manager.lic
```diff
@@ -80,10 +80,16 @@ class TrainingManager
 
   def check_repair
     return false if @skip_repair
-    return false if Time.now - @repair_timer_snap < @repair_timer
+    
+    if Time.now - @repair_timer_snap > @repair_timer
+      @repair_timer_snap = Time.now
+      return true
+    elsif @repair_every_counter+=1 >= @repair_every
```
rpherbig: The `els` is unnecessary (i.e. this can just be an `if`) if that would let you format the code to be clearer.

jayRyan24:
If we reset them both at the same time, then We don't need the ifs. Personally, I think the timer is redundant and not as useful as the repair every cycle approach.

jayRyan24:
Not sure what happened to my individual comments. Basically, if we keep both of them, then it allows for us to say repair every time cycle or total cycles.

training-manager.lic
jayRyan24: Read comments below, not sure what happened to all my individual comments.

jandersson:
Why is it redundant? How would I specify, repair once every 24 hours in the iterative system?

rpherbig:
Personally, I think the count is more useful (since 'number of hunting trips' is what determines how damage your gear is). That said, I can see how some people prefer the time-based approach, so I don't see the harm in leaving both in.

My first thought was that it should be an *or* check - we repair if either condition is met. In that case I think we would want to reset both vars each time we repair (regardless of why we repair). As long as each conditional checks for `nil`, the user can turn off a check they don't care about.

jayRyan24:
If you are spending X number of minutes hunting and then X number of minutes in x'ing training, just divide that by 24 hours and there's the amount of times you want to cycle before repairing. Doesn't this number get removed once you stop training manager? If so, you could end up never repairing your gear.

jandersson:
It will get removed (this could be an argument for making the timer a uservar). Right, id see the timer vs iterations as a relative vs absolute approach - both are useful.

PR #1727, opened 2017-06-19
Updated knitting to not force Crossing yarn purchase.  Updated Shaping to allow for non-workorder training

adamgoerdt:


crossing-training.lic
```diff
@@ -954,6 +984,17 @@ class CrossingTraining
...
   end
+  
+  def shape(chapter, unique_name, item)
+    return unless buy_wood
+
+    walk_to(@training_room)
+
+    wait_for_script_to_complete('shape', ['trash', chapter, unique_name, 'balsa', item])
+
+    fput('get my lumber')
```
rpherbig: Can you change this to `bput` (with match messages)?

crossing-training.lic
```diff
@@ -974,9 +1015,16 @@ class CrossingTraining
     dispose_trash(left_hand)
   end
 
+  def buy_wood
+    crafting_data = get_data('crafting')
+    return unless money_for_training?(700, 'Engineering')
+    order_item(crafting_data['shaping'][@settings.hometown]['stock-room'], 11)
+    fput('stow my lumber')
```
rpherbig: Can you change this to `bput` (with match messages)?

shape.lic
```diff
@@ -140,13 +144,26 @@ class Shape
...
+      logbook_item
+    when /stow/
+      stow_item(@noun)
+    when /trash/
+      dispose_trash(@noun)
+    end
     exit
   end
+  
+  def logbook_item
```
rpherbig: This looks nearly identical to the method in `sew.lic`. Is there any benefit to moving this to a `common-crafting` method?

crossing-training.lic
```diff
@@ -954,6 +984,17 @@ class CrossingTraining
       false
     end
   end
+  
+  def shape(chapter, unique_name, item)
+    return unless buy_wood
+
+    walk_to(@training_room)
```
rpherbig: Should this instead do something similar to https://github.com/rpherbig/dr-scripts/blob/master/workorders.lic#L237 ?

adamgoerdt:
Requested changes have been made

common-crafting.lic
```diff
@@ -142,4 +142,13 @@ module DRCC
     end
     true
   end
+  
+  def logbook_item(logbook)
+    get_item("#{logbook} logbook")
+    case bput("bundle my #{@noun} with my logbook", 'You notate the', 'This work order has expired', 'The work order requires items of a higher quality')
```
rpherbig: Does `@noun` work here? I thought this would need to be a parameter to the function...

crossing-training.lic
```diff
@@ -974,10 +1015,17 @@ class CrossingTraining
...
+    return unless money_for_training?(700, 'Engineering')
+    order_item(crafting_data['shaping'][@settings.hometown]['stock-room'], 11)
+    bput('stow my lumber','You put your')
+  end
+  
   def buy_yarn
     return unless money_for_training?(700, 'Outfitting')
-    order_item(16_667, 13)
-    fput('stow my yarn')
+    order_item(crafting_data['tailoring'][@settings.hometown]['stock-room'], 13)
```
rpherbig: Does this work? I don't see where `crafting_data` is declared in this function.

adamgoerdt: I thought I had that in there.  Added the method to pull the data, similar to buy_wood

common-crafting.lic
```diff
@@ -142,4 +142,13 @@ module DRCC
     end
     true
   end
+  
+  def logbook_item(logbook,noun)
+    get_item("#{logbook} logbook")
```
rpherbig: Where is `get_item` declared?

common-crafting.lic
```diff
@@ -142,4 +142,13 @@ module DRCC
...
+    case bput("bundle my #{noun} with my logbook", 'You notate the', 'This work order has expired', 'The work order requires items of a higher quality')
+    when 'This work order has expired', 'The work order requires items of a higher quality'
+      dispose_trash(noun)
+    end
+    stow_item("#{logbook} logbook")
```
rpherbig: Where is `stow_item` declared?

common-crafting.lic
adamgoerdt: I though it was part of DRCI, but I guess not.  Updated to use bputs

common-crafting.lic
```diff
@@ -142,4 +142,13 @@ module DRCC
...
     true
   end
+  
+  def logbook_item(logbook,noun)
+    DRC.bput("get my #{logbook} logbook", "You get")
+    case bput("bundle my #{noun} with my logbook", 'You notate the', 'This work order has expired', 'The work order requires items of a higher quality')
+    when 'This work order has expired', 'The work order requires items of a higher quality'
+      dispose_trash(noun)
+    end
+    DRC.bput("stow #{logbook} logbook", 'You put')
```
rpherbig: Will the logbook get misplaced if you have a crafting bag set?

adamgoerdt: For a logbook, does it matter?  If it does then we should probably move this method back into shape.lic and sew.lic

rpherbig: It _should_ be ok if the get function uses the same approach as the stow function (i.e. both use crafting bag or neither do).

rpherbig:
@rcuhljr Have you had a chance to review this?

common-crafting.lic
adamgoerdt: What about something like this: It will grab the logbook from wherever it is on the person, but stow it in the crafting container

rcuhljr:
Sorry I missed this for so long, this looks pretty awesome, and some good refactors. One thing I think we should be more thoughtful about ahead of time is not even trying to do shaping if their hometown settings don't have any data to try and save people from running shaping where it's not yet configured, but we don't do it for any other crafts yet.

PR #1733, opened 2017-06-20
Add waggle set support to astro

jayRyan24:
When casting Astro, it will recast the buffs even if they already exist. I added support for using waggle sets for this. The latter code should be removed after a short amount of time, this will give people the chance to prepare for it not being there.


jandersson:
Will need to move the RtR check (line 46) out of the deprecated `else` block to the buff.lic script line 35.. 

Heres what i tested:
```
  def check_spells(spells)                                                                                               
    check_osrel                                                                                                          
    spells.each do |name, data|                                                                                          
      next if DRSpells.active_spells[name] && (data['recast'].nil? || DRSpells.active_spells[name].to_i > data['recast'])
      next if name == 'Read the Ripples' && !Flags['rtr-expire']                                                         
```
Without the check RtR will be recast on every run of astrology

astrology.lic
```diff
@@ -37,12 +37,17 @@ class Astrology
...
-    settings.astrology_buffs['spells'].each do |spell|
-      if spell['abbrev'].casecmp('rtr').zero?
-        next if !Flags['rtr-expire'].nil? && !Flags['rtr-expire']
+    if settings.waggle_sets['astrology']
+      wait_for_script_to_complete('buff', ['astrology'])
+    else
+      #this should probably be deprecated, this doesn't allow you to NOT cast the spells if they are up already
+      settings.astrology_buffs['spells'].each do |spell|
+        if spell['abbrev'].casecmp('rtr').zero?
+          next if !Flags['rtr-expire'].nil? && !Flags['rtr-expire']
```
jandersson: Will need to move the RtR check (line 46) out of the deprecated `else` block to the buff.lic script line 35.. 

Heres what i tested:
```
  def check_spells(spells)                                                                                               
    check_osrel                                                                                                          
    spells.each do |name, data|                                                                                          
      next if DRSpells.active_spells[name] && (data['recast'].nil? || DRSpells.active_spells[name].to_i > data['recast'])
      next if name == 'Read the Ripples' && !Flags['rtr-expire']                                                         
```
Without the check RtR will be recast on every run of astrology

jandersson: I tried out the whole PR, it works good for me

PR #1752, opened 2017-06-30
Enable bankbot when bank is disabled

intangibles:
Change allows use of bankbot when bank bypass is on. This will also deposit all coins with the bankbot, not just the local currency, since no coins are exchanged with bank bypass on. Still retains threshold functionality of local currency.

Retains all previous functionality.

Usage is activated if bank bypass is on and a bankbot name is present.

if settings.sell_loot_skip_bank && settings.bankbot_name


sell-loot.lic
```diff
@@ -82,6 +98,9 @@ class SellLoot
     return unless @bankbot_room_id
     return unless @bankbot_deposit_threshold
 
+    tip_currency = @town_data[hometown]['currency']
+    keep_copper = @bankbot_deposit_threshold*-1 if keep_copper == -1
+
```
rcuhljr: The only thing we care about hometown for in this method was pulling out currency and copper on hand. It seems like it makes more sense to rename this method to `give_money_to_bankbot(currency, keep)` and kill the keep_copper hack and kill the `@bankbot_deposit_threshold` guard at the top of this method.

Doing that we change the above calls to the more reasonable 
```
['Lirums', 'Kronars', 'Dokoras'].each{|currency| give_money_to_bankbot(currency, currency == real_currency ? @bankbot_deposit_threshold : 0) }`

File: sell-loot.lic [hunk 2]
```diff
@@ -18,22 +18,38 @@ class SellLoot
...
 
+    temp_currency = @town_data[@actual_fucking_hometown]["currency"]
     sell_gems("#{settings.gem_pouch_adjective} pouch") if settings.sell_loot_pouch
 
     check_spare_pouch(settings.spare_gem_pouch_container, settings.gem_pouch_adjective) if settings.spare_gem_pouch_container
 
     sell_bundle if settings.sell_loot_bundle
 
-    return if settings.sell_loot_skip_bank
+    if settings.sell_loot_skip_bank && settings.bankbot_name
```
rcuhljr: I'm not a fan of this logical combo, If I have skip bank set I don't expect to be going to a bankbot. 
It seems like our options should be, Normal banking, No Banking, Bank Bot with exchange, Bank Bot without exchange. I'd kind of just like to add a `bankbot_skip_exchange: true` flag.

intangibles: I originally had that but figured I could pare down by assuming if you had a bankbot name listed then you wanted to use it.  I've got no problem adding back in.

rcuhljr: Yeah, but bankbots can go offline, I could see someone just toggling the skip_bank until their bankbot is back online. If' it's not too much trouble swapping to the explicit setting would be my preferred option.

sell-loot.lic
intangibles: No problem. I was trying not to interfere with the already established methods but will give it a shot!

Thanks!

common-money.lic
```diff
@@ -29,7 +29,11 @@ module DRCM
   end
 
   def wealth(hometown)
-    currency = get_data('town')[hometown]['currency']
```
rpherbig: This feels a little hacky to me. Could we have two methods (and the one that looks up your hometown currency could call the other one)?

common-money.lic
```diff
@@ -29,7 +29,11 @@ module DRCM
...
+    if hometown =~ /kronars|lirums|dokoras/
```
rpherbig: Should this regex ignore case?

sell-loot.lic
```diff
@@ -19,24 +19,31 @@ class SellLoot
...
 
-    return if settings.sell_loot_skip_bank
-    exchange_coins
+    @keep_coppers = parse(settings.sell_loot_money_on_hand.split(' ')) + @bankbot_deposit_threshold
 
-    keep_copper = parse(settings.sell_loot_money_on_hand.split(' '))
-    deposit_coins(keep_copper)
-    check_bankbot(settings.hometown, keep_copper)
+    if settings.sell_loot_skip_bank && settings.bankbot_skip_bank
+      ['lirums', 'kronars', 'dokoras'].each{|currency|
```
rpherbig: I think we have these currency names hardcoded somewhere else too. Maybe `bankbot`? Is it worth putting this in a common place?

common-money.lic
rcuhljr: Agreed.

sell-loot.lic
rcuhljr: I can't say I care much about three super common pieces of information that everyone knows, but it wouldn't be hard to move it into common-money.

intangibles: It actually may be a good idea. I had to go through and make sure the cases were the same for all the moneys.

common-money.lic
intangibles: Technically, see note below. I went through and made sure all the cases were the same.

intangibles:
closing this PR. mishap with rebase.

PR #1758, opened 2017-07-01
Added TDP Gain

adamgoerdt:


learned.lic
```diff
@@ -17,9 +17,16 @@ DRSkill.list.select { |item| item.current - item.baseline > 0 }.sort_by { |item|
   respond(output)
 end
 
+
+
 total = DRSkill.list.select { |item| item.current - item.baseline > 0 }.map { |skill| DRSkill.gained_exp(skill.name) }.inject(&:+).round(2)
+tdps = DRSkill.list.select { |item| item.current - item.baseline > 0 }.map { |skill| (DRSkill.gained_exp(skill.name) * DRSkill.getrank(skill.name))/200 }.inject(&:+).round(2)
```
rpherbig: Why does it `select` on `item.current - item.baseline > 0`? It is possible to lose TDPs if you die.

adamgoerdt:
I'm not really counting TDPs, just calculating based on exp gain.

If the select needs to be changed here it should be changed for the exp
portion as well.

Thoughts?

rpherbig:
Hey @rcuhljr - something to think about in the future. But it's not related to this change, so I'll merge this PR.

Issue #1757, opened 2017-07-01
Engineering work orders


mwcarew:
Getting an error when making pendants for engineering. Shape.lic isn't getting finished short leather cords or some such.

https://pastebin.com/xJXU5rvh

mwcarew:
This is with shaping work orders in the YAML and 111 Engineering. Done in Crossing.

BastionWT:
Also broken for me, same issue. This started happening in the last 2 days. Pendants recipe is not working.

adamgoerdt:
Updated recipe in PR 1759 should resolve this

rpherbig:
@leustyin I think the problem is that `wood pendant` isn't in `base-recipes`.

PR #1759, opened 2017-07-01
Swapped pendant in favor or earrings

adamgoerdt:
Fixes issue 1757


crossing-training.lic
```diff
@@ -1009,7 +1009,7 @@ class CrossingTraining
   def shape(chapter, unique_name, item)
     return unless buy_wood
 
-    find_shaping_room(@hometown, @engineering_room)
```
rpherbig: Is this supposed to change?

rpherbig:
I added the recipe and closed #1757. Do you still want to make this change?

adamgoerdt:
If the recipe fixes it then we don't need to swap the product.  I'll cancel
this PR

Sarvatt:
It's not buying the cord needed for the pendant and still getting hung up.

rpherbig:
Is this after restarting dependency?

Sarvatt:
Yep restarted the client and I had wood pendant in my base-recipes when the error hit again. The cords are in a crate and not bought via ORDER, not sure if that matters.

rpherbig:
I don't know if the script supports that. Good call.

PR #1764, opened 2017-07-03
Update Base-Spells with AP and Bard Spells

jandersson:
Data should have all the AP and bard spells now. I added a few cleric ones as well but not all of them.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -490,6 +490,10 @@ spell_data:
     skill: Targeted Magic
     heavy: true
     mana: 15
+  Drums of the Snake:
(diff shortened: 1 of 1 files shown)


data/base-spells.yaml
```diff
@@ -604,20 +685,41 @@ spell_data:
...
+    skill: Targeted Magic
+    abbrev: HOT
+    mana: 7
   Hands of Justice:
     skill: Utility
     abbrev: HOJ
     mana: 5
     recast: 1
   Hands of Lirisa:
-    skil: Augmentation
```
rpherbig: Nice catch

PR #1769, opened 2017-07-05
additional bput slip stalk confirmation messages

intangibles:
new confirmation message on bput slip stalk

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -270,7 +270,7 @@ module DRC
 
   def hide?(hide_type = 'hide')
     unless hiding?
-      case bput(hide_type, 'Roundtime', 'too busy performing', 'can\'t see any place to hide yourself', 'Stalk what', 'You\'re already stalking')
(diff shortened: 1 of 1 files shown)


rcuhljr:
So it was possible to slip without RT?

intangibles:
Yeah, unfortunately. I never caught this until things were too hard.  Maybe I should return hide? again on failure? Since there is no RT.

[combat-trainer]>stalk

You quickly slip into hiding to prepare to stalk.
The sky giant notices your attempt to hide!
The sky giant discovers you, ruining your hiding place!
Stalking is an inherently stealthy endeavor, try being out of sight.

intangibles:
Let's hold off on this PR for now then, I'll do some testing to see how returning on failure with no RT works.

intangibles:
Although it's possible you could get into a crazy stalk loop.  Maybe not such a good idea.

PR #1770, opened 2017-07-06
Commonize Buff Methods

jandersson:
I wanted to split up  a users waggle sets and only send parts of them to buff (this is for an astrology update).  I didn't see an easy way to do this because `buff.lic` takes a waggle set name as an argument and then casts the entire set from top to bottom. This is an issue if you want to only cast some of the spells, or perhaps modify the set data. To remedy this, I moved a number of methods over into common-arcana with `check_spells` renamed to `do_waggle_set`. Now `buff.lic` can still be called, or `do_waggle_set` can be called with spell data as an argument.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -65,17 +65,6 @@ module DRCA
     true
   end
 
-  def buff(spell, settings)
-    if spell['ritual']
-      ritual(spell)
-      return
-    end
-    prepare?(spell['abbrev'], spell['mana'], spell['symbiosis'])
-    find_charge_invoke_stow(settings.cambrinth, settings.stored_cambrinth, settings.cambrinth_cap, settings.dedicated_camb_use, spell['cambrinth'])
-    waitcastrt?
-    cast?(spell['cast'], spell['symbiosis'], spell['before'], spell['after'])
-  end
-
   def ritual(spell)
     DRCT.retreat
     DRC.release_invisibility
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -288,7 +288,7 @@ module DRCA
     release_cyclics if data['cyclic']
 
     if spell['ritual']
-      ritual(spell)
+      ritual(data)
       return
     end
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -276,7 +276,7 @@ module DRCA
 
     release_cyclics if data['cyclic']
 
-    if spell['ritual']
+    if data['ritual']
       ritual(data)
       return
     end
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -228,7 +228,7 @@ module DRCA
     end
   end
 
-  def cast_spells(settings, spells)
+  def cast_spells(spells, settings)
     infuse_om(!settings.osrel_no_harness, settings.osrel_amount)
     spells.each do |name, data|
       next if DRSpells.active_spells[name] && (data['recast'].nil? || DRSpells.active_spells[name].to_i > data['recast'])
@@ -240,7 +240,7 @@ module DRCA
       if data['ritual']
         ritual(data)
       else
-        cast_spell(settings, data)
+        cast_spell(data, settings)
       end
     end
   end
@@ -267,7 +267,7 @@ module DRCA
     echo "Run `#{$clean_lich_char}e autostart('moonwatch')` to avoid this in the future"
   end
 
-  def cast_spell(settings, data)
+  def cast_spell(data, settings)
     return unless data
     return unless settings
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -257,24 +257,24 @@ module DRCA
   end
 
   def set_moon_data(data)
-    if data['moon']
-      check_moonwatch
-      # Look for a moon that will be up for the next few minutes
-      unless moon = UserVars.moons.find { |moon_name, moon_data| UserVars.moons['visible'].include?(moon_name) && moon_data['timer'] >= 4 }.first
-        echo "No moon available to cast #{data['abbrev']}"
-        return
-      end
-      data['cast'] = "cast #{moon}"
+    return unless data['moon']
+
+    check_moonwatch
+    # Look for a moon that will be up for the next few minutes
+    unless moon = UserVars.moons.find { |moon_name, moon_data| UserVars.moons['visible'].include?(moon_name) && moon_data['timer'] >= 4 }.first
+      echo "No moon available to cast #{data['abbrev']}"
+      return  # Return nil to indicate that setting moon casting data failed
     end
+    data['cast'] = "cast #{moon}"
     data
   end
 
   def check_moonwatch
-    unless Script.running? 'moonwatch'
-      echo 'moonwatch is not running. Starting it now'
-      custom_require.call('moonwatch')
-      echo "Run `;e autostart('moonwatch')` to avoid this in the future"
-    end
+    return unless !Script.running? 'moonwatch'
+
+    echo 'moonwatch is not running. Starting it now'
+    custom_require.call('moonwatch')
+    echo "Run `#{$clean_lich_char}e autostart('moonwatch')` to avoid this in the future"
   end
 
   def cast_spell(settings, data)
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -256,7 +256,7 @@ module DRCA
     end
   end
 
-  def check_moons(data)
+  def set_moon_data(data)
     if data['moon']
       check_moonwatch
       # Look for a moon that will be up for the next few minutes
@@ -278,7 +278,7 @@ module DRCA
   end
 
   def cast_spell(settings, data)
-    return unless data = check_moons(data)
+    return unless data = set_moon_data(data)
       
     release_cyclics if data['cyclic']
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -237,11 +237,7 @@ module DRCA
         pause 15
       end
 
-      if data['ritual']
-        ritual(data)
-      else
-        cast_spell(data, settings)
-      end
+      cast_spell(data, settings)
     end
   end
 
@@ -265,6 +261,10 @@ module DRCA
     echo 'moonwatch is not running. Starting it now'
     custom_require.call('moonwatch')
     echo "Run `#{$clean_lich_char}e autostart('moonwatch')` to avoid this in the future"
+  end
+
+  def buff(spell, settings)
+    cast_spell(spell, settings)
   end
 
   def cast_spell(data, settings)
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -78,6 +78,7 @@ module DRCA
 
   def ritual(spell)
     DRCT.retreat
+    DRC.release_invisibility
     DRC.bput('stance set 100 0 80', 'Setting your')
 
     prepare?(spell['abbrev'], spell['mana'], spell['symbiosis'])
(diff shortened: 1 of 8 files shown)


common-arcana.lic
```diff
@@ -236,4 +236,57 @@ module DRCA
...
+      if data['ritual']
+        ritual(data)
+      else
+        cast_spell(settings, data)
+      end
+    end
+  end
+
+
+  def cast_spell(settings, data)
```
jandersson: I waffled between having the settings be an argument or having the method call get_settings.

jandersson:
just to show the motivation behind the change; here is what i have in `astrology.lic` initialize method
```ruby
    buffs = @settings.waggle_sets['astrology']
    # Pop out the RTR spell data and save it for later
    @rtr_data = buffs.select { |spell| spell.eql?('Read the Ripples') }
    buffs.reject! { |spell| spell.eql?('Read the Ripples') }
```
This lets astrology run its buff routine with RTR removed, then when its time to train with RTR, it gets cast individually. Now RTR doesn't need to be listed last in the astrology waggle_set and it lets you train using other methods before running RTR.

common-arcana.lic
rpherbig: Avoid calling get_settings in a `common-*` module

common-arcana.lic
```diff
@@ -236,4 +236,57 @@ module DRCA
       return parse_mana_message(mana_msg)
     end
   end
+
+  def check_osrel(osrel_no_harness, osrel_amount)
+    return unless DRSpells.active_spells['Osrel Meraud'] && DRSpells.active_spells['Osrel Meraud'] < 90
+    infuse_om(!osrel_no_harness, osrel_amount)
+  end
+
+  def do_waggle_set(settings, spells)
```
rpherbig: There's no reason this has to be called from waggle (or buff...). Maybe `cast_spells`?

common-arcana.lic
```diff
@@ -236,4 +236,57 @@ module DRCA
...
+    check_osrel(settings.osrel_no_harness, settings.osrel_amount)
```
rpherbig: Do we want to `check_osrel` before casting the entire set, or before casting each spell in the set?

common-arcana.lic
```diff
@@ -236,4 +236,57 @@ module DRCA
       return parse_mana_message(mana_msg)
     end
   end
+
+  def check_osrel(osrel_no_harness, osrel_amount)
```
rpherbig: Could this method be combined with `infuse_om`?

common-arcana.lic
```diff
@@ -236,4 +236,57 @@ module DRCA
...
+        ritual(data)
+      else
+        cast_spell(settings, data)
+      end
+    end
+  end
+
+
+  def cast_spell(settings, data)
+    if data['moon']
```
rpherbig: Is this `cast_spell` the same as what is used in combat-trainer or crossing-training?

common-arcana.lic
jandersson: Good point; the old name was better

common-arcana.lic
jandersson: Theyre different - they both deal with state in their respective scripts. I can see an issue with the names being the same.

common-arcana.lic
jandersson: This was the previous behaviour in `buff.lic`. I'm a little hesitant to change that without a cleric to test on.

common-arcana.lic
jandersson: Yea i think it can be. `check_osrel` could be made into a guard clause in `infuse_om`

jandersson:
Updated for latest feedback

rpherbig:
@rcuhljr look good to me. Can you give this a look?

common-arcana.lic
```diff
@@ -236,4 +237,52 @@ module DRCA
...
```
rcuhljr: I really kind of want this whole moon block moved into it's own method as a guard >.>

rcuhljr:
One little nitpick, but it's better than it was before.

jandersson:
Moved the moon stuff into its own method. Since it both checks for moons and sets part of the spell data, `cast_spell` can back out early if the data is not returned from `check_moons` when there are no moons.

common-arcana.lic
```diff
@@ -236,4 +238,56 @@ module DRCA
...
+    end
+  end
+
+  def set_moon_data(data)
+    if data['moon']
+      check_moonwatch
+      # Look for a moon that will be up for the next few minutes
+      unless moon = UserVars.moons.find { |moon_name, moon_data| UserVars.moons['visible'].include?(moon_name) && moon_data['timer'] >= 4 }.first
+        echo "No moon available to cast #{data['abbrev']}"
+        return
```
rpherbig: This probably needs to return `data`

common-arcana.lic
```diff
@@ -236,4 +238,56 @@ module DRCA
...
+      if data['ritual']
+        ritual(data)
+      else
+        cast_spell(settings, data)
+      end
+    end
+  end
+
+  def set_moon_data(data)
+    if data['moon']
```
rpherbig: You could make this a guard

common-arcana.lic
```diff
@@ -236,4 +238,56 @@ module DRCA
...
+        echo "No moon available to cast #{data['abbrev']}"
+        return
+      end
+      data['cast'] = "cast #{moon}"
+    end
+    data
+  end
+
+  def check_moonwatch
+    unless Script.running? 'moonwatch'
```
rpherbig: You could make this a guard

common-arcana.lic
```diff
@@ -236,4 +238,56 @@ module DRCA
...
+      echo 'moonwatch is not running. Starting it now'
+      custom_require.call('moonwatch')
+      echo "Run `;e autostart('moonwatch')` to avoid this in the future"
```
rpherbig: Should `;` be `$clean_lich_char`?

https://github.com/rpherbig/dr-scripts/blob/master/setupaliases.lic#L11

common-arcana.lic
jandersson: Id like to signal to `cast_spell` that `set_moon_data` failed to set the data by returning nothing. This way it can return early on line 281. If there's a better way to do this, I'm all for it.

rpherbig: I think returning nothing (`nil`) makes sense

jandersson: I dont like that it is destroying the data and isn't really clear why its doing that. Having a method be a guard and a setter function was a bit of a brainteaser.

jandersson: I will document the method to indicate what its doing. Code is documentation, but i think this is the instance where comments are good

mwest152:
I was using a hybrid method of astrology buffs and the waggle sets setup, and noticed I think I had a piercing gaze that expired mid Read the Ripples which kinda screwed things up.  I went back to the old setup.

jandersson:
Updated with latest changes, commented the nil return in `set_moon_data` and made certain lines into guards

common-arcana.lic
```diff
@@ -257,24 +257,24 @@ module DRCA
...
     data
   end
 
   def check_moonwatch
-    unless Script.running? 'moonwatch'
-      echo 'moonwatch is not running. Starting it now'
-      custom_require.call('moonwatch')
-      echo "Run `;e autostart('moonwatch')` to avoid this in the future"
-    end
+    return unless !Script.running? 'moonwatch'
```
rpherbig: This can be `return if`

common-arcana.lic
```diff
@@ -236,4 +238,56 @@ module DRCA
...
+      if data['ritual']
+        ritual(data)
+      else
+        cast_spell(settings, data)
+      end
+    end
+  end
+
+  def set_moon_data(data)
+    return unless data['moon']
```
rpherbig: Should this `return data ...`

common-arcana.lic
```diff
@@ -236,4 +238,56 @@ module DRCA
...
+  def check_moonwatch
+    return unless !Script.running? 'moonwatch'
+
+    echo 'moonwatch is not running. Starting it now'
+    custom_require.call('moonwatch')
+    echo "Run `#{$clean_lich_char}e autostart('moonwatch')` to avoid this in the future"
+  end
+
+  def cast_spell(settings, data)
+    return unless data = set_moon_data(data)
```
rpherbig: Am I correct that this will only ever return if `data` is `nil`? I wasn't expecting to see a conditional based on an assignment.

common-arcana.lic
```diff
@@ -236,4 +238,56 @@ module DRCA
...
+  end
+
+  def set_moon_data(data)
+    return unless data['moon']
+
+    check_moonwatch
+    # Look for a moon that will be up for the next few minutes
+    unless moon = UserVars.moons.find { |moon_name, moon_data| UserVars.moons['visible'].include?(moon_name) && moon_data['timer'] >= 4 }.first
+      echo "No moon available to cast #{data['abbrev']}"
+      return  # Return nil to indicate that setting moon casting data failed
```
rpherbig: I prefer to explicitly `return nil`. What do you think?

rpherbig:
How does `cast_spell` differ from the existing `buff` function?

common-arcana.lic
jandersson: Yep, thanks

common-arcana.lic
jandersson: Thats fine with me

common-arcana.lic
jandersson: I wanted to make the `if data['moon']` block into a guard as suggested earlier - this was the best solution i could think of. Any thoughts for a better way to handle this? The idea is to set `data['cast']` to cast on a moon if there is a visible moon, else return without casting.

jandersson:
Where is the `buff` function?

jandersson:
They're not as far as I can tell. Buff handles ritual data and cast_spell handles moon data. They should be combined i think. I like `cast_spell` because its more general.

common-arcana.lic
jandersson: I will break it into multiple lines, assign then check.

common-arcana.lic
```diff
@@ -236,4 +227,66 @@ module DRCA
...
+  def cast_spells(settings, spells)
+    infuse_om(!settings.osrel_no_harness, settings.osrel_amount)
+    spells.each do |name, data|
+      next if DRSpells.active_spells[name] && (data['recast'].nil? || DRSpells.active_spells[name].to_i > data['recast'])
+      while mana < 40
+        echo('Waiting on mana...')
+        pause 15
+      end
+
+      if data['ritual']
```
rpherbig: `cast_spell` does this check, is it still necessary here?

jandersson: It looks like belt & suspenders, ill remove it.

common-arcana.lic
```diff
@@ -64,19 +65,9 @@ module DRCA
     true
   end
 
-  def buff(spell, settings)
```
rpherbig: Were you still planning on leaving `buff` here (which would just call `cast_spell`)?

jandersson: Added it back

jandersson:
Reverted a commit to continue using `buff` instead of `cast_spells` until everyone can get their scripts updated.

rpherbig:
Looks great, thank you for the PR @jandersson .

A few thoughts that aren't part of this change, but I noticed while reviewing it:

* `ritual` should short circuit if `prepare?` fails (line 73) similar to how `cast_spell` behaves (line 284)
* There's some moon logic in `crossing-training` could probably be changed now to use `common-arcana`
* There's some moon logic in `combat-trainer` - same thing

If you're back in this area again, these would be great to clean up, but it's not essential.

jandersson:
Thanks, i'll keep that in mind if I'm editing these again.

PR #1787, opened 2017-07-15
base-recipes.yaml: Disable wood crook workorders.

Sarvatt:
2 people reported them as being broken today so just disable them for now.

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -819,7 +819,7 @@ crafting_recipes:
   noun: crook
   volume: 5
   type: shaping
-  work_order: true
+  work_order: false
   chapter: 9
 - name: a wood earcuff
   noun: earcuff


rpherbig:
Do you have a log?

This is one of the first recipes found in the Journeyman book. The `crook` recipe sticks in my mind since it's the one my guys always run into problems with when they need to upgrade from the Apprentice book.

I checked my logs and I've made a `crook` today.

Sarvatt:
I don't, neither person had a log just reported it getting stuck not being able to bundle it. I'll try to reproduce it myself.

rpherbig:
If it's failing on the `bundle` step, then it's more likely due to `common-crafting` being out of date. I suggest relogging the character to make sure everything is in sync.

Sarvatt:
Looking into it, I think the problem might be it trying to stow the crook in a container that can't hold a 19 length item if it is not good enough quality to bundle and getting stuck but I'm not sure yet. I'm still a few days off of being able to do the workorders. I haven't been able to get logs from anyone outside of a few more people saying it was hanging for them too, hopefully will have more info soon.

rpherbig:
This may or may not be related to #1797

Sarvatt:
I've been making crooks for 2 days now with no problems, I agree that it was probably another of the bugs that people hit. Lets just close this for now and I'll resend if I notice the problem.

PR #1788, opened 2017-07-17
Update clean-lumber.lic

fireballx777:
Added regex to recognize "woodcutting saw" in addition to just "wood saw".  Also changed saw/drawknife storage to the lumber storage container.

### DIFF
--- a/clean-lumber.lic
+++ b/clean-lumber.lic
@@ -24,7 +24,7 @@ class CleanLumber
   end
 
   def process_size(args, size)
-    while bput("get #{args.type} #{size} from my #{args.source}", 'You get', 'You carefully remove', 'What were you') != 'What were you'
+    while 'What were you' != bput("get #{args.type} #{size} from my #{args.source}", 'You get', 'You carefully remove', 'What were you')
       if size == 'deed'
         was_deed = true
         fput('tap deed')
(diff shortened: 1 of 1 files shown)


clean-lumber.lic
```diff
@@ -24,7 +24,7 @@ class CleanLumber
   end
 
   def process_size(args, size)
-    while bput("get #{args.type} #{size} from my #{args.source}", 'You get', 'You carefully remove', 'What were you') != 'What were you'
```
rpherbig: Why did you change the order of the things that are being compared?

clean-lumber.lic
```diff
@@ -35,23 +35,23 @@ class CleanLumber
         end
         size = right_hand.split.last
       end
-      unless bput('get my saw', 'wood saw', 'You get') == 'wood saw'
-        fput('stow saw')
+      unless /woodcutting saw|wood saw/ =~ bput('get my saw', 'wood saw', 'woodcutting saw', 'You get')
+        fput("put my saw in my #{args.storage}")
```
rpherbig: Is it always the case that the container in which you store lumber should be the same container in which you store your tools?

clean-lumber.lic
```diff
@@ -35,23 +35,23 @@ class CleanLumber
...
+      until 'you complete' == bput("cut #{size} with my saw", 'roundtime', 'you complete')
         waitrt?
       end
       pause
       waitrt?
-      fput('stow saw')
-      fput('get drawknife')
-      until bput("scrape #{size} with my drawknife", 'roundtime', 'work completes') == 'work completes'
+      fput("put my saw in my #{args.storage}")
+      fput('get my drawknife')
```
rpherbig: Can drawknives be stored on belts?

clean-lumber.lic
fireballx777: During testing, I was having issues with the comparisons not working, and switching the order got them to work -- I'm not sure why  If it's a big deal, I can go back and try them the original way again to try to get them working like that.

clean-lumber.lic
fireballx777: It would really be user preference, but I think the lumber container is the more likely tool storage location than the default stow container.

clean-lumber.lic
fireballx777: They can be stored on belts, yes,, but I used the same logic here as with the saw -- that the lumber container would be a more likely storage spot than the default stow container.  I haven't worked with the syntax of using a belt yet, and was just going for a simple change on this one.

clean-lumber.lic
rpherbig: Our style standard has been the order you saw. The order shouldn't matter, so I'm a bit concerned something else could be going on.

clean-lumber.lic
rpherbig: Works for me

clean-lumber.lic
rpherbig: Works for me

fireballx777:
Changed the comparisons back to convention.  Updated the drawknife and saw storage to use crafting_container from settings.

PR #1793, opened 2017-07-19
Communes

tulanegator:
commune.lic added to repos to quickly use various Cleric communes. it uses YAML settings already required by theurgy.lic. it does not gather or purchase materials but requires you already have most material in your theurgy_supply_container.

to see options:
;commune

to perform communes:
;commune tamsine
;commune eluned
;commune kertigen
;commune hodierna
;commune meraud
;commune truffenyi

Kertigen and Truffenyi may be used on others.

to carve beads (for use in Truffenyi or otherwise):
;commune carve (immortal or aspect)


rpherbig:
How much of the code in `theurgy.lic` were you able to reuse? If possible I'd like to change `theurgy.lic` to call into this when it does communes, though that would be outside the scope of the PR.

rpherbig:
There is a merge conflict in `data/base-spells.yaml` that needs to be resolved before this can be merged.

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+	  missing_item("#{@flint_lighter}")
+	end
+	empty_cleric_hands
+  end
+  
+
+  ######################## DEVOTION ########################
+	
+	def check_devotion()
+	  case bput('commune', @devotion_levels)
```
rpherbig: Could you check the index of the `bput` result in `@devotion_level` instead of the case/when block?

rpherbig:
Could you run Rubocop on this file? That would automatically update it with our script style.

`rubocop --auto-correct commune.lic`

commune.lic
```diff
@@ -0,0 +1,412 @@
+=begin
+  Documentation: ???
```
rpherbig: https://elanthipedia.play.net/Lich_script_repository#commune

Just make sure to update elanthipedia once this is merged.

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+  def sprinkle(item, target=@self)
+    case bput("sprinkle my #{item} on #{target}", 'You sprinkle', 'Sprinkle what', 'What were you referring to')
+	when 'You sprinkle'
+	  return true
+	else
+	  return false
+	end
+  end
+  
+  def wave_incense(target=@self)
```
rpherbig: If `wave_incense` is only ever called on line 302 without a parameter, why is there a parameter?

rpherbig: This comment may also apply to some of the other functions

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+    case bput("look at my #{item} in my #{@theurgy_supply_container}", 'You see nothing unusual', 'I could not find', 'What were you referring to')
+    when 'You see nothing unusual'
+	  return true
+    else
+	  echo("No #{item} found. Please make sure containers are open.")
+	  missing_item("#{item}")
+    end
+  end
+  
+  def holding?(noun)
```
rpherbig: This seems pretty useful and reusable... this might be a good candidate to be moved into `common-items` if we're duplicating the logic in other scripts...

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+	case bput("count holy water in my #{@water_holder}", 'left of the holy water', 'I could not find', 'What were you referring to')
+	when 'left of the holy water'
+	  return true
+	else
+	  echo('No holy water found. Please make sure containers are open.')
+	  missing_item('holy water')
+	end
+  end
+
+  def have_cleric_item?(item)
```
rpherbig: Is this the same as `DRCI.inside?`? https://github.com/rpherbig/dr-scripts/blob/master/common-items.lic#L79-L82

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+	else
+	  return false
+	end
+  end
+  
+  # must pass items in an array
+  def have_cleric_items?(items)
+    items.each do |item|
+	  if
+	    item == 'holy water' then have_holy_water?
```
rpherbig: I noticed that `have_holy_water?` and `have_cleric_item?` are predicates, yet their return value isn't used. Is the intent here that we exit if an item is missing?

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+	  ],[
+	   { name: 'Kertigen', regex: /^k.*/i, variable: true, description: 'Bless weapon or person.' },
+	   { name: 'target', regex: /\w/i, optional: true, description: "Target weapon or person. Defaults to self." }
+	  ],[
+       { name: 'Hodierna', regex: /^h.*/i, variable: true, description: 'Staunch bleeding.' }
+	  ],[
+       { name: 'Meraud', regex: /^m.*/i, variable: true, description: 'Increase mana in room.' }
+	  ],[
+	   { name: 'Truffenyi', regex: /^tr.*/i, variable: true, description: 'Get or return favor orb for you. Get favor orb for others (kneeling). Proper offering required in right hand.' },
+	   { name: 'target', regex: /\w/i, optional: true, description: "Target person. Defaults to self. O" }
```
rpherbig: Should "O" at the end of the description be removed?

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+	  ],
+	  [
+	   { name: 'Carve', regex: /^c.*/i, variable: true, description: 'Carve a prayer bead. Must have block or be able to forage a limb.' },
+	   { name: 'target', regex: /\w/i, optional: true, description: "Immortal for prayer bead. Defaults to Meraud. Must use full name." }
+	  ]
+    ]
+
+    args = parse_args(arg_definitions)
+
+    if args.to_h.empty?
```
rpherbig: I think this `if` can be removed. If the provided args do not match exactly 1 possible definition, `parse_args` will automatically display the possible definitions and exit the script: https://github.com/rpherbig/dr-scripts/blob/master/dependency.lic#L35-L41

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+      else
+	    commune_kertigen
+	  end
+	elsif args.Hodierna
+	  commune_hodierna
+	elsif args.Meraud
+	  commune_meraud
+	elsif args.Truffenyi
+	  if args.target
+	    commune_truffenyi("#{args.target}")
```
rpherbig: Do you need to do string interpolation here? I think passing `args.target` should be the same.

This applies to the other uses of `args.target` too.

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+       { name: 'Hodierna', regex: /^h.*/i, variable: true, description: 'Staunch bleeding.' }
+	  ],[
+       { name: 'Meraud', regex: /^m.*/i, variable: true, description: 'Increase mana in room.' }
+	  ],[
+	   { name: 'Truffenyi', regex: /^tr.*/i, variable: true, description: 'Get or return favor orb for you. Get favor orb for others (kneeling). Proper offering required in right hand.' },
+	   { name: 'target', regex: /\w/i, optional: true, description: "Target person. Defaults to self. O" }
+	  ],
+	  [
+	   { name: 'Carve', regex: /^c.*/i, variable: true, description: 'Carve a prayer bead. Must have block or be able to forage a limb.' },
+	   { name: 'target', regex: /\w/i, optional: true, description: "Immortal for prayer bead. Defaults to Meraud. Must use full name." }
```
rpherbig: Is Meraud a common enough choice to merit being the default? Or should this arg be required?

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+	  when 'You sense nothing special from your communing'
+	    @devotion_level = 0
+	  end
+	  return @devotion_level
+	end
+
+  ##################### CARVING ########################
+  
+  def create_bead(bead_type='meraud')
+    symbols = { 'kertigen' => 'raven', 'divyaush' => 'welkin', 'zachriedek' => 'raccoon',
```
rpherbig: I think you could move this to a higher scope and use it in the definition of the `target` arg (if you wanted to).

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+		      }
+	if symbols.key?("#{bead_type}")
+	  @shape = symbols["#{bead_type}"]
+	elsif symbols.value?("#{bead_type}")
+	  @shape = bead_type
+	else
+	  missing_item('a valid carving target')
+	end
+	echo "Carving: #{@shape}"
+    if GameObj.right_hand.noun == 'block'
```
rpherbig: Does the `exists?` check on line 216 already cover this?

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+  end
+  
+  def carve_bead(shape)
+    if exists?('shaper')
+	  empty_cleric_left_hand if GameObj.left_hand.noun != 'shaper'
+	  empty_cleric_right_hand if GameObj.right_hand.noun != 'block'
+	  fput('get my block') if GameObj.right_hand.noun != 'block'
+	  fput('get my shaper') if GameObj.left_hand.noun != 'shaper'
+	  while GameObj.right_hand.noun == 'block'
+	    case bput("shape my block to #{shape}", 'You fumble around but', 'Roundtime') # You skillfully begin', 'You adeptly begin', 'You carefully begin shaping', 'You confidently begin', 'Trying your best', 'With heavy strokes', 'You focus carefully', 'You focus slowly', 'You focus clumsily', 'You smile with satisfaction', 'You notice the', 'Pursing your lips', 'Your spirit soars with pride', 'You fumble around but'
```
rpherbig: Why are all of these messages commented out?

tulanegator: I was adding them as I tested but got new messages every test run. I found that using 'Roundtime' worked effectively without using each individual message. I will remove them.

commune.lic
tulanegator: Meraud orbs are required for Osrel Meraud so most Clerics need them after a death. It's my normal use for ;commune Truffenyi even though my Cleric gets favors from Kertigen.

commune.lic
tulanegator: It definitely should.

tulanegator:
Thank you for all of your input. I'll look into each of these issues and make changes. This is my first stand alone project so the input is really appreciated.

rpherbig:
Oh neat, apparently I can resolve the merge conflict from the web UI pretty easily.

And thank you for the contribution @tulanegator - we love to see pull requests :)

commune.lic
tulanegator: Yes the intent is to exit if a required item is missing. The 'else' in have_cleric_items? exits if the items required for the commune do not exist.

commune.lic
tulanegator: In case a future use of the script requires waving at something else. It could be eliminated in the current version of the script with no change in functionality.

commune.lic
tulanegator: Yes, I'm going to work on figuring that out... right now the check_devotion isn't even used. The intent is to be able to use it in the future for training and only use communes if your devotion is high enough (ex: don't use communes if devotion is lower than 13/16).

commune.lic
tulanegator: Eliminated

commune.lic
tulanegator: Thanks. Eliminated.

commune.lic
tulanegator: My newness to programming will show here. Are you saying that changing it to:

elseif args.Truffenyi
  commune_truffenyi("#{args.target}")
else
...

would have the same result? So if no "target" is defined it will just pass nothing and go to the default anyway?

commune.lic
tulanegator: Yes but it could carve a block, then tap a block, then carve a bead. The extra check just skips the tapping if the block already exists in the right hand. It was added so you can ;commune carve and then ;commune truffenyi without tapping or moving items. The action ;commune carve is intended to leave a prayer bead in your right hand.

commune.lic
tulanegator: Removed

commune.lic
rpherbig: I _think_ this code will not do that. The `if item == 'holy water'` is checked, and if true, then `have_holy_water?` will evaluate; but at that point we're already inside the `if` and the `else` can't be taken.

commune.lic
rpherbig: I err on the side of not implementing something until I need it, but this isn't a big deal.

commune.lic
rpherbig: I _think_ `theurgy.lic` does something like that, so you may see if you can reuse that code.

commune.lic
```diff
@@ -0,0 +1,412 @@
...
+	  ],[
+       { name: 'Hodierna', regex: /^h.*/i, variable: true, description: 'Staunch bleeding.' }
+	  ],[
+       { name: 'Meraud', regex: /^m.*/i, variable: true, description: 'Increase mana in room.' }
+	  ],[
+	   { name: 'Truffenyi', regex: /^tr.*/i, variable: true, description: 'Get or return favor orb for you. Get favor orb for others (kneeling). Proper offering required in right hand.' },
+	   { name: 'target', regex: /\w/i, optional: true, description: "Target person. Defaults to self. O" }
+	  ],
+	  [
+	   { name: 'Carve', regex: /^c.*/i, variable: true, description: 'Carve a prayer bead. Must have block or be able to forage a limb.' },
```
rpherbig: I recommend using the whole word `carve` here

commune.lic
rpherbig: There are two things here:
1. The only difference between `args.target` and `"#{args.target}"` is that the latter forces the value to be a string. However, args.target is already a string, so you don't need to do the string interpolation.
1. I'm not sure if you can change this to `commune_truffenyi(args.target)`. If that works, I think it would make the code clearer. If it doesn't, I recommend trying a different style that might also work:
```
def commune_truffenyi(target)
  target = target || @self
end
...
commune_truffenyi(args.target)
```

commune.lic
tulanegator: the have_holy_water? method that is called will check if holy water exists and, if it doesn't, it will exit the script. if have_holy_water? returns true (holy water DOES exist) then the script continues to run through each |item| and make sure we have the necessary items for the commune.

tulanegator:
I believe most of the requested changes are incorporated into the recently pushed "changes per comments." The biggest change is the devotion_check now checks against the index of @devotion_levels, rather than a 16 section when statement. I also cleaned up the way args.target passes for Kertigen and Truffenyi and used your suggestion of "target = target || @self"

Please let me know if you see any more necessary changes.

commune.lic
```diff
@@ -0,0 +1,367 @@
...
+custom_require.call(%w(common common-items common-arcana drinfomon))
+
+class Commune
+  include DRC
+  include DRCI
+  include DRCA
+
+  ######################## ITEM ACTIONS ########################
+
+  def get_cleric_item(item)
```
rpherbig: Does `get_cleric_item` need to return true or false? I don't see anywhere the return value is used.

commune.lic
```diff
@@ -0,0 +1,367 @@
...
+	end
+	empty_cleric_hands
+  end
+  
+
+  ######################## DEVOTION ########################
+	
+  def check_devotion()
+    result = bput('commune', @devotion_levels)
+    @devotion_level = @devotion_levels.index("#{result}")
```
rpherbig: You don't need to use string interpolation here (`"#{result}"`), since `result` is already a string.

commune.lic
```diff
@@ -0,0 +1,367 @@
...
+  def get_cleric_item(item)
+    case bput("get #{item} from my #{@theurgy_supply_container}", 'You get', 'You are already holding', 'I could not find', 'What were you referring to')
+	when 'You get' || 'You are already holding'
+	  return true
+	else
+	  return false
+	end
+  end
+
+  def stow_cleric_item(item)
```
rpherbig: Does `stow_cleric_item` need to return true or false? I don't see anywhere the return value is used.

If you are using the return value, I recommend naming this `stow_cleric_item?` to be clear that it is a predicate.

commune.lic
```diff
@@ -0,0 +1,367 @@
...
+    case bput("put my #{item} in my #{@theurgy_supply_container}", 'You put', 'What were you referring to')
+	when 'You put'
+	  return true
+	else
+	  return false
+	end
+  end
+
+  def empty_cleric_hands(hand='both')
```
rpherbig: Why does `empty_cleric_hands` take a parameter if it is never called with one?

commune.lic
```diff
@@ -0,0 +1,367 @@
...
+	else
+	  echo('No holy water found. Please make sure containers are open.')
+	  missing_item('holy water')
+	end
+  end
+
+  def have_cleric_item?(item)
+    case bput("look at my #{item} in my #{@theurgy_supply_container}", 'You see nothing unusual', 'I could not find', 'What were you referring to')
+    when 'You see nothing unusual'
+	  return true
```
rpherbig: Does this function need to return a boolean if the `else` is going to exit the script?

commune.lic
tulanegator: It does not need to return true or false. I changed it so if the item is missing it will exit the script.

tulanegator: I also changed stow_cleric_item to no longer return true or false. Same reasoning.

commune.lic
tulanegator: Thank you. That's fixed as well.

commune.lic
tulanegator: Fixed. it does not need true or false.

commune.lic
tulanegator: Good question. I eliminated the parameter and made empty_cleric_hands simply empty right and empty left.

commune.lic
tulanegator: It appears it does not. Thanks again. That's eliminated.

PR #1799, opened 2017-07-21
pick.lic: Make args optional

Sarvatt:
Without this fix ;pick would not run on its own unless you supplied the args.

>;pick
>
--- Lich: pick active.
[pick: ***INVALID ARGUMENTS DON'T MATCH ANY PATTERN***]
Provided Arguments: ''

  ;pick [pets] count
   pets         Disarm boxes and place them in the pet box container 
   count        How many pet boxes to make 
--- Lich: pick has exited.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -111,8 +111,8 @@ class LockPicker
   def setup
     arg_definitions = [
       [
-        { name: 'pets', regex: /pets/, description: 'Disarm boxes and place them in the pet box container' },
-        { name: 'count', regex: /\d+/, description: 'How many pet boxes to make' }
+        { name: 'pets', regex: /pets/, optional: true, description: 'Disarm boxes and place them in the pet box container' },
+        { name: 'count', regex: /\d+/, optional: true, description: 'How many pet boxes to make' }
       ]
     ]
 


rpherbig:
I _think_ this would mean you could `;pick pets` without providing a `count`. I'm not sure what the script would do in that case.

I was about to suggest we have an args option of `[]`, but then I noticed that @rcuhljr removed that in d8add39b65b4177974a0c7abb9db7916692e2b84. Why was that removed?

Sarvatt:
It made 1 pet when I did ;pick pets with this change

Sarvatt:
Could change the message to this in the worst case.

> count How many pet boxes to make (Default: 1)

Sarvatt:
he changed it because arguments didn't work anymore with the [] there

Sarvatt:
With the [] there it gave this error:

>;pick pets
--- Lich: pick active.
[pick: ***INVALID ARGUMENTS MATCH MULTIPLE PATTERNS***]

Sarvatt:
Already committed as https://github.com/rpherbig/dr-scripts/commit/d9173af9b981a8eb9dcd28518cc6408f0f0080a8

PR #1829, opened 2017-08-04
Corrected: Added Crossing Abbey, Muspar'i, additional Leth and Crossing areas!

elBlanquito:
List of areas: Gam chaga, ashu'shinvi, sand spiders, twisted dryad, dryad priestesses, germish'din, moss meys in leth, additional buccas in crossing, copperheads in crossing, resuscitant, crypt fiends, snaer hafwa, pale grey death squirrels, goblin shamans, blood vines in hib/boar clan, writing maidens, blightwater nyads, and giant blight ogres!

Forgot to add new rooms for buccas, will double tap that on next commit.

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -1354,3 +1354,175 @@ hunting_zones:
   - 11907
   - 13026
   - 13027
+
+  
+#################################################################
+# Added as of 3AUG2017 2335 EST; to be organized on next commit #
+#################################################################
+
+  ##Muspar'i ##the mobs not listed need some TLC from a mapdb guru
+  #
+  # 300-450
+  gam_chaga:
+  - 209
+  - 227
+  - 226
+  - 223
+  - 224
+  - 208
+  #
+  # 500-700 # needs a quick remap, nothing crazy.
+  ashu_shinvi:
+  - 166
+  - 162
+  #
+  # 80-100
+  sand_spider:
+  - 259
+  - 260
+  - 261
+  - 262
+  - 263
+  - 264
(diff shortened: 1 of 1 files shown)


data/base-hunting.yaml
```diff
@@ -1354,3 +1354,175 @@ hunting_zones:
...
+  # two separate hunting areas, so this is likely not correct. These are the
+  # more difficult of the two, as well. Please test and update.
+  dryad_priestess:
+  - 9768
+  - 9771
+  - 9770
+  - 9769
+  #
+  # 330-520
+  germish_din:
```
rpherbig: Is this a duplicate of `germishdin` (around line 836)?

data/base-hunting.yaml
```diff
@@ -1354,3 +1354,175 @@ hunting_zones:
...
+  - 9772
+  - 9773
+  - 9767
+  - 9774
+  - 9772
+  #
+  # 175-220-- epedia has the same values for these two, but they are in 
+  # two separate hunting areas, so this is likely not correct. These are the
+  # more difficult of the two, as well. Please test and update.
+  dryad_priestess:
```
rpherbig: Is this a duplicate of `dryads` (around line 429)?

elBlanquito: yes it is, but it breaks it up into two groups. 

Here is the rationalization: I don't want to break anybody before informing them that there is a change. There are two tiers here. E-pedia has them listed as having the same range, even though that's not the case. That's why I made a note of it. Yes.

PR #1832, opened 2017-08-05
Added support to astrology for full prediction pools

fireballx777:
Updated astrology to look for a new yaml setting to only use full pools.  If set to true, the script will still make observations, but will only predict skillsets if the prediction pool is full.  Default setting is still the old functionality (make predictions on any pool >0).

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -158,13 +158,9 @@ class Astrology
       'future events' => 'future events'
     }
 
-	if @full_pools
-		pools.reject { |_skill, size| (size < 10) }
-			.each { |skill, _size| align(skillset_to_pool[skill]) }
-	else
-		pools.reject { |_skill, size| size.zero? }
-			.each { |skill, _size| align(skillset_to_pool[skill]) }
-	end
+	target = @full_pools ? 10 : 1
+	pools.reject { |_skill, size| (size < target) }
+		.each { |skill, _size| align(skillset_to_pool[skill]) }
   end
 
   def check_heavens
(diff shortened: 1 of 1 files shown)


astrology.lic
```diff
@@ -157,8 +158,13 @@ class Astrology
       'future events' => 'future events'
     }
 
-    pools.reject { |_skill, size| size.zero? }
-         .each { |skill, _size| align(skillset_to_pool[skill]) }
+	if @full_pools
```
rpherbig: You could combine these cases with something like:
```
target = @full_pools ? 10 : 1
pools.reject { |_skill, size| size < target }
  	 .each { |skill, _size| align(skillset_to_pool[skill]) }
```

astrology.lic
```diff
@@ -215,10 +221,17 @@ class Astrology
     pause 2
     waitrt?
 
-    observed['pools']
-      .reject { |_skill, value| value.nil? }
-      .each { |skill, _value| align(skill) }
-    waitrt?
+	if @full_pools
```
rpherbig: You could also do this without an `if`

astrology.lic
```diff
@@ -215,10 +221,17 @@ class Astrology
...
+		observed['pools']
+		.reject { |_skill, value| (value < 10 || value.nil?) }
```
rpherbig: I'd prefer to see two separate `rejects` here. What do you think?

rpherbig:
Looks good to me. @jandersson or @kamarcum or @rcuhljr ?

jandersson:
Makes sense to me

rpherbig:
Not needed in this PR, but a style note: we could avoid duplicating the `target = @full_pools ? 10 : 1` logic by storing it into something like `@prediction_pool_target` (instead of storing `@full_pools`).

rpherbig:
Thank you for the PR @fireballx777 !

Issue #1777, opened 2017-07-10
astrology waggle bug - doesn't use foci for rtr


Missah:
Here is my astrology waggle setup (maybe I have set something wrong)

```
  astrology:
    Piercing Gaze:              
      abbrev: pg            
      mana: 10             
      cambrinth:            
      - 10
    Aura Sight:     
      abbrev: aus           
      mana: 10            
      cambrinth:            
      - 10
    Destiny Cipher:      
      abbrev: dc           
      mana: 430             
      focus: abacus        
      worn_focus: false
    Read The Ripples:      
      abbrev: rtr          
      mana: 420            
      focus: abacus        
      worn_focus: false
```

The script will use my abacus for casting dc...but will NOT use it for casting rtr:

```
[buff]>prepare dc 430
You feel intense strain as you try to manipulate the mana streams to form this pattern, and you are not certain that you will have enough mental stamina to complete it.
You raise your arms skyward, chanting the equation of the Destiny Cipher spell.
>
[buff]>get my abacus
You get a moonstone-beaded watered steel abacus etched with complex sigils from inside your leather pack.
>
[buff]>invoke my abacus 
You lift your abacus toward the sky and will the mana streams to bend into it.
Roundtime: 20 sec.
>
Your abacus steadily brightens, until it burns like a glowing white torch!
You direct the flow of power you've conjured up into your spell pattern, taking much of the strain of empowering it off you.
>
Your abacus fades over the course of a few seconds, returning to normal.
>
[buff]>stow my abacus
You put your abacus in your leather pack.
>
You feel fully prepared to cast your spell.
>
[buff]>cast
You close your eyes and take several slow, calming breaths.
The mental strain of this pattern is considerably eased by your ritual focus.
Your mind is cleared of distraction, the present and the mundane.  You are adrift in an unbroken sea of silent darkness.

A chorus of voices fills your mind, a cacophony of languages both known and unfamiliar.  They overlap, merge, submerge, muffle and distort.  You hone in on one voice at a time, memorizing its unique cadence before shifting your focus to the next voice.  As you come to know their message, their words, their intent, you realize the throng of voices is really one voice, one language, one truth.

The darkness recedes and you find your thoughts unified, your mind honed to a prophetic purpose.
>
[buff]>prepare rtr 420
You feel intense strain as you try to manipulate the mana streams to form this pattern, and you are not certain that you will have enough mental stamina to complete it.
[...]

PR #1834, opened 2017-08-06
Astrology: Handle RTR issues & add astrology_training setting

jandersson:
* Make astrology use common-arcana instead of directly calling `buff.lic`.
* Fix RTR related issues by moving flag check into `check_ripples` as a guard clause
* Store RTR data and remove it from the buff list. This lets the user specify the spell in any order in their buff spells list and frees up `train_astrology` to train in any order.
* Add an `astrology_training` setting/list which lets the user decide what to train and in what order. Possible settings are: 
```yaml
astrology_training:
- ways
- observe  # Default
- rtr      # Default
- events
- weather # Default
- moons   # Default
```
* Warn the user that `astrology_buffs` are old and will disappear soon


astrology.lic
```diff
@@ -28,27 +28,37 @@ class Astrology
...
-        buff(spell, settings)
-      end
+      buffs = settings.waggle_sets['astrology']
+      @rtr_data = buffs.select { |spell| spell.eql?('Read the Ripples') }
+                       .values
+                       .first
+      echo @rtr_data
+      buffs.reject! { |spell| spell.eql?('Read the Ripples') }
+      cast_spells(buffs, settings)
+    elsif settings.astrology_buffs
```
rpherbig: Is it about time we stop handling this style and output an error?

astrology.lic
```diff
@@ -276,13 +289,14 @@ class Astrology
     fix_standing
   end
 
-  def train_astrology
-    check_ripples
+  def train_astrology(settings)
+    return if DRSkill.getxp('Astrology') > 30
```
rpherbig: Should this check only be here, before we do any training, or checked after each step of training?

astrology.lic
```diff
@@ -28,27 +28,37 @@ class Astrology
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
     @predict_event = settings.predict_event
-	@full_pools = settings.astrology_use_full_pools
+    @full_pools = settings.astrology_use_full_pools
```
rpherbig: we could avoid duplicating the `target = @full_pools ? 10 : 1` logic by storing the calculated target value into something like `@prediction_pool_target` (and then not storing `@full_pools`).

jandersson: I can update that

astrology.lic
jandersson: Yea that would be nice to check after each step. Ill add that

astrology.lic
jandersson: I dont know of a good way to remove this without breaking anyones setup using `astrology_buffs`

jandersson: One solution is to echo a warning, saying its outdated and that it will be removed in a future update.

rpherbig: It should at least echo something to the user. This has been in here a while though... @rcuhljr thoughts?

rcuhljr: I'm declaring one month as our cutoff. Give warnings up until then but once someones been seeing it for at least 2-3 weeks no mercy shall be given.

astrology.lic
jandersson: Going to add in an `astrology_training` list similar to what was done with `appraisal_training`

astrology.lic
jandersson: Added a warning

jandersson:
Updated OP to reflect added training settings

profiles/base.yaml
```diff
@@ -561,10 +561,15 @@ divination_bones_storage:
   container:
   tied:
 astral_plane_training:
-  train_in_ap: false
+  train_in_ap: false  # Deprecated
   train_destination:
   train_source:
-predict_event: false
+predict_event: false  # Deprecated
+astrology_training:
```
rpherbig: Do you want to list here (but comment out) the non-default ones?

jandersson: That sounds reasonable to me.

jandersson: Reminder to add this to `base-empty`

astrology.lic
```diff
@@ -276,13 +286,27 @@ class Astrology
...
+      when 'weather'
+        check_weather
+      when 'events'
+        check_events(check_pools)
+      when 'moons'
+        check_moons
+      end
+    end
     predict_all(check_pools)
     fput('predict analyze')
```
jandersson: Todo: change to bput and add waitrt? after

rpherbig:
I ran Rubocop and it caused a conflict. I merged in a fix, but it looks like some tab characters still exist on lines 231-235. Can you fix those with your next commit?

jandersson:
Will do

jandersson:
Updated

profiles/base.yaml
jandersson: Done & Done

astrology.lic
```diff
@@ -24,31 +24,40 @@ class Astrology
...
+                       .first
+      buffs.reject! { |spell| spell.eql?('Read the Ripples') }
+      cast_spells(buffs, settings)
+    elsif settings.astrology_buffs
+      echo "******************"
+      echo "astrology_buffs has been deprecated in favor of 'astrology' waggle set and will be removed in a future version"
+      echo "******************"
+      buffs = settings.astrology_buffs['spells']
+      @rtr_data = buffs.select { |spell| spell['abbrev'].casecmp('rtr').zero? }.first
+      buffs.reject! { |spell| spell['abbrev'].casecmp('rtr').zero? }
```
rpherbig: You can use `reject` here (instead of `reject!)

jandersson: I would have to do reassignment then right? I assumed `!` let me do it in-place.

astrology.lic
```diff
@@ -276,15 +285,30 @@ class Astrology
...
-    check_ripples
-    check_astral
-    check_weather
-    check_events(check_pools) if @predict_event
-    check_moons
-    check_heavens
+  def train_astrology(settings)
+    return unless settings
+    return unless settings.astrology_training
+
```
rpherbig: Since this is a new setting, do you want to explicitly check for the case in which `settings.astrology_training` is `[]` and output a message (e.g. 'Go set settings.astrology_training - see base.yaml for an example')?

rpherbig: Oh, will the transform in dependency make this a non-issue? Or should we output a deprecation warning in that case?

jandersson: That makes sense, though it would be temporary code right? This is for the case that they update but have not reloaded dependency

jandersson: Ill leave a comment saying to remove it in a future update, if so

astrology.lic
rpherbig: The `.each` on the next line is chained, so you don't need to do any assignment

jandersson: Each modifies the array in-place? TIL. Thanks!

astrology.lic
jandersson: I think it makes more sense to check if its `nil?` and not `[]`. This means they have updated astrology via dependency but not reloaded base.yaml

astrology.lic
```diff
@@ -286,12 +286,17 @@ class Astrology
   end
 
   def train_astrology(settings)
+    return unless DRStats.moon_mage?
     return unless settings
-    return unless settings.astrology_training
+    return unless settings.astrology_training.none?
```
rpherbig: This will throw if `settings.astrology_training` is `nil` (`undefined method `none?' for nil:NilClass (NoMethodError)`).

jandersson: Does base-empty turn this into a list/array if it is blanked out or nil?

jandersson: I also cant seem to get that error to throw using ;exec 
```
IHR>;en 1.times{settings = get_settings; echo settings.asdfasdfa.empty?} 
--- Lich: exec1 active.                                                  
[exec1: ]                                                                
--- Lich: exec1 has exited.                                              
```

jandersson: I think i can move this line to go after the `nil?` check

jandersson:
```ruby
  def train_astrology(settings)$                                                              
    return unless settings$                                                                   
    if settings.astrology_training.nil? # TODO: Temporary. Delete when everyone is updated$   
      echo " Astrology training settings have been updated, please restart dependency!"$      
      exit$                                                                                   
    end$                                                                                      
    return unless settings.astrology_training.kind_of?(Array)$                                
    return if settings.astrology_training.none?$                                              
$                                                                                             
```

rpherbig:
Will this fix #1777?

rpherbig:
And do you have any more changes to make?

jandersson:
updated

jandersson:
No more changes to make, I was going to add in a fix for being too injured to use a telescope, but #1849 is covering that 👍 

#1777 should actually be fixed in PR #1770

jandersson:
I take back what I said about  #1777 being fixed in #1770. I think theyre unrelated, the issue was a typo/case error.

PR #1844, opened 2017-08-15
Added carving to validate (+2 bucca rooms)

elBlanquito:
Added carving to validate: Fixed a bug where carving was failing a validate check in workorders.lic at lines 56-58, causing it to default to shaping instead.

Crossing_Training will still default to shaping, but atleast manually calling stone from a before: , works.

I will continue investigating what else is needed to ensure calling Engineering from Crossing_Training will support carving stone.

### DIFF
--- a/data/base-mining.yaml
+++ b/data/base-mining.yaml
@@ -62,28 +62,8 @@ mining_buddy_rooms:
   - 7237
   - 7238
   # Therengia
-  # Populated by wolf spiders
-  breech_tunnels:
-  - 7009
-  - 7010
-  - 7011
-  - 7014
-  - 7015
-  - 7016
-  - 7013
-  - 7017
-  - 7018
-  - 7019
-  - 7020
-  - 7021
-  - 7040
-  - 7039
-  - 7038
-  - 7037
-  - 7035
-  - 7036
-  - 10008
-  - 10009
+  # Removed Breeach Tunnels from Theren Keep because
+  # inability to retreat from wolf spiders breaks mining_buddy
   dark_tunnels:
   - 8560
   - 8563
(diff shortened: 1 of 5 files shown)


data/base-mining.yaml
```diff
@@ -62,28 +62,8 @@ mining_buddy_rooms:
...
-  - 7021
-  - 7040
-  - 7039
-  - 7038
-  - 7037
-  - 7035
-  - 7036
-  - 10008
-  - 10009
+  # Removed Breeach Tunnels from Theren Keep because
```
rpherbig: "Breech"?

validate.lic
```diff
@@ -255,7 +255,7 @@ class DRYamlValidator
     return unless settings.train_workorders
 
     settings.train_workorders
-            .reject { |discipline| %w[Blacksmithing Tailoring Shaping Remedies].include?(discipline) }
+            .reject { |discipline| %w[Blacksmithing Tailoring Shaping Carving Remedies].include?(discipline) }
```
rpherbig: Glancing at the code, it doesn't look like Carving is currently supported with `train_workorders` (though it _is_ supported by `disciplines_to_skill` for Traders).

rpherbig:
In the future, you might want to consider making unrelated changes into separate PRs. That makes it easier for us to merge changes with no comments; and keep discussion focused on smaller related changes.

elBlanquito:
Should I make different branches for each block of changes and then push from each to keep them separate? I was making these commits one at a time and then putting in a pull-request via the main page of my fork on the website each time after each chunk. What is the ideal way to handle this?

I understand why removing the error checking would be a bad thing until carving is fully supported by work orders--working on this right now (because there are ample examples from the other crafting sub-disciplines to use for building this in an expeditious manner).

elBlanquito:
Dartellum squared me away, thank you!

PR #1846, opened 2017-08-15
Updated sew to include support for lightening

fireballx777:


sew.lic
```diff
@@ -242,9 +261,15 @@ class Sew
         next_command = "cut my #{@noun} with my scissors"
       end
     when 'Roundtime'
-      tool = 'sewing needles'
+      case lighten
```
rpherbig: This might read a little clearer as:
```
      if lighten
        tool = 'scissors'
        next_command = "cut my #{@noun} with my scissors"
      else
        tool = 'sewing needles'
      end
```

sew.lic
```diff
@@ -132,6 +137,17 @@ class Sew
     stow_item('book')
     do_seal(true)
   end
+  
+  def lighten
+    check_thread
```
fireballx777: One thing I can think of is creating a new method: 

```ruby
def study_tailoring(chapter, recipe)
  get_item('tailoring book')
  bput("turn my book to page #{find_recipe(chapter, recipe)}", 'You turn your', 'The book is   already')
  bput('study my book', 'Roundtime')
  stow_item('book')
end
```

This could replace chunks in `leather`, `sew`, `knit`, `seal`, `lighten`, and `reinforce`.  It would also eliminate the need for `turn_to` be integrating it directly (it's not called anywhere except the above).  I think the rest of the logic is different enough between sections that trying to combine it further would lead to messy `if` blocks.

rpherbig: That sounds pretty good to me

PR #1852, opened 2017-08-20
Updating Crossing-training for multiple cambrinth

fireballx777:
I added logic to crossing-training to allow using multiple cambrinth items when training magic skills.  The requires new yaml settings (example copied below) to identify an array of cambrinth items.  The logic right now is to identify how much mana to put into cambrinth,  iterate over the cambrinth items to determine how much each can hold, and charge/invoke them sequentially (2 charges per item).  I'd like to work on other scripts to utilize multiple cambrinth (notably combat-trainer and buff), but I first wanted to get other eyes on what I did here to make sure I'm not doing something way off-base.

I also kept the logic in there to support the previous methodology so as not to break the script for people who don't add the new settings.

Potential future updates: 
- As it is now, it will iterate through cambrinth items in the order listed.  It would make sense to sort them biggest to smallest, but since they can be manually listed in that order, this wasn't high priority.
- The script was set to use 1/4 of the mana as prep, and put the remainder into cambrinth (if it fit).  I changed that to 1/5, since I figured keeping less as prep makes sense if there's more room available for cambrinth.  It might make sense to have this as a user-defined option, similar to prep_scaling_factor. 

Sample yaml settings:
```cambrinth_items:
- :name: cambrinth armband
  :cap: 50
  :stored: false
- :name: cambrinth cuff
  :cap: 28
  :stored: false

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -382,15 +382,36 @@ class CrossingTraining
   def calculate_mana(min, more, discern_data, cyclic)
     total = min + more
     total = (total * @settings.prep_scaling_factor).floor
-    discern_data['mana'] = [(total / 4.0).ceil, min].max
+    discern_data['mana'] = [(total / 5.0).ceil, min].max
     remaining = total - discern_data['mana']
-    if remaining > @settings.cambrinth_cap
+    total_cambrinth_mana = 0
+    @settings.cambrinth_items.each do |item|
+      total_cambrinth_mana += item[:cap]
+    end
+    if remaining > total_cambrinth_mana && !!@settings.cambrinth_items
+      discern_data['mana'] = discern_data['mana'] + (remaining - @settings.cambrinth_cap)
+      remaining = total - discern_data['mana']
+    elsif remaining > @settings.cambrinth_cap && !@settings.cambrinth_items
       discern_data['mana'] = discern_data['mana'] + (remaining - @settings.cambrinth_cap)
       remaining = total - discern_data['mana']
     end
     if cyclic
       discern_data['cambrinth'] = nil
       discern_data['mana'] = discern_data['mana'] + remaining
+    elsif remaining > 0 && !!@settings.cambrinth_items
+      discern_data['cambrinth'] = []
+      @settings.cambrinth_items.each do |item|
+        step_size = (item[:cap]/2).floor
+        if remaining > step_size
+          discern_data['cambrinth'] << [remaining, step_size].min
+          remaining -= step_size
+          discern_data['cambrinth'] << [remaining, step_size].min
+          remaining -= step_size
+        elsif remaining > 0
+          discern_data['cambrinth'] << [remaining, step_size].min
+          remaining -= step_size
+        end
+      end
     elsif remaining > 0
       discern_data['cambrinth'] = []
       step_size = (remaining / 3.0).ceil
@@ -527,12 +548,23 @@ class CrossingTraining
       data['cast'] = "cast #{moon}"
     end
 
-    release_cyclics if data['cyclic']
+    release_cyclics if data['cyclic']    
 
     return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'])
 
-    find_charge_invoke_stow(@settings.cambrinth, @settings.stored_cambrinth, @settings.cambrinth_cap, @settings.dedicated_camb_use, data['cambrinth'])
-
+    if !!@settings.cambrinth_items
+      charges_array = []
+      data['cambrinth'].each_with_index do |mana, index|
+        charges_array[(index/2.0).floor] ||= []
+        charges_array[(index/2.0).floor] << mana
+        end
+      charges_array.each_with_index do |charges, index|
+        current_item = @settings.cambrinth_items[index]
+        find_charge_invoke_stow(current_item[:name], current_item[:stored], current_item[:cap], @settings.dedicated_camb_use, charges)
+        end
+    else
+      find_charge_invoke_stow(@settings.cambrinth, @settings.stored_cambrinth, @settings.cambrinth_cap, @settings.dedicated_camb_use, data['cambrinth'])
+    end
     waitcastrt?
 
     snapshot = DRSkill.getxp(skill) if data['symbiosis']
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -384,49 +384,47 @@ class CrossingTraining
     total = (total * @settings.prep_scaling_factor).floor
     discern_data['mana'] = [(total / 5.0).ceil, min].max
     remaining = total - discern_data['mana']
-    total_cambrinth_mana = 0
-    @settings.cambrinth_items.each do |item|
-      total_cambrinth_mana += item['cap']
+    if @settings.cambrinth_items
+      cambrinth_items = @settings.cambrinth_items
+    else
+      cambrinth_items = []
+      cambrinth_items[0] = {}
+      cambrinth_items[0]['name'], cambrinth_items[0]['cap'], cambrinth_items[0]['stored'] = map_cambrinth(@settings)
     end
-    if remaining > total_cambrinth_mana && @settings.cambrinth_items
-      discern_data['mana'] = discern_data['mana'] + (remaining - @settings.cambrinth_cap)
-      remaining = total - discern_data['mana']
-    elsif remaining > @settings.cambrinth_cap && !@settings.cambrinth_items
-      discern_data['mana'] = discern_data['mana'] + (remaining - @settings.cambrinth_cap)
+    total_cambrinth_cap = 0
+    cambrinth_items.each do |item|
+      total_cambrinth_cap += item['cap']
+    end
+    total_cambrinth_charges = 0
+    charges_count_floor = 4
+    cambrinth_items.each do |item|
+      item['charges'] = ((item['cap'].to_f / total_cambrinth_cap) * charges_count_floor).ceil
+      total_cambrinth_charges += item['charges']
+    end
+    if remaining > total_cambrinth_cap
+      discern_data['mana'] = discern_data['mana'] + (remaining - total_cambrinth_cap)
       remaining = total - discern_data['mana']
     end
     if cyclic
       discern_data['cambrinth'] = nil
       discern_data['mana'] = discern_data['mana'] + remaining
-    elsif remaining > 0 && @settings.cambrinth_items
-      discern_data['cambrinth'] = []
-      @settings.cambrinth_items.each do |item|
-        step_size = (item['cap']/2).floor
-        if remaining > item['cap']
-          2.times do
-            discern_data['cambrinth'] << step_size
-            remaining -= step_size
-          end
-        elsif remaining > step_size
-          2.times do
-            discern_data['cambrinth'] << (remaining/2).floor
-          end
-          remaining = 0
-        elsif remaining > 0
-          discern_data['cambrinth'] << remaining
-          remaining = 0
-        end
-      end
     elsif remaining > 0
-      discern_data['cambrinth'] = []
-      step_size = (remaining / 3.0).ceil
-      while remaining > 0
-        discern_data['cambrinth'] << [remaining, step_size].min
-        remaining -= step_size
+      total_cambrinth_mana = [remaining, total_cambrinth_cap].min
+      cambrinth_items.each_with_index do |item, index|
+        discern_data['cambrinth'] ||= []
+        charge_amount = (total_cambrinth_mana / total_cambrinth_charges) * item['charges']
+        discern_data['cambrinth'][index] = []
+        charge_amount.times do |i|
+          discern_data['cambrinth'][index][i%item['charges']] += 1
+        end
       end
     else
       discern_data['cambrinth'] = nil
     end
+  end
+
+  def map_cambrinth(settings)
+      return settings.cambrinth, settings.cambrinth_cap, settings.stored_cambrinth
   end
 
   def check_discern(data)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rpherbig** [comment]
A quick note on formatting, you can use three backticks (`) to do multi-line code snippets like this:

```
cambrinth_items:
- :name: cambrinth armband
  :cap: 50
  :stored: false
- :name: cambrinth cuff
  :cap: 28
  :stored: false
```

File: crossing-training.lic [hunk 1]
```diff
@@ -382,15 +382,36 @@ class CrossingTraining
   def calculate_mana(min, more, discern_data, cyclic)
     total = min + more
     total = (total * @settings.prep_scaling_factor).floor
-    discern_data['mana'] = [(total / 4.0).ceil, min].max
+    discern_data['mana'] = [(total / 5.0).ceil, min].max
     remaining = total - discern_data['mana']
-    if remaining > @settings.cambrinth_cap
+    total_cambrinth_mana = 0
+    @settings.cambrinth_items.each do |item|
+      total_cambrinth_mana += item[:cap]
```
rpherbig: I prefer the 'string' style (`item['cap']`) over symbol style (`item[:cap]`) myself. I think it makes the YAML file a bit easier to read (and is definitely more common).

crossing-training.lic
```diff
@@ -382,15 +382,36 @@ class CrossingTraining
...
+    end
+    if remaining > total_cambrinth_mana && !!@settings.cambrinth_items
```
rpherbig: How does `if ... && !!@settings.cambrinth_items` differ from `if ... && @settings.cambrinth_items`?

crossing-training.lic
```diff
@@ -382,15 +382,36 @@ class CrossingTraining
...
     end
     if cyclic
       discern_data['cambrinth'] = nil
       discern_data['mana'] = discern_data['mana'] + remaining
+    elsif remaining > 0 && !!@settings.cambrinth_items
+      discern_data['cambrinth'] = []
+      @settings.cambrinth_items.each do |item|
+        step_size = (item[:cap]/2).floor
+        if remaining > step_size
+          discern_data['cambrinth'] << [remaining, step_size].min
```
rpherbig: If you wanted to reduce code duplication you could use a loop: `2.times do { ... }`

rpherbig:
I'm finding all of the mana calc code pretty hard to grok. It's not your fault, this is complicated stuff. Do you think there are some refactors we could do to make this code easier to follow?

crossing-training.lic
fireballx777: Works for me.  Updated.

crossing-training.lic
fireballx777: I guess it doesn't.  For some reason I thought if I was checking for true/false, it needed to explicitly be a boolean, which is what I was doing with `!!`.  Not needed, though -- updated to remove.

crossing-training.lic
fireballx777: Thanks, cleaned that up a bit.

fireballx777:
Agree on the mana calc stuff being complicated.  I tried to update it to three distinct `if` cases (all within the  `if` block checking that `@settings.cambrinth_items` is true): if the remaining mana is greater than the cambrinth cap (it will charge at half the cap twice and move on to the next cambrinth), if the remaining mana is greater than half the cambrinth cap (it will charge half of the remaining twice), or if the remaining mana is greater than 0 (it will charge the remaining in one shot).

I'm willing to try to clean it up some more if you think it's needed, and I'm open to suggestions.  I could add some comments on what's going on, but I've been avoiding that as I haven't seen commented code in other sections of the repository.

rpherbig:
@rcuhljr @kamarcum @jandersson any comments?

rcuhljr:
I think we can simplify this a bit.

Firstly, we can automatically map `cambrinth:` and `cambrinth_cap:` into `cambrinth_items:` if the user doesn't have an entry for it, and then run all logic against the list, one case of which is only having one item defined.  Then we set up the algorithm as such...

Take the lower of your mana past initial discern or summed cambrinth cap.

Pop the first item off the list

Divide item cap by target mana times 4 unless cap is greater than remaining target, round up charge it evenly over that many charges checking that you haven't put enough in.

E.g. 25 mana belt, 48 mana arm.  5 prep 95 extra spell 73 is total cambrinth so prep at 27

    25/73*4 = 2, I have min(95,25) for this item, divide by 2, [13, 12]
    48/73*4 = 3, Min(70,48) for this item, divide over 3, [16,16,16]

E.g. 25 mana belt, 48 mana arm.  5 prep 5 extra spell 73 is total cambrinth so prep at 5

    25 > 5 =  split over 5 charges [1,1,1,1,1]

This snippet is useful for splitting up a value evenly over N charges.

```
irb(main):027:0> charges = Array.new(5,0)
=> [0, 0, 0, 0, 0]
irb(main):028:0> 14.times{|x|charges[x%5]+=1}
=> 14
irb(main):029:0> charges
=> [3, 3, 3, 3, 2]
```

fireballx777:
Good ideas, @rcuhljr; I'm working on re-writing this implementing that logic.

Where you divide item cap by target mana times 4, I'm calling the 4 `charges_floor`.  I checked a few use cases to see if 4 would be the optimal number here:


Cambrinth_items | Charges Floor = 3 | Charges Floor = 4 | Charges Floor = 5
-- | -- | -- | --
[100] | [34, 33, 33] | [25, 25, 25, 25] | [20, 20, 20, 20, 20]
[50, 32] | [25, 25], [16, 16] | [17, 17, 16], [16, 16] | [13, 13, 12, 12], [16, 16]
[32, 32] | [16, 16], [16, 16] | [16, 16], [16, 16] | [11, 11, 10], [11, 11, 10]
[100, 50] | [50, 50], [50] | [34, 33, 33], [25, 25] | [25, 25, 25, 25], [25, 25]
[50, 28] | [25, 25], [14, 14] | [17, 17, 16], [14, 14] | [13, 13, 12, 12], [14, 14]
[50, 28, 28] | [25, 25], [28], [28] | [25, 25], [14, 14], [14, 14] | [17, 17, 16], [14, 14], [14, 14]
[32, 28, 4] | [16, 16], [14, 14], [4] | [16, 16], [14, 14], [4] | [11, 11, 10], [10, 9, 9], [4]
[50, 28, 4] | [25, 25], [14, 14], [4] | [17, 17, 16], [14, 14], [4] | [13, 13, 12, 12], [14, 14], [4]

It does seem 4 is the best option, except possibly in edge cases where it will cause more charges than ideal.  I just wanted to toss that out there in case anyone thinks another number here would be preferable.

I am also considering calling a setting in `cambrinth_items` as a replacement for `dedicated_camb_use`, but I'm not sure if this makes sense.  I don't have the dedicated feat, so I can't speak to the typical use cases -- would people want the option to designate specific cambrinth items as dedicated, or does it make more sense to keep it as a general setting?  An advantage to defining it per item would be that I could modify this script to only iterate through cambrinth items where it's `false` or `spell`, and ignore items where it's `cyclic`.

rcuhljr:
I agree that it might be worth looking for ded camb usage, but I'd save that for later. Lets get this one working for now and we can revisit it down the road.

crossing-training.lic
```diff
@@ -382,27 +382,51 @@ class CrossingTraining
...
+    discern_data['mana'] = [(total / 5.0).ceil, min].max
     remaining = total - discern_data['mana']
-    if remaining > @settings.cambrinth_cap
-      discern_data['mana'] = discern_data['mana'] + (remaining - @settings.cambrinth_cap)
+    if @settings.cambrinth_items
+      cambrinth_items = @settings.cambrinth_items
+    else
+      cambrinth_items = []
+      cambrinth_items[0] = {}
+      cambrinth_items[0]['name'], cambrinth_items[0]['cap'], cambrinth_items[0]['stored'] = map_cambrinth(@settings)
```
rpherbig: Did you consider using this syntax for making a new Hash?

```
cambrinth_items[0] = {
  'name' => @settings.cambrinth,
  'cap' => @settings.cambrinth_cap
  'stored' => @settings.stored_cambrinth
}
```

profiles/base.yaml
```diff
@@ -160,6 +160,10 @@ cambrinth_cap: 32
 stored_cambrinth: false
 prep_scaling_factor: 0.98
 kneel_khri: false
+cambrinth_items:
+- name:
```
rpherbig: Is this an empty cambrinth declaration? Will having this in `base.yaml` mean the `cambrinth_items` logic will be used immediately for everyone (since the array now has an item in it)?

crossing-training.lic
fireballx777: Since I was assigning it in a separate section also, I figured it made sense to define a new method to map it, though I don't know how much more efficient it is, since it's only a few lines and I had to declare the array and hash before calling `map_cambrinth`, anyway.  I'm fine switching it back if you think it makes more sense.

profiles/base.yaml
fireballx777: I was worried about that, but I tested it out, and the `if @settings.cambrinth_items` seemed to be returning false even with the array in the base.yaml, presumably since the hash was undefined.

rpherbig: I ran this:
```
if get_settings.cambrinth_items
  echo 'true'
else
  echo 'false'
end
```
and got 'true'. That implies to me (though I haven't tested it) that the old style cambrinth data wouldn't be copied over and used.

fireballx777: I just checked, and you're right -- I'm not sure why it was behaving differently for me before.  I updated to check for `@settings.cambrinth_items[0]['name']` specifically, which still returns false with the empty cambrinth declaration I added to base.yaml.

crossing-training.lic
rpherbig: Personally, I prefer the style I suggested, but it's not a big deal.

rpherbig:
@rcuhljr looks good to me - any more comments from you before merging this?

crossing-training.lic
rcuhljr: You can do it all on one line really, 
```
cambrinth_items = [{
  'name' => @settings.cambrinth,
  'cap' => @settings.cambrinth_cap
  'stored' => @settings.stored_cambrinth
}]
```

crossing-training.lic
```diff
@@ -382,27 +382,51 @@ class CrossingTraining
...
+    end
+    total_cambrinth_cap = 0
+    cambrinth_items.each do |item|
+      total_cambrinth_cap += item['cap']
+    end
```
rcuhljr: You can just assign directly to total cap with `[{'cap' => 5},{'cap'=>14}].map{|x| x['cap']}.inject(&:+)`

crossing-training.lic
```diff
@@ -382,27 +382,51 @@ class CrossingTraining
...
-      while remaining > 0
-        discern_data['cambrinth'] << [remaining, step_size].min
-        remaining -= step_size
+      total_cambrinth_mana = [remaining, total_cambrinth_cap].min
+      cambrinth_items.each_with_index do |item, index|
+        discern_data['cambrinth'] ||= []
+        charge_amount = (total_cambrinth_mana / total_cambrinth_charges) * item['charges']
+        discern_data['cambrinth'][index] = []
+        charge_amount.times do |i|
+          discern_data['cambrinth'][index][i%item['charges']] += 1
```
rcuhljr: Has this been tested? Unless lich has messed with nil again  this seems like it will fail because you're trying to add 1 to nil (an undefined array entry).

fireballx777: It worked when testing before.  I just tried it again with some echos added before and after assigning values to the array to make sure it was assigning as intended:

```
        discern_data['cambrinth'][index] = []
        echo "discern_data['cambrinth'][index] for #{item['name']} is equal to #{discern_data['cambrinth'][index]}"
        charge_amount.times do |i|
          discern_data['cambrinth'][index][i%item['charges']] += 1
        end
        echo "discern_data['cambrinth'][index] for #{item['name']} is equal to #{discern_data['cambrinth'][index]}"
```

And I got this when running it in game:
```
[crossing-training: discern_data['cambrinth'][index] for cambrinth armband is equal to []]
[crossing-training: discern_data['cambrinth'][index] for cambrinth armband is equal to [9, 9, 9]]
[crossing-training: discern_data['cambrinth'][index] for cambrinth hairnet is equal to []]
[crossing-training: discern_data['cambrinth'][index] for cambrinth hairnet is equal to [9]]
[crossing-training: discern_data['cambrinth'][index] for cambrinth moonflower is equal to []]
[crossing-training: discern_data['cambrinth'][index] for cambrinth moonflower is equal to [9, 9]]
```

crossing-training.lic
fireballx777: Wasn't aware of this bit of code, thanks.  Updated.

crossing-training.lic
```diff
@@ -527,12 +547,22 @@ class CrossingTraining
...
+      @settings.cambrinth_items = [{
+        'name' => @settings.cambrinth,
+        'cap' => @settings.cambrinth_cap,
+        'stored' => @settings.stored_cambrinth
+      }]
+    end
+    @settings.cambrinth_items.each_with_index do |item, index|
+      if data['cambrinth'][index][0]
+        find_charge_invoke_stow(item['name'], item['stored'], item['cap'], @settings.dedicated_camb_use, data['cambrinth'][index])
+      end
```
fireballx777: I added an `if` here to check whether there's any mana assigned to `data['cambrinth'][index]` before calling `find_charge_invoke_stow` on it.  Without this, I noticed that it was invoking all cambrinth even if there was no mana being used with it (wasn't breaking the scripts, just causing unnecessary roundtime).

crossing-training.lic
rcuhljr: Yeah it's just lich being dumb and redefining nil + value to be value instead of an error.

fireballx777: Would it make more sense to add
`discern_data['cambrinth'][index][i%item['charges']] ||= 0`
before
`discern_data['cambrinth'][index][i%item['charges']] += 1`
?
That way it will explicitly assign it a value of 0 if it's nil.

rpherbig:
@rcuhljr looks good to me - any more comments from you before merging this?

rcuhljr:
Nah, it looks fine. Not sure if there's any validate guards we want to add around cambrinth settings.

PR #1860, opened 2017-08-23
bescort.lic: clear hands before trying to wear ice skates.

Sarvatt:
If you try to go2 through here with an item in hand you will fail to wear skates and fall in every room after.


Sarvatt:
steps to reproduce:
;go2 4583
make sure you have an item in your hand and you have skates somewhere.
;go2 4365

PR #1861, opened 2017-08-24
Update Attunement

jandersson:
Summary of changes:
* Remove necromancer check
* Add custom `attunement_rooms` setting which lets the user specify rooms that they want to power walk in. This is in part to mitigate the removal of the necromancer check, attunement training can be done in non-justice areas (or anywhere) by filling in the setting.
* Cast 'attunement' spell set before training


attunement.lic
```diff
@@ -22,10 +23,13 @@ class Attunement
...
       waitrt?
       return
-    end
-
-    room_list = if @stationary_skills_only || DRStats.necromancer?
+    end 
+                
+    room_list = if @stationary_skills_only
                   [Room.current.id]
+                elsif !get_settings.attunement_rooms.nil? && !get_settings.attunement_rooms.empty?
```
rpherbig: What do you think about assigning `@attunement_rooms` in `initialize`?

attunement.lic
```diff
@@ -22,10 +23,13 @@ class Attunement
...
+                  # TODO: Nil check isn't necessary after everyone has restarted dependency
+                  get_settings.attunement_rooms
```
rpherbig: Is `attunement_rooms` just a way to override what's in `base-town`?

attunement.lic
```diff
@@ -22,10 +23,13 @@ class Attunement
       pause 15
```
rpherbig: Should MMs get special treatment here, or adhere to the room list below and use the guild to determine `perc` vs `perc mana`?

profiles/Jonas-setup.yaml
```diff
@@ -1,31 +1,41 @@
 ---
-safe_room: 4377 # bnb
-# Hunting Settings
-
+attunement_rooms: #Non-justice attunement rooms near my safe room
```
rpherbig: Is there any risk that this could give away your safe room? Do attunement rooms need to also support being pulled from a UserVar to keep them out of your YAML?

attunement.lic
jandersson: Yea

attunement.lic
jandersson: Sounds good

profiles/Jonas-setup.yaml
jandersson: I'm not concerned about it, but I will remove the comment that its near my safe room

attunement.lic
jandersson: I updated the OP to show the motivation behind it

attunement.lic
jandersson: Thats a great idea, I hadn't noticed it before

jandersson:
Updated based on feedback

jandersson:
Covers #1717 via the buff set / waggle set

jandersson:
Updated documentation: https://elanthipedia.play.net/Lich_script_repository#attunement

PR #1864, opened 2017-08-26
Added 'engage' before collecting and setting to set combat_training_abilities

adamgoerdt:


combat-trainer.lic
```diff
@@ -1336,6 +1336,8 @@ class TrainerProcess
 
     @analyze_retry_count = settings.combat_analyze_retry_count
     echo("  @analyze_retry_count: #{@analyze_retry_count}") if $debug_mode_ct
+
+    @combat_training_abilities_target = settings.combat_training_abilities_target
```
rpherbig: Can you add an echo for this setting (similar to line 1338)?

adamgoerdt: All set

rpherbig:
@adamgoerdt Can you make sure to update elanthipedia with this new setting?

PR #1865, opened 2017-08-26
Zombie Support

jandersson:
This PR adds support for zombie creation and usage in combat. 

Settings for a zombie look like this:
```yaml
zombie:               
  make: false  # Create a zombie during LootProcess if you do not already have one. You have a zombie if the spell Call from Beyond is active
  summon: true  # Command your zombie to come to your current room (if CFB is active)
  behavior: defensive # tells the zombie what mobs to attack
  stance: balanced # sets the offensive/defensive stance of the zombie
waggle_sets:
  zombie:       # Store zombie related spell data here... we can add the zombie buff spell RPU and healing spell NR in a future update
    Call from Beyond:   
      mana: 30         
      cambrinth:       
      - 32     
```

To make a zombie, a mob is killed then the 'arise' ritual is performed on it. Once that is completed, the spell 'Call from Beyond' is cast. The zombie can be dismissed into the ether with the 'leave' command and recalled to your room with the 'come' command.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -536,6 +536,7 @@ class LootProcess
     return true if game_state.casting_consume
     return true if game_state.prepare_cfb
     return true if game_state.casting_cfb
+    false
   end
 
   def arrange_mob(mob_noun, game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -422,7 +422,7 @@ class LootProcess
       end
       if @make_zombie && can_perform_priority_ritual?(game_state)
         echo "Making zombie" if $debug_mode_ct
-        do_necro_ritual(mob_noun, 'arise', game_state) unless DRSpells.active_spells.include?('Call from Beyond')
+        do_necro_ritual(mob_noun, 'arise', game_state) unless game_state.cfb_active? 
       end
       do_necro_ritual(mob_noun, @ritual_type, game_state) if can_perform_training_ritual?(game_state)
     end
@@ -1229,12 +1229,9 @@ class PetProcess
     false
   end
  
-  def cfb_active?
-    DRSpells.active_spells.include?('Call from Beyond')
-  end
 
   def summon_zombie
-    return unless cfb_active?
+    return unless game_state.cfb_active?
     return unless @zombie['summon']
     return if @is_present
 
@@ -1243,13 +1240,13 @@ class PetProcess
 
   def dismiss_zombie
     return unless @is_present
-    return unless cfb_active?
+    return unless game_state.cfb_active?
 
     command_zombie('leave')
   end
 
   def update_behavior
-    return unless cfb_active?
+    return unless game_state.cfb_active?
     return unless @zombie['behavior']
     return if @zombie['behavior'].eql?(@current_behavior)
 
@@ -1257,7 +1254,7 @@ class PetProcess
   end
 
   def update_stance
-    return unless cfb_active?
+    return unless game_state.cfb_active?
     return unless @zombie['stance']
     return if @zombie['stance'].eql?(@current_stance)
 
@@ -1266,7 +1263,7 @@ class PetProcess
 
   def check_zombie
     return unless DRStats.necromancer?
-    return unless cfb_active?
+    return unless game_state.cfb_active?
 
     summon_zombie
     update_behavior
@@ -1274,7 +1271,7 @@ class PetProcess
   end
 
   def command_zombie(command)
-    return unless cfb_active?
+    return unless game_state.cfb_active?
 
     case bput("command zombie #{command}", 'willing it to come back to you', 'You have already shifted', 'That is not a valid stance', 'is already right beside you', 'zombie shambles off with a groan', 'You sense a flicker of acknowledgement through the link', 'you sense your .+ shift into a \w+ stance', 'you sense your .+ behavior shift')
     when 'is already right beside you'
@@ -2568,6 +2565,10 @@ class GameState
     else
       "get my #{weapon_name}"
     end
+  end
+
+  def cfb_active?
+    DRSpells.active_spells.include?('Call from Beyond')
   end
 
   private
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -948,8 +948,8 @@ class SpellProcess
     game_state.cast_timer = nil
 
     game_state.casting_weapon_buff = false
-    game_state.casting_consume = false if game_state.casting_consume
-    game_state.casting_cfb = false if game_state.casting_cfb
+    game_state.casting_consume = false
+    game_state.casting_cfb = false
 
     if game_state.casting_moonblade
       if left_hand =~ /moon/ || game_state.brawling? || game_state.offhand?
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -402,6 +402,7 @@ class LootProcess
   end
 
   def can_perform_priority_ritual?(game_state)
+    return false unless DRStats.necromancer?
     return false if game_state.prepare_cfb
     return false if game_state.casting_cfb
     return false if game_state.prepare_consume
@@ -520,7 +521,7 @@ class LootProcess
 
     game_state.wield_whirlwind_offhand
 
-    unless necro_casting?(game_state)
+    if can_perform_priority_ritual?(game_state)
       while bput("loot #{@custom_loot_type}".strip, 'You search', 'I could not find what you were referring to', 'and get ready to search it') == 'and get ready to search it'
         pause
         waitrt?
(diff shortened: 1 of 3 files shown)


combat-trainer.lic
```diff
@@ -391,9 +394,19 @@ class LootProcess
     end
   end
 
-  def should_perform_ritual(prepare_consume)
-    !prepare_consume &&
-      ((@necro_train_first_aid && DRSkill.getxp('First Aid') < 32) || DRSkill.getxp('Thanatology') < 32)
+  def can_perform_training_ritual?(game_state)
```
jandersson: I broke the old method, `should_perform_ritual` into two methods here. Training rituals are rituals which are done for training purposes, like harvest and dissect. They do nothing but train the thanatology skill.

combat-trainer.lic
```diff
@@ -391,9 +394,19 @@ class LootProcess
...
+    return false unless can_perform_priority_ritual?(game_state)
+    return false if @necro_train_first_aid && DRSkill.getxp('First Aid') < 32
+    return false if DRSkill.getxp('Thanatology') < 32
+    true
+  end
+
+  def can_perform_priority_ritual?(game_state)
```
jandersson: Priority rituals are healing and zombie making rituals. They are 'priority' because they do something besides training

combat-trainer.lic
```diff
@@ -417,12 +434,17 @@ class LootProcess
...
     return if game_state.construct?(mob_noun)
+
     echo "Attempting necromancer ritual #{ritual} on #{mob_noun}" if $debug_mode_ct
     do_necro_ritual(mob_noun, 'preserve', game_state) if %w[consume harvest arise].include?(ritual)
     perform_message = "perform #{ritual} on #{mob_noun}"
-    result = bput(perform_message, @rituals['preserve'], @rituals['dissect'], @rituals['harvest'], @rituals['consume'], @rituals['construct'], @rituals['failures'])
+    result = bput(perform_message, @rituals['arise'], @rituals['preserve'], @rituals['dissect'], @rituals['harvest'], @rituals['consume'], @rituals['construct'], @rituals['failures'])
     echo result if $debug_mode_ct
     case result
+    when @rituals['arise']
```
jandersson: This is the ritual for making a zombie

combat-trainer.lic
```diff
@@ -508,11 +530,19 @@ class LootProcess
     @loot_timer = Time.now
   end
 
+  def necro_casting?(game_state)
+    return false unless DRStats.necromancer?
+    return true if game_state.prepare_consume
+    return true if game_state.casting_consume
+    return true if game_state.prepare_cfb
+    return true if game_state.casting_cfb
```
jandersson: this could return false if all of the above conditions aren't met, but 'nil' is falsey.

combat-trainer.lic
```diff
@@ -1158,6 +1205,94 @@ class SpellProcess
   end
 end
 
+class PetProcess
```
jandersson: This is the heart of the PR. PetProcess is not ZombieProcess because necromancers can use other pets like mudmen, and are supposedly getting a permanent risen pet. My hope is that this process would handle mudman and risen, or anything else.

Zombies can be buffed with the RPU spell, but it does not show up in active spells. You have to look at the zombie to see if the buff spell is active. I think this process would be a good place to expand on that.

jandersson: Also, zombies can perform a 'special' attack if commanded. It was suggested that we detect if a zombie is advancing on something and then issue the the special attack command. I think this process would be a good place to handle that.

combat-trainer.lic
rpherbig: I'm in favor of getting some basic functionality in with this PR and enhancing it over time in future PRs.

combat-trainer.lic
rpherbig: This looks a bit like PR #1629. I wonder if that logic would help here, and you could prioritize healing > zombie > training when picking a new ritual each time.

combat-trainer.lic
rpherbig: I'm fine with this being explicit.

Does this method need to return `false` (unconditionall) at the very end?

combat-trainer.lic
rpherbig: Is this similar to `game_state.necro_casting?`?

combat-trainer.lic
```diff
@@ -680,6 +710,9 @@ class SpellProcess
     @necromancer_healing = settings.necromancer_healing
     echo("  @necromancer_healing: #{@necromancer_healing}") if $debug_mode_ct
 
+    @necromancer_zombie = settings.waggle_sets['zombie']
```
rpherbig: I'm not sure how I feel about having `combat-trainer` own this buff set; vs having `buff` execute it. I guess the latter would be pretty complicated...

combat-trainer.lic
```diff
@@ -915,6 +949,7 @@ class SpellProcess
 
     game_state.casting_weapon_buff = false
     game_state.casting_consume = false if game_state.casting_consume
+    game_state.casting_cfb = false if game_state.casting_cfb
```
rpherbig: Is the conditional necessary here (and line 951)?

combat-trainer.lic
```diff
@@ -1158,6 +1205,94 @@ class SpellProcess
...
+    
+    @is_present = false
+    @current_stance = nil
+    @current_behavior = nil
+  end
+
+  def execute(game_state)
+    check_zombie
+    if game_state.dismiss_pet?
+      game_state.next_clean_up_step
```
rpherbig: Should this line and the next be done in the opposite order? I feel like we should dismiss the zombie before signaling to proceed with the next step, but I suspect this is functionally fine.

combat-trainer.lic
```diff
@@ -1158,6 +1205,94 @@ class SpellProcess
...
+      dismiss_zombie
+      return true
+    end
+    false
+  end
+ 
+  def cfb_active?
+    DRSpells.active_spells.include?('Call from Beyond')
```
rpherbig: This is also checked on line 425. Would it make sense to have a `GameState` helper for this check?

combat-trainer.lic
rpherbig: Could you take a look at the empath pet logic and see if it would be easy to move into this Process? I don't think we need to actually move it now, but I want to make sure this is flexible enough (which it looks like it is).

rpherbig:
I noticed you have `abbrev: cfb` in your sample in the PR description. Do we need to add some code to pull that data from `base-spells`?

combat-trainer.lic
jandersson: It might be.. `buff` would have to know how to perform the rituals

combat-trainer.lic
jandersson: Hmm yea theyre pretty much the same. I will remove `necro_casting`

combat-trainer.lic
jandersson: Ill check, however the prioritization of healing > zombie is already in place, on line 419 and 424

jandersson:
It looks like it already pulls the data via `dependency::transform_settings`

combat-trainer.lic
jandersson: I'm not seeing anything to borrow from that PR, let me know if i overlooked anything

combat-trainer.lic
jandersson: Yes, there is a command issued to a guardian spirit after the spell is casted. That could be moved in to PetProcess in some new method `command_spirit`, where we check if the spell is active before issuing the command. I don't have an empath with the Guardian Spirit spell else I could offer more.

combat-trainer.lic
jandersson: I'm imagining the solution as a set of common necromancy methods, though combat-trainer is still needed for making the corpse

combat-trainer.lic
rpherbig: Sounds good to me

combat-trainer.lic
rpherbig: This feels a bit like [feature envy](https://softwareengineering.stackexchange.com/questions/212128/what-is-a-feature-envy-code-and-why-is-it-considered-a-code-smell) to me. What do you think?

rpherbig:
@rcuhljr this looks good to me. Have you had a chance to review it?

combat-trainer.lic
jandersson: I see it. I think it might be worth moving this check to game_state

rcuhljr:
Nothing major jumped out at me, although the double booleans in gamestate around tracking CFB states was kind of meh. Since this is new functionality the risk for it breaking existing things is low so I'm willing to refine this one on master.

rpherbig:
@jandersson I fixed a small bug in dc8bf4b46f549aff220487cdebfef1fec741824b, but people are still reporting some issues. Can you investigate?

rpherbig:
I saw enough people reporting issues that I went ahead and reverted the PR for now.

PR #1868, opened 2017-08-26
Added stow when no lockpicks found.

Dartellum:


pick.lic
```diff
@@ -346,6 +346,8 @@ class LockPicker
     when 'referring to?'
       echo '***OUT OF LOCKPICKS***'
       beep
+      fput('stow left')
```
rpherbig: What do you think about making these conditional? Something like:
```
fput('stow right') if checkright
fput('stow left') if checkleft
```

Dartellum:
Added.

PR #1871, opened 2017-08-28
Add Zombie Support to Combat-Trainer

jandersson:
Redo of #1865


combat-trainer.lic
```diff
@@ -395,9 +395,10 @@ class LootProcess
   end
 
   def can_perform_training_ritual?(game_state)
+    return false unless DRStats.necromancer?
+    return false unless @ritual_type
     return false if game_state.necro_casting?
-    return false if @necro_train_first_aid && DRSkill.getxp('First Aid') > 32
-    return false if DRSkill.getxp('Thanatology') > 32
+    return false if DRSkill.getxp('First Aid') > 31 && DRSkill.getxp('Thanatology') > 31 && DRSkill.getxp('Skinning') > 31
```
jandersson: Based on a discussion with other players, it was decided to always do training rituals unless these three skills are locked

rpherbig: Would the "cycle rituals" PR be good to get in first? It might make some of this logic easier

jandersson: Works for me

rpherbig: #1629 is _almost_ done, are you up for finishing it?

jandersson: Yea of course, how should I do that?

jandersson:
Incorporated changes from #1873

combat-trainer.lic
```diff
@@ -418,18 +422,23 @@ class LootProcess
...
     if @last_ritual.nil?
-      if @necro_heal
+      if @necro_heal && !game_state.necro_casting?
         game_state.wounds = check_health
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
         do_necro_ritual(mob_noun, 'consume', game_state) unless game_state.wounds.empty?
       end
+      if @make_zombie && !game_state.necro_casting? && !game_state.cfb_active?
+        echo 'Making zombie' if $debug_mode_ct
+        do_necro_ritual(mob_noun, 'arise', game_state)
```
rpherbig: `Arise` will consume the corpse right? If so, should we `return false` here so we don't try and do another ritual and/or continue skinning?

jandersson: Yea good idea, we should probably do that with the consume logic in the above block too

combat-trainer.lic
```diff
@@ -702,6 +716,9 @@ class SpellProcess
     @necromancer_healing = settings.necromancer_healing
     echo("  @necromancer_healing: #{@necromancer_healing}") if $debug_mode_ct
 
+    @necromancer_zombie = settings.waggle_sets['zombie']
```
rpherbig: Is this still a waggle set?

jandersson: Yea

jandersson: Should it not be?

rpherbig: Sorry, I misread something. You're good.

rpherbig:
Looks good to me - @rcuhljr want to look this over?

rcuhljr:
Sure try it again. Not directly related to this, but I really think drinfomon needs fail safes where if it doesn't know your class it re-runs info until it has it. Empath detection functionality is too important to trust people not to fuck up their autostarts and such.

jandersson:
Should I add a check if guild is nil? then run 'info' somewhere?

rpherbig:
I think that's something we should tackle separately

jandersson:
Here we go

PR #1872, opened 2017-08-28
add premie boarding message to bescort

jandersson:
Add a message for premium users boarding the crossing ferry


jandersson:
Also looks like rubocop was ran at some point

PR #1873, opened 2017-08-28
Cycle rituals

jandersson:
Continues #1629

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -235,6 +235,9 @@ class LootProcess
     @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
 
+    @cycle_rituals = false
+    @cycle_rituals = @ritual_type == "cycle"
+
     @rituals = get_data('spells').rituals
     echo("  @rituals: #{@rituals}") if $debug_mode_ct
 
@@ -243,9 +246,6 @@ class LootProcess
 
     @necro_heal = thanatology['heal'] || false
     echo("  @necro_heal: #{@necro_heal}") if $debug_mode_ct
-
-    @necro_train_first_aid = thanatology['train_first_aid'] || false
-    echo("  @necro_train_first_aid: #{@necro_train_first_aid}") if $debug_mode_ct
 
     @necro_store = thanatology['store'] || false
     echo("  @necro_store: #{@necro_store}") if $debug_mode_ct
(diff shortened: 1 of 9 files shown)


combat-trainer.lic
```diff
@@ -392,25 +392,25 @@ class LootProcess
...
+                    'preserve'
                   elsif DRSkill.getxp('Skinning') < DRSkill.getxp('First Aid')
-                    "harvest"
+                    'harvest'
                   else
-                    "dissect"
+                    'dissect'
                   end
-    return next_ritual
+    next_ritual
```
jandersson: It might not be necessary to explicitly return/assign `next_ritual`

jandersson: It does clarify the intent of the method though...

combat-trainer.lic
```diff
@@ -392,8 +392,25 @@ class LootProcess
...
 
   def should_perform_ritual(prepare_consume)
-    !prepare_consume &&
-      ((@necro_train_first_aid && DRSkill.getxp('First Aid') < 32) || DRSkill.getxp('Thanatology') < 32)
+    return false unless @ritual_type
+    return false if prepare_consume
+    return true if @ritual_type == 'dissect' && DRSkill.getxp('First Aid') < 32
+    return true if @ritual_type == 'harvest' && DRSkill.getxp('Skinning') < 32
+    return true if DRSkill.getxp('Thanatology') < 32
+    true
```
rpherbig: Should this be `false`? If it is `true`, then you don't need the `return true if` statements above.

combat-trainer.lic
```diff
@@ -392,8 +392,25 @@ class LootProcess
...
+  end
+
+  def cycle_rituals
```
rpherbig: What do you think about renaming this `determine_next_ritual`? I think you could then get rid of `next_ritual` and return directly.

combat-trainer.lic
```diff
@@ -392,8 +392,25 @@ class LootProcess
   end
 
   def should_perform_ritual(prepare_consume)
```
rpherbig: Should this be named `should_perform_ritual?` (note the '?')?

validate.lic
```diff
@@ -450,6 +450,12 @@ class DRYamlValidator
     warn('***YOU HAVE OUTDATED SETTINGS*** THE SETTING train_in_ap IS NO LONGER USED. READ MORE HERE: https://elanthipedia.play.net/Lich_script_repository#astrology')
   end
 
+  def assert_that_necro_train_first_aid_is_deprecated(settings)
+    return unless settings.thanatology['train_first_aid']
+
+    warn('***YOU HAVE OUTDATED SETTINGS*** THE SETTING thanatology["train_first_aid"] IS NO LONGER USED. READ MORE HERE: https://elanthipedia.play.net/Lich_script_repository#combat-trainer')
```
rpherbig: I'd feel better giving the user a more informative message since we know exactly what is wrong (and the `combat-trainer` documentation can be hard to read). What do you think?

jandersson: We could tell them to set the ritual to `dissect` (the FA training ritual) or use cycle

combat-trainer.lic
jandersson: yes

combat-trainer.lic
jandersson: i honestly think both names make sense, cycle ritual implies youre cycling, but determine might make the intent more clear

combat-trainer.lic
jandersson: yes! thanks

jandersson:
Updated based on latest feedback. After testing for a bit, the cycling logic is working

PR #1884, opened 2017-09-04
Update to combat-trainer to not require killing the last enemy before quitting

adamgoerdt:


rcuhljr:
What do you think about making the setting `skip_last_enemy` It would clean up your check and gives us some safety net in that someone overwriting it to blank reverts to the base behavior.

adamgoerdt:
I think kill makes more sense than skip, because the decision is whether to
kill the target or bug out.

I just tested with kill_last_enemy: cleared and it killed the target,
rather than leaving, so the default behavior sticks unless the setting is
explicitly false.

rcuhljr:
Yes it's going to work because we're explicitly checking `== false` which we'd want to replace with `!`.  We don't want to be checking `== false` or `== true` in ruby, and in general we want a missing parameter to function the same as a defaultly set parameter.

adamgoerdt:
Any particular reason you don't want to do the explicit check?  It seems
like it would satisfy any edge case, including someone putting 'True' or 'I
like tacos' in as a setting.  Anything other than an explicit `false` will
do the default behavior

rcuhljr:
It violates ruby's standard idiom of truthy/falsey and it's going to add inconsistency to how settings behave in the yaml which makes it harder to troubleshoot. In every yaml setting an empty setting is the same as false, except for just this one?

adamgoerdt:
Now I'm confused.  In your earlier comment you said you wanted the empty setting to function the same as the default setting.  Now you're saying you want it to evaluate to false.  Which is it?

rcuhljr:
That's why I said I want to reverse the setting. If the setting is false by default an empty setting behaves the same way.

adamgoerdt:
I'm still confused.  What's the desired behavior for null values in settings:  Evaluate to false, or evaluate to the default setting?

rcuhljr:
Evaluate to false. Which is why having a settings default be falsey is useful when possible. It prevents having to add stuff to the hacky base-empty workaround.

adamgoerdt:
My big issue is that a setting like 'skip_last_enemy' doesn't describe what the setting does, where 'kill_last_enemy' does.  So we either need a more descriptive setting that can default to false, like `skip_last_kill_at_end_of_hunt` which is intuitive, albeit a bit long, use `skip_last_kill`, which is shorter but not quite as intuitive, or do something similar to how skinning sets the default (`@skin = skinning['skin'] || false`)

rcuhljr:
I don't see a difference in informativeness between kill/skip, but again I don't care what the name is just what the meaning is. `end_hunt_without_killing` would also be fine in addition to your two other suggestions.

PR #1890, opened 2017-09-06
Update to 'cycle' ritual

izawwlgood:
This is to ensure that rituals are not skipped when than is locked but FA/Skinning are not.


jandersson:
Tested this for a while, can confirm it works

PR #1891, opened 2017-09-06
Use prep_time setting in common-arcana spell casting

jandersson:
Combat-trainer makes use of a prep_time setting. I added this functionality to common-arcana. The difference is that combat-trainer is non-blocking while common-arcana is blocking.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -282,10 +282,15 @@ module DRCA
     end
 
     return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'])
+    prepare_time = Time.now
 
     find_charge_invoke_stow(settings.cambrinth, settings.stored_cambrinth, settings.cambrinth_cap, settings.dedicated_camb_use, data['cambrinth'])
 
-    waitcastrt?
+    if data['prep_time']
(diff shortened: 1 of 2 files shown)


jandersson:
To clarify, this will let ;buff and `waggle_sets` snap cast spells. I wanted to snap cast clear vision on my moon mage since it just needs to be active for many spells.

common-arcana.lic
```diff
@@ -282,10 +282,15 @@ module DRCA
     end
 
     return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'])
+    prepare_time = Time.now
 
     find_charge_invoke_stow(settings.cambrinth, settings.stored_cambrinth, settings.cambrinth_cap, settings.dedicated_camb_use, data['cambrinth'])
 
-    waitcastrt?
+    if data['prep_time']
+      pause 0.1 until checkcastrt.zero? || Time.now - prepare_time >= data['prep_time']
```
jandersson: I haven't seen `checkcastrt` used in other scripts, so ill mention that it is a built in lich function that returns the current casting roundtime as a float

PR #1897, opened 2017-09-08
DRInfomon check info if guild is nil

jandersson:


drinfomon.lic
```diff
@@ -852,6 +852,7 @@ while line = script.gets
         end
       end
     end
+    fput('info') if DRStats.guild.nil? && !dead?
```
jandersson: Just realized the command above is using 'put' not 'fput'. Should I change this around? Not familiar with `put`

rpherbig: `fput` will wait for a response from the game, `put` is more appropriate here

jandersson: Got it & fixed

PR #1912, opened 2017-09-13
Updates to Bankbot to work with skip bank

intangibles:
updated bankbot changes

old PR #1752 w/ comments

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -29,16 +29,15 @@ module DRCM
   end
 
   def hometown_currency(hometown_name)
-    return currency = get_data('town')[hometown_name]['currency']
+    currency = get_data('town')[hometown_name]['currency']
   end  
 
-  def wealth(hometown)
-    if $CURRENCIES.join("|").match /#{hometown}/i
-      currency = hometown
-    else
-      currency = hometown_currency(hometown)
-    end
+  def check_wealth(currency)
     DRC.bput("wealth #{currency}", /\(\d+ copper #{currency}\)/i, /No #{currency}/i).scan(/\d+/).first.to_i
+  end
+
+  def wealth(hometown)
+    check_wealth(hometown_currency(hometown))
   end
 
   def ensure_copper_on_hand(copper, hometown)
--- a/common-money.lic
+++ b/common-money.lic
@@ -29,7 +29,7 @@ module DRCM
   end
 
   def hometown_currency(hometown_name)
-    currency = get_data('town')[hometown_name]['currency']
+    get_data('town')[hometown_name]['currency']
   end  
 
   def check_wealth(currency)
@@ -64,7 +64,7 @@ module DRCM
   end
 
   def debt(hometown)
-    currency = get_data('town')[hometown]['currency']
+    currency = hometown_currency(hometown)
     DRC.bput('wealth', /\(\d+ copper #{currency}\)/i, /Wealth:/i).scan(/\d+/).first.to_i
   end
 end
--- a/common-money.lic
+++ b/common-money.lic
@@ -8,6 +8,8 @@ custom_require.call(%w[common common-travel])
 module DRCM
   module_function
 
+  $CURRENCIES = %w(Kronars Lirums Dokoras)
+
   def minimize_coins(copper)
     denominations = [[10_000, 'platinum'], [1000, 'gold'], [100, 'silver'], [10, 'bronze'], [1, 'copper']]
     denominations.inject([copper, []]) do |result, denomination|
@@ -28,8 +30,16 @@ module DRCM
     amount
   end
 
+  def hometown_currency(hometown_name)
+    return currency = get_data('town')[hometown_name]['currency']
+  end  
+
   def wealth(hometown)
-    currency = get_data('town')[hometown]['currency']
+    if $CURRENCIES.join("|").match /#{hometown}/i
+      currency = hometown
+    else
+      currency = hometown_currency(hometown)
+    end
     DRC.bput("wealth #{currency}", /\(\d+ copper #{currency}\)/i, /No #{currency}/i).scan(/\d+/).first.to_i
   end
 
--- a/common-money.lic
+++ b/common-money.lic
@@ -8,8 +8,6 @@ custom_require.call(%w[common common-travel])
 module DRCM
   module_function
 
-  $CURRENCIES = %w(Kronars Lirums Dokoras)
-
   def minimize_coins(copper)
     denominations = [[10_000, 'platinum'], [1000, 'gold'], [100, 'silver'], [10, 'bronze'], [1, 'copper']]
     denominations.inject([copper, []]) do |result, denomination|
--- a/sell-loot.lic
+++ b/sell-loot.lic
@@ -22,41 +22,35 @@ class SellLoot
     @character_hometown = settings.hometown
     @hometown = town_data[settings.hometown]
     @bankbot_name = settings.bankbot_name
-    @bankbot_deposit_threshold = settings.bankbot_deposit_threshold
     @bankbot_room_id = settings.bankbot_room_id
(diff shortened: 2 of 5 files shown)


common-money.lic
```diff
@@ -28,8 +30,16 @@ module DRCM
     amount
   end
 
+  def hometown_currency(hometown_name)
+    return currency = get_data('town')[hometown_name]['currency']
```
rpherbig: You don't need to assign to `currency` here (or even use an explicit `return`)

common-money.lic
```diff
@@ -28,8 +30,16 @@ module DRCM
...
+  end  
+
   def wealth(hometown)
-    currency = get_data('town')[hometown]['currency']
+    if $CURRENCIES.join("|").match /#{hometown}/i
```
rpherbig: Does this mean we're calling `wealth` with an arg of a currency sometimes and a hometown sometimes?

sell-loot.lic
```diff
@@ -19,24 +19,45 @@ class SellLoot
...
 
-    return if settings.sell_loot_skip_bank
-    exchange_coins
+    return if settings.sell_loot_skip_bank && !bankbot_enabled
+
+    @keep_coppers = parse(settings.sell_loot_money_on_hand.split(' '))
+    @keep_coppers = @bankbot_deposit_threshold if bankbot_enabled
+
+    if bankbot_enabled
+      if @bankbot_name.nil? || @bankbot_room_id.nil?
```
rpherbig: Is this more appropriate to have in `validate.lic`?

sell-loot.lic
```diff
@@ -19,24 +19,45 @@ class SellLoot
...
-    check_bankbot(settings.hometown, keep_copper)
+    if settings.sell_loot_skip_bank && bankbot_enabled
+      $CURRENCIES.each{|currency|
+        give_money_to_bankbot(currency, currency == @local_currency ? @keep_coppers : 0)
+      }
+      return
+    end
+
+    exchange_coins
+    if bankbot_enabled == false
```
rpherbig: I recommend reversing the order of the conditionals, so the `if true` clause is first.

sell-loot.lic
```diff
@@ -60,30 +81,26 @@ class SellLoot
...
+      .reject { |currency| currency =~ /#{exchange_to}/i }
       .each { |currency| fput "exchange all #{currency} for #{exchange_to}" }
   end
 
   def deposit_coins(keep_copper)
     walk_to @hometown['deposit']['id']
     release_invisibility
     bput('wealth', 'Wealth:')
-    case bput('deposit all', 'you drop all your', 'You hand the clerk some coins', "You don't have any", 'There is no teller here', 'reached the maximum balance I can permit')
+    case bput('deposit all', 'you drop all your', 'You hand the clerk some coins', "You don't have any", 'There is no teller here')
```
rpherbig: Why was this line changed?

sell-loot.lic
```diff
@@ -19,24 +19,45 @@ class SellLoot
...
     check_spare_pouch(settings.spare_gem_pouch_container, settings.gem_pouch_adjective) if settings.spare_gem_pouch_container
 
     sell_bundle if settings.sell_loot_bundle
 
-    return if settings.sell_loot_skip_bank
-    exchange_coins
+    return if settings.sell_loot_skip_bank && !bankbot_enabled
+
+    @keep_coppers = parse(settings.sell_loot_money_on_hand.split(' '))
+    @keep_coppers = @bankbot_deposit_threshold if bankbot_enabled
```
rpherbig: If `bankbot_enabled` is true, should we instead keep the max of `settings.sell_loot_money_on_hand` and `@bankbot_deposit_threshold`?

rpherbig: If not, maybe we should have an entry in `validate.lic` that says you have `sell_loot_money_on_hand` set, but also `bankbot_enabled` is true. Something to indicate that `sell_loot_money_on_hand` is ignored.

common-money.lic
intangibles: Yeah, there are a variety of places that use hometown name.  Either by calling wealth directly or indirectly via ensure_copper_on_hand, or other methods I haven't seen yet.

For example:

crossing-repair:
ensure_copper_on_hand(@repair_withdrawal_amount, settings.hometown)

deposit.lic:
copper_on_hand = wealth(hometown)

sell-loot.lic
intangibles: Possibly, there were checks to return if those settings were missing in the method give_money_to_bankbot.  I figured it would be more efficient to move them to the top so we don't go through any methods if they are missing.

Would you rather have me have exit the method w/o posting an error message?

common-money.lic
rpherbig: Could we have two methods (and the one that looks up your hometown currency could call the other one)?

sell-loot.lic
rpherbig: Or we can assume that if they have `bankbot_enabled`, they filled in both name and id?

sell-loot.lic
intangibles: Whoops, I had a rebase mishap so used my old code on updated master and didn't notice this change.

I'll update.

common-money.lic
intangibles: Updated

sell-loot.lic
intangibles: Yeah, I'm okay with that.

sell-loot.lic
intangibles: updated

sell-loot.lic
intangibles: Ideally I'd like to have one setting.  How would you feel above removing bankbot_deposit_threshhold?

common-money.lic
intangibles: Err... deposit.lic is not a dependency script. it was a copy of something I was using before.  If that matters.  It may be easier to clean up the scripts that call the method instead.

sell-loot.lic
intangibles: So... after further review on this.  Both settings are needed because bank withdraw uses "1 silver" and bankbot uses "tip 100".  They don't mix very well.  (Or at all)

I recommend they stay independent and people can just figure it out.  One goes for bank withdrawals, one is for bank deposit.

Bankbot deposit threshhold only works for the local currency based on current hometown settings. (assuming you are deposting all currencies.)  I considered added the option to keep multiple currencies but decided it would be not be used.

intangibles:
re: deposit threshholds
 
So... after further review on this. Both settings are needed because bank withdraw uses "1 silver" and bankbot uses "tip 100". They don't mix very well. (Or at all)

I recommend they stay independent and people can just figure it out. One goes for bank withdrawals, one is for bank deposit.

Bankbot deposit threshhold only works for the local currency based on current hometown settings. (assuming you are deposting all currencies.) I considered added the option to keep multiple currencies but decided it would be not be used.

intangibles:
re: deposit thresholds part 2.

Something else that has been bothering me about bankbot_deposit_threshhold.  I actually never use this because if you keep have a threshhold of "1 plat", you could end up keeping 5,000 coppers.  Instead, i set all deposits to zero and have a small withdrawal script for any pieces i want to keep.

Not sure how others use it.

common-money.lic
```diff
@@ -29,18 +29,17 @@ module DRCM
   end
 
   def hometown_currency(hometown_name)
-    return currency = get_data('town')[hometown_name]['currency']
+    currency = get_data('town')[hometown_name]['currency']
```
rpherbig: Does `currency` need to be assigned here?

rpherbig:
@rcuhljr any comments?

common-money.lic
intangibles: Ahh. No it doesn't.  Thanks!

PR #1913, opened 2017-09-14
Update Drinfomon to handle unknown guild

jandersson:
After testing the previous commit, found some errors with the previous PR. Changed the following:
* Guild is initialized to 'unknown'. Switched this to nil instead
* removed the initial info command. Leaving it in place caused 'info' to be sent several times. 
* Used fput and pause 1 to give drinfomon time to update (otherwise it spams info)

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -30,7 +30,7 @@ class DRStats
   @@concentration ||= 0
   @@favors ||= 0
   @@tdps ||= 0
-  @@encumbrance ||= nil
+  @@encumbrance = nil
   @@balance ||= 8
   @@health ||= 100
   @@mana ||= 100
@@ -704,6 +704,12 @@ end
 
 put('exp all')
 pause 1
+info_check_time = Time.now - 30
+unless dead?
+  put('info')
+  info_check_time = Time.now
+  pause 1
+end
 
 while line = script.gets
   begin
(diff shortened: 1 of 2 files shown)


drinfomon.lic
```diff
@@ -704,7 +704,6 @@ end
 
 put('exp all')
 pause 1
-put('info') unless dead?
```
jandersson: Leaving this in caused a double send of 'info'. It seems to take a few seconds for infomon to parse and register the information displayed, causing the conditional on line 854 to be true

drinfomon.lic
```diff
@@ -852,7 +851,10 @@ while line = script.gets
         end
       end
     end
-    put('info') if DRStats.guild.nil? && !dead?
+    if DRStats.guild.nil? && !dead?
+      fput('info')
```
jandersson: using put here causes multiple 'info' commands to be sent to the game. I think we want to use fput since we want a response from the game

rcuhljr: No you just want to debounce the info check, don't do it more than once every 10-30 seconds.

jandersson: That sounds good

jandersson:
Added a timer to act as a debounce mechanism.. tested this on a new character by killing them and logging in and out, then departing. The info command was ran by drinfomon immediately after i departed

PR #1926, opened 2017-09-17
Completed armorsmithing recipes, added Shard to check_tithe pladin function

bmcneff:


crossing-training.lic
```diff
@@ -231,13 +231,21 @@ class CrossingTraining
...
     return unless withdraw_exact_amount?('5 silver', @settings.hometown)
 
     wait_for_script_to_complete('pay-debt')
 
     UserVars.tithe_timer = Time.now
-    walk_to(741)
-    bput('put 5 silver kronar in almsbox', 'You drop', 'But you do not', 'attend to thy own woes')
+    ## Hardcoding Shard and Crossing almsbox rooms for now, Crossing default
+    ## TODO: check almsbox yaml setting, default to Crossing
+    if @settings.hometown == 'Shard'
```
rpherbig: This seems like it should be in `base-town.yaml`. What do you think?

rpherbig:
In the future, you may want to consider making a separate PR for logically distinct changes. That way we can merge/review them independently.

crossing-training.lic
```diff
@@ -231,13 +231,21 @@ class CrossingTraining
     return unless @settings.tithe
     delta = Time.now - (UserVars.tithe_timer || Time.now - 14_400)
     return if (delta < 4 * 60 * 60 && DRStats.paladin?) || (delta < 60 * 10 && DRStats.cleric?)
+    currency = get_data('town')[@settings.hometown]['currency']
```
rpherbig: PR #1912 is adding a helper that does this for you, let's keep an eye on which PR gets merged first.

crossing-training.lic
bmcneff: I think you're probably right. I can move that over at some point today.

bmcneff: I'll check for the existence of the value for the town in 'hometown', defaulting to Crossing if none is found.  At some point I'll go visit each of them to map them and capture the lich room id.

crossing-training.lic
```diff
@@ -229,15 +229,18 @@ class CrossingTraining
 
   def check_tithe
     return unless @settings.tithe
+    return unless almsbox = @settings.tithe_almsbox || get_data('town')[@settings.hometown]['almsbox']['id']
```
rpherbig: Can you add `tithe_almsbox:` to `base.yaml`?

crossing-training.lic
```diff
@@ -229,15 +229,18 @@ class CrossingTraining
...
     delta = Time.now - (UserVars.tithe_timer || Time.now - 14_400)
     return if (delta < 4 * 60 * 60 && DRStats.paladin?) || (delta < 60 * 10 && DRStats.cleric?)
+    ## Will convert currency fetch to 'hometown_currency' when PR 1912 gets merged
```
rpherbig: I just merged #1912, could you update this to use the new helper?

Issue #1919, opened 2017-09-15
Please add "release spell" when hunt is completed


Missah:
I have my yaml set to practice astrology after a hunt.  I had prepped calm, the hunt ended, it moved me to the safe room to start up the astrology script.  However when I tried to prep and cast pg, I couldn't because I was already prepping calm.  (Please see log below)  Because pg wasn't cast, and there was cloud cover, I was unable to see the planet Dawgolesh which put me into a loop trying to find it over and over.

Would it be possible to add "release spell" after a hunt so that this doesn't happen again?

Thank you!

  **

```
[combat-trainer]>pre calm 10
You raise your arms skyward, chanting the equation of the Calm spell.
>
[combat-trainer]>charge my camb armband 19
You harness a moderate amount of energy and attempt to channel it into your cambrinth armband.
You are able to channel all the energy into the armband.
The cambrinth armband absorbs all of the energy.
Roundtime: 3 sec.
>
* Looking as if this were a bad idea, a young forest gryphon rakes a set of large, sharp claws at you.  You partially block with a small gryphon-pelt shield sealed with protective wax.  
[You're nimbly balanced and in very strong position.]
>
[combat-trainer]>retreat
You retreat back to pole range.
>
[combat-trainer]>retreat
You retreat from combat.
>
[combat-trainer]>stow my throwing.spike
>
A young forest gryphon snaps its beak suddenly, startling a sparrow into flight.
>
You put your spike in your star-spangled sack.
>
--- Lich: combat-trainer has exited.
[hunting-buddy]>retreat
You are already as far away as you can get!
>
[hunting-buddy: ***STATUS*** EXECUTE go2 7660]
--- Lich: go2 active.
[go2: ETA: 0:00:02 (10 rooms to move through)]
.
.
.
.

>
[hunting-buddy: ***STATUS*** EXECUTE astrology]
--- Lich: astrology active.
[astrology]>prepare pg 10
You have already fully prepared the Calm spell!
>
[astrology]>charge my camb armband 20
You harness a moderate amount of energy and attempt to channel it into your cambrinth armband.
You are able to channel all the energy into the armband.
The cambrinth armband absorbs all of the energy.
Roundtime: 3 sec.
>
[astrology]>invoke my camb armband spell
The cambrinth armband pulses with Lunar energy.  You reach for its center and forge a magical link to it, readying all of its mana for your use.  By directing your focus, you ensure that mana is drawn only when spells are cast.
Roundtime: 1 sec.
[...]

PR #1929, opened 2017-09-19
release prepared spells in common-arcana

jandersson:
Partially addresses #1919 though the problem still remains in combat-trainer.

* Release prepared spells in `common-arcana::cast_spell`

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -317,7 +317,7 @@ module DRCA
     return unless data # update_astral_data returns nil on failure
 
     release_cyclics if data['cyclic']
-    fput('release spell') unless checkprep == 'None'
+    bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell") unless checkprep == 'None'
 
     if data['ritual']
       ritual(data)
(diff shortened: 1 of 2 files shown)


rpherbig:
Is there a more generic place to put this? Maybe in `buff`?

jandersson:
Makes sense. I moved it to `cast_spell`.

common-arcana.lic
```diff
@@ -317,6 +317,7 @@ module DRCA
     return unless data # update_astral_data returns nil on failure
 
     release_cyclics if data['cyclic']
+    fput('release spell') unless checkprep == 'None'
```
rpherbig: Would you consider changing the `fput` to a `bput`?

See 0aa0651bf9b9cdad3c170e4cfd5dc4950651b6d0 for an example

jandersson: Done

common-arcana.lic
```diff
@@ -317,6 +317,7 @@ module DRCA
     return unless data # update_astral_data returns nil on failure
 
     release_cyclics if data['cyclic']
+    bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell") unless checkprep == 'None'
```
rpherbig: Should this be `DRC.bput(...)`?

jandersson: My mistake. Ive fixed that.

PR #1936, opened 2017-09-21
Trash the yarn at the end of knitting.

Dartellum:
I left the commented line in to discuss. It did not work with the leftover calculation, so I decided to just dispose of the yarn outright. Thoughts? My few runs work as expected now.


rpherbig:
Yarn is pretty cheap right?

Dartellum:
Yes, yarn is cheap.

rpherbig:
Then I'm not worried about trashing it.

PR #1945, opened 2017-09-24
Setting 'tithe_almsbox' to Crossing almsbox to preserve original defa…

bmcneff:
…ult behavior


rpherbig:
If I'm reading this right, having a default value in `base.yaml` would override the hometown-based selection. Why would we not want to default to your hometown?

bmcneff:
Oh, didn't realize hometown: Crossing was set in base.yaml.  Closing this

PR #1950, opened 2017-09-30
Remove astrology_buffs from astrology.lic

jandersson:
Removes outdated `astrology_buffs`. The warning has been in place for ~2-3 months? That seems like a fair amount of time.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -44,24 +44,15 @@ class Astrology
 
   def do_buffs(settings)
     return unless settings
+    return unless settings.waggle_sets['astrology']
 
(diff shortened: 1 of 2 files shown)


rpherbig:
Should the transform in `dependency.lic` also be removed?

rpherbig:
If you're up for it, I'd love to see a similar change made for `lockpick_buffs`

jandersson:
Yea that would be nice, I dont have any plans to do that at the moment. If someone gets to it before me, that would be great

jandersson:
Settings transform removed

Issue #1542, opened 2017-04-12
Astrology needs to stow held items


jandersson:
Astrology buffing fails because I'm holding an item in my hands when it starts. It fails because invoking a focus requires empty hands (one hand to hold the focus, other hand free).

jandersson:
Preferably, this could be handled by the focus logic, since I think invoking any focus requires a free hand

PR #1951, opened 2017-09-30
Use EquipmentManager in Astrology

jandersson:
Addresses #1542 
Anytime we cast a ritual spell or use a telescope, we need both hands free.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -178,6 +178,7 @@ class Astrology
   end
 
   def check_heavens
+    @equipmentmanager.empty_hands
     vis_bodies = visible_bodies
 
     night = vis_bodies.find { |body| body['constellation'] }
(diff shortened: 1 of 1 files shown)


astrology.lic
```diff
@@ -22,6 +22,7 @@ class Astrology
     # Make sure that our Cirlce is up-to-date - avoid using stale data for new characters who joined the guild
     bput('info', 'Circle:') if DRStats.circle.zero?
 
+    @equipmentmanager = EquipmentManager.new
```
rpherbig: Personally, I'd prefer `@equipment_manager`, but it's not a big deal

jandersson: I can change that

jandersson:
Renamed `@equipmanager` as requested.

Note: I think this may end up conflicting with #1950 depending on which one gets merged first.

PR #1962, opened 2017-10-10
Healing using remedies crafted with Alchemy system.

Dartellum:


heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
+# quiet
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#tendme
```
rpherbig: `tendme` should be `heal-remedy`

heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
...
+
+args = parse_args(arg_definitions)
+$debug_mode_hr = args.debug
+
+$remedies = get_data('remedies')
+
+def wounds
+  wounds = DRCH.check_health
+  pause 1
+#  echo (wounds)
```
rpherbig: Do you want to uncomment this but make it conditional on `$debug_mode_hr`?

heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
...
+  hurt = wounds.select{|level, _| level >= 4}.values.flatten
+  wound_grouping = { "right arm" => "limbs", "left arm" => "limbs", "left leg" => "limbs", "right leg" => "limbs", "right hand" => "limbs", "left hand" => "limbs", "right eye" => "eyes", "left eye" => "eyes" }
+  scar_grouping = { "head" => "face", "neck" => "face", "eyes" => "face", "chest" => "body", "abdomen" => "body", "back" => "body" }
+  hurt = hurt.map{ |raw| wound_grouping[raw] || raw}.uniq
+  scar = hurt.map{ |raw| scar_grouping[raw] || raw}.uniq
+  echo ("Scars: ") if $debug_mode_hr
+  if $debug_mode_hr
```
rpherbig: How is this different than line 29?

heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
...
+    scar.each do |key| 
+      puts key
+    end
+  end
+  hurt.each do |key|
+    puts key if $debug_mode_hr
+    echo("In break loop") if $debug_mode_hr
+    apply_wounds(key)
+  end
+  pause 180 # pause 3 minutes before using scar remedies
```
rpherbig: Why does this need to pause?

heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
...
+  end
+end
+  
+def apply_wounds(key)
+  echo ("In apply_wounds function and key is #{key}") if $debug_mode_hr
+  rem_wounds = $remedies['remedies']['external'][key]
+  echo (rem_wounds) if $debug_mode_hr
+  rem_wounds.each do |key|
+    echo("In apply loop") if $debug_mode_hr
+    if DRCI.exists?("#{key}")
```
rpherbig: Is the string interpolation necessary, or could this be `if DRCI.exists?(key)`?

heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
...
+end
+  
+def apply_wounds(key)
+  echo ("In apply_wounds function and key is #{key}") if $debug_mode_hr
+  rem_wounds = $remedies['remedies']['external'][key]
+  echo (rem_wounds) if $debug_mode_hr
+  rem_wounds.each do |key|
+    echo("In apply loop") if $debug_mode_hr
+    if DRCI.exists?("#{key}") 
+      fput("get #{key}")
```
rpherbig: Could this use `DRC.bput`?

And the same question for each other `fput`...

heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
...
+    end
+  end
+end
+
+def apply_scars(key)
+  echo ("In apply_scar function and key is #{key}") if $debug_mode_hr
+  rem_scars = $remedies['remedies']['scars'][key]
+  echo (rem_scars) if $debug_mode_hr
+  rem_scars.each do |key|
+    echo("In apply loop") if $debug_mode_hr
```
rpherbig: Could this be extracted to a helper function? It looks like it overlaps a lot with the same part of `apply_wounds`.

heal-remedy.lic
```diff
@@ -0,0 +1,93 @@
...
+      case key
+        when /salve/, /ungent/, /poultices/, /ointment/
+          fput ("rub #{key}")
+        when /potion/, /tonic/, /elixir/, /draught/
+          fput ("drink #{key}")
+       end
+       fput("stow #{key}")
+    else
+      echo("*** No more #{key}! ***")
+#      fput("Trying general scar remedies.")
```
rpherbig: Is there more functionality you plan to implement?

heal-remedy.lic
Dartellum: Fixed.

heal-remedy.lic
Dartellum: I did.

heal-remedy.lic
Dartellum: The pause is so scars actually show from the use of the remedies so the scar remedies work.

heal-remedy.lic
Dartellum: I was going to add in general remedies if no specific ones were found. I need to work through this piece.

heal-remedy.lic
Dartellum: No round time is given for getting, applying, and stowing the remedies. Not sure what bput would offer for this.

heal-remedy.lic
Dartellum: I thought I tried it without the string interpolation and it did not work. I can test again.

Dartellum: First test it did not work. Second run through it did without #{}. Still testing. :)

heal-remedy.lic
Dartellum: Can I change the key to equal general to run through the scar|general items? Such as: key == general ?

heal-remedy.lic
Dartellum: I was also thinking, maybe with a yaml setting, one could call this from within tendme. Thoughts?

heal-remedy.lic
rcuhljr: Roundtime isn't really related to the use of bput, it's more about asserting state. Even when you don't expect roundtime, this is a healing script for when people are injured so being able to gracefully deal with RT from fleas, lag, boomer traps, and player intervention.

Dartellum: Okay, changed to DRC.bput instead of fput.

heal-remedy.lic
Dartellum: I could only do this for one of the echos. I tried for another but it did not know what key was and I do not believe I want to pass key back and forth, do I?

heal-remedy.lic
rpherbig: Do you need to `check_health` again or were scars already captured?

heal-remedy.lic
rpherbig: To get the benefit of bput, you'll want to add parameters for acceptable output text: `DRC.bput("get #{key}", 'foo', 'bar')`.

heal-remedy.lic
Dartellum: No, grabbing the areas where the initial wounds were captured.

heal-remedy.lic
```diff
@@ -0,0 +1,124 @@
...
+  echo (wounds)  if $debug_mode_hr
+  hurt = wounds.select{|level, _| level >= 4}.values.flatten
+  wound_grouping = { "right arm" => "limbs", "left arm" => "limbs", "left leg" => "limbs", "right leg" => "limbs", "right hand" => "limbs", "left hand" => "limbs", "right eye" => "eyes", "left eye" => "eyes" }
+  scar_grouping = { "right arm" => "limbs", "left arm" => "limbs", "left leg" => "limbs", "right leg" => "limbs", "right hand" => "limbs", "left hand" => "limbs", "head" => "face", "neck" => "face", "eyes" => "face", "chest" => "body", "abdomen" => "body", "back" => "body" }
+  hurt = hurt.map{ |raw| wound_grouping[raw] || raw}.uniq
+  scar = hurt.map{ |raw| scar_grouping[raw] || raw}.uniq
+  echo ("Wounds: ") if $debug_mode_hr
+  if $debug_mode_hr
+    hurt.each do |key| 
+      puts key
```
jandersson: Should this be `echo` instead of `puts`?
```
R>;en puts 'hello'            
--- Lich: exec1 active.       
--- Lich: exec1 has exited.   
R>                            
```

heal-remedy.lic
```diff
@@ -0,0 +1,124 @@
...
+def wounds
+  wounds = DRCH.check_health
+  pause 1
+  echo (wounds)  if $debug_mode_hr
+  hurt = wounds.select{|level, _| level >= 4}.values.flatten
+  wound_grouping = { "right arm" => "limbs", "left arm" => "limbs", "left leg" => "limbs", "right leg" => "limbs", "right hand" => "limbs", "left hand" => "limbs", "right eye" => "eyes", "left eye" => "eyes" }
+  scar_grouping = { "right arm" => "limbs", "left arm" => "limbs", "left leg" => "limbs", "right leg" => "limbs", "right hand" => "limbs", "left hand" => "limbs", "head" => "face", "neck" => "face", "eyes" => "face", "chest" => "body", "abdomen" => "body", "back" => "body" }
+  hurt = hurt.map{ |raw| wound_grouping[raw] || raw}.uniq
+  scar = hurt.map{ |raw| scar_grouping[raw] || raw}.uniq
+  echo ("Wounds: ") if $debug_mode_hr
```
jandersson: You could move this into the conditional below, instead of having two conditionals

heal-remedy.lic
Dartellum: Added.

heal-remedy.lic
Dartellum: yes, puts is correct.

heal-remedy.lic
Dartellum: Which part are you referring? The echo or the groupings?

jandersson: the echo statement could be added below line 30

Dartellum: The echo is the header for the list of wounds.

jandersson: I understand, just pointing out that you dont need to do `if $debug_mode_hr` twice, if the header is below line 30

Dartellum: if $debug_mode_hr
    echo ("Wounds: ")
    hurt.each do |key| 
      puts key
    end
  end

Done.

Dartellum:
Any other suggestions?

heal-remedy.lic
```diff
@@ -0,0 +1,124 @@
...
+        when /salve/, /ungent/, /poultices/, /ointment/
+          DRC.bput ("rub #{key}", "You")
+        when /potion/, /tonic/, /elixir/, /draught/
+          DRC.bput ("drink #{key}", "You")
+       end
+       DRC.bput("stow #{key}", "You")
+    else
+      echo("*** No more #{key}! ***")
+      DRC.bput("Trying general scar remedies.")
+      apply_general_scar(general)
```
rpherbig: Where is `general` defined?

Dartellum: the next def is where it is, line 99.

rpherbig: That's where the function `apply_general_scar` is declared. I'm asking where the argument to that function, `general`, is declared.

Dartellum: I want to pass the word general to the function so it goes to the general key in the yaml. Does it need quotes?

rpherbig: Ah, I see. Yes, you'll need quotes around it. That will be a string with the contents 'general'. As written right now it's a variable named general (which doesn't exist).

Dartellum:
Added a single quote around general.

PR #1972, opened 2017-10-17
base-athletics.yaml: Add silverwater ladder for 350-500 skill.

Sarvatt:


Sarvatt:
Submitting this for Subvert

PR #1974, opened 2017-10-17
combat-trainer.lic: Allow specifying an ambush location for better teaching

Sarvatt:
Changes by Majorly.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1721,6 +1721,9 @@ class AttackProcess
     @offhand_thrown = settings.offhand_thrown
     echo("  @offhand_thrown: #{@offhand_thrown}") if $debug_mode_ct
 
+    @ambush_location = settings.ambush_location
+    echo("  @amubsh_location: #{@ambush_location}") if $debug_mode_ct
+
     Flags.add('ct-ranged-ready', 'You think you have your best shot possible now')
     Flags.add('ct-face-what', 'Face what')
   end
@@ -1784,7 +1787,11 @@ class AttackProcess
 
       if (game_state.backstab? || game_state.ambush?) && hiding?
         if (game_state.backstab? && game_state.no_stab_current_mob) || (game_state.ambush? && !game_state.backstab?)
-          command += ' back'
+          if @ambush_location
+            command += " #{@ambush_location}"
+          else
+            command += ' back'
+          end
         else
           command.sub!(verb, 'backstab')
         end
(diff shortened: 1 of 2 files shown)


rcuhljr:
Has anyone verified this has any effect? They removed location targeting teaching for TM, I can't imagine it still works for weapons, and the difficulty of attacking from stealth is against creature perception regardless of body part attacked.

Sarvatt:
He said it was a very noticeable difference in backstab skill learned picking different body parts to ambush, the wiki agrees https://elanthipedia.play.net/Backstab_skill. I thought it was odd they removed this mechanic from everything else but ambush too but that seems to be the case.

Sarvatt:
It's just for the amount of backstab skill in the bucket on non bipeds, not stealth.

rcuhljr:
Oh, ambush training is just trash for backstab exp compared to khri so I didn't consider that.

combat-trainer.lic
```diff
@@ -1721,6 +1721,9 @@ class AttackProcess
     @offhand_thrown = settings.offhand_thrown
     echo("  @offhand_thrown: #{@offhand_thrown}") if $debug_mode_ct
 
+    @ambush_location = settings.ambush_location
```
rpherbig: This needs an entry in `base.yaml`

I recommend defaulting it to `back`

combat-trainer.lic
```diff
@@ -1721,6 +1721,9 @@ class AttackProcess
...
+    echo("  @amubsh_location: #{@ambush_location}") if $debug_mode_ct
```
rpherbig: "ambush"

combat-trainer.lic
```diff
@@ -1784,7 +1787,11 @@ class AttackProcess
 
       if (game_state.backstab? || game_state.ambush?) && hiding?
         if (game_state.backstab? && game_state.no_stab_current_mob) || (game_state.ambush? && !game_state.backstab?)
-          command += ' back'
+          if @ambush_location
+            command += " #{@ambush_location}"
```
rpherbig: I think you can simplify this to `command += @ambush_location`

rpherbig: Though if you default `@ambush_location` to `'back'`, you can get rid of this entire conditional

Sarvatt: Very true, getting rid of the conditional will break people who haven't restarted dependency to get the updated base though. Should I leave the conditional in for a few days then back it out when its reasonable to expect everyone has updated, or just drop it?

rpherbig: I'd leave it in for a few days, with another PR to remove it

combat-trainer.lic
Sarvatt: accounted for by https://github.com/rpherbig/dr-scripts/pull/1974/commits/c12852567a3e754a06716a664eac546b3f02350a

Sarvatt:
Addressed everything on the branch now except the conditional and I have the conditional removal ready to go

Issue #1975, opened 2017-10-18
astrology.lic


Redheadmoonie:
Can you please remove the npc check for astrology when using rituals like RTR. The script causes me to stand and retreat, losing my RTR spell and having to wait a long time before recasting. Thanks

rpherbig:
Fixed by #1976: acf5a30d5fc9a05d06192dfa11e4c614988a30a2

Issue #1904, opened 2017-09-11
Read the Ripples fails with NPCs in the room


mwest152:
Someone did some detective work and found the check is in buff.lic I think 'return if DRRoom.npcs.empty

Not sure if it checks the ignored_npc: list or not.  I tried adding the NPCs to that list and it didnt seem to work.

Specifically with RTR, if you stand while its up it falls off.  So retreating because of NPCs causes you to stand and try to retreat.

Happens with all rituals around NPCs as far as I can tell, but this one specifically you can't use while standing.

LOG BELOW:
-------------------------------

```
[astrology]>put bones in my pack

You put your bones in your grey pack.
> 
The brilliant silver glow dissipates from Lucelle.
> 
[astrology]>retreat

You are already as far away as you can get!
> 
[astrology]>stance set 100 0 80

Setting your Evasion stance to 100%, your Parry stance to 0%, and your Shield stance to 80%.  You have 10 stance points left.
> 
[astrology]>prepare RtR 400

You feel intense strain as you try to manipulate the mana streams to form this pattern, and you are not certain that you will have enough mental stamina to complete it.
However, if you utilize a ritual focus: That won't affect your current attunement very much.

You raise your head skyward, chanting the equation of the Read the Ripples spell.
> 
[astrology]>get my abacus

> 
You get a moonstone-beaded watered steel abacus etched with complex sigils from inside your grey pack.
> 
[astrology]>invoke my abacus 

You lift your abacus toward the sky and will the mana streams to bend into it.
Roundtime: 20 sec.
R> 

Your abacus steadily brightens, until it burns like a glowing white torch!
You direct the flow of power you've conjured up into your spell pattern, taking much of the strain of empowering it off you.
R> 

Your abacus fades over the course of a few seconds, returning to normal.
R> 

[astrology]>stow my abacus

You put your abacus in your grey pack.
> 
[astrology]>retreat

You are already as far away as you can get!
> 
[astrology]>sit

> 
You feel fully prepared to cast your spell.
> 
You sit down.

s> 
[astrology]>cast

You gaze towards the heavens seeking their silent guidance.
The mental strain of this pattern is considerably eased by your ritual focus.
At the ritual's peak, your prophetic connection blooms a thousand-fold.  You are alone.  An infinitesimal speck in space and time adrift in an infinite sea of possibility.  The course of your past, present and future are dictated by ceaseless currents beyond any mortal control.

[...]

PR #1976, opened 2017-10-18
Ignore listed NPC's when trying to retreat during rituals

jandersson:
When doing rituals, the retreat command is ignored if all npc's in the room match the list of ignored npc's

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -77,7 +77,7 @@ module DRCA
     DRCT.retreat(ignored_npcs)
     waitcastrt?
     return unless cast?(spell['cast'], spell['symbiosis'], spell['before'], spell['after'])
-    DRCT.retreat(ignored_npc)
+    DRCT.retreat(ignored_npcs)
   end
 
   def cast?(cast_command = 'cast', symbiosis = false, before = [], after = [])
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -65,7 +65,7 @@ module DRCA
     true
   end
   def ritual(spell, ignored_npcs)
-    DRCT.retreat(ignored_npcs)
+    DRCT.retreat(ignored_npcs) unless spell['skip_retreat']
     DRC.release_invisibility
     DRC.bput('stance set 100 0 80', 'Setting your')
 
@@ -74,10 +74,11 @@ module DRCA
 
     invoke(spell['focus'], nil)
     stow_focus(spell['focus'], spell['worn_focus'], spell['tied_focus'])
-    DRCT.retreat(ignored_npcs)
+    DRCT.retreat(ignored_npcs) unless spell['skip_retreat']
+
     waitcastrt?
     return unless cast?(spell['cast'], spell['symbiosis'], spell['before'], spell['after'])
-    DRCT.retreat(ignored_npcs)
+    DRCT.retreat(ignored_npcs) unless spell['skip_retreat']
   end
 
   def cast?(cast_command = 'cast', symbiosis = false, before = [], after = [])
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -65,8 +65,8 @@ module DRCA
     true
   end
 
-  def ritual(spell)
-    DRCT.retreat
+  def ritual(spell, settings = nil)
+    DRCT.retreat(settings.ignored_npcs)
     DRC.release_invisibility
     DRC.bput('stance set 100 0 80', 'Setting your')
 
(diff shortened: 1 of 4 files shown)


common-arcana.lic
```diff
@@ -65,8 +65,8 @@ module DRCA
     true
   end
 
-  def ritual(spell)
-    DRCT.retreat
+  def ritual(spell, settings = nil)
+    DRCT.retreat(settings.ignored_npcs)
```
rpherbig: Will this throw if settings is nil?

jandersson: strangely, it wont
```
>;en echo nil.something
--- Lich: exec1 active.
[exec1: ]
--- Lich: exec1 has exited.
```

jandersson: compared to the interpreter. I'm not arguing that what i have is correct or acceptable
```
2.0.0-p648 :001 > nil.something
NoMethodError: undefined method `something' for nil:NilClass
        from (irb):1
        from /home/jonas/.rvm/rubies/ruby-2.0.0-p648/bin/irb:12:in `<main>'
2.0.0-p648 :002 >
```

rpherbig: Oh right, `lich.rbw` modifies the behavior of `nil`...

rpherbig:
Does this fix #1975?

common-arcana.lic
jandersson: Im down to refactor it to not rely on that

jandersson:
yes and #1904

jandersson:
Updated - haven't tested

common-travel.lic
```diff
@@ -72,8 +72,9 @@ module DRCT
     room_num == Room.current.id
   end
 
-  def retreat
+  def retreat(ignored_npcs=Array.new)
     return if DRRoom.npcs.empty?
+    return if DRRoom.npcs.to_set.subtract(ignored_npcs).empty?
```
rpherbig: You don't need `to_set`, you could do something like `(DRRoom.npcs - @ignored_npcs).empty?`

This might not be a bad helper to commonize either... it's used in `combat-trainer` and `oshu-manor` (maybe others)

jandersson: Well if there are multiple `owls` or `town guards` in the room or something - to set will dedupe it

jandersson: Ohh TIL i learned
```
>;en echo DRRoom.npcs                                                  
--- Lich: exec1 active.                                                
[exec1: ["falcon", "eagle", "hawk", "kestrel", "eagle", "merlin"]]     
--- Lich: exec1 has exited.                                            
>;en echo DRRoom.npcs - ['eagle']                                      
--- Lich: exec1 active.                                                
[exec1: ["falcon", "hawk", "kestrel", "merlin"]]                       
--- Lich: exec1 has exited.                                            
```

jandersson:
I'm thinking about including some additional logic for RtR specifically
```
  def ritual(spell, ignored_npcs)
    safe_spells = ['Read the Ripples']
    DRCT.retreat(ignored_npcs) unless safe_spells.include?(spell['name'])
    DRC.release_invisibility
    DRC.bput('stance set 100 0 80', 'Setting your')
```
Is this too belt and suspenders? I think most people asking for this feature are targetting the RtR spell because retreating breaks it and starts the 30 minuteish cooldown timer.

jandersson:
While the ignored_npcs thing will work, it also forces you to maintain a list of npcs to get the spell to function properly. Theres really 0 reason to retreat when casting this spell

rpherbig:
Should there be a (new) `retreat: false` attribute on RtR in `base-spells`?

jandersson:
That works

rpherbig:
You could add `retreat: true` to all rituals and break the explicit ritual-retreat dependency, but I'm not convinced that's worth it...

jandersson:
I agree - I think RtR is the only spell that it really breaks

jandersson:
Updated and tested with these npc's in the room (which aren't in my ignored npc's list)
```
You also see a
  prairie falcon, a crested eagle, a red-winged hawk, a kestrel, a golden eagle and a merlin.
Obvious paths: east.
```

common-travel.lic
```diff
@@ -72,8 +72,9 @@ module DRCT
     room_num == Room.current.id
   end
 
-  def retreat
+  def retreat(ignored_npcs=Array.new)
     return if DRRoom.npcs.empty?
+    return if (DRRoom.npcs - ignored_npcs).empty?
```
rpherbig: Is line 76 still necessary?

jandersson: Good call, took it out

PR #1980, opened 2017-10-18
Updating afk to allow depart options

FluxBorn:
This is specifically important for necromancers who may choose to depart guild, or depart death.

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -21,6 +21,7 @@ pause 10
 settings = get_settings
 health_threshold = settings.health_threshold
 depart_on_death = settings.depart_on_death
+depart_type = settings.depart_type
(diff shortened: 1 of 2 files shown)


afk.lic
```diff
@@ -62,7 +63,7 @@ loop do
     echo 'Afk - detected death departing in 1 minute'
     echo '*' * 30
     pause 60
-    fput('depart item') if depart_on_death
+    fput('depart #{depart_type}') if depart_on_death
```
jandersson: I think you need to use double quotes for string interpolation

Sarvatt:
depart_type: item should be added to base.yaml for this

FluxBorn:
I've added it to the base.yaml, and updated with double quotes so string interpolation works.

PR #1983, opened 2017-10-19
base-{picking,town}.yaml: Add lockpick shop info for multiple towns.

Sarvatt:


data/base-picking.yaml
```diff
@@ -103,3 +103,20 @@ picking:
     slim copper: 2706
     quality copper: 10824
     slim diamondique: 27060
+    # Riverhaven
+    ordinary: 100
```
rpherbig: I'm not quite sure what it'll do if the same key appears multiple times. I suspect it'll only store the last one. Can you double check how this behaves?

Sarvatt: With this base-picking.yaml I went to crossing, bought an empty lockpick ring and attempted to pick a box. It refilled the ring with 250 kronar stout lockpicks in crossing that are also in the haven list for 200. Is that the right way to test it?

rpherbig: I'm not concerned that it will buy the wrong lockpicks. Rather, I'm concerned that it will withdraw the wrong amount of coin.

I confirmed that if the same key appears, the last value is used. That means that for the Crossing, only 200 copper will be withdrawn per `stout` lockpick (instead of 250). I suggest depositing everything, then double checking how much coin gets withdrawn.

jandersson: Would it make more sense to just defer to the crossing cost instead? Since its higher, you will be able to afford both. Maybe leave a comment saying that it exists in riverhaven and the cost is 100 incase the schema ever gets modified

rpherbig: That was my thinking too

Sarvatt:
Would this latest commit be enough then? Or is another way preferred?

rpherbig:
I don't think the logic will ever be fixed to deal with duplicates. We're using a Hash which disallows duplicate keys by its very nature.

I recommend keeping the Crossing costs and adding a comment to that section that Riverhaven has different costs but we need to keep the higher for something something reason.

Sarvatt:
That should be it then, thanks!

rpherbig:
I still don't like having duplicates in the list. Could you just add the Riverhaven comment to the Crossing section (and remove the Riverhaven items/costs)?

PR #1985, opened 2017-10-19
Set healing thresholds for necro healing

FluxBorn:
update combat trainer to allow user to set a healing threshold to avoid healing scratches if you so choose. Updated base yaml with the setting defaulting to heal everything.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -428,8 +428,8 @@ class LootProcess
       if @necro_heal && !game_state.necro_casting?
         game_state.wounds = check_health
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
-        if @heal_threshold <= game_state.wounds.keys.max
-          unless game_state.wounds.empty?
+        unless game_state.wounds.empty?
+		  if @heal_threshold <= game_state.wounds.keys.max
             do_necro_ritual(mob_noun, 'consume', game_state)
             return false
           end
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -432,11 +432,11 @@ class LootProcess
       if @necro_heal && !game_state.necro_casting?
         game_state.wounds = check_health
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
-        echo "wound_level_threshold: #{@wound_level_threshold}"
+        echo "wound_level_threshold: #{@wound_level_threshold}" if $debug_mode_ct
         unless game_state.wounds.empty?
 		      if @wound_level_threshold <= game_state.wounds.keys.max
-                do_necro_ritual(mob_noun, 'consume', game_state)
-                return false
+            do_necro_ritual(mob_noun, 'consume', game_state)
+            return false
           end
         end
       end
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -428,10 +428,12 @@ class LootProcess
       if @necro_heal && !game_state.necro_casting?
         game_state.wounds = check_health
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
-        unless game_state.wounds.empty?
-          do_necro_ritual(mob_noun, 'consume', game_state)
-          return false
-        end
+		if @heal_threshold <= game_state.wounds.keys.max
+			unless game_state.wounds.empty?
+			  do_necro_ritual(mob_noun, 'consume', game_state)
+			  return false
+			end
+		end
       end
       if @make_zombie && !game_state.necro_casting? && !game_state.cfb_active?
         echo 'Making zombie' if $debug_mode_ct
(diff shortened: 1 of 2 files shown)


izawwlgood:
Can confirm this works as intended, and will likely be useful for Empath combat healing as well.

combat-trainer.lic
```diff
@@ -428,10 +428,12 @@ class LootProcess
       if @necro_heal && !game_state.necro_casting?
         game_state.wounds = check_health
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
-        unless game_state.wounds.empty?
-          do_necro_ritual(mob_noun, 'consume', game_state)
-          return false
-        end
+		if @heal_threshold <= game_state.wounds.keys.max
```
jandersson: any chance you can set your indents to 2 spaces as per the Ruby style?

jandersson: Rubocop will fix this automagically for you if you're interested: https://github.com/bbatsov/rubocop

Running `rubocop -a` in the scripts directory will autocorrect all script files

FluxBorn: Nifty, I will use this next time.

combat-trainer.lic
```diff
@@ -428,9 +428,11 @@ class LootProcess
       if @necro_heal && !game_state.necro_casting?
         game_state.wounds = check_health
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
-        unless game_state.wounds.empty?
-          do_necro_ritual(mob_noun, 'consume', game_state)
-          return false
+        if @heal_threshold <= game_state.wounds.keys.max
```
rpherbig: Does this work if you have no wounds?

My understanding is that `game_state.wounds.keys.max` would be `nil` in that case. Comparing `1 <= nil` should result in an exception.

combat-trainer.lic
```diff
@@ -428,9 +428,11 @@ class LootProcess
...
+          unless game_state.wounds.empty?
```
rpherbig: Is this conditional still necessary?

profiles/base.yaml
```diff
@@ -82,6 +82,7 @@ buff_nonspells:
 dance_actions_stealth:
 thanatology:
 necromancer_healing:
+heal_threashold: 1
```
rpherbig: Should this be `heal_threshold`?

profiles/base.yaml
```diff
@@ -82,6 +82,7 @@ buff_nonspells:
 dance_actions_stealth:
 thanatology:
 necromancer_healing:
```
rpherbig: The setting name `heal_threshold` is rather generic. What would you think about putting `necro` or `necromancer` somewhere in the name?

Or maybe it should be a sub-setting of `thanatology` or `necromancer_healing`? @jandersson might have a better idea on that one...

profiles/base.yaml
FluxBorn: Correct, updating the typo

combat-trainer.lic
FluxBorn: very true, and I missed it in testing. I will switch the logic around to check for this.

combat-trainer.lic
FluxBorn: I switched the logic around to have the unless check for nill before moving to the if > threshold

izawwlgood:
If you can use heal_threshold for Empath battle healing too, it might make sense to simply use the same flag? If there's potential crosstalk problems, then yeah, necro_self_heal_threshold or somesuch?

profiles/base.yaml
FluxBorn: So I may add this to the regular empath healing logic as well, so not sure at that point if it should be a separate variable, or if all healing yaml should be rolled up..

rpherbig: Maybe `self_heal_threshold`?

I'm concerned it could be confused with the existing `health_threshold` setting

FluxBorn: Hmm wound_level_threshold?

FluxBorn: I suppose making it a sub of necromancer_healing would be better.

FluxBorn: Moved under necromancer_healing, and renamed wound_level_threshold

combat-trainer.lic
```diff
@@ -428,9 +432,12 @@ class LootProcess
       if @necro_heal && !game_state.necro_casting?
         game_state.wounds = check_health
         echo "Severity to Wounds: #{game_state.wounds}" if $debug_mode_ct
+        echo "wound_level_threshold: #{@wound_level_threshold}"
```
rpherbig: Should this be conditional on `if $debug_mode_ct`?

combat-trainer.lic
```diff
@@ -428,9 +432,12 @@ class LootProcess
...
         unless game_state.wounds.empty?
-          do_necro_ritual(mob_noun, 'consume', game_state)
-          return false
+		      if @wound_level_threshold <= game_state.wounds.keys.max
```
rpherbig: It looks like there is unusual indentation here, could you run Rubocop again?

rpherbig:
Is this PR adding a blank file (`necroheal.lic`)?

profiles/base.yaml
FluxBorn: let me know if the new variable name and location work for you.

PR #1990, opened 2017-10-21
Hunting buddy subconfigs

charleyc-dr:
When hunting-buddy calls out to combat-trainer based on an extra YAML file, this change makes it so that combat-trainer receives the name of the subconfig as an argument.

This makes the behavior much more inuitive: for example, prior to this change, the "offensive_spells" setting from the subconfig would be ignored by hunting-buddy unless the subconfig explicitly passed its own name as an arg in each hunting_info object.


rpherbig:
Looks good to me

@jandersson @rcuhljr any comments?

jandersson:
Is it going to add the filename to args if it already exists?

charleyc-dr:
Yes, it will.  In theory this shouldn't matter, but to be safe I'll add a check.

PR #1991, opened 2017-10-21
Separate moth hunting

jayRyan24:
Separated fuligin from void umbral moths... the ranks don't overlap well.


rpherbig:
This will break anyone using the current hunting zone.

What about instead adding two "new" zones (leaving the old combined one) and adding a check to validate?

Another option would be to leave the "old" name as the easier hunting zone (so as to not accidentally throw anyone into the harder zone) and add a "new" zone for the harder mobs.

What do you think?

jayRyan24:
I think I'm the only one hunting moths currently. I think the idea of adding umbral_moths and then another harder zone is confusing and causes issues. Other zones currently do that as well, which is kind of bad long term. I think I'll ask in chat for a while and if no one else is hunting there, the pr can go through. If some random person is having issues, we'll tell them how to fix it. Seems like the best course of action. The combined zone is terrible in the first place because the rank different is immense.

rpherbig:
@jayRyan24 Sounds good. Let me know when you feel you've done enough due diligence checking if anyone else is using this zone and I'll merge the change.

rpherbig:
@jayRyan24 Any updates on this?

jayRyan24:
Yeah, I'm the only person hunting there right now.

PR #1992, opened 2017-10-21
workorders.lic: Allow crafting shaping recipes with volumes > 10, and add furniture support.

Sarvatt:
This adds support for crafting items > 10 volumes, and adds furniture recipes. It was assumed nothing would be bigger than an order of balsa lumber before.

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -260,8 +260,6 @@ class WorkOrders
   def shape_items(info, item, quantity)
     ensure_copper_on_hand(5000, @hometown)
     recipe, items_per_stock, spare_stock, scrap = find_recipe(info, item, quantity)
-    stock_name = 'stock-name'
-    stock_number = 'stock-number'
 
     quantity.times do |count|
       if items_per_stock.zero? || (count % items_per_stock).zero?
@@ -270,9 +268,9 @@ class WorkOrders
           go_door
           pause 0.5 until Room.current.id
         end
-        bput("get my #{info[stock_name]} lumber", 'What were', 'You get')
-        while bput("count my #{info[stock_name]} lumber", 'You count.*\d+', 'I could not').scan(/\d+/).first.to_i < recipe['volume']
-          order_item(info['stock-room'], info[stock_number])
+        bput("get my #{info['stock-name']} lumber", 'What were', 'You get')
+        while bput("count my #{info['stock-name']} lumber", 'You count.*\d+', 'I could not').scan(/\d+/).first.to_i < recipe['volume']
+          order_item(info['stock-room'], info['stock-number'])
           bput('combine', 'combine')
         end
 
@@ -280,7 +278,7 @@ class WorkOrders
         buy_parts(recipe['part'], info['part-room'])
         find_shaping_room(@hometown, @engineering_room)
       end
-
+	  echo('shape', ['log', recipe['chapter'], recipe['name'], info['stock-name'], recipe['noun']])
       wait_for_script_to_complete('shape', ['log', recipe['chapter'], recipe['name'], info['stock-name'], recipe['noun']])
     end
 
--- a/workorders.lic
+++ b/workorders.lic
@@ -263,9 +263,9 @@ class WorkOrders
     stock_name = 'stock-name'
     stock_number = 'stock-number'
 
-	quantity.times do |count|
+    quantity.times do |count|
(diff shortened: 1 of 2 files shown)


Sarvatt:
Tabs need to be fixed, sorry about that.

Sarvatt:
Just noticed it's not withdrawing enough money for big stuff, please don't pull this yet.

Sarvatt:
Ok that was user error so scratch that about it not being ready to pull. I forced the wood type over to maple which is more expensive and provides half as much volume on top of allowing 5 item workorders in my yaml, that is why I didn't have enough money. Not a situation people will hit.

workorders.lic
```diff
@@ -260,15 +260,22 @@ class WorkOrders
...
     quantity.times do |count|
-      if (count % items_per_stock).zero?
+      if items_per_stock.zero? || (count % items_per_stock).zero?
         dispose("#{info['stock-name']} lumber") if count > 0 && spare_stock
         if count > 0
           go_door
           pause 0.5 until Room.current.id
         end
-        order_item(info['stock-room'], info['stock-number'])
+        fput("get my #{info[stock_name]} lumber")
```
rpherbig: Could you use `bput` here?

workorders.lic
```diff
@@ -260,15 +260,22 @@ class WorkOrders
   def shape_items(info, item, quantity)
     ensure_copper_on_hand(5000, @hometown)
     recipe, items_per_stock, spare_stock, scrap = find_recipe(info, item, quantity)
+    stock_name = 'stock-name'
+    stock_number = 'stock-number'
 
     quantity.times do |count|
-      if (count % items_per_stock).zero?
+      if items_per_stock.zero? || (count % items_per_stock).zero?
         dispose("#{info['stock-name']} lumber") if count > 0 && spare_stock
```
rpherbig: Should this be `dispose("#{info[stock_name]} lumber") if count > 0 && spare_stock`?

workorders.lic
```diff
@@ -260,15 +260,22 @@ class WorkOrders
   def shape_items(info, item, quantity)
     ensure_copper_on_hand(5000, @hometown)
     recipe, items_per_stock, spare_stock, scrap = find_recipe(info, item, quantity)
+    stock_name = 'stock-name'
```
rpherbig: What do you think about `stock_name = info['stock-name']` instead?

workorders.lic
```diff
@@ -260,15 +260,22 @@ class WorkOrders
...
+    stock_number = 'stock-number'
```
rpherbig: What do you think about `stock_number = info['stock-number']` instead?

workorders.lic
Sarvatt: for some reason switching to bput("get my #{info[stock_name]} lumber", 'What were', 'You get') kills it, after the first item it crashes with 

> --- Lich: shape has exited.
> --- Lich: error: undefined local variable or method `stock' for #<Scripting::WorkOrders:0x2f3a990>
> 	test-wo:268:in `block in shape_items'
> 	test-wo:266:in `times'

rpherbig: That's a very unexpected error, could you post the modified code +/- maybe 20 lines of code?

Sarvatt: Scratch that, it was actually this change that broke it,  not the bput.

-        dispose("#{info['stock-name']} lumber") if count > 0 && spare_stock
+        dispose("#{info[stock-name]} lumber") if count > 0 && spare_stock

workorders.lic
Sarvatt: Fixing that ended up breaking the script, and it's not something I changed from the original though

workorders.lic
Sarvatt: When I change these two to that, they end up being blank. It'll do a get my (extra space here) lumber instead of get my maple lumber.

workorders.lic
rpherbig: That would need to be `info[stock_name]` (note the underscore) to work.

workorders.lic
rpherbig: That usually indicates a typo in the thing you're using for the lookup. Maybe `-` (dash) vs `_` (underscore)?

rpherbig:
@Sarvatt it looks good to me. Are you confident in your testing?

Sarvatt:
Gave it 14 hours of testing and it is working fine, has my thumbs up now.

PR #2004, opened 2017-10-24
Clean up and add new patterns to buy_item

charleyc-dr:
Adds matches for some shops in Ratha that weren't covered.
Cleans up general interface and expands currency options.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -13,22 +13,23 @@ module DRCI
   def buy_item(room, item)
     DRCT.walk_to(room)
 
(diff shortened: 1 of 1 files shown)


common-items.lic
```diff
@@ -13,22 +13,23 @@ module DRCI
...
+                /That'll cost you (.*) (?:kronar|lirum|dokora)s?/,
+                'You decide to purchase',
+                'Buy what',
+                /I'll give that to you for (.*) (?:kronar|lirum|dokora)s?/,
+                /I ask that you give (.*) copper (?:kronar|lirum|dokora)s?/,
+                /it'll be (.*) (?:kronar|lirum|dokora)s?/]
+    
+    match = DRC.bput("buy #{item}", *patterns)
+    if match
+      patterns.each {|p| break if p.match(match)}
```
rpherbig: Do you need to explicitly match against the patterns? That is, won't `Regexp.last_match(1)` already be the amount based on the match done within `bput`?

charleyc-dr: I initially thought the same thing, but it turns out Regexp#last_match only retains the match within the current method scope.  I tried exactly that and the match is always nil.  Relevant docs:

https://ruby-doc.org/core-2.1.1/Regexp.html#method-c-last_match

rpherbig: Good to know!

PR #2006, opened 2017-10-24
Dump junk improvements.

Sarvatt:
Make item count before deciding to dump junk configurable
Lower the default to 10 items in the room

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -291,6 +291,8 @@ class LootProcess
     @dump_junk = settings.dump_junk
     echo("  @dump_junk: #{@dump_junk}") if $debug_mode_ct
     @dump_timer = Time.now - 300
+    # remove || 9 after a few days when base.yaml migrates
+    @dump_item_count = settings.dump_item_count || 9
 
     @last_rites = settings.last_rites
     echo("  @last_rites: #{@last_rites}") if $debug_mode_ct
@@ -301,7 +303,7 @@ class LootProcess
   end
 
   def execute(game_state)
-    if (Time.now - @dump_timer > 300) && @dump_junk && DRRoom.room_objs.count >= 13
+    if (Time.now - @dump_timer > 300) && @dump_junk && DRRoom.room_objs.count >= @dump_item_count
       fput 'DUMP JUNK'
       @dump_timer = Time.now
     end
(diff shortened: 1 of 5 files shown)


combat-trainer.lic
```diff
@@ -291,6 +291,8 @@ class LootProcess
     @dump_junk = settings.dump_junk
     echo("  @dump_junk: #{@dump_junk}") if $debug_mode_ct
     @dump_timer = Time.now - 300
+    # remove || 9 after a few days when base.yaml migrates
+    @dump_item_count = settings.dump_item_count || 9
```
Sarvatt: Is it possible for people to get in a state where base.yaml isn't updated but the script is? I feel like it has happened before so I put in this safeguard to be on the safe side.

mech-lore.lic
```diff
@@ -29,6 +29,15 @@ class MechLore
     fput "wear my #{@settings.hand_armor}"
   end
 
+  def dump_junk
```
rpherbig: What would you think about moving this into `common`?

Maybe something like:

```
 def dump_junk?(dump_junk, junk_timer, dump_item_count)
    return unless dump_junk
    return unless Time.now - junk_timer > 300
    return unless DRRoom.room_objs.size > dump_item_count

    fput('dump junk')
  end
```

And then the usage would be:

```
def blah
  ...
  @junk_timer = Time.now if dump_junk?(@settings.dump_junk, @junk_timer, @settings.dump_item_count)
  ...
end
```

Is that too verbose?

mech-lore.lic
```diff
@@ -29,6 +29,15 @@ class MechLore
...
+    return unless @settings.dump_junk
+    return unless Time.now - @junk_timer > 300
+    return unless DRRoom.room_objs.size > @settings.dump_item_count
+
+    fput('dump junk')
```
rpherbig: Can this be a `bput`?

combat-trainer.lic
rpherbig: 👏

rpherbig: It'd be super awesome if you could make the PR to remove this as soon as this PR is merged, with a note to let it sit for a few days. Otherwise I'll completely forget about this.

PR #2012, opened 2017-10-26
combat-trainer.lic: Remove dump_item_count safety precaution.

Sarvatt:
should be fine to remove by the 29th.


Sarvatt:
good to go.

PR #2013, opened 2017-10-26
setupaliases.lic: add lfr alias to look at raw data of remote rooms.

Sarvatt:
usage: lfr <roomnumber>. 

Happy to change the name, in my head its look far room instead of just look room like the lr alias.

### DIFF
--- a/setupaliases.lic
+++ b/setupaliases.lic
@@ -22,6 +22,9 @@
   # echos the full data of the room you're standing in
   ['lr', "#{$clean_lich_char}e echo Room[Room.current.id].inspect"],
 
+  # echos the full data of a specified remote room
+  ['lfr', "#{$clean_lich_char}e echo Room[\\?].inspect"],
+
   # these work as a pair, for recording room numbers (like setting up a hunting area); run cb, then rec in each room you want to record
   ['cb', "#{$clean_lich_char}e $temp = []"],
   ['rec', "#{$clean_lich_char}e echo $temp << Room.current.id"],


setupaliases.lic
```diff
@@ -22,6 +22,9 @@
   # echos the full data of the room you're standing in
   ['lr', "#{$clean_lich_char}e echo Room[Room.current.id].inspect"],
 
+  # echos the full data of a specified remote room
+  ['lfr', "#{$clean_lich_char}e echo Room[\\?].inspect"],
```
rpherbig: I expected this to be `\?` not `\\?` - did you run this to be sure?

jandersson:
Is there any way to have `lr` handle an optional argument? As in typing `lr` looks at the current room and `lr 1234` looks at room 1234. Just a thought, I don't know if its possible.

setupaliases.lic
jandersson: I believe its to escape the first slash so ;alias can insert the argument there. See the `fz` command

jandersson: `fz` was broken until sarvatt added the second slash

jandersson:
Not trying to hold this up by the above question - just brainstorming `lfr 2323` or `lr 2323` works great for me.

setupaliases.lic
Sarvatt: yeah I ran it first to be sure, had to fix up a few others not working without the extra escape

jandersson: I added the `fz` command manually via ;alias then assumed it would work with ;setupaliases without any changes. Good fix 👍

PR #2015, opened 2017-10-26
Check if RTR is active 100 times in 1 second

jandersson:
Some users have experienced RTR not showing up as an active spell and
pre-emptively exiting the RTR training portion of astrology. This patch
will poll active spells at most 100 times before moving on from RTR
training.

Personally, Ive been sitting on this for a couple months because I thought it was just an issue with my machine, but some other users have reported similar behaviour. This has already been tested.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -128,6 +128,15 @@ class Astrology
     waitrt?
   end
 
+  def rtr_active?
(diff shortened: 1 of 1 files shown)


jandersson:
I understand this is a symptom of a bigger issue, however the patch will at least make the spell work in the meantime.

PR #2017, opened 2017-10-26
Added mechanism to retrieve and store tools on belt, like all the cra…

Dartellum:
…fting scripts.

### DIFF
--- a/commune.lic
+++ b/commune.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#commune
 =end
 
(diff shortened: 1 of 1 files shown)


rpherbig:
Should the commented lines of code be deleted?

Dartellum:
I can remove the commented lines out. Does it need common-crafting? It works as it is?

rpherbig:
`custom_require` makes sure that the library is loaded before we try to use it. It can accidentally succeed without the `custom_require` call, but will fail for users who haven't already loaded the library through another script.

PR #2022, opened 2017-10-27
Added check for researching so athletics does not occur if so.

Dartellum:
Making the climbing section work as it should if research was started. Climb practice does not work with research. I added return if @researching to line 659.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -63,9 +63,9 @@ class CrossingTraining
   def dump_junk
     return unless @settings.dump_junk
     return unless Time.now - @junk_timer > 300
-    return unless DRRoom.room_objs.size > @settings.dump_item_count
+    return unless DRRoom.room_objs.size > 13
 
-    bput('dump junk', 'You should just kick', 'You have marked this room', 'You cannot summon')
+    fput('dump junk')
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -63,9 +63,9 @@ class CrossingTraining
   def dump_junk
     return unless @settings.dump_junk
     return unless Time.now - @junk_timer > 300
-    return unless DRRoom.room_objs.size > @settings.dump_item_count
+    return unless DRRoom.room_objs.size > 13
 
-    bput('dump junk', 'You should just kick', 'You have marked this room', 'You cannot summon')
+    fput('dump junk')
```
Sarvatt: You had unintentional changes in here and reverted the dump junk changes from earlier

Dartellum: I had an earlier version then the one that was already pushed. I was in testing. :) Should I make the changes mentioned in the

Sarvatt: yeah switch to the red ones and this whole hunk will go away :)

Dartellum: Just did, please check. :)

Sarvatt:
All good, thank ya!

PR #2026, opened 2017-10-27
Add all shrines in Ratha to base-shrines.yaml

charleyc-dr:


data/base-shrines.yaml
```diff
@@ -10,13 +10,6 @@ shrines:
     push: image
     path_out:
     - east
-  - location: The shrine of Tamsine's Rest
```
rpherbig: Why was this removed?

charleyc-dr: Duplicate; it's the same as lines 6-12.

rpherbig: 👌

PR #2030, opened 2017-10-28
bug-grabber.lic: Adjust rooms, use DRCT.walk_to

Sarvatt:
Adjust rooms to account for number changes, same as find-darkbox.


Aracheon:
Making this adjustment on my end allows the script to run, but it doesn't seem to be evaluating the bug variable properly. Even with bugs in the room, it loops with the following:

[bug-grabber]>put  on shrine

Sarvatt:
nothing stands out to me, I'll have to mess with it when it runs next. These changes will be needed regardless, DRC.walk_to does not even exist and the room numbers have changed this year.

PR #2032, opened 2017-10-29
Add support for waggle_sets and heal spell

jandersson:
If the user has defined a waggle_set named 'healing' then use that to
cast the heal spell.


jandersson:
Going to make some more changes to this an re-open later

PR #2033, opened 2017-10-29
Update check_health for disease, poison, parasites

jandersson:
Check health will return a hash with keys for wounds, disease, poison,
and parasites. Disease is a boolean, poison is a boolean, and parasites
is a hash with parasite names for keys and locations for values.

Usages have been updated with placeholder logic until everyone gets updated with common-healing


rpherbig:
It looks like `heal-remedy.lic` also needs updated.

common-healing.lic
```diff
@@ -162,13 +162,33 @@ module DRCH
   module_function
 
   def check_health
+    all_parasites = [
```
rpherbig: What do you think about moving this outside the module like the other matchers?

common-healing.lic
```diff
@@ -182,7 +202,19 @@ module DRCH
...
+    parasites = Hash.new { |h, k| h[k] = [] }
+    parasites_line.split(',').each do |parasite|
+      parasite =~ all_parasites_re
+      parasite_name = Regexp.last_match.to_s
+      parasite =~ /on your (?<part>(?:left|right)?\s?(?:\w+))/
+      location = Regexp.last_match[:part].to_s
+      parasites[parasite_name] << location
+    end
+
+    health = { 'wounds' => wounds, 'poisoned' => poisoned, 'diseased' => diseased,
```
rpherbig: I _think_ you could get away with 'returning' the hash (and not need the assignment to `health`), but I'm not sure if that makes it clearer or not...

safe-room.lic
```diff
@@ -51,7 +51,8 @@ class SafeRoom
       return true
     end
```
rpherbig: Do you want a TODO comment here?

common-healing.lic
```diff
@@ -162,13 +162,33 @@ module DRCH
...
+    poisoned = false
+
     DRC.bput('health', 'You have')
+    pause 0.5
     data = reget(300).reverse
-    wounds_line = nil
     data.each do |line|
-      if line =~ /^You have/ && line !~ /^You have no significant injuries\./ && line !~ /^You have .* lodged .* into your/ && line !~ /^You have .* poisoned/
+      case line
+      when /^You have (?!no significant injuries)(?!.* lodged into your)(?!.* infection)(?!.* poisoned)(?!.* #{all_parasites_re})/
```
rpherbig: I'm a little confused, why are the infection, poison, and parasite lines matched here and in their own `when`?

jandersson: Its a negative look ahead match on 'You have' but not followed by anything after `?!`

common-healing.lic
jandersson: That makes sense - `health` isnt being used

data/base-remedies.yaml
```diff
@@ -0,0 +1,70 @@
+---
```
jandersson: I'm not sure why this is being treated as a new file

jandersson:
Updated per code review. Also had some interesting merge issues 😅  due to cherry-picking old commits but managed to figure it out.

rpherbig:
@jandersson Please make the 'remove deprecated code' PR when you have time. Thanks!

PR #2036, opened 2017-10-30
forestry-buddy.lic: Updates to get rid of mining references.

Sarvatt:
change script name in the header
switch to using lumber_buddy_chop_every_room instead of mining_buddy_chop_every_room

### DIFF
--- a/forestry-buddy.lic
+++ b/forestry-buddy.lic
@@ -1,5 +1,5 @@
 =begin
-  Documentation: https://elanthipedia.play.net/Lich_script_repository#mining-buddy
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#forestry-buddy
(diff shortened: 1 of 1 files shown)


rcuhljr:
You may want to grand father in the old yaml setting with a loud warning for awhile.

rpherbig:
Even though the setting was blatantly named wrong? I kind of suspect no one was using this script until now...

Sarvatt:
yeah lumber_buddy_chop_every_room was in base too, i found this from someone saying it wasn't doing anything on lnet

rcuhljr:
If the base.yaml value was named wrong, no need then. I used this script for ~6 months 2-3 times a day. It's just the harvest_every_room settings are basically a terrible choice.

PR #2037, opened 2017-10-30
spinner.lic: add new bput message, make it actually work with lower s…

Sarvatt:
…kill.

It only worked if you had over 600 skill to always get the action and device in one appraisal.

### DIFF
--- a/spinner.lic
+++ b/spinner.lic
@@ -10,15 +10,15 @@ class Spinnerette
   include DRCI
 
   def initialize
-    @action = ''
-    @device = ''
+    @action = nil
+    @device = nil
     @junk_items = %w[ruazin khaddar felt silk linen wool cotton]
     start_game
   end
 
   def start_game
-    @action = ''
-    @device = ''
+    @action = nil
+    @device = nil
     fput 'push button'
     pause 1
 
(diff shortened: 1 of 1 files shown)


spinner.lic
```diff
@@ -10,44 +10,50 @@ class Spinnerette
...
+    @device = nil
     fput 'push button'
     pause 1
 
     appraise_spinner
   end
 
   def appraise_spinner
-    case bput('appraise spinner', '^You think you can do something to the .+, but you\'re not sure what.', '^You think you can .+ one of the items', '^You think you can .+ the .+ to increase your chances of spinning a good fabric.', '^You think you can .+ the .+ to increase', '^You think you can .+ the .+ to increase', '^You\'re pretty sure you\'ve improved your odds as much as you possibly can.')
+    case bput('appraise spinner', '^You think you can do something to the .+, but you\'re not sure what.', '^You think you can .+ one of the items', '^You think you can .+ the .+ to increase your chances of spinning a good fabric.', '^You think you can .+ the .+ to increase', '^You think you can .+ the .+ to increase', '^You\'re pretty sure you\'ve improved your odds as much as you possibly can.', 'You are really not sure what you should do next to the spinneret.'. 'You are pretty sure you can PUSH the BUTTON')
```
rpherbig: It looks like `next to the spinneret.'. 'You are pretty sure` should be `next to the spinneret.', 'You are pretty sure` (note the `.` that should be a `,`).

rpherbig: Fixed in eb740bf72cd03d57a7f68d5fdf54aee677a5e04b

PR #2039, opened 2017-10-31
Fix croc_swamp

skyleronken:
The croc swamp has changed. Every room now has reeds. Looking at reeds will always return the same result. However, studying reeds will only return something of interest in the room where you can exit.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -726,7 +726,7 @@ class Bescort
       end
     else
       loop do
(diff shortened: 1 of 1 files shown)


Sarvatt:
only the exit room has reeds in it for me, did you have a reed in your inv? anything happen if you tap my reed? it's good that whatever the problem was is fixed, I'm just curious what the real problem was

PR #2042, opened 2017-11-01
smash-pumpkins.lic: Various updates.

Sarvatt:
Use the fix_standing helper from common.
Check for not enough coins to continue and withdraw from the bank, or exit if not enough coin.
switch an fput to a bput to put the item in the bucket
change the safe room to the lobby now that the entrance is magic free in 2017.

### DIFF
--- a/smash-pumpkins.lic
+++ b/smash-pumpkins.lic
@@ -11,7 +11,7 @@ junk_list = get_settings.pumpkin_junk.map { |x| /\b#{x}/i }
 DRCT.walk_to(10_570)
 loop do
   break if checkbleeding
-  break if checkhealth < 75
+  break if checkhealth < 50
 
   until checkright || checkleft
     DRC.fix_standing
--- a/smash-pumpkins.lic
+++ b/smash-pumpkins.lic
@@ -51,9 +51,9 @@ loop do
       #   fput("combine")
       # end
       case DRC.bput("stow #{thing}", 'You put', 'to fit in the')
-	  when 'to fit in the'
-	    exit
-	  end
+      when 'to fit in the'
+        exit
+      end
     end
     pause
   end
--- a/smash-pumpkins.lic
+++ b/smash-pumpkins.lic
@@ -44,6 +44,7 @@ loop do
     case thing
     when *junk_list
       DRC.bput("put #{thing} in bucket", 'You drop')
+    else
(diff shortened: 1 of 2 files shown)


Sarvatt:
I found a few more fixes I might as well squeeze in here, please don't pull just yet

Sarvatt:
There we go. The timing on the second tend is iffy, 20 seconds seems to catch it every time though.

smash-pumpkins.lic
```diff
@@ -1,48 +1,67 @@
...
+    DRCI.stow_hands
+  end
   pause
 
   DRCT.walk_to(10_570)
   while thing = DRC.left_hand || DRC.right_hand
     case thing
     when *junk_list
-      fput("put #{thing} in bucket")
+      DRC.bput("put #{thing} in bucket", 'You drop')
```
rpherbig: Does `dispose_trash` do what you need?

smash-pumpkins.lic
```diff
@@ -1,48 +1,67 @@
...
     else
       # if thing =~ /ticket/
```
rpherbig: Can this be deleted?

smash-pumpkins.lic
Sarvatt: Tried changing it to DRCI.dispose_trash(thing) but it just does this
[smash-pumpkins]>put my tattered scarf in bucket
You drop a tattered scarf in a dented metal bucket.
>
[smash-pumpkins]>stow tattered scarf
Stow what?  Type 'STOW HELP' for details.
...
[smash-pumpkins: checked against [/You put/i, /to fit in the/i]]
[smash-pumpkins: for command stow skeleton earrings]

smash-pumpkins.lic
Sarvatt: done

smash-pumpkins.lic
rpherbig: That looks like what I'd expect to happen. Line 46 does a 'put' and line 47 does a 'stow'.

Sarvatt: I see the problem, I removed the else by mistake, they weren't both meant to be run. Fixed up now.

rpherbig: Cool. I'm still in favor of `dispose_trash`, but it isn't a big deal.

PR #2045, opened 2017-11-02
smash-pumpkins.lic: minor change for breaking pumpkins

Sarvatt:


rpherbig:
Is this a _breaking_ change?

PR #2051, opened 2017-11-04
base-items.yaml: Add new scroll types from HE.

Sarvatt:
stack-scrolls doesn't work unless these are added.

### DIFF
--- a/data/base-items.yaml
+++ b/data/base-items.yaml
@@ -123,6 +123,8 @@ scroll_nouns:
 - moldering scroll
 - fine scroll
 - illuminated scroll
+- amber-hued scroll
+- blood-red scroll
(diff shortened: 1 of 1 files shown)


Sarvatt:
Alternative is to define custom_scrolls: in base.yaml with these 2 types, looks like that would work for stack-scrolls.lic

Sarvatt:
Added the alternative way in https://github.com/rpherbig/dr-scripts/pull/2052

rpherbig:
I prefer this PR, to add them to `base-item.yamls`, because these is not intended to be overriden by the user.

The other PR, to add them to `base.yaml`, will cause problems for some users if they have already specified `custom_scrolls:` in their YAML.

PR #2053, opened 2017-11-05
Fix issue with setting less than 3 defenses in stances section

charleyc-dr:
This has been a repeated question in Lich chat.  The original counterintuitive behavior is that setting less than 3 defenses in a stance section causes only one defense to be set, usually with injurious consequences.


rpherbig:
Would a check in `validate.lic` be helpful here?

PR #2060, opened 2017-11-06
sanowret: Stop multiple gazes due to laggy DRStats.concentration updates

Sarvatt:
the concentration hit from using the sanowret is taking upwards of 5 seconds to register and the script was trying to loop gaze every 1 second until it did update. The 10 seconds is arbitrary but I wanted to be on the safe side, it is a passive script so an extra few seconds will not hurt. Also, it wasn't handling the already gazed message before so it was giving an error splat on the second gaze.

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -23,7 +23,7 @@ class SanowretCrystal
     return if hiding?
 
     if DRSkill.getxp('Arcana') <= 10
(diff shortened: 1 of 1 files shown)


Sarvatt:
the concentration hit from using the sanowret is taking upwards of 5 seconds to register and the script was trying to loop gaze every 1 second until it did update. Also, it wasn't handling the already gazed message before so it was giving an error splat on the second gaze.

PR #2061, opened 2017-11-06
mech-lore.lic: Remove timer usage, this is a short lived script.

Sarvatt:
This script would never run 300 seconds anyway, just try to dump junk if the setting exists and there is enough trash on the ground.

### DIFF
--- a/mech-lore.lic
+++ b/mech-lore.lic
@@ -29,15 +29,6 @@ class MechLore
     fput "wear my #{@settings.hand_armor}"
   end
 
-  def dump_junk
-    return unless @settings.dump_junk
-    return unless Time.now - @junk_timer > 300
-    return unless DRRoom.room_objs.size > @settings.dump_item_count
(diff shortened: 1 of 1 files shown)


rpherbig:
@rcuhljr @jandersson any thoughts on this one?

Sarvatt:
@junk_timer was never initialized in this and it isn't working right now, adding a @junk_timer = Time.now to the initialization just seemed silly because this doesn't run for the 300 second dump junk timer on any character I've seen.

Sarvatt:
ripping out the dump junk support from mech-lore would be an option too. crossing-training calls mech-lore and does a dump_junk itself right after

https://github.com/rpherbig/dr-scripts/blob/master/crossing-training.lic#L642

rpherbig:
Removing it seems reasonable to me

Sarvatt:
There we go

Issue #2062, opened 2017-11-07
combat-trainer box looting issue


mwest152:
Not sure why I'm running into this, could be barb specific, but i keep getting hung with a box and my box bag being full.  Only noticed on my barb, but thats my only guy regularly filling his box bag.
```
[combat-trainer]>stow my throwing.hammer

You put your hammer in your spidersilk backpack.
> 
[combat-trainer]>get my throwing.hammer

You get a steel throwing hammer from inside your spidersilk backpack.
> 
[combat-trainer]>loot goods

You search the Xala'shar magus.
The magus was carrying a crust-covered wooden chest!

> 
[combat-trainer]>stow my throwing.hammer

You put your hammer in your spidersilk backpack.
> 
[combat-trainer]>stow chest

You pick up a crust-covered wooden chest.
There isn't any more room in the haversack for that.

[combat-trainer]>get my throwing.hammer

The corpse of a hulking Xala'shar magus slowly melts away into a greenish-black pool of ooze, laced with fractured bones and several stray teeth.

> 
A hulking Xala'shar magus flexes its heavily-bandaged fingers with a low groan and softly mutters, "The price one pays to attain true power."
> 
You need a free hand to pick that up.
> 
[hunting-buddy: ***STATUS*** 22 minutes of hunting remaining or waiting on Polearms]

[combat-trainer: *** No match was found after 15 seconds, dumping info]

mwest152:
**more complete log by request**

-------------

You put your hammer in your spidersilk backpack.
> 
[combat-trainer]>get my throwing.hammer

You get a steel throwing hammer from inside your spidersilk backpack.
> 
[combat-trainer]>loot goods

You search the Xala'shar magus.
The magus was carrying a crust-covered wooden chest!

> 
[combat-trainer]>stow my throwing.hammer

You put your hammer in your spidersilk backpack.
> 
[combat-trainer]>stow chest

You pick up a crust-covered wooden chest.
There isn't any more room in the haversack for that.

> 
> 
> 
> 
> 
[combat-trainer]>get my throwing.hammer

The corpse of a hulking Xala'shar magus slowly melts away into a greenish-black pool of ooze, laced with fractured bones and several stray teeth.

> 
A hulking Xala'shar magus flexes its heavily-bandaged fingers with a low groan and softly mutters, "The price one pays to attain true power."
> 
You need a free hand to pick that up.
> 
A hulking Xala'shar magus glowers darkly at you.
[...]

PR #2073, opened 2017-11-10
Fix dropping boxes

jayRyan24:
2 or less is too little and it will sometimes miss, 4+ is too much.


jayRyan24:
Direct fix for https://github.com/rpherbig/dr-scripts/issues/2062

We have an issue in that the stow command does two things, and the bput returns a good and bad thing. So, we pick it up, but fail to put it in. We can't switch case catch the failure because the success of picking it up bypasses the when condition.

combat-trainer.lic
```diff
@@ -326,6 +326,8 @@ class LootProcess
       end
     end
     pause 0.25
+    bput("drop #{item}", 'You drop') if reget(3, 'There isn\'t any more room')
```
rpherbig: Is 3 lines of lookback enough?

Would using a Flag (like we do for pouches) be more reliable?

jayRyan24: 2 is too little, 4 is too much, 3 felt just right. We could do a flag too.

PR #2080, opened 2017-11-14
drinfomon.lic: add bank at the HE spider to hib totals.

Sarvatt:


Sarvatt:
guess I worded it wrong, this just lets transactions at the spider bank count for your hib balance in ;banks, doesn't add it to it

PR #2082, opened 2017-11-16
 pick.lic: check learning status between each pick attempt

Sarvatt:
not just when getting new boxes.


Sarvatt:
This needs more work, I'll reopen it some other time.

jandersson:
Astrology could use something very similar to this - it will predict and drain pools even if locked

PR #2085, opened 2017-11-18
Adding support for warhorn lure

skyleronken:
Adding warhorn lure on a 5 minute timer. Only executes if dancing (i.e too few creatures) and every 5 minutes.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -24,6 +24,15 @@ class SetupProcess
     @warhorn = settings.warhorn
     echo("  @warhorn: #{@warhorn}") if $debug_mode_ct
     @last_warhorn = Time.now - 305
+    if @warhorn
+      if DRCI.wearing?(@warhorn)
+        @warhorn_get_verb = "remove"
+        @warhorn_store_verb = "wear"
+      else
+        @warhorn_get_verb = "get"
+        @warhorn_store_verb = "stow"
+      end
+    end
   end
 
   def execute(game_state)
@@ -72,22 +81,17 @@ class SetupProcess
 
   def blow_warhorn()
     return unless @warhorn
-    return if Time.now - @last_warhorn < 300
+    return if Time.now - @last_warhorn < 600
 
-    if DRCI.wearing?(@warhorn)
-        get_verb = "remove"
-        store_verb = "wear"
-    else
-        get_verb = "get"
-        store_verb = "stow"
-    end
-
-    case bput("#{get_verb} my #{@warhorn}", 'You get', 'You remove', 'What were you referring to', 'You need a free hand', 'Remove what')
+    case bput("#{@warhorn_get_verb} my #{@warhorn}", 'You get', 'You remove', 'What were you referring to', 'You need a free hand', 'Remove what')
     when 'You get', 'You remove'
       bput('exhale warhorn lure','You sound a series of bursts from the warhorn','Your lungs are tired from having sounded a warhorn so recently.')
       @last_warhorn = Time.now
       waitrt?
-      bput("#{store_verb} my #{@warhorn}",'You put','You attach')
+      bput("#{@warhorn_store_verb} my #{@warhorn}",'You put','You attach')
+    when 'What were you referring to'
+      echo "#{warhorn} NOT FOUND! Removing from hunt."
+      @warhorn = ''
     end
   end
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -74,8 +74,7 @@ class SetupProcess
     return unless @warhorn
     return if Time.now - @last_warhorn < 600
 
-    case bput("tap my #{@warhorn}", 'inside your', 'that you are wearing', 'I could not find what you were referring to.')
-    when 'that you are wearing'
+    if DRCI.wearing?(@warhorn)
         get_verb = "remove"
         store_verb = "wear"
     else
@@ -83,11 +82,12 @@ class SetupProcess
         store_verb = "stow"
     end
 
-    if ['You remove','You get'].include? bput("#{get_verb} my #{@warhorn}", 'You get', 'You remove', 'What were you referring to', 'You need a free hand', 'Remove what')
+    case bput("#{get_verb} my #{@warhorn}", 'You get', 'You remove', 'What were you referring to', 'You need a free hand', 'Remove what')
+    when 'You get', 'You remove'
       bput('exhale warhorn lure','You sound a series of bursts from the warhorn','Your lungs are tired from having sounded a warhorn so recently.')
       @last_warhorn = Time.now
       waitrt?
-      fput("#{store_verb} my #{@warhorn}")
+      bput("#{store_verb} my #{@warhorn}",'You put','You attach')
     end
   end
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -20,6 +20,12 @@ class SetupProcess
     @cycle_armors = settings.cycle_armors
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @last_cycle_time = Time.now - 125
+    
+    @warhorn = settings.warhorn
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -44,6 +50,9 @@ class SetupProcess
     was_retreating = game_state.retreating?
     game_state.update_room_npcs
     if game_state.dancing?
+      if @use_warhorn
```
jayRyan24: This could be done cleaner two ways. 

```
blow_warhorn if @use_warhorn
```
or
```blow_warhorn``` and in the method
```
def blow_warhorn()
  return unless @blow_warhorn
  return if Time.now - @last_warhorn < 300
```

The second one follows the norm for the rest of Dependency.

File: combat-trainer.lic [hunk 2]
```diff
@@ -20,6 +20,12 @@ class SetupProcess
     @cycle_armors = settings.cycle_armors
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @last_cycle_time = Time.now - 125
+    
+    @warhorn = settings.warhorn
+    echo("  @warhorn: #{@warhorn}") if $debug_mode_ct
+    @use_warhorn = settings.use_warhorn
```
jayRyan24: No need for use_warhorn setting. This would just need to be 

`@warhorn` and then the check inside the methods would be `if @warhorn`

With how ruby doesn't know booleans, just checking if it exists or not is enough. If the warhorn is named, then it knows to use a warhorn.

combat-trainer.lic
```diff
@@ -65,6 +74,16 @@ class SetupProcess
 
   private
 
+  def blow_warhorn()
+    return if Time.now - @last_warhorn < 300
+    if bput("get #{@warhorn}", 'You get', 'What were you referring to', 'You need a free hand') == 'You get'
```
jayRyan24: This needs to support worn warhorns as well. What do you think about 

```
warhorn:
  name:
  is_worn:
```

combat-trainer.lic
skyleronken: Sounds good. But do you mean:
return unless @warhorn

to be consistent with dropping a flag yaml setting

combat-trainer.lic
skyleronken: done

combat-trainer.lic
```diff
@@ -65,6 +70,17 @@ class SetupProcess
 
   private
 
+  def blow_warhorn()
```
Sarvatt: just throwing it out there, how about something like ```  def blow_warhorn()
    return unless @warhorn
    return if Time.now - @last_warhorn < 600
    case bput('tap my warhorn', 'inside your', 'that you are wearing', 'I could not find what you were referring to.')
	when 'inside your'
	  if bput('get my warhorn', 'You get', 'What were you referring to', 'You need a free hand') == 'You get'
        bput('exhale warhorn lure','You sound a series of bursts from the warhorn','Your lungs are tired from having sounded a warhorn so recently.')
        @last_warhorn = Time.now
        waitrt?
        fput("stow #{@warhorn}")
    when 'that you are wearing'
	   if bput('remove my warhorn', 'you list the messages here') == 'successful message'
         bput('exhale warhorn lure','You sound a series of bursts from the warhorn','Your lungs are tired from having sounded a warhorn so recently.')
         @last_warhorn = Time.now
         waitrt?
         fput("stow #{@warhorn}")
	else
	  return
    end
  end```

skyleronken: done

combat-trainer.lic
```diff
@@ -65,6 +70,27 @@ class SetupProcess
...
+    return unless @warhorn
+    return if Time.now - @last_warhorn < 600
+
+    case bput("tap my #{@warhorn}", 'inside your', 'that you are wearing', 'I could not find what you were referring to.')
```
rpherbig: Would using `DRCI.wearing?` simplify this?

combat-trainer.lic
```diff
@@ -65,6 +70,27 @@ class SetupProcess
...
+    when 'that you are wearing'
+        get_verb = "remove"
+        store_verb = "wear"
+    else
+        get_verb = "get"
+        store_verb = "stow"
+    end
+
+    if ['You remove','You get'].include? bput("#{get_verb} my #{@warhorn}", 'You get', 'You remove', 'What were you referring to', 'You need a free hand', 'Remove what')
```
rpherbig: I prefer the `case`/`when` syntax for this kinds of situations. What do you think?

combat-trainer.lic
```diff
@@ -20,6 +20,10 @@ class SetupProcess
     @cycle_armors = settings.cycle_armors
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @last_cycle_time = Time.now - 125
+    
+    @warhorn = settings.warhorn
```
rpherbig: Can you update `base.yaml` with a default value for this setting?

combat-trainer.lic
```diff
@@ -65,6 +70,27 @@ class SetupProcess
...
+    else
+        get_verb = "get"
+        store_verb = "stow"
+    end
+
+    if ['You remove','You get'].include? bput("#{get_verb} my #{@warhorn}", 'You get', 'You remove', 'What were you referring to', 'You need a free hand', 'Remove what')
+      bput('exhale warhorn lure','You sound a series of bursts from the warhorn','Your lungs are tired from having sounded a warhorn so recently.')
+      @last_warhorn = Time.now
+      waitrt?
+      fput("#{store_verb} my #{@warhorn}")
```
rpherbig: Can this use `bput`?

combat-trainer.lic
skyleronken: based on @jayRyan24 feedback, we will be using a null value as a false check for using a warhorn. Im a Ruby noob though, so I can add it if thats not going to cause issues.

combat-trainer.lic
skyleronken: Done!

combat-trainer.lic
skyleronken: Done!

combat-trainer.lic
skyleronken: done!

combat-trainer.lic
jayRyan24: Still should be there in base. It needs to just be `warhorn:` in the base yaml. This let's people know the different settings.

combat-trainer.lic
```diff
@@ -65,6 +70,27 @@ class SetupProcess
...
+
+    if DRCI.wearing?(@warhorn)
+        get_verb = "remove"
+        store_verb = "wear"
+    else
+        get_verb = "get"
+        store_verb = "stow"
+    end
+
+    case bput("#{get_verb} my #{@warhorn}", 'You get', 'You remove', 'What were you referring to', 'You need a free hand', 'Remove what')
```
jayRyan24: What do you think about a message when no warhorn is found?

```
when 'What were you referring to'
  echo "#{@warhorn} NOT FOUND! Removing from hunt."
  @warhorn = ''
end
```

This would stop it from tapping each time through again and save a little time triggering the `return unless @warhorn` before the tap.

combat-trainer.lic
```diff
@@ -65,6 +70,27 @@ class SetupProcess
 
   private
 
+  def blow_warhorn()
+    return unless @warhorn
+    return if Time.now - @last_warhorn < 300
+
+    if DRCI.wearing?(@warhorn)
```
jayRyan24: Would this logic be better done once? Maybe in the init?

combat-trainer.lic
skyleronken: done!

combat-trainer.lic
skyleronken: done

combat-trainer.lic
skyleronken: done!

PR #2089, opened 2017-11-18
Add full container item to unlootable list.

jayRyan24:
When the container is full, the item that gets dropped now is added to the unlootable list. This stops the spamming of getting and failing to stow boxes and any other item.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -375,7 +375,10 @@ class LootProcess
       end
     end
     pause 0.25
-    bput("drop #{item}", 'You drop') if Flags['container-full']
+    if Flags['container-full']
+      bput("drop #{item}", 'You drop')
+      game_state.unlootable(item)
+    end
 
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -375,7 +375,10 @@ class LootProcess
       end
     end
     pause 0.25
-    bput("drop #{item}", 'You drop') if Flags['container-full']
+    if Flags['container-full']
+      bput("drop #{item}", 'You drop')
```
Sarvatt: should this be drop my #{item}"? drop item could default to the one on the ground and fail

jayRyan24: This shouldn't be na issue, and everything but the adding of the item to unlootables is already in there working fine. The item comes from a list of room objects. So, 99.9999% chance that the same two items don't drop from the same critter, and if it's on the no loot list, it will already need to apply to all.

PR #2091, opened 2017-11-19
[combat-trainer] Don't manipulate constructs

Ry7n:
Ensures if you run `;combat-trainer construct` you won't manipulate mobs, and adds a catch if you accidentally do (e.g. in mixed groups).

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1434,14 +1434,11 @@ class ManipulateProcess
     @threshold = settings.manipulate_threshold
     echo("  @threshold: #{@threshold}") if $debug_mode_ct
 
-    @construct_mode = settings.construct(false)
-    echo("  @construct_mode: #{@construct_mode}") if $debug_mode_ct
-     
     @last_manip = Time.now - 200
   end
 
   def execute(game_state)
-    return if game_state.danger || @threshold.nil? || @construct_mode
+    return if game_state.danger || @threshold.nil? || game_state.construct_mode?
     @filtered_npcs = game_state.npcs
     manipulate if should_manipulate?
   end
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -1434,11 +1434,14 @@ class ManipulateProcess
     @threshold = settings.manipulate_threshold
     echo("  @threshold: #{@threshold}") if $debug_mode_ct
 
+    @construct_mode = settings.construct(false)
```
rpherbig: What would you think about moving this into a predicate in `GameState`? There's a lot of duplication with `AttackProcess` here

combat-trainer.lic
```diff
@@ -1456,14 +1459,19 @@ class ManipulateProcess
     npc_occurrences = Hash.new(0)
 
     @filtered_npcs.each do |npc|
+      next if construct?(npc)
```
rpherbig: Does this need to be `game_state.construct?(npc)`?

combat-trainer.lic
```diff
@@ -1456,14 +1459,19 @@ class ManipulateProcess
...
 
       index = npc_occurrences[npc]
       ordinal_string = $ORDINALS[index]
 
-      bput("manipulate friendship #{ordinal_string} #{npc}", 'You\'re already manipulating', 'beyond your ken', 'You attempt to empathically manipulate', 'You strain')
-      manipulate_count += 1
-      npc_occurrences[npc] += 1
+      case bput("manipulate friendship #{ordinal_string} #{npc}", 'You\'re already manipulating', 'beyond your ken', 'You attempt to empathically manipulate', 'You strain', 'no life force')
+      when 'no life force' then
+        construct(npc)
```
rpherbig: Does this need to be `game_state.construct(npc)`?

combat-trainer.lic
Ry7n: Sure thing.

PR #2094, opened 2017-11-19
Fixing minor error.

jayRyan24:
Minor error with settings turned to game_state. Hotfix!

PR #2095, opened 2017-11-19
[combat-trainer.lic] Fix variable undefined error in AttackProcess

Ry7n:
Fixes the issue described in #2094:

```
--- Lich: error: undefined local variable or method `game_state' for #<Scripting::AttackProcess:0x007fb369b17d48>
	combat-trainer:1767:in `initialize'
	combat-trainer:2101:in `new'
```

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -20,7 +20,7 @@ class SetupProcess
     @cycle_armors = settings.cycle_armors
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @last_cycle_time = Time.now - 125
-    
+
     @warhorn = settings.warhorn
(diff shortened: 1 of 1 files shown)


rpherbig:
I feel like there's some tech debt we could clean up in this area. One, merge the `is_empath`, `construct_mode?`, and maybe even the undead check (~ line 1797). Two, don't check this logic every time the Process executes (although the undead check is also reliant on a buff...).

But that's for future us. The PR itself is fine for now.

Ry7n:
@rpherbig I agree! I actually dislike construct and undead modes because they don't play nice with mixed hunting grounds (Reavers, Bone Wolves).

My thought was to instead use local construct arrays in game state like necros do. But that's a fairly big change.

PR #2096, opened 2017-11-20
[combat-trainer.lic] Fix inverted logic

Ry7n:
Switching from `if charged_maneuver.empty?` to `if charged_maneuver` inverted the logic.

```ruby
if ''.empty? # true
if nil       # false
```

Switched the affected checks from `if` to `unless` in order to restore the logic.


rpherbig:
Thanks for the PR :)

PR #2097, opened 2017-11-20
Add script to do the Truffenyi commune quest.

Sarvatt:
This is so oddly specific that I'm not sure it makes sense for dr-scripts instead of repos but thought I'd share just in case

### DIFF
--- a/truffenyi-commune-quest.lic
+++ b/truffenyi-commune-quest.lic
@@ -1,0 +1,69 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#truffenyi-commune-quest
+=end
+
+custom_require.call(%w(common))
+
+echo "This script does the actions for the truffenyi commune Cleric quest. It should be started after drinking the vial two times."
+fput("stow #{checkright || checkleft}") if DRC.left_hand || DRC.right_hand
+
+while line = get
+  if line =~ /Your vision fades and you suddenly see yourself working in front of a glowing forge/
+    fput('pray Divyaush')
+  end
+  if line =~ /Your vision fades and you see yourself toiling in a dusty field. Each withered crop/
+    fput('pray Berengaria')
+  end
+  if line =~ /Your vision fades and you see yourself huddled in front of a fire in an icy cavern/
+    fput('pray Kunio')
+  end
+  if line =~ /Your vision fades and you see yourself surrounded by occupied cots. The air is still/
+    fput("pray Peri'el")
+  end
+  if line =~ /Your vision fades and you see yourself alone on a raft, calm seas in all directions./
+    fput('pray Lemicus')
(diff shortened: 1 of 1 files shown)


truffenyi-commune-quest.lic
```diff
@@ -0,0 +1,69 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#truffenyi-commune-quest
+=end
+
+custom_require.call(%w(common))
+
+echo "This script does the actions for the truffenyi commune Cleric quest. It should be started after drinking the vial two times."
+fput("stow #{checkright || checkleft}") if DRC.left_hand || DRC.right_hand
```
jandersson: What do you think about [creating an instance of EquipmentManager](https://github.com/rpherbig/dr-scripts/blob/1c8306610f09bbfa783d980f41b582c1f21f2b17/favor.lic#L12) and calling the `stow_hands` method on it?

truffenyi-commune-quest.lic
```diff
@@ -0,0 +1,69 @@
...
+    fput('pray Zachriedek')
+  end
+  if line =~ /Your stomach grumbles and you realize that you're holding a/
+    fput("drop #{checkright || checkleft}") if DRC.left_hand || DRC.right_hand
+  end
+  if line =~ /you have my attention, though really you are never far from my sight/
+    echo('All done!')
+    exit
+  end
+end
```
jandersson: Does this incur roundtime? You could throw in a `waitrt?` here to block other scripts who are waiting for it to finish.

Sarvatt: there is no RT for any of the actions, but its expected that this is run afk basically.. I used fputs because there was no way to know the success messaging for every one from one quest run, should I add waitrt?'s before each fput maybe just in case someone is running another script?

rpherbig: That might be the safest option

Or maybe a `waitrt?` at the top before parsing a line?

Sarvatt: Like this?
```while line = get
  waitrt?
  if line =~ /Your vision fades and you suddenly see yourself working in front of a glowing forge/
    fput('pray Divyaush')``` I don't want to get this wrong, I only tested what I submitted and am worried about changing it because I can never test it again :)

Sarvatt: actually nevermind, thats easy to test by doing something giving rt then doing a say Your vision fades and you suddenly see yourself working in front of a glowing forge

Sarvatt: done and tested, thanks :)

truffenyi-commune-quest.lic
Sarvatt: done

PR #2099, opened 2017-11-21
Added gate.lic

skyleronken:
script to handle moongate and teleport

### DIFF
--- a/gate.lic
+++ b/gate.lic
@@ -1,0 +1,194 @@
+custom_require.call(%w[common common-arcana events spellmonitor])
+
+# expects a 'gate' waggle set
+# expects raw channeling feat 
+
+class Moongate
+  include DRC
+  include DRCA
+
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'person', regex: /\w+/, description: 'Person to gate' },
+        { name: 'release', regex: /release/i, optional: true, description: 'Release gate once person comes through' },
+        { name: 'teleport', regex: /teleport/i, optional: true, description: 'Cast teleport instead of moongate' }
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+    @settings = get_settings
+    @lnet = (Script.running + Script.hidden).find { |val| val.name == 'lnet' }
+    gate_prep(args.person, args.release, args.teleport)
+
+  end
+
+  def gate_prep(person, release, teleport)
+    send_message(person,"Lets try to gate!")
+    check_buffs
+    moon = get_moon
+    unless moon
+      send_message(person,"No moons, no gates!")
+      exit
+    end
+    focus_moonbeam(moon)
+    send_message(person,"Locating and shifting...")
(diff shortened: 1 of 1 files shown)


gate.lic
```diff
@@ -0,0 +1,194 @@
...
+      [
+        { name: 'person', regex: /\w+/, description: 'Person to gate' },
+        { name: 'release', regex: /release/i, optional: true, description: 'Release gate once person comes through' },
+        { name: 'teleport', regex: /teleport/i, optional: true, description: 'Cast teleport instead of moongate' }
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+    @settings = get_settings
+    @lnet = (Script.running + Script.hidden).find { |val| val.name == 'lnet' }
```
rpherbig: I was working towards extracting LNet messaging into a common place but never quite finished. This would be the third place it's used (after bankbot and lichbot), so it might be worth finishing that effort...

gate.lic
```diff
@@ -0,0 +1,194 @@
...
+  def check_buffs
+    wait_for_script_to_complete('buff',['gate'])
+  end
+
+  def send_message(character, message)
+    @lnet.unique_buffer.push("chat to #{character} #{message}")
+  end
+
+  # copied from crossing-trianing
+  def check_discern(data)
```
rpherbig: This would be another good thing to commonize

gate.lic
```diff
@@ -0,0 +1,194 @@
...
+      end
+    else
+      discern_data['cambrinth'] = nil
+    end
+  end
+
+  def cast_spell(spell, mana, target, snap = false)
+
+    prepare?(spell, mana, false)
+    snap ? pause(5) : waitcastrt?
```
rpherbig: Is this logic still necessary, or will the `prep_time` attribute on a spell definition serve the same purpose?

gate.lic
```diff
@@ -0,0 +1,194 @@
+custom_require.call(%w[common common-arcana events spellmonitor])
```
rpherbig: Would you mind adding a documentation block as a header?

rpherbig:
Would you be up for running `rubocop -a` on this file?

gate.lic
skyleronken: done

gate.lic
skyleronken: yeah I agree. im not really sure how lnet interaction works. I copied this out of bankbot/lichbot.

gate.lic
skyleronken: agreed. However, these use yaml data which as I understand it the common-* scripts should not include functionality that depends on yaml.

gate.lic
skyleronken: I was trying to avoid yaml dependencies for this since this method is used for some explicit casting (i.e low prep CV, etc)

skyleronken:
ran rubocop

PR #2105, opened 2017-11-21
Adding restock.lic

skyleronken:
Added script and base data to enable restocking of consumable items like ammo

### DIFF
--- a/data/base-consumables.yaml
+++ b/data/base-consumables.yaml
@@ -1,3 +1,4 @@
+---
 Crossing:
   arrow:
     name: boar-tusk arrows
--- a/data/base-consumables.yaml
+++ b/data/base-consumables.yaml
@@ -1,0 +1,74 @@
+Crossing:
+  arrow:
+    name: boar-tusk arrows
+    room: 8263 
+    price: 62
+    size: 5
+    stackable: true
+    quantity: 30
+  bolt:
+    name: crossbow bolts
+    room: 8263
+    price: 112
+    size: 15
+    stackable: true
+    quantity: 30
+  tk_ammo:
+    name: short-handled kris
+    room: 8263
+    price: 250
+    size: 1
+    stackable: false
+    quantity: 3
+  rock:
+    name: small rocks
+    room: 8263
+    price: 12
+    size: 10
+    stackable: true
+    quantity: 30
+Shard:
+  arrow:
+    name: long arrows
+    room: 13144
+    price: 213
+    size: 15
+    stackable: true
+    quantity: 30
+  bolt:
+    name: crossbow bolts
+    room: 13144
+    price: 136
+    size: 15
+    stackable: true
+    quantity: 30
+Riverhaven:
+  arrow:
+    name: barbed arrows
+    room: 13115
+    price: 180
+    size: 15
+    stackable: true
+    quantity: 30
+  bolt:
+    name: crossbow bolts
+    room: 13115
+    price: 130
+    size: 15
+    stackable: true
+    quantity: 30
+Hibarnhvidar:
+  tk_ammo:
+    name: big-fist katar
+    room: 12172
+    cost: 2706
+    size: 1
+    stackable: false
+    quantity: 3
+  rock:
+    name: smooth rocks
+    room: 12172
+    cost: 45
+    size: 30
+    stackable: true
+    quantity: 30
--- a/restock.lic
+++ b/restock.lic
@@ -4,39 +4,10 @@
 
 custom_require.call(%w[common common-items common-money])
 
-$NUM_MAP = {
-    'zero'=> 0,
-    'one' => 1,
-    'two'=> 2,
-    'three'=> 3,
-    'four'=> 4,
-    'five'=> 5,
-    'six'=> 6,
-    'seven'=> 7,
-    'eight'=> 8,
-    'nine'=> 9,
-    'ten'=> 10,
-    'eleven'=> 11,
-    'twelve'=> 12,
-    'thirteen'=> 13,
-    'fourteen'=> 14,
-    'fifteen'=> 15,
-    'sixteen'=> 16,
-    'seventeen'=> 17,
-    'eighteen'=> 18,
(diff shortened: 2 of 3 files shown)


data/base-consumables.yaml
```diff
@@ -0,0 +1,74 @@
+Crossing:
```
rpherbig: YAML files _should_ start with `---` on its own line. It won't cause an error if it's missing, but it's best practice.

restock.lic
```diff
@@ -0,0 +1,166 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#restock
+=end
+
+custom_require.call(%w[common common-items common-money])
+
+$NUM_MAP = {
```
rpherbig: It looks like [workorders.lic](https://github.com/rpherbig/dr-scripts/blob/master/workorders.lic#L496) has something similar. Is it time to find a common place for this?

restock.lic
```diff
@@ -0,0 +1,166 @@
...
+      if remaining < item['quantity']
+        num_needed = item['quantity'] - remaining
+        buy_num = (num_needed / item['size'].to_f).ceil
+        coin_needed += buy_num * item['price']
+        item['buy_num'] = buy_num
+        items_to_restock.push(item)
+      end
+    end  
+
+    DRCM.ensure_copper_on_hand(coin_needed, @hometown) if coin_needed > 0
```
rpherbig: What do you think about adding `include DRCM` and `include DRCI` around line 39?

restock.lic
```diff
@@ -0,0 +1,166 @@
...
+      item['buy_num'].times do
+        DRCI.buy_item(item['room'],item['name'])
+        DRCI.stow_hands
+      end 
+    end 
+
+  end
+
+  def setup
+    @settings = get_settings
```
rpherbig: What do you think about _not_ making this an instance variable, and making `@restock` an instance var instead?

restock.lic
```diff
@@ -0,0 +1,166 @@
...
+  def count_stackable_item(item)
+    count_msg = bput("count my #{item}", "I could not find what you were referring to.", "tell you much of anything.", 'and see there \w+ .+ left.')
+    case count_msg
+    when 'I could not find what you were referring to.'
+      count = 0
+    when "tell you much of anything."
+      echo "#{item} is marked as stackable but is not!"
+      count = count_nonstackable_item(item) 
+    else
+      count_txt = count_msg.match(/and see there \w+ (.+) left./).captures.first.gsub('-',' ')
```
rpherbig: Could you use `Regexp.last_match(1)` here to simplify the code?

restock.lic
```diff
@@ -0,0 +1,166 @@
...
+      count = text2num(count_txt) 
+    end
+    waitrt?
+    return count
+  end
+
+  def count_nonstackable_item(item)
+    /inside your (.*).|I could not find/ =~ bput("tap my #{item}", 'inside your (.*).', 'I could not find')
```
rpherbig: Would the `inside?` helper from common-item simplify this?

rpherbig:
Do you have an example of this in action, i.e. a character's YAML that uses the `restock` setting?

rpherbig:
Would you mind running `rubocop -a` on this script?

data/base-consumables.yaml
skyleronken: done!

restock.lic
skyleronken: done

restock.lic
skyleronken: done!

restock.lic
skyleronken: done

restock.lic
skyleronken: this portion of code is to determine the container that is containing item you are trying to count. Using tap as a means to figure out the container not verify the item necessarily.

skyleronken:
@rpherbig should be good to go. made a ton of changes based on your input (included moving text2num to common.lic) and ran rubocop.

skyleronken:
Sample YAML. Includes:
arrow - demonstrating explicit use of quantity
bolt - demonstrate no overrides. All defaults from base-consumables.yaml
tk_ammo - demonstrates nonstackable purchases and overrides
rock - demonstrate a purchase quantity thats not equal to the purchasable stack size
thing - demonstrate non existent default as well as undefined custom definitions
dagger - demonstrate a custom definition 

---
restock:
  arrow:
    quantity: 30
  bolt:
  tk_ammo:
    quantity: 3
  rock:
    quantity: 15
  thing:
  dagger:
    name: wide-bladed dagger
    size: 1
    stackable: false
    room: 8263
    price: 200
    quantity: 3

restock.lic
```diff
@@ -0,0 +1,107 @@
...
+
+      next unless remaining < item['quantity']
+      num_needed = item['quantity'] - remaining
+      buy_num = (num_needed / item['size'].to_f).ceil
+      coin_needed += buy_num * item['price']
+      item['buy_num'] = buy_num
+      items_to_restock.push(item)
+    end
+
+    ensure_copper_on_hand(coin_needed, @hometown) if coin_needed > 0
```
rpherbig: Do you need this to be conditional? I thought `ensure_copper_on_hand` would take care of that

restock.lic
```diff
@@ -0,0 +1,107 @@
...
+    coin_needed = 0
+
+    items.each do |item|
+      remaining = if item['stackable']
+                    count_stackable_item(item['name'])
+                  else
+                    count_nonstackable_item(item['name'])
+                  end
+
+      next unless remaining < item['quantity']
```
rpherbig: Is this the same as `next if remaining >= item['quantity']`?

restock.lic
skyleronken: I just wanted to avoid the 'wealth' command in the event that there was nothing needing to be restocked.

restock.lic
skyleronken: yes. pretty sure rubocop made that change.

PR #2110, opened 2017-11-22
Use GG to generate favor offerings for Truffenyi's commune

charleyc-dr:
Tested by @Sarvatt ; worked the one time it was tested... probably others should try this out before committing.


Sarvatt:
Moving truff up near the top of the commune list was the one suggestion I had and he changed that, it's working theurgy a lot faster now

rpherbig:
Looks good to me. Let me know when you feel confident it's tested sufficiently.

charleyc-dr:
Alright, @Sarvatt 's been running this for a day and has seen no issues.  I'm confident enough that I think we should push this.

Issue #1736, opened 2017-06-21
Tie/Untie for clean-lumber.lic


BardWyrrd:
Request clean-lumber.lic be updated to tie/untie wood saw and drawknife from carpenter's belts.

Dartellum:
That is something i could add to it. I would then need to make sure the proper saw is in hand since you  can have a wood saw or bone saw on the belt (or both). Let me think on it.

PR #2111, opened 2017-11-22
Added tool belt support

Dartellum:
I would like some one else to test as well. Worked for me processing lots of lumber and I have two saws, bone and woodcutting. Still like another test without the belt though. This should complete #1736: Tie/Untie for clean-lumber.lic.

### DIFF
--- a/clean-lumber.lic
+++ b/clean-lumber.lic
@@ -20,11 +20,10 @@ class CleanLumber
     ]
     args = parse_args(arg_definitions)
     settings = get_settings
-    tool_storage = settings.crafting_container
     @bag = settings.crafting_container
     @belt = settings.engineering_belt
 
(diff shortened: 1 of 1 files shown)


clean-lumber.lic
```diff
@@ -21,10 +21,20 @@ class CleanLumber
     args = parse_args(arg_definitions)
     settings = get_settings
     tool_storage = settings.crafting_container
+    @bag = settings.crafting_container
```
rpherbig: What is the difference between `tool_storage` and `@bag`?

Dartellum: looking at it now, nothing. tool_storage was what was in there. I added the known pieces to make tool belts work. I can take tool_storage out and change any place referenced to @bag.

Dartellum: Testing the changes.

PR #2112, opened 2017-11-22
Added tool belt support

Dartellum:


theurgy.lic
```diff
@@ -482,15 +493,15 @@ class TheurgyActions
...
     while right_hand.include? 'limb'
       bput('carve my limb with my knife',
            'You begin to hack', 'With fluid strokes, you', 'You continue to whittle',
            'narrowing the connection', 'With a final deep cut')
       waitrt?
     end
-    put_in_container('carving knife')
-    bput('get shaper', 'You get', 'I could not find', 'What were you referring to')
+    stow_item('carving knife')
+    get_item('shaper') if GameObj.left_hand.noun != 'shaper'
```
rpherbig: Why is this conditional here? How would you already have your shaper out?

Dartellum:
It was in the older version that I copied over. I will test without.

Regards,
484-300-4833 (h)
484-431-3355 (m)

PR #2113, opened 2017-11-22
bescort.lic: Ensure you have 1 platinum on hand for the dirigible.

Sarvatt:


bescort.lic
```diff
@@ -597,12 +597,12 @@ class Bescort
     case mode
     when 'aesry'
       manual_go2(12_862)
-      if wealth('Shard') < 5000
+      if wealth('Shard') < (DRStats.circle * 50)
```
rpherbig: What do you think about doubling the expected amount of coin to account for people who just circled but haven't run `info`?

Another idea would be to add a minimum amount of coin in case `drinfomon` goes wonky, which happens, albeit rarely. Thoughts?

Sarvatt: 5000+(DRStats.circle * 50) maybe? that's 1 plat for a circle 100, is that too high? How about just doing 1 plat unconditionally in that case so even a circle 200 will have enough on hand?

rpherbig: Is there any reason low circle characters would be using the dirigible?

Sarvatt: I can think of higher level characters who may not have 1 platinum lirum in the aesry bank for the return trip at least :) i'll just submit it as 1 plat

PR #2126, opened 2017-11-25
Theurgy items update

charleyc-dr:
- Adds item prices and withdraws only as much money as necessary.
- Splits out individual product shops rather than using 'theurgy_supply_shop' (which is inconsistent in multiple ways).  Removes dependency on base-towns.yaml .
- Adds support for buying non-blessed wine.
- Adds wine shop for Shard.
- Adds support for P5.

I've tested this a few times in P5 and Shard and am happy with the behavior there.  I have not tested in Crossing.

### DIFF
--- a/data/base-theurgy.yaml
+++ b/data/base-theurgy.yaml
@@ -22,9 +22,11 @@ Crossing:
   parchment_shop:
     id: 5735
     price: 6000
+    method: buy_parchment
   taffelberry_shop:
     id: 8280
     price: 15
+    method: buy_taffelberries
   altar:
     id: 11690
   dirt_foraging:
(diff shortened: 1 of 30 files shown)


Sarvatt:
Just dropped all my existing items and did a full theurgy run in crossing with no problems 👍

data/base-theurgy.yaml
```diff
@@ -44,25 +73,69 @@ Shard:
...
+    needs_bless: true
   flint_shop:
     id: 13290
+    price: 50
   altar:
     id: 5339
+  dirt_foraging:
+    id: 5339
+
+_P5 Shared: &p5
```
rpherbig: Nice

PR #2129, opened 2017-11-25
[base.yaml] Update lumber, ore and rocks

Ry7n:
Adds all rare and the useful common ore types to base.yaml in order to ensure auto-harvesting does
not drop something potentially valuable. Some of these may not drop from from harvesting,
but better safe than sorry.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -320,27 +320,55 @@ forests_to_chop:
 lumber_skip_populated: true
 lumber_buddy_chop_every_room: false
 lumber_buddy_tree_list:
-# MISC
-#- Durian
 # common
-# - Ash #shortbows
-# - Birch #composite
-# - Maple #longbows
-# - Oak #martial
-# rare
-- Cherry #shortbows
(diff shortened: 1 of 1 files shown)


Sarvatt:
most of these weren't there because they aren't worth the time to harvest, maybe comment out the common ones?

Ry7n:
I didn't realize oravir is available in the forging society. I'll comment out that one. The rest of the commons (lead, silver, bamboo) are all best-in-class for crafting and not available in the societies. If I missed one of those, let me know.

jandersson:
It makes sense to me. RE: harvesting unwanted materials: I can see some of these not being reasonable defaults (lead and covellite jumped out), but It would encourage people to configure the script on the other hand.

Ry7n:
> lead and covellite

FWIW, my thinking here is that lead is required for pewter, which pays more than double for work orders for both Alchemy and Forging, as well as being the best non-rare weighting metal. Covellite is third best (behind two rare metals) for crafting tongs.

PR #2131, opened 2017-11-25
Theurgy prayer mat support

charleyc-dr:


charleyc-dr:
BTW, having used this a bit and thought over the intent more generally, I think this is kind of missing the point of a prayer mat.  If we're outside of combat and in an area with an easily accessible altar, the altar offers a larger set of available rituals and moving to the altar is relatively cheap.  It seems like the big advantage to the prayer mat is that you can use it in your hunting room if there aren't creatures there, which is a big boost to the available devotion/Theurgy gains in combat.

PR #2138, opened 2017-11-27
Prayer mat rituals in combat

charleyc-dr:
Note: I named the yaml setting `prayer_mat` rather than reusing `theurgy_prayer_mat` because the presence of `theurgy_prayer_mat` turns on prayer mat use in theurgy.lic, which shouldn't be linked to using the mat in combat.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1667,11 +1667,23 @@ class TrainerProcess
   end
 
   def pray_mat(game_state)
+    unless @prayer_mat && @theurgy_supply_container
+      echo '*** Cannot run PrayerMat; need prayer_mat and theurgy_supply_container defined'
+      @training_abilities.delete('PrayerMat')
+      return
+    end
     waitrt?
     retreat
     @equipment_manager.stow_weapon(game_state.weapon_name)
 
-    fput "get #{@prayer_mat} from my #{@theurgy_supply_container}"
+    unless bput(
+      "get #{@prayer_mat} from my #{@theurgy_supply_container}",
+      'You get', 'What were you referring to'
+    ) == 'You get'
+      echo '*** Cannot run PrayerMat; no prayer mat in theurgy_supply_container'
+      @training_abilities.delete('PrayerMat')
+      return
+    end
     fix_standing
     bput("unroll #{@prayer_mat}",
          'reverently lay your', 'need to be holding that first')
@@ -1679,8 +1691,9 @@ class TrainerProcess
     bput("kneel #{@prayer_mat}", 'You humbly kneel')
     bput("kiss #{@prayer_mat}", 'You bend forward to kiss')
 
-    fput "get wine from my #{@theurgy_supply_container}"
-    bput("pour wine on #{@prayer_mat || 'altar'}",
+    bput("get wine from my #{@theurgy_supply_container}",
+         'You get', 'What were you referring to')
+    bput("pour wine on #{@prayer_mat}",
          'You quietly pour', 'Pour what')
     bput("put wine in my #{@theurgy_supply_container}",
          'You put', 'What were you referring to')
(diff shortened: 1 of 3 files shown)


rpherbig:
Do many people have one prayer mat to use in combat and a different one to use out of combat?

combat-trainer.lic
```diff
@@ -1661,6 +1666,34 @@ class TrainerProcess
...
+    fput "get #{@prayer_mat} from my #{@theurgy_supply_container}"
+    fix_standing
+    bput("unroll #{@prayer_mat}",
+         'reverently lay your', 'need to be holding that first')
+
+    bput("kneel #{@prayer_mat}", 'You humbly kneel')
+    bput("kiss #{@prayer_mat}", 'You bend forward to kiss')
+
+    fput "get wine from my #{@theurgy_supply_container}"
+    bput("pour wine on #{@prayer_mat || 'altar'}",
```
rpherbig: What does 'altar' mean in this context?

rpherbig:
Have you run this when you don't have `@theurgy_supply_container` defined or when you are out of wine?

combat-trainer.lic
```diff
@@ -1661,6 +1666,34 @@ class TrainerProcess
...
+
+    fput "get #{@prayer_mat} from my #{@theurgy_supply_container}"
+    fix_standing
+    bput("unroll #{@prayer_mat}",
+         'reverently lay your', 'need to be holding that first')
+
+    bput("kneel #{@prayer_mat}", 'You humbly kneel')
+    bput("kiss #{@prayer_mat}", 'You bend forward to kiss')
+
+    fput "get wine from my #{@theurgy_supply_container}"
```
rpherbig: Can this (and line 1674) use `bput`?

profiles/base.yaml
```diff
@@ -121,6 +121,7 @@ custom_loot_type:
 last_rites: false
 ambush_location: back
 warhorn:
+prayer_mat:
```
rpherbig: What do you think about `combat_trainer_prayer_mat`?

combat-trainer.lic
charleyc-dr: It means I am a bad copy-paster 😛

profiles/base.yaml
charleyc-dr: Counteroffer: what I *actually* want is to use this setting for both theurgy.lic and for combat-trainer.lic .  The problem is that would screw up anyone who has already adopted `theurgy_prayer_mat` (but the only side effect would be that they'd pray at their local altar).  I'd like to make a `theurgy_use_prayer_mat` setting that just uses `prayer_mat` instead -- as you noted, no one has two prayer mats. OK to get rid of the old setting and sub in a new one?

combat-trainer.lic
charleyc-dr: Per your first comment: I've run when out of wine and it's fine (just "What were you referring to" and "Pour what?").  I can update to bput shortly.

profiles/base.yaml
rpherbig: I don't have the code in front of me, but couldn't the player turn off the commune in theurgy if they only wanted to use their prayer mat in combat-trainer?

combat-trainer.lic
rpherbig: 👌

combat-trainer.lic
rpherbig: 👌

profiles/base.yaml
charleyc-dr: Unfortunately I set it up so that the presence of `theurgy_prayer_mat` causes it to be used.  And it's several rituals that get subbed out.

rpherbig: Got it.

To back up a step then, I'm not sure I understand your counteroffer. Are you suggesting adding `theurgy_use_prayer_mat` as a boolean which is used in theurgy and determines whether or not to use the prayer mat?

charleyc-dr: That's right; I've gone ahead and done this.  Take a look and see if you think the new setup is reasonable.

PR #2146, opened 2017-11-28
Harness when Arcana full. 

jayRyan24:
Changes the use of Cambrinth to Harness when Arcana is full. Does not Harness if attunement is above 30. 

I've been running this all night, but the benefits really push themselves for someone that has a lot of attunement. Using harness over Cambrinth teaches Attunement, and this with the use of Perc/Pow during combat will lock attunement quick and require less power walking, and when at 800+ attunement, and magic secondary, and worst of all prime, can take a very long time, even for just a few mind levels. 

This needs more testing, and some feedback on implementation. 

Also, note that due to how harnessing works and the loss of mana from leakage (means less with higher and higher attunement, and with some cyclics running, it was decided to harness all the mana after the spell has been completed. Another check we can do here is wait for the spell to have, let's say, 3 RT remaining. 3 RT due to the fact that it's the RT for harnessing, minimum, that I've seen.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -837,7 +837,7 @@ class Bescort
     end
     if entering_ap
       focus_shard(shard) if DRStats.circle < 100
-      3.times { harness_mana(['20']) }
+      harness_mana(['20', '20', '20'])
       waitcastrt?
       if DRStats.circle > 99
         unless cast?('cast grazhir')
--- a/bescort.lic
+++ b/bescort.lic
@@ -801,7 +801,7 @@ class Bescort
   def power_walk(pattern)
     fput 'exit' if dead? || Flags['ap-danger']
     if Flags['harness-check']
-      harness_mana(['20']) unless Flags['harness-check'].include?('effortlessly')
+      harness_mana([20]) unless Flags['harness-check'].include?('effortlessly')
       Flags.reset('harness-check')
     end
     result = bput('pow', pattern)
@@ -837,7 +837,7 @@ class Bescort
     end
     if entering_ap
       focus_shard(shard) if DRStats.circle < 100
-      harness_mana(['20', '20', '20'])
+      harness_mana([20, 20, 20])
       waitcastrt?
       if DRStats.circle > 99
         unless cast?('cast grazhir')
--- a/bescort.lic
+++ b/bescort.lic
@@ -801,7 +801,7 @@ class Bescort
   def power_walk(pattern)
     fput 'exit' if dead? || Flags['ap-danger']
     if Flags['harness-check']
-      harness_mana(20) unless Flags['harness-check'].include?('effortlessly')
+      harness_mana(['20']) unless Flags['harness-check'].include?('effortlessly')
       Flags.reset('harness-check')
     end
     result = bput('pow', pattern)
@@ -837,7 +837,7 @@ class Bescort
     end
     if entering_ap
       focus_shard(shard) if DRStats.circle < 100
-      3.times { harness_mana(20) }
+      3.times { harness_mana(['20']) }
       waitcastrt?
       if DRStats.circle > 99
         unless cast?('cast grazhir')
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -24,7 +24,7 @@ module DRCA
 
     loop do
       pause 5 while mana <= amount
-      harness_mana(amount) if harness
+      harness_mana(['amount']) if harness
       break if success.include?(DRC.bput("infuse om #{amount}", success, failure))
       pause 0.5
       waitrt?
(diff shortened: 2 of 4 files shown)


common-arcana.lic
```diff
@@ -24,17 +24,20 @@ module DRCA
 
     loop do
       pause 5 while mana <= amount
-      harness_mana(amount) if harness
+      harness_mana(['amount']) if harness
```
rpherbig: Won't this pass 'amount' as an arg, instead of the value of the `amount` variable?

common-arcana.lic
```diff
@@ -341,4 +348,12 @@ module DRCA
 
     cast?(data['cast'], data['symbiosis'], data['before'], data['after'])
   end
+
+  def check_to_harness(should_harness)
+    return false unless should_harness.equals? 'true'
+    return false unless DRSkill.getxp('Attunement') < DRSkill.getxp('Arcana')
```
rpherbig: Do you think `return false if DRSkill.getxp('Attunement') >= DRSkill.getxp('Arcana')` would be clearer?

common-arcana.lic
```diff
@@ -341,4 +348,12 @@ module DRCA
...
+    return false unless DRSkill.getxp('Attunement') < 30
```
rpherbig: Do you think `return false if DRSkill.getxp('Attunement') >= 30` would be clearer?

common-arcana.lic
jayRyan24: Yes, fixed it. Thanks!

common-arcana.lic
```diff
@@ -341,4 +348,12 @@ module DRCA
 
     cast?(data['cast'], data['symbiosis'], data['before'], data['after'])
   end
+
+  def check_to_harness(should_harness)
+    return false unless should_harness.equals? 'true'
```
rpherbig: Why not `return false unless should_harness`?

common-arcana.lic
jayRyan24: Sure, changed it.

common-arcana.lic
jayRyan24: Thought about this, but I figured maybe someone would want to turn it off and on, and if they do so in their yamls, they'll have to remember that this works with removal and not with false, so they can swap false and true comfortably without having to know how the internals of Ruby work. It seems more user friendly with this, if that makes sense.

jayRyan24: I actually worked it the way you are suggesting, but like I said, this is more user friendly. I think that when left on, this setup can actually hurt someone pretty hard without the attunement to support it.

common-arcana.lic
jayRyan24: These are clearer, but throughout this app we have unless and I've had it suggested to use unless over if quite a few times, lol. I don't care either way, it's the same thing to me.

jayRyan24: If is standard, so it's always clearer, I think. Unless is if not.

rpherbig: I prefer `if` whenever possible.

That said, I prefer `unless` to `if !...`

jayRyan24: Gotcha, I'll remember that for the future for sure.

common-arcana.lic
rpherbig: I'm not sure I understand. The simplified version should work for both true and false (and nil).

jayRyan24: Hm, then maybe I am not understanding.

If I put should_harness, doesn't that check, like our flags, if something is there or not?
It's just checking if it has a string?

jayRyan24: There's no outright boolean value, so it's saying "if it's got something there, it's true"

jayRyan24: OK, thanks for that stuff, got it sorted.

bescort.lic
```diff
@@ -801,7 +801,7 @@ class Bescort
   def power_walk(pattern)
     fput 'exit' if dead? || Flags['ap-danger']
     if Flags['harness-check']
-      harness_mana(20) unless Flags['harness-check'].include?('effortlessly')
+      harness_mana(['20']) unless Flags['harness-check'].include?('effortlessly')
```
rpherbig: Why is this now a string?

jayRyan24: It's an array?

jayRyan24: Seemed better to make harness_mana use an array rather than check if there was an array coming in and doing something different for it.

jayRyan24: Am I doing it wrong?

jayRyan24: Works for my testing?

rpherbig: An Array can contain integers. If we really want a number, then I think it should be a number. This happens to work because all we do is some string formatting, but would break if we did any math later (see discern logic).

rpherbig: There is a separate conversation to be had around whether or not this should take an array. Specifically, I'm a bit leery around changing the signature of existing `common-*` methods since they can cause problems for people running multiple characters.

jayRyan24: Yeah, but it seems pointless to have to check for an array when we can just use it, seems better to just say "restart dependency' than to put it in in a clunky way.

jayRyan24: I see what you mean by number/string, my mistake, duh.

PR #2147, opened 2017-11-28
Fix bugged pilgrim's badge purchase method

charleyc-dr:
Local variable `description` was lost in the creation of `pilgrimage.lic`, so this never worked.


rpherbig:
I blame the original author.

Nice catch!

PR #2150, opened 2017-11-29
What kind of cleric even

charleyc-dr:


Sarvatt:
Ranger screws up the ranger gods name, go figure! Thanks for fixing it :)

PR #2153, opened 2017-11-29
Added an s to flower so the inlcude? works. Also added the s to the ba…

Dartellum:
…se recipes missing it. This change was necessary due to it not working for Telarann, and after testing on Anshelm (my remedy guy), I found this true. The changes correct it and make it work; Telarann tested as well.


rcuhljr:
This isn't going to work, `exists?` doesn't function like that, you need to give exists a string to tap against. it can't tap `red flower || blue flower`

PR #2155, opened 2017-11-29
Add script to transfer items from one container to another.

Sarvatt:


Sarvatt:
I guess common isn't even needed here

transfer-items.lic
```diff
@@ -0,0 +1,48 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#transfer-items
+=end
+
+custom_require.call(%w[common])
```
rpherbig: Is this needed?

transfer-items.lic
```diff
@@ -0,0 +1,48 @@
...
+
+class ItemTransfer
+  include DRC
```
rpherbig: Is this needed?

transfer-items.lic
```diff
@@ -0,0 +1,48 @@
...
+    fput("look in my #{@source}")
+
+    while line = get
+      if line =~ /In the .*#{@source} you see/
+        items = line
+        break
+      end
+    end
+
+    items = items.sub(' and ', ',')
```
rpherbig: What do you think about something like this? (in place of lines 34-44)

```
    items
      .sub(' and ', ',')
      .split(',')
      .map { |full_name| full_name.split(' ') }
      .map { |parts| parts.last.sub('.', '') }
      .each do |item|
        fput("get #{item} from my #{@source}")
        fput("put #{item} in my #{@destination}")
      end
```

Sarvatt: that works great actually, thank you for the nicer way to do it

rpherbig:
Please update https://elanthipedia.play.net/Lich_script_repository#transfer-items when you get a chance

Sarvatt:
done, thank you

PR #2156, opened 2017-11-30
Make the time between cycling armors configurable with cycle_armors_time

Sarvatt:


Sarvatt:
I would like to increase the default time here, but kept it functionally the same as it is now for this pull request and will send another to paint that bikeshed another color.

combat-trainer.lic
```diff
@@ -19,7 +19,9 @@ class SetupProcess
     echo("  @priority_defense: #{@priority_defense}") if $debug_mode_ct
     @cycle_armors = settings.cycle_armors
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
-    @last_cycle_time = Time.now - 125
+    @cycle_armors_time = settings.cycle_armors_time || 125
```
Sarvatt: the || 125 will be removed in a followup, I just wanted to be extra safe for now to not break people.

Sarvatt: actually it should probably go now? shouldn't be needed

rpherbig:
Do you also need to change `combat-trainer`, line 106?

Sarvatt:
What a doofus, thanks for catching that!

PR #2157, opened 2017-11-30
base-crafting.yaml: Switch all shaping to use maple instead of balsa.

Sarvatt:
Workorders pay better for maple, and shaping is not hardcoded to use balsa because it is 10 pieces (aka the max pieces any of the previous recipes would have needed) anymore now that it supports buying enough of any wood to complete any recipe. 80 workability for maple vs 85 for balsa is insignificant. This will add a minor few silver extra cost to non workorders if they use a recipe that needs more than 5 pieces but workorder profits will more than make up for that.

### DIFF
--- a/data/base-crafting.yaml
+++ b/data/base-crafting.yaml
@@ -421,9 +421,9 @@ shaping:
     repair-room: 19209
     repair-npc: Rangu
     stock-room: 8864
-    stock-number: 11
-    stock-name: balsa
-    stock-volume: 10
+    stock-number: 10
+    stock-name: maple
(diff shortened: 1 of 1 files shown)


rpherbig:
Could you test this with a circle 1 with storebought tools to make sure the 80 vs 85 isn't a big deal, or are you pretty confident already?

Sarvatt:
just did 3 maple nose ring workorders with 0 ranks (well 4 ranks by the end) on a f2p commoner with none too low to go in the logbook, I'm confident it's safe.

PR #2159, opened 2017-11-30
sew.lic: allow setting a material type for cloth

### LINKED ISSUE #2065: sew does not take the material into account for cloth
it hardcodes "get my cloth" and such in places, which can grab chamois cloths instead of crafting material.

Sarvatt:
default to burlap if not provided.
Fixes #2065

### DIFF
--- a/sew.lic
+++ b/sew.lic
@@ -26,7 +26,7 @@ class Sew
         { name: 'chapter', regex: /\d+/i, variable: true, description: 'Chapter containing the item.' },
         { name: 'recipe_name', display: 'recipe name', regex: /^[A-z\s\-\']+$/i, variable: true, description: 'Name of the recipe, wrap in double quotes if this is multiple words.' },
         { name: 'noun', regex: /\w+/i, variable: true },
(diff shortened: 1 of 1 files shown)


Sarvatt:
Tested with `;sew stow sewing 3 "a cloth backpack" backpack` with a cham cloth as my "get cloth" item and burlap in a crafting bag. And also with  `;sew stow sewing 3 "a cloth backpack" backpack wool` with wool in a crafting bag and a cham cloth at the top of the get cloth list.

rcuhljr:
So what happens if I have heavy wool cloth, fine wool cloth, and heavy silk cloth?

Sarvatt:
it just goes by the material, wool cloth so that's something you'd have to sort out beforehand if you wanted to use a specific one like heavy, no different than now.

Sarvatt:
ahh you mean it would work for any cloth before, gotcha

Sarvatt:
Closing this and the other, I'm going to make the material type non-optional which will require the base update in the workorders PR.

PR #2160, opened 2017-11-30
workorders: pass cloth type to sew, define cloth name in base

Sarvatt:
Also allow sewing workorders from chapter 3 (bags and accessories), and discard the proper cloth. info['stock-name'] was not defined.


Sarvatt:
This needs to go in after https://github.com/rpherbig/dr-scripts/pull/2159 to be functional, but no harm if not. material type would just get ignored.

PR #2165, opened 2017-12-02
Adding support for articulated tongs

Onbusi:
Adding support in forge.lic for articulated/clockwork tongs.  Put use_tongs_as_shovel: false in base.yaml

### DIFF
--- a/forge.lic
+++ b/forge.lic
@@ -295,26 +295,27 @@ class Forge
 
   def add_fuel
     if @articulated
-	  case bput('adjust my tongs', 'lock the tongs into a fully extended', 'You have no idea how you')
-	  when 'You have no idea how you'
-	    echo '*** YOU DO NOT HAVE ARTICULATED TONGS ***'
-		exit
+      case bput('adjust my tongs', 'lock the tongs into a fully extended', 'You have no idea how you')
+      when 'You have no idea how you'
+        echo '*** YOU DO NOT HAVE ARTICULATED TONGS ***'
+        exit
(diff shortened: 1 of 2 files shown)


Sarvatt:
can you switch to spaces for indent instead of tabs please?

forge.lic
```diff
@@ -293,19 +294,42 @@ class Forge
   end
 
   def add_fuel
-    stow_item('tongs')
```
rpherbig: I wonder if the code would be any clearer if you wrapped the `stow` and `get` here with an `if @articulated` check. Something like:

```
if @articulated
  # do new stuff
else
  # do old stuff
end

# push fuel like the old code

if @articulated
  # put away with new code
else
  # put away with old code
end
```

Onbusi: Alright, I'll work on doing that and testing it today

rpherbig:
@Onbusi are you still working on this?

BinuDR:
If this is still being worked on, could you include clockwork tongs too?

>tap clock tongs
You tap some clockwork-enhanced tongs with an attached shovel head that you are holding.
>adjust tong
You lock the tongs into a fully extended position.  Then you reach alongside the tong's body, and pull free the head of a shovel.
>tap tong
You tap some clockwork-enhanced tongs with an attached shovel head that you are holding.
>adjust tong
With a yank you fold the shovel head back alongside the tong's body.  Then you unlock the arms, readying them for use as tongs once more.

PR #2167, opened 2017-12-02
Sewing fixes.

### LINKED ISSUE #2065: sew does not take the material into account for cloth

Sarvatt:
Add material arg support to sew.lic for tailoring
Make workorders.lic pass material type to sew for tailoring workorders
add sewing and knitting default material types to base-crafting.yaml
add support for chapter 3 in the tailoring book, cloth containers and accessories
Fixes #2065

### DIFF
--- a/data/base-crafting.yaml
+++ b/data/base-crafting.yaml
@@ -301,6 +301,7 @@ tailoring:
     sew-stock-number: 8
     sew-stock-name: burlap
     knit-stock-number: 13
+    knit-stock-name: wool
     pattern-book: tailoring
     idle-room:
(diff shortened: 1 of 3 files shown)


Sarvatt:
argh, this wont work either without defining yarn too, will reopen when fixed.

rpherbig:
@Sarvatt it looks like one more place needed updated: https://github.com/rpherbig/dr-scripts/blob/master/crossing-training.lic#L1024

PR #2171, opened 2017-12-03
Update sanowret-crystal.lic

Indras1:
If there is another sanowret crystal in the room with you (i.e. on a table in a player house), "gaze sanowret crystal" will fail.  Adding "my" qualifier fixes the issue.

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -23,9 +23,9 @@ class SanowretCrystal
     return if hiding?
 
     if DRSkill.getxp('Arcana') <= 10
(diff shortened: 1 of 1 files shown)


Sarvatt:
what happens if someone named Myth is in the room?

Sarvatt:
i've been hesitant to push this change because i'd rather stop it in a house than spam gaze my repeatedly at someone in a public room

rpherbig:
Can we match on the "gaze at a person" text and do something different?

Sarvatt:
ahh it wouldn't gaze at the person at least, just give gaze help messaging every 10 seconds. but yeah will look into fixing that
J>gaze embrace sanowret crystal
Usage: GAZE <character|creature|item|direction>
       GAZE MY <feet|foot|fingers|hands|nose|arm|leg>
       GAZE AT <character> [IN] <type>
               Where type can be: mournful, happy, sadly, grim,
               worship, dreamy, thought, timid, hope, sympathetic,
               silent, defiantly, awe, wonder, admiration, amazement,
               longingly and fondly.

PR #2179, opened 2017-12-05
base.yaml: Change theurgy to stop at 33/34 instead of running all rit…

Sarvatt:
…uals.

It runs well past mindlock otherwise.


rpherbig:
@charleyc-dr thoughts on this?

charleyc-dr:
This is good with me.  I'd like to make the change that by default `;theurgy` stops running communes (but keeps doing rituals) once we're mindlocked, but either way stopping completely at 33/34 seems like the right default behavior.

PR #2181, opened 2017-12-05
Add Shard Gates to Bescort

jandersson:
This update will let us use the Shard gates for citizens of Ilithi. You need to be a citizen of Ilithi to use the gates at night. If you are a citizen and the gates are closed, you can knock on them to pass through. You can't be invisible when you do this. 

The west, south, and north gates will get updates to use the following stringprocs: 
```
timeto = StringProc.new("if Script.exists?('bescort') && UserVars.citizenship == 'Ilithi' then 0.2 else nil end")    
wayto = StringProc.new("start_script('bescort', ['shard_gate']); wait_while{ running?('bescort') }")             
```
We'll use the `citizenship` UserVar to let people set their citizenship, since this is readable by go2.

In the future it would be nice if users could set citizenship via their YAML settings and have that translated into a UserVar. I would be interested in hearing about solutions for doing this. For now it will be a manual setting.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -167,8 +167,8 @@ class Bescort
         { name: 'ferry1', regex: /ferry1/i, description: 'Ferry between Hibarnhvidar and the Ain Ghazal' },
         { name: 'mode', options: %w[hibarnhvidar ainghazal], description: 'Which town are you going to?' }
       ],
-      [ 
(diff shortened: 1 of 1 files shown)


Sarvatt:
Not related to this pull request, but does anyone have thoughts on where a proper place to add a yaml setting to configure this would be in the future?

PR #2182, opened 2017-12-05
Remove an unmappable room with multiple locations from orc scouts hun…

Sarvatt:
…ting area.

There are 2 possible locations to this room and go2 gets stuck in a loop if you end up in it. It is unmappable because you can't peer in the rose thicket to differentiate the rooms.


Sarvatt:
Make that 3 possible locations, still.

PR #2183, opened 2017-12-06
Theurgy restock

charleyc-dr:
A few big changes:
- Moves functions from restock.lic into common-items.lic, and renames to be more generic.  Note that I didn't change restock.lic to use them yet since I'd like to not break everyone.  Once everyone's had a few days to pick up changes to common-items.lic we can move things over.
- Uses the above functions to restock theurgy supplies based on config.  This resolves the last big remaining TODO.  Note that we can't easily use restock.lic due to the need to bless certain items as we go.  Might be possible to merge these together at a later date but this was already feeling ambitious.
- Also adds support for cleaning anloral pins.  Requires looking up the last favor god's aspect to avoid accidentally breaking someone's pin (it breaks if you clean it and your last favor was from a different god).
- Adds holy oil / flint to P5.


theurgy.lic
```diff
@@ -378,7 +442,7 @@ class TheurgyActions
   def perform_commune(commune)
     stow_hands
     # Recheck item presence since supplies may have been used up.
-    return false if commune[:items].any? { |i| !on_hand?(i) }
+    return false if commune[:items].any? { |i| count(i) == 0 }
```
Sarvatt: return false if commune[:items].any? { |i| count(i).zero? }

theurgy.lic
```diff
@@ -625,7 +689,7 @@ class TheurgyActions
 
   def meditate_bead
     return false unless get_prayer_chain?
-    if count_prayer_beads.zero?
+    if count_prayer_beads == 0
```
Sarvatt: why the change from if count_prayer_beads.zero?

theurgy.lic
```diff
@@ -208,28 +223,54 @@ class TheurgyActions
...
-    elsif item[:any_container]
-      exists?(item[:name])
+  def count(item)
+    if item[:stackable]
+      if item[:any_container]
+        count_item_parts(item[:name])
+      else
+        count_item_parts("#{item[:name]} in my #{@theurgy_supply_container}")
+      end
     else
```
Sarvatt: this could be 
elseif item[:any_container]
  count_items(item[:name])
else
  count_items_in_container(item[:name], @theurgy_supply_container)
to shut up rubocop.

rpherbig:
In the future I'd very much prefer to see a single PR with a single logical change. It gets hard to review with so much changing at once.

common-items.lic
```diff
@@ -91,6 +91,39 @@ module DRCI
     result =~ /You tap|You thump/
   end
 
+  def count_item_parts(item)
```
rpherbig: What is the benefit to moving this from `restock` to here, rather than updating `restock` to handle the functionality you need?

theurgy.lic
```diff
@@ -208,28 +223,54 @@ class TheurgyActions
...
 
-  def on_hand?(item)
-    if item[:name] == 'oil'
-      inside_container?('holy oil') || inside_container?('some oil')
-    elsif item[:any_container]
-      exists?(item[:name])
+  def count(item)
+    if item[:stackable]
+      if item[:any_container]
+        count_item_parts(item[:name])
```
rpherbig: This will break for anyone with multiple characters when the script gets updated but `common-items` has not been reloaded. If possible, it would be nice to add in the helper and then change the script a bit later.

theurgy.lic
```diff
@@ -18,9 +15,12 @@ class TheurgyActions
   def initialize(settings)
     @settings = settings
     @hometown = @settings.hometown
-    @data = get_data('theurgy')[@hometown]
+    all_theurgy_data = get_data('theurgy')
+    @data = all_theurgy_data[@hometown]
+    @immortal_to_aspect = all_theurgy_data['immortal_to_aspect']
     @safe_room = @settings.safe_room
     @immortal_aspect = @settings.immortal_aspect
+    @last_favor_aspect = last_favor_aspect
```
rpherbig: Why check this here at `initialize` time, when they may not be doing the `clean_anloral` ritual? I.e. what do you think about moving this down into the ritual function?

data/base-theurgy.yaml
```diff
@@ -1,4 +1,48 @@
...
+  Damaris: panther
+  Tamsine: cat
+  Meraud: wolf
+  Truffenyi: ox
+  Glythtide: ram
+  Chadatru: lion
+  Faenella: wren
+  Hodierna: unicorn
+  Eluned: dolphin
+  "Hav'roth": cobra
```
rpherbig: I don't think you need the `"` around this. Have you tried without?

data/base-theurgy.yaml
```diff
@@ -1,4 +1,48 @@
 ---
+immortal_to_aspect:
```
rpherbig: A similar mapping is done in `commune.lic`: https://github.com/rpherbig/dr-scripts/blob/master/commune.lic#L147-L159. Is there a benefit to updating that script, too?

common-items.lic
charleyc-dr: The problem is that there's a lot of logic involved in deciding which items we need, e.g. we don't need to restock taffelberries if we know GG, and we don't need wine if we're blacklisting that ritual, etc.  It would make sense to make the core behavior of restock into a script that could be called with a "shopping list", but that was beyond my level of ambition here.

theurgy.lic
charleyc-dr: Done.

theurgy.lic
charleyc-dr: Reverted.

theurgy.lic
charleyc-dr: Done.

theurgy.lic
charleyc-dr: Good idea; submitting a separate PR for this.

theurgy.lic
charleyc-dr: Done.

data/base-theurgy.yaml
charleyc-dr: Done; haven't tested the change yet, though.

data/base-theurgy.yaml
charleyc-dr: Done; haven't tested this yet, though.

common-items.lic
```diff
@@ -91,6 +91,39 @@ module DRCI
...
+              echo "ERROR: count_item_parts called on non-stackable item: #{item}"
+              count_items(item)
+            when *match_messages
+              DRC.text2num(Regexp.last_match(1).tr('-', ' '))
+            end
+    waitrt?
+    count
+  end
+
+  def count_items(item)
```
rpherbig: #2192 has been merged

common-items.lic
rpherbig: It sounds like we need to make restock more dynamic, but that's a PR for another day

rpherbig:
Will this PR include updating `restock.lic` to use the new `common-items` functions?

rpherbig:
@charleyc-dr what's left to do here?

charleyc-dr:
I just updated the common methods to agree with new restock; seems to be working fine, so I'll send another PR for that update.  After that I'm happy with this.  I'd prefer to update restock in a separate PR.

PR #2187, opened 2017-12-07
hunting-buddy.lic: Add during: option.

Sarvatt:
This is to run arbitrary scripts only while in combat. For instance, I have one that summons a wolf companion, and watches for triggers to react to so you can automate the process of getting them older and dismisses it when the script is killed. Baby wolves can't deliver items to other people, and it takes around 80 hours for them to get older so this seems like the most fitting place to add it. I'm sure it could be used for other things too.

Usage:
```hunting_info:
- :zone: black_goblins
  args:
  - d1
  stop_on:
  - Small Edged
  - Stealth
  :duration: 50
  before:
  - mech-lore
  - athletics
  during:
  - combat-ranger-companion
  after:
  - athletics

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -60,6 +60,7 @@ class HuntingBuddy
       args = info['args']
       before_actions = info['before']
       after_actions = info['after']
+      during_actions = info['during']
       duration = info[:duration]
       stop_on_skills = info['stop_on']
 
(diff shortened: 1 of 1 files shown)


Sarvatt:
Another use for this would be to run a script that monitors for an enraged tusky coming in the room and prioritizes killing those, and have it quit after hunting.

rpherbig:
@Sarvatt would you be up for contributing your YAML that uses `during` (and hopefully also `combat-ranger-companion.lic`)?

Issue #2088, opened 2017-11-18
Astrology should skip predictions when mind locked


jandersson:
The current behaviour of astrology is to drain all pools during a single run. This seems like a waste. Instead we could check mindstate on each prediction and exit the script when mindlocked or above 30 mindstate.

PR #2195, opened 2017-12-09
Astrology: Stop predicting when nearly mindlocked

jandersson:
This change makes the `predict_all` method return early if Astrology experience is over 30 mindstate. Continuing to predict would be a waste of prediction pools.


jandersson:
This closes #2088

PR #2196, opened 2017-12-09
Astrology: Remove predictions from check_heavens

jandersson:
Removing this lets the predict_all handle all predictions done in this script

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -240,12 +240,6 @@ class Astrology
 
     pause 2
     waitrt?
-
-    observed['pools']
-      .reject { |_skill, value| value.nil? }
-      .reject { |_skill, value| value < @prediction_pool_target }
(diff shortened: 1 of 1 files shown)


astrology.lic
```diff
@@ -240,12 +240,6 @@ class Astrology
 
     pause 2
     waitrt?
-
-    observed['pools']
```
rpherbig: @jandersson Does this mean `observed` is no longer necessary? See line 203.

jandersson: It isn't used anymore, the state of prediction pools is tracked with the `predict all status` command in game.

rpherbig: There's a pretty big block of code to assign to it, could you look into if any of it can be removed?

jandersson: Sure thing

PR #2201, opened 2017-12-11
base-town.yaml: start tracking guildleader room numbers.

Sarvatt:
complete for crossing and riverhaven (except for thieves guild in haven), the other places have what is available and need more mapping to fix but the mapdb has been locked. Help is needed finding the room numbers for the thief guilds outside of crossing.


Sarvatt:
scratch this, new pull request coming changing the formatting

PR #2202, opened 2017-12-11
base-town.yaml: Add more guild leaders for levelup to use.

Sarvatt:


data/base-town.yaml
```diff
@@ -82,16 +82,16 @@ Crossing:
   almsbox:
     id: 741
   guild_leaders:
-    Bard: 7892
+    Bard: 7879
```
rpherbig: Nice catch

data/base-town.yaml
```diff
@@ -199,6 +199,14 @@ Shard:
     inner_room:
   almsbox:
     id: 13143
+  guild_leaders:
```
rpherbig: Would you be up for including the entire list for each city (similar to https://github.com/rpherbig/dr-scripts/commit/d82da1536095c75fb23c1428896efbb516eab40e#diff-e65384bcbc952bdd340d0bf9c06b452cR84) so we can see at a glance which ones are missing?

Sarvatt: Done, also made leth use crossing guildleaders and boar clan use the same as hib ones.

data/base-town.yaml
```diff
@@ -265,7 +279,18 @@ Riverhaven:
   stamina:
     outer_room: 578
     inner_room:
-  # repairs: 8737
```
Sarvatt: deleted this comment because repairs are already defined

PR #2205, opened 2017-12-11
Combat Trainer: Dont train mm attunement if nearly locked

jandersson:
At 1200 attunement ranks this gives 2 mindstates, so stopping at 32 mindstate
seems appropriate


rpherbig:
This should already default to 32: https://github.com/rpherbig/dr-scripts/blob/5a7a327d364f6555cd6900f151835cfe2051289a/profiles/base.yaml#L70

jandersson:
I see it, thanks for catching that

PR #2212, opened 2017-12-13
base-spells.yaml: match any noun for holy foci.

Sarvatt:


jandersson:
Would it make sense to match this on any word?

Sarvatt:
http://rubular.com/r/VB5jIednVI

PR #2219, opened 2017-12-14
validate.lic: add test for cycle_armor names being correct.

Sarvatt:
--- Lich: validate active.
  Validating settings for 1 settings files
  Checking 50 different potential errors in file Sarvatt-setup.yaml
[validate: ERROR:< Skill name in cycle_armors is not valid: Chain  >]
  WARNINGS:0 ERRORS:1
  All done!
--- Lich: validate has exited.

### DIFF
--- a/validate.lic
+++ b/validate.lic
@@ -88,6 +88,14 @@ class DRYamlValidator
     echo("Valid skills are #{@valid_weapon_skills}")
   end
 
+  def assert_that_cycle_skills_named_properly(settings)
+    return unless settings.cycle_armors
+
+    settings.cycle_armors
+            .reject { |skill| ['Light Armor', 'Chain Armor', 'Brigandine', 'Plate Armor'].include?(skill) }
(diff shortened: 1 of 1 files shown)


validate.lic
```diff
@@ -88,6 +88,14 @@ class DRYamlValidator
     echo("Valid skills are #{@valid_weapon_skills}")
   end
 
+  def assert_that_cycle_skills_named_properly(settings)
```
rpherbig: Maybe `assert_that_cycle_armors_are_skills`?

validate.lic
```diff
@@ -88,6 +88,14 @@ class DRYamlValidator
...
+    return unless settings.cycle_armors
+
+    settings.cycle_armors
+            .reject { |skill| ['Light Armor', 'Chain Armor', 'Brigandine', 'Plate Armor'].include?(skill) }
+            .each_key { |skill| error("Skill name in cycle_armors is not valid: #{skill}") }
```
rpherbig: What do you think about checking that the values (i.e. items) are in the gear list? Too much?

Issue #2223, opened 2017-12-15
elder dillo bescort support - fix included


xtremenerdrage:
Just submitted a pull request

can someone update bescort elder dillos with the following fix - basically since juvies and elders have the same room, it looks for the exit and matches the area based on the exit and continues to exit.
entering is easier since it just looks for paths/trails and enters

starts at line 378

  def desert(mode)
    priority_directions = %w[northeast east north northwest south southeast west southwest]
    if mode !~ /exit/i && Room.current.id != 6760
      echo('Desert script must be started in 6760')
      exit
    end
    case mode
    when /oasis/
      echo 'Not implemented'
      exit
    when /juvenile/
      desert_enter
      find_room_maze
    when /adult/
      desert_enter
      move_direction?(priority_directions) until DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')
      move 'go trail'
      find_room_maze
    when /elder/
	  desert_enter
      move_direction?(priority_directions) until DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')
      move 'go trail'
	  move_direction?(priority_directions) until DRRoom.room_objs.include?('faint path leading to some nearby dunes')
	  move 'go path'
      find_room_maze
      exit
    when /exit/
      if XMLData.room_title.include?('Sand Valley') # Adult Area
        move_direction?(priority_directions) until DRRoom.room_objs.include?('faint trail leading to some nearby dunes')
        move 'go trail'
        desert(mode)
      end
      if XMLData.room_title.include?('High Dunes') # Juvie Area or Elder
        move_direction?(priority_directions) until DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')  || DRRoom.room_objs.include?('faint path that stretches towards a sand-filled valley')
		if DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')
		else
		  move 'go path'
		  desert(mode)
		  exit
		end
        move 'south'
        move 'northeast'
        walk_to(6760)
      end
    end
  end

Issue #2222, opened 2017-12-15
elder dillos


xtremenerdrage:
adult and juvies are there, can we add elders?

  # https://elanthipedia.play.net/Elder_desert_armadillo 700-1100
  elder_armadillos:
    base: 6760
    area: desert
    enter: elder

already in base-hunting

  # https://elanthipedia.play.net/Adult_desert_armadillo 725-1000
  adult_armadillos:
    base: 6760
    area: desert
    enter: adult
  # https://elanthipedia.play.net/Juvenile_desert_armadillo 600-740
  desert_juvenile_armadillos:
    base: 6760
    area: desert
    enter: juvenile

xtremenerdrage:
elder_armadillos:
base: 6760
area: desert
enter: elder

can these be added to base-hunting

PR #2224, opened 2017-12-15
Update bescort.lic elder dillos

xtremenerdrage:
fixes elder dillo travel starting at line 378

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -394,7 +394,12 @@ class Bescort
       move 'go trail'
       find_room_maze
     when /elder/
-      echo 'Not implemented'
+	  desert_enter
+      move_direction?(priority_directions) until DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')
+      move 'go trail'
+	  move_direction?(priority_directions) until DRRoom.room_objs.include?('faint path leading to some nearby dunes')
+	  move 'go path'
+      find_room_maze
       exit
     when /exit/
       if XMLData.room_title.include?('Sand Valley') # Adult Area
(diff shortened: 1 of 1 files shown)


rpherbig:
Does this fix both #2222 and #2223?

xtremenerdrage:
No, 2222 is just an update to base-hunting
2223 is an update to bescort, sorry I didn't include the push to base-hunting in this pull

bescort.lic
```diff
@@ -394,23 +394,35 @@ class Bescort
...
-        move_direction?(priority_directions) until DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')
+      if XMLData.room_title.include?('High Dunes') # Juvie Area or Elder
+        move_direction?(priority_directions) until DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')  || DRRoom.room_objs.include?('faint path that stretches towards a sand-filled valley')
+		if DRRoom.room_objs.include?('faint trail that stretches towards a sand-filled valley')
+		else
+		  move 'go path'
+		  desert(mode)
+		  exit
+		end
         move 'south'
```
rpherbig: Should these three lines be in the `if` block?

xtremenerdrage: Yes, the move until above will stop you in the trail or path room
if 'faint trail...' it goes to move south - Juvies
else 'go path' then restarts the exit script - elders
the exit is because otherwise it loops the end twice 

I tested 
;bescort desert elders then exit 
;bescort desert adult then exit
;bescort desert juvenile then exit

Have also been running 3 hunting zones since I changed, running a stealth script in juvies, backtraining in adult, and mains in elders - hasn't had any issues yet.

I guess more specific logic would be a better standard, but I'm not sure of syntax
move until 'faint trail' || 'faint path'
if 'faint trail' continue
else if 'faint path' do the go path desert mode

Not sure if else if is a thing?

PR #2226, opened 2017-12-16
Adding rezz.lic

skyleronken:
Adding rezz.lic; rubocop completed.

### DIFF
--- a/rezz.lic
+++ b/rezz.lic
@@ -1,0 +1,184 @@
+#   Documentation: https://elanthipedia.play.net/Lich_script_repository#rezz
+
+custom_require.call(%w[common common-arcana events drinfomon spellmonitor])
+
+# rezz waggle set with POM if you want
+
+class Rezz
+  include DRC
+  include DRCA
+
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'person', regex: /\w+/, description: 'Person to rezz' }
+      ]
+    ]
+
+    @rejuv_mana = 0
+    args = parse_args(arg_definitions)
+    @settings = get_settings
+    rezz_prep(args.person)
+  end
+
+  def rezz_prep(person)
+    # cast buffs if set
+    check_buffs
+
+    # perceive body & rejuv
+    rejuv(person) unless is_silver?(person)
+
+    # cast rezz
+    mana = cast_rezz
+
+    # infuse untill see spirit
+    find_soul(person, mana)
+
+    # snap cast soul bonding
+    cast_soulbonding(person)
+
+    # check for silver again. recast rejuv (or infuse again) if not
+    rejuv(person) unless is_silver?(person)
+
+    # gesture
+    bput("gest #{person}", 'As you intone a quiet prayer')
+  end
+
+  def find_soul(person, infuse_amt)
+    loop do
+      pause 5 while mana <= 40
+      harness_mana([infuse_amt]) unless skip_harness?
+      case bput("infuse rezz #{infuse_amt}", /You clearly sense the spirit of #{person} in the Void/i, 'You sense a spirit nearby, but you are unable to make it out clearly.', 'Your sense of sight has not fully transitioned to the spirit world', "You don't have enough harnessed mana to infuse that much.")
+      when 'You sense a spirit nearby, but you are unable to make it out clearly.'
+        next
+      when /You clearly sense the spirit of #{person} in the Void/i
+        break
+      when 'Your sense of sight has not fully transitioned to the spirit world'
+        pause 2
+        next
+      when "You don't have enough harnessed mana to infuse that much."
+        next
+      else
+        next
+      end
+    end
+  end
+
+  def is_silver?(person)
+    /As you concentrate, a thin (.*) nimbus/ =~ bput("perc #{person}", 'As you concentrate, a thin (.*) nimbus')
+    color = Regexp.last_match(1)
+    echo "Memories are #{color}"
+    waitrt?
+    return false unless color.eql? 'silver'
+    true
+  end
+
+  def rejuv(person)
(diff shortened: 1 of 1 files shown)


rpherbig:
> rubocop completed
❤️

rezz.lic
```diff
@@ -0,0 +1,184 @@
...
+    @settings = get_settings
+    rezz_prep(args.person)
+  end
+
+  def rezz_prep(person)
+    # cast buffs if set
+    check_buffs
+
+    # perceive body & rejuv
+    rejuv(person) unless is_silver?(person)
```
rpherbig: What do you think about moving the `is_silver?(person)` check into `rejuv` as a guard clause?

rezz.lic
```diff
@@ -0,0 +1,184 @@
...
+      end
+    end
+    pause 1
+    waitrt?
+    UserVars.discerns[data['abbrev']] = discern_data
+    data['mana'] = discern_data['mana']
+    data['cambrinth'] = discern_data['cambrinth']
+  end
+
+  def calculate_mana(min, more, discern_data, cyclic)
```
rpherbig: Is this copied from `crossing-training.lic`? (which has another copy in `gate.lic`)

rezz.lic
```diff
@@ -0,0 +1,184 @@
...
+        remaining -= step_size
+      end
+    else
+      discern_data['cambrinth'] = nil
+    end
+  end
+
+  def cast_spell(spell, mana, target, snap = false)
+    prepare?(spell, mana, false)
+    snap ? pause(5) : waitcastrt?
```
rpherbig: This is already supported in `common-arcana` with the `prep_time` attribute on a spell. What do you think about using that?

rezz.lic
```diff
@@ -0,0 +1,184 @@
...
+
+  def cast_soulbonding(person)
+    cast_spell('sb', 1, person, true)
+  end
+
+  def check_if_buff_active(spell, time)
+    DRSpells.active_spells[spell] && DRSpells.active_spells[spell] > time
+  end
+
+  # copied from crossing-trianing
```
rpherbig: "training"

rezz.lic
```diff
@@ -0,0 +1,184 @@
...
+      end
+    end
+  end
+
+  def is_silver?(person)
+    /As you concentrate, a thin (.*) nimbus/ =~ bput("perc #{person}", 'As you concentrate, a thin (.*) nimbus')
+    color = Regexp.last_match(1)
+    echo "Memories are #{color}"
+    waitrt?
+    return false unless color.eql? 'silver'
```
rpherbig: You could instead return `color.eql? 'silver'`

rezz.lic
```diff
@@ -0,0 +1,184 @@
...
+    wait_for_script_to_complete('buff', ['rezz'])
+  end
+
+  def skip_harness?
+    attunement = DRSkill.getrank('Attunement')
+    if attunement < 541
+      return false
+    elsif attunement > 600
+      return true
+      elseif check_if_buff_active('Persistence of Mana', 6)
```
rpherbig: Is the duration of PoM based on your Attunement skill? I noticed that if your skill is < 541 it won't check for PoM

rezz.lic
skyleronken: done!

rezz.lic
skyleronken: I have no problem with this. If I recall correctly there was an issue with doing this. I think it was a business rule. Something about yaml settings in common libs. I can move this to common-arcana if that is okay.

rezz.lic
skyleronken: the common-arcana cast_spell function does not support specifying a target to cast against.

rezz.lic
skyleronken: done!

rezz.lic
skyleronken: With rezz you dont need to harness mana before infusing a spell if the following rules apply:
- You have more than 600 attunement
OR
- You have less than 600 but more than 541 attunement AND have POM active.

rezz.lic
skyleronken: hah! fixed

rezz.lic
rpherbig: I think it will work if you `spell['cast] = "cast #{target}"` beforehand

skyleronken: Okay, duh. I made the changes.

rezz.lic
skyleronken: i made the changes. going to test a little. Need to find a dead person though lol.

rpherbig: Ah, you're right. It's the `prep_scaling_factor` and `cambrinth_cap` settings. I think we had much the same conversation in #2099 not too long ago.

Let's leave the copy/paste in for _this_ PR, but I would very much like to see the next one be extracting that logic. The catch would be that those two settings have to be passed in as parameters, not pulled out of settings like they are today.

rezz.lic
rpherbig: Got it, thank you

rezz.lic
skyleronken: Im still going to test adding to common-arcana. I followed the same methodology in other common-arcana functions that require settings but providing it as an argument to the function. So you can call check_discern(data, settings) now which in turn will pass to calculate_mana.

jandersson: Took care of the commonizing of discern, the signature is as you mention
`data = check_discern(data, settings)`

skyleronken:
@rpherbig latest commit includes all of the discussed changes

However, it is dependent upon @jandersson DRCA discerns PR which is pending its co-dependent cast_spell update. Once thats resolved this should be good to g.

PR #2228, opened 2017-12-16
Add flag for whether the character knows rezz to skill-recorder.lic

charleyc-dr:
Needed for deciding whether to take the back entrance into the abyss.


rpherbig:
Could something similar be done for GG and Bless to simplify `theurgy.lic`?

PR #2239, opened 2017-12-18
Updated failsafe to use same number as ;afk

adamgoerdt:


combat-trainer.lic
```diff
@@ -712,7 +712,7 @@ class SafetyProcess
 
   def execute(game_state)
     custom_require.call('tendme') if bleeding? && !Script.running?('tendme')
-    fput 'exit' if health < 40
+    fput 'exit' if health < get_settings.health_threshold
```
rpherbig: It's not as simple as using `get_settings` here because then it won't be using the same settings as the rest of `combat-trainer` (i.e. the script could have been executed with args). You'll want to pass the _right_ settings into `SafetyProcess` (see line 2154).

adamgoerdt:
In this case I think it's fine because people don't override
health_threshold in their hunt yamls

rpherbig:
We don't really know if people do or not. I could see someone wanting to change the threshold for a safer or riskier hunt.

Sarvatt:
Can we push this and make an issue to extend it so it works in every yaml? This would be a great feature to have and it has been asked for.

PR #2240, opened 2017-12-18
Added ability to set priority weapons in a hunt

adamgoerdt:


combat-trainer.lic
```diff
@@ -138,7 +138,19 @@ class SetupProcess
       echo('skipping summoned weapons because no moonblade available') if $debug_mode_ct
       weapon_training = weapon_training.reject { |skill, _| game_state.summoned_info(skill) }
     end
-    new_weapon_skill = weapon_training.min_by { |skill, _| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }.first
+
+    new_weapon_skill=""
+    echo "Priority weapons: #{get_settings.priority_weapons}" if $debug_mode_ct
+    get_settings.priority_weapons.each do | weapon |
```
rpherbig: Same problem as #2239 re: use of `get_settings`.

combat-trainer.lic
```diff
@@ -138,7 +138,19 @@ class SetupProcess
...
+      if DRSkill.getxp(weapon)==0 && weapon_training[weapon]
+        new_weapon_skill=weapon
```
rpherbig: Should it `break` after finding a priority weapon? I think this would make the _last_ item in the last the _highest_ priority, which seems backwards.

combat-trainer.lic
```diff
@@ -138,7 +138,19 @@ class SetupProcess
       echo('skipping summoned weapons because no moonblade available') if $debug_mode_ct
       weapon_training = weapon_training.reject { |skill, _| game_state.summoned_info(skill) }
     end
-    new_weapon_skill = weapon_training.min_by { |skill, _| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }.first
+
+    new_weapon_skill=""
+    echo "Priority weapons: #{get_settings.priority_weapons}" if $debug_mode_ct
+    get_settings.priority_weapons.each do | weapon |
+      if DRSkill.getxp(weapon)==0 && weapon_training[weapon]
```
rpherbig: You could use method chaining instead of putting these conditionals in the `.each`

combat-trainer.lic
```diff
@@ -138,7 +138,19 @@ class SetupProcess
       echo('skipping summoned weapons because no moonblade available') if $debug_mode_ct
       weapon_training = weapon_training.reject { |skill, _| game_state.summoned_info(skill) }
     end
-    new_weapon_skill = weapon_training.min_by { |skill, _| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }.first
```
rpherbig: I'm pretty sure there's a way to use `.sort` here to make this code a lot simpler. Something like "sort by the index of the weapon skill in `priority_weapons`" (and account for skills that aren't in the priority list), lower is better.

combat-trainer.lic
```diff
@@ -138,7 +138,19 @@ class SetupProcess
...
+
+    new_weapon_skill=""
+    echo "Priority weapons: #{get_settings.priority_weapons}" if $debug_mode_ct
```
rpherbig: This should be in `initialize`

adamgoerdt: I wouldn't think so.  new_weapon_skill should be scoped locally to the method, not to the class

combat-trainer.lic
adamgoerdt: I'm not sure how to do that.  Can you demonstrate?

combat-trainer.lic
rcuhljr: He means the echo, you're echoing out the fact that priority weapons is set every time that determine next to train is called, instead of once when the setup class is initialized.

combat-trainer.lic
jayRyan24: I think he means something like this? 

```new_weapon_skill = get_settings.priority_weapons.select{|weapon| weapon_training[weapon] && DRSkill.getxp (weapon)}```

Comment (jayRyan24) [reply combat-trainer.lic]: ```array.flatten.compact.uniq``` type of thing

Comment (rpherbig) [reply combat-trainer.lic]: Something like:
```
    get_settings.priority_weapons
    .select { |weapon| DRSkill.getxp(weapon)==0 }
    .select { |weapon| weapon_training[weapon] }
    .each { | weapon | new_weapon_skill=weapon }
```

Although you'll want to account for some of the other comments (like [this](https://github.com/rpherbig/dr-scripts/pull/2240/files/66a604e2cde436ffa059594c2c4a250115c76b13#r157391826) and [this](https://github.com/rpherbig/dr-scripts/pull/2240/files/66a604e2cde436ffa059594c2c4a250115c76b13#r157391904)).

rpherbig:
Has this been overtaken by #2684 ?

sungobund:
Yes, I implemented all of the features this PR set out to do in #2684

jayRyan24:
From what @dr-leustyin says, #2684 doesn't meet the requirements he set out. The training is slightly different as in it doesn't prioritize 0's first and pull them out immediately. I don't know the details, maybe he can chime in.

dr-leustyin:
I looked again at 2684 and this does exactly what I was trying to do, only better.  This PR can be terminated

PR #2241, opened 2017-12-18
Barb buffing is smarter.com

### LINKED ISSUE #2172: barb meditations will repeat if already up
Barb meditations will repeat over and over if already active
example, if Med tenacity is already up you get this message when it does Med tenacity

```
[combat-trainer]>med tenacity
You begin to meditate upon the chakrel amulet, your inner fire swelling as you center your mind, body, and spirit.
Roundtime: 11 sec.

Peering into the flame you visualize swords, teeth and worse all rending your flesh.

You draw the flames close and bring forth a physical hardening to your body.

Your concentration slowly dissolves, leaving you vulnerable to physical harm.
The flame infuses your flesh, reinforcing it against physical damage!
You feel a jolt as your vision snaps shut.

[combat-trainer]>med tenacity
You begin to meditate upon the chakrel amulet, your inner fire swelling as you center your mind, body, and spirit.
Roundtime: 7 sec.
```

The jolt shows that it's already up, but the expired message also comes up, causing it to med tenacity again.

jayRyan24:
Changed the buffs to flag for rebuffing if the buff isn't active upon CT start up, makes it smarter and allows meditations to be active while going into combat. This fixes #2172. Also added in the ability to start multiple meditations in a row, before they would overlap and cause issues.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1405,7 +1405,7 @@ class AbilityProcess
     @barb_buffs.each do |name|
       ability_data = get_data('spells').barb_abilities[name]
       Flags.add("ap-#{name}-expired", ability_data['expired_message'])
-      Flags["ap-#{name}-expired"] = true
+      Flags["ap-#{name}-expired"] = true unless DRSpells.active_spells[name]
       echo "setup_barb_buff_flags:: adding #{Flags["ap-#{name}-expired"]} flag with an expiration message of: #{ability_data['expired_message']}" if $debug_mode_ct
     end
   end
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -1405,7 +1405,7 @@ class AbilityProcess
     @barb_buffs.each do |name|
       ability_data = get_data('spells').barb_abilities[name]
       Flags.add("ap-#{name}-expired", ability_data['expired_message'])
-      Flags["ap-#{name}-expired"] = true
+      Flags["ap-#{name}-expired"] = true unless DRSpells.active_spells[name]
```
rpherbig: What will this do if you don't have `med power` or Powermonger?

jayRyan24: Fuck Barbarian skills.

jayRyan24: Yeah, tested this, you need at least Power to see spells being active. What do you think could be done to handle this then?

jayRyan24: PM won't do anything for this, the person just needs POWER. Actually, I don't think this is a big deal, it just won't see the spell as active and it will use it on startup. People that can use Meditations in combat 100% have Power at least, and it will function the same as it did before. The skill will attempt to be used and since it's already active, it will just reset the flag. So, actually this isn't an issue.

rpherbig: I agree with all of the above

PR #2245, opened 2017-12-20
Update ranger-companion.lic

Indras1:
Added some raccoon-specific messages, specifically fixing an issue where:
"A baby raccoon stands up then paces back and forth nervously."

would trigger a "pet wolf" but not a "pet raccoon."  My raccoon tends to sit all the time on his own, so he often stands up to pace, triggering me to try to pet the nearest wolf.

### DIFF
--- a/ranger-companion.lic
+++ b/ranger-companion.lic
@@ -10,8 +10,7 @@ class Companion
   include DRC
 
   def initialize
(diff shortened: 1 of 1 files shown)


Sarvatt:
looks good to me and doesn't break wolves, please pull this :)

PR #2250, opened 2017-12-24
Lower the default prep_scaling_factor in the sample yamls.

Sarvatt:
Backfiring when starting out with the default 98% discerns is a very common issue on lnet.

### DIFF
--- a/profiles/SampleBard-setup.yaml
+++ b/profiles/SampleBard-setup.yaml
@@ -120,6 +120,7 @@ training_spells:
     cyclic: true
   Augmentation:
     abbrev: ECRY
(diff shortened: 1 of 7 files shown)


Sarvatt:
Not sure if this is wanted and happy to just close it, but it's one issue I'm tired of coming up when people start out. 98% is too high for any caster I've played.

rpherbig:
I've created a MU of every class but Empath and Trader and the default worked for me. That said, I'm not opposed to changing it.

I gave a quick glance at the checked in YAMLs and `0.95` seems to be pretty common. What is your reasoning behind using `0.85`?

PR #2251, opened 2017-12-25
bescort.lic: Hide when taking transportation.

Sarvatt:
Thieves need to point you out to steal, at least this will stop anonymous theft.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -367,9 +367,11 @@ class Bescort
 
     case bput('go barge', 'What were you referring to', "One of the barge's crew members watching", "You can't do that right now")
     when /What were you referring to/, /You can't do that right now/
+      hide?
       waitfor 'A sand barge pulls'
       take_sandbarge(start_location, end_location)
     when /One of the barge's crew members watching/
+      hide?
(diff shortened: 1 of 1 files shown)


rpherbig:
What do you think about making a helper to do this? That will help us remember the next time something gets added to `bescort`.

Sarvatt:
Do you think that's necessary? Every new entry is a cut and paste of a previous one from what it looks like, they would pick up the hide :)

jandersson:
It would be nice to have a generic ride method, if it doesn't make it too rigid

Sarvatt:
I'll just close this then, not up for that

jandersson:
Well i meant in general. I think making a generic ride method is another PR

rpherbig:
I'm not opposed to merging this if it helps people out.

Sarvatt:
I was running this for a while and now that I stopped I got stolen from waiting for a ferry and couldn't join it so things just hung up, would still like this to get pulled if possible.

jandersson:
This would be nice for caravan trading as well

PR #2252, opened 2017-12-25
Base-Spells: Add trader magic

jandersson:


jandersson:
Adding spells as I go. Will leave a message when done

jandersson:
All spells (except metamagic/metaspells) added

jandersson:
Going through the spells now and updating cast messaging

data/base-spells.yaml
```diff
@@ -1559,6 +1647,11 @@ spell_data:
     abbrev: TR
     mana: 5
     cyclic: true
+  Turmar Illumination:
+    abbrev: TU IL
```
rpherbig: https://elanthipedia.play.net/Turmar_Illumination says the abbrev here is `TURI`. Could you double check this in game?

bmcneff: Validated that the abbrev is TU IL

bmcneff: Technically both work:

>prep TURI 5
You raise your hands skyward, chanting the equation of the Turmar Illumination spell.

jandersson: Thanks for that. We should probably use the official abbreviation

PR #2254, opened 2017-12-25
Updating missing remedies item in base-recipies

Firekast:
Adding missing recipe "some neck tonic". It is only one of two difficulty 7 - challenging recipes where all ingredients are sold by the society. From the order of the remedies recipes, it looks like everything was just added in order from the Wiki, but this one was just skipped.

This had originally blocked having a viable workorder at any difficulty at 601 alchemy with no techniques. Stopgap is to switch to "easy" workorders for a time with this being the only option available.

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -3849,6 +3849,15 @@ crafting_recipes:
   container: bowl
   herb1: yelith
   herb1_stock:
+- name: some neck tonic
+  noun: tonic
+  volume: 1
+  type: remedies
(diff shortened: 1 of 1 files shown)


rpherbig:
Have you run `remedy.lic` with this recipe?

Firekast:
It ran through 20 times successfully over a few hours of training-manager (the only easy workorders I've done, easy to track). Each time it asked for easy remedies work until it got the new recipe, then completed it successfully and continued onto the next.

PR #2256, opened 2017-12-26
Added box loot tracking

skyleronken:
will track box looting. 
- Counts pet boxes and non-pet boxes at beginning of combat-trainer.
- If box limits are defined in yaml, will track each looted box.
- Once limit is set, will stop looting boxes.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -431,7 +431,7 @@ class LootProcess
     items_to_loot = []
     @lootables
       .select { |item| game_state.lootable?(item) }
-      .reject { |item| @box_nounts.include?(item) && @box_loot_limit && @current_box_count >= @box_loot_limit }
+      .reject { |item| @box_nouns.include?(item) && @box_loot_limit && @current_box_count >= @box_loot_limit }
       .each do |item|
         item_reg = item.split.join('.*')
         matches = DRRoom.room_objs.grep(/\b#{item_reg}$/)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -333,6 +333,15 @@ class LootProcess
     echo("  @last_rites: #{@last_rites}") if $debug_mode_ct
     @last_rites_timer = Time.now - 600
 
+    if settings.box_loot_limit
+      @stop_hunt_on_box_limit = settings.stop_hunt_on_box_limit
+      @box_nouns = get_data('items').box_nouns
+      @box_loot_limit = settings.box_loot_limit
+      @current_box_count = get_boxes(settings.picking_box_source).size + get_boxes(settings.picking_pet_box_source).size
+      echo("  @current_box_count: #{@current_box_count}")
+      echo("  @box_loot_limit: #{@box_loot_limit}")
+    end
+
     Flags.add('using-corpse', 'begins arranging', 'completes arranging', 'kneels down briefly and draws a knife', 'cruelly into the body and carving out a chunk', 'makes additional cuts, purposeful but seemingly at random')
     Flags.add('pouch-full', 'You think the .* pouch is too full to fit another gem into', 'You\'d better tie it up before putting')
     Flags.add('container-full', 'There isn\'t any more room')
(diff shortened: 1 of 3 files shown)


combat-trainer.lic
```diff
@@ -374,12 +383,18 @@ class LootProcess
...
+    case bput("stow #{item}", 'You pick up', 'You get', 'You need a free hand', 'There isn\'t any more room', 'You just can\'t', 'push you over the item limit', 'You stop as you realize the .* is not yours', 'Stow what', 'already in your inventory', 'The .* is not designed to carry anything') 
+    when 'already in your inventory'
       if @gem_nouns.include?(item)
         bput('stow gem', 'You pick up', 'You get', 'You need a free hand', 'You just can\'t', 'push you over the item limit', 'You stop as you realize the .* is not yours', 'Stow what', 'already in your inventory')
       else
         bput("stow other #{item}", 'You pick up', 'You get', 'You need a free hand', 'You just can\'t', 'push you over the item limit', 'You stop as you realize the .* is not yours', 'Stow what', 'already in your inventory')
       end
+    when 'You pick up','You get'
+      if @box_loot_limit and @box_nouns.include?(item)
```
rpherbig: We prefer `&&` to `and`

combat-trainer.lic
```diff
@@ -423,6 +438,7 @@ class LootProcess
         tried_loot ||= !matches.empty?
         matches.each { |_| items_to_loot.push(item); }
       end
+    items_to_loot = items_to_loot - @box_nouns if @box_loot_limit and @current_box_count >= @box_loot_limit
```
rpherbig: You could instead add `.reject { |item| ... }`  to the function chain on line 433.

profiles/base.yaml
```diff
@@ -183,7 +183,7 @@ cambrinth_items:
 loot_additions:
 loot_subtractions:
 loot_specials:
-
```
rpherbig: You need to add `stop_hunt_on_box_limit`

combat-trainer.lic
```diff
@@ -333,6 +333,15 @@ class LootProcess
     echo("  @last_rites: #{@last_rites}") if $debug_mode_ct
     @last_rites_timer = Time.now - 600
 
+    if settings.box_loot_limit
+      @stop_hunt_on_box_limit = settings.stop_hunt_on_box_limit
+      @box_nouns = get_data('items').box_nouns
+      @box_loot_limit = settings.box_loot_limit
+      @current_box_count = get_boxes(settings.picking_box_source).size + get_boxes(settings.picking_pet_box_source).size
+      echo("  @current_box_count: #{@current_box_count}")
```
rpherbig: Please also echo `stop_hunt_on_box_limit`

combat-trainer.lic
```diff
@@ -333,6 +333,15 @@ class LootProcess
     echo("  @last_rites: #{@last_rites}") if $debug_mode_ct
     @last_rites_timer = Time.now - 600
 
+    if settings.box_loot_limit
+      @stop_hunt_on_box_limit = settings.stop_hunt_on_box_limit
```
rpherbig: I don't see where `stop_hunt_on_box_limit` is used, can it be removed?

combat-trainer.lic
```diff
@@ -333,6 +333,15 @@ class LootProcess
     echo("  @last_rites: #{@last_rites}") if $debug_mode_ct
     @last_rites_timer = Time.now - 600
 
+    if settings.box_loot_limit
```
rpherbig: I _think_ the `if` is unnecessary. In the 'missing' case, we assign them to `nil` which is what they would be if unassigned.

combat-trainer.lic
skyleronken: done

combat-trainer.lic
skyleronken: I think that would be better. trying this:
      .reject { |item| @box_nounts.include?(item) && @box_loot_limit && @current_box_count >= @box_loot_limit }

Will test.

combat-trainer.lic
skyleronken: ah, thats for the stopping functionality. removing it.

combat-trainer.lic
skyleronken: might be weird, but id rather it not do the rummaging if this setting isnt being used.

skyleronken:
Done testing. This appears to work fine with the changes. 

Ready for merge.

combat-trainer.lic
```diff
@@ -333,6 +333,14 @@ class LootProcess
...
+      @box_nouns = get_data('items').box_nouns
+      @box_loot_limit = settings.box_loot_limit
+      @current_box_count = get_boxes(settings.picking_box_source).size + get_boxes(settings.picking_pet_box_source).size
```
rpherbig: If these two sources are the same container, will this double your number of boxes?

skyleronken: hmm good catch.

profiles/base.yaml
skyleronken: no. removed.

skyleronken:
@rpherbig Made a change to ensure we don't double count. 

Also added hunting-buddy option to stop a hunt on box limit.

combat-trainer.lic
```diff
@@ -2186,6 +2188,20 @@ class CombatTrainer
     ]
   end
 
+  def get_process(process)
```
rpherbig: I wonder if this is overly generic... would it be easier to just expose a method like `current_box_count` on `CombatTrainer` and let it look up the count on `LootProcess`?

skyleronken: I personally dont think so. Having a method like this really opens up combat-trainer and allows us to interact with it at a 'super user' level. 

However, having more specific getter style methods is a bit cleaner when analyzing other scripts that may be interacting. 

If your concerned with using the get_process method I can make the helper function.

rpherbig: Maybe my bigger concern is that other scripts would have to know which process to query for what, rather than treating `combat-trainer` as a black box.

Regardless, it's not critical right now. We can wait and see a few use cases before deciding.

PR #2265, opened 2017-12-28
Added orc chiefs as well as better orc descriptions

jayRyan24:
Adding a separate tag with the specific rooms which spawn chiefs, also added better descriptions of the rooms. Shouldn't break anyone currently using **orc_raiders2**.

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -745,6 +745,7 @@ hunting_zones:
   - 8702
   - 8703
   # https://elanthipedia.play.net/Orc_raider 365-450
+  # downstairs mixed with reivers
   orc_raiders:
   - 8704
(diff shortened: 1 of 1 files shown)


data/base-hunting.yaml
```diff
@@ -745,20 +745,25 @@ hunting_zones:
...
   - 8709
+  # upstairs just raiders
   orc_raiders2:
   - 8710
   - 8711
   - 8712
   - 8713
   - 8714
   - 8715
+  # https://elanthipedia.play.net/Orc_clan-chief
```
rpherbig: Could you add the skill range to this comment?

jayRyan24: Added and adjusted comments.

PR #2268, opened 2017-12-28
drinfomon: Check exp all after death

jandersson:
Fixes an error with running circlecheck immediately after dying


jayRyan24:
I would love this same type of thing to go into the afk script logout. Right before logging out, when it alerts Statusmon, it should also do this and maybe ;learned!

PR #2269, opened 2017-12-28
Spellmonitor: require drinfomon

jandersson:
Require drinfomon to avoid an error if spellmonitor is started before drinfomon


Sarvatt:
this is getting ripped out in the next day or so

rpherbig:
Is that once they update Barbs and Thieves?

Sarvatt:
yeah the old code path is being used for thieves/barbs right now and will need to be ripped out as soon as it changes over

PR #2274, opened 2017-12-28
Implements an %INCLUDE directive for YAMLs

charleyc-dr:
Been asked to upload this a few times, so here it is.  Documentation is in generate-profile.lic .


generate-profile.lic
```diff
@@ -0,0 +1,120 @@
+=begin
+This script takes partial YAML from profile-gen/ and resolves all %INCLUDE
```
rpherbig: Do you plan to document this on elanthipedia?

generate-profile.lic
```diff
@@ -0,0 +1,120 @@
...
+directives to produce new YAML files in profiles/ .
```
rpherbig: Now, if we could just wire this into `dependency` and have it run at settings-load time this could all be transparent to the user!

generate-profile.lic
```diff
@@ -0,0 +1,120 @@
...
+      echo("Loading file #{file}") if @debug
+      unless match = /^(#{checkname}|ALL)-(.+?)(-include)?\.yaml$/.match(file)
+        raise ArgumentError, "Bad filename: #{file}"
+      end
+      name = match[2]
+      is_include = !!match[3]
+      if match[1] == 'ALL' && !is_include
+        raise ArgumentError, 'Files starting in ALL- must be includes.'
+      end
+      body = File.readlines("./scripts/profile-gen/#{file}")
```
rpherbig: Does this script need to be trusted?

generate-profile.lic
```diff
@@ -0,0 +1,120 @@
...
+      end
+
+      profile[:resolvable] = profile[:deps].all? do |dep|
+        @profiles[dep] && @profiles[dep][:include]
+      end
+    end
+
+    @profiles.each do |name, prof|
+      if !prof[:resolvable] && !prof[:include]
+        raise ArgumentError, "Profile #{name} contains non-resolvable includes."
```
rpherbig: Can you output _which_ includes were not able to be resolved?

generate-profile.lic
```diff
@@ -0,0 +1,120 @@
...
+
+%INCLUDE foo
+
+See Risima-* in profile-gen for an example usage.
+=end
+
+class ProfileGenerator
+  def initialize(debug)
+    @debug = debug
+    load_profilegen
```
rpherbig: What do you think about returning `profiles` from `load_profilegen` and then passing it into `resolve_includes` and `write_profiles`?

generate-profile.lic
```diff
@@ -0,0 +1,120 @@
...
+  def write_profiles
+    @profiles.values.select { |profile| !profile[:include] }.each do |profile|
+      filename = "./scripts/profiles/#{profile[:file]}"
+      echo("Writing #{filename}") if @debug
+      File.open(filename, 'wb') { |f| f.puts(profile[:src]) }
+    end
+  end
+end
+
+ProfileGenerator.new(true)
```
rpherbig: Does `true` mean this is always in debug mode?

generate-profile.lic
```diff
@@ -0,0 +1,120 @@
+=begin
```
rpherbig: In the past we've used the `/util` directory for these kinds of "doesn't interact with the game" scripts... but I'm not opposed to leaving it here, either.

generate-profile.lic
charleyc-dr: I like this idea, but it's probably too expensive to call this every time we run get_settings unless functionality is added to limit the scope (which could be done).  E.g. if `;generate-profile foo` only parsed the deps for Name-foo.yaml and Name-setup.yaml .

generate-profile.lic
charleyc-dr: Yes, it does.

generate-profile.lic
charleyc-dr: Currently yes.

generate-profile.lic
rpherbig: Agreed. I'm just making a mental note of what I want for Christmas next year.

generate-profile.lic
rpherbig: Could you add a note to the documentation?

rpherbig:
@charleyc-dr how's it going with this PR?

charleyc-dr:
Sorry, been lazy lately, plus tiny Necromancers are very confusing!  Picking this back up now.

generate-profile.lic
charleyc-dr: Done.

generate-profile.lic
charleyc-dr: Done; also added line number to the error message.

generate-profile.lic
charleyc-dr: Done.

generate-profile.lic
charleyc-dr: I wasn't particularly planning on it, but I can if we start getting questions.  Right now I'm considering this an "advanced" option and I'm not sure I want to point people towards it.

generate-profile.lic
charleyc-dr: If this were moved to `/util`, what would the command be to run the script?  `;util/gen`?

Sarvatt: well util/* isn't shipped with dr-scripts unless its a git checkout, so...

charleyc-dr:
This is broken now! Don't submit until I figure out what I messed up, please :)

charleyc-dr:
Ok, fixed.  Missed a `@profiles` in there.

generate-profile.lic
charleyc-dr: Switched this to an arg instead.

charleyc-dr:
At this point I'm happy with the state of this change.  Please LMK if you think any other improvements are in order 😃

generate-profile.lic
rpherbig: Oh, right. Good point - that depends on the target users for this.

PR #2280, opened 2017-12-29
Updated workorders to read volume from deeds

trall541:
workorders will pull the volume from a deed where appropriate instead of using analyze, this removes the minimum appraisal required to use your own ingots and gives a slight optimization in removing analyze RT

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -598,6 +598,11 @@ class WorkOrders
     res.scan(/\d+/).first.to_i
   end
 
+  def deed_ingot_volume
+    res = bput('read my deed', 'Volume:\s*\d+')
+    res.scan(/\d+/).first.to_i
+  end
+
   def forge_items_with_own_ingot(info, item, quantity)
     recipe, = find_recipe(info, item, quantity)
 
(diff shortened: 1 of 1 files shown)


workorders.lic
```diff
@@ -623,11 +630,11 @@ class WorkOrders
         stow_hands
         exit
       else
+		volume = deed_ingot_volume
         fput('tap deed')
         pause
         fput("get #{@use_own_ingot_type} ingot") unless checkleft || checkright
       end
-      volume = ingot_volume
```
rpherbig: Do you need to do another `unless volume` check here?

workorders.lic
```diff
@@ -606,13 +611,15 @@ class WorkOrders
...
       else
+        volume = deed_ingot_volume
         fput('tap deed')
         pause
         fput("get #{@use_own_ingot_type} ingot") unless checkleft || checkright
       end
     end
-
-    volume = ingot_volume
+    if(!volume)
```
rpherbig: What do you think about `volume = ingot_volume unless volume`?

workorders.lic
trall541: this spot will only get here if the item is a deed, so its really not needed at all

workorders.lic
trall541: that works too :)

PR #2282, opened 2017-12-30
Strip special characters from command check

adamgoerdt:


status-monitor.lic
```diff
@@ -130,6 +130,7 @@ class GameFilter
   end
 
   def command_check(line)
+    line = line.gsub(/_|~|-|=|\./,"")
```
rpherbig: I'd prefer to see the `.gsub` in the method chain on the next line. What do you think?

status-monitor.lic
```diff
@@ -119,7 +119,7 @@ class GameFilter
     echo("\a")
     pause 0.25
     command_check(line)
-    fput @responses.first
+    fput @responses.first if get_settings.status_monitor_respond
```
rpherbig: I'd prefer to see the setting stored into an instance var in `initialize`. What do you think?

status-monitor.lic
adamgoerdt: Makes sense.  Updated

PR #2283, opened 2017-12-30
Trade: Add script for caravan trading

jandersson:
This script handles running a caravan route in Zoluren. It will trade one contract at a time, but has the potential to be updated to run fancier routes with several contracts at a time. The script assumes you have a worn feedbag for feeding the animals.

### DIFF
--- a/trade.lic
+++ b/trade.lic
@@ -136,8 +136,7 @@ class Trade
     previous, _shortest_paths = Map.dijkstra(Room.current.id, destination)
     path = [destination]
     path.push(previous[path[-1]]) until previous[path[-1]].nil?
-    path.reverse!
-    return path
+    return path.reverse!
   end
 
   def step(dir)
--- a/trade.lic
+++ b/trade.lic
@@ -1,7 +1,7 @@
 =begin
   Documentation: https://elanthipedia.play.net/Lich_script_repository#trade
 =end
-custom_require.call(%w(common common-items common-money common-travel drinfomon equipmanager))
+custom_require.call(%w[common common-items common-money common-travel drinfomon equipmanager])
 
 class Trade
   include DRC
@@ -28,13 +28,13 @@ class Trade
     end
     contract = parse_contract
     unless contract.presented
-      load_caravan(contract.origin_clerk) 
+      load_caravan(contract.origin_clerk)
       recall_caravan
       walk_to(contract.origin_outpost)
     end
     feed_caravan
     take_caravan_to(contract.destination_outpost)
-    bput("get my contract", 'You get')
+    bput('get my contract', 'You get')
     deliver_contract(contract.destination_clerk)
     pay_dues(contract.destination_clerk)
     deposit_coins(500, Room.current.id) if contract.destination_town == @hometown
@@ -46,7 +46,7 @@ class Trade
 
   def deposit_coins(withdraw_amt, return_to_id)
     return if wealth(@hometown) < withdraw_amt
-    
+
     walk_to(@town_data[@hometown]['deposit']['id'])
     case bput('deposit all', 'you drop all your', 'You hand the clerk some coins', "You don't have any", 'There is no teller here', 'reached the maximum balance I can permit')
     when 'There is no teller here'
@@ -58,9 +58,9 @@ class Trade
 
   def find_closest_id(entity)
     rooms = @town_data.to_h.values.select { |data| data[entity] }
-                             .map{ |data| data[entity]['id'] }
+                      .map { |data| data[entity]['id'] }
     echo "rooms: #{rooms}" if debugging?
-    return sort_destinations(rooms).first
+    sort_destinations(rooms).first
   end
 
   def parse_contract
@@ -70,7 +70,7 @@ class Trade
     contract = OpenStruct.new
     contract.presented = true
     contract.expired = false
-    fput("read my contract")
+    fput('read my contract')
     while line = get
       echo line if debugging?
       if line =~ destination_re
@@ -101,21 +101,21 @@ class Trade
       echo "Contract destination town: #{contract.destination_town}"
       echo "Contracted presented to clerk? #{contract.presented}"
     end
-    return contract
+    contract
   end
 
   def recall_caravan
     recall_messages = [
-        /You seem to recall that you left your (?<description>.*) (?<name>.*) right behind you/,
-        /You recall that your (?<description>.*) (?<name>.*) should be located in/,
-        /You don't recall where you left your caravan\./,
-        /You don't recall having a caravan\./
+      /You seem to recall that you left your (?<description>.*) (?<name>.*) right behind you/,
+      /You recall that your (?<description>.*) (?<name>.*) should be located in/,
+      /You don't recall where you left your caravan\./,
+      /You don't recall having a caravan\./
     ]
-    response = bput("recall caravan", recall_messages)
+    response = bput('recall caravan', recall_messages)
     echo("Response: #{response}") if debugging?
     case response
     when /You don't recall where you left your caravan/, /You don't recall having a caravan/
-      echo("No caravan found") if debugging?
+      echo('No caravan found') if debugging?
     when /You seem to recall that you left your (?<description>.*) (?<name>.*) right behind you/
       /You seem to recall that you left your (?<description>.*) (?<name>.*) right behind you/ =~ response
       @caravan.noun = name
@@ -135,7 +135,7 @@ class Trade
     previous, _shortest_paths = Map.dijkstra(Room.current.id, destination)
     path = [destination]
     path.push(previous[path[-1]]) until previous[path[-1]].nil?
-    return path.reverse
+    path.reverse
   end
 
   def step(dir)
@@ -149,7 +149,7 @@ class Trade
   end
 
   def get_next_dir(path)
-    return Room.current.wayto[path[path.index(Room.current.id)+1].to_s]
+    Room.current.wayto[path[path.index(Room.current.id) + 1].to_s]
   end
 
   def take_caravan_to(outpost_id)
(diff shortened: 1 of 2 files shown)


trade.lic
```diff
@@ -0,0 +1,258 @@
+custom_require.call(%w(common common-items common-money common-travel drinfomon equipmanager))
```
rpherbig: Do you plan to document this on elanthipedia?

jandersson: I do, Ill add a link to the header

trade.lic
```diff
@@ -0,0 +1,258 @@
...
+    end
+    contract = parse_contract
+    unless contract.presented
+      load_caravan(contract.origin_clerk) 
+      recall_caravan
+      walk_to(contract.origin_outpost)
+    end
+    feed_caravan
+    take_caravan_to(contract.destination_outpost)
+    fput("get my contract")
```
rpherbig: Could this use `bput`?

trade.lic
```diff
@@ -0,0 +1,258 @@
...
+  end
+
+  def find_closest_id(entity)
+    rooms = @town_data.to_h.values.select { |data| data[entity] }
+                             .map{ |data| data[entity]['id'] }
+    echo "rooms: #{rooms}" if $debug
+    return sort_destinations(rooms).first
+  end
+
+  def parse_contract(name=nil)
```
rpherbig: I don't see where `name` is ever passed in. Can it be removed?

trade.lic
```diff
@@ -0,0 +1,258 @@
...
+    destination_re = /The guild office at (?:The )?(?<destination>.*) requires .*/
```
rpherbig: Could these be declared once, say at the top of the file? (like we do in `common-healing`)

trade.lic
```diff
@@ -0,0 +1,258 @@
...
+      echo "Contract origin: #{contract.origin_town}"
+      echo "Contract origin clerk room: #{contract.origin_clerk}"
+      echo "Contract origin minister room: #{contract.origin_minister}"
+      echo "Contract destination town: #{contract.destination_town}"
+      echo "Contracted presented to clerk? #{contract.presented}"
+    end
+    return contract
+  end
+
+  def recall_caravan
```
rpherbig: It feels like this should return some sort of Caravan hash (similar to how `parse_contract` works). What do you think?

trade.lic
```diff
@@ -0,0 +1,258 @@
...
+      echo "@caravan_description: #{@caravan_adjective}" if $debug
+    end
+  end
+
+  def compute_path(destination)
+    previous, _shortest_paths = Map.dijkstra(Room.current.id, destination)
+    path = [destination]
+    path.push(previous[path[-1]]) until previous[path[-1]].nil?
+    path.reverse!
+    return path
```
rpherbig: What do you think about returning `path.reverse` instead of using `.reverse!` on the line above?

trade.lic
```diff
@@ -0,0 +1,258 @@
...
+      @caravan_adjective = description
+      echo "@caravan_description: #{@caravan_adjective}" if $debug
+    end
+  end
+
+  def compute_path(destination)
+    previous, _shortest_paths = Map.dijkstra(Room.current.id, destination)
+    path = [destination]
+    path.push(previous[path[-1]]) until previous[path[-1]].nil?
+    path.reverse!
```
rpherbig: Maybe I'm reading this wrong, but does this build the path in reverse order, then reverse it? Can you reference `path[0]` and build it in-order?

trade.lic
```diff
@@ -0,0 +1,258 @@
+custom_require.call(%w(common common-items common-money common-travel drinfomon equipmanager))
+$debug = UserVars.trade_debug
```
rpherbig: This is a global and so accessible outside of this script. I recommend a more unique name than `debug` (for example, `debug_mode_trade`. What do you think?

trade.lic
jandersson: I agree, then i can get rid of most of the instance variables

trade.lic
jandersson: Right, maybe it would make more sense just as a method in the class instead...
```ruby
def debugging?
  return UserVars.trade_debug
end
```

trade.lic
jandersson: I cant think of a good way to build it starting from the source, at least not by looking up the index of the room pointing to it. There might be several indices that point to the starting room for example

```
[test: source: 1005]
[test: destination: 895]
[test: previous[1006]: 1005]
[test: previous.index(1005): 1004]
[test: previous.index(source): 1004]
[test: forward_path: [1005, 1004, 1003, 1001, 1000, 999, 998, 997, 996, 995, 992]]
[test: backward_path: [1005, 1006, 1007, 1642, 1626, 1625, 1623, 1619, 1620, 1621, 806, 805, 807, 808, 809, 810, 770, 771, 772, 862, 905, 906, 904, 907, 903, 902, 895]]
```

jandersson: The last pseudo code entry discusses the same method of reconstructing the path by reverse iteration as well: 

https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm#Pseudocode

rpherbig: 👍

trade.lic
rpherbig: I don't have a preference on that. My gut reaction is that won't make the code clearer, but it could...

trade.lic
```diff
@@ -0,0 +1,262 @@
...
+        /^You don't have any coins on you to pay dues\./,
+        /^You count out some coins, and the clerk notes that your dues are settled\./,
+        /^You count out some coins, and the clerk notes that your dues are now (.*)\./,
+        /^The clerk says, "You do not owe dues, .*!"/
+    ]
+    walk_to(clerk_id)
+    bput("pay clerk", clerk_messages)
+  end
+
+  def feed_caravan
```
rpherbig: This doesn't take an argument (see line 35)

trade.lic
```diff
@@ -0,0 +1,262 @@
...
+      nearest_minister = find_closest_id('trade_minister')
+      get_new_contract(nearest_minister)
+    end
+    contract = parse_contract
+    unless contract.presented
+      load_caravan(contract.origin_clerk) 
+      recall_caravan
+      walk_to(contract.origin_outpost)
+    end
+    feed_caravan(caravan)
```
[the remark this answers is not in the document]
jandersson: Fixing that - strange that it didn't throw an error :thinking:

trade.lic
```diff
@@ -0,0 +1,262 @@
...
+
+  def initialize
+    @town_data = get_data('town')
+    @equipment_manager = EquipmentManager.new
+    settings = get_settings
+    @contract_container = settings.trade_contract_container
+    @hometown = settings.hometown
+    bput("open my #{@contract_container}", 'You open', 'That is already')
+    @equipment_manager.empty_hands
+    @caravan = OpenStruct.new
```
rpherbig: If you're going to loop, this feel like something that should be in `run` instead of `initialize` (so it clears `@caravan` before each run)

trade.lic
```diff
@@ -0,0 +1,262 @@
...
+    Flags.add('caravan-upgrade', /Unfortunately, this load would kill/)
+    bput("give my contract to clerk", /The .* clerk accepts your contract and peruses it\./)
+    recall_caravan if Flags['caravan-upgrade']
+    bput("put my contract in my #{@contract_container}", 'You put')
+    Flags.delete('caravan-upgrade')
+  end
+end
+
+trade = Trade.new
+loop { trade.run }
```
rpherbig: Generally scripts have been designed to run once, with a second script responsible for looping (and stopping, etc). I'm not saying you have to change anything, I'm just not sure it would be intuitive that this loops indefinitely.

trade.lic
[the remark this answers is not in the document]
rpherbig: I think everything references `@caravan` instead, so there wouldn't be any errors. Yay, Ruby!

trade.lic
jandersson: I was originally going to add a TradingBuddy class to this that would manage the start stop based on some user preferences, inspired by hunting buddy. I'll remove this loop and let a future commit handle the looping

trade.lic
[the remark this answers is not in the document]
jandersson: PEBKAC .. (symlinking issue)

trade.lic
```diff
@@ -0,0 +1,259 @@
...
+    UserVars.trade_debug
+  end
+
+  def deposit_coins(withdraw_amt, return_to_id)
+    return if wealth(@hometown) < withdraw_amt
+
+    walk_to(@town_data[@hometown]['deposit']['id'])
+    case bput('deposit all', 'you drop all your', 'You hand the clerk some coins', "You don't have any", 'There is no teller here', 'reached the maximum balance I can permit')
+    when 'There is no teller here'
+      return
```
rpherbig: Do you need to `walk_to(return_to_id)` if you fail to find a teller?

trade.lic
```diff
@@ -0,0 +1,259 @@
...
+
+  def find_closest_id(entity)
+    rooms = @town_data.to_h.values.select { |data| data[entity] }
+                      .map { |data| data[entity]['id'] }
+    echo "rooms: #{rooms}" if debugging?
+    sort_destinations(rooms).first
+  end
+
+  def parse_contract
+    destination_re = /The guild office at (?:The )?(?<destination>.*) requires .*/
```
rpherbig: Do any of these regex's need anchors? I.e. would they pick up a whisper or something?

jandersson: Good call. Ill add some anchors in

trade.lic
jandersson: Yea, that sounds better. It was a copy and paste from sell-loot i believe.

rpherbig:
@jandersson looks good to me. How do you feel?

jandersson:
I think its fine for a first version. :+1:

PR #2285, opened 2018-01-01
truffenyi-commune-quest.lic: Add missing Drogor vision.

Sarvatt:


truffenyi-commune-quest.lic
```diff
@@ -48,6 +48,9 @@ while line = get
   if line =~ /In your vision the waters pull away from the shore, gathering upward in a massive wall/
     fput('pray Drogor')
   end
+  if line =~ /Your vision fades and you see yourself facing a crackling fire next to the shore/
+    fput('pray Drogor')
```
rpherbig: Does Drogor have two visions; is line 49 wrong; or something else?

Sarvatt: The old one was from the wiki, I figured it was safe it leave it in just in case it is still there.

PR #2292, opened 2018-01-04
Added a devour healing routine for necromancers to safe-room. 

jayRyan24:
This happens a bunch at lower circles, which is why I originally added it to my scripts.

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -40,7 +40,7 @@ class SafeRoom
       hometown = town_data[settings.hometown]
       wait_at_empath(hometown['npc_empath']['id'])
       fix_standing
-    elsif settings.necro_safe_room_use_material
+    elsif settings.necro_safe_room_use_devour
       devour_healing(settings)
     end
 
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -40,7 +40,7 @@ class SafeRoom
       hometown = town_data[settings.hometown]
       wait_at_empath(hometown['npc_empath']['id'])
       fix_standing
-    elsif safe_room_necro_use_material
+    elsif settings.necro_safe_room_use_material
       devour_healing(settings)
       return
     end
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -2,13 +2,14 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#safe-room
 =end
 
-custom_require.call(%w[common common-healing common-money common-travel drinfomon events])
+custom_require.call(%w[common common-arcana common-healing common-money common-travel drinfomon events])
 
 class SafeRoom
   include DRC
   include DRCH
   include DRCM
   include DRCT
+  include DRCA
 
   def initialize
     arg_definitions = [
@@ -72,7 +73,7 @@ class SafeRoom
     walk_to(settings.safe_room)
     release_invisibility
     devours = count_items_in_container('material', settings.thanatology['harvest_container'])
-    return unless devours > 0
+    return if devours.zero?
 
     stow_hands
     return if right_hand
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -86,7 +86,7 @@ class SafeRoom
 
       result = bput("get material in my #{settings.thanatology['harvest_container']}", 'You get', 'You are already', 'What were you') unless right_hand
       break if result =~ 'What were you'
-      DRCA.cast_spell(settings.necromancer_healing['Devour'], settings)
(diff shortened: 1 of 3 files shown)


jayRyan24:
Could use feedback on whether this should be an explicit setting, or based on if devour is found as a setting in the yaml. I almost thing it should be an explicit setting.

jandersson:
What about a setting like `necromancer_heal_with_material` ?

jayRyan24:
Smart, lol

jayRyan24:
@jandersson and @izawwlgood, How's this look to you guys? I've used it on Melborne but once you can cast Devour better, it becomes unused. It is an option for lower necros.

safe-room.lic
```diff
@@ -40,6 +40,9 @@ class SafeRoom
       hometown = town_data[settings.hometown]
       wait_at_empath(hometown['npc_empath']['id'])
       fix_standing
+    elsif settings.necro_safe_room_use_material
+      devour_healing(settings)
+      return
```
jandersson: Should this be here? This will skip over give_and_take below

safe-room.lic
```diff
@@ -63,6 +66,31 @@ class SafeRoom
...
+      while mana < 40
+        echo('Waiting on mana...')
+        pause 10
+      end
+
+      result = bput("get material in my #{settings.thanatology['harvest_container']}", 'You get', 'You are already', 'What were you') unless right_hand
+      break if result =~ 'What were you'
+      DRCA.cast_spell(settings.necromancer_healing['Devour'], settings)
+      devours -= 1 unless right_hand
+    end
```
jandersson: Do we want to put material back into the harvest container in case we finish healing with some left in hand?

safe-room.lic
jayRyan24: Removed it, thought give_and_take was something else.

safe-room.lic
jayRyan24: Good call, added it. Though it shouldn't pull out a second if there is already some in hand. I use the `unless right_hand` to avoid getting another out.

jayRyan24:
@rpherbig this should be good to go, I've been using it.

safe-room.lic
```diff
@@ -63,6 +65,33 @@ class SafeRoom
...
   end
 
+  def devour_healing(settings)
+    return unless settings.necromancer_healing['Devour']
+    return unless settings.thanatology['harvest_container']
+    
+    walk_to(settings.safe_room)
+    release_invisibility
+    devours = count_items_in_container('material', settings.thanatology['harvest_container'])
+    return unless devours > 0
```
rpherbig: What do you think about `return if devours.zero?`?

safe-room.lic
```diff
@@ -63,6 +65,33 @@ class SafeRoom
...
+    
+    while need_healing? && devours > 0
+      while mana < 40
+        echo('Waiting on mana...')
+        pause 10
+      end
+
+      result = bput("get material in my #{settings.thanatology['harvest_container']}", 'You get', 'You are already', 'What were you') unless right_hand
+      break if result =~ 'What were you'
+      DRCA.cast_spell(settings.necromancer_healing['Devour'], settings)
```
rpherbig: You need to add `common-arcana` to `custom_require` at the top. If you also add `include DRCA` to the class, you don't need to prefix this function call.

validate.lic
```diff
@@ -355,6 +355,13 @@ class DRYamlValidator
     error("thanatology['ritual_type']: '#{settings.thanatology['ritual_type']}' is invalid, only [preserve, harvest, fetish, cut, dissect, consume, arise, and cycle] are supported at this time.")
   end
```
rpherbig: Is it worth adding a validator for the case in which the user has `thanatology['store'] = false` but `necro_safe_room_use_material: true`?

safe-room.lic
jayRyan24: I feel like if I had done that to begin with, you'd be asking me why not `return unless devours > 0`.

rpherbig: No, my default position is to use `if` over `unless`. It generally reads cleaner (obviously there are exceptions).

validate.lic
jayRyan24: Done, good call.

safe-room.lic
jayRyan24: Done.

safe-room.lic
jayRyan24: Ok, gotcha.

PR #2294, opened 2018-01-04
Drca update

jandersson:
The cambrinth data was being indexed incorrectly. The previous style was looking up the first bit of the first element in the array in some cases. This doesn't throw an error, so it wasn't noticed. Since we're interested in knowing if the cambrinth data is an array of arrays or an array of integers, we can look at the first element each time.

Current versions of ruby use a unified integer format for Fixnum
and Bignum. Ruby 2.0.0 is using Fixnum and Bignum. We should check both for
compatability.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,3 +53,6 @@ Metrics/PerceivedComplexity:
 
 Metrics/ParameterLists:
   Enabled: false
+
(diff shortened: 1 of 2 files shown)


rpherbig:
Were you going to disable the cop for this?

jandersson:
Yes, should I include that in the same PR?

PR #2295, opened 2018-01-04
Added the ability for spells/spell sets to override the new harness. 

jayRyan24:
This will make it better for crossing-training and buff as people can't use it due to nerve damage and leakage, and also want forced max cast buffing. Should also be spread tout to force buffing spells in combat-trainer as well. Currently have it set to force cambrinth use with buff.lic and crossing-training.lic.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1151,10 +1151,10 @@ class SpellProcess
         data = { 'abbrev' => 'heal', 'mana' => @empath_spells['HEAL'].first, 'cambrinth' => @empath_spells['HEAL'][1..-1] }
       end
       @wounds = {}
-      prepare_spell(data, game_state) if data
+      prepare_spell(data, game_state, true) if data
     elsif @empath_spells['VH']
       data = { 'abbrev' => 'vh', 'mana' => @empath_spells['VH'].first, 'cambrinth' => @empath_spells['VH'][1..-1] }
-      prepare_spell(data, game_state)
+      prepare_spell(data, game_state, true)
     end
   end
 
@@ -1167,7 +1167,7 @@ class SpellProcess
     game_state.casting_cfb = true
     game_state.prepare_cfb = false
     data = @necromancer_zombie['Call from Beyond']
-    prepare_spell(data, game_state)
+    prepare_spell(data, game_state, true)
   end
 
   def check_consume(game_state)
@@ -1181,7 +1181,7 @@ class SpellProcess
     data = @necromancer_healing['Devour'] || @necromancer_healing['Consume Flesh']
     game_state.casting_consume = true
     game_state.prepare_consume = false
-    prepare_spell(data, game_state)
+    prepare_spell(data, game_state, true)
   end
 
   def check_offensive(game_state)
@@ -1233,7 +1233,7 @@ class SpellProcess
     game_state.reset_stance = true
   end
 
-  def prepare_spell(data, game_state)
+  def prepare_spell(data, game_state, force_cambrinth = false)
     return unless data
     game_state.cast_timer = Time.now
     @prep_time = data['prep_time']
@@ -1290,8 +1290,8 @@ class SpellProcess
     game_state.casting = true
     game_state.cambrinth_charges(data['cambrinth'])
     if data['cambrinth']
-      @should_harness = check_to_harness(@harness_for_attunement)
-      @should_invoke = data['cambrinth'] unless @should_harness
+      @should_harness = check_to_harness(@harness_for_attunement) && force_cambrinth
+      @should_invoke = data['cambrinth'] unless @should_harness 
     end
     @custom_cast = data['cast']
     @symbiosis = data['symbiosis']
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -839,7 +839,7 @@ class SpellProcess
     @hide_type = settings.hide_type
     echo("  @hide_type: #{@hide_type}") if $debug_mode_ct
 
-    @buff_force_cambrinth = settings.crossing_training_force_cambrinth
+    @buff_force_cambrinth = settings.combat_trainer_buffs_force_cambrinth
     echo("  @buff_force_cambrinth: #{@buff_force_cambrinth}") if $debug_mode_ct
 
     @perc_health_timer = Time.now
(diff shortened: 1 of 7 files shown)


jayRyan24:
Any questions for these? Been running the common changes for a couple weeks and the combat-trainer changes all day on several chars.

jayRyan24:
Maybe get @Sarvatt and @jandersson  to test them out/look over the code.

jayRyan24:
Want to make sure these places are good for overrides.

jandersson:
I feel like this makes more sense as a setting rather than a parameter to pass around

jayRyan24:
I don't think so. I think you are overriding, it default to false if not overrode.

jayRyan24:
So, you don't need it to be present.

jayRyan24:
This also let's us set it up as a paremeter in a spell. Let's say you want a specific spell in offensive_spells to cast, but you want to force it to use cambrinth, I think this would let you do it spell spell as a harmless type setting.

jayRyan24:
I'm not sure how you'd make it a setting.

jandersson:
Is the goal of this to harness mana instead of using cambrinth?

jayRyan24:
So, cambrinth is the default. I added in the setting to harness instead of cambrinth to train attunement. The issue is that this causes problems with crossing-training, buffs, and buff_spells due to mana leakage. The remedy to this is to set up a way to override the harness training for specific spells when needed. 

The reversal of this would be to be changed holistically for all spells, we would change the setting from cambrinth to harness, and you swap out which you want to use based on the spell type. This isn't a tough change either. 

What do you guys think we should do?

jayRyan24:
Most people will want to balance them, but sometimes you want to force cambrinth use regardless of exp. Maxxing buffs, heals, etc. But also, you will want to default to cambrinth for most people.

jandersson:
OK, I think I understand. I still disagree with passing in a new parameter that sounds like something the user has specified they want to do. If this is not a setting, how does a script determine when to enable or disable this?

jayRyan24:
The user can specify all they want, but in certain cases it's always best to use cambrinth. Now, I do get what you're saying there, but that's a whole other set of settings. We'd need:

buff_spells_always_use_cambrinth: true
crossing_training_always_use_cambrinth: true

These are possibilities if that's where you can see this going, or we can just make it the default without an override. 

I think in the case of healing, you always want to be using cambrinth to maximize. 

This is a good topic for discussion and is one I was hoping to have. This is currently in the "do as I say, not as I do" category. Adding these settings isn't difficult.

jayRyan24:
@jandersson Made those changes. I still think we should force the use of cambrinth for healing and necro stuff, it's important for certain spells to go full, no leakage.

jayRyan24:
bloody hell, ignore, one moment

rpherbig:
Should this be an attribute on a spell definition rather than a setting that applies to all spells in some context?

jayRyan24:
I think that would become tedious. You'd then have to deal with dealing with each spell individually, and in different contexts. I think, how it is implemented now. That we default to cambrinth unless overrode, and then you can force an override  `if you want`, makes it work flawlessly. I am so confused as to why this is an issue.

jayRyan24:
@adamgoerdt @Sarvatt @skyleronken @charleyc-dr  

What do you all think?

rpherbig:
This doesn't solve a problem I'm having, so I don't have a lot to say about the approach. If it does what you need, I'm ok with a few settings.

profiles/base.yaml
```diff
@@ -167,6 +167,9 @@ safe_room_empath:
 
 # Spellcasting (and Khri) settings
 use_harness_when_arcana_locked: false
+crossing_training_force_cambrinth: true
+buff_spells_force_cambrinth: true
```
rpherbig: I'm not sure if I like `buff` in the setting name... when I initially saw this, I assumed it was used by `buff.lic`. Maybe it's fine given the next setting has `waggle` in the name (though I really dislike the naming confusion between `buff` and `waggle`, it has nothing to do with this PR). Thoughts?

jayRyan24:
So the way the system works is it defaults to always use cambrinth. If you set the setting to use harness to train attunement, it follows the algorithm and trains attunement with harness. This is a very good change as it basically keeps those two magics moving without having to powerwalk. There is a problem with mana leakage which causes nerve damage, this is a big issue at lower circles for people that use crossing-training, and also I've noticed buff_spells are being cast lower than they should be due to mana leakeage. 

This fixes that by overriding the harness logic with always use cambrinth for this particular spell or set of spells, i.e., healing spells, buff spells, and crossing-training spells. I added a simple override for this, which isn't required, and defaults to false.

profiles/base.yaml
jayRyan24: Yup, I agree and I was like 'ok, I'll use waggle'. I'll figure it out.

Sarvatt:
This is a workaround because there are different casting methods for different scripts, as it is now using harness instead of cambrinth in combat works amazingly because it doesn't harness mana until just before casting, but in crossing-training and ;buff it harnesses the mana instead of using cambrinth right after prepping which can be a problem when you're holding 50 mana for 30+ seconds and some leaks giving you nerve damage. I really like how he implemented this change which gives us lots of control over exactly how we want things to work, and just works for lower level people out of the box.

Sarvatt:
I have been testing it out it for about 5 hours now and am happy with how it works.

jayRyan24:
I think jonas's suggestions of giving settings for more control and me not saying 'this is how you should do it' is better as you gain attunement you leak less and this can be turned back off.

profiles/base.yaml
jayRyan24: I made the change.

combat-trainer.lic
```diff
@@ -839,7 +839,7 @@ class SpellProcess
     @hide_type = settings.hide_type
     echo("  @hide_type: #{@hide_type}") if $debug_mode_ct
 
-    @buff_force_cambrinth = settings.buff_spells_force_cambrinth
+    @buff_force_cambrinth = settings.crossing_training_force_cambrinth
```
rpherbig: Should this be `combat_trainer_buffs_force_cambrinth`?

rpherbig:
Looks good to me. Are you ready for me to merge it @jayRyan24 ?

jayRyan24:
I am yeah, you can merge this and then also the siphon vit one which just piggy-backs off this.

jayRyan24:
Thanks for looking over it with the comb!

PR #2298, opened 2018-01-05
Adding siphon vitality to combat-trainer.

jayRyan24:
Helps during transitioning. We may need to set a player-defined threshold, but right now it's set to 90. 

Originally thought this was the case, but this is really rocking and makes using blood burst a viable option as a TM spell with auto cleanup.


jayRyan24:
Note: This needs to be absorbed AFTER my other PR, #2295. Which, by the way, is working swell.

combat-trainer.lic
```diff
@@ -1127,8 +1127,15 @@ class SpellProcess
...
     return if game_state.casting
-    return unless DRStats.empath?
-    return if health > 95 && @wounds.empty?
+    return unless DRStats.empath? || DRStats.necromancer?
+    return if health > 95 && @wounds.empty? && DRStats.empath?
+    return if health > 90 && DRStats.necromancer?
+
+    if DRStats.necromancer? && @necromancer_healing['Siphon Vitality']
+      @necromancer_healing['Siphon Vitality']['prep'] = 'target'
+      prepare_spell(@necromancer_healing['Siphon Vitality'], game_state, true)
```
jandersson: I dont think true/false should be a parameter to prepare_spell. The intention should be captured well enough by modifying the spell data or game_state, e.g. line 1135

combat-trainer.lic
```diff
@@ -1151,10 +1158,10 @@ class SpellProcess
         data = { 'abbrev' => 'heal', 'mana' => @empath_spells['HEAL'].first, 'cambrinth' => @empath_spells['HEAL'][1..-1] }
       end
       @wounds = {}
-      prepare_spell(data, game_state) if data
+      prepare_spell(data, game_state, true) if data
```
jandersson: See above comment

combat-trainer.lic
```diff
@@ -1151,10 +1158,10 @@ class SpellProcess
...
     elsif @empath_spells['VH']
       data = { 'abbrev' => 'vh', 'mana' => @empath_spells['VH'].first, 'cambrinth' => @empath_spells['VH'][1..-1] }
-      prepare_spell(data, game_state)
+      prepare_spell(data, game_state, true)
```
jandersson: See above comment

combat-trainer.lic
jayRyan24: Not sure what the issue is. prepare_spell uses cambrinth or attunement based on a setting and then based on the mind state of said skill, this is a forced cambrinth use for that. I am seriously so confused as to why this is an issue? It makes perfect sense, and defaults to false if not. There is actually no requirement for it to be put through, but if put through, it forces it. I would need a whole other setting for this. That is, unless we change all spells to be able to force cambrinth use regardless of the mind state of attunement vs arcana.

rpherbig:
This feels like it belongs with Consume Flesh and Devour logic (i.e. necromancer healing). Why was this added to `check_health` (previously empath-only) instead?

jayRyan24:
@rpherbig It was added there because it doesn't require a consume ritual being done, the other reason was that the check_health felt like a good place to put a restore of vit that didn't require a ritual. I think check_health could have been there if it didn't have other requirements.

jayRyan24:
What do you think about this instead?

```
  def check_consume(game_state)
    return unless DRStats.necromancer?
    return unless @necromancer_healing
    return if game_state.casting

    if health > @siphon_vit_threshold && @necromancer_healing['Siphon Vitality']
      @necromancer_healing['Siphon Vitality']['prep'] = 'target'
      prepare_spell(@necromancer_healing['Siphon Vitality'], game_state, true)
      return
    end

    return unless game_state.prepare_consume

    echo "@necromancer_healing.key?('Consume Flesh'): #{@necromancer_healing.key?('Consume Flesh')}" if $debug_mode_ct
    echo "@necromancer_healing.key?('Devour'): #{@necromancer_healing.key?('Devour')}" if $debug_mode_ct
    data = @necromancer_healing['Devour'] || @necromancer_healing['Consume Flesh']
    game_state.casting_consume = true
    game_state.prepare_consume = false
    prepare_spell(data, game_state)
  end
```

jayRyan24:
This is tested with a BLB dance script. It dances with BLB, which is maintaining BLB and SV and devour. This is working real well.

https://pastebin.com/uz9qNsY8

jayRyan24:
On quick fix.

combat-trainer.lic
```diff
@@ -1181,8 +1184,14 @@ class SpellProcess
     return unless DRStats.necromancer?
     return unless @necromancer_healing
     return if game_state.casting
-    return unless game_state.prepare_consume
 
+    if health <= @siphon_vit_threshold && @necromancer_healing['Siphon Vitality'] && DRRoom.npcs.any?
```
rpherbig: Should `DRRoom.npcs` here be `game_state.npcs`? The latter will filter out anything in `settings.ignored_npcs` (which I _think _ would include a zombie or construct).

jayRyan24: Yup, good call. Thanks!

jayRyan24: Done.

combat-trainer.lic
```diff
@@ -1181,8 +1184,14 @@ class SpellProcess
     return unless DRStats.necromancer?
     return unless @necromancer_healing
     return if game_state.casting
-    return unless game_state.prepare_consume
 
+    if health <= @siphon_vit_threshold && @necromancer_healing['Siphon Vitality'] && game_state.npcs?
```
rpherbig: I don't think `game_state.npcs?` will work since the function is named `npcs`. Maybe `game_state.npcs.empty?`

PR #2299, opened 2018-01-05
appraisal.lic: add bundle support.

Sarvatt:


appraisal.lic
```diff
@@ -39,6 +41,15 @@ class Appraisal
     @equipment_manager.items.each { |item| break unless appraise_item?(item) }
   end
 
+  def train_appraisal_with_bundle
+    case bput("appraise my bundle quick", 'Roundtime', 'Appraise what', 'You cannot appraise')
```
jandersson: Whats the difference between using quick and not using it?

appraisal.lic
```diff
@@ -39,6 +41,15 @@ class Appraisal
...
+    when 'Appraise what', 'You cannot appraise'
```
jandersson: This doesn't seem necessary

appraisal.lic
```diff
@@ -39,6 +41,15 @@ class Appraisal
...
+      return
+    else
+      waitrt?
```
jandersson:
```
bput('appraise bundle...
waitrt?
```
would accomplish the same thing, theres no need for a case/switch here

appraisal.lic
Sarvatt: it's harder so should train higher

appraisal.lic
Sarvatt: you're right, fixed this up.

appraisal.lic
Sarvatt: this too, thanks.

PR #2300, opened 2018-01-05
More spell messaging for Words of the Wind

Maledar:
Added 2 more messaging for ritual Words of the Wind


data/base-spells.yaml
```diff
@@ -144,6 +144,10 @@ cast_messages:
 - You point a crooked finger at
 - Mentally steeling yourself in preparation for the unnatural action
 - ^You mold an? .* of (?:the available (?:starlight|moonlight)|your starlight aura)
+- You gaze towards the sky above seeking the strongest flows of Air
```
Sarvatt: this is in here twice?

Maledar: Um, crap meant to do it once

Issue #1228, opened 2017-01-01
Make `buff.lic` support khri and barb buffs


rpherbig:
It would be nice of buff supported khri and barb buffs. Then we could use it to define pre-activity buff sets (i.e. locksmithing, athletics, etc.). Several scripts have logic to pre-buff (astrology, pick, etc.) that would be much cleaner to do via a buff set.

rpherbig:
Merged in #2303

PR #2303, opened 2018-01-06
Initial commit of barb and thief abilities added to waggle sets. 

jayRyan24:
I moved the start_barb_buffs method to common-arcana out from combat-trainer. I added the khri and barb activations to buff.lic, and I added a setting to force barb meditations to sit for lower skills barbs.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1450,7 +1450,7 @@ class AbilityProcess
       .select { |name| Flags["ap-#{name}-expired"] }
       .each do |name|
         Flags.reset("ap-#{name}-expired")
-        start_barb_buff(name)
+        activate_barb_buff(name)
         echo "Activated barb_buff #{name}!" if $debug_mode_ct
         waitrt?
         pause 6 if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1451,20 +1451,10 @@ class AbilityProcess
       .each do |name|
         Flags.reset("ap-#{name}-expired")
         start_barb_buff(name)
+        echo "Activated barb_buff #{name}!" if $debug_mode_ct
         waitrt?
         pause 6 if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
       end
-  end
-
-  def start_barb_buff(name)
-    ability_data = get_data('spells').barb_abilities[name]
-    case bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand')
-    when 'You should stand'
-      fix_standing
-      start_barb_buff(name)
-      return
-    end
-    echo "Activated barb_buff #{name}!" if $debug_mode_ct
   end
 end
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -51,6 +51,19 @@ module DRCA
     ['Your body is willing', 'You have not recovered'].include?(result)
   end
 
+  def start_barb_buff(name, sit = false)
+    ability_data = get_data('spells').barb_abilities[name]
+    is_meditation = ability_data['type'].eql? 'meditation'
+    DRC.bput('sit', 'You sit', 'You are already', 'You rise') if sit && is_meditation && !sitting?
+    case bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand')
+    when 'You should stand'
+      fix_standing
+      start_barb_buff(name)
+      return
+    end
+    fix_standing if sitting?
+  end
+
   def prepare?(abbrev, mana, symbiosis = false, command = 'prepare')
     return false unless abbrev
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -40,6 +40,14 @@ module DRCA
     end
   end
 
+  def start_khris(khris, settings)
+    khris
+      .map { |name| "Khri #{name}" }
+      .each do |name|
+        activate_khri?(settings.kneel_khri, name)
+      end
+  end
+
   def activate_khri?(kneel, ability)
     return false if DRSpells.active_spells[ability]
     DRCT.retreat if kneel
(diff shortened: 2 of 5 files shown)


jayRyan24:
Example of working for Barb: https://pastebin.com/ZmkmaKyM

jayRyan24:
```
leustyin [12:20 PM] 
@blackheart Yeah, works for my thief as expected
```

jayRyan24:
@rpherbig Can this get a look over?

common-arcana.lic
```diff
@@ -51,6 +59,29 @@ module DRCA
...
+    ability_data = get_data('spells').barb_abilities[name]
+    is_meditation = ability_data['type'].eql? 'meditation'
+    DRC.bput('sit', 'You sit', 'You are already', 'You rise') if sit && is_meditation && !sitting?
+    case bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand')
+    when 'You should stand'
+      fix_standing
+      activate_barb_buff(name)
+      return
+    end
+    fix_standing if sitting?
```
rpherbig: You don't need the conditional here, there's already one inside of `fix_standing`

common-arcana.lic
```diff
@@ -40,6 +40,14 @@ module DRCA
     end
   end
 
+  def start_khris(khris, settings)
```
rpherbig: With changes to `common-arcana`, we're going to see errors for people with multiple characters. Would you be up for putting just these changes in a separate PR?

combat-trainer.lic
```diff
@@ -1450,22 +1450,12 @@ class AbilityProcess
       .select { |name| Flags["ap-#{name}-expired"] }
```
rpherbig: Should line 1442 use the new Khri helper in `common-arcana`?

rpherbig:
This will fix #1228

common-arcana.lic
```diff
@@ -51,6 +59,29 @@ module DRCA
...
+        pause 6 if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
+      end
+  end
+
+  def activate_barb_buff(name, sit = false)
+    ability_data = get_data('spells').barb_abilities[name]
+    is_meditation = ability_data['type'].eql? 'meditation'
+    DRC.bput('sit', 'You sit', 'You are already', 'You rise') if sit && is_meditation && !sitting?
+    case bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand')
+    when 'You should stand'
```
rpherbig: Is there a 'you should sit' case we could handle to negate the need for a meditation-specific setting?

common-arcana.lic
```diff
@@ -51,6 +59,29 @@ module DRCA
...
+        activate_barb_buff(name, settings.sit_meditations)
+        waitrt?
+        pause 6 if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
+      end
+  end
+
+  def activate_barb_buff(name, sit = false)
+    ability_data = get_data('spells').barb_abilities[name]
+    is_meditation = ability_data['type'].eql? 'meditation'
+    DRC.bput('sit', 'You sit', 'You are already', 'You rise') if sit && is_meditation && !sitting?
```
rpherbig: Should this retreat before sitting?

common-arcana.lic
jayRyan24: This shouldn't affect anyone as the methods are only called by people using buff. Barbs will be affected, but they just need to restart. So you're asking me to rename the activate_barb_buff back to start_barb_buff and then make a separate PR for the name change? Not sure how that's different from this?

common-arcana.lic
jayRyan24: Fixed.

common-arcana.lic
jayRyan24: Done.

common-arcana.lic
jayRyan24: Done.

combat-trainer.lic
jayRyan24: Can you activate a Khri that is already active? It uses a timer system there, which is why I didn't push it through. I think the Khri system might need a barb style setup. I think that's outside the scope of this PR. It can definitely get done though.

rpherbig: Yeah, that does seem outside the scope of this PR.

common-arcana.lic
rpherbig: I forgot that people would have to opt-in to this by changing up their `buff` set. That's fine then.

common-arcana.lic
```diff
@@ -51,6 +59,29 @@ module DRCA
...
+      .each do |name|
+        next if DRSpells.active_spells[name]
+        activate_barb_buff(name)
+        waitrt?
+        pause 6 if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
+      end
+  end
+
+  def activate_barb_buff(name)
+    case bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand', 'You must be sitting')
```
rpherbig: I'm pretty sure this `bput` needs to be `DRC.bput`, could you double check that?

common-arcana.lic
```diff
@@ -51,6 +59,29 @@ module DRCA
     ['Your body is willing', 'You have not recovered'].include?(result)
   end
 
+  def start_barb_abilities(skills, settings)
+    skills
+      .each do |name|
+        next if DRSpells.active_spells[name]
```
rpherbig: Another option would be to `.reject { |name| DRSpells.active_spells[name] }.each do ....`. To me that makes it a bit clearer what each function in the chain is doing.

Note: I'm just pointing out a style difference, you don't have to change this.

jayRyan24: No, this is what I want from reviews! Thanks a lot!

jayRyan24: Definitely wanting to use more of what Ruby has to offer, I just always forget about select/reject.

common-arcana.lic
jayRyan24: It does! But my question is how's it working without? It's just *smart*?

PR #2304, opened 2018-01-06
support removing lodged items in tendme

skyleronken:


tendme.lic
```diff
@@ -20,9 +20,19 @@ def bind_open_wounds?
...
+           .dup
+           .map { |x| x.scan(/lodged \w* into your [^\,\.]+/).to_a }
+           .flatten
+           .compact
 
   bleeders = bleeders.keep_if { |item| !item.include?('tended') && !item.include?('inside') && !item.include?('Area') }
 
+  lodged.each do |message|
+    message =~ /into your ([^\,\.]+)/
+    DRCH.bind_wound(Regexp.last_match(1))
```
rpherbig: Did you test what happens if you try this and your hands are full?

Do other scripts break if this runs and then the bolt (or whatever was lodged) is now in your hand?

PR #2306, opened 2018-01-06
Adding dynamic dance_skill

skyleronken:
determine next dance skill based on lowest exp weapons

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2230,6 +2230,7 @@ class GameState
   $thrown_skills = ['Heavy Thrown', 'Light Thrown']
   $aim_skills = %w[Bow Slings Crossbow]
   $ranged_skills = $thrown_skills + $aim_skills
+  $non_dance_skills = $ranged_skills + ['Brawling', 'Offhand Weapon']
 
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -2513,6 +2517,10 @@ class GameState
   end
 
   def dance
+    if @dynamic_dance_skill 
+      filtered_skills = weapon_training.reject { |skill, _| $non_dance_skills.include?(skill) }
+      @dance_skill = filtered_skills.min_by { |skill, _| DRSkill.getrank(skill) }.first
```
rpherbig: Should this `min_by` exp and rank?

rpherbig: Similar to https://github.com/skyleronken/dr-scripts/blob/907c309f5f10ac6aafb9cfa3a2f08890be4c29f5/combat-trainer.lic#L141 is what I was thinking of

skyleronken: I removed the exp portion to avoid switching weapons too frequently. You could theoretically enter swapping hell if the mindstate were part of this determination.

rpherbig: Ah, good point. Thank you for the explanation.

PR #2310, opened 2018-01-06
Based on community feedback, this is an issue I needed to fix. 

jayRyan24:
Harnessing mana didn't release it on spell failure, so the next spell would definitely fail.

This occurs when **harness mana** is set to be used for attunement training! Just pointing out the issue's starting point.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -914,7 +914,10 @@ class SpellProcess
       echo('SpellProcess::clean_up') if $debug_mode_ct
       game_state.next_clean_up_step
       release_cyclics
-      bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell") if checkprep != 'None'
+      if checkprep != 'None'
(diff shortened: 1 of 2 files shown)


jayRyan24:
Any place that `release spell` was called in CT and associated scripts, I just added `release mana`.

jayRyan24:
@rpherbig This should be an easy merge.

combat-trainer.lic
```diff
@@ -958,7 +961,10 @@ class SpellProcess
     return if game_state.cast_timer.nil? || (Time.now - game_state.cast_timer) <= 70
 
     game_state.cast_timer = nil
-    fput('release spell') if game_state.casting
+    if game_state.casting
+      fput('release spell')
```
rpherbig: I'll go ahead and merge this PR. In a future PR, would you be up for changing this and the `fput('release spell')` in common-arcana to be `bput('release spell')` like in combat-trainer?

jayRyan24: Yeah, want me to just do that now?

Issue #2279, opened 2017-12-29
Set a room(s) for athletics to train outdoors


izawwlgood:
Help spread us out so we don't all accumulate in the rooms near the undergondola for training athletics.

athletics_room_outdoors_training: for example.

jayRyan24:
https://github.com/rpherbig/dr-scripts/pull/2311

jayRyan24:
This was merged @izawwlgood. You can close it.

PR #2311, opened 2018-01-06
Adding support for the issue created: #2279. 

jayRyan24:
Adds in a collecting room to athletics. 

#2279.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -251,6 +251,7 @@ appraisal_training:
 full_pouch_container:
 classes_to_teach:
 avoid_athletics_in_justice: false
+athletics_coutdoorsmanhship_room:
 attunement_rooms:
 
 
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -251,7 +251,7 @@ appraisal_training:
 full_pouch_container:
 classes_to_teach:
 avoid_athletics_in_justice: false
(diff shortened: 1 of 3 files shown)


jayRyan24:
Working well for me so far.

jayRyan24:
@rpherbig Seems pretty straight forward. Am using it currently.

profiles/base.yaml
```diff
@@ -251,6 +251,7 @@ appraisal_training:
 full_pouch_container:
 classes_to_teach:
 avoid_athletics_in_justice: false
+athletics_outdoorsmanship_rooms:
```
rpherbig: If this is an array, it needs an entry in `base-empty.yaml`

jayRyan24: Done. What's this actually do?

rpherbig: If someone overrides a setting in their YAML such as `athletics_outdoorsmanship_rooms:` then `get_settings.athletics_outdoorsmanship_rooms` will return `nil`, not an array. Sometimes that's ok, but sometimes that causes an error (e.g. if you're using an Array-specific function in your script).

Base-empty is almost solely for settings which are arrays and hashes. It provides a default value if the setting is `nil` when loaded.

jayRyan24: Ah, thanks for the explanation.

PR #2320, opened 2018-01-07
Combat-Trainer: Add setting for allowing spell mana/cambrinth/harness data to be calculated dynamically

jandersson:
This works almost the same as with the logic in crossing-trainer. 

Add the attribute `use_auto_mana: true` to spell data and it will update the mana/cambrinth data based on info from `discern spellname` and the `calculate_mana` method from DRCA. If symbiosis is true then it will step up the mana/cambrinth data one mana at a time until a cast teaches at least as much as defined in `symbiosis_learning_threshold` (default is 2).

If the setting name is terrible (`use_auto_mana`) let me know, I dont care one way or the other what its called :P

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -842,6 +842,15 @@ class SpellProcess
 
     @perc_health_timer = Time.now
 
+    @check_discern_settings = OpenStruct.new(
+      'prep_scaling_factor' => settings.prep_scaling_factor,
+      'cambrinth_items' => settings.cambrinth_items,
+      'cambrinth' => settings.cambrinth,
+      'cambrinth_cap' => settings.cambrinth_cap,
+      'stored_cambrinth' => settings.stored_cambrinth,
+    )
+    echo(" @check_discern_settings: #{@check_discern_settings}")
+
     @symbiosis_learning_threshold = settings.symbiosis_learning_threshold
     echo(" @symbiosis_learning_threshold: #{@symbiosis_learning_threshold}")
 
@@ -1257,7 +1266,7 @@ class SpellProcess
   def prepare_spell(data, game_state, force_cambrinth = false)
     return unless data
 
-    data = check_discern if data['use_auto_mana']
+    data = check_discern(data, @check_discern_settings) if data['use_auto_mana']
     game_state.cast_timer = Time.now
     @prep_time = data['prep_time']
     echo("prepare spell: #{data}") if $debug_mode_ct
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1,7 +1,7 @@
 =begin
   Documentation: https://elanthipedia.play.net/Lich_script_repository#combat-trainer
 =end
-
+require 'ostruct'
 custom_require.call(%w[common common-arcana common-healing common-items common-summoning common-travel drinfomon equipmanager events spellmonitor])
 
 class SetupProcess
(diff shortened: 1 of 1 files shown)


jayRyan24:
Setting name seems file, other options could be:

`use_discerned_mana:` (though this sounds like only what is discerned)
`calculate_mana:` Sounds like a command to the system to calculate it's own cost.
`use_calculated_mana:` Similar to above, but less like a command.

jandersson:
Closing while I work out a bug. I'll re-open when I've got it fixed

jandersson:
Bug fixed

combat-trainer.lic
```diff
@@ -842,6 +842,18 @@ class SpellProcess
 
     @perc_health_timer = Time.now
 
+    @check_discern_settings = OpenStruct.new(
```
rpherbig: You _probably_ need to `require 'ostruct'` at the top of the file if you're going to do this. Have you tested this using a Hash instead of OpenStruct?

jandersson: I used a hash initially but switched to openstruct due to errors from DRCA.  Common-arcana::check_discern takes user settings which is an OpenStruct. 

I can add the require to the top. I think the reason I did not see any errors is that dependency does this.

jandersson: The alternative I considered was making settings an instance variable, but none of the other Combat processes seem to do this.

jandersson: using the dot operator on hashes was the error if memory serves
```
2.0.0-p647 :002 > settings_hash = {'a'=> 'b'}
 => {"a"=>"b"}
2.0.0-p647 :003 > require 'ostruct'
 => true
2.0.0-p647 :004 > settings_struct = OpenStruct.new('a' => 'b')
 => #<OpenStruct a="b">
2.0.0-p647 :005 > settings_hash.a
NoMethodError: undefined method `a' for {"a"=>"b"}:Hash
        from (irb):5
        from /home/jonas/.rvm/rubies/ruby-2.0.0-p647/bin/irb:12:in `<main>'
```

rpherbig: Storing `settings` seems pretty reasonable to me

jandersson: OK, that is what I would like to do as well. Is it alright if I store settings in an instance variable in this PR and then refactor the SpellProcess to use the settings instance var in another PR?

combat-trainer.lic
```diff
@@ -774,6 +773,8 @@ class SpellProcess
     @equipment_manager = equipment_manager
     echo('New SpellProcess') if $debug_mode_ct
 
+    @settings = get_settings
```
rpherbig: We don't want to call `get_settings` here as it will be missing any args that were passed to `combat-trainer`. SpellProcess should be passed `settings` as an argument, which would be what we want to store.

wscherer420:
Is there a profile with this setup that i can see?  I added use_auto_mana: true and removed all the mana from my spells in the YAML but i just began casting spells at minimum.  So i'm assuming there's more to it than just putting use_auto_mana: true into my YAML.

Sarvatt:
you'd only cast spells at minimum if you use symbiosis, it ramps up the mana until you learn 2 mindstates per cast

wscherer420:
I am using symbiosis.  So are you saying it ought to still ramp up the mana even when using symbiosis?

PR #2328, opened 2018-01-08
Add glyph-of-mana.lic, to keep glyph of mana up always.

Sarvatt:
This is meant to be used as a during: in hunting_info so it stays up in combat, like

```hunting_info:
- :zone: cinder_beasts
  args:
  - d1
  stop_on:
  - Small Edged
  :duration: 55
  before:
  - go2 12345
  - mech-lore
  - athletics
  during:
  - glyph-of-mana


Sarvatt:
This is super low priority fwiw, not really worth the review effort until other things go in. Not many paladins around who care about this :)

PR #2331, opened 2018-01-08
Add paladin support; prayer badge, prayer to chadatru, glyphs.

jayRyan24:
Initial commit of paladin supports. I added prayer badge support to combat-trainer under abilityprocess. I added support for prayer to chadatru to crossing-training. Prayer badge is 31 minutes cooldown, and prayer to chadatru is 1 hour cooldown. I need the number for Shard and Mer'Kresh to make it happen for them.


jayRyan24:
I made the prayer badge a UserVar because I want to expand that into crossing-training if people think we should. 

I added the Paladin skills to AbilityProcess because I think we can get some good use out of this location for things like Smite. It's about to be expanded.

combat-trainer.lic
```diff
@@ -1425,15 +1426,55 @@ class AbilityProcess
     @barb_buffs = @buffs.delete('barb_buffs') || []
     echo("  @barb_buffs: #{@barb_buffs}") if $debug_mode_ct
     setup_barb_buff_flags
+
+    @paladin_use_badge = settings.paladin_use_badge
+    echo("  @kneel_khri: #{@kneel_khri}") if $debug_mode_ct
```
rpherbig: Should this be `@paladin_use_badge`?

combat-trainer.lic
```diff
@@ -1425,15 +1426,55 @@ class AbilityProcess
...
+    if DRCI.wearing?('pilgrim badge')
+      @paladin_badge_get_verb = 'remove'
+      @paladin_badge_store_verb = 'wear'
+    else
+      @paladin_badge_get_verb = 'get'
+      @paladin_badge_store_verb = 'stow'
+    end
+
+    UserVars.paladin_last_badge_use ||= Time.now
+    @badge_reuse_cooldown = 1860 # 31 minutes
```
rpherbig: I wonder if this should be in TrainerProcess... then you get the timer for free. It wouldn't be based on a skill check, but we have a precedent for that (FavorOrb: https://github.com/jayRyan24/dr-scripts/blob/60839aab83cc6b9829b9ce4f37bd870174c5bbb7/combat-trainer.lic#L1580)

combat-trainer.lic
```diff
@@ -1425,15 +1426,55 @@ class AbilityProcess
...
 
   private
 
+  def check_paladin_skills(game_state)
+    return unless DRStats.paladin? 
+
+    pray_badge if @paladin_use_badge
+  end
+
+  def pray_badge
```
rpherbig: There looks to be quite a bit of overlap with the `pray_badge` function in `theurg.lic`. Could this be in a common file (and maybe return a bool for success/fail)?

crossing-training.lic
```diff
@@ -243,6 +248,32 @@ class CrossingTraining
     bput("put 5 silver #{currency} in almsbox", 'You drop', 'But you do not', 'attend to thy own woes')
   end
 
+  def check_chadatru
+    return unless DRStats.paladin? && @pray_to_chadatru
+    return unless Time.now - UserVars.chadatru_prayer_last > @chadatru_prayer_cooldown
+
+    case @hometown
+    when 'Crossing'
+      walk_to('5844')
```
rpherbig: This feels like some data that should be in `base-hometown` or `base-theurgy` (so you don't need a big case/when block). What do you think?

pilgrimage.lic
```diff
@@ -34,6 +34,7 @@ class Pilgrimage
       bput("wear my #{description}", 'You put on')
       move('out')
     else
+      continue if DRStats.paladin?
```
rpherbig: What does this line do?

profiles/base.yaml
```diff
@@ -255,6 +256,7 @@ classes_to_teach:
 avoid_athletics_in_justice: false
 athletics_outdoorsmanship_rooms:
 attunement_rooms:
+pray_to_chadatru: false
```
rpherbig: Could this be renamed to something more specific? The name alone doesn't tell me if this is _all_ prayer or just some.

rpherbig:
What do you think about adding a line to `new-character.lic`? Something like "ask a Cleric to buy you a pilgrim's badge then run 'pilgrimage.lic'" (see https://github.com/rpherbig/dr-scripts/blob/master/new-character.lic#L51)

combat-trainer.lic
jayRyan24: Yeah, caught it in my current iteration already.

combat-trainer.lic
jayRyan24: Is this used to train a skill? I thought it was used to maintain high levels of soul.

jayRyan24: oh sorry, misread. I want the overall timer so we can use it during crossing-training too.  Think I left a comment to this effect in the main comments section previous.

crossing-training.lic
jayRyan24: Probably would be a good idea. I can do that.

pilgrimage.lic
jayRyan24: Skips the error message and exit so we can do the rest of the script. I did it this way because then we'd need to pull the failure logic out and put it in the main, this just follows suit.

profiles/base.yaml
jayRyan24: I am not sure what you mean by all or some. Any suggestions?

combat-trainer.lic
```diff
@@ -1424,16 +1425,97 @@ class AbilityProcess
...
+
+    pray_badge
+    check_glyphs
+  end
+
+  def setup_glyph_flags
+    @paladin_glyphs
+      .reject { |name|
+        unless get_data('spells').paladin_abilities[name]['type'].eql? 'glyph'
+          echo "***THE {glyph} GLYPH IS NOT SUPPORTED AT THIS TIME***"
```
rpherbig: Should this be `#{glyph}`?

combat-trainer.lic
```diff
@@ -1424,16 +1425,97 @@ class AbilityProcess
...
+  def check_paladin_skills(game_state)
+    return unless DRStats.paladin? 
+
+    pray_badge
+    check_glyphs
+  end
+
+  def setup_glyph_flags
+    @paladin_glyphs
+      .reject { |name|
```
rpherbig: The block inside of `.reject` needs to return a bool. While this does, it's a bit unclear (I _think_ it's returning the `echo` result which ends up being true.

combat-trainer.lic
```diff
@@ -1424,16 +1425,97 @@ class AbilityProcess
...
+      .select { |name| Flags["glyph-#{name}-expired"] }
+      .each do |name|
+        if use_glyph(name)
+          Flags.reset("glyph-#{name}-expired")
+          echo "Used glyph #{name}!" if $debug_mode_ct
+          waitrt?
+          return
+        else
+          echo "***YOU DO NOT KNOW THE GLYPH #{name}. REMOVING FROM HUNT***"
+          @paladin_glyphs.delete(name)
```
rpherbig: It feels like something weird may happen if you delete an item from the array while in the midst of enumerating it. Did you test what Ruby does in that case?

rpherbig:
In the future, you might want to consider splitting up independent features (prayer vs badge in this case) into separate PRs. There is a lot of code to review.

new-character.lic
```diff
@@ -90,7 +90,13 @@ class NewCharacter
   def check_theurgy
     echo '***CHECKING THEURGY***'
     return unless @settings.crossing_training.include?('Theurgy')
-    return unless DRStats.cleric?
+
+    if DRStats.paladin?
```
rpherbig: Will the `return unless` on line 92 prevent this from running? Or do Paladins put Theurgy in their training list?

It feels like this might go better in the 'manual steps' section.

data/base-spells.yaml
```diff
@@ -309,6 +309,13 @@ barb_abilities:
     activated_message: ^You .* to meditate
     expired_message: Focus meditation slips away from your mind
 
+paladin_abilities:
```
rpherbig: Do you plan to update https://github.com/rpherbig/dr-scripts/blob/master/glyph-of-mana.lic to use this data?

crossing-training.lic
```diff
@@ -53,6 +56,7 @@ class CrossingTraining
     @equipment_manager = EquipmentManager.new(@settings)
 
     UserVars.crossing_timers ||= {}
+    UserVars.chadatru_prayer_last ||= Time.now - 3600
```
rpherbig: Should this be `... - @chadatru_prayer_cooldown`?

common-arcana.lic
```diff
@@ -83,6 +83,13 @@ module DRCA
     fix_standing
   end
 
+  def use_glyph(name)
```
rpherbig: If this will return true/false (a 'predicate'), the Ruby convention is to have a `?` at the end of the function name (so `use_glyph?`).

common-arcana.lic
```diff
@@ -83,6 +83,13 @@ module DRCA
...
+    ability_data = get_data('spells').paladin_abilities[name]
+    result = DRC.bput("#{ability_data[name]['use_command']}", ability_data[name]['activated_message'], 'only to look foolish', 'this glyph is already in effect here')
+    return false if result =~ /only to look foolish/
```
rpherbig: Can you `return result !~ /only to look foolish/` instead of lines 89 and 90? If so, the `return` is implicit and could be removed as well (so the line would just be `result =~ /only to look foolish/`)

combat-trainer.lic
jayRyan24: Ah, it should be #{name}, thanks! I did have the variables named glyph before this.

combat-trainer.lic
jayRyan24: Changed it, thanks, let me know if that is OK or if you know of a better method.

combat-trainer.lic
jayRyan24: Good call. Let me know if this solution is acceptable. Tested it out and subtracting an empty array from an array shouldn't cause an issue.

new-character.lic
jayRyan24: Good call, done. I looked for badge and din't know about manual_changes.

data/base-spells.yaml
jayRyan24: Not really, this should negate the need for that script.

crossing-training.lic
jayRyan24: No, so the chadatru is a one hour cooldown, and we can keep track of it so if we start up we do it automatically, but if we restart we don't waste time doing it again. The cooldown checks that it's been an hour since the `chadatru_prayer_last` time. The only time there can be an overlap is on initial run, where if you've run it recently, update the scripts and get this functionality, and then run it you will overlap, but it shouldn't be an issue. Or if you delete that UserVar you will run into the previous issue.

common-arcana.lic
jayRyan24: OK, will add.

common-arcana.lic
jayRyan24: Right, changed it.

jayRyan24:
Sorry, this thing ballooned. :( I will keep what you said in mind in the future!

combat-trainer.lic
rpherbig: Is the boolean here backwards? Since you're using `.reject`, a value of `false` means 'keep this item' and `true` means 'reject this item`.

If so, you could either swap `true` and `false` or change `.reject` to `.select`.

crossing-training.lic
rpherbig: I'm not sure I was clear: what I was suggesting is `UserVars.chadatru_prayer_last ||= Time.now - @chadatru_prayer_cooldown`.

Why use the magic number '3600' here when `@chadatru_prayer_cooldown` is defined above to be '3600'. Are these two logically unrelated and just happen to be the same value?

pilgrimage.lic
rpherbig: The only way to get into this block is if you don't have a badge. Why do you want to go visit all the shrines if you are a Paladin without a badge?

jayRyan24: Yeah, I didn't thoroughly read this script enough.

combat-trainer.lic
jayRyan24: ah yes, duh

jayRyan24: I'm going to change it to select, I think the reject and then unless read too difficult.

crossing-training.lic
jayRyan24: Happen to be the same value, the 3600 was added before the variable. Yup, I'll change it.

jayRyan24: Except sometimes magic numbers make me feel like a magician? 🥇

PR #2337, opened 2018-01-08
Remove buff from common-arcana

jandersson:
This feels like what `alias` is for


jandersson:
Alternatively I can change all usages of `buff` to `cast_spell`. I see it used in pick.lic, lichbot.lic, scouting.lic, sorcery.lic and steal.lic

rpherbig:
I like that idea

rpherbig:
I'll merge this in a few days since it affects a `common-*` file

PR #2341, opened 2018-01-08
Adding bonding quest for paladins.

jayRyan24:
More of an escort than a hand holding, at this time.

### DIFF
--- a/paladin-quests.lic
+++ b/paladin-quests.lic
@@ -16,12 +16,13 @@ class PaladinQuests
       exit
     end
 
-  arg_definitions = [
+    arg_definitions = [
       [
         { name: 'warding', regex: /warding/i, description: 'Complete the Glyph of Warding Quest.' }
       ],
       [
-        { name: 'bonding', regex: /bonding/i, description: "Complete the Glyph of Bonding Quest." }
+        { name: 'bonding', regex: /bonding/i, description: "Complete the Glyph of Bonding Quest. This is more of an quester helper/escort. You will need to participate!" }, 
+        { name: 'town', options: %w[shard crossing hib], description: 'Town in which to complete the quest: Shard, Crossing, or Hib.' }
       ],
       [
         { name: 'light', regex: /light/i, description: 'Complete the Glyph of Light Quest.' }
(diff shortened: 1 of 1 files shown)


jayRyan24:
More of an escort for the bonding quest at this time.

paladin-quests.lic
```diff
@@ -46,9 +47,20 @@ class PaladinQuests
     if args.warding
       warding
     elsif args.bonding
-      tamsine
+      case args.town
```
rpherbig: What do you think about moving all of this logic into the `bonding` function? That way `initialize` is really just a dispatcher

paladin-quests.lic
```diff
@@ -46,9 +47,20 @@ class PaladinQuests
...
+      when 'shard'
+        args.town = 'Shard'
```
rpherbig: In general I'm not a fan of mutating "input" (which I consider `args` to be). More a style thing than functional here, though.

paladin-quests.lic
```diff
@@ -75,9 +87,54 @@ class PaladinQuests
...
+    walk_to(room)
+
+    case bput("ask #{leader} about glyph", 'learn the Glyph of Bonding', 'the Citizenship Office')
+    when 'the Citizenship Office'
+      echo '***YOU DO NOT MEET THE REQUIREMENTS***'
+      echo "***PLEASE SUPPLY THIS LOG AS AN ISSUE IF YOU GET A 'NO MATCH FOUND' MESSAGE***"
+      exit
+    end
+
+    echo "***PAUSING SCRIPT. FIND THE ROOM DISPLAYED IN THE VISION. IT WILL BE IN THIS TOWN! UNPAUSE WHEN YOU FIND IT!***"
```
rpherbig: What do you think about saying _how_ to unpause? I don't know if that's something most people are familiar with (`... ;unpause paladin-quests ...`)

paladin-quests.lic
jayRyan24: It's super weird that inputting Shard or shard will result in shard. Doesn't matter how the argument accepts it either. Do you know why this is and how to stop it from happening? It can't match shard to Shard in data-town, and I can't get .capitalize to work, maybe too old of a version?

paladin-quests.lic
jayRyan24: Actually meant to add this, thanks for the reminder!

paladin-quests.lic
jayRyan24: Agreed

PR #2345, opened 2018-01-10
Validate priority defense

charleyc-dr:
The recent change to the handing of `priority_defense` exposed the fact that many people had "Parry" or "Shield" for that value (which previously was also handled badly, but less obviously).  Add a validation check for that.


validate.lic
```diff
@@ -512,6 +518,7 @@ class DRYamlValidator
     @valid_weapon_skills = @valid_ranged_skills + @valid_melee_skills + ['Targeted Magic']
     @valid_summon_skills = @valid_thrown_skills + @valid_melee_skills - ['Offhand Weapon', 'Brawling']
     @valid_research_skills = ['Arcana', 'Life Magic', 'Holy Magic', 'Lunar Magic', 'Elemental Magic', 'Arcane Magic', 'Attunement', 'Warding', 'Augmentation', 'Utility']
+    @valid_defensive_skills = ['Shield Usage', 'Parry Ability', 'Evasion']
```
rpherbig: It's not part of this PR, but you've made me think that another good validator would be to check the defense skills in the `stances` setting. Just a thought for the future...

validate.lic
```diff
@@ -160,6 +160,12 @@ class DRYamlValidator
             .each { |skill| error("Invalid training_manager_priority_skills: skill name '#{skill}'") }
   end
 
+  def assert_that_priority_defense_skill_is_valid(settings)
+    return unless settings.priority_defense
+
+    error("Invalid priority_defense skill: skill name '#{settings.priority_defense}'") unless @valid_defensive_skills.include?(settings.priority_defense)
```
rpherbig: Style-wise, I'd recommend turning the `unless` clause into another guard (like line 164). It's not a big deal though.

PR #2359, opened 2018-01-11
Update astrology.lic

Rhayle:
added perceive mana, planets to moons area. removed individual moons for single moons noun. added astrology_pool_target for setting pool size  before prediction.

Add astrology_pool_target to yaml set between 1-10. or set astrology_use_full_pools to true

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -30,7 +30,7 @@ class Astrology
     @have_telescope = settings.have_telescope
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
-    @prediction_pool_target = settings.astrology_pool_target || (settings.astrology_use_full_pools ? 10 : 1)
+    @prediction_pool_target = settings.astrology_use_full_pools ? 10 : astrology_pool_target
     @rtr_data = nil
     echo " Flags['rtr-expire'].nil? #{Flags['rtr-expire'].nil?}" if UserVars.astrology_debug
     echo " Flags['rtr-expire'] = #{Flags['rtr-expire']}" if UserVars.astrology_debug
--- a/astrology.lic
+++ b/astrology.lic
@@ -73,6 +73,7 @@ class Astrology
   end
 
   def check_attunement
+    return if DRSkill.getxp('Attunement') > 30
     ['','mana','moons','planets'].each do |target|
       bput("perceive #{target}", 'roundtime')
       waitrt?
--- a/astrology.lic
+++ b/astrology.lic
@@ -30,7 +30,7 @@ class Astrology
     @have_telescope = settings.have_telescope
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
-    @prediction_pool_target = settings.astrology_use_full_pools ? 10 : astrology_pool_target
+    @prediction_pool_target = settings.astrology_use_full_pools ? 10 : settings.astrology_pool_target
     @rtr_data = nil
     echo " Flags['rtr-expire'].nil? #{Flags['rtr-expire'].nil?}" if UserVars.astrology_debug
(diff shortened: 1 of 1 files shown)


astrology.lic
```diff
@@ -73,8 +73,8 @@ class Astrology
   end
 
   def check_moons
-    %w[xibar yavash katamba].each do |moon|
-      bput("pow #{moon}", 'roundtime')
+    %w[mana moons planets].each do |moon|
```
rpherbig: Should the arg to the block still be `moon`? Maybe `target`?

astrology.lic
```diff
@@ -73,8 +73,8 @@ class Astrology
   end
 
   def check_moons
```
rpherbig: Should this function be renamed?

Sarvatt:
Could you add `astrology_pool_target:` to profiles/base.yaml while adding the new option?

astrology.lic
```diff
@@ -329,8 +329,10 @@ class Astrology
         check_weather
       when 'events'
         check_events(check_pools)
-      when 'moons'
-        check_moons
+      when 'moons'  #deprecated - left for backwards compat. only
```
rpherbig: Oh, good catch. I think we should add a validator for `astrology_training`. Or maybe two - one that checks for valid values and one for the deprecated 'moon' value. What do you think?

Rhayle: ill work on that this weekend. Also, the above modifications are incorrect and do not correctly check pool size. It needs a parenthesis set. Please do not merge.

astrology.lic
```diff
@@ -329,8 +329,10 @@ class Astrology
...
+        check_attunement
+      when 'attunement'
```
jandersson: Youll want to change `astrology_training` in base.yaml to use `attunement` instead of `moons`
https://github.com/rpherbig/dr-scripts/blob/master/profiles/base.yaml#L757

astrology.lic
```diff
@@ -72,9 +72,10 @@ class Astrology
     result.compact.select { |data| data['circle'] <= DRStats.circle }
   end
 
-  def check_moons
-    %w[xibar yavash katamba].each do |moon|
-      bput("pow #{moon}", 'roundtime')
+  def check_attunement
```
rpherbig: Does this _only_ train Attunement? If so, why is it in the Astrology trainer (and not in the Attunement training script)?

jandersson: perc planets and perc moons train astrology

jandersson: perc mana is attunement training only

Rhayle: actually I just added them to finish up the attunement. The actual attunement.lic for moon mages only adds perc mana but with all 4 it drives attunement faster. I also did not know if attunement is called during hunting in which case you would not want the additional calls. perc moons is around 40 seconds at circle 1, perc planets is 20 seconds. perc mana is around 40 seconds. If attunement.lic is actually being called during combat, it may be more advantageous to change it to only call perc with a 10 second or so rt instead of perc mana and allow this to be used for actual moon mage attunement. Also, its a really good way to eat the reset timer on astrology.

jandersson: Im not against it all, I think the change is quite nice

rpherbig:
This looks fine to me. @Rhayle @jandersson are you ready for me to merge this (and #2363 at the same time)?

jandersson:
I have no problem with it. Thank you for the contribution!

Rhayle:
Can we wait until I get the validate complete?

astrology.lic
Rhayle: is it the actual attunement script called from combat? if so I may modify that to drop moon mage attunement as a 30+ second rt during combat sucks

rpherbig: No. `combat-trainer` has two types of Attunement training, neither of which call the `attunement` script:
* Perc: https://github.com/rpherbig/dr-scripts/blob/master/combat-trainer.lic#L1706
* PercMana: https://github.com/rpherbig/dr-scripts/blob/master/combat-trainer.lic#L1910-L1915

PR #2361, opened 2018-01-12
Merge combat spell training with base-spells

jandersson:
Much like the other spell merging, this compares the abbreviation and
the spell name with data base spells. If there is a match, then the data
is merged, keeping the users data.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -866,6 +866,17 @@ class SetupFiles
      .select { |spell| spells[spell] }
      .each_key { |name| s.buff_spells[name] = spells[name].merge(s.buff_spells[name]) }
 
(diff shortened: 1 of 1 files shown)


jandersson:
Im not seeing `combat_spell_training`, so I think this will cause an error if we try to iterate over nil. Closing until thats sorted.

jandersson:
I think all the spell transforming could be moved to a method `transform_spells` or `merge_spells`

rpherbig:
I'm fine with that. Note that `lockpick_buffs` should be deprecated in favor of a waggle/buff set at some point.

jandersson:
Ill take care of that

PR #2363, opened 2018-01-12
Update base.yaml to reflect changes in astrology

Rhayle:
Added astrology_pool_target: 1
Added astrology_use_full_pools: false
Added attunement option to astrology array.
Removed moons from astrology array.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -748,13 +748,15 @@ astral_plane_training:
   train_destination:
   train_source:
 predict_event: false  # Deprecated
-astrology_use_full_pools:
+astrology_use_full_pools: false
+astrology_pool_target: 1
 astrology_training:
 #- ways
 - observe
 - rtr
 - weather
-- moons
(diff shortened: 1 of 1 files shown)


profiles/base.yaml
```diff
@@ -748,13 +748,15 @@ astral_plane_training:
   train_destination:
   train_source:
 predict_event: false  # Deprecated
-astrology_use_full_pools:
+astrology_use_full_pools: false
+astrology_pool_target: 1
```
Sarvatt: this should be astrology_pool_target: not astrology_pool_target: 1? having it 1 will always make it 1 with your other change.

jandersson: conditional assignment would also accomplish this:
```ruby
target = astrology_use_full_pools ? 10 : astrology_pool_target
```

Rhayle: Updated Astrology.lic to match.

PR #2367, opened 2018-01-12
drinfomon.lic: Make DRRooms.pcs work for heritage house cantrip users.

Sarvatt:
This should cover all of the ones on https://elanthipedia.play.net/Veiled_Identity now.

Crannach whose body glows with starlight

>;en echo DRRoom.pcs
--- Lich: exec1 active.
[exec1: ["starlight", "Yndass"]]
--- Lich: exec1 has exited.


Sarvatt:
not sure if there is a better way to do this?

PR #2372, opened 2018-01-14
Add waggle 'pick' set

jandersson:
We can use a waggle set called 'pick', the name of the script, to buff.
If its defined then it takes precedence over the locksmithing_buffs
setting.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -32,9 +32,9 @@ class LockPicker
       return
     end
 
-      fput 'sit'
+    fput 'sit'
 
-      do_buffs
+    do_buffs
 
     boxes.each do |box|
       if stop_picking?
--- a/pick.lic
+++ b/pick.lic
@@ -32,11 +32,11 @@ class LockPicker
       return
     end
 
-      lockpick_buffs = if @settings.waggle_sets['pick']
-                         @settings.waggle_sets['pick']
-                       else
-                         @settings.lockpick_buffs
-                       end
+    lockpick_buffs = if @settings.waggle_sets['pick']
+                       @settings.waggle_sets['pick']
+                     else
+                       @settings.lockpick_buffs
+                     end
 
     fput 'sit'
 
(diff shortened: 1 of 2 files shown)


rpherbig:
What do you think about deprecating the lockpick_buff setting? I'm thinking you add an echo that says to use a buff set; and a validator saying it's depracated.

jandersson:
How should we handle non-spells? Those would still be casted by locksmithing_buffs

rpherbig:
Doesn't waggle now support those? @chuno

jandersson:
Added a validation check and updated to use khri/meditations

pick.lic
```diff
@@ -170,18 +174,30 @@ class LockPicker
       fput("whisper #{@settings.lockpick_buff_bot} buff hol")
     end
 
-    buffs['spells'].each do |spell|
-      echo "Buffing: #{spell}" if UserVars.lockpick_debug
-      cast_spell(spell, @settings)
+    if @settings.waggle_sets['pick']
+      if DRStats.barbarian?
```
rpherbig: Why execute this logic instead of calling `buff` with the `pick` argument?

pick.lic
```diff
@@ -170,18 +174,30 @@ class LockPicker
...
+        start_khris(buffs, @settings) 
+      else
+        cast_spells(buffs, @settings)
+      end
+    else
+      #TODO: Remove this sometime in end of February. That should be enough time for users to move to waggle
+      echo "*** You have outdated settings, please make a waggle_set called 'pick' ***"
+      buffs['spells'].each do |spell|
+        echo "Buffing: #{spell}" if UserVars.lockpick_debug
+        buff(spell, @settings)
```
rpherbig: Wasn't the `buff` function deleted?

pick.lic
```diff
@@ -32,7 +32,11 @@ class LockPicker
       return
     end
 
-    lockpick_buffs = @settings.lockpick_buffs
+    lockpick_buffs = if @settings.waggle_sets['pick']
```
rpherbig: I don't think you really need this var any more (or the arg to `do_buffs`). Since `@settings` is available you can do the check in `do_buffs`.

PR #2376, opened 2018-01-15
Update stealing script to support multiple provinces

charleyc-dr:
Also fix a few bugs; in particular, don't write multiple copies of a new item.


steal.lic
```diff
@@ -104,7 +115,7 @@ class Steal
   end
 
   def do_buff(buffs, settings)
-    buffs['spells'].each { |spell| cast_spell(spell, settings) }
+    buffs['spells'].each { |spell| buff(spell, settings) }
```
jandersson: the buff method is deprecated and going to be removed #2337

jandersson: OK, now it is removed :P

charleyc-dr: Oops, ok, reverting this!  Was an unintentional reversion during merge.

charleyc-dr:
Per discussion in Slack, going to do a bit more work to refine the "province" notion.  I've got this done but need to test it before I send it out.

charleyc-dr:
Ok, I've split things up by town and added the option to list the towns you're willing to steal from.

rpherbig:
Does the default `stealing_towns` (in base.yaml) match how stealing is currently behaving?

data/base-stealing.yaml
```diff
@@ -4648,29 +4646,21 @@
...
 - id: '409'
   item: bone backer
   item_in: in crate
   pawnable: false
-  province: zoluren
+  town: Crossing
   room: 8864
   trivial_max: 1750
   trivial_min: 741
-- id: '410'
```
rpherbig: Why was this removed?

profiles/base.yaml
```diff
@@ -269,7 +269,13 @@ steal_past_mindlock: false
 stealing_buffs:
 npc_stealing_attempt_count: 0
 dont_steal_list:
-
+stealing_towns:
```
rpherbig: Do these default stealing_towns match how stealing is currently behaving?

steal.lic
```diff
@@ -134,6 +137,19 @@ class Steal
     move('go door')
   end
 
+  def walk_to_bin
+    case @hometown
+    when 'Crossing'
```
rpherbig: This feels like something that should be in `base-town.yaml`. What do you think?

steal.lic
```diff
@@ -134,6 +137,19 @@ class Steal
     move('go door')
   end
 
+  def walk_to_bin
```
rpherbig: If this is a predicate, could you name it `walk_to_bin?` ?

steal.lic
```diff
@@ -335,36 +372,36 @@ class Steal
               'you decide to wait a while before trying again',
               'Roundtime')
     when /this trivial theft/
-      update_target(target, 'trivial')
+      update_target(target, 'trivial') if count == 1
```
rpherbig: Why did you add the conditional here?

steal.lic
```diff
@@ -20,7 +20,13 @@ class Steal
 
     @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
     @stealing_bag = settings.stealing_bag
+    @hometown = settings.hometown
+    @stealing_towns = settings.stealing_towns
     @bin_stolen = settings.bin_stolen
+    unless %w[Crossing].include?(@hometown) || !@bin_stolen
```
jandersson: It would  be nice to couple the towns with the existence of a bin room. That way the script doesn't have to be updated for any new bin data

data/base-stealing.yaml
charleyc-dr: It's a dupe of item 394.

profiles/base.yaml
charleyc-dr: Yes, these are exactly the towns where items are already in the db.

steal.lic
charleyc-dr: These were being called multiple times per theft.  Usually that's harmless, but for a new item it was making N copies if you stole N times.

charleyc-dr:
Ok, I've switched to using a bin location in base-town.yaml.

PR #2377, opened 2018-01-15
Adding Override Yaml

Zelaskal:
Feature request from folks in lnet.  Essentially adds a base override that's processed after base.yaml but before any character-setup yamls.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -552,6 +552,12 @@ class ScriptManager
     message_body += "\n"
     message_body += "Dependency Version: #{$DEPENDENCY_VERSION}"
     message_body += "\n"
+    
+    message_body += "\n****\n" * 3
+	overrideFile = File.path('./scripts/profiles/override.yaml')
+	message_body += overrideFile.to_s
+    message_body += "\n"
+    message_body += File.read(overrideFile)
 
     Dir['./scripts/profiles/*.yaml'].map { |item| File.path(item) }.select { |path| path.include?("#{checkname}-") }.each do |path|
       message_body += "\n****\n" * 3
--- a/dependency.lic
+++ b/dependency.lic
@@ -9,7 +9,7 @@ require 'yaml'
 require 'ostruct'
 require 'digest/sha1'
 
-$DEPENDENCY_VERSION = '1.2'
+$DEPENDENCY_VERSION = '1.2.1'
 
 no_pause_all
 no_kill_all
@@ -824,7 +824,11 @@ class SetupFiles
 
   def load_settings
     profile = "./scripts/profiles/#{checkname}-setup.yaml"
+    localoverrides = "./scripts/profiles/override.yaml" # Overrides things in base before calling character setup
     settings = @base_files['base.yaml'].dup
+    if File.exist?(localoverrides)
+		  settings.merge!(safe_load_yaml(localoverrides) || {})
+    end
     if File.exist?(profile)
       settings.merge!(safe_load_yaml(profile) || {})
       @extras.each do |subfile|
--- a/outdoorsmanship.lic
+++ b/outdoorsmanship.lic
@@ -16,7 +16,9 @@ class Outdoorsmanship
   end
 
   def train_outdoorsmanship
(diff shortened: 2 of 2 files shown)


Dartellum:
I do not see the purpose of this. Any changes I want are handled in my specific yaml. Why would I want settings changed before my character-setup yaml is accessed?

rpherbig:
I'm also curious what the use case is for this

Zelaskal:
The chat in lnet was around having "globals" for all of your characters.  Some suggestions were made to modify the base.yaml, but I suggested this instead.  Apparently some people wanted consistency across characters for a handful of settings.

Somehow my mindstate check for outdoorsmanship made this PR too.  That'll teach me to branch before creating PRs.

rpherbig:
I think the profile generator that @charleyc-dr made could be a more flexible way to solve this problem.

https://github.com/rpherbig/dr-scripts/blob/master/generate-profile.lic has a bunch of documentation in the header, and there are some examples in the `/profile-gen` directory.

What do you think?

charleyc-dr:
Yep, the profile generator will definitely work for this.  Just make a `profile-gen/ALL-base-setup-include.yaml` file, move your character yamls into `profile-gen` as well, and add the line `%INCLUDE base-setup` to the top of each one.  Then running `;gen` will create yamls with your common info at the top.

dependency.lic
```diff
@@ -552,6 +552,12 @@ class ScriptManager
     message_body += "\n"
     message_body += "Dependency Version: #{$DEPENDENCY_VERSION}"
     message_body += "\n"
+    
+    message_body += "\n****\n" * 3
+	overrideFile = File.path('./scripts/profiles/override.yaml')
```
rpherbig: Will this file already be output by line 545?

outdoorsmanship.lic
```diff
@@ -16,7 +16,9 @@ class Outdoorsmanship
   end
 
   def train_outdoorsmanship
-    if DRSkill.getrank('Outdoorsmanship') < 20
+    if DRSkill.getxp('Outdoorsmanship') >= 30
```
rpherbig: As noted in the PR, this change should removed

rpherbig:
I can't really argue with the simplicity of this approach

dependency.lic
```diff
@@ -552,6 +552,12 @@ class ScriptManager
...
+	message_body += overrideFile.to_s
+    message_body += "\n"
+    message_body += File.read(overrideFile)
```
jayRyan24: Not sure about the override naming convention for this, it's not an end-all, I just find it possibly misleading. 

Possible alternatives:

1. common.yaml
2. base-common.yaml
3. common-settings.yaml

Just some examples.

jayRyan24:
> It does make me wonder if there's a way to make ;gen more accessible/easier to use...

I wonder if you could put a YAML list inside of the generator, maybe some of the includes list the ../<name>-*.yamls and copies them into the gen folder. This could automate it and make it a tad easier.

rcuhljr:
Seems fine, make sure to add the file contents to `;e help_me` output please. I do think it makes it a little more convoluted to find bugs with yaml because it adds another layer where settings can change but not my problems to debug anymore :)

rcuhljr:
As far as naming, base-override.yaml or base-personal.yaml or something like that would be my preference.

jayRyan24:
Yeah, it just needs something to say we're changing base, because all YAML files across the board are override files so to speak. 

If he doesn't respond to this and get it merged, I'll take it over, @rpherbig.

rpherbig:
I'm on board with a different name to make it clearer that `base.yaml` is 'owned' by the repo (i.e. subject to change without notice) and `base-whateverThisIs.yaml` is 'owned' by the user (i.e. we will never overwrite it).

Zelaskal:
Sorry folks, been a busy day.  Not sure if I will have time this weekend to put this back together or not. @jayRyan24 if you want to take this over you won't hurt my feelings.  I will close out this pull request either way.

PR #2378, opened 2018-01-15
Ambush choke in combat

charleyc-dr:
(Plus a script for filling a dirt stacker)


Sarvatt:
Could you add this to https://github.com/rpherbig/dr-scripts/blob/master/validate.lic#L344 ?

combat-trainer.lic
```diff
@@ -1876,6 +1881,40 @@ class TrainerProcess
...
+      ) != 'Roundtime'
+      pause
+      waitrt?
+    end
+    if dirt_in_hand
+      bput("put my dirt in my #{@dirt_stacker}",
+           'dumping some dirt', 'What were you referring')
+    end
+
+    false
```
rpherbig: I realize you're following the precedent of the other training abilities, but I'm not convinced the return value is used anywhere. What do you think?

fill-dirt.lic
```diff
@@ -0,0 +1,31 @@
+custom_require.call(%w[common common-items drinfomon])
```
rpherbig: Do you plan to document this on elanthipedia?

fill-dirt.lic
```diff
@@ -0,0 +1,31 @@
...
+                             'over three quarters full',
+                             'over half full',
+                             'over a quarter full',
+                             'has a little dirt',
+                             'out of dirt') == 'it is full of dirt'
+  end
+
+  def fill_stacker
+    loop do
+      forage?('dirt')
```
rpherbig: Do you want to break if `forage?` fails?

combat-trainer.lic
rcuhljr: You'd need a return up in the case that calls this like `ambush_stun` has, but that's only necessary if the action here could cause a real change in game state, I don't think choke can kill mobs so there's no need.

fill-dirt.lic
charleyc-dr: Sure, I can do that!

combat-trainer.lic
charleyc-dr: You're right, totally not needed.  Removed.

fill-dirt.lic
charleyc-dr: Good idea; done.

fill-dirt.lic
rpherbig: Can you add a header documentation block to the script?

charleyc-dr: Done!

PR #2379, opened 2018-01-15
Astrology Changes

Rhayle:
Added attunement and specifiable prediction pool levels to astrology.lic
Added new keys to for astrology changes to base.yaml
Added validation settings for all astrology_training array keys to validate.lic

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -30,7 +30,7 @@ class Astrology
     @have_telescope = settings.have_telescope
     @ap_source = settings.astral_plane_training['train_source']
     @ap_destination = settings.astral_plane_training['train_destination']
-    @prediction_pool_target = settings.astrology_pool_target || (settings.astrology_use_full_pools ? 10 : 1)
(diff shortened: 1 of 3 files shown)


Rhayle:
I figured out how to roll all changes into one PR. hehe. I'll get the hang of this stuff yet.

astrology.lic
```diff
@@ -72,9 +72,10 @@ class Astrology
     result.compact.select { |data| data['circle'] <= DRStats.circle }
   end
 
-  def check_moons
-    %w[xibar yavash katamba].each do |moon|
-      bput("pow #{moon}", 'roundtime')
+  def check_attunement
+    return if DRSkill.getxp('Attunement') > 30
```
rpherbig: I already clicked the merge button, but it just occurred to me that you might want to move the Attunement check inside the `.each` loop. What do you think?

Rhayle: I thought of that too earlier. but decided against it as it would still kill time off the observation timer.

PR #2387, opened 2018-01-16
first-aid.lic: Empty hands on script exit instead of just completion.

Sarvatt:
So you stow your book when killing the script manually also.


rpherbig:
Have you considered not killing scripts manually?

Issue #2371, opened 2018-01-14
validate should check if sell_loot_money_on_hand is valid


Sarvatt:
sell_loot_money_on_hand: 1000 will fail because it does a keep_coppers_bank = parse(keep_money_by_currency.split(' ')) expecting the right format (which would be 1 gold or 10 silver), but it is not obvious what the formatting should be.

jandersson:
What is the correct format?

rpherbig:
`sell_loot_money_on_hand: [amount as integer] [currency as string]`

jandersson:
Sorry to be a pain, but is that `3000 kronars` or `3000 bronze`?

Sarvatt:
copper bronze silver gold platinum, it reduces it to the least amount of coins, sell_loot_money_on_hand: 3000 bronze would withdraw 3 platinum

jandersson:
This is resolved

rpherbig:
#2388

PR #2388, opened 2018-01-16
Check if sell_loot_money_on_hand is correct

jandersson:
We check if the first value is a number and the second value is a string
or substring matching the coin denominations

Handles #2371

### DIFF
--- a/validate.lic
+++ b/validate.lic
@@ -510,6 +510,13 @@ class DRYamlValidator
     warn('***YOU HAVE OUTDATED SETTINGS*** THE SETTING lockpick_buffs IS NO LONGER USED. Please make a waggle_set called \'pick\'')
   end
 
+  def assert_that_sell_loot_money_on_hand_is_correct(settings)
+    settings.sell_loot_money_on_hand =~ /(\d+) (\w+)/
+    return if Regexp.last_match(1) && %w[copper silver gold platinum].any? { |denom| denom[Regexp.last_match(2)] }
+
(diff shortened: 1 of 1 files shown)


validate.lic
```diff
@@ -510,6 +510,13 @@ class DRYamlValidator
     warn('***YOU HAVE OUTDATED SETTINGS*** THE SETTING lockpick_buffs IS NO LONGER USED. Please make a waggle_set called \'pick\'')
   end
 
+  def assert_that_sell_loot_money_on_hand_is_correct(settings)
+    settings.sell_loot_money_on_hand =~ /(\d+) (\w+)/
+    return if Regexp.last_match(1) && %w[copper silver gold platinum].any? { |denom| denom[Regexp.last_match(2)] }
```
rpherbig: What about `bronze`?

rpherbig: Does this handle partial matches (e.g. `3 silv` instead of `3 silver`)?

jandersson: Missed bronze, thanks!

jandersson: Yes, Ive tested it on `3 silver`, `3 silv` which pass and `3 basdf` which fails. 
```
2.0.0-p647 :001 > 'silver'['sil']
 => "sil"
```

jandersson: Other things that fail correctly are only including a number because it fails the regex match on the first line

jandersson: We may want to check if the substring is from the beginning of the string.. as in 'silver'['i'] will be a positive match but incorrect denomination.

rpherbig: 👍

PR #2389, opened 2018-01-16
Fix validation check

jandersson:
Reading Regexp.last_match after a regex was just evaluated will clear the previous last matches.


jandersson:
Sorry about the double PR's

PR #2394, opened 2018-01-17
astrology.lic: Have the moons target run attunement while people tran…

Sarvatt:
…sition.


Sarvatt:
Does this make sense to keep in for a month or so for people to notice the echo spam?

jandersson:
Yea that makes sense to me

Sarvatt:
conflicting after the rubocop update, I'll fix it in the morning.

PR #2397, opened 2018-01-18
DRCA: Handle injuries and using a telescope

jandersson:
Let the calling script decide how to handle receiving an empty array

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -305,7 +305,7 @@ module DRCA
     planets.each do |planet|
       case DRC.bput("center telescope on #{planet}", 'You put your eye', 'The pain is too much')
       when 'The pain is too much'
-        return observed_planets
+        break
       end
       observed_planets << planet unless Flags['planet-not-visible']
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -303,7 +303,10 @@ module DRCA
     observed_planets = []
 
     planets.each do |planet|
-      DRC.bput("center telescope on #{planet}", 'You put your eye')
+      case DRC.bput("center telescope on #{planet}", 'You put your eye', 'The pain is too much')
+      when 'The pain is too much'
+        return observed_planets
```
rpherbig: Does this need to stow the telescope before returning? (see line 315)

jandersson: Yes good call. I tested this in combat-trainer, and it must have figured out that it needed to stow it.

jandersson: I will just break out of the for loop then

Dartellum:
Sounds good. I set my safe_room_health_threshold to 5 for my Moon Mage because of this.

PR #2400, opened 2018-01-19
cleric-quests.lic: Fix up tamsine for the common-money changes.

Sarvatt:


cleric-quests.lic
```diff
@@ -139,7 +141,7 @@ class ClericQuests
   end
 
   def tamsine
-    ensure_copper_on_hand(25, 'Crossing')
+    ensure_copper_on_hand(25, @settings)
```
rpherbig: This might result in not having Kronars if your hometown is a Dok/Lir town.

I think we need to do something like this: https://github.com/rpherbig/dr-scripts/blob/master/fir.lic#L123-L125

What do you think?

Sarvatt:
Like this?

PR #2401, opened 2018-01-19
Combat-Trainer: Support for zombie creation recast timer

Morzain:
Adds support for recasting zombie before it has worn off. Also added a regex match for a message that was missing it.


jandersson:
What happens when a zombie already exists? Do we want to do anything like unsummon it?

combat-trainer.lic
```diff
@@ -2970,6 +2973,10 @@ class GameState
     DRSpells.active_spells.include?('Call from Beyond')
   end
 
+  def cfb_recast?
```
jandersson: Any particular reason to make this part of GameState? It seems like its only used in LootProcess. Would it make more sense to define it there?

PR #2404, opened 2018-01-19
base-hunting.yaml: Add more rooms to louts.

Sarvatt:


Sarvatt:
These were probably removed when the east gate map was messed up a while back, but they work fine now.

PR #2406, opened 2018-01-19
cleric-quests.lic: Pause if you've completed tamsine too recently to …

Sarvatt:
…do eluned. Happens when doing both quests in a row.

### DIFF
--- a/cleric-quests.lic
+++ b/cleric-quests.lic
@@ -129,7 +129,11 @@ class ClericQuests
     bput('meditate', 'You close your eyes and begin to meditate')
     waitfor('once again changing and reforming into your Human form')
 
-    bput('dive deeper water', 'You take a deep breath')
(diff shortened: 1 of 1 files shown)


cleric-quests.lic
```diff
@@ -129,7 +129,11 @@ class ClericQuests
     bput('meditate', 'You close your eyes and begin to meditate')
     waitfor('once again changing and reforming into your Human form')
 
-    bput('dive deeper water', 'You take a deep breath')
+    case bput('dive deeper water', 'You take a deep breath', 'You get a sense that you have completed this')
+    when 'You get a sense that you have completed this'
+      pause 300
```
Sarvatt: Note: I guesstimated on this. I waited 4 minutes and I could complete it though, so made it 5 to be on the safe side.

PR #2407, opened 2018-01-19
safe-room.lic: Move on from pc empath for other messages.

Sarvatt:
2 other common ones in use:
XXXXXX whispers, "You're not even injured, ya doofus" (in case someone runs ;safe-room force)
XXXXXX whispers, "All set Sarvatt"

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -176,7 +176,7 @@ class SafeRoom
 
     fput("whisper #{empath} heal")
     fput("listen to #{empath}")
-    Flags.add('doneheal', '"Done!"')
+    Flags.add('doneheal', '"Done!"', '"All set', '"You're not even injured')
     24.times do
       pause 5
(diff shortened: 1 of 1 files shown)


Sarvatt:
I need to test if this works first, please don't pull.

Sarvatt:
Fixed and tested.

jandersson:
Can we just check if the empath whispers to you? Adding each message someone can think of would add up.

Sarvatt:
I was thinking about that but the problem is the heal script that a bunch of people use has this in it if they are too injured to heal `fput('whisper "#{target} I'm a little too injured for that right now.  Give me a moment to heal up and I'll get right to you")` and then it runs healme until your health goes under a certain score before healing.

jandersson:
Good call, I hadn't considered that

Issue #952, opened 2016-10-19
get_item crafting helper shouldn't be exiting


rcuhljr:
```
      when 'What do you', 'What were you'
        echo("You seem to be missing: #{name}")
        exit
      when "can't quite lift it"
```

This robs agency from the scripts and kills them before they can put crafting gear away. This was leading to knitting needles being thrown away.

Dartellum:
Should exit come out of the script?

PR #2409, opened 2018-01-20
Made shape kill workorders when a low quality item was made. Update g…

Dartellum:
…et method for workorders to stop after 500 tries.

### DIFF
--- a/shape.lic
+++ b/shape.lic
@@ -152,6 +152,7 @@ class Shape
       stow_item(@noun)
     when /trash/
       dispose_trash(@noun)
+      stop_script('workorders')
     end
     exit
   end
--- a/shape.lic
+++ b/shape.lic
@@ -152,7 +152,6 @@ class Shape
       stow_item(@noun)
     when /trash/
       dispose_trash(@noun)
-      stop_script('workorders')
     end
     exit
   end
--- a/workorders.lic
+++ b/workorders.lic
@@ -258,7 +258,7 @@ class WorkOrders
   end
 
   def shape_items(info, item, quantity)
-    ensure_copper_on_hand(5000, @settings)
+    ensure_copper_on_hand(10000, @settings)
     recipe, items_per_stock, spare_stock, scrap = find_recipe(info, item, quantity)
 
     quantity.times do |count|
--- a/workorders.lic
+++ b/workorders.lic
@@ -282,7 +282,7 @@ class WorkOrders
       wait_for_script_to_complete('shape', ['log', recipe['chapter'], recipe['name'], info['stock-name'], recipe['noun']])
       log_num = bput('read my log', 'You must bundle and deliver \d+ more').scan(/\d+/).first.to_i
       if log_num == quantity
-        exit
+        break
       end  
     end
 
--- a/workorders.lic
+++ b/workorders.lic
@@ -683,11 +683,11 @@ class WorkOrders
 
   def request_work_order(recipes, npc_rooms, npc, npc_last_name, discipline, logbook, diff)
     match_names = recipes.map { |x| x['name'] }
-    $i = 0
+    i = 0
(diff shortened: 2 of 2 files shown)


workorders.lic
```diff
@@ -683,9 +683,11 @@ class WorkOrders
 
   def request_work_order(recipes, npc_rooms, npc, npc_last_name, discipline, logbook, diff)
     match_names = recipes.map { |x| x['name'] }
+    $i = 0
```
jandersson: These dont need to be global variables

workorders.lic
```diff
@@ -683,9 +683,11 @@ class WorkOrders
...
+    $num = 500
 
     stow_hands
-    loop do
+    while $i <= $num do
```
jandersson: How does this loop exit? It looks like it runs 500 times no matter what.

jandersson: You can do `500.times do` which would remove the need for the loop variables

workorders.lic
```diff
@@ -715,7 +717,9 @@ class WorkOrders
         end
         fput('get logbook') unless [left_hand, right_hand].grep(/logbook/i).any?
       end
+      $i +=1
     end
+    exit
```
jandersson: careful with this. see https://github.com/rpherbig/dr-scripts/issues/952

shape.lic
```diff
@@ -152,6 +152,7 @@ class Shape
       stow_item(@noun)
     when /trash/
       dispose_trash(@noun)
+      stop_script('workorders')
```
jandersson: Same with `exit`. See https://github.com/rpherbig/dr-scripts/issues/952

Dartellum:
The case statements contain the returns when the workorder is found.
Working for my guys; I have not had a situation all day where they ask for
workorders endlessly.

Regards,
484-300-4833 (h)
484-431-3355 (m)

Dartellum:
This exit is only used once the 500 times completes without finding a
workorder.

Regards,
484-300-4833 (h)
484-431-3355 (m)

Dartellum:
This stop script kills workorders since shape bundles and not workorders
(like remedies bundle in workorders and not in remedy.lic). This only
occurs when the quality is not good enough for the workorder and you will
not make one on the next one that will.

Regards,
484-300-4833 (h)
484-431-3355 (m)

workorders.lic
rpherbig: In that case, you would also need a `break`. I'm a fan of `500.times do`.

shape.lic
rpherbig: It is unintuitive to me that the 'lower' script (`shape`) would stop its 'calling' script (`workorders`). I think a Flag in `workorders` would be a better choice, so it can watch for low quality items.

workorders.lic
rpherbig: This one should be ok since no crafting script is running. This loop is the sole responsibility of `workorders`, so it can exit. Make sure there's no logbook in hand, though.

shape.lic
Dartellum: Working on a better method.

workorders.lic
```diff
@@ -280,6 +280,10 @@ class WorkOrders
       end
 
       wait_for_script_to_complete('shape', ['log', recipe['chapter'], recipe['name'], info['stock-name'], recipe['noun']])
+      log_num = bput('read my log', 'You must bundle and deliver \d+ more').scan(/\d+/).first.to_i
+      if log_num == quantity
+        exit
```
rpherbig: Does `workorders` need to do any cleanup like lines 289 or 290 before it exits?

rpherbig: I wonder if this should `break` instead of `exit` and then `workorders` will clean up as it normally does? I'm not sure what other side effects that would have.

PR #2411, opened 2018-01-20
base-hunting.yaml: Expand rock guardians hunting area.

Sarvatt:


Sarvatt:
reordered them slightly also so you start from the entrace and move inwards, instead of starting from the very end of it.

PR #2412, opened 2018-01-20
Allow bin_town override in steal.lic

charleyc-dr:
Heard from someone who wants to run from Boar Clan to do Crossing stealing, so adding an override for that.

Also making it so that if an item is specified to the command, then we don't do NPC stealing on that run.


steal.lic
```diff
@@ -20,8 +20,8 @@ class Steal
 
     @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
     @stealing_bag = settings.stealing_bag
-    hometown_data = get_data('town')[settings.hometown]
-    @bin_id = hometown_data['thief_bin'] && hometown_data['thief_bin']['id']
+    bin_data = get_data('town')[settings.bin_town || settings.hometown]['thief_bin']
```
rpherbig: Can you add `bin_town` to `base.yaml`?

steal.lic
```diff
@@ -20,8 +20,8 @@ class Steal
...
+    @bin_id = bin_data && bin_data['id']
     @stealing_towns = settings.stealing_towns
     @bin_stolen = settings.bin_stolen
     if @bin_stolen && !@bin_id
```
rpherbig: Does line 28 also need to be updated?

jayRyan24:
Is this needed with the new changes to steal implemented?

rpherbig:
@charleyc-dr

PR #2419, opened 2018-01-20
pick.lic: Shut up outdated warning if lockpick_buffs is not set

Sarvatt:


pick.lic
```diff
@@ -174,16 +174,17 @@ class LockPicker
...
-      buffs['spells'].each do |spell|
-        echo "Buffing: #{spell}" if UserVars.lockpick_debug
-        cast_spell(spell, @settings)
-      end
-      buffs['khri']
-        .map { |name| "Khri #{name}" }
-        .each { |name| activate_khri?(@settings.kneel_khri, name) }
-      buffs['meditate'].each do |med|
-        bput("meditate #{med}", 'You feel a jolt as your vision snaps shut', 'Your inner fire lacks the strength')
+      unless buffs.empty?
```
jandersson: You might be missing a closing `end`

jandersson: You could `echo unless buffs.empty?` then you wouldn't need a closing `end`

PR #1933, opened 2017-09-20
Made spelling correction on junliar from junilar.

Dartellum:

PR #2426, opened 2018-01-22
Changed two remedy workorders from false to true. I made both this mo…

Dartellum:
…rning. I do not recall why I made them false.

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -3941,7 +3941,7 @@ crafting_recipes:
   noun: ointment
   volume: 1
   type: remedies
-  work_order: false
+  work_order: true
   chapter: 5
   container: mortar
   herb1: genich
@@ -3981,7 +3981,7 @@ crafting_recipes:
   noun: poultices
   volume: 1
   type: remedies
-  work_order: false
+  work_order: true
   chapter: 5
   container: mortar
   herb1: genich


rpherbig:
According to #1933, these two were removed because:
> I removed genich from workorders since Shard has it as grenich. I bugged it and posted in Lore folder. 

Has this bug been fixed?

Dartellum:
Good catch. Sadly, it is not corrected in Shard. I wonder if I want to
program around this since more recipes are needed so I do not loop through
s many times looking for one.

Regards,
484-300-4833 (h)
484-431-3355 (m)

Dartellum:
Actually, I just combined them and they turn into genich from Shard.
You get some dried grenich from inside your embroidered backpack.

                         |
>combine

                        |
You combine the stacks of herbs together
You glance down to see some dried genich in your left hand.

Regards,
484-300-4833 (h)
484-431-3355 (m)

On Tue, Jan 23, 2018 at 8:15 AM, Dr. Wes Snyder V <dr.wes.snyder.v@gmail.com

rpherbig:
If you're not sure why a line of code is the way it is, there is a good tool to help you investigate: https://git-scm.com/docs/git-blame. In short, `git blame` will tell you the last commit that changed each line. In some files that are changed a lot, that won't help. In files like `base-crafting` that don't change much, it's very effective.

As for grenich vs genich, did you run a workorder in Shard or just test combining the herbs by hand?

Dartellum:
Test combined by hand. My Remedies person is F2P. Dartellum does not have
the tools or skill to use.

Regards,
484-300-4833 (h)
484-431-3355 (m)

Issue #2243, opened 2017-12-18
Commonize crossing-trainer::cast_spell


jandersson:
It looks like all the logic in crossing-trainer can be moved to DRCA.cast_spell.

jandersson:
The two things left that I can see are: commonize the auto mana logic and handle composting.

One thing ive been wondering about casting compost is how to handle the `walk_to @settings.compost_room` line. I would think that should not be commonized and maybe crossing-trainer should handle that before the spell is cast? But im curious to hear any other thoughts.

MahtraDR:
This is fully in commons today...

PR #2428, opened 2018-01-23
Common-Arcana: Add locate prereq check

jandersson:
Locate requires clear vision active before casting. This logic is ported
from crossing-training::cast_spell.


jandersson:
Part of the effort of resolving #2243

PR #2429, opened 2018-01-23
Move nerve check to its own method

jandersson:
Part of commonizing the cast_spell method. The success check was removed
to allow for healing if any nerve damage exists.

One more item for #2243

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -217,6 +217,13 @@ class CrossingTraining
     end
   end
 
+  def check_nerves
+    wounds = check_health['wounds']
+    if wounds.any? { |_k, v| v.include?('skin') }
+      wait_for_script_to_complete('safe-room', ['force'])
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -518,12 +528,6 @@ class CrossingTraining
 
     success = cast?(data['cast'], data['symbiosis'], data['before'], data['after'])
 
-    wounds = check_health['wounds']
-    if !success && wounds.any? { |_k, v| v.include?('skin') }
-      wait_for_script_to_complete('safe-room', ['force'])
-      walk_to(@training_room)
```
rpherbig: I noticed this `walk_to` isn't in the extracted method. Was that intentional?

jandersson: Dangit, no its not. Ill get that added back in.

PR #2438, opened 2018-01-26
Adding a default stance, user editable.

jayRyan24:


profiles/base.yaml
```diff
@@ -34,6 +34,7 @@ stances:
   - Evasion
   - Shield Usage
   - Parry Ability
+default_stance: 100 0 80
```
rpherbig: Should common-arcana's `ritual` be updated to use this setting?

jayRyan24: Not sure that would work without a separate setting, my default stance would be to use parry and not shield use, but if you use parry with a focus, you are limited to only using Evasion and will get hit.

rpherbig: `ritual` currently uses `DRC.bput('stance set 100 0 80', 'Setting your')`. Are you saying it shouldn't do that?

jayRyan24: If the player changes it, which is what the entire point of this is, then it could in turn change this setting and ruin their ritual if it Parried with a not parry item in hand. This would need to be a ritual specific stance setting.

This is here because when combat starts, there is about 10s where the train sets this stance and then waits. This is terrible in higher levels as the things can start from polearm engage, and things like moths engage nearly instantly. You could get injured at the very beginning of battle. This default_stance is for that purpose. Now, the default_ritual_stance could be something different, and honestly should go up as the player gains Defending ranks.

rpherbig: That makes sense, thank you for the explanation

PR #2442, opened 2018-01-31
theurgy.lic: Specify pilgrim badge

Sarvatt:


rpherbig:
Should `pilgrimage.lic` also be updated?

Sarvatt:
good call, updated that too.

Issue #2451, opened 2018-02-04
Feature request: Allow overriding hometown for auto healers.


Sarvatt:
Everyone hunting snowbeasts and endrus serpents goes back to crossing for dokt because hometown: Leth  is so limited, it would be nice to override just the auto healer with a city closer to your hunting area.

Sarvatt:
would also be useful to use the leth auto healer instead of dokt on characters that can swim the segoltha.

PR #2454, opened 2018-02-04
safe-room.lic: Allow customizing which town to use an autopuff in.

Sarvatt:
Fixes: #2451

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -41,9 +41,7 @@ class SafeRoom
 
       town_data = get_data('town')
       if settings.force_healer_town
-        temp_settings = settings
-        temp_settings.hometown = settings.force_healer_town
-        hometown = town_data[temp_settings.hometown]
+        hometown = town_data[settings.force_healer_town]
       else
         hometown = town_data[settings.hometown]
       end
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -40,7 +40,13 @@ class SafeRoom
       Flags.add('idle', 'you have been idle too long')
 
       town_data = get_data('town')
-      hometown = town_data[settings.hometown]
+      if settings.force_healer_town
+        temp_settings = settings
+        temp_settings.hometown = settings.force_healer_town
+        hometown = town_data[temp_settings.hometown]
+      else
+        hometown = town_data[settings.hometown]
+      end
       wait_at_empath(hometown['npc_empath']['id'])
       fix_standing
     elsif settings.necro_safe_room_use_devour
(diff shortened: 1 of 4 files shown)


safe-room.lic
```diff
@@ -40,7 +40,13 @@ class SafeRoom
       Flags.add('idle', 'you have been idle too long')
 
       town_data = get_data('town')
-      hometown = town_data[settings.hometown]
+      if settings.force_healer_town
+        temp_settings = settings
+        temp_settings.hometown = settings.force_healer_town
```
rpherbig: Did you consider `hometown = town_data[settings.force_healer_town]`?

validate.lic
```diff
@@ -503,6 +503,18 @@ class DRYamlValidator
...
+  def assert_that_force_healer_town_is_valid(settings)
+    return unless settings.force_healer_town
+
+    valid_hometowns = get_data('town')
+                      .to_h
+                      .keys
+                      .map(&:to_s)
+
+    return if valid_hometowns.include? settings.force_healer_town
+    warn("Your force_healer_town setting of '#{settings.force_healer_town}' is not valid. Valid values are: #{valid_hometowns}")
```
rpherbig: Did you consider checking that the hometown also has an empath set?

Sarvatt: Done, and tested that it works.

rpherbig: Should this be `error` instead of `warn`?

validate.lic
```diff
@@ -503,6 +503,30 @@ class DRYamlValidator
...
+  end
+
+  def assert_that_force_healer_town_has_npc_empath(settings)
+    return unless settings.force_healer_town
+
+    town_data = get_data('town')
+    hometown = town_data[settings.force_healer_town]
+    if hometown['npc_empath'].nil?
+      echo('No npc empath exists in your force_healer_town so healing will not work, please pick another town.')
```
rpherbig: Should this be `error` instead of `echo`?

validate.lic
```diff
@@ -503,6 +503,30 @@ class DRYamlValidator
...
+                      .to_h
+                      .keys
+                      .map(&:to_s)
+
+    return if valid_hometowns.include? settings.force_healer_town
+    warn("Your force_healer_town setting of '#{settings.force_healer_town}' is not valid. Valid values are: #{valid_hometowns}")
+  end
+
+  def assert_that_force_healer_town_has_npc_empath(settings)
+    return unless settings.force_healer_town
```
rpherbig: Should this test be generalized to also check `settings.hometown` if there is no `force_healer_town`?

Sarvatt:
Echo changed to error, warn changed to error, and I made the npc empath validation check generic so it works for all settings. Verified it works by trying both hometown: Arthe Dale and force_healer_town: Arthe Dale independently.

PR #2456, opened 2018-02-04
Added message for wrong type of material.

Dartellum:
This catch is required if you make a cloth item and the work order wanted leather of the item.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -150,8 +150,8 @@ module DRCC
 
   def logbook_item(logbook, noun, container)
     DRC.bput("get my #{logbook} logbook", 'You get')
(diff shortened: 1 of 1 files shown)


rpherbig:
@Dartellum is there some change that needs to be made to `workorders.lic` so this doesn't happen in the future?

Dartellum:
If you mean for cloth versus leather, we could implement a new argument,
type or material, and check the workorder if it contains cloth if cloth or
leather depending on what the person uses and has customized their local
base-recipes. Maybe read the log when it is returned from getting a
workorder and checking if cloth or leather is listed.

Regards,
484-300-4833 (h)
484-431-3355 (m)

PR #2461, opened 2018-02-04
Added a capture so when done it does not wait 15 seconds and dump mes…

Dartellum:
…sage.

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -280,8 +280,11 @@ class WorkOrders
       end
 
       wait_for_script_to_complete('shape', ['log', recipe['chapter'], recipe['name'], info['stock-name'], recipe['noun']])
-      log_num = bput('read my log', 'You must bundle and deliver \d+ more').scan(/\d+/).first.to_i
-      break if log_num == quantity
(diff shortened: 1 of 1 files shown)


rpherbig:
Won't this only catch the case in which you fail the _first_ craft due to quality? That is, if a workorder needs 5 items and you succeed on the first but fail on the second, this case will never be triggered.

Is success/failure binary enough that this is ok?

Dartellum:
Right, just the first time. If you fail a subsequent one it will not catch.
Since the bundling is handled in sew, shape, knit, and maybe a few others,
trying to do like it is done for remedies is not feasible. Remedies are
bundled in workorders and not in remedy.lic. Since the called script should
not kill the calling script, this method is one way. Perhaps another
variable that is updated? The logic would still do log_num == quantity but
another variable is set to the log_num (that is one lower now), then check
against this new variable and then either break or set it to the log_num
until complete.

log_num == quantity
 break
log_num < quantity
  new_quantity = log_num

Maybe?

Regards,
484-300-4833 (h)
484-431-3355 (m)

PR #2464, opened 2018-02-05
crossing-training.lic: Don't train stealth above 30 ranks.

Sarvatt:
It is just a waste of time to train this higher, and is really obvious behavior. People think it actually trains when they first join lich without reading the scripting reference saying its low ranks only leading to spam at the empaths guild.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -738,11 +738,17 @@ class CrossingTraining
   end
 
   def train_stealth
-    hide_in 851
-    hide_in 850
-    hide_in 764
-    hide_in 5992
-    hide_in 992
+    if DRSkill.getrank('Stealth') >= 25
+      echo('Skill too high to learn this way, removing Stealth from training')
+      @settings.crossing_training.delete('Stealth')
+      nil
+    else
+      hide_in 851
+      hide_in 850
+      hide_in 764
+      hide_in 5992
+      hide_in 992
+    end
   end
 
   def train_thievery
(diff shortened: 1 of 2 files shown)


crossing-training.lic
```diff
@@ -738,11 +738,17 @@ class CrossingTraining
...
   def train_stealth
-    hide_in 851
-    hide_in 850
-    hide_in 764
-    hide_in 5992
-    hide_in 992
+    if DRSkill.getrank('Stealth') >= 25
+      echo('Skill too high to learn this way, removing Stealth from training')
+      @settings.crossing_training.delete('Stealth')
+      nil
```
Sarvatt: hmm this was return but rubocop changed it to nil

jandersson: Since ruby returns the last thing evaluated, `return` and `nil` would be the same

crossing-training.lic
```diff
@@ -738,11 +738,16 @@ class CrossingTraining
...
 
   def train_stealth
-    hide_in 851
-    hide_in 850
-    hide_in 764
-    hide_in 5992
-    hide_in 992
+    if DRSkill.getrank('Stealth') >= 30
+      echo('Skill too high to learn this way, removing Stealth from training')
+      @settings.crossing_training.delete('Stealth')
```
rpherbig: I prefer something like
```
if ...
    ...
    return
end

hide_in ...
```

It's a completely stylistic thing, so don't worry about changing it, I just wanted to point out an alternative.

Sarvatt: that's what I had originally, I'll use that style in the future.

PR #2468, opened 2018-02-05
mm: Updates from user feedback

jandersson:
Return false if If the lookup for the message 'You gesture' fails; the the array
slicing will fail.

Return a room id in `locate?` as the truthy value


jandersson:
Also included an example for the `locate` and `rs` spell sets in my profile

PR #2472, opened 2018-02-07
Organize base hunting

jayRyan24:
Organized by

1. Province
2. Min rank, descending


rpherbig:
Thanks for doing this, it's long overdue.

Can you confirm that this is purely a refactor (i.e. no data was modified)?

jayRyan24:
The base hunting data hasn't been modified except for comments. The only thing that could cause issues is my anchors which I'm going to use to delete terrible names like orc_warcats->orc_scouts, or getting things along the same naming convention like xalas_1, xalas_2, xalas_3 instead of xalas, xala_conjuror, Xala_3, or whatever terrible names were given. Some I left alone, a project for another day, but I changed the ones that piss me off enough.

PR #2475, opened 2018-02-07
base-hunting.yaml: Fix up anchor usage

Sarvatt:


Sarvatt:
yes, and I ran hz after to verify all the rooms were correct

PR #2478, opened 2018-02-08
Copy deposit coins method to DRCM

jandersson:
This is copied from sell-loot. Having it in common-money will allow
other scripts to use it.


rpherbig:
Do you have any other scripts in mind that would use this, or is this future-proofing?

jandersson:
I was looking at how to deposit coins withdrawn after repairing and realized there was no common method for this.

rpherbig:
Will this version integrate with the one in `trade.lic`?

jandersson:
Yea trade is another script that would use it.

PR #2479, opened 2018-02-08
Crossing-Repair: Deposit coins after repairing

jandersson:
When done repairing, deposit all coins that were withdrawn to pay for
the repairs. If the user has a high withdraw amount set, there could be
a good deal left over coin.

### DIFF
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -21,6 +21,8 @@ class CrossingRepair
       while tickets_left?(repairer)
       end
     end
+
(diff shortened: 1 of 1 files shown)


crossing-repair.lic
```diff
@@ -90,14 +92,18 @@ class CrossingRepair
     @equipment_manager.empty_hands
 
     settings = get_settings
+    @hometown = settings.hometown
+    @skip_bank = settings.skip_bank
```
jandersson: This is the wrong variable name, just a moment while i correct this.

PR #2481, opened 2018-02-08
test_afk: Update test_harness and test_afk

jandersson:
Tests were failing due to missing methods in the Script class.
The get_settings method is often called without an argument, i.e.
`settings = get_settings` which wasn't matching the signature in
Script::get_settings.

Copied the setup_settings method from another test file to test_afk
so that the script is ran with base.yaml loaded.

Added a new test to check for exiting with default settings enabled and
modified current tests to use updated settings such as depart_on_death.

### DIFF
--- a/test/test_afk.rb
+++ b/test/test_afk.rb
@@ -54,11 +54,15 @@ class TestAfk < Minitest::Test
   end
 
   def test_exits_if_low_health
-    skip "Afk has changed and this test needs to be updated to modify health while its running"
     setup_settings({})
+    $history = ['You continue to braid your vines.', 'You are attacked by a Grue! Your injuries are severe.']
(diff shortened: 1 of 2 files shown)


test/test_afk.rb
```diff
@@ -34,10 +54,13 @@ def test_departs_if_dead
   end
 
   def test_exits_if_low_health
+    setup_settings({})
     expected_messages = ['health', 'avoid all', 'exit']
-    self.health = 20
 
     run_script('afk')
+    sleep 0.4
```
jandersson: This sleep is causing an issue with threading. Im going to update this test so that it is skipped instead. `afk` pauses during start up until health and spirit are above a threshold. For this test to work, there needs to be a way to modify the health while `afk` is running.

PR #2483, opened 2018-02-09
Add tests for check_health

jandersson:
This adds tests for DRCH::check_health and updates the test harness with a reget method, since reget is used by check_health.

### DIFF
--- a/test/test_check_health.rb
+++ b/test/test_check_health.rb
@@ -1,0 +1,95 @@
+require 'minitest/autorun'
+load 'test/test_harness.rb'
+
+include Harness
+
+class TestCheckHealth < Minitest::Test
+  def setup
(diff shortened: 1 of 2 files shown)


test/test_check_health.rb
```diff
@@ -0,0 +1,95 @@
...
+        'Empath whispers, "Done!"',
+        'Your body feels at full strength.',
+        'Your spirit feels full of life.',
+        'You have no significant injuries.'
+    ]
+    $history = $server_buffer.dup
+    assert_healthy(DRCH.check_health)
+  end
+
+  def test_that_diseased_person_is_diseased
```
jandersson: This test will fail. There is something up with these particular messages that are throwing off check_health. Yay testing :)

PR #2484, opened 2018-02-09
Update check_health so tests pass

jandersson:
* gsub! doesn't work on nil (unless it was overloaded >.>)
* Moved wound_line and parasite_line logic inside an if block to prevent
operating on nil
* New disease line to check for
* ~An issue with checking health and being completely wound free but
but having wounds earlier in the message buffer. Added an explicit check
for 'You have no wounds'~
* The vitality is the last line in the reversed message history. Check_health will break on that instead.


jandersson:
Failed a new test - will need to keep working on it :)

PR #2485, opened 2018-02-09
Add healer script

jandersson:
Based on fest-healer. Modified to use healme, common-healing and
common-arcana to offload lots of the logic. Additions include the
pausing and starting of scripts when healing, meaning one could use this
while running crossing-training.

### DIFF
--- a/healer.lic
+++ b/healer.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#healer
 =end
 no_pause_all
-custom_require.call %w[events common common-arcana common-healing common-items]
+custom_require.call %w[common common-arcana common-healing common-items]
 
 class Healer
   include DRC
--- a/healer.lic
+++ b/healer.lic
@@ -1,5 +1,5 @@
 =begin
-  Documentation:
+  Documentation: https://elanthipedia.play.net/Lich_script_development#healer
 =end
 no_pause_all
 custom_require.call %w[events common common-arcana common-healing common-items]
--- a/healer.lic
+++ b/healer.lic
@@ -1,0 +1,139 @@
+=begin
+  Documentation:
+=end
+no_pause_all
+custom_require.call %w[events common common-arcana common-healing common-items]
+
+class Healer
+  include DRC
+  include DRCH
+  include DRCI
+  include DRCA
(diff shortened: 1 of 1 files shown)


healer.lic
```diff
@@ -0,0 +1,139 @@
...
+    echo "Healing #{name}" if debug?
+    @patient = name
+    return if @touched && @touched >= Time.now
+
+    unless @touched
+      @touched = Time.now + 12
+      pause 1
+      waitrt?
+      fput("touch #{name}")
+      put("transfer #{name} vit quick")
```
bmcneff: fput, not put

jandersson: This is from the original script. My understanding is that put is when we dont care about a response from the game. As someone using the script for several hours, this was never an issue.

bmcneff: 👍

jandersson: Appreciate the review nonetheless :) I think the argument on `put` for vitality is, 'we'll get it if we get it, its time to move on to wounds'

healer.lic
```diff
@@ -0,0 +1,139 @@
+=begin
+  Documentation:
```
rpherbig: Will this be documented on elanthipedia?

healer.lic
```diff
@@ -0,0 +1,139 @@
...
+=end
+no_pause_all
+custom_require.call %w[events common common-arcana common-healing common-items]
```
rpherbig: Why is `events` required?

healer.lic
```diff
@@ -0,0 +1,139 @@
...
+
+class Healer
+  include DRC
+  include DRCH
+  include DRCI
+  include DRCA
+
+  def empty_hands
```
rpherbig: What do you think about using `stow_hands` from DRCI instead?

healer.lic
```diff
@@ -0,0 +1,139 @@
...
+    echo 'Emptying hands' if debug?
+    @left = checkleft
+    echo "@left: #{@left}" if debug?
+    @right = checkright
+    echo "@right: #{@right}" if debug?
+    fput('stow left') if checkleft
```
rpherbig: What do you think about using `stow_hands` from DRCI here?

PR #2487, opened 2018-02-10
Fixed ignite.

Dartellum:


profiles/Dartellum-setup.yaml
```diff
@@ -348,7 +348,10 @@ buff_spells:
     - - 20
     harmless: true
     cast: cast air
-    Ignite:
+  Ignite:
+    mana: 20
+    cambrinth:
+    - - 30
```
rpherbig: Does this syntax work? I would have expected `- 30`

jayRyan24: @Dartellum

Dartellum: Cambrinth is now an array, thus the double dashes.

PR #2494, opened 2018-02-11
join-thieves.lic: Update to handle various states of joining, asking …

Sarvatt:
…3 times is required.

also include DRCT for walk_to to work.


Sarvatt:
tested this out on 2 commoners to be sure it worked, pretty sure I got all the messaging the beggar can give outside of when you can actually join for the bput there. Asking about thieves before asking about join 3 times is required and was missing.

PR #2496, opened 2018-02-11
Combat trainer - health threshold

rpherbig:
Fixed #2239

@Sarvatt can you test this for me?


Sarvatt:
I set health_threshold: 99 in a backhunt and it logged me out on the first hit, looks to be working properly.

PR #2498, opened 2018-02-11
Trash only the item forged

jandersson:
This helps prevents throwing out items that may have ended up stuck
in hands such as valuable crafting tools. The smith script already
disposes the ingot.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -979,8 +979,8 @@ class CrossingTraining
 
   def smith(full_name)
     wait_for_script_to_complete('smith', ['bronze', full_name, 'buy'])
-    dispose_trash(right_hand)
-    dispose_trash(left_hand)
+    noun = get_data('recipes')['crafting_recipes'].select { |item| item['name'] == full_name }.first['noun']
+    dispose_trash(noun)
   end
 
   def buy_wood


jandersson:
I did this after a background script caused me to end up with my makers mark in my hands at this point, which was thrown out.

crossing-training.lic
```diff
@@ -979,8 +979,8 @@ class CrossingTraining
 
   def smith(full_name)
     wait_for_script_to_complete('smith', ['bronze', full_name, 'buy'])
-    dispose_trash(right_hand)
-    dispose_trash(left_hand)
+    noun = get_data('recipes')['crafting_recipes'].select { |item| item['name'] == full_name }.first['noun']
```
rpherbig: Should this use `recipe_lookup` (from `DRCC`) like smith.lic does?

jandersson: I wasn't aware of recipe_lookup - that sounds like a good idea to me. I'll update this to use that function instead.

jandersson:
Oh boy, looks like a lot has changed. This might still be worth closing.

jandersson:
Closing because all crafting has been offloaded to separate scripts.

PR #2500, opened 2018-02-11
base-shrines.yaml: Add Berengaria altar south of leth.

Sarvatt:


data/base-shrines.yaml
```diff
@@ -74,7 +74,10 @@ shrines:
     path_out:
     - w
     - out
-  Therenborough:
+  - The altar to Berengaria in the Gash
+    start_room: 2258
+    push: altar
+ Therenborough:
```
rpherbig: This indentation looks off to me, is there a space missing?

Sarvatt: there is indeed a space missing in front of Therenborough:

PR #2505, opened 2018-02-12
Updated to catch prompts separated by spaces, such as J U M P

adamgoerdt:


status-monitor.lic
```diff
@@ -132,12 +132,14 @@ class GameFilter
...
   def command_check(line)
     line.gsub(/_|~|-|=|\./, '')
-        .scan(/[A-Z]{3,}/)
-        .select { |item| @valid_commands.include?(item) }
-        .each do |prompt|
-          pause 3
-          fput(prompt)
-          pause 5
+        .scan(/[A-Z\s]{3,}/)
+        .each do |match|
```
rpherbig: It's a style thing, but I'd prefer to see some more method chaining here. Something like `.scan{...}.map { |match| match.gsub(/\s/,'') }.each {...}`.

The `if` check could also be extracted into `.select { |match| @valid_commands.include?(match) }`.

PR #2507, opened 2018-02-12
Discern.lic

Rhayle:
A script to be added to autorun, if desired, capable of running discern and populating the discern array before executing other scripts.


discern.lic
```diff
@@ -0,0 +1,46 @@
+custom_require.call(%w[common common-travel events spellmonitor drinfomon])
```
rpherbig: Please add a documentation header so we can put additional information on elanthipedia

discern.lic
```diff
@@ -0,0 +1,46 @@
...
+    include DRC
+    include DRCA
+
+    def initialize()
+    settings = get_settings
+    if(settings.force_discerns)
+        UserVars.discerns = {}
+    end
+    echo 'Reading spell list'
+    buffspelllist = settings.buff_spells
```
rpherbig: What do you think about `buff_spell_list`? (and similar for the next two vars).

discern.lic
```diff
@@ -0,0 +1,46 @@
...
+    offspelllist =  settings.offensive_spells
+    ctspelllist = settings.training_spells
+     buffspelllist.each {|x| if (x[1]['use_auto_mana'])
```
rpherbig: What do you think about extracting the `if` into a chained method like `buff_spell_list.select{|x| (x[1]['use_auto_mana'])}.each {|x| ...}`?

discern.lic
```diff
@@ -0,0 +1,46 @@
...
+        @abbrev = x[1]['abbrev'] ? x[1]['abbrev'] : x[0]
```
rpherbig: Why is this an instance var (`@`)? Could it be local instead (`abbrev`)?

discern.lic
rpherbig: This also applies to the other blocks

discern.lic
rpherbig: This also applies to the other blocks

rpherbig:
@jandersson do you have a few minutes to review this?

astrology.lic
```diff
@@ -46,14 +46,21 @@ class Astrology
...
     @equipment_manager.empty_hands
     # Pop out rtr data from buffs and save it for later
     buffs = settings.waggle_sets['astrology']
     @rtr_data = buffs.select { |spell| spell.eql?('Read the Ripples') }
                      .values
                      .first
     buffs.reject! { |spell| spell.eql?('Read the Ripples') }
+    
+    buffs.keys.each{ |key| 
+        if(buffs[key]['use_auto_mana'])
```
rpherbig: Same comment here about method chaining, something like `buffs.keys.select{|key| buffs[key]['use_auto_mana']}.each{...}`

rpherbig: Or maybe `buffs.select{|key, _v| buffs[key]['use_auto_mana']}.each_value{|value| check_discern(value,settings)}`

Since this is an array, you don't really need to `.keys` and then later `buffs[key]`.

buff.lic
```diff
@@ -30,6 +30,12 @@ class Waggle
     elsif DRStats.thief?
       start_khris(spells, settings)
     else
+      spells.keys.each{ |key|
```
rpherbig: Same comment here about method chaining

astrology.lic
```diff
@@ -46,14 +46,21 @@ class Astrology
...
+    
     @equipment_manager.empty_hands
     # Pop out rtr data from buffs and save it for later
     buffs = settings.waggle_sets['astrology']
     @rtr_data = buffs.select { |spell| spell.eql?('Read the Ripples') }
                      .values
                      .first
     buffs.reject! { |spell| spell.eql?('Read the Ripples') }
+    
+    buffs.keys.each{ |key|
```
jandersson: I think rubocop will change this to `each_key`. Would you mind running rubocop on this? if you do `bundle install` while in the repository root directory it will install rubocop. Then you can do `mv astrology.lic astrology.rb` run rubocop in autocorrect mode, `rubocop -a astrology.rb` then `mv astrology.rb astrology.lic`. Alternatively you can do `rubocop -a` which will autocorrect all scripts.

discern.lic
```diff
@@ -0,0 +1,46 @@
+custom_require.call(%w[common common-travel events spellmonitor drinfomon])
+
+class AutoDiscern
+    include DRC
+    include DRCA
+
+    def initialize()
+    settings = get_settings
+    if(settings.force_discerns)
```
rpherbig: Is this a new setting? It feels more like it should be an argument to the script...

discern.lic
```diff
@@ -0,0 +1,46 @@
...
+
+    def initialize()
+    settings = get_settings
+    if(settings.force_discerns)
+        UserVars.discerns = {}
+    end
+    echo 'Reading spell list'
+    buffspelllist = settings.buff_spells
+    offspelllist =  settings.offensive_spells
+    ctspelllist = settings.training_spells
```
rpherbig: Do you need to store each of these spell lists into a local var? What do you think about something like `settings.training_spells.each {|x| ...}` instead?

discern.lic
jandersson: If its a new setting, it needs to be added to base.yaml

Rhayle:
I cannot get it to run rubocop.

discern.lic
```diff
@@ -0,0 +1,40 @@
...
+
+    ctspelllist.keys.each do |x|
+      next unless ctspelllist[x]['use_auto_mana']
+      @abbrev = ctspelllist[x]['abbrev']
+      echo "Discerning: '#{@abbrev}'"
+      check_discern(ctspelllist[x], settings)
+      echo " Current Discern data valid until '#{UserVars.discerns[@abbrev]['time_stamp'] + 24 * 60 * 60}'"
+    end
+end
+end
```
jandersson: This might throw an error, I can find a matching statement that begins this

jandersson:
Youll want to expand all the previous comments by clicking 'show outdated' as rubocop seems to have beautified your code quite a bit :)

Rhayle:
I finally got rubocop to run.

astrology.lic
```diff
@@ -54,6 +54,11 @@ class Astrology
                      .values
                      .first
     buffs.reject! { |spell| spell.eql?('Read the Ripples') }
+
+    buffs.keys.each do |key|
```
rpherbig: What do you think about using the hash directly instead of `.keys`? Something like this:
```ruby
    buffs.each do |key, value|
      check_discern(value, settings) if value['use_auto_mana']
    end
```

rpherbig: What do you think about using method chaining for the `if`? Something like this:
```ruby
    buffs.select { |_key, value| value['use_auto_mana'] }
         .each { |_key, value| check_discern(value, settings) }
```

discern.lic
```diff
@@ -0,0 +1,42 @@
...
+
+custom_require.call(%w[common common-travel events spellmonitor drinfomon])
+
+class AutoDiscern
+  include DRC
+  include DRCA
+
+  def initialize
+    settings = get_settings
+    UserVars.discerns = {} if settings.force_discerns
```
rpherbig: I'm not sure what I think of `force_discerns` being a setting. That implies that you want to re-discern every time this script is run. Is that a common use case?

This feels more like a script argument to me, so you can specify it (or not) differently each time the script is run. What do you think?

discern.lic
```diff
@@ -0,0 +1,42 @@
...
+    echo 'Reading spell list'
+
+    settings.buff_spells.each do |x|
```
rpherbig: I have a similar thought here about method chaining, something like:
```ruby
settings.buff_spells
    .select { |x| x[1]['use_auto_mana'] }
    .each do |x| ...
```

What do you think?

discern.lic
Rhayle: grumble no I thought I removed the force setting it was only for me to test with.  i figure you can run ;vars delete discerns instead of this setting

astrology.lic
Rhayle: can you explain this syntax with the if statement. It looks to me like that would be applied to everything following the statement than what is before the statement.

rpherbig: The original code iterates over every item, but only runs `check_discern` if a condition is met.

The `.select` syntax only keeps items for which the condition is met, then runs `check_discern` over everything that was kept.

See https://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-select for the technical details.

Rhayle:
ok. See what you think of this. I've made requested changes and fixed errors. The .select .each did not work in part of the discern scripts so I corrected to use key value

.ruby-version
```diff
@@ -1 +1 @@
-2.0.0-p647
+2.4
```
rpherbig: Was this supposed to be checked in?

Rhayle: no. sorry. I thought I had it in the ignores.
How do I remove it?

rpherbig:
Sorry for the confusion there. Thank you so much for your contribution @Rhayle !

Issue #282, opened 2016-03-01
Integrate with CircleCI or TravisCI


rpherbig:

jandersson:
What were the permission issues with travis?

rpherbig:
You have to give travisCI access to _every_ repo you have access to. I.e. if I'm in a group for both my personal repos (`rpherbig/*`) and work repos (`sep/*`), travis gets them all.

PR #2509, opened 2018-02-12
Add travis-ci config

jandersson:
This is a simple config file that would let us use travis-ci for this repository. With that we can run tests against all pull requests, potentially run rubocop on them. I think this might help mitigate future breakages. I was able to run this successfully on my fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: ruby
+rvm:
+- 2.0.0


rcuhljr:
We looked into this a few years ago, at the time it didn't work without giving travis-ci unreasonable permissions, but @rpherbig did the investigation.

rcuhljr:
https://github.com/rpherbig/dr-scripts/issues/282

jandersson:
I see - well at the least I can keep running it on my fork to mitigate any mistakes i might make :P

rpherbig:
You have to give travisCI access to _every_ repo you have access to. I.e. if I'm in a group for both my personal repos (`rpherbig/*`) and work repos (`sep/*`), travis gets them all.

jandersson:
Oh right, that is still a thing

PR #2513, opened 2018-02-13
Sets the threshold to 0 for necro health.

jayRyan24:
Sets the threshold to 0 for necro health to clean even if not needed by settings.

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -14,7 +14,6 @@ class SafeRoom
 
   def initialize
     arg_definitions = [
-      [ { name: 'force', regex: /force/i, description: 'Run healing logic even if there are no visible wounds.' } ],
       [ 
         { name: 'skip', regex: /(s|skip)/i, description: 'Skip running to the safe room, useful for Empaths and Necros.' },
         { name: 'force', regex: /force/i, optional: true, description: 'Run healing logic even if there are no visible wounds. (Optional)' }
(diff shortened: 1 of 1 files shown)


safe-room.lic
```diff
@@ -14,10 +14,15 @@ class SafeRoom
 
   def initialize
     arg_definitions = [
-      [{ name: 'force', regex: /force/i, description: 'Run healing logic even if there are no visible wounds.' }],
-      [{ name: 'skip', regex: /(s|skip)/i, description: 'Skip running to the safe room, useful for Empaths and Necros.' }],
+    arg_definitions = [
+      [ { name: 'force', regex: /force/i, description: 'Run healing logic even if there are no visible wounds.' } ],
```
rpherbig: Would it be easier to just set `force` to `optional: true` instead of adding two more arg sets?

safe-room.lic
```diff
@@ -14,10 +14,15 @@ class SafeRoom
 
   def initialize
     arg_definitions = [
```
rpherbig: Are there now two `arg_definitions = [` lines?

safe-room.lic
jayRyan24: I thought if we did this the force would require the skip first?

safe-room.lic
jayRyan24: It's a copy and paste error.

PR #2516, opened 2018-02-14
Update Bescort.lic

BinuDR:
Use equipment manager to change into swimming gearset before swimming the Segoltha, then change back to standard when done.  My swimming gearset is my standard set minus my torso armor.  It makes swimming much easier.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -744,7 +744,7 @@ class Bescort
         break
       end
     end
-    EquipmentManager.new.wear_equipment_set?('swimming')
+    EquipmentManager.new.wear_equipment_set?('standard')
   end
 
   def croc_swamp(mode)
--- a/bescort.lic
+++ b/bescort.lic
@@ -716,6 +716,8 @@ class Bescort
   end
 
   def segoltha(mode)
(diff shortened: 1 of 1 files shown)


bescort.lic
```diff
@@ -716,6 +716,8 @@ class Bescort
   end
 
   def segoltha(mode)
+    EquipmentManager.new.empty_hands
+    EquipmentManager.new.wear_equipment_set?('swimming')
```
rpherbig: What does this do if there is no `swimming` set defined? I assume it either would remove all gear or change nothing?

BinuDR: Yes.

rpherbig: Sorry, I'm not sure I understand. Which thing does it do if that set is not defined?

BinuDR: The Yes was in response to your original inquiry of if the 3rd addition was supposed to be:
'EquipmentManager.new.wear_equipment_set?('standard')'

If there is no 'swimming' gearset defined, it does this:
`echo("Could not find gear set #{set_name}")`

PR #2517, opened 2018-02-14
base-filters.yaml: Add some more various messages to ignore.

Sarvatt:


data/base-filters.yaml
```diff
@@ -413,3 +413,9 @@ filter_strings:
 - '^\w+ shrugs and says, "I will not repair something that isn.t broken\."$'
 - '^(?:One|A few) small orbiting sliver of lunar magic'
 - '^I could not find what you were referring to\.$'
+- '^You sense that you were able to channel'
+- '^That would accomplish nothing\.$'
+- '^The translucent image of a .* glistens briefly in the .* depths of the orb hovering near you\.$'
+- '^.* came through a town gate\.$'
+- '^.* came through the western gate\.$'
```
rpherbig: Is this the same as line 137?

Sarvatt: hmm, indeed. wonder why it showed up in debug.

Sarvatt:
Ok, done adding stuff :)

PR #2518, opened 2018-02-14
discern.lic: Convert crlf line endings to lf to fix up git on non-windows.

Sarvatt:


Sarvatt:
git tries to fix this up automatically and discern.lic always has changes that need to be commited otherwise.

PR #2522, opened 2018-02-15
 Added checks for mixed-case prompts

adamgoerdt:


status-monitor.lic
```diff
@@ -139,10 +139,21 @@ class GameFilter
...
           if @valid_commands.include?(match)
             pause 3
-            fput(match)
+            fput(match.downcase)
             pause 5
           end
         end
+
+    ## Look for mixed-case prompts like J_u_M_p
+    tokens = line.split(" ")
```
rpherbig: Some method chaining and/or Rubocop might make this a bit clearer, but that's a style thing

Issue #2000, opened 2017-10-23
Use lich fork by default


rpherbig:
It's not quite as simple as adding a call to `use_lich_fork` to the end of https://github.com/rpherbig/dr-scripts/blob/master/dependency.lic#L1089

We'd want to remove some of the announces and other stuff. Remove the echos (probably).

Change this to true? https://github.com/rpherbig/dr-scripts/blob/master/dependency.lic#L181

Can we remove the use_lich_fork, use_mainline, and toggle_lich_fork?

This would only catch new users. How do we catch all the people not using it already?

MahtraDR:
@rpherbig Shall we resume pursuing this? The initial install is the single most complex part our new users face, and the fork switch catches a lot of people.

rpherbig:
I'm in favor of simplifying the setup process. Everyone really should be using the fork.

MahtraDR:
@rpherbig @KatoakDR What do you think here? The current recommended install setup uses `dr-scripts_install` to automate this.

It's obviously not the same thing as having use_lich_fork be the default, but it is also working quite well now. 

Revive this? With @KatoakDR's changes to dependency, and the upcoming changes to downloading files, maybe we could revisit having the use_lich_fork toggle be the default?

rpherbig:
The new setup script uses the fork right?

MahtraDR:
> The new setup script uses the fork right?

Yes, the setup script enables the fork for new users. 

What would be a benefit would be if dependency itself  checked lich version, and gave a warning if it doesn't detect the fork. Even if it doesn't auto-toggle the fork, users would know. (I'd love that check at login, make dependency require the fork to run beyond basic functions)

MahtraDR:
@rcuhljr @KatoakDR, @rpherbig and myself were discussing this topic again. We're discussing folding the process in https://github.com/rpherbig/dr-scripts/wiki/First-Time-Setup into `;dependency install` and the topic came up as to why it wasn't already that way. If there was a reason we weren't aware of. @rpherbig Suggested we ask you. Any opinions/ideas here? Thanks!

rcuhljr:
Early on the fork was less divergent from the mainline and less important to proper lich functionality, also it felt kind of presumptuous to me at the time to say 'hey we're also going to swap out the entire lich you're using'. kind of thing. I suspect that ship has sailed. We had no reasons for not doing it besides philosophical. I can't imagine any technical reason that would limit the change.

[...]

PR #2523, opened 2018-02-15
setupalias.lic: Use e instead of en in hz ez fz aliases so it works w…

Sarvatt:
…hen not using the lich fork.

These work fine with just ;e, they only need get_data from dependency that is already trusted.


Sarvatt:
nope, that never happened

rcuhljr:
I vote we do that instead (because I'm not doing anything).

rpherbig:
Lucky #2000

PR #2524, opened 2018-02-16
Fixing the safe-room force issues.

jayRyan24:
Adding in the previous force logic as well as the new option skip and then force that necros and empaths can use.

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -16,7 +16,7 @@ class SafeRoom
     arg_definitions = [
       [ 
         { name: 'skip', regex: /(s|skip)/i, optional: true, description: 'Skip running to the safe room, useful for Empaths and Necros.' },
-        { name: 'force', regex: /(f|force)/i, optional: true, description: 'Run healing logic even if there are no visible wounds. (Optional)' }
+        { name: 'force', regex: /(f|force)/i, optional: true, description: 'Run healing logic even if there are no visible wounds.' }
       ]
     ]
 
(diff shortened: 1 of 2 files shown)


rcuhljr:
It needs testing, but I think this is all you should need.
```
 arg_definitions = [
      [
        { name: 'skip', regex: /(s|skip)/i, optional:true, description: 'Skip running to the safe room, useful for Empaths and Necros.' },
        { name: 'force', regex: /force/i, optional: true, description: 'Run healing logic even if there are no visible wounds.' }
      ]
    ]
```

jayRyan24:
I've done dual optional thing, and it wouldn't work.

jayRyan24:
This might work though, I'll test this tonight. 

```
 arg_definitions = [
      [
        { name: 'skip', regex: /(s|skip)/i, optional:true, description: 'Skip running to the safe room, useful for Empaths and Necros.' },
        { name: 'force', regex: /force/i, optional: true, description: 'Run healing logic even if there are no visible wounds.' }
      ], []
    ]
```

rcuhljr:
Are you sure you weren't trying it with flex args enabled? those might mess with it but they shouldn't. I don't see any reason all optional args won't work.

jayRyan24:
I'll try again tonight, but I've had issues with using full optionals. I think it happened last in the paladin quest walkthrough script, don't remember the exact conditions.

jayRyan24:
OK, so my issues before were the separated settings

```
[
  setting one optional
],
[
  setting two optional
]
```
thanks for the help.

PR #2527, opened 2018-02-16
Added ^Your spell horribly backfires.

Dartellum:


data/base-spells.yaml
```diff
@@ -81,6 +81,7 @@ prep_messages:
 cast_messages:
 - ^Your spell backfires
 - Your spell barely backfires
+- ^Your spell horribly backfires
```
rpherbig: Could lines 82-84 be combined into `- ^Your spell .*backfires`?

Dartellum:
Changed. I do not see why not.

PR #2528, opened 2018-02-16
Added message for picking up bone stacks and added more logic for det…

Dartellum:
…ecting failure to bundle to exit.


rpherbig:
@Dartellum how was this tested?

Dartellum:
Testing on all my 7 of my guys. I added the break today removing the exit so the remedy tools are repaired instead of existing on failure. Same logic as shaping, knitting, and sewing.

PR #2530, opened 2018-02-16
Add offhand_thrown

jandersson:
Adds a missing setting to base


profiles/base.yaml
```diff
@@ -129,6 +129,7 @@ ambush_location: back
 warhorn:
 prayer_mat:
 paladin_use_badge: false
+offhand_thrown:
```
rpherbig: What do you think of making this `offhand_thrown: false` instead?

jandersson: Makes sense to me

PR #2531, opened 2018-02-16
Add pray_to_chadatru

jandersson:
Adds a missing setting to base.yaml


profiles/base.yaml
```diff
@@ -822,6 +822,7 @@ theurgy_exp_threshold: 33
 theurgy_bead_carving_tool:
 theurgy_use_prayer_mat: false
 theurgy_prayer_mat_room:
+pray_to_chadatru: false
```
rpherbig: Could you also remove `pray_at_chadatru_altar` on line 260?

jandersson: Sure thing

PR #2532, opened 2018-02-16
Update test_all_settings_in_base_yaml

jandersson:
Check membership in base-empty.

Ignore profiles with custom/non-repository settings.


jandersson:
One alternative to skipping peoples profiles entirely was to enumerate all the custom settings. Let me know if that's preferred.

rpherbig:
There are an infinite number of custom settings. This seems fine.

PR #2545, opened 2018-02-24
cleric-quests.lic: Fix check for eluneds and add an echo

Sarvatt:
2018-01-19 14:28:30 -0500:[cleric-quests: *** No match was found after 15 seconds, dumping info]
2018-01-19 14:28:30 -0500:[cleric-quests: messages seen length: 2]
2018-01-19 14:28:30 -0500:[cleric-quests: message: The vision fades, and you find that you cannot move yourself to go in.]
2018-01-19 14:28:30 -0500:[cleric-quests: message: You stare deep into the deeper water, focusing on your reflection.  The eyes of the man in the water stare back up at you, and the surface suddenly breaks, revealing a younger$
2018-01-19 14:28:30 -0500:[cleric-quests: checked against [/You take a deep breath/i]]
2018-01-19 14:28:30 -0500:[cleric-quests: for command dive deeper water]

### DIFF
--- a/cleric-quests.lic
+++ b/cleric-quests.lic
@@ -138,8 +138,9 @@ class ClericQuests
     bput('meditate', 'You close your eyes and begin to meditate')
     waitfor('once again changing and reforming into your Human form')
 
-    case bput('dive deeper water', 'You take a deep breath', 'You get a sense that you have completed this')
-    when 'You get a sense that you have completed this'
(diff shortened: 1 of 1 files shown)


Sarvatt:
sitting around for 5 minutes with no feedback just makes it look like the script is broken.

PR #2547, opened 2018-02-26
Changed up the necro material logic. 

jayRyan24:
It will no longer check the count every iteration, checks at the beginning and then fills the missing material.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -288,6 +288,9 @@ class LootProcess
     @necro_store = thanatology['store'] || false
     echo("  @necro_store: #{@necro_store}") if $debug_mode_ct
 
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -288,6 +288,9 @@ class LootProcess
     @necro_store = thanatology['store'] || false
     echo("  @necro_store: #{@necro_store}") if $debug_mode_ct
 
+    @current_harvest_count = rummage('C material', @necro_container).size if DRStats.necromancer?
```
rpherbig: `@necro_container` looks like it is initialized _after_ it is used here (see line 294). Can you confirm this line works as expected?

jayRyan24: Thank you, when I copied it over to the repo version, I actually changed the order purposely and didn't realize I had done this. Thanks!

PR #2549, opened 2018-02-28
base-recipes.yaml: Fix up competition shortbow noun.

Sarvatt:


Sarvatt:
I went through all the other bows to be sure there weren't any other messed up nouns and it was just this one.

PR #2550, opened 2018-02-28
Adding Butchery for Necros to combat-trainer.

jayRyan24:
The butchery addition for necromancers to combat-trainer.

1) You can set the `ritual_type` as butcher and it will butcher the entire corpse until done. 
2) You can set the `ritual_type` as cycle, and then override the default dissect to butcher for when locked with FA and Skinning. This could be further expanded to let butcher be the total default, and replace dissect which would only occur `if skinning > 0 and FA >= skinning`. Just need feedback on peoples thoughts about it.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -276,8 +276,8 @@ class LootProcess
     @cycle_rituals = @ritual_type == 'cycle'
     echo("  @cycle_rituals: #{@cycle_rituals}") if $debug_mode_ct
 
-    @butcher_by_default = settings.butcher_by_default
-    echo("  @butcher_by_default: #{@butcher_by_default}") if $debug_mode_ct
+    @butcher_and_dissect = settings.butcher_and_dissect
+    echo("  @butcher_by_default: #{@butcher_and_dissect}") if $debug_mode_ct
 
     @rituals = get_data('spells').rituals
     echo("  @rituals: #{@rituals}") if $debug_mode_ct
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -581,7 +581,7 @@ class LootProcess
   def butcher_corpse(mob_noun, ritual, game_state)
     return unless ritual.eql?('butcher')
     echo "Butchering the #{mob_noun}'s corpse!" if $debug_mode_ct
-    echo " Only butchering it once!" if $debug_mode_ct && butcher_once
+    echo " Only butchering it once!" if $debug_mode_ct && @ritual_type.eql?('cycle')
 
     do_necro_ritual(mob_noun, 'preserve', game_state)
     @equipment_manager.stow_weapon(game_state.weapon_name)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -276,7 +276,7 @@ class LootProcess
     @cycle_rituals = @ritual_type == 'cycle'
     echo("  @cycle_rituals: #{@cycle_rituals}") if $debug_mode_ct
 
-    @dissect_and_butcher = settings.butcher_and_dissect
+    @dissect_and_butcher = settings.dissect_and_butcher
     echo("  @dissect_and_butcher: #{@dissect_and_butcher}") if $debug_mode_ct
 
     @rituals = get_data('spells').rituals
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -584,14 +584,12 @@ class LootProcess
 
     @equipment_manager.stow_weapon(game_state.weapon_name)
 
-    only_once = true if @ritual_type.eql?('cycle')
-
     loop do
       perform_message = "perform #{ritual} on #{mob_noun}"
       result = bput(perform_message, @rituals['butcher'], @rituals['failures'])
       break if result =~ /too battered and beat up/i
       bput("drop my #{right_hand}", 'You drop', 'You discard', 'Please rephrase')
-      break if only_once
+      break if @ritual_type.eql?('cycle')
     end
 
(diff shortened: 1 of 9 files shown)


combat-trainer.lic
```diff
@@ -276,6 +276,9 @@ class LootProcess
     @cycle_rituals = @ritual_type == 'cycle'
     echo("  @cycle_rituals: #{@cycle_rituals}") if $debug_mode_ct
 
+    @butcher_and_dissect = settings.butcher_and_dissect
+    echo("  @butcher_by_default: #{@butcher_and_dissect}") if $debug_mode_ct
```
rpherbig: Should `@butcher_by_default` be `@butcher_and_dissect`?

combat-trainer.lic
```diff
@@ -557,6 +578,26 @@ class LootProcess
     echo "Last ritual performed: #{@last_ritual}" if $debug_mode_ct
   end
 
+  def butcher_corpse(mob_noun, ritual, game_state)
+    return unless ritual.eql?('butcher')
+    echo "Butchering the #{mob_noun}'s corpse!" if $debug_mode_ct
+    echo " Only butchering it once!" if $debug_mode_ct && butcher_once
+
+    bput("perform preserve on #{mob_noun}", @rituals['preserve'], @rituals['failures'])
```
rpherbig: Should this be `do_necro_ritual(mob_noun, 'preserve', game_state)`?

combat-trainer.lic
```diff
@@ -557,6 +578,26 @@ class LootProcess
...
+
+    loop do
+      perform_message = "perform #{ritual} on #{mob_noun}"
+      result = bput(perform_message, @rituals['butcher'], @rituals['failures'])
+      break if result =~ /too battered and beat up/i
+      bput("drop my #{right_hand}", 'You drop', 'You discard', 'Please rephrase')
+      break if @ritual_type.eql?('cycle')
+    end
+
+    bput("perform dissect on #{mob_noun}", @rituals['dissect'], @rituals['failures']) if @ritual_type.eql?('cycle')
```
rpherbig: Same question about `do_necro_ritual`?

combat-trainer.lic
```diff
@@ -557,6 +578,26 @@ class LootProcess
...
+    echo " Only butchering it once!" if $debug_mode_ct && butcher_once
+
+    bput("perform preserve on #{mob_noun}", @rituals['preserve'], @rituals['failures'])
+    @equipment_manager.stow_weapon(game_state.weapon_name)
+
+    loop do
+      perform_message = "perform #{ritual} on #{mob_noun}"
+      result = bput(perform_message, @rituals['butcher'], @rituals['failures'])
+      break if result =~ /too battered and beat up/i
+      bput("drop my #{right_hand}", 'You drop', 'You discard', 'Please rephrase')
```
rpherbig: Is there a common noun we could use (like `material`?) instead of `right_hand`?

jayRyan24: No, and it varies from butchery to butchery. It can be one of any body part, so right_hand works the best. Is this not good practice? This seems like the best way to do this, or anything for that matter.

combat-trainer.lic
jayRyan24: Yes, I'll add that, duh.

combat-trainer.lic
jayRyan24: The items are something like:

> kobold head
> kobold arm
> kobold leg
> kobold torso

jayRyan24:
so, the new logic is as such
1) `ritual_type: butcher`

Always fully butcher the corpse, no dissection

2) `ritual_type: cycle`  &&  `dissect_and_butcher: false`

As normal, only dissect

3) `ritual_type: cycle`  &&  `dissect_and_butcher: true`

Butcher once, then dissect

4) `ritual_type: dissect`  &&  `dissect_and_butcher: false`

Dissect as normal.

5) `ritual_type: dissect`  &&  `dissect_and_butcher: true`

Butcher once, then dissect

combat-trainer.lic
rpherbig: Just make sure you retest it, `do_necro_ritual` does alter some internal state.

combat-trainer.lic
rpherbig: I usually prefer something uniquely 'butchery' to avoid accidentally dropping something important in case we get into a bad state. I'm less concerned when the action is stow, for example.

`right_hand` is fine, though.

rpherbig:
@jayRyan24 This looks good to me. Let me know when you're ready for me to merge it (I don't know if you have any more changes or testing in mind).

combat-trainer.lic
jayRyan24: Right, but it can be too many different items. I can check if right_hand contains one of the body part nouns, and then stow the item if not? What do you think?

rpherbig: I wouldn't worry about it

jayRyan24:
I am currently testing it via @izawwlgood while at work. He is having issues, but I don't think it's based on the code but more his environment. I will let ya know, thanks!

izawwlgood:
My issues so far have been limited to user errors D:

jayRyan24:
Ran this for a full hunt. It harvest, dissected, butchered as expected. I then ran a butcher hunt, and it butchered mercilessly as expected. Handled weapon usage and body parts as expected. I am OK with this going through. @rpherbig

jayRyan24:
@izawwlgood I hope so, I don't know how else it could have happened. I need to get more people using it to verify, but I've been rolling with it pretty well all day, and the changes from this morning seem to be working as intended.

jayRyan24:
@halprind I also think Z was testing and he posted some logs of it working.

PR #2557, opened 2018-03-03
Butcher now checks for multiple failure messages.

jayRyan24:
Needed to check if A we have a result, and B that the result isn't a failure. I didn't realize the other failures were possible. Either way, it should have been put in this way to begin with, I didn't the first time due to it being more cycles.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -593,7 +593,7 @@ class LootProcess
     loop do
       perform_message = "perform #{ritual} on #{mob_noun}"
       result = bput(perform_message, @rituals['butcher'], @rituals['failures'])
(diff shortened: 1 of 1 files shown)


jayRyan24:
One sec, this is backwards.

jayRyan24:
I could also do

```
    loop do
      case bput(perform_message, @rituals['butcher'], @rituals['failures'])
      when @rituals['butcher']
        bput("drop my #{right_hand}", 'You drop', 'You discard', 'Please rephrase')
        break if @dissect_and_butcher && !@ritual_type.eql?('butcher')
        next
      end
      break
    end
```

Which do you think is cleaner? The top one is cleaner to me, but I think **bput** kind of does it already, and I'm reinventing the wheel.

PR #2559, opened 2018-03-03
base-town.yaml: Add ranger guildmaster room in hib

Sarvatt:


Sarvatt:
Included another change by mistake

PR #2561, opened 2018-03-04
horse-trainer.lic: Add reporting how much time each class takes.

Sarvatt:
It seems to vary with skill, good to know to update the wiki at various skill levels.

### DIFF
--- a/horse-trainer.lic
+++ b/horse-trainer.lic
@@ -19,11 +19,14 @@ class TeachHorse
   end
 
   def instruct(skill)
+    @start = Time.now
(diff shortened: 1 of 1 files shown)


horse-trainer.lic
```diff
@@ -19,11 +19,14 @@ class TeachHorse
   end
 
   def instruct(skill)
+    @start = Time.now
```
rpherbig: These don't need to be instance vars if they're all in the same scope.

Sarvatt: thanks for the heads up!

PR #2563, opened 2018-03-04
Create register.lic

BinuDR:
Register searching script.

Usage:  ;register <material>
Output: lines from the register that contain that material

;register silverwood

<NOT INCLUDED HERE, THE OUTPUT OF READING THE REGISTER>

[REGISTER]>stow right
You put your register in your hunting pack.
>
   Results:
    35 -- a deed for some silverwood lumber (6V 97Q - Silverwood --Binu)
--- Lich: REGISTER has exited.

### DIFF
--- a/register.lic
+++ b/register.lic
@@ -1,55 +1,46 @@
 =begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#register?
+
   Search a deed register for the specified crafting material.
   Output all lines that match.
 =end
 
-custom_require.call(%w[common common-items drinfomon])
+custom_require.call(%w[common common-items])
 
 class Register
   include DRC
   include DRCI
 
   def initialize
-    settings = get_settings
-	@search_results = []
-
     arg_definitions =
       [
         [
-          { name: 'query', regex: /^[A-z\s\-\']+$/i, description: 'crafting material to search for, use quotes for multiple words' }
+          { name: 'query', regex: /^[A-z0-9\s\-\'\.]+$/i, description: 'crafting material to search for, use quotes for multiple words' }
         ]
       ]
-
     args = parse_args(arg_definitions)
-
-    if args.query
-      stow_hands
-	  if get_item('deed register', get_settings.crafting_container)
-	    fput("turn my register to contents")
-        search(args.query)
-	  else
-	    echo "Can\'t find your deed register."
-		exit
-	  end
-      stow_hands
-	end
-	respond('   ' + "Results:")
-	if @search_results.any?
-	  display_results(@search_results)
+    container = get_settings.crafting_container
+    
+    stow_hands
+    if get_item('deed register', container)
+	  bput("turn my register to contents", 'You flip your deed register to the contents page.', 'But a deed register is already at the table of contents!')
+      search(args.query)
 	else
-	  respond('   ' + "No matches found.")
+	  echo "Can\'t find your deed register."
+	  exit
 	end
+    stow_hands
   end
 
   def display_results(results)
     results.each do |data|
-      respond("#{data}")
+      respond(data)
     end
   end
 
   def search(query)
     contents = []
-    fput("read my register")
+    bput("read my register", 'Stored Deeds:')
     pause
     loop do
 	  line = get?
@@ -68,7 +59,13 @@ class Register
 	    break
 	  end
 	end
-	@search_results = contents
+	
+	respond('   ' + "Results:")
+	if contents.any?
+	  display_results(contents)
+	else
+	  respond('   ' + "No matches found.")
+	end
   end
   
 end
--- a/register.lic
+++ b/register.lic
@@ -52,7 +52,7 @@ class Register
 	    contents << "   This is not your deed register!"
 		break
 	  end
-	  if line.match(/\W#{query}\W/i)
+	  if line.match(/\W#{query}\W*/i)
 	    contents << line
 	  end
 	  if line.match("Maximum:  ")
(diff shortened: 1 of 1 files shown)


register.lic
```diff
@@ -0,0 +1,76 @@
+=begin
+  Search a deed register for the specified crafting material.
```
rpherbig: Will you be adding documentation for this script to elanthipedia?

register.lic
```diff
@@ -0,0 +1,76 @@
...
+    arg_definitions =
+      [
+        [
+          { name: 'query', regex: /^[A-z\s\-\']+$/i, description: 'crafting material to search for, use quotes for multiple words' }
+        ]
+      ]
+
+    args = parse_args(arg_definitions)
+
+    if args.query
```
rpherbig: Do you need to check `args.query` here? It is the only arg and not-optional.

register.lic
```diff
@@ -0,0 +1,76 @@
...
+=end
+
+custom_require.call(%w[common common-items drinfomon])
+
+class Register
+  include DRC
+  include DRCI
+
+  def initialize
+    settings = get_settings
```
rpherbig: The performance hit is probably negligible, but you could `get_settings` after you parse args - so settings are never retrieved if args are invalid.

register.lic
```diff
@@ -0,0 +1,76 @@
...
+          { name: 'query', regex: /^[A-z\s\-\']+$/i, description: 'crafting material to search for, use quotes for multiple words' }
+        ]
+      ]
+
+    args = parse_args(arg_definitions)
+
+    if args.query
+      stow_hands
+	  if get_item('deed register', get_settings.crafting_container)
+	    fput("turn my register to contents")
```
rpherbig: Would you be up for making this a `bput`?

register.lic
```diff
@@ -0,0 +1,76 @@
...
+	if @search_results.any?
+	  display_results(@search_results)
+	else
+	  respond('   ' + "No matches found.")
+	end
+  end
+
+  def display_results(results)
+    results.each do |data|
+      respond("#{data}")
```
rpherbig: Does string interpolation do anything here?

register.lic
```diff
@@ -0,0 +1,76 @@
...
+    end
+  end
+
+  def search(query)
+    contents = []
+    fput("read my register")
```
rpherbig: Would you be up for making this a `bput`?

register.lic
BinuDR: Yes.

register.lic
BinuDR: I guess not.

register.lic
BinuDR: Ok

register.lic
BinuDR: What's the benefit of bput over fput?

rpherbig: `bput` handles round time more reliably

register.lic
rpherbig: Excellent!

Could you add a link to the header like `Documentation: https://elanthipedia.play.net/Lich_script_repository#register`?

register.lic
```diff
@@ -0,0 +1,76 @@
+=begin
+  Search a deed register for the specified crafting material.
+  Output all lines that match.
+=end
+
+custom_require.call(%w[common common-items drinfomon])
```
rpherbig: What is `drinfomon` used for?

register.lic
```diff
@@ -0,0 +1,76 @@
...
+	end
+  end
+
+  def display_results(results)
+    results.each do |data|
+      respond("#{data}")
+    end
+  end
+
+  def search(query)
```
rpherbig: You don't need to make this change, but I wanted to point out a style thing. You could have `search` return the results (instead of assigning them to `@search_results`) and then you don't need to manage an instance var.

register.lic
BinuDR: Apparently not.  When I was first writing the script, I had issues with it not finding a line if it had the query in the notes you can write in yourself.

IE: searching for 3.68 would not match
    84 -- a deed for a steel ingot (90V 99Q - 3.68 density MCS --Binu)

It's been a few weeks and now I don't remember which change I made fixed that issue.  I'll test without the interpolation and if it's not necessary I'll remove it.

register.lic
BinuDR: Not sure.  I used scroll-search.lic to learn how to do some of the things I wasn't sure of, and it was used in that.  I'll take it out and see if it still works.

register.lic
BinuDR: Same answer as above.

register.lic
BinuDR: I see what you mean.  I can just call search and have it do all the printing that I currently have in the initialize section.  That would be cleaner.  Change made.

register.lic
BinuDR: Changing this one makes the script very laggy when reading the register.

BinuDR: Never mind.  It was removing the pause that made it slow.

Overall though, the bputs do seem to slow down the script just a tad.

BinuDR: See comment above.

register.lic
```diff
@@ -0,0 +1,73 @@
...
+	  line = get?
+	  if line.match("You haven't stored any deeds in this register.")
+	    contents << "   There are no deeds in your register."
+		break
+	  end
+	  if line.match("You shouldn't read somebody else's deed book.")
+	    contents << "   This is not your deed register!"
+		break
+	  end
+	  if line.match(/\W#{query}\W/i)
```
rpherbig: A regex of `\Wlumber\W` _requires_ there to be a non-word character after the word `lumber`. You can use `\W*` if you want _zero or more_ non-word characters.

BinuDR: Thanks, it works properly now.

PR #2570, opened 2018-03-11
base-hunting.yaml: Add newly mapped clay mobs to theren.

Sarvatt:


Sarvatt:
the space removal was because boars had an extra space there, sorry to add an unrelated change

PR #2582, opened 2018-03-15
base.yaml: Add shard_thief_password and commented out stealing towns.

Sarvatt:
just the towns currently in base-stealing.yaml

### DIFF
--- a/data/base-stealing.yaml
+++ b/data/base-stealing.yaml
@@ -5926,11 +5926,4 @@
   pawnable: false
   room: 13721
   town: Horse Clan
-- exceptionally_max: 305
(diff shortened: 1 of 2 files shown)


data/base-stealing.yaml
```diff
@@ -5926,11 +5926,4 @@
   pawnable: false
   room: 13721
   town: Horse Clan
-- exceptionally_max: 305
```
rpherbig: Why was this removed?

Sarvatt: It is in a citizen only shop in shard and shouldn't really go in there, sorry for lumping it together with the other change

PR #2585, opened 2018-03-18
Update steal.lic

aruthas:
Added capability for binning in Shard

### DIFF
--- a/steal.lic
+++ b/steal.lic
@@ -126,7 +126,7 @@ class Steal
     @bin_items.each do |item|
       if @slow_bin_speed
         bput("get #{item} from my #{@stealing_bag}", 'You get')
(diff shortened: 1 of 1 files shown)


steal.lic
```diff
@@ -126,7 +126,7 @@ class Steal
     @bin_items.each do |item|
       if @slow_bin_speed
         bput("get #{item} from my #{@stealing_bag}", 'You get')
-        bput("put #{item} in bin", 'nods toward you as your .* falls into the azurite bin')
+        bput("put #{item} in bin", 'nods toward you as your .* falls into the azurite bin','nods toward you as your .* falls into the sturdy bin.')
```
Sarvatt: How about this instead, it may end up working in other towns also and is simpler.
```
bput("put #{item} in bin", 'nods toward you as your .* falls into the .* bin')

aruthas:
Ok made the changes, that should work much better in the future.

PR #2586, opened 2018-03-18
base-stealing.yaml: Update thieving data, add new leth arthe and shar…

Sarvatt:
…d items.

### DIFF
--- a/data/base-stealing.yaml
+++ b/data/base-stealing.yaml
@@ -684,7 +684,7 @@
   room: 19376
   pawnable: false
   id: '45'
-  acceptable_min: 217
+  acceptable_min: 224
   acceptable_max: 409
   exceptionally_max: 183
   exceptionally_min: 150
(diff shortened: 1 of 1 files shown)


data/base-stealing.yaml
```diff
@@ -5870,49 +5952,55 @@
...
   item_in: on rack
   pawnable: false
   room: 9596
   town: Shard
+  poorly_max: 609
+  poorly_min: 573
 - acceptable_max: 305
   acceptable_min: 305
-  id: '589'
+  id: 589
```
rpherbig: Why did this change from string to int? Will that cause problems?

Sarvatt: The only thing I can think of is that I used ruby 2.3 to run update-stealing. I'll just close this.

rpherbig: 4cb256ae59f485023a79042f4c4eceda405abae5

PR #2590, opened 2018-03-20
Add message for shard collection guy at bank.

yuoree:
Kept hanging up in shard - figured out guy taking money at bank has different message.


rpherbig:
Thank you for the contribution @yuoree !

PR #2595, opened 2018-03-23
Add currency exchange conversion method

jandersson:
Converts one form of currency to another and subtracts the exchange
fee. Default exchange fee is 5% as that is the fee in most of the
commonly used banks. The bank keeps any fractional amount. Assumes the
amount is in copper.


jandersson:
I wanted to update pay-debt.lic to pay debts in other provinces, but saw that we need a way to convert and exchange currency first :) so baby-steps.

PR #2600, opened 2018-03-25
dusk-labyrinth.lic: Update for the new quest that switched to monkeys.

Sarvatt:


rpherbig:
You might consider naming the Flag something more generic like 'target-found' or 'no-animal' so in the future we only need to change the regex/string.

rpherbig:
It's not a big deal, but there's also still one reference to rats: `rat_dir` on line 18.

jandersson:
The entire capturing group with `rat_dir` can be removed - it was a leftover from a previous iteration on that script where it tried to follow the direction the rat went in.

jandersson:
The pet noun could also be generalized in the regex by just matching on `\w+` instead of `monkey` or `rat`

PR #2602, opened 2018-03-29
 Add option to pay debt in another province

jandersson:
* Add optional args for specifying province.
* Deposit all hometown currency in hometown bank, withdraw
  needed amount and do 'exchange all' at the currency changer
* Renamed several variables for clarity.
* Updated convert_currency to specify a 'direction', e.g. How much money
  do I need to withdraw to receive X amt from exchange, or How much money
  will I get when I exchange X. This is done by making the fee positive or negative

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -28,25 +28,33 @@ module DRCM
     amount
   end
 
-  def convert_currency(amount, from, to, fee = 0.05)
+  def convert_currency(amount, from, to, fee)
+    # When determining how much coin is needed to receive X amount in another currency
+    # Use a negative fee percentage
+    # When determining how much coin will be received after the exchange
+    # Use a positive fee percentage
     exchange_rates = {
-      'dokoras' => {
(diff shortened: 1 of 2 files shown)


common-money.lic
```diff
@@ -28,25 +28,33 @@ module DRCM
     amount
   end
 
-  def convert_currency(amount, from, to, fee = 0.05)
+  def convert_currency(amount, from, to, fee)
```
rpherbig: Do you plan to put the actual convert mechanics (not just calculations) in `common-money`?

jandersson: That could be useful

jandersson: I guess the mechanics used in pay-debt are a test to see if thats a good/working approach: deposit all the money on hand then withdraw only what you need to make the exchange.

jandersson:
I want to run two tests with the `sack` arg in and out of the hometown province. Then Ill stamp it

PR #2614, opened 2018-04-07
sanowret-crystal.lic: Don't use while invisible.

Sarvatt:


Sarvatt:
it checks if the invisible indicator is lit up in the xml, yes that works for DR also. I use that indicator in profanity. 

alias :invisible? :checkinvisible
...
def checkinvisible
	XMLData.indicator['IconINVISIBLE'] == 'y'
end

This was tested by someone who hit the problem also

PR #2618, opened 2018-04-13
Unify crafting waggle sets to remove duplication

Sarvatt:
I don't see any reason to have to have separate waggle sets for each craft, if you want to buff one you'd want to buff them all?

### DIFF
--- a/sew.lic
+++ b/sew.lic
@@ -52,7 +52,7 @@ class Sew
     @mat_type = args.material
     @stamp = @settings.mark_crafted_goods
 
-    wait_for_script_to_complete('buff', ['sew'])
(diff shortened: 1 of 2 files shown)


jandersson:
Its for a more granular approach to buffing. Everyone likes to train in their own way. I would add in a general 'crafting' buff set in addition to the existing ones.

Sarvatt:
make it call buff twice then?

jandersson:
Yea

jandersson:
With the exception of PHK, crafting buff spells target a single crafting skill

Sarvatt:
oh that makes more sense, this should just be closed then

Issue #2353, opened 2018-01-10
Cambrinth invoking should invoke only the amount charged or desired.


mithgrek:
Currently, cambrinth use by DRS assumes that a cambrinth item starts at zero and has only the mana it is currently charging.  The default 'invoke' will invoke all mana, and if there is any residual from a failed prior cast or the Trader metaspell, too much mana will be put in and spell will backfire.  

Recommendation is to invoke only the amount charged if the charged amount is less than the total amount of the device.  (or, just always invoke the charged amount).  Feedback on drscripts channel suggested that lower level characters may have issues with this.  I tested with my lowest skilled magic alt (a warrior mage with about 150 arcana), and he could precisely invoke a 32 mana armband that he did not have the spell to wear and charge.

For the occasional minor failure of attuning to +/- a mana, I don't think this is an issue.  If the cambrinth is full, then issue is moot and we'll charge one less.  If the cambrinth has more mana than we expected, and we invoke a lesser amount and gain an extra 1, then cast will probably succeed and if it doesn't, we're still better off than invoking the full item anyway.

Implementation could be by calculating "cambrinth_mana" and inserting into this line of common-arcana
If lower level characters are an issue, it could be conditional.  Lower level chars are less likely to have leftover mana in a cambrinth anyway.  Traders are a temporary exception due to spell preview, but this is temporary.

While Trader preview is exacerbating the issue with the metaspell, this is an existing issue.  Mana starved chars notice it more.  

Pseudocode (with an arcana skill check)
  cambrinth_mana = arcana-skill > 100 ? what_we_wanted : nil

From:
    DRC.bput("invoke my #{cambrinth} #{dedicated_camb_use}", get_data('spells').invoke_messages)
to:
  DRC.bput("invoke my #{cambrinth} #{cambrinth_mana} #{dedicated_camb_use}", get_data('spells').invoke_messages)

PR #2621, opened 2018-04-17
Common-Arcana: Invoke exact amount from cambrinth items

jandersson:
The setting lets a magic user invoke the exact amount
of mana from their cambrinth items. This is useful for
Traders in particular where their cambrinth items
are charged passively and can lead to backfiring.

Only common-arcana and the base setting is included in this update to give users time to get drca updated.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -151,11 +151,11 @@ module DRCA
     !Flags['spell-fail']
   end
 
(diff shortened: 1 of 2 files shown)


jandersson:
Ill get a second PR rolling soon which includes updates to the scripts using these functions.

jandersson:
Resolves issue #2353

PR #2625, opened 2018-04-18
Base update

jandersson:
Adds some missing defaults


jandersson:
I did not add these to base-empty because some scripts are checking if they are truthy/falsey instead of if they are empty

rpherbig:
We might want to change those scripts in the future. We can update `base-empty` then.

PR #2626, opened 2018-04-18
Fix up error in my yaml.

Sarvatt:


jandersson:
Thanks - Its one of the few errors that popped up here: https://travis-ci.org/jandersson/dr-scripts . Just trying to make all those tests pass, eventually.

PR #2633, opened 2018-04-22
Fix paladin glyphs

jayRyan24:
**common-arcana, base-spells**
I had originally pushed through this change to support future and changed glyphs, but this looks like a moot thing based on the current feedback and what's changing. So, I removed it.

**combat-trainer**
Changed the functionality to use a boolean for the glyph and removed all the old support for the previous implementation.

### DIFF
--- a/profiles/Valkiss-setup.yaml
+++ b/profiles/Valkiss-setup.yaml
@@ -175,15 +175,10 @@ priority_defense: Shield Usage
 skip_last_kill: true
 
 ## Paladin ##
+paladin_use_mana_glyph: true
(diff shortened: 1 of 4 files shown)


profiles/Valkiss-setup.yaml
```diff
@@ -175,16 +175,11 @@ priority_defense: Shield Usage
 skip_last_kill: true
 
 ## Paladin ##
+paladin_use_mana_glyph: true
```
rpherbig: Can you also update `base.yaml`?

PR #2644, opened 2018-04-30
Added smite to training_abilities.

jayRyan24:


Sarvatt:
new training_abilities need to be added to validate, https://github.com/rpherbig/dr-scripts/blob/master/validate.lic#L344

rpherbig:
@jayRyan24 Since the exp timer is well known, you might considering updating `training_abilities` in `base.yaml` with it. You might also want to update the listening logic. Neither of those changes hold up this PR, though.

rpherbig:
Thanks for getting this in so quickly!

PR #2648, opened 2018-05-01
Update thieving data.

Sarvatt:


data/base-stealing.yaml
```diff
@@ -5454,9 +5466,11 @@
   acceptable_min: 284
   poorly_max: 291
   poorly_min: 291
+  vpoorly_max: 387
+  vpoorly_min: 387
 - id: '522'
   item: hip pouch
-  item_in: ' catalog'
+  item_in: " catalog"
```
rpherbig: This doesn't look right. Do you know anyone in Shard that could double check it?

data/base-stealing.yaml
```diff
@@ -5767,7 +5791,7 @@
   trivial_max: 1750
   trivial_min: 573
 - id: '556'
-  item: ' mesh handguards'
+  item: " mesh handguards"
```
rpherbig: Similar comment to above

rpherbig:
Other than the two comments, this looks fine to merge.

Sarvatt:
I figured the change from single to double quotes were harmless and just someone trying to manually steal with different quotes to update the info. Reverted that change.

PR #2669, opened 2018-05-14
add shard_embrasure

yuoree:
for lower level option till shard_wall


jandersson:
Thanks for the contribution!

PR #2674, opened 2018-05-16
add wyvern cliff climbing + tweak

yuoree:
Add climbing section to climb both the Undergondola stuff and the Cliffs in wyverns/ice archons, few extra mindstates.

Add the ability to directly call undergondola or wyvern via argument, expandable for future.


Sarvatt:
Are these do-able at the same 540 climbing ranks? If not (which I think not) how about making it automatically do the wyvern ones instead at a higher rank? Also the args should be lower case if they go in like the rest of the scripts.

yuoree:
I don't know what the low end for archon cliffs is, and it's a pretty dangerous area w/ archons and all. So didn't want it to be something automatically forced on people, just an option they can call. I know for sure it moves athletics at 900 and 1600. Lowercase, can do.

PR #2682, opened 2018-05-23
Use a list of pc empaths when healing

jandersson:
Let the user specify multiple empaths incase one goes offline.
Also sorts the list of empaths by nearest first.
If the user already has safe_room_empath and safe_room_id defined,
then these are merged into the pc_empath list.

Example setting
```yaml
---
safe_room_empaths:
- name: Dolores
  id: 1234
- name: Teddy
  id: 5678
- name: Bernard
  id: 7373
```

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -896,6 +896,7 @@ class SetupFiles
     s.safe_room ||= UserVars.safe_room
     s.safe_room_id ||= UserVars.safe_room_id
     s.safe_room_empath ||= UserVars.safe_room_empath
+    s.safe_room_empaths ||= UserVars.safe_room_empaths
     s.slack_username ||= UserVars.slack_username
     s.bankbot_name ||= UserVars.bankbot_name
     s.bankbot_room_id ||= UserVars.bankbot_room_id
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -37,6 +37,7 @@ class SafeRoom
       sort_destinations(pc_empaths.map { |empath| empath['id'] })
         .map { |room| pc_empaths.find { |empath| empath['id'] == room } }
         .each do |empath|
+        next unless DRRoom.pcs.include?(empath['name'].capitalize!) || in_game?(empath['name'])
         next unless use_pc_empath?(empath['id'], empath['name'])
         ensure_copper_on_hand(settings.safe_room_tip_threshold || 0, settings)
         tip(settings.safe_room_tip_threshold, settings.safe_room_tip_amount, empath['name'])
@@ -67,6 +68,11 @@ class SafeRoom
     end
 
     give_and_take(settings.safe_room_id, settings.safe_room_give, settings.safe_room_take)
+  end
+
+  def in_game?(character)
+    # TODO: Remove and use common-validation::in_game? when all players have had a reasonable amount of time to reload common-scripts
+    "  #{character}." == bput("find #{character}", 'There are no adventurers in the realms that match the names specified', "^  #{character}.$")
   end
 
   def need_healing?
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -26,9 +26,22 @@ class SafeRoom
 
     return unless args.force || need_healing?
 
-    if use_pc_empath?(settings.safe_room_id, settings.safe_room_empath)
-      ensure_copper_on_hand(settings.safe_room_tip_threshold || 0, settings)
-      tip(settings.safe_room_tip_threshold, settings.safe_room_tip_amount, settings.safe_room_empath)
+    pc_empaths = settings.safe_room_empaths || []
+    if settings.safe_room_id && settings.safe_room_empath
+      # Add to pc_empaths list if its not in there already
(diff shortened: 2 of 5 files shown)


safe-room.lic
```diff
@@ -26,9 +26,22 @@ class SafeRoom
 
     return unless args.force || need_healing?
 
-    if use_pc_empath?(settings.safe_room_id, settings.safe_room_empath)
-      ensure_copper_on_hand(settings.safe_room_tip_threshold || 0, settings)
-      tip(settings.safe_room_tip_threshold, settings.safe_room_tip_amount, settings.safe_room_empath)
+    pc_empaths = settings.safe_room_empaths || []
```
rpherbig: What about for people that want to use UserVars to store their empath list for the sake of privacy?

safe-room.lic
```diff
@@ -26,9 +26,22 @@ class SafeRoom
...
+      # Add to pc_empaths list if its not in there already
+      pc_empaths << { 'name' => settings.safe_room_empath, 'id' => settings.safe_room_id } unless pc_empaths.any? { |empath| empath['name'] == settings.safe_room_empath && empath['id'] == settings.safe_room_id }
+    end
+
+    if !pc_empaths.empty?
+      # Sort pc empaths list by nearest first
+      sort_destinations(pc_empaths.map { |empath| empath['id'] })
+        .map { |room| pc_empaths.find { |empath| empath['id'] == room } }
+        .each do |empath|
+        next unless use_pc_empath?(empath['id'], empath['name'])
```
rpherbig: Could this be extracted into a `reject` clause before the `each`?

jandersson: I dont think so - its re ordering an array of hashes based on the array of room numbers returned from sort destinations, but not discarding anything.

jandersson: I could look into replacing `find` with `select` or `reject` to cover the edge case where two empaths share the same room number. Let me know if I misunderstood your question.

jayRyan24:
KILLER!

safe-room.lic
rpherbig: Rereading the code, I think I'm wrong.

I hadn't thought about the case where two empaths share the same room. Since we're sorting destinations based on distance, I assume the two empaths would always be adjacent in the ordering, so we shouldn't need to handle that in a special way.

safe-room.lic
rpherbig: Is this line now unnecessary?

jandersson: Yes - thank you base-empty.

jandersson: Guessing you mean the `|| []`

jandersson: We still need to initialize pc_empaths

safe-room.lic
```diff
@@ -26,9 +26,24 @@ class SafeRoom
...
+      pc_empaths << { 'name' => settings.safe_room_empath, 'id' => settings.safe_room_id } unless pc_empaths.any? { |empath| empath['name'] == settings.safe_room_empath && empath['id'] == settings.safe_room_id }
+    end
+
+    if !pc_empaths.empty?
+      # Sort pc empaths list by nearest first
+      sort_destinations(pc_empaths.map { |empath| empath['id'] })
+        .map { |room| pc_empaths.select { |empath| empath['id'] == room } }
+        .flatten
+        .each do |empath|
+        next unless DRRoom.pcs.include?(empath['name'].capitalize!) || in_game?(empath['name'])
```
rpherbig: What do you think about `.select { |empath| in_game?(empath['name']) }` outside of the `.each`? The code would be cleaner, but technically that would check before you actually try to run to the empath. Do you think that's an issue?

safe-room.lic
```diff
@@ -26,9 +26,24 @@ class SafeRoom
...
+        next unless use_pc_empath?(empath['id'], empath['name'])
+        ensure_copper_on_hand(settings.safe_room_tip_threshold || 0, settings)
```
rpherbig: The `|| 0` looks strange, maybe that should be part of `ensure_copper_on_hand`?

safe-room.lic
rpherbig: Oh, that could result in a lot of spam to the game to check the status for every empath even if you only need one. Nevermind.

jandersson: It would make it more readable but I like the behaviour in-game where it just checks lazily when necessary

safe-room.lic
jandersson: What about changing the function prototype to have `safe_room_tip_threshold=0`?

dependency.lic
```diff
@@ -896,6 +896,7 @@ class SetupFiles
     s.safe_room ||= UserVars.safe_room
     s.safe_room_id ||= UserVars.safe_room_id
     s.safe_room_empath ||= UserVars.safe_room_empath
+    s.safe_room_empaths ||= UserVars.safe_room_empaths
```
rpherbig: @jandersson FYI: 73871cca1e757f5b8f93be2af80f001b00518c8a

PR #2683, opened 2018-05-23
Added walk_to(@training_room) after tithing.

Dartellum:


crossing-training.lic
```diff
@@ -256,6 +256,7 @@ class CrossingTraining
     UserVars.tithe_timer = Time.now
     walk_to(almsbox)
     bput("put 5 silver #{currency} in almsbox", 'You drop', 'But you do not', 'attend to thy own woes')
+    walk_to(@training_room)
```
jayRyan24: I don't agree with this. This seems like an extra step and doesn't work with a good flow. Wouldn't it be more efficient for each script/activity/training item to handle where it goes for its training? As an example, this is sending the person the crossing to check the tithing, then it goes back to the training room, then it goes back to the crossing to do pray to chadatru. This type of back and forth could be applied to the rest of the scripts as well. The scripts end where they end, the newly starting script/training should handle where it goes for it's own training.

Dartellum: My guy will sit at the almsbox for a very long time. He is F2P. In other words, I do not want anyone hanging out in Crossing if they do not have to.

jayRyan24: Why is he sitting at the almsbox?

Dartellum: Because he went to tithe and nothing else triggers until skills drain. He will move from their when a skill drains enough to go train.

jayRyan24: I think he is just waiting for the next thing to do? I think introducing inefficiency due to the lack of something for a F2P to do seems iffy.

jayRyan24: Yeah, I get what you're saying, but it seems like he needs something else to do. Again, it just seems iffy to purposely add inefficiency to compensate for inefficiency.

Dartellum: :shrug: many have F2P clerics. But I am open to suggestions.

jayRyan24: Yeah, I did too, but I once they sat around I gave them other things to do.

jayRyan24: Oh yeah, `requires_movement` or something like that. I'll be fair, this change would in no way impact me at all as I don't use this system, but just thought I'd put my figurative two coppers in.

jayRyan24:
Should this be closed?

PR #2684, opened 2018-05-23
Add weapon priority settings to combat trainer

sungobund:
This is an attempt to implement #2240 - the ability to specify weapon priorities. There was lots of great feedback there and I tried to incorporate it all here. 

The use of the array `min_by` to figure out the weapon skill is very clever. I built on that by adding a simple flag for priority weapons and inserting it into the middle of the `min_by` calculation.

The full weapon skill selection logic should now work as follows:
- Always pick the weapon with the lowest field experience, priority weapon or not [**unchanged**]
- If a priority weapon is specified, use that [**added**]
- If a priority weapon is not specified, pick the weapon skill with the lowest rank [**unchanged**]

I think this gets us what we want. If no priority weapon is selected, everything is unchanged. If one is selected, we will still always train the lowest exp weapon, and when they are all equal, we will train a priority weapon. The one downside is we will never train the lowest ranked weapon if a priority weapon is set, but if we set a priority weapon I assume that is the behavior we want.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -16,6 +16,8 @@ class SetupProcess
     echo("  @stance_override: #{@stance_override}") if $debug_mode_ct
     @priority_defense = settings.priority_defense
     echo("  @priority_defense: #{@priority_defense}") if $debug_mode_ct
+    @priority_weapons = settings.priority_weapons
+    echo("  @priority_weapons: #{@priority_weapons}") if $debug_mode_ct
     @cycle_armors = settings.cycle_armors
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @cycle_armors_time = settings.cycle_armors_time
@@ -137,7 +139,7 @@ class SetupProcess
       echo('skipping summoned weapons because no moonblade available') if $debug_mode_ct
       weapon_training = weapon_training.reject { |skill, _| game_state.summoned_info(skill) }
     end
-    new_weapon_skill = weapon_training.min_by { |skill, _| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }.first
+    new_weapon_skill = weapon_training.min_by { |skill, _| [DRSkill.getxp(skill), @priority_weapons.include?(skill) ? -1 : 0, DRSkill.getrank(skill)] }.first
     game_state.update_weapon_info(new_weapon_skill)
 
     game_state.update_target_weapon_skill
--- a/profiles/SampleBarbarian-setup.yaml
+++ b/profiles/SampleBarbarian-setup.yaml
@@ -103,6 +103,10 @@ weapon_training:
   Polearms: spear
   Large Edged: bastard sword
   Twohanded Edged: bastard sword
+priority_weapons:
+  - Large Edged
+  - Polearms
+  - Heavy Thrown
 use_stealth_attacks: true
 skinning:
   skin: true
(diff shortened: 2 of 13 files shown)


rcuhljr:
Not related to this PR at all, but thank you for contributing and welcome aboard.

jayRyan24:
Thanks a lot for contributing!

sungobund:
No problem. This is a great way to learn the codebase. Hoping to start biting off larger things soon. I also noticed there is a slackbot - is there a slack workspace where you all hang out?

Thank you guys for the great project!

jayRyan24:
Who are you in game?

jayRyan24:
How is this different, in function, from #2240?

combat-trainer.lic
```diff
@@ -137,7 +139,7 @@ class SetupProcess
       echo('skipping summoned weapons because no moonblade available') if $debug_mode_ct
       weapon_training = weapon_training.reject { |skill, _| game_state.summoned_info(skill) }
     end
-    new_weapon_skill = weapon_training.min_by { |skill, _| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }.first
+    new_weapon_skill = weapon_training.min_by { |skill, _| [DRSkill.getxp(skill), @priority_weapons.include?(skill) ? -1 : 0, DRSkill.getrank(skill)] }.first
```
rpherbig: If you do not override the `priority_weapons` setting in your character YAML, then I think `@priority_weapons` will be `undefined`. Does this line throw in that case (e.g. it tries to check `includes?` on `undefined`)?

rpherbig: We use `base-empty` to prevent these kinds of problems from happening. If I'm right, you'll want to update that.

combat-trainer.lic
```diff
@@ -16,6 +16,8 @@ class SetupProcess
     echo("  @stance_override: #{@stance_override}") if $debug_mode_ct
     @priority_defense = settings.priority_defense
     echo("  @priority_defense: #{@priority_defense}") if $debug_mode_ct
+    @priority_weapons = settings.priority_weapons
```
rpherbig: Do you have a character YAML that uses this setting you would be willing to upload?

When someone adds a new setting, I like to have an example of it being used (for others to use as a reference). It's not necessary for merging, though.

sungobund: Sure, happy to upload my character setup yaml

combat-trainer.lic
sungobund: Yeah you are definitely right. Thanks for catching that, I will update `base-empty.yaml` with an empty array. 

Out of curiosity, I did add to `base.yaml` just the tag `priority_weapons:`. I see now that `base-empty.yaml` is being used to provide an empty default value. Is the purpose of `base.yaml` to provide just a set of default keys (and sometimes sensible defaults)?

combat-trainer.lic
jayRyan24: Make sure to comment out anything you don't want shared publicly. Safe rooms, etc.

sungobund: I decided to add it to the sample files. If I add my character setup, will it then be auto sync'ed through `dependency` in case I have local changes I would like to keep?

combat-trainer.lic
rpherbig: Pretty much. `base-empty` also makes it easier for users to 'null out' a setting without having to know what the default value should be.

PR #2685, opened 2018-05-24
Add validation for combat_spell_training

sungobund:
This got me pretty good. For my `combat_spell_training` I was following the pattern set with `buff_spells` in naming the key as a spell name, so for a few days none of my magics were training in combat and I couldn't figure out why. Finally did, and decided to add in a validation to clue in others.

One question - do these abbreviations also need to appear anywhere else like they do for `training_spells` used in the `crossing-training` script? I saw a validation for that, but I haven't yet dug into the specifics of how combat-training works with spells yet.

### DIFF
--- a/validate.lic
+++ b/validate.lic
@@ -88,6 +88,19 @@ class DRYamlValidator
     bad_names.each { |skill_name| error("Invalid weapon_training: skill name '#{skill_name}'") }
     echo("Valid skills are #{@valid_weapon_skills}")
   end
+  
+  def assert_that_combat_spell_training_spells_are_skills(settings)
+    return unless settings.combat_spell_training
+    
+    valid_combat_spell_training_skills = %w[Augmentation Warding Utility]
+
+    bad_names = settings.combat_spell_training.keys
+                        .reject { |skill_name| valid_combat_spell_training_skills.include?(skill_name) }
+    return if bad_names.empty?
+
+    bad_names.each { |skill_name| error("Invalid combat_spell_training: skill name '#{skill_name}'") }
+    echo("Valid skill names are #{valid_combat_spell_training_skills}")
+  end
 
   def assert_that_cycle_armors_are_skills(settings)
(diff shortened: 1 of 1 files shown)


validate.lic
```diff
@@ -88,6 +88,19 @@ class DRYamlValidator
...
     echo("Valid skills are #{@valid_weapon_skills}")
   end
+  
+  def assert_that_combat_spell_training_spells_are_skills(settings)
+    return unless settings.combat_spell_training
+    
+    valid_combat_spell_training_skills = %w[Augmentation Warding Utility]
+
+    bad_names = settings.combat_spell_training.keys
+                        .reject { |skill_name| valid_combat_spell_training_skills.include?(skill_name) }
```
jayRyan24: I think this is a little more clear, what do you think? (It follows the convention of the rest of the software too, normally layered method calls are on a separate line like this.)
```
bad_names = settings.combat_spell_training
                                   .key
                                   .reject { |skill_name| valid_combat_spell_training_skills.include?(skill_name) }
```

jandersson:
> One question - do these abbreviations also need to appear anywhere else like they do for training_spells used in the crossing-training script? I saw a validation for that, but I haven't yet dug into the specifics of how combat-training works with spells yet.

Off the top of my head, It should not be needed for offensive_spells or buff_spells or waggle_sets. Your spell data gets merged with base-spells based on either the abbreviation or the spell name. Any values you specify are not overwritten of course, it just 'fills in the blanks'.

validate.lic
sungobund: I am happy to follow the convention in the rest of code base, no problem. I like the in lining of the each too, I will make that change.

sungobund:
Updated for the one method call per line syntax suggestion and inlining of each block. Also updated the weapon_training to match the same stylistically since it was close in lexical scope.

rpherbig:
Thank you for your contribution @sungobund !

PR #2687, opened 2018-05-24
Rezz script rewrite

BinuDR:
I rewrote the script to fix a number of issues, the biggest being mana usage due to it using discern values.

I tried to account for all possible issues that might arise while rezzing someone.  One I know I didn't account for is if someone drags the corpse off.

### DIFF
--- a/rezz.lic
+++ b/rezz.lic
@@ -25,7 +25,8 @@ class Rezz
     args = parse_args(arg_definitions)
     @settings = get_settings
     
-    args.player ? @target = DRRoom.pcs.find { |name| /^#{args.player}/i =~ name }.to_a : find_dead
+    args.player ? @target = DRRoom.pcs.find { |name| /^#{args.player}/i =~ name }.to_a : @target = find_dead
+    @target = nil if @target.empty?
     unless @target
       echo '*** No bodies! ***'
       exit
@@ -50,7 +51,7 @@ class Rezz
     Flags.add('dead', /H.. soul (is|is not) bound to h.. body./)
     Flags.add('decaying', 'will decay in about a minute', 'decay in a few minutes', 'decay in several minutes')
     Flags.add('necro', 'has no protection from the gods. judgment')
-    
+	
     @target.each do |name|
       @rejuv['cast'] = "cast #{name}"
       rejuv_process(name)
@@ -64,8 +65,7 @@ class Rezz
   
   def find_dead
     /You take a moment to look for everybody in the area and see (.*)/ =~ bput('look people', "You look around and notice that you're the only one in the area", /You take a moment to look for everybody in the area and see (.*)/)
-    @target = Regexp.last_match(1).sub(/ and /, ', ').scan(/the body of (.+?(?=who))/i).flatten.map { |obj| obj.split.last }
-    @target = nil if @target.empty?
+    Regexp.last_match(1).sub(/ and /, ', ').scan(/the body of (.+?(?=who))/i).flatten.map { |obj| obj.split.last }
   end
   
   def rejuv(player, quick = false)
@@ -78,11 +78,10 @@ class Rezz
   
   def vigil(player)
     @vigil = { 'abbrev' => 'vigil', 'prep_time' => 10, 'cast' => "cast #{player}", 'mana' => 20 } unless @vigil
-    cast_spell(@vigil, @settings)
-    vigil(player) unless DRSpells.active_spells['Vigil']
+    vigil(player) unless cast_spell(@vigil, @settings)
   end
   
-  def find_soul(player)
+  def find_soul?(player)
     Flags.add('rotting', /#{player}.s body'?s? (grows paler|seems to|appears to|appearance takes)/)
     10.times do
       pause 5 while DRStats.mana < 40
@@ -95,7 +94,7 @@ class Rezz
     return false
   end
   
-  def keep_going
+  def keep_going?
     respond("*** Not found after #{@infuse_attempts} infuses so far. Do you want to continue? ***")
     respond("*** To continue: ;send yes ***")
     respond("*** To give up: ;send no ***")
@@ -117,18 +116,21 @@ class Rezz
     Flags.reset('dead')
     Flags.reset('decaying')
     Flags.reset('necro')
-    bput("perc #{player}", /#{player} has (.*) favors* with (his|her) god/, /has no favor with h.. god!/)
+    bput("perc #{player}", /#{player} has (.*) favors* with h.. god/, /has no favor with h.. god!/)
     vigil(player) if Flags['decaying']
     (!Flags['rejuv-silver'] && Flags['dead']) ? rejuv(player) : nil
   end	
   
   def rezz_process(player)
     cast_spell(@rezz, @settings) unless DRSpells.active_spells['Resurrection']
-    bput('cast', get_data('spells').cast_messages) until DRSpells.active_spells['Resurrection']
-    waitfor('You prepare to look for the lost spirits of the dead')
+    while checkprep == 'Resurrection'
+      pause 1
+      bput('cast', get_data('spells').cast_messages)
+    end
+    pause 3
     
-    until find_soul(player)
-      unless keep_going
+    until find_soul?(player)
+      unless keep_going?
         bput("whisper #{player} I can't find your soul.  You need more favors!", 'You whisper', 'Who are you trying to whisper to?')
         return
       end
--- a/rezz.lic
+++ b/rezz.lic
@@ -1,118 +1,131 @@
-#   Documentation: https://elanthipedia.play.net/Lich_script_repository#rezz
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#rezz
+  
+  Define a waggle sets for rezz and rejuv if you don't want to use the default values in base-spells.yaml
+  
+  EXAMPLE
+  waggle_sets:
+    rezz:
+      Resurrection:
+        mana: 17
+    rejuv:
+      Rejuvenation:
+        mana: 10
+        cambrinth:
+        - 15
+
+  In order to infuse directly from your mana pool add the following to your yaml:
+  osrel_no_harness: true
+  
+=end
 
 custom_require.call(%w[common common-arcana events drinfomon spellmonitor])
-
-# rezz waggle set with POM if you want
 
 class Rezz
   include DRC
   include DRCA
-
+  
   def initialize
+    DRStats.cleric? ? nil : begin echo '*** Nice try, heathen! ***'; exit end
+	
     arg_definitions = [
       [
         { name: 'person', regex: /\w+/, description: 'Person to rezz' }
       ]
     ]
-
-    @rejuv_mana = 0
+	
     args = parse_args(arg_definitions)
     @settings = get_settings
-    rezz_prep(args.person)
+    
+    @person = args.person.capitalize
+	/the body of (#{@person}.*) who/ =~ bput('look people', /the body of (#{@person}.*) who/)
+	@person = Regexp.last_match(1)
+    
+    @rezz = if @settings.waggle_sets['rezz']
+               @settings.waggle_sets['rezz'].values.first
+           else
+	         get_data('spells').spell_data['Resurrection']
+	       end
+    	   
+    @rejuv = if @settings.waggle_sets['rejuv']
+               @settings.waggle_sets['rejuv'].values.first
+           else
+	         get_data('spells').spell_data['Rejuvenation']
+	       end
+
+    @rejuv['cast'] = "cast #{@person}"
+	
+	check_favor
+    rejuv
+    cast_rezz
+    find_soul
+    soul_bond
+    rejuv(true)
+    gesture
   end
-
-  def rezz_prep(person)
-    # cast buffs if set
-    check_buffs
-
-    # perceive body & rejuv
-    rejuv(person)
-
-    # cast rezz
-    mana = cast_rezz
-
-    # infuse untill see spirit
-    find_soul(person, mana)
-
-    # snap cast soul bonding
-    cast_soulbonding(person)
-
-    # check for silver again. recast rejuv (or infuse again) if not
-    rejuv(person)
-
-    # gesture
-    bput("gest #{person}", 'Roundtime')
+  
+  def check_favor
+    Flags.add('no-favor', /(#{@person}.*) has no favor with (.*) god/)
+    bput("perc #{@person}", /#{@person} has (.*) favors with (his|her) god/)
+    Flags['no-favor'] ? begin echo "*** #{@person} has no favors! ***"; exit end : nil
   end
-
-  def find_soul(person, infuse_amt)
-    loop do
-      pause 5 while mana <= 40
-      harness_mana([infuse_amt]) unless skip_harness?
-      case bput("infuse rezz #{infuse_amt}", /You clearly sense the spirit of #{person} in the Void/i, 'You sense a spirit nearby, but you are unable to make it out clearly.', 'Your sense of sight has not fully transitioned to the spirit world', "You don't have enough harnessed mana to infuse that much.")
-      when 'You sense a spirit nearby, but you are unable to make it out clearly.'
-        next
-      when /You clearly sense the spirit of #{person} in the Void/i
-        break
-      when 'Your sense of sight has not fully transitioned to the spirit world'
-        pause 2
-        next
-      when "You don't have enough harnessed mana to infuse that much."
-        next
-      else
-        next
-      end
+  
+  def rejuv(quick = false)
+    quick ? begin @rejuv['cambrinth'] = []; @rejuv['prep_time'] = 5; @rejuv['mana'] = 5 end : nil
+    Flags.add('wrong-name', "You can't cast that at yourself")
+    Flags.add('silver', 'A thin silver nimbus surrounds')
+    Flags.add('alive', 'Nothing happens')
+    until Flags['silver'] || Flags['alive'] || Flags['wrong-name']
+	  pause 1 while DRStats.mana < 50
+      cast_spell(@rejuv, @settings)
+  	  pause 0.5
     end
+    Flags['alive'] ? begin echo "*** #{@person} is already alive, dumbass! ***"; exit end : nil
+    Flags['wrong-name'] ? begin echo "*** Name spelled incorrectly. ***"; exit end : nil
+    Flags.reset('rejuv-silver')
   end
-
-  def is_silver?(person)
-    /As you concentrate, a thin (.*) nimbus/ =~ bput("perc #{person}", 'As you concentrate, a thin (.*) nimbus')
-    color = Regexp.last_match(1)
-    echo "Memories are #{color}"
-    waitrt?
-    color.eql? 'silver'
-  end
-
-  def rejuv(person)
-    return if is_silver?(person)
-    data = get_data('spells').spell_data['Rejuvenation']
-    data['cast'] = "cast #{person}"
-    if @rejuv_mana.zero?
-      data = check_discern(data, @settings)
-      @rejuv_mana = data['mana']
-    end
-
-    loop do
-      cast_spell(data, @settings)
-      break if is_silver?(person)
-    end
-  end
-
-  def check_buffs
-    wait_for_script_to_complete('buff', ['rezz'])
-  end
-
-  def skip_harness?
-    attunement = DRSkill.getrank('Attunement')
-    return false if attunement < 541
-    return true if attunement > 600
-    return true if check_if_buff_active('Persistence of Mana', 6)
-    false
-  end
-
+    
   def cast_rezz
-    data = get_data('spells').spell_data['Resurrection']
-    data = check_discern(data, @settings)
-    cast_spell(data, @settings)
-    data['mana']
+    Flags.add('rezz-ready', 'You prepare to look for the lost spirits of the dead')
+    Flags.add('rezz-not-prepped', 'This pattern may only be cast with full preparation')
+    cast_spell(@rezz, @settings)
+    if Flags['rezz-not-prepped']
+      while Flags['rezz-not-prepped'] 
+	    Flags.reset('rezz-not-prepped')
+        pause 1
+	    bput('cast', get_data('spells').cast_messages, )
+	  end
+    end
+    while !Flags['rezz-ready']
+      pause 0.5
+    end
   end
-
-  def cast_soulbonding(person)
-    data = { 'abbrev' => 'SB', 'prep_time' => 5, 'cast' => "cast #{person}", 'mana' => 1 }
+  
+  def find_soul
+    Flags.add('soul-found', /sense the spirit of #{@person} in the Void/i)
+    while !Flags['soul-found']
+      pause 1 while DRStats.mana < 50
+      @settings.osrel_no_harness ? nil : harness_mana([@rezz['mana']])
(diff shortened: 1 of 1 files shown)


rezz.lic
```diff
@@ -1,118 +1,131 @@
...
+    
+  def cast_rezz
+    Flags.add('rezz-ready', 'You prepare to look for the lost spirits of the dead')
+    Flags.add('rezz-not-prepped', 'This pattern may only be cast with full preparation')
+    cast_spell(@rezz, @settings)
+    if Flags['rezz-not-prepped']
+      while Flags['rezz-not-prepped'] 
+	    Flags.reset('rezz-not-prepped')
+        pause 1
+	    bput('cast', get_data('spells').cast_messages, )
```
BinuDR: There's an extra comma at the end of that statement, before the ).  I will take that out.

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
     ]
-
-    @rejuv_mana = 0
+	
     args = parse_args(arg_definitions)
     @settings = get_settings
-    rezz_prep(args.person)
+    
+    @person = args.person.capitalize
+	/the body of (#{@person}.*) who/ =~ bput('look people', /the body of (#{@person}.*) who/)
```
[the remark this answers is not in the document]
BinuDR: It's still too greedy, if there is another person lying down after the target, it will pick up way too much.
New solution:
/the body of (#{@person}[a-z]*) who is lying down/

jandersson:
> I rewrote the script to fix a number of issues, the biggest being mana usage due to it using discern values.

Thanks for the contribution - I want to suggest that you break big rewrites like this into smaller pull requests so that smaller sets of logical changes can get reviewed, tested, merged. Just a suggestion though, again thanks for contributing.

BinuDR:
@jandersson:
I rewrote the whole script because the old one had major issues, and it was a challenge for myself so that I could learn more ruby and more about the various lich scripts, methods, how to pull info from various places (drskills, drstats, etc).

rezz.lic
```diff
@@ -1,118 +1,131 @@
-#   Documentation: https://elanthipedia.play.net/Lich_script_repository#rezz
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#rezz
+  
+  Define a waggle sets for rezz and rejuv if you don't want to use the default values in base-spells.yaml
```
jandersson: Can you move this to the elanthipedia documentation?

jayRyan24:
Best way is to dive right in. I honestly have no clue how you would break this into smaller parts, it's a whole rewrite to me, but I'm a maniac.

rezz.lic
BinuDR: There is no current elanthipedia documentation.  I'll create it there and remove it from here once we finalize how it's going to be setup here.

BinuDR:
@jayRyan24 (Not sure why there isn't a reply under this comment of yours)
"Oh, I also think you should check for attempts as you can OOM on people that don't have it. I would even use it as an argument, or base it on the "perc" feedback. But I think 5-10 is good, if I can't find them in that range, I'm not going to find them."

OOM as in running out of mana?  Yeah I thought I handled that with a mana check somewhere?  If not, I'll add it in:
pause 1 while DRStats.mana < 50
Something like that.

rezz.lic
BinuDR: I made all of the different methods because that's how the original script was setup for the most part.

I'll look over the suggested changes.  Thanks for the feedback!

jayRyan24:
Yeah, what I actually meant is that you can never raise them. I've been in a situation where I just can't raise. Another issue is where it takes too long and things keep failing to happen in the correct time and you'll eventually never get out of the loop. I don't know, I think giving it a max number of tries is like "ok, I can't do this" instead of relying on the player to stop it.

rezz.lic
jayRyan24: I think the program adage is, "refactor, reduce", or I could be pulling that out of my ass.

rezz.lic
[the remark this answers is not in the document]
jandersson: What if this is an invasion triage and i want to raise a specific person in the room?

I agree with @jayRyan24, its already an arg to the script. What you could do is add an additional argument, or have the lack of a `person` argument tell the script to just raise anyone whose dead in the room.

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
+  
+  EXAMPLE
+  waggle_sets:
+    rezz:
```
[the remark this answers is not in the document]
BinuDR: Yeah, I was thinking of that but I was having issues with learning how to access hashes properly.  I'll change it.

rezz.lic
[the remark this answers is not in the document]
jayRyan24: That's another option yeah, just `;rezz` to call it.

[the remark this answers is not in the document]
jandersson: Disregard, I misunderstood :)

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
-    rezz_prep(args.person)
+    
+    @person = args.person.capitalize
+	/the body of (#{@person}.*) who/ =~ bput('look people', /the body of (#{@person}.*) who/)
+	@person = Regexp.last_match(1)
+    
+    @rezz = if @settings.waggle_sets['rezz']
+               @settings.waggle_sets['rezz'].values.first
+           else
+	         get_data('spells').spell_data['Resurrection']
```
jandersson: Do we want to print something to the user here? Such as "Using minimum prep values. Please make a waggle set called 'rezz' to configure your rezz spell settings."

rezz.lic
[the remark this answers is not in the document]
jayRyan24: Oh, this is in case the body gets drug away, lol. Ok, I get it. I was like, wtf is the purpose of this and then I put his original comment with this, 2+2=4. 🥇 for my ability to understand.

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
+	       end
+    	   
+    @rejuv = if @settings.waggle_sets['rejuv']
+               @settings.waggle_sets['rejuv'].values.first
+           else
+	         get_data('spells').spell_data['Rejuvenation']
```
jandersson: See comment for 'rezz'

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
-    # snap cast soul bonding
-    cast_soulbonding(person)
-
-    # check for silver again. recast rejuv (or infuse again) if not
-    rejuv(person)
-
-    # gesture
-    bput("gest #{person}", 'Roundtime')
+  
+  def check_favor
```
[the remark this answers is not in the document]
BinuDR: When I first added that, it was so I could have the script echo some message about if the person had low favors, but I didn't do that in the end.  I just have it so it exits the script if the person has no favors, rather than preparing the body fully and then finding out when trying to gesture.

Now that I think about it though, it should rejuv and then exit if they have no favors.

rezz.lic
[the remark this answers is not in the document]
BinuDR: Just me being picky about having the person's full name for the fun of it.  I didn't know how to do string scan, and I'm sure I could have pulled the Also here line out of some lich generated data source somewhere.  I'm just not sure what variable stores that also here: line.

I didn't set it up to raise everyone, but I could do that.  I'd prefer to leave it as an optional argument.

Either ;rezz <target>
Or ;rezz all (or something like that.

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
+      @settings.osrel_no_harness ? nil : harness_mana([@rezz['mana']])
+      bput("infuse rezz #{@rezz['mana']}", /sense the spirit of #{@person} in the Void/i, 'You sense a spirit nearby, but you are unable to make it out clearly.')
+    end
   end
-
-  def cast_soulbonding(person)
-    data = { 'abbrev' => 'SB', 'prep_time' => 5, 'cast' => "cast #{person}", 'mana' => 1 }
+  
+  def soul_bond
+    data = { 'abbrev' => 'SB', 'prep_time' => 8, 'cast' => "cast #{@person}", 'mana' => 5 }
```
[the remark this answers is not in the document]
jayRyan24: I'm almost certain that they'll have the required skill to min prep insta cast it. I would insta cast it, min prep is all you ever need, and if someone ever comes across it not working, figure out how/why and then adjust slightly to accommodate.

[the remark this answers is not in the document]
jayRyan24: 1 mana, 10 min skill required, I couldn't see a world where this wouldn't be possible at 30, and to be honest, most don't do this quest until much later. Just my thoughts.

rezz.lic
jandersson: I think yelling at the user to define their spell settings should be good enough. Doing it at min prep is likely to be a horrible experience and should motivate them :D

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
+	    Flags.reset('rezz-not-prepped')
+        pause 1
+	    bput('cast', get_data('spells').cast_messages, )
+	  end
     end
-
-    loop do
-      cast_spell(data, @settings)
-      break if is_silver?(person)
+    while !Flags['rezz-ready']
```
jandersson: You can use `until` instead of `while not`. Rubocop may auto correct this

rezz.lic
```diff
@@ -1,118 +1,131 @@
...
-  def cast_rezz
-    data = get_data('spells').spell_data['Resurrection']
-    data = check_discern(data, @settings)
-    cast_spell(data, @settings)
-    data['mana']
+  
+  def find_soul
+    Flags.add('soul-found', /sense the spirit of #{@person} in the Void/i)
+    while !Flags['soul-found']
+      pause 1 while DRStats.mana < 50
```
jandersson: Previously this value was 40, and it still was not enough for my 50th cleric. I think 40 makes sense, but can argue for an even lower threshold based on my previous experience.

jandersson: Id argue 10 is fine. You want to keep harness/infusing until you find the soul or its no longer feasible to maintain rezz. Waiting to restore mana while you have a cyclic going thats draining it faster than you restore it will just make the user kill the script.

BinuDR:
@jayRyan24 
"Yeah, what I actually meant is that you can never raise them. I've been in a situation where I just can't raise. Another issue is where it takes too long and things keep failing to happen in the correct time and you'll eventually never get out of the loop. I don't know, I think giving it a max number of tries is like "ok, I can't do this" instead of relying on the player to stop it."

I have raised people where it took me 300-500+ infused mana to raise them.

I do agree that maybe we should have something to deal with too many infuses.
Keep track of it, if it gets above a certain amount echo a message to screen.  Maybe a prompt that they need to answer?
"X# infuses so far, do you want to continue?"

What's a good way to handle waiting for a response?

rezz.lic
BinuDR: I don't have rubocop installed, working on it.

rezz.lic
BinuDR: Whatever everyone thinks is best.

rezz.lic
jandersson: Don't worry about it, I was hesitant to point out style things because rubocop is going to correct it all in the end. I thought you might be interested though :)

jandersson: I notice there are a TON of review comments in here, so ill refrain to avoid fatiguing you :P But do check out the Ruby Style Guide

jayRyan24:
Something like:

```
respond("#{@infuse_total} infuses so far, do you want to continue?")
respond(";cont to continue, or ;exit to exit")

line = get until line.strip =~ /^(cont|exit)$/i

case line
when /^cont$/i
  continue
when /^exit$/i
  respond("Exiting!")
  fput('release rezz')
  fput('release mana')
  fput("Get some favors you dead beat!")
  exit
end
```

BinuDR:
From testing I did a long time ago, 50 mana is the minimum needed to find a soul if they have enough favors.  When I use my friend's cleric, I prep rezz at 17 and infuse 17 mana 3 times to find people.
Maybe prepping at higher mana would allow you to find souls with less mana?  I've never tested.

The cleric I have access to is able to infuse 17 mana a LOT of times before running low, but he's 100th with 700 utility and 800 attunement.

BinuDR:
Maybe we can use the # of favor info to adjust how much to infuse at a time?

If it's really low, infuse a bit less at a time so you don't run out of mana as quickly.

BinuDR:
Darn it!  I found a more efficient way to deal with the name that doesn't match it if they're not dead, so now I don't get to have this line in the script anymore.

    Flags['alive'] ? begin echo "*** #{@target} is already alive, dumbass! ***"; exit end : nil

It's so sad.

BinuDR:
Ok, new changes:
;rezz by itself will raise all corpses in the room
;rezz will only raise that person.

Put both Rejuv and Rezz in 1 waggle set, and now it gives a message if you don't have it setup.

Now the script only does the 'look people' if you run it without any arguments so it can check for dead people.

Made the methods more efficient and removed some of the unnecessary ones.

They're now called with rejuv_process and rezz_process.

If soul isn't found in 10 infuses, it gives the option to give up or keep going. If you keep going, it will do 10 more and ask again as many times as you let it continue.

I added the script runtime displaying at the end because I was curious. That can stay or go, whatever is appropriate.

The only thing I can think of that isn't handled is if the body is either dragged off or someone else raises them first.
If they get raised while you're still rejuving, it will deal with it. But not after that.

I could add some more messages to the 'alive' flag. "You can't cast that at yourself." will show up both if you try to rejuv or soul bond someone who isn't there.

Oh, just added one more flag for if the person is a necro. I ran across a dead one while testing and noticed that they have a different message when you perceive them.

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
+			 end
+
+    Flags.add('no-favor', /has no favor with h.. god!/)
+    Flags.add('rejuv-silver', 'A thin silver nimbus surrounds', 'As you concentrate, a thin silver nimbus flickers into view around')
+    Flags.add('alive', 'Nothing happens')
+    Flags.add('necro', 'has no protection from the gods. judgment')
+
+    @target.each do |name|
+      @rejuv['cast'] = "cast #{name}"
+	  rejuv_process(name)
```
jayRyan24: An alignment issue here.

jayRyan24: I think this matches the @target.each do, which is not aligning correctly.

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
+			   echo "Using default values for REJUV, make a 'rezz'=>Rejuvenation waggle set to avoid this."
+	           get_data('spells').spell_data['Rejuvenation']
+			 end
+
+    Flags.add('no-favor', /has no favor with h.. god!/)
+    Flags.add('rejuv-silver', 'A thin silver nimbus surrounds', 'As you concentrate, a thin silver nimbus flickers into view around')
+    Flags.add('alive', 'Nothing happens')
+    Flags.add('necro', 'has no protection from the gods. judgment')
+
+    @target.each do |name|
```
jayRyan24: Is this ever going to be an array? Wouldn't it be simpler, because this isn't an array of multiple characters to raise, for this to just be put here without the `each do`? It's just confusing, but I could also just not be reading it correctly.

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
-
-    # check for silver again. recast rejuv (or infuse again) if not
-    rejuv(person)
-
-    # gesture
-    bput("gest #{person}", 'Roundtime')
+  def find_dead
+    /You take a moment to look for everybody in the area and see (.*)/ =~ bput('look people', "You look around and notice that you're the only one in the area", /You take a moment to look for everybody in the area and see (.*)/)
+    @target = Regexp.last_match(1).sub(/ and /, ', ').scan(/the body of ([a-z[:blank:]]+) who is [a-z[:blank:]]+?,?./i).flatten.map { |obj| obj.split.last }
+	@target == [] ? @target = nil : nil
```
jayRyan24: So this appears that it's setting @target to nil regardless of true or false, so why the check?

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
-        pause 2
-        next
-      when "You don't have enough harnessed mana to infuse that much."
-        next
-      else
-        next
-      end
+  def rejuv(player, quick = false)
+    Flags.reset('rejuv-silver')
+    quick ? begin @rejuv['cambrinth'] = []; @rejuv['prep_time'] = 5; @rejuv['mana'] = 10 end : nil
```
jayRyan24: I'm going to be honest. These if checks for begins are slick, but they are also tough to read, maybe an occasional use would be OK, but this many is becoming overused.

jayRyan24: Could be a personal preference thing too, not entirely sure. :D

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
+    until Flags['rejuv-silver'] || Flags['alive'] || Flags['necro']
+	  pause 5 while DRStats.mana < 40
```
jayRyan24: Line 64 doesn't align with line 65 properly. I'd double check that the alignments and formatting of everything matches up across the entire file.

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
-      data = check_discern(data, @settings)
-      @rejuv_mana = data['mana']
-    end
-
-    loop do
-      cast_spell(data, @settings)
-      break if is_silver?(person)
+  def find_soul(player)
+    10.times do
+	  pause 5 while DRStats.mana < 40
```
jayRyan24: Alignment issues with this method too.

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
-    data['mana']
-  end
-
-  def cast_soulbonding(person)
-    data = { 'abbrev' => 'SB', 'prep_time' => 5, 'cast' => "cast #{person}", 'mana' => 1 }
-    cast_spell(data, @settings)
+  def keep_going
+    respond("*** Not found after #{@infuse_attempts} infuses so far. Do you want to continue? ***")
+    respond("*** To continue: ;send yes ***")
+	respond("*** To give up: ;send no ***")
```
jayRyan24: This repond's alignment.

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
+  def rejuv_process(player)
+    Flags.reset('alive')
+    Flags.reset('no-favor')
+    bput("perc #{player}", /#{player} has (.*) favors* with (his|her) god/, /has no favor with h.. god!/)
+    Flags['rejuv-silver'] ? nil : rejuv(player)
+  end	
+
+  def rezz_process(player)
+    cast_spell(@rezz, @settings)
+	bput('cast', get_data('spells').cast_messages) until DRSpells.active_spells['Resurrection']
```
jayRyan24: Alignments in this method also.

rezz.lic
```diff
@@ -1,119 +1,126 @@
...
-      @rejuv_mana = data['mana']
-    end
-
-    loop do
-      cast_spell(data, @settings)
-      break if is_silver?(person)
+  def find_soul(player)
+    10.times do
+	  pause 5 while DRStats.mana < 40
+      @settings.osrel_no_harness ? nil : harness_mana([@rezz['mana']])
```
jayRyan24: As a side note here, I usually cast rezz at 20 mana, and harness at 15. Usually I've went with around 75% mana on harness to rezz ratio.

jayRyan24:
```
[rezz]>cast Solmiw
You gesture at Solmiw.
Solmiw's soul is already bound to his body.
[sanowret-crystal]>exhale my sanowret crystal
You exhale softly on your sanowret crystal, and scintillating
You hear in your mind a quiet recollection of wisdom.  Though
  were to GAZE into the crystal, you come away from the experi
  scintillating lights fade again.
>
Mekare's feet are dripping water everywhere.
>'found an issue
>
You say, "Found an issue."
>'easy fix
You say, "Easy fix."
>
You hear the ghostly voice of Solmiw ask, "What happened?"
```

rezz.lic
BinuDR: Yes.  If you just run ;rezz, it will look at the people in the room and make an array of all the dead.

rezz.lic
BinuDR: If @target is empty, it sets it to nil.  Otherwise it does nothing.

Otherwise it would send an empty array of names back and that would get past the check at line 24 or 25.  Then it would send the empty array to the @target.each do |name| section.

args.player ? @target = DRRoom.pcs.find { |name| /^#{args.player}/i =~ name }.to_a : find_dead
!@target ? begin echo "*** No bodies! ***"; exit end : nil

jayRyan24: Oh yes, I'm reading it. So if true do something else do nothing. This is how I would handle that, and it would follow convention with the rest of Dependency.

`@target = nil if @target.empty?`

rezz.lic
BinuDR: I was just trying to keep the code condensed.  I have no issues writing them out in normal blocks.

I like those ternary if statements, but they don't work if you need to do more than 1 command in either section.  I found some example with the begin/end and used it.  But the style guide says that it's a bad idea for code readability.

rezz.lic
jayRyan24: Gotcha, thanks.

rezz.lic
jayRyan24: Yeah, they're slick but could become cumbersome to read. I think this is all personal preference though. Keep doing it, I would only care ultimately if the code owners cared.

rezz.lic
BinuDR: "jayRyan24 3 days ago  Contributor
I also hope he gets froggy and makes it raise all dead in the room. :D But that's another pR. :P"

I got froggy.  ;rezz with no arguments will make an array of all the dead in the room.

rezz.lic
BinuDR: It sets it to nil only if true.  If false it does nothing.
What happened otherwise is that if I ran the script with no bodies in the room, it would pass an empty array to @targets and then it would get by the check on like 24 that would echo that there are no bodies and exit the script.

rezz.lic
BinuDR: Yeah.  I like those ternary ifs, but after reading a little of the ruby style guide it says that doing what I did is bad.  I'll change it.

rezz.lic
BinuDR: Why not just prep at 15 too then?  Prepping at 20 in that case just makes it use more mana each time it pulses.

rezz.lic
BinuDR: Oh, that looks better.  I'll change it to that.

BinuDR:
@jayRyan24 Does it hang the script if you attempt to SB a corpse that's already bound?

If not, I'm not going to worry about tracking it just to save a few mana/seconds.

rezz.lic
BinuDR: Ok, I changed it to that.

rezz.lic
BinuDR: @jayRyan24 See previous comment.

BinuDR:
I took care of all of the indentation issues, I hope.

I also fixed a few of the style issues as well as made the ternary if statements with begin/end into multi line if statements so that they're more readable.

I changed one of the flags so that the script doesn't have to attempt to rejuv before realizing that the player is already alive.
That change might require some code to handle if someone rezzes them while you're trying to find their soul. Not certain.

Let me know if there are any other things I can improve!

rpherbig:
Looks good to me.

@BinuDR @jayRyan24 @jandersson any more planned changes or should I merge this?

Sarvatt:
It works unlike the current script and is a huge improvement, would like to see it merged here. I've tested it twice now in the past day. Anything more can be a new feature.

BinuDR:
@rpherbig
Go ahead and merge it.
I'm only updating now if I find better ways to handle stuff, find bugs, or if there are feature requests.

PR #2693, opened 2018-05-26
pick.lic: Drop too-hard boxes unless picking_box_storage is defined.

Sarvatt:


Sarvatt:
This is kind of implicit behavior that isn't obvious, if you have picking_box_storage: defined it will keep them otherwise it wont. Would it be better to have an extra option like picking_discard_too_hard: true maybe for people that cut and paste others setups?

PR #2696, opened 2018-05-27
Train sorcery in combat spell training

jandersson:
Sheath and wield items before and after casting via
a flag in gamestate, `casting_sorcery`

Example settings
---
```yaml
combat_spell_training:
  Sorcery:
    name: Ghost Shroud
    mana: 21
    cyclic: true
    day: true # Keeps from conflicting with SLS, another cyclic
```


Sarvatt:
Awesome! How about adding it to validate too on this line

https://github.com/rpherbig/dr-scripts/blob/master/validate.lic#L93

jandersson:
Sounds good

Issue #2554, opened 2018-03-02
bescort needed for the m'riss/lang balloon


Sarvatt:
The map is hardcoded to use escort.lic to do it and that has a typo in it, trying to join baloon.

PR #2697, opened 2018-05-27
bescort.lic: Add support for the mriss/lang balloon.

Sarvatt:
Yes baloon is misspelled on purpose internally, keeping in tradition with escort.lic. :)

Closes: #2554

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -139,6 +139,10 @@ class Bescort
         { name: 'mode', options: %w[island ratha], description: 'Where do you need to get to?' }
       ],
       [
+        { name: 'baloon', regex: /baloon/i, description: 'The air ship between Langenfirth and Mriss' },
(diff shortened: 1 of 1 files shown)


Sarvatt:
i'm open to making it balloon too. just find this funny since escort.lic never worked with it and keeping history :)

rpherbig:
I'd much prefer 'balloon'.

Sarvatt:
fixed :)

bescort.lic
```diff
@@ -658,6 +664,36 @@ class Bescort
...
+  def take_balloon(mode)
+    case mode
+    when 'mriss'
+      manual_go2(8793)
+      if wealth('Therenborough') < 10_000
+        echo('Get money you slob!')
+        return
+      end
+    when 'langenfirth'
+      if wealth('Mer'Kresh') < 10_000
```
rpherbig: This needs to either be double-quotes or escape the `'`... :sad_trombone:

Issue #2706, opened 2018-06-01
safe-room.lic is not moving on to an npc empath if a pc empath is defined but not available.


Sarvatt:
after 07495065fcec2, if you have safe_room_empath and safe_room_id defined but they are not in the game , safe-room will just exit instead of falling back to a npc empath

PR #2707, opened 2018-06-01
 Use npc empath if no pc empaths are available

jandersson:
Fixes the bug in #2706 

I'll write some tests for safe-room to try and prevent this in the future.

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -32,18 +32,7 @@ class SafeRoom
       pc_empaths << { 'name' => settings.safe_room_empath, 'id' => settings.safe_room_id } unless pc_empaths.any? { |empath| empath['name'] == settings.safe_room_empath && empath['id'] == settings.safe_room_id }
     end
 
-    if !pc_empaths.empty?
-      # Sort pc empaths list by nearest first
-      sort_destinations(pc_empaths.map { |empath| empath['id'] })
-        .map { |room| pc_empaths.select { |empath| empath['id'] == room } }
-        .flatten
-        .each do |empath|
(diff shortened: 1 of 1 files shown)


safe-room.lic
```diff
@@ -193,6 +182,22 @@ class SafeRoom
     bput('lie', 'You lie') if standing?
   end
 
+  def use_pc_empaths?(pc_empaths, settings)
+    return if pc_empaths.empty?
```
rpherbig: ❓  Should this be `return false`?

safe-room.lic
```diff
@@ -32,18 +32,7 @@ class SafeRoom
...
-        .map { |room| pc_empaths.select { |empath| empath['id'] == room } }
-        .flatten
-        .each do |empath|
-        next unless DRRoom.pcs.include?(empath['name'].capitalize!) || in_game?(empath['name'])
-        next unless use_pc_empath?(empath['id'], empath['name'])
-        ensure_copper_on_hand(settings.safe_room_tip_threshold || 0, settings)
-        tip(settings.safe_room_tip_threshold, settings.safe_room_tip_amount, empath['name'])
-        break
-      end
+    if use_pc_empaths?(pc_empaths, settings)
```
rpherbig: 💡 What do you think about adding a comment to note this is deliberately empty?

safe-room.lic
jandersson: Yea. im guessing predicates should be explicit about true/false

safe-room.lic
jandersson: That sounds good to me.

jandersson:
This should be good to go, I did some manual testing

PR #2708, opened 2018-06-02
Pet buffing support added to buff_spells.

jayRyan24:
This will allow you to buff your pet once right after it is up. This doesn't support maintained buffs due to the issue that Vigor's up-time on a Warrior is not perceptible. You don't see the buff fall off.

Min prep RUP lasts a very long time, not a typical buff, around 50 mana and by the time you need Vigor for your warrior, you are big enough to cast it for the full duration of 40 minutes which is longer than most people hunt.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1219,7 +1219,7 @@ class SpellProcess
       if data['pet_type']
         check_spell_timer?(data) && pet_active?(data['pet_type'])
       elsif data['recast_every']
-        check_spell_timer(data)
+        check_spell_timer?(data)
       elsif data['expire']
         true
       else
@@ -1310,7 +1310,7 @@ class SpellProcess
                    .select { |spell| spell['max_threshold'] ? game_state.npcs.length <= spell['max_threshold'] : true }
                    .select { |spell| spell['expire'] ? Flags["ct-#{spell['abbrev']}"] : true }
                    .select { |spell| game_state.dancing? ? spell['harmless'] : true }
-                   .select { |spell| spell['recast_every'] ? check_spell_timer(spell) : true }
+                   .select { |spell| spell['recast_every'] ? check_spell_timer?(spell) : true }
                    .select { |spell| @cast_only_to_train ? DRSkill.getxp(spell['skill']) <= 32 : true }
                    .select { |spell| spell['cast_only_to_train'] ? DRSkill.getxp(spell['skill']) <= 32 : true }
                    .select { |spell| spell['cyclic'] ? !DRSpells.active_spells[spell['name']] : true }
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1159,10 +1159,6 @@ class SpellProcess
     Time.now - (@spell_timers[data['pet_type']] || @spell_timers[data['abbrev']] || Time.at(0)) >= data['recast_every']
   end
 
-  def pet_active?(type)
-    DRRoom.npcs.include?(type)
-  end
-
   def check_buff_conditions?(name, game_state)
     return false if $weapon_buffs.include?(name) && (game_state.aimed_skill? || game_state.brawling?)
     # Resonance does not show up in Active Spells, so we need to track the last weapon on which it was cast to avoid recasting it continuously
@@ -1213,7 +1209,7 @@ class SpellProcess
 
     name, data = recastable_buffs.find do |name, data|
       if data['pet_type']
-        check_spell_timer?(data) && pet_active?(data['pet_type'])
+        check_spell_timer?(data) && DRRoom.npcs.include?(data['pet_type'])
       elsif data['recast_every']
         check_spell_timer?(data)
       elsif data['expire']
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -906,9 +906,6 @@ class SpellProcess
     @symbiosis_learning_threshold = settings.symbiosis_learning_threshold
     echo(" @symbiosis_learning_threshold: #{@symbiosis_learning_threshold}") if $debug_mode_ct
 
-    @pet_buffs = settings.pet_buffs
-    echo(" @pet_buffs: #{@pet_buffs}") if $debug_mode_ct
-
     Flags.add('ct-spelllost', 'Your pattern dissipates with the loss of your target')
     Flags.add('ct-need-bless', ' passes through the .* with no effect')
 
@@ -1152,15 +1149,22 @@ class SpellProcess
 
     Flags.reset("ct-#{@reset_expire}") if @reset_expire
 
-    prepare_spell(@pet_buffs, game_state, true) if @pet_buffs
     return unless @command
     pause 0.5 until DRRoom.npcs.include?('warrior')
     fput("command #{@command}")
     @command = nil
   end
 
-  def check_spell_timer(data)
-    Time.now - (@spell_timers[data['abbrev']] || Time.at(0)) >= data['recast_every']
+  def check_spell_timer?(data)
+    Time.now - (@spell_timers[data['pet_type']] || @spell_timers[data['abbrev']] || Time.at(0)) >= data['recast_every']
+  end
+
+  def pet_active?(type)
+    if DRStats.moon_mage?
+      DRRoom.room_objs.include?(type)
+    else
+      DRRoom.npcs.include?(type)
+    end
   end
 
   def check_buff_conditions?(name, game_state)
@@ -1212,7 +1216,9 @@ class SpellProcess
                        .select { |name, _data| check_buff_conditions?(name, game_state) }
 
     name, data = recastable_buffs.find do |name, data|
-      if data['recast_every']
+      if data['pet_type']
+        check_spell_timer?(data) && pet_active?(data['pet_type'])
+      elsif data['recast_every']
         check_spell_timer(data)
       elsif data['expire']
         true
@@ -1358,7 +1364,13 @@ class SpellProcess
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    @spell_timers[data['abbrev']] = Time.now if data['recast_every']
+    if data['recast_every'] 
+      if data['pet_type']
+        @spell_timers[data['pet_type']] = Time.now
+      else
+        @spell_timers[data['abbrev']] = Time.now 
+      end
+    end
 
     if data['moon']
       check_moonwatch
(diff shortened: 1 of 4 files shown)


rcuhljr:
❓ I haven't looked into combat-trainer in awhile, but it feels like this would just keep recasting the pet buff because nothing is setting `@pet_buffs` false? 

💡 If you ever want to extend to support recasting you could look at using the recast_every: setting and add a flag like `buff_pet: petname` to a regular buff spell entry, So it gets picked like every other buff spell where if a pet with the matching name is out and recast_every timer is up it casts that spell on that pet. That would even allow buffing multiple pets. (Can anyone even get those now with necro risen?) but it would prevent needing another spell section.

jayRyan24:
Oh this one does yeah, I had it implemented with the check after the command, but moved it up due to supporting other guilds. Good catch. Fuck, OK. I can implement it the way that was suggested, didn't consider adding another spell entry, was making it it's own thing. I'll do it this way tomorrow.

rcuhljr:
You'll make several peoples days when this goes live.

jayRyan24:
I actually didn't implement it with buff_spells to begin with before, because my pet was getting rocked in between the cast time, so if it waited for a buff_spell time, it was too injured. That can be look at a couple ways:

First: Maybe if it couldn't live that long, then I shouldn't be hunting there.
Second: This is most beneficial during those transitions, lol.

But yeah, I remember why I did begin to add it there, then made a `petbuff_process`, but then just decided the quickest way to get it casting was prepped right after the pet went up.

rcuhljr:
How tight is the timing? If you made it the top priority checked in buff spells it _should_ be the next buff to cast after the pet. I don't remember if we weave in tm/debils with buffs or if buffs are all priority one.

jayRyan24:
Yeah, I moved it to the top, but some things were going off before it during the AbilityProcess.

jayRyan24:
It buffs first, I moved it to the top and it should be good. 

```
buffs: &buffs
  << : *vigpet
  << : *em
  << : *ic
  << : *ags
  #<< : *vig
  << : *tranq
  << : *heal
  #<< : *pop
```

rcuhljr:
Is spell process higher than ability process? you could have execute for spell process return true? false? whatever it is that tells it to skip all lower priority processes whenever it casts a pet spell and come back to spell process.

rcuhljr:
Although then you'd need to know all the pet spells... Hmmm.

jayRyan24:
So I use the pet_type tag to set the timer, this is due to the fact that for Empaths, the pet buff of choice is Vigor which could also be cast on yourself, so you'd have overlapping timers with the abbreviation "vigor".

rcuhljr:
Ah right spell name is the key to that timer hash. That approach could work but is it limited to one buff per pet then? Would it make sense to split the buffs out in the process setup to pet buffs and people buffs and have two sets of timers?

jayRyan24:
Shouldn't be an issue but wondering what guilds even have multiple buffs? Is that worth it? Again not hard just dunno if it's useful.

jayRyan24:
I don't believe their pet buff which makes them undead killers will be useful to Empaths. If it is release and is, I'll split the code up to make it work with multiple buffs on the pet, but for not I think this is OK. I honestly don't believe we'll ever have multiple pet buffs.

rcuhljr:
I meant instinct for evasion and other buffs we can cast on others. Or are pets not targetable with general cast on other spells and only explicit spells which are flagged for pets?

jayRyan24:
I don't know about the pet flagging spells. I do know Vigor is castable on myself, others, and also my pet. I assume someone would need to have the feats to make whatever spells group castable, of whatever. I don't think it's something that would be ever brought up/needed. If there's ever a call for it, maybe expanding would be fine, but I think this is all that is really needed at this point.

rcuhljr:
Totally not relevant to the current PR, but I think we've crossed wires. Rangers (and others) have plenty of buffs they can defaultly cast on themselves and others (and I assume pets). These aren't group buffs and shouldn't require any feats. The only thing I don't know is if elanthipedia is correct, where spells like vigor https://elanthipedia.play.net/Vigor list creature as allowed target, while instinct https://elanthipedia.play.net/Instinct doesn't list creature so maybe only a very limited set of spells are pet castable in which case we should never bother trying to support multiple spells. Just need to find someone to loan an instinct scroll to an empath at some point.

rcuhljr:
@rpherbig looks good for a merge? probably use a set of eyes other than mine.

jayRyan24:
Ah OK, I thought most buffs required feats.

jayRyan24:
I should add I've been running it all night with the Empath. I don't have a necro or MM to test the others. I know @jandersson was interested in this specifically for necro buffing, but also mentioned MM. So I think he'll jump in and test once it's merged, and can fix any issues (though I can't imagine any) that could arise. It doesn't break anything, that is for sure. 😉

jayRyan24:
@rcuhljr Thanks for all the feedback, it made this a high quality add over a mediocre one.

jayRyan24:
wait a moment, found an issue

jandersson:
I dont believe SLS is technically a pet in the sense that it can be targeted or casted at. Necromancers should enjoy the RPU buff support though.

rcuhljr:
Thanks for taking the extra effort to refactor to the other approach.

jayRyan24:
Ok, I had changed the method `check_spell_timer` which is a predicate method to be `check_spell_timer?`. I didn't change the other 2 in the rest of the code. I had in my main base, but not in the dr-scripts base.

jayRyan24:
> Thanks for taking the extra effort to refactor to the other approach.

I had tried a few other methods and based on the lack of available flags, or the best way to set a timer, I hadn't pushed those. This is the better approach as it keeps all buffs in a central area for CT. This is definitely the better approach. I kind of threw mine up here, not happy with it, but having it "work". I was hoping to have happen what did happen. 

A lot of times in development outside eyes are the best thing because you are so focused on your way of how it should be, you can't see the simplest and best way to make it work which is staring right at you.

jayRyan24:
> I dont believe SLS is technically a pet in the sense that it can be targeted or casted at. Necromancers should enjoy the RPU buff support though. 

Thought you mentioned there is a buff for them? Either way, this will still work for them if it does. If not, it won't technically hurt anything and we can safely remove the extra DRStats check later if not.

jayRyan24:
I also admit to programming in a more "it works" methodology for this than I would professionally. However that stands.

jandersson:
> Thought you mentioned there is a buff for them? Either way, this will still work for them if it does. If not, it won't technically hurt anything and we can safely remove the extra DRStats check later.

There is but it is a spell you cast on yourself.

jayRyan24:
OK, that still is fine then. Does it require an active sphere? If so, this will work great, and can be used at the check. The cast just needs to be `cast: cast`.

jandersson:
> Does it require an active sphere? 
I don't believe so, but I don't have a lot of experience with it. The buff spell isn't particularly popular because it carries high penalties/risks with it.

jayRyan24:
OK. Do you want me to remove that support then?

profiles/Melborne-setup.yaml
```diff
@@ -133,6 +133,14 @@ magic_training:
 
 buff_spells:
   #<< : *qe
+  Reverse Putrefaction:
+    pet_type: zombie
+    recast_every: 2100 # 35 minutes
+    cast: cast zombie
```
jandersson: RPU only works with `cast offense`  and `cast defense`. This might be a bit misleading

jayRyan24: OK, thanks for that heads up!

jayRyan24:
I went ahead and removed that method call. No need for it.

combat-trainer.lic
```diff
@@ -1155,8 +1155,8 @@ class SpellProcess
     @command = nil
   end
 
-  def check_spell_timer(data)
-    Time.now - (@spell_timers[data['abbrev']] || Time.at(0)) >= data['recast_every']
+  def check_spell_timer?(data)
+    Time.now - (@spell_timers[data['pet_type']] || @spell_timers[data['abbrev']] || Time.at(0)) >= data['recast_every']
```
rpherbig: 💡 Extracting `@spell_timers[data['pet_type']] || @spell_timers[data['abbrev']] || Time.at(0)` into a well-named variable would add a lot of readability to this function.

combat-trainer.lic
```diff
@@ -1354,7 +1356,13 @@ class SpellProcess
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    @spell_timers[data['abbrev']] = Time.now if data['recast_every']
+    if data['recast_every']
```
rpherbig: 💡 I think extracting a variable here would add some readability, something like:
```
    if data['recast_every'] 
      timer_type = data['pet_type'] ? 'pet_type' : 'abbrev'
      @spell_timers[data[timer_type]] = Time.now
    end
```

profiles/Ssarek-setup.yaml
```diff
@@ -92,6 +92,15 @@ vig: &vig
     cambrinth:
     - 20
     - 20
+vigpet: &vigpet
```
rpherbig: 💡 Since we're going to be pointing at this as an example of pet buffs being used, what do you think about renaming this to `vigor_pet_buff` (or something similarly more-verbose-but-clearer)?

validate.lic
```diff
@@ -551,6 +551,18 @@ class DRYamlValidator
     error('sell_loot_money_on_hand is invalid. The proper format is: <amount> <denomination> e.g. 3 silv or 4 bronze')
   end
 
+  def assert_that_pet_buffs_have_spell_requirements(settings)
+    return unless settings.buff_spells
+    
+    settings.buff_spells.select { |_name, data| data['pet_type'] }
+                        .reject{ |_name, data| data.include? ('recast_every') }
+                        .each{ |_name, data| error("Pet buffs require a custom recast timer. Please include a \"recast_every:\" setting for #{_name}.") }
```
rpherbig: 💡 By convention, the underscore in front of a variable name means it is not used. Since `_name` is used in the error message, inside the `.each` it should be named `name`. The same for the next block.

rpherbig:
All of my comments are cosmetic, so I'll merge this. A future PR with those things changed would be welcome but not required 😁

validate.lic
jayRyan24: Ah, this is something I couldn't find info for. I figured that the one without was the focus. Good to know this.

PR #2709, opened 2018-06-02
Adding a setting for pouch tying.

jayRyan24:
Adding this for discussion. 

Pros:

1) A lot of new players ask about this consistently. 
2) I don't have a person to take my pouches, so sometimes I have issues where gems get stowed and I have to "start over". I do then have to go out and kill something until a gem, or pick a box, etc. I then need to manually tie up a new pouch. This would just tie always if setting is true. 

Cons:

1) Yet another YAML setting. 
2) Not as slick at the first approach.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -311,6 +311,9 @@ class LootProcess
     @gem_nouns = get_data('items').gem_nouns
     echo("  @gem_nouns: #{@gem_nouns}") if $debug_mode_ct
 
+    @tie_pouch = settings.tie_gem_pouches
+    echo("  @tie_pouch: #{@tie_pouch}") if $debug_mode_ct
+
     @spare_gem_pouch_container = settings.spare_gem_pouch_container
     echo("  @spare_gem_pouch_container: #{@spare_gem_pouch_container}") if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


rcuhljr:
Let's just make it at a setting at this point, at least when someone has it auto tie a pouch they didn't want tied it will be on their head. I did like when it was auto detected, although to clarify on point two if you ever end up with gems outside a pouch you should already have the gems you need to start a tied pouch with, just dump the container into the pouch and tie it.

jayRyan24:
That is not be the scenario which causes me to not have a pouch with a gem or something, I forget exactly which does because that's obvious there, but it does happen occasionally. I think the big issue for me is the first one.

jandersson:
Just to chime in from my own experience:

When using auto-tie and the current pouch is full, it drops the gem in hand and swaps to a new pouch. An issue pops up if the gem on the ground disappears during the pouch swap. When that happens, the new pouch does not get tied because it needs at least one item in it. At this point the player needs to manually tie the current pouch or the rest of the new pouches will be untied. I noticed this happens when hunting mobs that steal stuff from the ground like Adan'f or with hunting buddies.

jandersson:
A setting would be nice because then it would enforce all pouches get tied in the above case

PR #2714, opened 2018-06-04
Added lines to exit from Junk Yard

dr-leustyin:


tulanegator:
This update breaks the ;steal script (especially for F2P thieves) when it causes you to enter the junk yard. The ;steal script automatically drops items in the junk yard, leaves the junk yard, then stops collecting stolen items.

Can we add a check after 60 seconds to determine if you're still in the junk yard and, if so, exit you at that point?

rpherbig:
@dr-leustyin @tulanegator does this need reverted until it can be updated?

dr-leustyin:
Just pushed a fix

PR #2715, opened 2018-06-04
Replaced sell-loot with a run to the bank

dr-leustyin:


dr-leustyin:
The crux of the issue is that if you get caught by a guard while ;go2 is running, pay-debt or sell-loot will hang for quite a bit of time before their first walk_to call before killing it, so you sit in the same spot for a minute and then see this:

```--- Lich: 'go2' has been stopped by pay-debt.
--- Lich: go2 has exited.
--- Lich: go2 active.
```

The update on line 55 fixes this issue by killing the ;go2 process that's blocking ;sell-loot from running (pay-debt or sell-loot will kill this process themselves to get moving, as shown above, so I don't see any added side effects by killing it manually in line 55)

As for the sell-loot process, I think it was chosen so that the player would end up with `sell_loot_money_on_hand` coins instead of 300.  If that's the desire it can stay, although I think it would make sense to move most of that logic into common-money and call those methods rather than spinning up a 'sell-loot' process after getting arrested.

If that's the case I can amend this PR to keep sell-loot in there, but then work on another one to move that logic over to common-money and update the jail-buddy code then

Dartellum:
Make sure whatever you do will work with bankbots.

dr-leustyin:
ensure_copper_on_hand is bankbot aware

PR #2716, opened 2018-06-04
Adding broken out performance script

dr-leustyin:


performance.lic
```diff
@@ -0,0 +1,156 @@
+custom_require.call(%w(common common-travel))
```
rpherbig: ❌ Please add a documentation header with a link to elanthipedia

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+  include DRCT
+  
+  def initialize
+    no_pause_all
+    startTime = Time.new.to_i
+
+    @settings = get_settings
+    @song_list = get_data('perform').perform_options
+    UserVars.song = @song_list.first.first unless UserVars.song
+    fput('release ecry') if DRStats.bard?
```
rpherbig: ❓ Why release ECRY? Is it so the performance buff will not cause you to play a more-difficult-than-normal song?

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+
+    instrument = @settings.instrument
```
rpherbig: ❌ This looks like a new setting. Please update `base-yaml` with a default value for it.

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+      bput("get my #{instrument}","You get")
+      bput("remove my #{hand_armor}","You pull off","Remove what")
+      bput("stow my #{hand_armor}","You put","Stow what")
+    end
+    
+    if play_song
+      loop do 
+        runTime = Time.new.to_i - startTime
+        break if runTime > 600
+        break if DRSkill.getxp('Performance') > 30
```
rpherbig: 💡 The `DRSkill.getxp('Performance') > 30` pattern is repeated three times. What do you think about extracting it to a helper function?

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+          else
+            exit
+          end
+        when /Perhaps you should find somewhere drier/
+          exit
+        end
+      end
+    end
+     
+    if instrument != nil
```
rpherbig: 💡 You can simplify this to `if instrument`

rpherbig: Or even `bput("stow my #{instrument}", "You put") if instrument`

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+      return false
+    when 'You\'re already playing a song'
+       echo "Already playing"
+      fput('stop play')
+      return play_song
+    when 'You cannot play'
+      wait_for_script_to_complete('safe-room')
+    when 'dirtiness may affect your performance'
+      if DRSkill.getrank('Performance') < 20
+        echo "Skipping cleaning of zills due to low rank of Performance: #{DRSkill.getrank('Performance')}" if UserVars.crossing_trainer_debug
```
rpherbig: 💡  Should this still be relying on `UserVars.crossing_trainer_debug`?

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+            exit
+          end
+        when /Perhaps you should find somewhere drier/
+          exit
+        end
+      end
+    end
+     
+    if instrument != nil
+      bput("stow my #{instrument}","You put")
```
rpherbig: 💡 By convention, strings that are 'simple' (no apostrophe, no interpolation) should be single-quoted like `'You put'`. Rubocop will fix this for us.

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+      loop do 
+        runTime = Time.new.to_i - startTime
+        break if runTime > 600
+        break if DRSkill.getxp('Performance') > 30
+        line = get
+        pause 0.05 unless line
+        case line
+        when /You begin (.*)/, /You continue (.*)/, /You find yourself relaxing (.*)/, /You concentrate on your sense of rhythm (.*)/, /You continue (.*)/
+          if DRSkill.getxp('Performance') > 30
+            fput('stop play')
```
rpherbig: ❓ Is this necessary due to the `before_dying`?

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+      bput("stow my #{instrument}","You put")
+    end
+     
+  end
+  
+  def play_song
+    
+    @did_play = true
+    result = bput("play #{UserVars.song}", 'dirtiness may affect your performance', 'slightest hint of difficulty', 'fumble slightly', 'Your zills is submerged in the water','You begin a', 'You struggle to begin', 'You\'re already playing a song', 'You effortlessly begin', 'You begin some', 'You cannot play', 'Play on what instrument', 'now isn\'t the best time to be playing', 'Perhaps you should find somewhere drier before trying to play')
+     echo result
```
rpherbig: ❓ Should this be guarded by a conditional like `if UserVars.something_debug`?

performance.lic
```diff
@@ -0,0 +1,156 @@
...
+     case result
+    when 'Play on what instrument'
+      fput('get zills')
```
rpherbig: ❓ Should these two lines use the `instrument` setting instead of zills?

performance.lic
rpherbig: ❌ Since you're relying on `DRStats`, please add `drinfomon` to the `custom_require` list

dr-leustyin: That's right for releasing ECRY.  Without this you can end up in situations where the script can't accurately play a song, since you might be playing considerably different songs if ECRY is up vs when its not.  Releasing ECRY also extends the range of training with zills by about 150 ranks

dr-leustyin: Is the `require` call necessary or just convention?  The script works without it

rpherbig: ECRY: Cool, thanks for the explanation

performance.lic
dr-leustyin: I've never had an issue with it not pulling the instrument out, but have definitely run into instances where my zills were stowed instead of worn, for some reason.  I think this was put in to guard against that.

performance.lic
dr-leustyin: Looks like some of my debugging messages were left.  I'll remove them

performance.lic
dr-leustyin: Done

performance.lic
```diff
@@ -0,0 +1,153 @@
...
+          end
+        when /Perhaps you should find somewhere drier/
+          exit
+        end
+      end
+    end
+     
+  end
+  
+  def done_training
```
rpherbig: 💡 By convention, a function which returns a boolean should have a `?` at the end of its name, so `done_training?`

performance.lic
```diff
@@ -0,0 +1,153 @@
...
+        break if done_training
+        line = get
+        pause 0.05 unless line
+        case line
+        when /You begin (.*)/, /You continue (.*)/, /You find yourself relaxing (.*)/, /You concentrate on your sense of rhythm (.*)/, /You continue (.*)/
+          if done_training
+            exit
+          end
+        when /You finish (.*)/
+          if not done_training
```
rpherbig: 💡 By convention, we've been using `! && ||` rather than `not and or`.

Another style would be `unless done_training?`.

Probably my most preferred option would be to reverse the order of the `then` and `else` so you don't need a `!` in the first place:
```
if done_training?
  exit
else
  play_song
end
```

rpherbig:
All of my comments are 💡 , so I'll merge this. A future PR with those things changed would be welcome but not required 😁

performance.lic
```diff
@@ -0,0 +1,153 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#performance
+=end
+
+custom_require.call(%w(common drinfomon spellmonitor))
```
rpherbig: @dr-leustyin This list needs to include `events`

PR #2720, opened 2018-06-05
Fixed bug which caused exits during ;steal

dr-leustyin:


afk.lic
```diff
@@ -48,7 +48,14 @@ loop do
   if line =~ /For a current status of how many items you have, use LOOK PORTAL./
     echo "Looks like you're in the junk yard.  We'll log you out in 60 seconds.  You can ';kill afk' to stay in the game, just remember to restart it once you get out of the junk yard"
     pause 60
-    exit_game("Exiting from the Junk Yard")
+
+    fput('look')
+    lines = reget(30)
+    lines.each do | line |
+      if line =~ /Junk Yard/
```
rpherbig: ❓ Have you considered using the room title instead of `reget`? That is how `steal.lic` checks for the Junk Yard: https://github.com/rpherbig/dr-scripts/blob/6b24e56b6da7a4f6e126058d0f1537e9796b4870/steal.lic#L263

dr-leustyin: Didn't know that was an option.  This is more clear

PR #2722, opened 2018-06-06
Updated jail-buddy to grab a couple of mindstates while waiting for t…

dr-leustyin:
…he judge

### DIFF
--- a/jail-buddy.lic
+++ b/jail-buddy.lic
@@ -54,7 +54,7 @@ def get_out_of_jail(old_loc)
   Script.running.find_all { |s| !s.paused? && !s.no_pause_all }.each(&:pause)
   stop_script('go2') if Script.running?('go2')
   echo('** DON\'T PANIC JAIL BUDDY HAS YOUR BACK! **')
-  fput('stand') if not standing?
+  DRC.fix_standing
   start_script('outdoorsmanship',['10 "dust bunny"'])
   
   waitfor('PLEAD INNOCENT or PLEAD GUILTY')
--- a/jail-buddy.lic
+++ b/jail-buddy.lic
@@ -54,7 +54,12 @@ def get_out_of_jail(old_loc)
   Script.running.find_all { |s| !s.paused? && !s.no_pause_all }.each(&:pause)
   stop_script('go2') if Script.running?('go2')
   echo('** DON\'T PANIC JAIL BUDDY HAS YOUR BACK! **')
+  fput('stand') if not standing?
+  start_script('outdoorsmanship',['10 "dust bunny"'])
+  
   waitfor('PLEAD INNOCENT or PLEAD GUILTY')
+  stop_script('outdoorsmanship') if Script.running?('outdoorsmanship')
+  
   fput('PLEAD INNOCENT')
 
   loop do
--- a/outdoorsmanship.lic
+++ b/outdoorsmanship.lic
@@ -9,21 +9,44 @@ class Outdoorsmanship
   include DRCI
 
   def initialize
+    arg_definitions = [
+      [
+        { name: 'mindstates', regex: /\d+/, optional: true, description: 'Number of mindstates or collection attempts to perform before exiting. Defaults to 3' },
+        { name: 'collect_item', regex: /\w+/i, optional: true, description: 'Item to collect. Will use forage_item if not set' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
+
     settings = get_settings
-    @forage_item = settings.forage_item
+    
+    @targetxp = 3
+    @targetxp = args.mindstates.to_i if args.mindstates
+    
+    if args.collect_item
+      @forage_item = args.collect_item
+    else
+      @forage_item = settings.forage_item
+    end
+
+    start_exp = DRSkill.getxp('Outdoorsmanship')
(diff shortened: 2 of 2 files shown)


outdoorsmanship.lic
```diff
@@ -9,21 +9,44 @@ class Outdoorsmanship
...
+        { name: 'mindstates', regex: /\d+/, optional: true, description: 'Number of mindstates or collection attempts to perform before exiting. Defaults to 3' },
+        { name: 'collect_item', regex: /\w+/i, optional: true, description: 'Item to collect. Will use forage_item if not set' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
+
     settings = get_settings
-    @forage_item = settings.forage_item
+    
+    @targetxp = 3
```
rpherbig: 💡 These two lines could be combined, e.g. `@targetxp = args.mindstates ? args.mindstates.to_i : 3`

jail-buddy.lic
```diff
@@ -54,7 +54,12 @@ def get_out_of_jail(old_loc)
   Script.running.find_all { |s| !s.paused? && !s.no_pause_all }.each(&:pause)
   stop_script('go2') if Script.running?('go2')
   echo('** DON\'T PANIC JAIL BUDDY HAS YOUR BACK! **')
+  fput('stand') if not standing?
+  start_script('outdoorsmanship',['10 "dust bunny"'])
```
rpherbig: 💡 I _think_ you can make this a bit clearer with `start_script('outdoorsmanship', [10, "dust bunny"])`

jail-buddy.lic
```diff
@@ -54,7 +54,12 @@ def get_out_of_jail(old_loc)
...
+  
   waitfor('PLEAD INNOCENT or PLEAD GUILTY')
+  stop_script('outdoorsmanship') if Script.running?('outdoorsmanship')
```
rpherbig: ❓  Do we need a `waitrt?` after stopping the script?

jail-buddy.lic
```diff
@@ -54,7 +54,12 @@ def get_out_of_jail(old_loc)
   Script.running.find_all { |s| !s.paused? && !s.no_pause_all }.each(&:pause)
   stop_script('go2') if Script.running?('go2')
   echo('** DON\'T PANIC JAIL BUDDY HAS YOUR BACK! **')
+  fput('stand') if not standing?
```
rpherbig: :x: There's a [helper in common.lic](https://github.com/rpherbig/dr-scripts/blob/master/common.lic#L260) that will do this (note you'd call it with the prefix, e.g. `DRC.fix_standing`).

outdoorsmanship.lic
```diff
@@ -9,21 +9,44 @@ class Outdoorsmanship
...
   def train_outdoorsmanship
-    if DRSkill.getrank('Outdoorsmanship') < 20
-      forage?('rock')
-      dispose_trash 'rock'
-    else
-      item = @forage_item
-      collect(item)
-      waitrt?
-      kick_pile? unless kick_pile?("#{item}s")
+    attempt = 0   # Failsafe counter to prevent an infinite loop if item isn't collectable in the room
```
rpherbig: 💡 I'm not opposed to this change, but I feel like `collect(...)` should maybe be a predicate and handle this.

jail-buddy.lic
dr-leustyin: That's what I tried first, but start_script didn't parse the arguments correctly.  This way worked

jail-buddy.lic
dr-leustyin: I don't think so.  I ended up in the 'Pleading' room in RT during testing and 'PLEAD INNOCENT' waited for RT to end before pleading

outdoorsmanship.lic
dr-leustyin: I typically avoid the ternary operator because I think it reduces readability, although what I have in there isn't much better.  Up to you if you want me to correct it.

outdoorsmanship.lic
dr-leustyin: I thought about that, but figured that would mean the method name would need to change to `collect?`, and I worry about breaking other scripts, especially if people have stuff that's not in the repo that uses that method

outdoorsmanship.lic
rpherbig: I find 'simple' (and I consider this to be 'simple enough') ternaries easier to read, but it's definitely a preference thing.

💡 I wonder if coalesce would work here? `@targetxp = args.mindstates || 3`

jail-buddy.lic
rpherbig: Sounds good

jail-buddy.lic
rpherbig: Sounds good

outdoorsmanship.lic
rpherbig: If you were to do this, my recommendation would be to have one PR that _only adds_ `collect?` (and could even be mostly a copy/paste of `collect`) and then another PR a few days later that converts all the scripts to use `collect?` instead of `collect` (and deletes `collect`).

This gives people running multiple characters time to have re-run `common` before the consumer script changes.

PR #2724, opened 2018-06-06
Added capability for chained aim_actions

dr-leustyin:
Added capability to allow cor chained aim_actions:

```aim_fillers:
  Crossbow:
  - gouge left
  - gouge left
  Bow:
  - get bola; lob left; stow bola
  - gouge left
  - gouge left
  Slings:
  - get bola; lob left; stow bola
  - gouge left
  - gouge left```

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2246,8 +2246,11 @@ class AttackProcess
       game_state.loaded = false
       waitrt?
     elsif game_state.loaded
-      fput(game_state.next_aim_action)
-      waitrt?
+      actions = game_state.next_aim_action.split(";")
+      actions.each do | action |
+        fput(action.strip)
+        waitrt?
+      end
     else
       if game_state.dual_load?
         pause 0.5 until load_return = bput('load arrows', 'You reach into', 'already loaded', 'Such a feat would be impossible without the winds to guide', 'but are unable to draw upon its majesty')

### CODE REVIEW

**Comment by rpherbig** [comment]
How would
```
Bow:
  - get bola; lob left; stow bola
```

differ from:
```
Bow:
  - get bola
  - lob left
  - stow bola
```

**Comment by dr-leustyin** [comment]
When separated they're not guaranteed to be sequential, since three
executions of AttackProcess would need to run to execute all three, and you
can't know what might have changed in the interim.  The update guarantees
sequentiality

**Comment by rcuhljr** [comment]
Note aim actions also aren't guaranteed to all occur, so you could get bola, lob left, then not stow bola and leave it laying on the ground. Also I believe left hand actions while aiming need to occur rapidly to avoid the game cancelling your aim for stick bows?

**Comment by rcuhljr** [comment]
I thought `;` was reasonable because I think some mud front ends support multiple commands separated by semicolons already, so people in this domain might find it more understandable and expected behavior.

**Comment by rpherbig** [comment]
Yeah, I was more hoping to think up a better way to represent it in YAML (nested arrays or such).

Does this work?
```
  Bow:
  - - get bola
    - lob left
    - stow bola
  - gouge left
  - gouge left
```

Even if it does, is it any more readable?

rcuhljr:
I prefer nested arrays for keeping with convention of other things we do in the yaml, but I'm at :bulb: not :x:.

PR #2731, opened 2018-06-07
Added town override

dr-leustyin:


sell-loot.lic
```diff
@@ -50,23 +68,13 @@ class SellLoot
...
     else
-      deposit_coins(keep_coppers_bank)
+      deposit_coins(keep_money_by_currency)
     end
   end
 
   def parse(setting)
-    arg_definitions = [
-      [
-        { name: 'amount', regex: /\d+/i, variable: true, description: 'Number of coins to keep' },
```
rpherbig: ❓ Why did you delete these args? I use them sometimes...

sell-loot.lic
```diff
@@ -80,15 +88,15 @@ class SellLoot
...
+  def deposit_coins(keep_money_by_currency)
     walk_to @hometown['deposit']['id']
     release_invisibility
     bput('wealth', 'Wealth:')
     case bput('deposit all', 'you drop all your', 'You hand the clerk some coins', "You don't have any", 'There is no teller here', 'reached the maximum balance I can permit')
     when 'There is no teller here'
       return
     end
-    minimize_coins(keep_copper).each { |amount| withdraw_exact_amount?(amount, @settings) }
+    fput("withdraw #{keep_money_by_currency}")
```
rpherbig: ❓ This doesn't look equivalent. Why did this change?

sell-loot.lic
```diff
@@ -80,15 +88,15 @@ class SellLoot
       .each { |currency| fput "exchange all #{currency} for #{exchange_to}" }
   end
 
-  def deposit_coins(keep_copper)
+  def deposit_coins(keep_money_by_currency)
```
rpherbig: ❓ Could, or should, this be removed in favor of the function in [common-money](https://github.com/rpherbig/dr-scripts/blob/eb5fec5570e7281f32f3408672a1b9d288097e2a/common-money.lic#L119)?

sell-loot.lic
dr-leustyin: It's equivalent due to the 'deposit all' on line 95.  There are no coins to minimize at this point, and withdraw_exact_amount runs to your hometown instead of the bank you're already in

dr-leustyin: Good question.  I can't think of anything, but don't know what edge cases there could be.

jandersson: You may want to branch on the town arg here, as the above two methods seem to handle bankbot transactions, being invisible, withdrawing too quickly, and not having enough money. fput would cover the last two but not the first two

jandersson: e.g. 
if args.town -> fput
else -> minimize+ensure

dr-leustyin: Bankbots get handled on line 65.  This method is only called if no bankbot is set, and line 93 includes the `release_invisibility` check

dr-leustyin: Sorry, bankbots get handled on line 68

jandersson: `withdraw_exact_amount?` withdraws from bankbots and releases invisibility

jandersson: :bulb: im just pointing out that this is going to remove those two things. I dont know who uses them or needs them. The whole banking thing probably needs an overhaul. things seem too tightly coupled (like movement is tied to withdrawing as you pointed out)

dr-leustyin: Right, but you won't be in the bank depositing if you've got a bank bot or invisible.  Those checks have already been made.

Agree on cleaning up some of the banking code.  Maybe that makes more sense to tackle first, but that would be a pretty big task

jandersson: I think branching on your feature, e.g. args.town would still work though? Else just do business as usual

jandersson: I disagree - I added the invisibility check because I train with that as a cyclic spell and it was breaking other things. Id end up stuck at the bank because i cant withdraw. I really think theres an easy solution to this without removing existing functionality.

jandersson: Sorry, I'm probably coming across too strong, didn't intend for that. I just wanted to point out some code that might be affected. Carry on :)

sell-loot.lic
rpherbig: If you call `;sell-loot 1 plat`, then (in the old code) the args are checked _before_ your settings. I use it when I want to end up with a certain amount on hand which is different than my setting.

rpherbig:
I like the 'town override' concept itself. My discomfort with this pair of PRs is that they seem to be making quite a few changes at once:
* Town override
* Introducing a new `deposit_coins` that may? may not? conflict with `common-money`
* Change the way money is fetched for repairs (from `ensure_copper_on_hand` to `get_money_from_bank`)
* Remove the `sell-loot` args
* Change the way money is withdrawn (from `minimize_coins` to `fput`)

Put together, it's a lot to review and I don't feel confident I understand all of the implications of these changes.

dr-leustyin:
Tabling this one for the time bing

PR #2732, opened 2018-06-07
Added town override

dr-leustyin:


crossing-repair.lic
```diff
@@ -22,7 +22,19 @@ class CrossingRepair
       end
     end
 
-    deposit_coins(@keep_copper, settings) unless settings.sell_loot_skip_bank
+    deposit_coins(@settings.sell_loot_money_on_hand) unless @settings.sell_loot_skip_bank
+  end
+  
+  def deposit_coins(keep_money_by_currency)
```
rpherbig: ❓ Is this being used instead of the function in `common-money`?

dr-leustyin: It is.  This was copied nearly verbatim from the sell-loot PR.  It should probably be moved to common-money under a different name, but I was in the zone adding town args.  I could cancel this PR, combine it with 2731, and move this method if it's easier, but was thinking they should be separate since they're distinct functionality changes.

Thoughts?

crossing-repair.lic
```diff
@@ -87,11 +87,30 @@ class CrossingRepair
...
+    if town.start_with?("Hib")
+      town = "Hibarnhvidar"
+    elsif town.start_with?("River")
+      town = "Riverhaven"
+    elsif town.start_with?("Theren")
+      town = "Therenborough"
+    end
+    settings = get_settings
+
+    settings['hometown'] = town if town
```
rpherbig: 💡 I'm not a fan of mutating `settings`, I would have expected something more like `@hometown = town || settings.hometown`

dr-leustyin: I'm not a huge fan of this either.  I think the common-money functions need to be updated to take more specific parameters.  I'll work on getting that refactored, but it'll probably be a pretty big PR because a lot of other scripts touch that code

PR #2733, opened 2018-06-08
Added some checks into chained actions

dr-leustyin:


rpherbig:
❓ What prompted this change?

rpherbig:
I'm a bit less comfortable with this feature now. The pattern matching feels very similar to `before`/`after` for spells: https://github.com/rpherbig/dr-scripts/blob/master/data/base-spells.yaml#L785-L789

I'll merge it since this will fix some problems, but I'd really like to see a better way to represent these actions and/or their matches.

PR #2734, opened 2018-06-08
Added new weararmor script

dr-leustyin:


weararmor.lic
```diff
@@ -0,0 +1,17 @@
+# Simple script to wear a gear set
```
rpherbig: 💡 I prefer this to be a link to elanthipedia

PR #2736, opened 2018-06-08
sorcery.lic: Cast 3 times always instead of exiting when mana is low.

Sarvatt:


rpherbig:
❓ Hmm, this seems like something people might have opinions on. Have you polled LNet or Slack to make sure no one is relying on the current behavior?

Sarvatt:
I did ask once so far and people suggested they'd rather learn more each time it runs instead of it stopping, but I'll ask again another night. My use case is I only train sorcery in crossing-training or in a before: with runs of this script on a character who can't do symb so I only get 2 spells per run of it max before it bails and the later runs bail right away.

PR #2737, opened 2018-06-08
sew.lic: Add magic training support while crafting.

Sarvatt:
Uses the crafting_training_spells: yaml setting, I initially made it use training-spells with discerns but charging cambrinth isn't worth the time unless you really want to.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -877,9 +877,9 @@ class SetupFiles
                                        end
     end
 
-    s.crafting_spell_training
+    s.crafting_training_spells
      .each do |skill, data|
-      s.crafting_spell_training[skill] = if spells[data['name']]
+      s.crafting_training_spells[skill] = if spells[data['name']]
                                          spells[data['name']].merge(data)
                                        elsif spells[data['abbrev']]
                                          spells[data['abbrev']].merge(data)
--- a/dependency.lic
+++ b/dependency.lic
@@ -880,12 +880,12 @@ class SetupFiles
     s.crafting_training_spells
      .each do |skill, data|
       s.crafting_training_spells[skill] = if spells[data['name']]
-                                         spells[data['name']].merge(data)
-                                       elsif spells[data['abbrev']]
-                                         spells[data['abbrev']].merge(data)
-                                       else
-                                         data
-                                       end
+                                            spells[data['name']].merge(data)
+                                          elsif spells[data['abbrev']]
+                                            spells[data['abbrev']].merge(data)
+                                          else
+                                            data
+                                          end
     end
 
     s.lockpick_buffs['spells']
--- a/dependency.lic
+++ b/dependency.lic
@@ -877,6 +877,17 @@ class SetupFiles
                                        end
     end
 
+    s.crafting_spell_training
+     .each do |skill, data|
+      s.crafting_spell_training[skill] = if spells[data['name']]
+                                         spells[data['name']].merge(data)
+                                       elsif spells[data['abbrev']]
+                                         spells[data['abbrev']].merge(data)
+                                       else
+                                         data
+                                       end
+    end
+
     s.lockpick_buffs['spells']
      .each_with_index
      .select do |spell, i|
--- a/sew.lic
+++ b/sew.lic
@@ -310,9 +310,9 @@ class Sew
   end
 
   def cast_spell?
-    return if !@training_spells
-    return if !Flags['magic-ready']
-    return if mana < 40
+    return false if !@training_spells
+    return false if !Flags['magic-ready']
+    return false if mana < 40
 
     unless @settings.cambrinth_items[0]['name']
       @settings.cambrinth_items = [{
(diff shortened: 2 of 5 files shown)


Sarvatt:
don't review this yet please, it's a mess :)

Sarvatt:
This is ready for review now, apologies for submitting it too early.

sew.lic
```diff
@@ -298,9 +301,77 @@ class Sew
         tool = 'sewing needles'
       end
     end
+    magic_routine
```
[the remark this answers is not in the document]
Sarvatt: that is handled in the updated version

sew.lic
```diff
@@ -51,11 +52,13 @@ class Sew
     @noun = args.noun
     @mat_type = args.material
     @stamp = @settings.mark_crafted_goods
+    @training_spells = @settings.crafting_training_spells
```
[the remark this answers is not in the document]
Sarvatt: base and base-empty updated now, not sure about merging data in dependency, I'm not seeing where that applies?

sew.lic
```diff
@@ -298,9 +304,90 @@ class Sew
...
+    cast?(@data['cast'], @data['symbiosis'], @data['before'], @data['after'])
+    Flags.reset('magic-ready')
+    @preparing = false
+  end
+
+  def prepare_spell
+    return if Flags['magic-ready']
+
+    @preparing = true
+    needs_training = %w[Warding Utility Augmentation]
```
Sarvatt: I'm not sure if we want to allow Sorcery here. Thoughts?

dr-leustyin: I'm thinking no, unless you want to deal with the logic to get blown-off hands fixed as well

dr-leustyin: Also, there's potential for lost crafting tools if both hands go away

jayRyan24: The main issue is that if you're crafting in town, which is probably something many Traders are doing, it's illegal to use Sorcery. So, another compounded issue.

jayRyan24:
Do this across the board so I can make a trader, thanks my dude.

sew.lic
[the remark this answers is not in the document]
rpherbig: ❓ : https://github.com/rpherbig/dr-scripts/blob/master/dependency.lic#L843 has some logic to merge spell settings with `base-spells`.

You'll probably want to add a test to `validate.lic` too.

sew.lic
```diff
@@ -367,6 +455,7 @@ class Sew
     when /trash/
       dispose_trash(@noun)
     end
+    magic_cleanup if @training_spells
```
rpherbig: 💡 I recommend moving `if @training_spells` into `magic_cleanup` as a guard clause

sew.lic
rpherbig: I'm on board with not allowing Sorcery.

rpherbig:
> I want to abstract this eventually to use in all crafting scripts and potentially other things like performance that Leus brought up wanting another spell set where you don't use cambrinth since you can cast while performing as long as you don't use it, but I wanted to start here first. I'm not sure if I'll be able to pull it out into common-arcana properly myself.

I'm fine with merging this code into `sew.lic`, but I'd really like to see it refactored into a more reusable state, if possible, before we roll it out to the other crafts.

sew.lic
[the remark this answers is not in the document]
Sarvatt: done and done

dependency.lic
```diff
@@ -877,6 +877,17 @@ class SetupFiles
                                        end
     end
 
+    s.crafting_training_spells
+     .each do |skill, data|
+      s.crafting_training_spells[skill] = if spells[data['name']]
+                                         spells[data['name']].merge(data)
```
jayRyan24: Is the alignment of all of this off, or is it just me?

validate.lic
```diff
@@ -117,6 +117,13 @@ class DRYamlValidator
             .each { |skill_name| error("Invalid crafting_training_spells: skill name '#{skill_name}'. Valid skill names are '#{valid_crafting_training_spell_skills}'") }
   end
 
+  def assert_that_crafting_training_spells_have_no_cambrinth(settings)
+    return unless settings.crafting_training_spells
+
+    settings.crafting_training_spells.select { |_spell, data| data['cambrinth'] }
+            .each { |_spell, _data| warn("#{_name} in crafting_training_spells uses cambrinth which increases crafting time, you may want to remove it.") }
```
rpherbig: ❌  By convention, an underscore in a variable name means it is not used. Seeing both `_spell` and `_data` made me wonder what was being reported in the `.each`. I don't think `_name` is defined.

Sarvatt: I mean it worked, I tested it, but I don't know how to fix it properly so I'll just revert that one.

rpherbig: I think the test was reasonable. I suspect `_name` should have been `_data['name']`.

PR #2739, opened 2018-06-09
Moved the action arrays outside of the loop. 

jayRyan24:
This way they won't be made anew each iteration. I then moved them to the init and made them instance based as they are used each time through, so they may as well be made once only.

This shouldn't affect anything, right? @dr-leustyin

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2088,6 +2088,15 @@ class AttackProcess
     @ambush_location = settings.ambush_location
     echo("  @ambush_location: #{@ambush_location}") if $debug_mode_ct
 
+    @get_actions = ['get','wield']
+    echo("  @get_actions: #{@get_actions}") if $debug_mode_ct
+
+    @rt_actions = ['gouge','attack','jab','feint','draw','lunge','slice']
+    echo("  @rt_actions: #{@rt_actions}") if $debug_mode_ct
+
+    @stow_actions = ['stow','sheath','put']
+    echo("  @stow_actions: #{@stow_actions}") if $debug_mode_ct
+
     Flags.add('ct-ranged-ready', 'You think you have your best shot possible now')
     Flags.add('ct-face-what', 'Face what')
   end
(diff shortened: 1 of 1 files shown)


jayRyan24:
As a side question, why can't the matches just be all of them, and always pass them?

dr-leustyin:
What do you mean, 'always pass them'?

jayRyan24:
Why do we only pass them to the bput method based on if we need them? Why not just send the whole shabang at once? Just make them the whole set and pass them all?

jayRyan24:
```
@matches = [/^You /, "You get", "You draw out", "You pick up", "You get","You draw out", "You pick up", "Roundtime", "You put", "You sheathe"]```
```
actions = game_state.next_aim_action.split(";")
actions.each do | action |
  bput(action.strip, @matches)
  pause 0.25
  waitrt?
end
```

**Comment by rpherbig** [comment]
@jayRyan24 over in #2733, @dr-leustyin said
> The default /^You /, is better, but incomplete as it would sometimes be triggered by 'You are fully rested' or 'You block with your shield' messaging, so I added some more specific checks for a non-exhaustive set of potential actions.

Which implies to me that we only _sometimes_ want to use `/^You /`, and use the others when possible.

**Comment by jayRyan24** [comment]
Ah yes, thanks.

File: combat-trainer.lic [hunk 1]
```diff
@@ -2088,6 +2088,15 @@ class AttackProcess
     @ambush_location = settings.ambush_location
     echo("  @ambush_location: #{@ambush_location}") if $debug_mode_ct
 
+    @get_actions = %w[get wield]
+    echo("  @get_actions: #{@get_actions}") if $debug_mode_ct
```
rpherbig: 💡 I'm not sure if there's much value in echoing something that isn't variable. On the other hand, it's not that much extra logging.

rpherbig:
I'm fine with merging this @jayRyan24 - or did you want to wait on any further discussion with @dr-leustyin ?

combat-trainer.lic
jayRyan24: Thought this too, but it could be helpful with debugging what is expected vs what we put in. It's only output once, so I figured it would be ok.

jayRyan24:
It's practically the same thing, just we don't constantly make new arrays.

jayRyan24:
At a minimum, this could be done.
```
    elsif game_state.loaded
      actions = game_state.next_aim_action.split(";")
      get_actions = %w[get wield]
      rt_actions = %w[gouge attack jab feint draw lunge slice]
      stow_actions = %w[stow sheath put]
      actions.each do | action |
        first_word = action.split(" ")[0].downcase
        matches = [/^You /]
        matches = ["You get","You draw out","You pick up"] if get_actions.include?(first_word)
        matches = ["Roundtime"] if rt_actions.include?(first_word)
        matches = ["You put", "You sheathe"] if stow_actions.include?(first_word)
        bput(action.strip, matches)
        pause 0.25
        waitrt?
      end
```

jayRyan24:
@dr-leustyin

dr-leustyin:
I've got no problems with the updates

jayRyan24:
@rpherbig Think we're good.

PR #2741, opened 2018-06-10
Added reference to common-money

dr-leustyin:


Sarvatt:
This fails if you get busted with jail-buddy loaded before another script has loaded common-money, i hit that earlier today getting busted right at login after the autostart loaded.

PR #2748, opened 2018-06-12
sew.lic: Update to use common-arcana methods.

Sarvatt:


Sarvatt:
I removed releasing symbiosis, because if you have one researched you can lose the researched one if you try to release it again within 30 seconds.

rpherbig:
Remind me to merge this in a few days so users have a chance to get the new `common-arcana` with the helpers

PR #2752, opened 2018-06-12
Update all crafting scripts to support magic training.

Sarvatt:


Sarvatt:
Please don't pull yet, I want to test them all out first but wanted them here for review.

carve.lic
```diff
@@ -126,9 +131,39 @@ class Carve
       command = "cut #{@my} #{@noun} with my #{@main_tool}"
     end
     waitrt?
+    magic_routtine
```
jandersson: typo

forge.lic
```diff
@@ -457,6 +465,35 @@ class Forge
       analyze_item
     end
   end
+
+  def magic_routine
```
jandersson: 💡 Is it worth commonizing this?

Sarvatt:
tested all of them and fixed the problems I hit, they are all working properly. I'm happy with this :)

Sarvatt:
I actually don't think it would be an issue pulling it in before people get updated common-arcana, they have to manually opt-in to having a list of crafting_training_spells or else the magic routines are a noop. magic_routine returns if it's not set and nothing from common-arcana is used in that case.

Sarvatt:
Sorry for submitting so many at once, but they are basically just cut and pastes of each other. Forge and remedy are the only ones that aren't. I did try to test every scenario on them and I'm confident they are good to go now after fixing up problems I hit.

carve.lic
```diff
@@ -41,6 +43,7 @@ class Carve
 
     wait_for_script_to_complete('buff', ['carve'])
     Flags.add('carve-assembly', 'another finished wooden (hilt|haft)', 'another finished (long|short) wooden (pole)')
+    Flags.add('magic-ready', 'You feel fully prepared to cast your spell.')
```
rpherbig: 💡 Flags are global. That means each of these scripts is reading the same Flag. Since you `.add` at the top, this is ok (`add` sets it to false), but just be aware.

If you wanted to be extra safe, you might add a prefix or suffic to make each Flag uniquely named.

Sarvatt: should they be deleted on script close then?

Sarvatt: unique makes sense, I'll change that just in case.

Sarvatt: it explains why I needed to use if Flags['magic-ready'].include?('You feel fully prepared to cast your spell.') at least, Flags['magic-ready'] was true at odd times while testing.

PR #2755, opened 2018-06-13
safe-room.lic: use in_game? from common-validation.

Sarvatt:


safe-room.lic
```diff
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#safe-room
 =end
 
-custom_require.call(%w[common common-arcana common-healing common-items common-money common-travel drinfomon events])
+custom_require.call(%w[common common-arcana common-healing common-items common-money common-travel common-validation drinfomon events])
 
 class SafeRoom
   include DRC
```
rpherbig: ❌ I _think_ you need to `include CharacterValidator` here, can you test this?

rpherbig: @Sarvatt Hmm, that probably won't work because `CharacterValidator` is a class, not a module.

When I try to run `safe-room` I get:
```
--- Lich: error: undefined method `in_game?' for                   |
              #<Scripting::SafeRoom:0x30d1c10 @health_threshold=25>            
                    safe-room:188:in `block in use_pc_empaths?'                
                    safe-room:187:in `each'                                    
--- Lich: safe-room has exited.
```

rpherbig: I _think_ this will work:
```
module DRCV
  module_function

  def in_game?(character)
    "  #{character}." == bput("find #{character}", 'There are no adventurers in the realms that match the names specified', "^  #{character}.$")
  end
end
```

PR #2758, opened 2018-06-14
common-validation and safe-room changes to use in_game method

bmcneff:


common-validation.lic
```diff
@@ -64,6 +68,6 @@ class CharacterValidator
   end
 
   def in_game?(character)
-    "  #{character}." == bput("find #{character}", 'There are no adventurers in the realms that match the names specified', "^  #{character}.$")
+    "  #{character}." == DRC.bput("find #{character}", 'There are no adventurers in the realms that match the names specified', "^  #{character}.$")
```
jandersson: :bulb: You can remove DRC here since you included it at the top of the class.

PR #2759, opened 2018-06-14
Make crafting magic flags unique.

Sarvatt:


rcuhljr:
:bulb:  Why are you doing `Flags['carve-magic-ready'].include?('You feel fully prepared to cast your spell.')` instead of `Flags['carve-magic-ready']` in the if block?

Sarvatt:
if Flags['magic-ready'] was unreliable when I was testing it, working sometimes and not working others but include? worked and I have 0 clue why it was happening. I'll try it again, but yeah the point of this commit is to just make the flags unique to each script.

rcuhljr:
Yeah someone should track down what error you were seeing because It'd be preferable not to have anti patterns in the code base or unexpected behavior in libraries.

PR #2762, opened 2018-06-15
Athletics.lic - added climbing in xala'shar

BinuDR:
I've added climbing for the climbs in xala'shars if your hometown is set to Crossing and you have 560 or more athletics.
I need more people to test if that limit is high enough for the very last couple of climbs.

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -23,7 +23,8 @@ class Athletics
     arg_definitions = [
       [
         { name: 'wyvern', regex: /^wy.*/i, optional: true, description: 'Climb both undergondola and wyvern cliffs' },
-        { name: 'undergondola', regex: /^un.*/i, optional: true, description: 'Climb branch etc undergondola' }
+        { name: 'undergondola', regex: /^un.*/i, optional: true, description: 'Climb branch etc undergondola' },
+		{ name: 'xalas', regex: /^xala.*/i, optional: true, description: 'Climb xalas in zoluren in instead of undergondola' }
       ]
     ]
 
@@ -33,6 +34,8 @@ class Athletics
       climb_wyvern
     elsif args.undergondola
       climb_branch
+	elsif args.xalas
+	  climb_xalas
     elsif have_climbing_rope
       train_with_rope
     elsif @swimming_target
@@ -207,11 +210,10 @@ class Athletics
   end
 
   def crossing_athletics
-    if DRSkill.getrank('Athletics') >= 560
-	  climb_xalas
-    elsif DRSkill.getrank('Athletics') >= 400
+    if DRSkill.getrank('Athletics') >= 400
       climb_branch
     else
+      start_script('performance') unless Script.running?('performance')
       until done_training?
         @athletics_options
           .reject { |data| @settings.avoid_athletics_in_justice && data['justice'] }
@@ -290,6 +292,8 @@ class Athletics
           remaining_attempts -= 1
         end
         return if remaining_attempts.zero?
+      else
+        start_script('performance') unless Script.running?('performance')
       end
       Flags.add('ct-climbing-finished', 'You finish practicing your climbing')
       Flags.add('ct-climbing-combat', 'You are engaged in combat')
@@ -333,10 +337,9 @@ class Athletics
         walk_to(19_464)
         break if done_training?
         walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-        train_outdoorsmanship
-        train_outdoorsmanship
-        train_outdoorsmanship
-        train_outdoorsmanship
+		4.times do
+          train_outdoorsmanship
+		end
       end
     end
   end
@@ -355,12 +358,12 @@ class Athletics
       walk_to(6443)
       break if done_training?
       walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-      train_outdoorsmanship
-      train_outdoorsmanship
-      train_outdoorsmanship
+	  4.times do
+        train_outdoorsmanship
+	  end
     end
   end
-
+  
   def climb_xalas
     start_script('skill-recorder') unless Script.running?('skill-recorder')
     walk_to(6154)
@@ -368,9 +371,10 @@ class Athletics
       walk_to(12838)
       walk_to(6154)
       break if done_training?
+      walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
       4.times do
         train_outdoorsmanship
-      end      
+      end
     end
   end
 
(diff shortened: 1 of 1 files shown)


Sarvatt:
Definitely going to need some testing here to be sure its climbable at 560, it's also a shame to cut off undergondola which is amazing training from 540-1000. how about making it an arg like wyvern and undergondola is now so people can call it with athletics xala in a before? athletics.lic isn't used by crossing-training except in the case of ropes, so its pretty safe adding an arg like that

jandersson:
At some point I think athletics should be refactored to be hometown agnostic

BinuDR:
Per Jonas: 650 athletics, ~10 mindstates per run through, no RT, no getting stuck.

rpherbig:
The code itself looks fine. I'll wait to merge until @BinuDR says its tested and ready to go.

BinuDR:
I just need more testing and some people to vote whether I should leave it as is or make it an argument.
One idea I had which I didn't implement yet was to use DRCT.sort_destinations to determine if the north side or the south side of the segoltha is closer.

If the north side is closer it could run and climb xalas and then run to climb the undergondola.  Or just do xalas.

rpherbig:
@BinuDR - any updates on this?

BinuDR:
We might need to do a little more testing on the lower and upper limits of what it teaches.  I can still lock at 980+ but it takes a bit longer.  At that point the undergondola might be faster.

Do you think it's a good idea to put a range on it?  Or just make it a variable like Sarvatt suggested?

rpherbig:
I don't know enough about this to have an informed opinion. Maybe @jayRyan24 or @dr-leustyin have some ideas?

dr-leustyin:
I'm think it would be nice to be able to break out something like

```athletics_options:
- undergondola
- wyvern
- xala
```

and then the script would run through each of them.  Would decouple athletics from hometown, which is nice at high level, and would let f2p pick the things that they can get to.

Something like 
```atheltics_options:
- crossing_walls
- climb_practice
```

would run the walls once and then use the climb practice target until locked.

Doesn't need to be part of this commit, but having the xala_climbing and wyvern_climbing options mean this is something we could do pretty easily in later updates.

BinuDR:
Alright, I'll work on making it a variable you can call instead of an automatic destination based on hometown.

dr-leustyin:
Maybe do xala first, then undergondola. I personally do that, plus wyvern
in my climbs. Reduces the need for the outdoorsmanship training since
travel takes care of the cooldown

jayRyan24:
So, looks like this is the gains I get, 4-5. I wouldn't collect in that location though, due to the company.

```Athletics:   1028 56.34% thinking       (5/34)```
```
[athletics: Circuit Exp Gains: 4]
[athletics]>collect rock
>
```
```
  limbs, silent testimony to their makers lack of skill.  You also see some braided vines.                        |
Also here: Bloodwraith Cayra and Deadly Shadow Fahijeck who has a fiery visage.                                   |
Obvious paths: southwest.                                                                                         |
Room Number: 6154                                                                                                 |
>                                                                                                                 |
[go2: travel time: 0:00:01]                                                                                       |
--- Lich: go2 has exited.                                                                                         |
[athletics: Circuit Exp Gains: 5]                                                                                 |
[athletics]>collect rock                                                                                          |
You manage to collect a pile of rocks.                                                                            |
Roundtime: 15 sec.                                                                                                |
>                                                                                                                 |
You feel fully rested.                                                                                            |
>                                                                                                                 |
Fahijeck's feet are soaked.                                                                                       |
>                                                                                                                 |
Cayra begins to carefully examine a padded heavy titanese hauberk.                                                |
>                                                                                                                 |
 * Three plaintive bell-tolls carried on the wind harbinger Maltris's arrival.                                    |
>                                                                                                                 |
Cayra begins to carefully examine a puffy-white targe sealed with protective wax.                                 |
[athletics]>kick pile                                                                                             |
[Assuming you mean a pile of rocks.]                                                                              |
You take a step back and run up to the pile of rocks.  Bringing your foot smashing down onto the pile, you send   |
  bits and pieces of debris scattering in every direction.  Looking around for the now defunct mountain of glory, |
  you fail to find any signs of its presence.                                                                     |
[athletics]>collect rock                                                                                          |
You begin scanning the area before you, trying to find just the right spot to forage.                             |
You manage to collect a pile of rocks.                                                                            |
```

**Comment by jayRyan24** [comment]
Ssarek gained at least 6. didn't get stuck, Athletics:  624.

**Comment by jayRyan24** [comment]
@BinuDR Look at what I put in for athletic outdoorsmanship rooms. You could just implement that into yours.  Gives all the people in athletics Crossing xala to use random rooms and not all be clumped together.

**Comment by Sarvatt** [comment]
https://github.com/rpherbig/dr-scripts/blob/master/athletics.lic#L336 @BinuDR

**Comment by jayRyan24** [comment]
https://github.com/rpherbig/dr-scripts/blob/master/profiles/Chuno-setup.yaml#L26-L31

```
athletics_outdoorsmanship_rooms: 
- ####
- ####
- ####
- ####
- ####
```

**Comment by Sarvatt** [comment]
this needs to be refreshed against the current athletics

**Comment by BinuDR** [comment]
Okay, it's been updated.

File: athletics.lic [hunk 1]
```diff
@@ -353,9 +358,23 @@ class Athletics
       walk_to(6443)
       break if done_training?
       walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-      train_outdoorsmanship
-      train_outdoorsmanship
-      train_outdoorsmanship
+	  4.times do
```
Sarvatt: this should be 3.times, the travel to and from the area takes over 15 seconds thats why its shorter

BinuDR: Okay

Sarvatt:
other than the 3.times change for wyvern this looks good to me, but it will need a rubocop run after to fix the spacing issues, he can't do that on the website.

jayRyan24:
@BinuDR Just a FYI. A lot of the suggestions here have bloated the scope of this PR. Possibly the best course of action is to get your base changes merged and then make a new PR with the focus on breaking athletics out from the Town into a more listed approach. What do you think?

BinuDR:
It looks like I added a lot but I really didn't.  I had to update the version of athletics that I had originally used because some other changes were added in.

These are all the additions I really made.  That plus I condensed the outdoordsmanship sections to loops in the other methods.

+        { name: 'xalas', regex: /^xala.*/i, optional: true, description: 'Climb xalas in zoluren in instead of undergondola' }

+    elsif args.xalas
+      climb_xalas

+  def climb_xalas
+    start_script('skill-recorder') unless Script.running?('skill-recorder')
+    walk_to(6154)
+    until done_training?
+      walk_to(12838)
+      walk_to(6154)
+      break if done_training?
+      walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
+      4.times do
+        train_outdoorsmanship
+      end
+     end
+     end

BinuDR:
@rpherbig This is ready unless you want me to separate my changes and make a 2nd PR for the rest.

PR #2764, opened 2018-06-15
Delete crafting magic flags on script exit.

Sarvatt:


carve.lic
```diff
@@ -175,6 +175,7 @@ class Carve
     end
     waitrt?
     magic_cleanup
+    Flags.delete('carve-magic-ready')
```
rpherbig: ❓  Should the Flag deletion be part of `magic_cleanup`? (in each of these files)

Sarvatt:
whoops! yes you're right, magic_cleanup used to be called before every prep and not just at exit like it is now so it slipped past me. thanks!

PR #2765, opened 2018-06-15
Update all callers of settings.hometown to automatically capitalize

Sarvatt:


jandersson:
:bulb: What do you think about doing this in `transform_settings` in dependency? This would do it the capitalization anytime get_settings is done

Sarvatt:
much better idea :)

rpherbig:
`athletics.lic` should also be doing case-insensitive checking as a matter of best practice

PR #2768, opened 2018-06-15
skill-recorder.lic: Check for khri flight and unyielding for athletic…

Sarvatt:
…s buffs.

### DIFF
--- a/skill-recorder.lic
+++ b/skill-recorder.lic
@@ -17,7 +17,7 @@ loop do
   pause 2
   clear
   UserVars.athletics = DRSkill.getrank('Athletics')
-  if DRSpells.active_spells['Athleticism']
+  if DRSpells.active_spells['Athleticism'] || DRSpells.active_spells['Khri Flight'] || DRSpells.active_spells['Unyielding']
     UserVars.athletics = UserVars.athletics * 1.1
   end
   UserVars.scouting = DRSkill.getrank('Scouting')


skill-recorder.lic
```diff
@@ -17,7 +17,7 @@ loop do
   pause 2
   clear
   UserVars.athletics = DRSkill.getrank('Athletics')
-  if DRSpells.active_spells['Athleticism']
+  if DRSpells.active_spells['Athleticism'] || DRSpells.active_spells['Khri Flight'] || DRSpells.active_spells['Unyielding']
     UserVars.athletics = UserVars.athletics * 1.1
```
Sarvatt: what do you think about making this 15% instead of 10%?

Sarvatt:
note: I am unsure if Unyielding is how the barb meditiation shows up in active spells, I have yet to find a barb that actually knows it and it is not in barb_abilities

rpherbig:
My barb's abilities show up as `Flashflood` (I don't have Unyielding to test with), so this code seems fine.

skill-recorder.lic
rpherbig: Since we don't know how much mana was put into Athleticism, 10% seems safer.

Sarvatt: yeah 10% makes the most sense since it's the bare minimum. I'll be honest, I didn't even know having a buff up would effect your climbing in ;go2 until I saw this script today and the only levels where I would have noticed it I had 15% at least which is why I was asking. (segoltha and undergondola)

PR #2775, opened 2018-06-17
Added performance to athletics training

dr-leustyin:


athletics.lic
```diff
@@ -210,6 +210,7 @@ class Athletics
     if DRSkill.getrank('Athletics') >= 400
       climb_branch
     else
+      start_script('performance') if not Script.running?('performance')
```
rpherbig: 💡 `!` instead of `not`

PR #2776, opened 2018-06-17
Fixed bug in 'careful' disarming

dr-leustyin:
Eliminated messages which cause a 'disarm careful' even if the trap is too difficult

### DIFF
--- a/data/base-picking.yaml
+++ b/data/base-picking.yaml
@@ -73,9 +73,6 @@ picking:
   - A pitiful snowball encased in the Flames
 
   disarm_careful:
-  - covered black scarab of some unidentifiable substance clings to the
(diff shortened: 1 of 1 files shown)


data/base-picking.yaml
```diff
@@ -73,9 +73,6 @@ picking:
   - A pitiful snowball encased in the Flames
 
   disarm_careful:
-  - covered black scarab of some unidentifiable substance clings to the
```
rpherbig: ❓ Should any of these be moved to `disarm_too_hard`?

dr-leustyin:
Not in my opinion. The error here is mixing in traps with difficulties,
which this PR resolves. It wouldn't make sense to dump a trivial scarab
trap, for example.

PR #2786, opened 2018-06-19
Adding offhand-trainables during aiming.

jayRyan24:


combat-trainer.lic
```diff
@@ -2243,6 +2244,7 @@ class AttackProcess
     game_state.selected_maneuver = charged_maneuver unless game_state.loaded
     game_state.loaded = false if game_state.mob_died && !(game_state.casting_consume || game_state.prepare_consume)
     game_state.clear_aim_queue if Flags['ct-ranged-ready']
+    game_state.loaded = false if Flags['ct-aim-failed']
```
dr-leustyin: I think this should be game_state.aiming, not loaded

combat-trainer.lic
```diff
@@ -2259,16 +2261,33 @@ class AttackProcess
...
-      actions.each do |action|
-        first_word = action.split(' ')[0].downcase
-        matches = [/^You /]
-        matches = ['You get', 'You draw out', 'You pick up', "You're already holding that"] if @get_actions.include?(first_word)
-        matches = ['Roundtime',"You aren't close enough to attack"] if @rt_actions.include?(first_word)
-        matches = ['You put', 'You sheathe'] if @stow_actions.include?(first_word)
-        bput(action.strip, matches)
-        pause 0.25
-        waitrt?
+      if game_state.use_offhand_trainables?
```
dr-leustyin: Rather than a setting, What about checking whether `@offhand_trainables` is set?

combat-trainer.lic
jayRyan24: There is no `aiming` method in GameState. There is a `done_aiming?` which is a check to see if the aiming action array is empty (which is triggered to be empty by the line above this one if the aim is complete. So, setting loaded to false here will cause another "load" which will then begin to aim. This is just simply how the method works for this.

combat-trainer.lic
```diff
@@ -2259,16 +2261,33 @@ class AttackProcess
...
+        skill = game_state.determine_offhand_weapon
+        skill = 'Brawling' if game_state.current_weapon_skill.eql?'Bow'
+        if skill.eql?'Brawling'
+          3.times{bput('gouge left', 'You')}
+          return
+        end
+
+        weap = game_state.wield_offhand_weapon(skill)
```
dr-leustyin: If skill is Light Thrown I think it should lob, otherwise it should jab.

combat-trainer.lic
```diff
@@ -2259,16 +2261,33 @@ class AttackProcess
...
+        %w[lob jab].each do |action|
+          result = bput("#{action} left", 'Roundtime', 'close enough')
+          fput('eng') if result =~ /close enough/i
+          bput("get my #{weap}", 'You pick up') if action.eql?'lob'
```
dr-leustyin: I _think_ I may have had an issue at some point when trying to pick up a bonded weapon instead of using `invoke`

combat-trainer.lic
jayRyan24: That's what this done, but it's in the game_state. `use_offhand_trainables?` is a method that returns if the array is empty. I guess I could just make an accessor to @offhand_trainables in GameState.

combat-trainer.lic
jayRyan24: I think it should do both? We kind of get double bang for our buck?

jayRyan24: I could definitely make different action arrays based on the skill containing throw.

dr-leustyin: Not everyone will train LT (even though they should) so they would waste time lobbing at something they couldn't hit.  Another side effect would be if they have super high LT but are doing Xbow in a backhunt.  LT would murder things

combat-trainer.lic
dr-leustyin: Does it try to re-load the weapon after a failure?

jayRyan24: Yes, that's just the way the system works as is. So, it will hit the last else in the if chain in the `attack_aimed` method. It will hit the "already loaded" and then go about the normal aiming, etc.

combat-trainer.lic
jayRyan24: Fair points! So, how do we want to handle this?

jayRyan24: Wait, this is why I have a trainables array. They simply don't include Light Thrown.

combat-trainer.lic
```diff
@@ -2259,16 +2261,33 @@ class AttackProcess
...
-        matches = [/^You /]
-        matches = ['You get', 'You draw out', 'You pick up', "You're already holding that"] if @get_actions.include?(first_word)
-        matches = ['Roundtime',"You aren't close enough to attack"] if @rt_actions.include?(first_word)
-        matches = ['You put', 'You sheathe'] if @stow_actions.include?(first_word)
-        bput(action.strip, matches)
-        pause 0.25
-        waitrt?
+      if game_state.use_offhand_trainables?
+        skill = game_state.determine_offhand_weapon
+        skill = 'Brawling' if game_state.current_weapon_skill.eql?'Bow'
```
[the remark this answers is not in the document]
jayRyan24: Aaaaaah. Ok, I just need a list of attacks for each weapon. Lemme know.

combat-trainer.lic
dr-leustyin: Not me.  I use bola, throwing hammer, and javelin.  There's a penalty for using non-specifically-throwable items.  You can lob the nightstick, but you're getting an additional penalty

combat-trainer.lic
jayRyan24: hm, you can pick up or invoke any weapon, this seems terribly weird if it's a bug. :( I don't know how to handle that with the system's current implementation. I'd need to add in a method to handle getting a weapons current info from game_states equipment manager.

combat-trainer.lic
```diff
@@ -2259,16 +2261,42 @@ class AttackProcess
...
-        pause 0.25
-        waitrt?
+      if game_state.aiming_trainables
+        skill = game_state.determine_aiming_skill
+
+        if skill.eql?('Brawling')
+          3.times{ return if !execute_aiming_action?('gouge', game_state) }
+          return
+        elsif skill.eql?('Tactics')
+          %w[bob weave circle].each do |action|
```
rpherbig: 💡 These are the same actions on line 1779. Maybe we need a `$TACTICS_ACTIONS` somewhere...

combat-trainer.lic
```diff
@@ -2259,16 +2261,42 @@ class AttackProcess
...
+            return if !execute_aiming_action?(action, game_state)
+          end
+          return
+        end
+
+        weap = game_state.wield_offhand_weapon(skill)
+        if skill.include?('Thrown')
+          actions = %w[lob lob lob]
+        else
+          actions = %w[jab feint draw]
```
rpherbig: 💡 If we're using weak attacks, we only `jab` (see line 3097).  Do we want to `feint` and `draw` here? Or have a conditional on `use_weak_attacks`?

combat-trainer.lic
```diff
@@ -2259,16 +2261,42 @@ class AttackProcess
...
-        matches = ['Roundtime',"You aren't close enough to attack"] if @rt_actions.include?(first_word)
-        matches = ['You put', 'You sheathe'] if @stow_actions.include?(first_word)
-        bput(action.strip, matches)
-        pause 0.25
-        waitrt?
+      if game_state.aiming_trainables
+        skill = game_state.determine_aiming_skill
+
+        if skill.eql?('Brawling')
+          3.times{ return if !execute_aiming_action?('gouge', game_state) }
```
rpherbig: 💡 There might be some room to re-use some code between Brawling and Tactics. After all, this line is the same as `%w[gouge gouge gouge]`

combat-trainer.lic
```diff
@@ -2259,16 +2261,42 @@ class AttackProcess
...
+        if skill.include?('Thrown')
+          actions = %w[lob lob lob]
+        else
+          actions = %w[jab feint draw]
+        end
+        actions.each do |action|
+          break if !execute_aiming_action?(action, game_state)
+          bput("get my #{weap}", 'You pick up') if action.eql?('lob')
+        end
+        game_state.sheath_offhand_weapon(skill) if left_hand
```
rpherbig: 💡 You might consider adding a guard clause to `sheath_offhand_weapon` (something like `return unless weapon`) and then you don't need the conditional here

combat-trainer.lic
```diff
@@ -2259,16 +2261,42 @@ class AttackProcess
...
+          3.times{ return if !execute_aiming_action?('gouge', game_state) }
+          return
+        elsif skill.eql?('Tactics')
+          %w[bob weave circle].each do |action|
+            return if !execute_aiming_action?(action, game_state)
+          end
+          return
+        end
+
+        weap = game_state.wield_offhand_weapon(skill)
```
rpherbig: 💡 I'd prefer `weapon` or `weapon_name`

combat-trainer.lic
rpherbig: Or there might be a more general pattern here. The skill determines the array of actions, and you can make `wield_offhand_weapon` and `sheath_offhand_weapon` a little more defensive then always call them, even for Brawling/Tactics.

I suspect this block would get somewhat smaller, and the pattern would be more apparent. I don't know if it would be any clearer overall though.

combat-trainer.lic
```diff
@@ -2259,16 +2261,42 @@ class AttackProcess
...
-      actions.each do |action|
-        first_word = action.split(' ')[0].downcase
-        matches = [/^You /]
-        matches = ['You get', 'You draw out', 'You pick up', "You're already holding that"] if @get_actions.include?(first_word)
-        matches = ['Roundtime',"You aren't close enough to attack"] if @rt_actions.include?(first_word)
-        matches = ['You put', 'You sheathe'] if @stow_actions.include?(first_word)
-        bput(action.strip, matches)
-        pause 0.25
-        waitrt?
+      if game_state.aiming_trainables
```
rpherbig: 💡 You could remove a level of indentation by combining lines 2263 and 2264 into `elsif game_state.loaded && game_state.aiming_trainables` and then line 2288 becomes ``elsif game_state.loaded`.

combat-trainer.lic
```diff
@@ -3077,6 +3129,17 @@ class GameState
     @construct_mode
   end
 
+  def determine_aiming_skill
+      held_types = ['Small Edged','Large Edged', 'Small Blunt','Large Blunt', 'Staves', 'Polearms']
+      options = @aiming_trainables
+      options = options - held_types if weapon_skill.eql?'Bow' or weapon_skill.eql?'Crossbow'
```
rpherbig: 💡 `options = options - held_types` could be `options -= held_types`

profiles/base.yaml
```diff
@@ -115,6 +115,7 @@ dump_junk: false
 dump_item_count: 9
 use_barb_combos: false
 whirlwind_trainables:
+aiming_trainables:
```
rpherbig: Do you have a YAML with an example of this setting?

combat-trainer.lic
```diff
@@ -3077,6 +3129,17 @@ class GameState
     @construct_mode
   end
 
+  def determine_aiming_skill
+      held_types = ['Small Edged','Large Edged', 'Small Blunt','Large Blunt', 'Staves', 'Polearms']
```
rpherbig: 💡 Seeing these weapon skills makes me think of how weapon skills are listed in https://github.com/rpherbig/dr-scripts/blob/master/validate.lic#L600 - maybe there's a better place for that data

combat-trainer.lic
jayRyan24: I was going to do this, but then figured we'd deprecate the other code. Then realized it has it's uses, so left it in. I did the suggested.

combat-trainer.lic
jayRyan24: I thought of doing this, with something like whirlwind_trainables, etc. I'll look more into it.

rpherbig:
Most of these comments are 💡 , I just want to know 1) if there's an example YAML; and 2) if you feel you've tested this thoroughly.

dr-leustyin:
I tested it today as well on two characters with no issues.  Really like this addition

combat-trainer.lic
jayRyan24: Ok, handled this.

combat-trainer.lic
jayRyan24: Done.

combat-trainer.lic
jayRyan24: Yeah, I actually have a modified version of weak attacks which does a sample of jab or feint. I know Leu does Draw on his so I have this setup. What I can do is jab/feint/draw unless weak attacks, then jab only?

combat-trainer.lic
jayRyan24: Actually, the weapon in sheath-offhand_weapon was misleading, as it takes in a skill and uses the weapon_list[skill] for handling that. We could always have a weapon, but it might have failed to pick it up? I can just remove the condition, either way it's going to be an issue. I also added a conditional in case the brawling/tactics for the previous changes suggested, so that should handle that.

combat-trainer.lic
jayRyan24: Ah, yeah, I do this in my other PR. Just spaced, thanks!

profiles/base.yaml
jayRyan24: ah yes!

jayRyan24:
I've been running it on 3 chars over night and all day today

```
  Parry Ability:    369 38% riveted       (28/34)     Small Edged:    331 84% pondering      (7/34)
    Large Edged:    240 49% learning       (3/34) Twohanded Edged:    240 11% thoughtful     (4/34)
    Small Blunt:    242 46% thoughtful     (4/34)     Large Blunt:    237 93% learning       (3/34)
Twohanded Blunt:    239 75% thoughtful     (4/34)          Slings:    200 44% learning       (3/34)
            Bow:    200 99% thinking       (5/34)        Crossbow:    200 67% learning       (3/34)
         Staves:    200 41% learning       (3/34)        Polearms:    331 67% pondering      (7/34)
   Light Thrown:    331 48% concentrating  (9/34)    Heavy Thrown:    337 93% pondering      (7/34)
       Brawling:    332 18% captivated    (26/34)  Offhand Weapon:    200 35% thoughtful     (4/34)
  Melee Mastery:    396 20% engrossed     (27/34) Missile Mastery:    382 52% riveted       (28/34)
```
@dr-leustyin Get the new changes and run them through til tomorrow with me. They aren't logic changes, just mainly cosmetic, so they shouldn't be an issue, I just want to double check.

jayRyan24:
I ran this overnight and all day on three chars. Working fine. No known issues.

PR #2793, opened 2018-06-20
Added combat picking

dr-leustyin:


Sarvatt:
Could you add Locks to https://github.com/rpherbig/dr-scripts/blob/master/validate.lic#L381 also?

combat-trainer.lic
```diff
@@ -1779,6 +1782,32 @@ class TrainerProcess
       bput(%w[weave bob circle].sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty?
     when 'Analyze'
       analyze(game_state, 0)
+    when 'Locks'
+      return if game_state.loaded
```
rpherbig: 💡 I highly recommend extracting this into a helper method (leaving this function to mostly be managing control flow)

combat-trainer.lic
```diff
@@ -1779,6 +1782,32 @@ class TrainerProcess
...
+      return if left_hand
+      boxes = get_boxes(@pet_box_source)
+      return if boxes.length == 0
+      box = boxes.first
+      # Check what's in the right hand so we can get it back
+      rhitem = @equipment_manager.item_by_desc(right_hand) if right_hand
```
rpherbig: ❓ Wouldn't the `rhitem` normally be a weapon? We have a method on `game_state` that can 'get rid' of the weapon safely (and one that gets it back).

combat-trainer.lic
```diff
@@ -1779,6 +1782,32 @@ class TrainerProcess
...
+      @equipment_manager.stow_weapon("#{rhitem.adjective} #{rhitem.name}") if rhitem
+      return if right_hand  # Couldn't empty right hand for some reason (typically a summoned weapon).  No picking allowed
```
rpherbig: ❓ If you check the summoning code, there's some logic to check if your current weapon is summoned. I suggest checking for that, outputting a message, and returning. What do you think?

combat-trainer.lic
```diff
@@ -1779,6 +1782,32 @@ class TrainerProcess
...
+      bput("get #{box} from my #{@pet_box_source}","You get", "What were")
+      retreat
+      response =  bput("pick my #{box} blind", 'not even locked', 'Roundtime', 'Find a more appropriate tool','You realize that would be next to impossible')
+      if response == 'not even locked'
+        # Dump the box so that combat-trainer lootprocess can take over
+        fput("open my #{box}")
```
rpherbig: ❓ Could any or all of these three be converted to the more reliable `bput`?

rpherbig: And/or line 1810

combat-trainer.lic
```diff
@@ -1779,6 +1782,32 @@ class TrainerProcess
       bput(%w[weave bob circle].sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty?
     when 'Analyze'
       analyze(game_state, 0)
+    when 'Locks'
+      return if game_state.loaded
+      return if left_hand
+      boxes = get_boxes(@pet_box_source)
```
rpherbig: ❓ Would you want to check if `@pet_box_source` is defined?

validate.lic
```diff
@@ -378,7 +378,7 @@ class DRYamlValidator
     return unless settings.training_abilities
 
     settings.training_abilities
-            .reject { |skill, _cooldown| ['PercMana', 'Perc', 'Perc Health', 'Astro', 'App', 'App Quick', 'App Careful', 'Tactics', 'Hunt', 'Pray', 'Scream', 'Khri Prowess', 'Stealth', 'Ambush Stun', 'Ambush Choke', 'Favor Orb', 'Charged Maneuver', 'Meraud', 'Recall', 'Barb Research Debilitation', 'Barb Research Augmentation', 'Barb Research Warding', 'Collect', 'Analyze', 'PrayerMat', 'Smite'].include?(skill) }
+            .reject { |skill, _cooldown| ['PercMana', 'Perc', 'Perc Health', 'Astro', 'App', 'App Quick', 'App Careful', 'Tactics', 'Hunt', 'Pray', 'Scream', 'Khri Prowess', 'Stealth', 'Ambush Stun', 'Ambush Choke', 'Favor Orb', 'Charged Maneuver', 'Meraud', 'Recall', 'Barb Research Debilitation', 'Barb Research Augmentation', 'Barb Research Warding', 'Collect', 'Analyze', 'PrayerMat', 'Smite', 'Locks'].include?(skill) }
```
rpherbig: ❓ Similar to the above question, do you want to add a new test that if `Locks` is in `training_abilities`, `pet_box_source` is also defined?

Sarvatt: I can follow up adding that after if this is a blocker.

rpherbig: 👍

combat-trainer.lic
dr-leustyin: Holding a summoned weapon is one of the things that broke the logic when I was first working on / testing this.  I'm not sure if there are other things that might also get in the way that might not be summoned weapons.

I think breaking the pick execution if there's something unexpected in your hands is a safe guard

rpherbig: :+1:

PR #2798, opened 2018-06-20
Restock supports custom hometowns for buying items.

jayRyan24:
Initial commit for adding custom hometowns to restock. If a hometown is found, it will then go to that hometown instead. I added two new money methods to handle things. Also fixed a bug in the base consumables for hib's items needing price and not cost.

### DIFF
--- a/restock.lic
+++ b/restock.lic
@@ -36,8 +36,7 @@ class Restock
       item['buy_num'] = buy_num
       items_to_restock.push(item)
     end
-    money_string = coin_needed.to_s + ' copper ' + town_currency(town)
-    get_money_from_specific_bank(money_string, town) if coin_needed > 0
+    get_money_from_specific_bank("#{coin_needed.to_s} copper #{town_currency(town)}", town) if coin_needed > 0
     items_to_restock.each do |item|
       item['buy_num'].times do
         buy_item(item['room'], item['name'])
--- a/restock.lic
+++ b/restock.lic
@@ -1,26 +1,35 @@
 #   Documentation: https://elanthipedia.play.net/Lich_script_repository#restock
-
 custom_require.call(%w[common common-items common-money])
-
 class Restock
   include DRC
   include DRCM
   include DRCI
-
   def initialize
     setup
-
     items = parse_restockable_items
+
+    custom_loc_items = items.select{ |k| k['hometown'] }
+
+    items -= custom_loc_items
+    restock_items(items)
+
+    unique_hometowns = custom_loc_items.map{ |k| "#{k['hometown']}" }.uniq
+    unique_hometowns.each do |hometown|
+      custom_list = custom_loc_items.select{ |k| k['hometown'].eql?(hometown) }
+      custom_loc_items -= custom_list
+      restock_items(custom_list)
+    end
+  end
+  def restock_items(item_list)
     items_to_restock = []
     coin_needed = 0
 
-    items.each do |item|
+    item_list.each do |item|
       remaining = if item['stackable']
                     count_stackable_item(item['name'])
                   else
                     count_nonstackable_item(item['name'])
                   end
-
       next unless remaining < item['quantity']
       num_needed = item['quantity'] - remaining
       buy_num = (num_needed / item['size'].to_f).ceil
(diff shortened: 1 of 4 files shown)


Sarvatt:
the common-money change should be broken out into it's own commit that goes in a few days before the rest to not break people who log in new characters with an existing one still logged in

jayRyan24:
That's a lot of work. If they break, they just need to announce it and then start over again.

jayRyan24:
I removed the common-money methods which are called and put them in a separate PR. They are located in #2801 .

restock.lic
```diff
@@ -1,58 +1,65 @@
...
       num_needed = item['quantity'] - remaining
       buy_num = (num_needed / item['size'].to_f).ceil
       coin_needed += buy_num * item['price']
       item['buy_num'] = buy_num
       items_to_restock.push(item)
     end
-
-    ensure_copper_on_hand(coin_needed, @settings) if coin_needed > 0
-
+    town = @settings.hometown
```
rpherbig: ❓ What would you think about having `town` be a second parameter to `restock_items`? Then line 14 would pass in `@settings.hometown` and line 20 would pass in `hometown`

restock.lic
```diff
@@ -1,58 +1,65 @@
...
     setup
-
     items = parse_restockable_items
+
+    custom_loc_items = items.select{ |k| k['hometown'] }
+
+    items -= custom_loc_items
+    restock_items(items)
+
+    unique_hometowns = custom_loc_items.map{ |k| "#{k['hometown']}" }.uniq
```
rpherbig: ❓ Why is there string interpolation here instead of mapping to `k['hometown']` directly?

restock.lic
```diff
@@ -1,58 +1,65 @@
...
+    unique_hometowns.each do |hometown|
```
rpherbig: 💡 You don't need `unique_hometowns` as a variable. Method chaining here could look like
```ruby
custom_loc_items.map{ |k| "#{k['hometown']}" }
.uniq
.each do ...
```

restock.lic
jayRyan24: Not sure, but I fixed it.

restock.lic
jayRyan24: Yup, I used it differently before this, so I had a use of it then. Thanks!

restock.lic
```diff
@@ -1,58 +1,65 @@
...
+      custom_list = custom_loc_items.select{ |k| k['hometown'].eql?(hometown) }
```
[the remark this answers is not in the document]
jayRyan24: ok, so it didn't do it in that order this time, so yeah, doesn't even matter

restock.lic
jayRyan24: Yup, good change. Thank you.

restock.lic
```diff
@@ -85,25 +90,21 @@ class Restock
...
   end
-
   def valid_item_data?(item_data)
-    return false unless item_data.key?('name')
-    return false unless item_data.key?('size')
-    return false unless item_data.key?('room')
-    return false unless item_data.key?('price')
-    return false unless item_data.key?('stackable')
-    return false unless item_data.key?('quantity')
+    return false unless item_data.key?('name')     &&\
```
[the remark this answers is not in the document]
jayRyan24: Sure thing.

jayRyan24:
Yup, that's the plan Stan. Thanks for the review!

jayRyan24:
Oh, tested all the changes. Continues to work.

restock.lic
```diff
@@ -37,10 +36,8 @@ class Restock
       item['buy_num'] = buy_num
       items_to_restock.push(item)
     end
-    town = @settings.hometown
-    town = items_to_restock.first['hometown'] if items_to_restock.first.key?('hometown')
-    coin_string = coin_needed.to_s + ' copper ' + town_currency(town)
-    get_money_from_specific_bank(coin_string, town) if coin_needed > 0
+    money_string = coin_needed.to_s + ' copper ' + town_currency(town)
```
rpherbig: 💡 This is a case where I prefer string interpolation, something like `"#{coin_needed.to_s} copper #{town_currency(town)}"`

rpherbig:
Looks good to me. Remind me to merge this at some point 😁

restock.lic
jayRyan24: Just so used to stringstream, lol. I gotcha.

jayRyan24:
I've added this to https://elanthipedia.play.net/mediawiki/index.php/Lich_script_repository

jayRyan24:
@rpherbig this can be pushed any time, it's been a few days.

PR #2799, opened 2018-06-21
Normalized some skills to use setting vs hard-coded value

dr-leustyin:


Sarvatt:
combat_training_abilities_target: 32 needs to be added to base.yaml

dr-leustyin:
It's already there: https://github.com/rpherbig/dr-scripts/blob/master/profiles/base.yaml#L73

PR #2800, opened 2018-06-21
 Added option to kill combat if certain skills drop too low

dr-leustyin:


hunting-buddy.lic
```diff
@@ -203,12 +211,15 @@ class HuntingBuddy
   def all_skills_at_cap?(stop_on_skills)
     stop_on_skills && stop_on_skills.flatten.all? { |skill| DRSkill.getxp(skill) >= 32 }
   end
+  def killon_skills_too_low?(kill_on_skills)
```
rpherbig: 💡 Maybe `kill_on_skills_too_low?`?

hunting-buddy.lic
```diff
@@ -230,6 +241,10 @@ class HuntingBuddy
         echo('***STATUS*** stopping due to skills')
```
rpherbig: 💡 Maybe `...due to capped skills`?

rpherbig: And/or you could include `stop_on_skills` in the output too

profiles/Leustyin-moths.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - weararmor
+  - release mof
+  # - buff main
+  - buff aeg
+  - buff aeg
+  after:
+  - release cyclic
+  stop_on:
+  - Staves
+  kill_on:
```
rpherbig: ❓ I'm not a fan of the name `kill_on` - I'm not sure it's clear enough what it does. Especially since it can coexist with `stop_on`, I can easily see users being unclear on how they differ. Have you thought of any other possible names for this?

Sarvatt: stop_when_low: maybe?

PR #2801, opened 2018-06-21
Added methods for supporting money from non-hometowns...

jayRyan24:
Added two methods. One to get the current of an input town and not just the hometown. Another is to get money from a specific town bank.

@Sarvatt 💋

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -129,4 +129,23 @@ module DRCM
     minimize_coins(keep_copper).each { |amount| withdraw_exact_amount?(amount, settings) }
     DRC.bput('check balance', 'your current balance is')
   end
+
+  def town_currency(town)
+    get_data('town')[town]['currency']
+  end
+
+  def get_money_from_specific_bank(amount_as_string, town)
+    DRCT.walk_to(get_data('town')[town]['deposit']['id'])
+    DRC.release_invisibility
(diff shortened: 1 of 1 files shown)


common-money.lic
```diff
@@ -129,4 +129,23 @@ module DRCM
     minimize_coins(keep_copper).each { |amount| withdraw_exact_amount?(amount, settings) }
     DRC.bput('check balance', 'your current balance is')
   end
+
+  def town_currency(town)
+    get_data('town')[town]['currency']
+  end
+
+  def get_money_from_specific_bank(amount_as_string, town)
```
rpherbig: ❓ If `get_money_from_specific_bank` is going to return a bool, could you add `?` to the end of the name to indicate it is a predicate?

jayRyan24: Sure thing, I pilfered from `get_money_from_bank` which also is a predicate method, do you want a ? there too? (Though it will break stuff for users.)

jayRyan24: @rpherbig

rpherbig: I would not change the method that is already being used

jayRyan24: ok

rpherbig:
Thank you for splitting this into two PRs @jayRyan24

PR #2804, opened 2018-06-21
Added items. Changed yardsticks

Glimmereyes:
Disabled Yardsticks as workorders since they require more volumes than store bought bronze ingots provide. (default material)
Added missing Blacksmithing items. Grouped items together.


Sarvatt:
In the future if you do this, could you make the changes in 2 separate commits on that branch to make it easier to review? Reviewing this diff is rough and seeing the 2 changes separated would help.

PR #2805, opened 2018-06-21
Move the performance during first-aid logic out of crossing-training …

Sarvatt:
…so it works running the script standalone.


Sarvatt:
scrap this, I need to figure out how to make it account for cleaning zills instead of pulling out the book without just addinig a pause i did locally.

PR #2806, opened 2018-06-21
Added an offensive spells threshold. 

jayRyan24:
Added the ability to stop training offensive spells at a defined threshold. Suggested by Binu.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -841,6 +841,9 @@ class SpellProcess
     @training_spells_max_threshold = settings.combat_spell_training_max_threshold
     echo("  @training_spells_max_threshold: #{@training_spells_max_threshold}") if $debug_mode_ct
 
+    @offensive_spells_max_threshold = settings.offensive_spell_training_max_threshold
+    echo("  @offensive_spells_max_threshold: #{@offensive_spells_max_threshold}") if $debug_mode_ct
+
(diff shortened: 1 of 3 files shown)


profiles/base.yaml
```diff
@@ -115,6 +115,7 @@ dump_junk: false
 dump_item_count: 9
 use_barb_combos: false
 whirlwind_trainables:
+offensive_spell_training_max_threshold:
```
dr-leustyin: I think this should default to 32, then line 1304 in combat-trainer could be simplified

combat-trainer.lic
```diff
@@ -1298,7 +1301,9 @@ class SpellProcess
   def check_offensive(game_state)
     return if game_state.casting
     return if game_state.npcs.empty?
+    return if @offensive_spells_max_threshold && game_state.npcs.length > @offensive_spells_max_threshold
```
dr-leustyin: game_state.npcs.length should be magic mindstates

profiles/base.yaml
jayRyan24: I just copied the other implementation. I agree but followed suit.

combat-trainer.lic
jayRyan24: I completely misread the other threshold, wtf.

PR #2808, opened 2018-06-22
cleric-quest.lic: Fix eluned harder, 5 minutes is not enough sometimes.

Sarvatt:
Just retry every minute until it works.

### DIFF
--- a/cleric-quests.lic
+++ b/cleric-quests.lic
@@ -138,11 +138,8 @@ class ClericQuests
     bput('meditate', 'You close your eyes and begin to meditate')
     waitfor('once again changing and reforming into your Human form')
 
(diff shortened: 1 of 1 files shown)


cleric-quests.lic
```diff
@@ -138,11 +138,8 @@ class ClericQuests
     bput('meditate', 'You close your eyes and begin to meditate')
     waitfor('once again changing and reforming into your Human form')
 
-    case bput('dive deeper water', 'You take a deep breath', 'you find that you cannot move yourself to go in')
-    when 'you find that you cannot move yourself to go in'
-      echo("Pausing for 5 minutes until you can complete another ritual because you did tamsine's too recently.")
```
rpherbig: ❓ I liked this `echo`. Why was it removed?

Sarvatt: readded it, sorry :)

PR #2818, opened 2018-06-23
missing a void moth room, found because of Binu's map

jayRyan24:


jayRyan24:
btw, I have the order which I think is best here, so 171 before 166.

PR #2823, opened 2018-06-24
workorders.lic: Allow setting outfitting_tools: in your yaml so they …

### LINKED ISSUE #2821: Workorders should support repairing outfitting tools that aren't beltworn
I noticed that `;workorders tailoring repair` will only repair knitting tools or anything under `outfitting_belt`...
We should add in a setting for `outfitting_tools` so these can be repaired:

```yaml
outfitting_tools:   
- knitting needles  
- sewing needles    
- awl               
- scissors          
- stamp             
- hide scraper      
- slickstone        
- yardstick         
```

Sarvatt:
…get repaired

Fixes #2821

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -306,6 +306,7 @@ forging_tools:
 - shovel
 knitting_tools:
 - knitting needle
+outfitting_tools:
 shaping_tools:
 - carving knife
 - shaper
--- a/workorders.lic
+++ b/workorders.lic
@@ -82,7 +82,7 @@ class WorkOrders
                      end
     when 'tailoring'
       @outfitting_room = @settings.outfitting_room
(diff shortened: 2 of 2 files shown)


jandersson:
<3

Sarvatt:
figured this would break the least amount of people, suddenly adding default outfitting_tools: would break everyone relying on current behavior.

rpherbig:
❓ Does this need an entry in `base-empty.yaml`?

jandersson:
Adding it to base-empty means the || will always return the empty array

You could do something like `tools = (settings.outfitting_tools + settings.knitting_tools).uniq`

Sarvatt:
Would you prefer populating the default tools for sewing or @jandersson's way?

PR #2830, opened 2018-06-24
Removed the left from always being in the execute_aiming_action

jayRyan24:
People thought it was annoying. It will now preload the left per the skill.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2306,15 +2306,15 @@ class AttackProcess
       weapon_name = game_state.wield_offhand_weapon(skill)
 
       actions = if skill.include?('Thrown')
-                  %w[lob lob lob]
+                  %w['lob left' 'lob left' 'lob left']
                 elsif skill.eql?('Brawling')
-                  %w[gouge gouge gouge]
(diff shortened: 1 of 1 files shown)


jayRyan24:
Don't merge yet, it's not working.

combat-trainer.lic
```diff
@@ -2306,15 +2306,15 @@ class AttackProcess
       weapon_name = game_state.wield_offhand_weapon(skill)
 
       actions = if skill.include?('Thrown')
-                  %w[lob lob lob]
+                  %w['lob left' 'lob left' 'lob left']
```
rpherbig: 💡 The `%w` syntax doesn't work for multiple-word-phrases. I suspect you want `['lob left', 'lob left', 'lob left']`

rpherbig:
❓ It looks like the only thing that functionally changes is the Tactics case, which would have been bob left/weave left/circle left but are now bob/weave/circle. Is that right?

Sarvatt:
that looks right to me.

jayRyan24:
Yes, the only thing really affected by this is the `tactics + left`. I tried single quotes and it failed, though I might have had another change which made it do so.

PR #2832, opened 2018-06-25
Added the ability for spells to have the attribute to use_stealth.

jayRyan24:
I defaulted Vivi to have this be set to true.


jayRyan24:
I could use some necros testing this. I will test it tonight with Ssarek.

I could also use some feedback on whether I should keep it as is or change to check if true.

`game_state.hide_on_cast = true if data['use_stealth']` vs `game_state.hide_on_cast = true if data['use_stealth'] == true` 

If it checks if true, it will allow a person to turn it off without commenting out the setting or removing it completely. I think I would need `== true` and not just check if it's there. Maybe they're the same thing, I get confused with Ruby some times. It could be functioning exactly how I think it should. 

Also doublechecking that we're not checking `use_stealth_attacks` is set anymore. I kind of think that's what the setting in the spell is now for so there's no need, and the method used before `use_stealth_attacks?` checked the `use_stealth?` based on current exp so I think it's basically equivalent. 

Input needed.

jayRyan24:
@rpherbig

combat-trainer.lic
```diff
@@ -1200,6 +1200,8 @@ class SpellProcess
       @training_cyclic_timer = @training_cast_timer
     end
 
+    game_state.hide_on_cast = true if data['use_stealth']
```
rpherbig: 💡 Could this just as easily be `game_state.hide_on_cast = data['use_stealth']`?

rpherbig:
Looks good to me. Do you feel like it's been tested enough?

combat-trainer.lic
jayRyan24: Yeah, I'll change that later tonight.

jayRyan24:
Sure, let me change it tonight and then tomorrow I'll get back to you. Thanks.

jayRyan24:
Working for my Empath fine. I will test over night and also am trying to get necros involved.

jayRyan24:
Working for a couple days without issue. Ready for merge.

jayRyan24:
@rpherbig

PR #2836, opened 2018-06-25
safe-room.lic: Update devour healing logic for the spell changes

Sarvatt:
wait for the buff to drop before checking if you should recast it since it is a pulsing heal spell now.

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -102,9 +102,7 @@ class SafeRoom
       break if result =~ /^What were you/
       cast_spell(settings.necromancer_healing['Devour'], settings)
       devours -= 1 unless right_hand
-      while DRSpells.active_spells['Devour']
-        pause 5
-      end
+      pause 5 while DRSpells.active_spells['Devour']
     end
 
     bput("put my material in my #{settings.thanatology['harvest_container']}", 'You put', 'What were') if right_hand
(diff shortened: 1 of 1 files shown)


safe-room.lic
```diff
@@ -102,6 +102,9 @@ class SafeRoom
       break if result =~ /^What were you/
       cast_spell(settings.necromancer_healing['Devour'], settings)
       devours -= 1 unless right_hand
+      while DRSpells.active_spells['Devour']
+        pause 5
+      end
     end
```
rcuhljr: Do we prevent a race condition with lag by using `pause 5 until !DRSpells.active_spells['Devour']`?

Sarvatt: that is definitely cleaner :)

rcuhljr:
How long does devo stay up if there's not enough for it to fully expend itself? Would they get stuck there until it expires from time?

Sarvatt:
that's a really good question. It takes 20-30 seconds to fully wear off if you do need healing from what I'm told. This does need fixing, and I'm thinking just a dumb 30 second pause between checks is better.

jayRyan24:
Could be an issue where you don't need 30, but 15. Also, you are casting the spell and not checking that it isn't already active. Possibly this is a better option, what do you think?
```
    while need_healing? && devours > 0
      while mana < 40
        echo('Waiting on mana...')
        pause 10
      end

      result = bput("get material in my #{settings.thanatology['harvest_container']}", 'You get', 'You are already', 'What were you') unless right_hand
      break if result =~ /^What were you/
      unless DRSpells.active_spells['Devour']
        cast_spell(settings.necromancer_healing['Devour'], settings) 
        devours -= 1 unless right_hand
      end
      pause 5
    end
```

PR #2841, opened 2018-06-26
workorders.lic: restock straight pins when sewing, they are consumable.

Sarvatt:
Tested many many times to get it right :)


Sarvatt:
Me not realizing sew already did it, someone reported getting stuck from not having pins now that we turned on some cloth workorders. It looks like sew never buys the initial pins if you don't have any was the problem

rcuhljr:
Ah yeah I could see that. We could probably add in pin purchasing if you have none to `sew.lic`.

PR #2844, opened 2018-06-26
Added missing Knitting recipes

Glimmereyes:
Added Knitted Slippers and Knitted Towel.


Sarvatt:
This change looks good to go, verified it.

PR #2845, opened 2018-06-26
Make sure that aiming_trainables are also weapons in the current hunt

dr-leustyin:


rpherbig:
💡 This feels like a good test to add to `validate.lic`

PR #2851, opened 2018-06-27
Add support for light crossbows to aiming trainables

jayRyan24:
Added in a new setting for letting the user say whether they're crossbow is light or not. There are too many different types of crossbows to keep a list of all possible versions. To be safe we default to false, but the user can adjust it. The `determine_aiming_skill` now checks that the new **game_state** setting is true, if so, it allows the use of attacking with the offhand instead of just thrown, tactics, and brawling. One of the benefits of lighter crossbows.


dr-leustyin:
❤️

PR #2853, opened 2018-06-27
Add craft.lic, crafting from crossing-training broken out on its own.

Sarvatt:
I would like to gut crafting from crossing-training with this eventually, but best to give it some time to settle and be sure it works in all scenarios, even ones I didn't test. This is nice for just crafting 1 item in a before instead of doing a workorder.

### DIFF
--- a/craft.lic
+++ b/craft.lic
@@ -1,0 +1,301 @@
+custom_require.call(%w[drinfomon equipmanager common common-crafting common-items common-money common-travel])
+
+class Craft
+  include DRC
+  include DRCC
+  include DRCI
+  include DRCM
+  include DRCT
+  
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'craft', options: %w[forging outfitting engineering alchemy], description: 'What craft to train' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
+    
+    @settings = get_settings
+    @hometown = @settings.hometown
+    @training_room = (@settings.training_rooms || [@settings.safe_room]).sample
+    @settings.storage_containers.each { |container| fput("open my #{container}") }
+    @equipment_manager = EquipmentManager.new(@settings)
+    @craft_max_mindstate = @settings.craft_max_mindstate
+    @yarn_quantity = @settings.yarn_quantity
+
+    if args.craft=="engineering"
+      exit if DRSkill.getxp("Engineering") > @craft_max_mindstate
+      train_engineering
+    end
+
+    if args.craft=="outfitting"
+      exit if DRSkill.getxp("Outfitting") > @craft_max_mindstate
+      train_outfitting
+    end
+    
+    if args.craft=="forging"
+      exit if DRSkill.getxp("Forging") > @craft_max_mindstate
+      train_forging
+    end
+
+    if args.craft=="alchemy"
+      exit if DRSkill.getxp("Alchemy") > @max_mindstate
+      train_alchemy
+    end
+
+  end
+
+  def train_outfitting
+    if @settings.train_workorders.include?('Tailoring')
+      return unless money_for_training?(5000, 'Outfitting')
+      wait_for_script_to_complete('workorders', ['Tailoring'])
+      wait_for_script_to_complete('sell-loot')
+      walk_to(@training_room)
+      return
+    end
+
+    rank = DRSkill.getrank('Outfitting')
+    if rank <= 25 # Tier 1  Extremely Easy
+      sew(5, 'knitted socks', 'socks')
+    elsif rank <= 50 # Tier 2 Very Easy
+      sew(5, 'knitted mittens', 'mittens')
+    elsif rank <= 100 # Tier 3  Easy
+      sew(5, 'knitted hat', 'hat')
+    elsif rank <= 175 # Tier 4  Simple
+      sew(5, 'knitted gloves', 'gloves')
+    elsif rank <= 300 # Tier 5  Basic
+      sew(5, 'knitted hose', 'hose')
+    elsif rank <= 425 # Tier 6  Somewhat Challenging
+      sew(5, 'knitted cloak', 'cloak')
+    elsif rank <= 650 # Tier 7  Challenging
+      sew(5, 'knitted blanket', 'blanket')
+    else # Tier 12  Extremely Difficult
+      echo('*** NOT YET IMPLEMENTED ***')
+    end
+  end
+
+  def train_engineering
+    if @settings.train_workorders.include?('Shaping')
+      return unless money_for_training?(5000, 'Engineering')
+      wait_for_script_to_complete('workorders', ['Shaping'])
+      wait_for_script_to_complete('sell-loot')
+      walk_to(@training_room)
+      return
+    elsif @settings.train_workorders.include?('Carving')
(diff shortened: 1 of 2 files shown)


craft.lic
```diff
@@ -0,0 +1,291 @@
+custom_require.call(%w[drinfomon equipmanager common common-crafting common-items common-money common-travel])
```
rpherbig: 💡 Please add a header with a link to elanthipedia

craft.lic
```diff
@@ -0,0 +1,291 @@
...
+      train_alchemy
+    end
+  end
+
+  def train_outfitting
+    if @settings.train_workorders.include?('Tailoring')
+      return unless money_for_training?(5000, 'Outfitting')
+      wait_for_script_to_complete('workorders', ['Tailoring'])
+      wait_for_script_to_complete('sell-loot')
+      walk_to(@training_room)
```
rpherbig: ❓ This feels more like the responsibility of `crossing-training`. What do you think?

craft.lic
```diff
@@ -0,0 +1,291 @@
...
+      shape(7, 'wood comb', 'comb')
+    elsif rank <= 1400 # Tier 11 - Very Difficult
+      shape(7, 'wood haircomb', 'haircomb')
+    else # Tier 12  Extremely Difficult
+      echo('*** NOT YET IMPLEMENTED ***')
+    end
+  end
+
+  def train_forging
+    eligible = @settings.train_workorders & %w[Blacksmithing Weaponsmithing]
```
rpherbig: 💡 Should engineering (shaping/carving) be refactored to use this pattern?

craft.lic
```diff
@@ -0,0 +1,291 @@
...
+      smith('a metal ankle band')
+    elsif rank <= 300 # Tier 5 - Basic
+      smith('a metal lockpick ring')
+    elsif rank <= 425 # Tier 6 - Somewhat Challenging
+      smith('a metal armband')
+    elsif rank <= 550 # Tier 7 - Challenging
+      smith('some metal clippers')
+    elsif rank <= 700 # Tier 8 - Complicated
+      smith('some squat knitting needles')
+      # Journeyman book: triangular wire sieve, bent <metal> scissors, knobby sewing needles, squat knitting needles, narrow <metal> awl
```
rpherbig: 💡 These comments (listing possible items to craft) can be removed

craft.lic
```diff
@@ -0,0 +1,291 @@
...
+      return unless money_for_training?(5000, 'Alchemy')
+      wait_for_script_to_complete('workorders', ['Remedies'])
+      wait_for_script_to_complete('sell-loot')
+      walk_to(@training_room)
+      return
+    end
+
+    beep
+    echo('*** NOT YET IMPLEMENTED ***')
+    @settings.crossing_training.delete('Alchemy')
```
rpherbig: ❓ This seems like something that a crafting script shouldn't be doing. In addition, it only modifies `@settings` for _this script_, so `crossing-training` would be unaffected.

And line 182 (maybe other lines too).

craft.lic
```diff
@@ -0,0 +1,291 @@
...
+  end
+
+  def buy_wood
+    crafting_data = get_data('crafting')
+    return unless money_for_training?(1000, 'Engineering')
+    order_item(crafting_data['shaping'][@hometown]['stock-room'], 10)
+    bput('stow my lumber', 'You put your', 'Stow what')
+  end
+
+  def buy_yarn(skipcoin = false)
```
rpherbig: ❓ I don't see any code paths in which `buy_yarn` is called with an argument. Why is `skipcoin` here?

craft.lic
```diff
@@ -0,0 +1,291 @@
...
+
+    bput('stow my yarn', 'You put', 'Stow what')
+  end
+
+  def shape(chapter, unique_name, item)
+    check_wood
+
+    find_shaping_room(@hometown, @training_room)
+    check_listening
+    check_teaching
```
rpherbig: ❓ What happens if you listen and try to teach?

profiles/base.yaml
```diff
@@ -333,7 +333,8 @@ alchemy_room:
 use_own_ingot_type:
 deed_own_ingot: false
 crafting_training_spells:
-
+craft_max_mindstate: 31
+yarn_quantity: 100
```
rpherbig: ❓ Why is this a setting? i.e. when would a user want to change this from the default?

craft.lic
Sarvatt: line 231, so if you only have 3 gold you dont run to the bank after each purchase

craft.lic
Sarvatt: true, also the getting money should be removed from crossing training since workorders handles it directly in a customizable way.

Sarvatt: it's more here for the future goal of just completely gutting crafting out of crossing-training

craft.lic
Sarvatt: good call. also needed in present day crossing-training

craft.lic
Sarvatt: I brought the workorder support in strictly because alchemy only works with workorders, I'll remove alchemy from this completely and we'll handle it on the crossing-training side

profiles/base.yaml
Sarvatt: which one? Carrying more yarn means less trips to the shop and it's light. but some people might run workorders also which trashes yarn so there is no point buying extra. the max mindstate is so I can have it not craft if i'm already at a high mindstate for some reason (having run workorders in crossing-training beforehand) I only want 12 mindstates, happy to remove it though.

craft.lic
Sarvatt: done

craft.lic
Sarvatt: deleted the workorders handling from this completely based on later comments, it will stay in crossing-training

Sarvatt: I have a commit ready to make this change after the money change to crossing-training goes in.

craft.lic
Sarvatt: I'm assuming its assumed someone wouldn't have both classes_to_teach and listen: true in their yaml at the same time, this was lifted straight out of current crossing-training.

rpherbig: Ok

craft.lic
rpherbig: Sure, line 231 (or, rather, now line 190). But that passes along `skipcoin` to itself recursively. To clarify my question, I don't see anywhere that `skipcoin` is changed from its default of `false`. Should line 190 be `true`?

profiles/base.yaml
rpherbig: Ok

rpherbig:
Looks good aside from the one ❓ on `skipcoin`.

A 💡 for handling alchemy: what do you think about keeping it in the arg list and then telling the user its not supported? That cuts down on things like 'did I misspell alchemy?' or 'is this a bug that it's not supported?'

craft.lic
Sarvatt: so I wasn't sure about this, and from testing I saw passing anything but false there made it true and skip withdrawing. buy_yarn(true) does work too, i'll change it to that

Sarvatt:
good idea, doing that now

craft.lic
Sarvatt: fixed

PR #2856, opened 2018-06-27
Update forge.lic - add call to buff

BinuDR:
Add call to buff script to add in support for using khri while crafting.


BinuDR:
And also any other buffs someone wants to cast before forging.

PR #2857, opened 2018-06-28
crossing-training.lic: Remove useless money checks, workorders alread…

Sarvatt:
…y does it.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -778,7 +778,6 @@ class CrossingTraining
 
   def train_outfitting
     if @settings.train_workorders.include?('Tailoring')
-      return unless money_for_training?(5000, 'Outfitting')
       wait_for_script_to_complete('workorders', ['Tailoring'])
       wait_for_script_to_complete('sell-loot')
       walk_to(@training_room)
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -778,7 +778,6 @@ class CrossingTraining
 
   def train_outfitting
     if @settings.train_workorders.include?('Tailoring')
-      return unless money_for_training?(5000, 'Outfitting')
```
rpherbig: ❓ There is a functional difference here. Specifically, if `money_for_training?` returns false, it short-circuits the sell-loot execution and travel. If we rely on workorders to handle the money check, we're running those two. Is that ok?

Sarvatt: that should be ok, money_for_training? does a ensure_copper_on_hand which goes and withdraws the money, which could be a different amount than workorders actually needs. Workorders gets a workorder before doing its own money withdraws so if your workorder_cash_on_hand: is not 5 gold we're adding an extra step going to the bank here. money_for_training? would only return false if your bank account doesn't have 5 gold in it. I'm happy to not include this change though, it's does seem safer to just keep it after thinking it through.

PR #2859, opened 2018-06-29
Update base.yaml - skip exchange setting added

BinuDR:
Added:
sell_loot_skip_exchange: false


Sarvatt:
can you add them in the same pull request since it will break people if they arent pulled at the same time?

BinuDR:
I don't know how to do that, and it shouldn't break anyone because it'll be a nil value if it's not in the yaml or base and should go right on exchanging money.

PR #2862, opened 2018-06-29
Fix up some of the CI errors.

Sarvatt:
Please don't pull yet, testing the CI.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -849,5 +849,3 @@ force_healer_town:
 shard_thief_password:
 forge_container:
 spell_cycle:
-
-restock_shop:
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -334,7 +334,8 @@ alchemy_room:
 use_own_ingot_type:
 deed_own_ingot: false
 crafting_training_spells:
-
+craft_max_mindstate: 31
+yarn_quantity: 100
 
 
 # Workorder settings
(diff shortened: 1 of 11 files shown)


Sarvatt:
Ready for review now.

Sarvatt:
Both broken yamls were my own characters so it's ok to change them :)

profiles/base.yaml
```diff
@@ -852,5 +852,4 @@ shard_thief_password:
 forge_container:
 spell_cycle:
 
-restock_shop:
-
+restock_shop: {}
```
rpherbig: ❓ This is already defined in `base-empty` as `{}`. Why is it defined that way here too? I expected it to remain `restock_shop:`...

Sarvatt:
https://github.com/rpherbig/dr-scripts/blob/master/test/test_validate.rb#L22 only loads base.yaml and doesn't know about base-empty, I don't know how to fix that.

rpherbig:
I suspect something like this would work (refactoring is left as an exercise to the reader)

```ruby
$test_settings = OpenStruct.new(YAML.load_file('profiles/base.yaml').merge(YAML.load_file('profiles/base-empty.yaml')).merge(settings))`
```

or maybe

```ruby
$test_settings = OpenStruct.new(YAML.load_file('profiles/base.yaml').merge(settings))`
YAML.load_file('profiles/base.yaml').empty_values.each { |name, value| $test_settings[name] ||= value }
```

Sarvatt:
I just removed the test_validate.rb changes, figure I can submit at least part of the fixes.

PR #2868, opened 2018-06-30
test-validate: load base empty

jandersson:
Loads the base empty file in the same manner as dependency.


jandersson:
This fixes the `nil.empty?` error. At the moment another, separate test is failing.

PR #2870, opened 2018-06-30
Go2: Use dr specific tags for go2 targets

jandersson:
`;go2 targets` a list of locations by town. DR has its own set
of tags disjoint from gemstones.


jandersson:
One moment, id like move the ternary operator out of the for loop

PR #2879, opened 2018-07-01
Tailoring - chapter 3 workorders

yuoree:
Turned on all chapter 3 workorders for higher level outfitting.

Still a gap between difficulty 08 (small rucksack) and 10 (rucksack). Likely need to pick a difficulty 09 or two from another chapter at some point.

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -2005,55 +2005,55 @@ crafting_recipes:
   noun: blanket
   volume: 9
   type: tailoring
-  work_order: false
+  work_order: true
   chapter: 3
 - name: a small cloth rucksack
   noun: rucksack
   volume: 7
   type: tailoring
-  work_order: false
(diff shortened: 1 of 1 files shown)


rpherbig:
@yuoree were (some) of these tested by running `workorders.lic`?

jandersson:
Ive tested the rucksack (by chance i made the same change this morning)

yuoree:
I've tested the ones I can on a couple chars, in shard. I've also done cloth armor stuff on my own version, but after talking to Sarvatt yesterday sticking to chapter 3 for now, since less parts.

Sarvatt:
this looks good to me, it's not turning on lower level chapter 3 things so it's only addiing items for people to craft not changing behavior for people still knitting.

Sarvatt:
chapter 2 has a few difficulty 9 items we could turn on

PR #2886, opened 2018-07-01
steal.lic: Add data builder mode that removes redundant actions if yo…

Sarvatt:
…u are just updating the stealing database.

### DIFF
--- a/steal.lic
+++ b/steal.lic
@@ -43,8 +43,8 @@ class Steal
     arg_definitions = [
       [],
       [
-        { name: 'item', regex: (/\w+|\w+\s\w+/), variable: true, description: 'item to steal' },
-        { name: 'container', regex: (/.+/i), variable: true, description: 'Where the item is to steal e.g in catalog, on counter' }
+        { name: 'item', regex: /\w+|\w+\s\w+/, variable: true, description: 'item to steal' },
+        { name: 'container', regex: /.+/i, variable: true, description: 'Where the item is to steal e.g in catalog, on counter' }
       ]
     ]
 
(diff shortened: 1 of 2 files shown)


steal.lic
```diff
@@ -35,7 +35,7 @@ class Steal
     @slow_bin_speed = settings.slow_bin_speed
 
     @difficulty_keys = %w[exceptionally very rather acceptable poorly vpoorly nearly trivial]
-
+    @data_builder_mode = settings.stealing_data_builder_mode
```
jandersson: If this is a new setting it should be added to base.yaml

jandersson: that was fast

steal.lic
```diff
@@ -49,11 +49,11 @@ class Steal
     ]
 
     args = parse_args(arg_definitions)
-
+    @data_builder_mode_toggle = false
```
jandersson: When is this set to true?

jandersson: Is it the same as `@data_builder_mode`?

jandersson: disregard

steal.lic
```diff
@@ -148,9 +148,12 @@ class Steal
   end
 
   def build_target_data(item, container)
+    @data_builder_mode_toggle = true if @data_builder_mode
```
jandersson: I think its confusing to have this buried in a function - is there a reason it cant be in setup?

Sarvatt:
thanks for the feedback, I'm trying to push this for Byd and he's updated it with enough changes that I should close this until I get them all integrated.

PR #2893, opened 2018-07-04
Add nexus script

jandersson:
This script will walk to a nexus point and perform the given sacrifice. It stores
the last time you sacrificed to pave the way for some future additions such as
keeping track of the sacrifice timer, since that is unknown.

The nexus alias has been removed from setupaliases as its been consumed by `;nexus list` which does the same thing

### DIFF
--- a/nexus.lic
+++ b/nexus.lic
@@ -1,0 +1,51 @@
+=begin
+  Documentation: 
+=end
+custom_require.call(%w[common common-travel])
+class Nexus
+  include DRC
+  include DRCT
+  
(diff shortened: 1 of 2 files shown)


jandersson:
We could add this into training manager, run a sacrifice every few hours or so (opt in). The timer is still unknown, but it seems to be less than 24 hours. But it may be dependent on the sacrifice, lots of unknowns.

jandersson:
Bugs!

jandersson:
false alarm

PR #2899, opened 2018-07-06
crossing-trainer.lic: Fix extra end left over

Sarvatt:


Sarvatt:
@rpherbig

Sarvatt:
tested working for 30 minutes. no immediate error running it like before.

PR #2903, opened 2018-07-06
common-healing.lic: Lower the check_health reget to 50, 300 is way to…

Sarvatt:
…o much.

It takes 6 lines at most, 44 extra lines in the reget array every time its called is maybe a bit excessive still but better than 294 :)


Sarvatt:
The person who implemented this at 300 lines thinks 50 is even too much, it's still better than 300.

PR #2907, opened 2018-07-08
combat-trainer.lic: Add missing end after the last update.

Sarvatt:


Sarvatt:
@rpherbig this broke combat-trainer usage, needs to be pulled asap if possible

PR #2909, opened 2018-07-08
Fixed to check properly. 

jayRyan24:
Return is true if nil (not summoning), and false (if summoned).

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2009,8 +2009,7 @@ class TrainerProcess
 
   def pickbox(game_state)
     return unless @pet_box_source && @pet_count > 0
-    return if game_state.loaded || game_state.offhand?
-    return unless game_state.weapon_is_summoned?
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2009,7 +2009,8 @@ class TrainerProcess
 
   def pickbox(game_state)
     return unless @pet_box_source && @pet_count > 0
-    return if game_state.loaded || game_state.weapon_is_summoned? || game_state.offhand?
+    return if game_state.loaded || game_state.offhand?
+    return unless game_state.weapon_is_summoned?
```
rpherbig: ❓ Does this mean 'return if weapon is not summoned'? That seems backwards. Or is the logic for `weapon_is_summoned?` inverted?

jayRyan24: Yeah, this is weird. I wanted to use the `summoned_info` method so I had it all weird. Put a !nil? return for the method, so it can move back up to `if summoned`.

PR #2917, opened 2018-07-11
Convert pouch selling for trading to a script

jandersson:
Essentially a copy and paste of the logic from crossing-training to a separate script.

This also updates the use of speculate finesse to a modern use of
trader magic spell finesse via a waggle set (the scripts name).

Another note is that this has potential to expand into running accept-sell multiple times as a counterpart to the new give pouches script.

### DIFF
--- a/sell-pouches.lic
+++ b/sell-pouches.lic
@@ -1,0 +1,43 @@
+custom_require.call(%w[common common-items common-travel])
+
+class PouchSeller
+  include DRC
+  include DRCI
+  include DRCT
+
+  def initialize
+    settings = get_settings
+    pouch_container = settings.sale_pouches_container
+    hometown = settings.hometown
+
+    exit unless have_pouch?(pouch_container)
+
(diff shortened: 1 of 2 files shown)


sell-pouches.lic
```diff
@@ -0,0 +1,43 @@
+custom_require.call(%w[common common-items common-travel])
```
rpherbig: 💡 Elanthipedia link

sell-pouches.lic
```diff
@@ -0,0 +1,43 @@
...
+
+    bput("get pouch from my #{container}", 'You get a', 'What were you') == 'You get a'  
+  end
+
+  def sell_pouch(shop_id)
+    return unless shop_id
+
+    walk_to(shop_id)
+    wait_for_script_to_complete('buff', [Script.current.name])
+    fput('sell pouch')
```
rpherbig: 💡 Could these two lines use `bput`?

jandersson:
Thanks for the feedback - I'll address those items in a new pull request

PR #2922, opened 2018-07-12
performance.lic: Don't clean zills during first-aid.

Sarvatt:
Not sure of how to get around this and people are hitting errors. I removed the script pausing also because it was only implemented for first-aid. The other users of this script (study-art and athletics) are fine wwithout pausing.

### DIFF
--- a/first-aid.lic
+++ b/first-aid.lic
@@ -31,7 +31,7 @@ class FirstAid
 
   def compendium_charts
     pause 2 # to give performance time to complete before_dying if stopped from the previous script
-    start_script('performance') unless Script.running?('performance')
+    start_script('performance', ['noclean']) unless Script.running?('performance')
     pause 4 # to catch the dirty zill message from performance.lic
     bput('get my compendium', 'You get', 'You are already holding')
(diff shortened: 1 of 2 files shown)


rpherbig:
Have you considered a "don't clean" art for performance, ie something that would indicate hands are full (first aid) vs empty (study art)?

Sarvatt:
Yep that was my first thought but nothing else would use it so I went the easy way. Updated to add an arg and use it, and i just tested it.

```--- Lich: first-aid active.
>
--- Lich: performance active.
[performance]>play psalm
>
Your zills's dirtiness may affect your performance.
You begin a quiet psalm on your asini zills with only the slightest hint of difficulty.
[first-aid]>get my cookbook
You get a sturdy wine-red cookbook titled "Gastronomy" along the spine from inside your multi-strapped carryall.
[first-aid]>open my cookbook
That is already open!
[first-aid]>turn my cookbook to Gryphon
You turn to the section on Gryphon physiology. 
[first-aid]>study my cookbook
You continue studying the Gryphon chart, gradually absorbing more of the knowledge contained within.
Roundtime: 20 seconds.
>

PR #2926, opened 2018-07-14
Updated bput to save time

dr-leustyin:


common.lic
```diff
@@ -66,7 +66,7 @@ module DRC
     clear
     put message
     timer = Time.now
-    while (response = get?) || (Time.now - timer < 15)
+    while (response = get?) || (Time.now - timer < 5)
```
rpherbig: ❌ It looks like a unit test needs to be updated too

dr-leustyin: Test updated.  Also cleaned up the method to be a bit less spammy in practice

Sarvatt:
wrong test, i think you want     ```assert_in_delta 5, @test.thread_variable_get('runtime'), 1.0``` in def test_bput_returns_error_code_after_timeout

Sarvatt:
oh there were 2 failing before, that should fix the one still failing though

Sarvatt:
changing ```assert_in_delta 15, @test.thread_variable_get('runtime'), 1.0``` to ```assert_in_delta 5, @test.thread_variable_get('runtime'), 1.0```

PR #2929, opened 2018-07-14
dependency.lic: Switch everyone over to the lich fork.

### LINKED ISSUE #2000: Use lich fork by default

Sarvatt:
Closes #2000

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -178,7 +178,7 @@ class ScriptManager
     @thievery_queue = []
     @shop_update_queue = []
 
-    @enable_fork = LICH_VERSION =~ /f/
+    @enable_fork = true
 
     @versions = nil
   end
@@ -246,13 +246,8 @@ class ScriptManager
       @remove_autos.each { |script| Settings['autostart'].delete(script) }
       @remove_autos = []
     end
-    if @disable_fork
-      stop_download_lich
-      @disable_fork = nil
-    elsif @enable_fork
-      download_lich
-      @enable_fork = nil
-    end
+    download_lich if @enable_fork
+    @enable_fork = nil
 
     submit_thieving_update(*@thievery_queue.pop) unless @thievery_queue.empty?
     submit_shop_update(@shop_update_queue.pop) unless @shop_update_queue.empty?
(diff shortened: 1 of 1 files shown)


Sarvatt:
Spent a few hours testing different scenarios with this such as a new lich install, install the new dependency here which updates to the fork, go to dev mode (so dependency doesn't get overwritten), replace lich.rbw with the old one, restart again to see if it switches you to the fork automatically. I've also disabled the fork on another instance of lich and logged in in dev mode with this dependency and it updatted me to the fork automatically.

dependency.lic
```diff
@@ -1170,19 +1169,12 @@ def toggle_developer_mode
...
 
 def use_lich_main
-  echo 'Now switching onto the lich mainline...'
-  start_script('repository', ['set-lich-updatable'])
-  pause
-  start_script('repository', ['download-lich'])
-  pause
-  pause 1 while Script.running?('repository')
-  echo('New version of lich installed, please restart Lich.')
+  echo 'Deprecated setting, only the fork is supported...'
```
rpherbig: ❓ This removes the ability to switch off of the fork?

Sarvatt: yeah, do we want to support that? As it is here it will just update to the fork again the next restart, I thought that was what we wanted?

rpherbig: My thought was that we'd make the fork the default and they have to manually move back to the base lich.

I haven't given it a lot of thought and don't have a strong preference right now.

jandersson:
Are you sure people not on the fork cant use dependency?

Sarvatt:
only people using ruby > 2.3, aka all linux users pretty much

Sarvatt:
the problem is $MODERN_LICH is false and $MODERN_RUBY is true so it's trying to use the old trust stuff in bootstrap.lic

Sarvatt:
but 4.6.49 lich you get not using the fork removed trust for newer ruby

PR #2930, opened 2018-07-14
Alleviate server lag issues from waitcastrt? (arcana only)

13and3:


rpherbig:
@13and3 @jandersson @Sarvatt any idea why the unit test is failing?

jandersson:
waitcast_rt_and_pause is not in the test harness is my guess.

jayRyan24:
Can we close this?

PR #2935, opened 2018-07-15
outdoorsmanship.lic: Add magic training.

Sarvatt:


outdoorsmanship.lic
```diff
@@ -38,16 +41,49 @@ class Outdoorsmanship
...
+    crafting_prepare_spell(data, @settings)
+    @preparing = true
+  end
+
+  def magic_cleanup
+    return unless @training_spells
+
+    bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
+    bput('release mana', 'You release all', "You aren't harnessing any mana")
+    Flags.delete('outdoors-magic-ready')
```
rpherbig: 💡 `before_dying`?

outdoorsmanship.lic
```diff
@@ -17,19 +18,21 @@ class Outdoorsmanship
...
     @forage_item = if args.collect_item
                      args.collect_item
                    else
-                     settings.forage_item
+                     @settings.forage_item
                    end
 
     start_exp = DRSkill.getxp('Outdoorsmanship')
     @end_exp = [start_exp + @targetxp, 29].min
+    Flags.add('outdoors-magic-ready', 'You feel fully prepared to cast your spell.')
```
rpherbig: 💡 We should probably establish some standard like "every script has a setup function which checks args, adds flags, get settings, ..." and "every script has a before_dying if it needs to delete flags, ..."

Sarvatt: very much agreed going forward, i submitted this before doing the before_dying thing.

Sarvatt: it makes sense to always delete the flags if the script dies even if it dies from being killed instead of just a graceful exit.

Sarvatt: added this to the review document in https://github.com/rpherbig/dr-scripts/wiki/Things-to-look-for-in-code-review

PR #2938, opened 2018-07-15
validate.lic: Add validation that dance_skill is correct

Sarvatt:
This came up on lnet yesterday.

[validate: ERROR:< Dance skill 'greatsword' is not valid. Valid skills are  '["Offhand Weapon", "Brawling", "Polearms", "Large Blunt", "Twohanded Blunt", "Staves", "Small Blunt", "Small Edged", "Large Edged", "Twohanded Edged"]'  >]

### DIFF
--- a/validate.lic
+++ b/validate.lic
@@ -88,6 +88,13 @@ class DRYamlValidator
             .each { |skill_name| error("Invalid weapon_training: skill name '#{skill_name}'. Valid skills are '#{@valid_weapon_skills}'") }
   end
 
+  def assert_that_dance_skill_is_skill(settings)
+    return unless settings.dance_skill
+    return if settings.dance_skill.include?('#{@valid_melee_skill}')
+
+    error("Dance skill '#{settings.dance_skill}' is not valid. Valid skills are  '#{@valid_melee_skills}'")
+  end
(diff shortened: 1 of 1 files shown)


validate.lic
```diff
@@ -88,6 +88,12 @@ class DRYamlValidator
             .each { |skill_name| error("Invalid weapon_training: skill name '#{skill_name}'. Valid skills are '#{@valid_weapon_skills}'") }
   end
 
+  def assert_that_dance_skill_is_skill(settings)
```
rpherbig: 💡 What do you think about making this name a bit more specific: `...dance_skill_is_melee_skill`?

validate.lic
```diff
@@ -88,6 +88,12 @@ class DRYamlValidator
...
+    return unless settings.dance_skill
+
+    error("dance_skill: skill name '#{settings.dance_skill}' is not valid. Valid skills are '#{@valid_melee_skills}'") unless @valid_melee_skills.include?(settings.dance_skill)
```
rpherbig: 💡 I suggest moving `unless @valid_melee_skills.include?(settings.dance_skill)` into a guard clause

PR #2939, opened 2018-07-15
validate.lic: Add 2 more validators for correct currency notation in yamls.

Sarvatt:


validate.lic
```diff
@@ -609,6 +609,20 @@ class DRYamlValidator
     error("The following restock shop items match no recipe data: #{unknown_item_recipes.map { |item| item['recipe'] }}")
   end
 
+  def assert_that_workorder_cash_on_hand_is_correct(settings)
+    return unless settings.workorder_cash_on_hand
+    return if settings.workorder_cash_on_hand.is_a?(Integer)
+
+    error('workorder_cash_on_hand is invalid. The proper format is 5000 for 5 gold')
```
rpherbig: 💡 You might make the phrasing a bit clearer that '5000 for 5 gold' is an _example_ and not a requirement.

PR #2941, opened 2018-07-15
Delete flags before script ending in various scripts.

Sarvatt:
No reason to keep the regex checking going on every line after they are done.

### DIFF
--- a/carve.lic
+++ b/carve.lic
@@ -180,4 +180,9 @@ class Carve
   end
 end
 
+before_dying do
+    Flags.delete('carve-assembly')
+    Flags.delete('carve-magic-ready')
+end
+
 Carve.new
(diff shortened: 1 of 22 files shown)


rpherbig:
❓ Would these work in `before_dying`? That could reduce code duplication (and make sure we don't miss any exits in the middle of a script).

Sarvatt:
before_dying does indeed work, I added a bunch more and made sure not to delete any that come from common files that need to persist. I'll run this overnight and see how it looks in the morning compared to now.

Sarvatt:
I'm actually happy with this as-is, didn't need to run it for a long time. I could just run the scripts then kill them at the point where Flags.add added them to verify its working.

PR #2944, opened 2018-07-17
Add an optional Armor Swap Hysteresis Logic

b-Y-d:
Stops swapping armors once all armors in swap cycle are above 32/34.
Equips the user's 'default' armor set once it toggles on hysteresis.
Will resume armor swapping after any of them drop below 25/34.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -116,10 +116,8 @@ class SetupProcess
         @equipment_manager.wear_items(@equipment_manager.desc_to_items(@cycle_armors[@default_armor]))
         @last_worn_type = @default_armor
       end
-      return true
-    else
-      return false
     end
+    true
   end
   
   def check_armor_swap(game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -129,7 +129,6 @@ class SetupProcess
     toggle_armor_hysteresis?
     return if @armor_hysteresis
     return if Time.now - @last_cycle_time < @cycle_armors_time
-    return if Time.now - @last_cycle_time < @cycle_armors_time
     return if game_state.loaded
     armor_types = @cycle_armors.map { |skill, _| skill }
     next_armor_type = armor_types.min_by { |skill| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -104,8 +106,29 @@ class SetupProcess
...
   end
-
+  
+  def toggle_ah(mode)
+    @armor_hysteresis = mode
+    echo("*** armor hysteresis: #{@armor_hysteresis} ***")
+  end
+  
+  def toggle_armor_hysteresis?
+    if @armor_hysteresis == true
```
rpherbig: 💡 You don't need `== true`

combat-trainer.lic
```diff
@@ -104,8 +106,28 @@ class SetupProcess
 
     game_state.wield_whirlwind_offhand
   end
-
+  
+  def toggle_ah(mode)
```
rpherbig: 💡 I'd prefer the function name to be fully spelled out for the sake of clarity

combat-trainer.lic
```diff
@@ -104,8 +106,28 @@ class SetupProcess
...
+    @armor_hysteresis = mode
+    echo("*** armor hysteresis: #{@armor_hysteresis} ***")
+  end
+  
+  def toggle_armor_hysteresis?
```
rpherbig: ❌ The `?` at the end of the function name means this is a predicate. This function does not explicitly return a boolean and its return value is never used. Please remove the `?`

combat-trainer.lic
```diff
@@ -22,6 +22,8 @@ class SetupProcess
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @cycle_armors_time = settings.cycle_armors_time
     echo("  @cycle_armors_time: #{@cycle_armors_time}") if $debug_mode_ct
+    @armor_hysteresis = settings.armor_cycle_hysteresis || false
```
rpherbig: ❓ The name of this setting implies that the user would use it to turn the hysteresis feature on or off. However, its value is overwritten. What is the intent of this variable?

combat-trainer.lic
```diff
@@ -104,8 +106,28 @@ class SetupProcess
 
     game_state.wield_whirlwind_offhand
   end
-
+  
+  def toggle_ah(mode)
+    @armor_hysteresis = mode
+    echo("*** armor hysteresis: #{@armor_hysteresis} ***")
```
rpherbig: ❌ Please add `if $debug_mode_ct`

combat-trainer.lic
rpherbig: 💡 'toggle' in the name here threw me off since this function does not toggle (i.e. flip), rather it sets the var to a given value. Do you think another name could be clearer?

b-Y-d:
Changes Made

Sarvatt:
I like how you fixed this up. this looks good to me right now.

PR #2947, opened 2018-07-18
Common-Arcana: Pause after casting

jandersson:
This will give more time for the game to return a backfire message.
Currently the method returns too quickly in some instances, reporting a
backfire as a successful cast.

Note: This commit is cherry-picking from a Jan 21 commit on my fork.


jandersson:
The pause amount may be too generous, it could be lowered with some testing or removed entirely with some updated logic for backfire checks.

rpherbig:
How thoroughly has this been tested?

jandersson:
Its been tested for ~ 7 months

Sarvatt:
tested by me for 18 hours also, and it does fix the backfires with use_auto_mana

PR #2951, opened 2018-07-19
combat-trainer.lic: Adjust the hysteresis setting name to match the o…

Sarvatt:
…thers, and add it to base.yaml

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -22,7 +22,7 @@ class SetupProcess
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @cycle_armors_time = settings.cycle_armors_time
     echo("  @cycle_armors_time: #{@cycle_armors_time}") if $debug_mode_ct
-    @armor_hysteresis = settings.armor_cycle_hysteresis
+    @armor_hysteresis = settings.cycle_armors_hysteresis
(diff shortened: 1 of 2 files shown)


rpherbig:
Seems reasonable. And the feature is young enough that I doubt very many people are using this setting. @Sarvatt can you post something on LNet a few times throughout the day?

FYI @b-Y-d

combat-trainer.lic
```diff
@@ -22,7 +22,7 @@ class SetupProcess
     echo("  @cycle_armors: #{@cycle_armors}") if $debug_mode_ct
     @cycle_armors_time = settings.cycle_armors_time
     echo("  @cycle_armors_time: #{@cycle_armors_time}") if $debug_mode_ct
-    @armor_hysteresis = settings.armor_cycle_hysteresis
+    @armor_hysteresis = settings.cycle_armors_hysteresis
```
rpherbig: 💡 If you wanted to be extra safe, you could add a validator that `armor_cycle_hysteresis` is not set to true.

This feature is young enough that it's probably not necessary.

Sarvatt: I really think it's not necessary, only Byd is using it. I noticed today when I went to add after updating it that the name was a little askew.

Sarvatt: I did ask Byd what he thought about this change beforehand and he said he's all for it.

PR #2956, opened 2018-07-20
Update skill-recorder.lic

aruthas:
skill_checker was updated so I could use the thief_passages.  I was fixing shard and this seemed the easiest way to fix it for all thieves

### DIFF
--- a/skill-recorder.lic
+++ b/skill-recorder.lic
@@ -1,18 +1,13 @@
 =begin
   Documentation: https://elanthipedia.play.net/Lich_script_repository#skill-recorder
 =end
-
 custom_require.call(%w[drinfomon spellmonitor common])
-
 no_pause_all
 no_kill_all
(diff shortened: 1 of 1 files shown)


Sarvatt:
how about just putting all the passages assignments in the same conditional as crossing_leth instead of checking 4 more variables every 2 seconds?

aruthas:
Only reason I did it that way, is because you dont need athletics to use these other passages

Sarvatt:
well they both could be condensed into one if DRStats.guild == 'Thief' && DRStats.circle > 5 at any rate

Sarvatt:
looks good to me and this is needed to fix a bunch of problems with the map in shard that he already has staged @rpherbig

aruthas:
Did I do something wrong?  It says some of the checks havent been completed

PR #2958, opened 2018-07-21
outdorsmanship.lic: Bump the default collects to 4 so a spell can go …

Sarvatt:
…off for lower level people who can get 3 mindstates in 2 collects, and releases symb in cleanup


Sarvatt:
without this change, some people are preparing a spell and moving on before its ever cast. 4 mindstates or 4 collects should be ok as a default. I've had a few people who hit that situation test it.

PR #2961, opened 2018-07-21
combat-trainer.lic: Add combat teaching support for training_abilities.

Sarvatt:


validate.lic
```diff
@@ -395,7 +395,7 @@ class DRYamlValidator
     return unless settings.training_abilities
 
     settings.training_abilities
-            .reject { |skill, _cooldown| ['PercMana', 'Perc', 'Perc Health', 'Astro', 'App', 'App Quick', 'App Careful', 'Tactics', 'Hunt', 'Pray', 'Scream', 'Khri Prowess', 'Stealth', 'Ambush Stun', 'Ambush Choke', 'Favor Orb', 'Charged Maneuver', 'Meraud', 'Recall', 'Barb Research Debilitation', 'Barb Research Augmentation', 'Barb Research Warding', 'Collect', 'Analyze', 'PrayerMat', 'Smite', 'Locks'].include?(skill) }
+            .reject { |skill, _cooldown| ['PercMana', 'Perc', 'Perc Health', 'Astro', 'App', 'App Quick', 'App Careful', 'Tactics', 'Hunt', 'Pray', 'Scream', 'Khri Prowess', 'Stealth', 'Ambush Stun', 'Ambush Choke', 'Favor Orb', 'Charged Maneuver', 'Meraud', 'Recall', 'Barb Research Debilitation', 'Barb Research Augmentation', 'Barb Research Warding', 'Collect', 'Analyze', 'PrayerMat', 'Smite', 'Locks', 'Teach'].include?(skill) }
             .each_key { |skill| error("Ability in training_abilities is not valid: #{skill}") }
```
rpherbig: ❓ Is there value to a validator that checks that the value of `combat_teaching_skill` is a valid skill? Or warns when it's say, Stealth, Thievery, Sorcery, Arcane, ...?

combat-trainer.lic
```diff
@@ -1727,7 +1727,7 @@ class TrainerProcess
     @skill_map = { 'Hunt' => 'Perception', 'Pray' => 'Theurgy', 'PercMana' => 'Attunement',
                    'Perc' => 'Attunement', 'Astro' => 'Astrology', 'App' => 'Appraisal', 'PrayerMat' => 'Theurgy',
                    'App Quick' => 'Appraisal', 'App Careful' => 'Appraisal', 'Tactics' => 'Tactics', 'Analyze' => 'Tactics',
-                   'Scream' => 'Bardic Lore', 'Perc Health' => 'Empathy', 'Khri Prowess' => 'Debilitation',
+                   'Scream' => 'Bardic Lore', 'Perc Health' => 'Empathy', 'Khri Prowess' => 'Debilitation', 'Teach' => 'Scholarship',
```
rpherbig: ❓ Is the goal for the teacher to learn Scholarship or the teachee to learn the skill? I.e. should this be controlled by a timer or something other than skill experience like 'Favor Orb'?

combat-trainer.lic
```diff
@@ -1816,6 +1819,12 @@ class TrainerProcess
       pickbox(game_state)
     when 'Hunt'
       bput('hunt', 'You take note of ', 'You find yourself unable to hunt in this area') unless game_state.retreating?
+    when 'Teach'
```
rpherbig: ❓ Is my understanding correct that the 'exp timer' value for 'Teach' is 'how often to check the room for students'?

combat-trainer.lic
```diff
@@ -1816,6 +1819,12 @@ class TrainerProcess
...
+      UserVars.friends.each do |character|
```
rpherbig: 💡 Should this be a setting that gets merged with a UserVar in dependency.lic? Or are you confident people would only want to use a UserVar here?

Sarvatt: that uservar is populated by hunting_buddies: in your yaml, I don't really see value in making it separately configurable outside of that since you want to be hunting with them for it to work

combat-trainer.lic
Sarvatt: yeah, didn't want to make this too smart here. I don't care if I miss a teaching pulse while hiding (which stops the class) enough to juggle interrupting more important things like combat if it drops off, perfectly happy to just check to teach every minute or two.

combat-trainer.lic
Sarvatt: its just to trickle learn scholarship and the skill you pick for both parties, mostly useful for people who don't leave combat

validate.lic
Sarvatt: yep good call, I'll make one. Don't think that should block this though.

combat-trainer.lic
```diff
@@ -1783,6 +1783,9 @@ class TrainerProcess
     @analyze_retry_count = settings.combat_analyze_retry_count
     echo("  @analyze_retry_count: #{@analyze_retry_count}") if $debug_mode_ct
 
+    @combat_teaching_skill = settings.combat_teaching_skill
```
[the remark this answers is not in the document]
Sarvatt: i'd rather not add all the logic to stop teaching and rotate through skills personally, using one skill per yaml here to train stuff I can't learn on that hunt via teaching and also getting scholarship which ends up locked after an hour hunt.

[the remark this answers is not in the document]
rpherbig: :+1:

validate.lic
rpherbig: :+1:

combat-trainer.lic
rpherbig: :+1:

combat-trainer.lic
rpherbig: :+1:

combat-trainer.lic
rpherbig: :+1: I missed that, thanks

PR #2964, opened 2018-07-22
Adding thicket vipers and serpents

izawwlgood:
Hunting area outside riverhaven in the pejek bog

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -966,6 +966,13 @@ hunting_zones:
   - 3073
   - 3072
   - 3070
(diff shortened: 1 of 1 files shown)


data/base-hunting.yaml
```diff
@@ -966,6 +966,13 @@ hunting_zones:
   - 3073
   - 3072
   - 3070
+  #https://elanthipedia.play.net/Giant_thicket_viper
```
rpherbig: 💡 It would be nice to have an approximate range of ranks in here too (e.g. line 976)

PR #2972, opened 2018-07-23
athletics.lic: Call outdoorsmanship.lic directly so you can train magic

Sarvatt:


athletics.lic
```diff
@@ -349,9 +349,7 @@ class Athletics
         walk_to(19_464)
         break if done_training?
         walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-        4.times do
-          train_outdoorsmanship
-        end
+        start_script('athletics', ['4'])
```
rpherbig: ❓ Should these really be starting `athletics`?

PR #2976, opened 2018-07-24
Minor update to MAB

izawwlgood:
Revised to skill: Utility and added expiration messaging.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -1129,9 +1129,10 @@ spell_data:
     mana: 15
     recast: 1
   Magnetic Ballista:
-    skill: Targeted Magic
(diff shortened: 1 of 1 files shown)


data/base-spells.yaml
```diff
@@ -1129,9 +1129,10 @@ spell_data:
     mana: 15
     recast: 1
   Magnetic Ballista:
-    skill: Targeted Magic
+    skill: Utility
```
rpherbig: ❓ Isn't MAB both TM and Utility? If so, why the change?

BinuDR: Probably so that it will recast on the expire message.

PR #2977, opened 2018-07-24
add arrow and bolt to Hib/Boar Clan

tulanegator:
just moved someone to Hib and realized this wasn't added. tested the change and it worked. technically the purchase is in Boar Clan but I don't know of a shop in Hib itself.


Sarvatt:
thanks for this, ilithi long arrow purchasing has been getting stuck for me, this is close enough to not care if restock needs to happen

PR #2981, opened 2018-07-24
common.lic: Allow arcana in listen? for barbs and thieves

Sarvatt:


Sarvatt:
smartlisten would need to be updated too anyway

PR #2984, opened 2018-07-27
base-theurgy.lic: Add favor altar location info for crossing, to eve…

Sarvatt:
…tually be consumed by an alternate method to get favors in any city using ;find to buy primers.


Sarvatt:
once i get this added to ;favor as an arg to get whatever type of orb you want I want to automate the first part of the truffenyi commune quest getting the 3 orbs, that's the motivation for this. being able to get orbs outside crossing would be nice in general when we get more data added.

rpherbig:
Nice

PR #2986, opened 2018-07-28
base-theurgy.yaml: Various changes to accomidate favor altars

Sarvatt:
Add aspect_to_primer to map the npc order numbers, and add the primer npc name and start tracking free holy water rooms.

I have an updated favor.lic to use the altars, but would like to get the data in first so more people can test it easily.

### DIFF
--- a/data/base-theurgy.yaml
+++ b/data/base-theurgy.yaml
@@ -43,6 +43,47 @@ immortal_to_aspect:
   Berengaria: cow
   Divyaush: welkin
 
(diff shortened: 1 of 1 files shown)


Sarvatt:
https://github.com/Sarvatt/dr-scripts/commit/b0c1701adb081d54b6cd0a6d2eb22ab2daf2b82b is the current favor.lic i'm going to submit after writing validate updates for it if you want to see what context these are used in

PR #2987, opened 2018-07-28
Update to force it to exhale if not worn

BinuDR:
Checks if crystal is worn when the response would be to gaze at the crystal, that way it won't gaze at a non-worn crystal and then put it away.

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -22,7 +22,7 @@ class SanowretCrystal
     return if @invalid_rooms.include?(Room.current.id)
     return if hiding? || invisible?
 
-    if DRSkill.getxp('Arcana') <= 10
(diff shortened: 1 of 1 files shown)


BinuDR:
This is for sanowret-crystal.lic.  I forgot to include that in the subject

sanowret-crystal.lic
```diff
@@ -22,7 +22,7 @@ class SanowretCrystal
     return if @invalid_rooms.include?(Room.current.id)
     return if hiding? || invisible?
 
-    if DRSkill.getxp('Arcana') <= 10
+    if DRSkill.getxp('Arcana') <= 10 && @worn_crystal
```
jandersson: Does this affect f2p users? (they cannot wear the crystal)

jandersson: That is my understanding at least.

PR #2988, opened 2018-07-28
Updated my yamls, removed obsolete ones.

jayRyan24:


jandersson:
The CI fails because of custom settings, can you comment those and add the character names to the ignore list in the tests?

jandersson:
This is the ignore list: https://github.com/rpherbig/dr-scripts/blob/master/test/test_valid_yaml.rb#L19

jayRyan24:
Ok, I commented them already in the yamls but added the names to the list.

PR #2990, opened 2018-07-28
Add favor altar support to favor.lic

Sarvatt:
And use it in training-manager if use_favor_altars: true is set in your yaml. This opens up all non-neutral aspects and is significantly faster than doing the puzzle when you get a higher number of favors. It also will open up non-crossing favors when the data is populated as well as being used to get orbs for the first stage of the truffenyi commune quest in a future script. holy_water and primer_npc needs to be defined in the base files for other towns, and the primer npcs wander so they will need to be added to find.lic. shard is already in.


Sarvatt:
Hmm, the crossing temple is hardcoded for turning in orbs, there is a general altar in shard that works too. I'll break that out to base-theurgy.

Sarvatt:
actually I'd rather add that later, and not potentially introduce more breakage over what i've tested. this is already a pretty big change

favor.lic
```diff
@@ -2,17 +2,88 @@
...
+    end
+  end
+
+  def use_altar
+    unless @water_holder
+      echo('You must purchase a water container and set water_holder: in your yaml to use altars for favors')
+      exit
+    end
+    if @primer_number > 300
+      ensure_copper_on_hand(750, @settings)
```
rpherbig: 💡 Is 750 copper so much that we can't just always grab that?

Sarvatt: nope, would you prefer that? I figured it might save a trip if someone uses 100-199 primers with a light aspect god with the default 3 silver on hand money.

profiles/base.yaml
```diff
@@ -554,6 +554,8 @@ trail_override:
 ignored_npcs:
 
 favor_god: Hodierna
+# Offer primers on favor altars instead of doing the puzzle, allows more immortal choices
+use_favor_altars:
```
rpherbig: ❓ Should this be `use_favor_altars: false`?

training-manager.lic
```diff
@@ -122,19 +123,27 @@ class TrainingManager
...
       fput('put my orb on altar')
       if favor_count + 1 < @settings.favor_goal
-        wait_for_script_to_complete('favor')
+        run_favors
         fput('stow my orb')
       end
     end
   end
 
+  def run_favors
```
rpherbig: ❓ Why is the use of `use_favor_altars` part of `training-manager` and not `favor`?

favor.lic
rpherbig: Eh, I don't like magic numbers, but it's not a big deal.

training-manager.lic
Sarvatt: because its making training-manager pass the arg to automatically use favor altars over using the puzzle, people using ;favor can call it with an arg to turn it on manually

Sarvatt: ok, so, use case here is, I want to use ;favor immortalname called from another script in the future to make truffenyi-commune-quest handle getting the 3 orbs at the start of the quest, that was the whole goal of adding altar favors. i'd rather it not be baked in to favor

PR #2992, opened 2018-07-29
Adding the ability to set a pet_on_hand limit.

jayRyan24:
You can set a limit now and it will maintain this amount.

Just typing `;pick pets` will do this.

You can force a number of pet boxes to be picked if you have the variable set. 

Just typing `;pick pets 2 force` will do this.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -112,9 +112,10 @@ class LockPicker
   def setup
     arg_definitions = [
       [
-        { name: 'pets', regex: /pets/, optional: true, description: 'Disarm boxes and place them in the pet box container' },
-        { name: 'count', regex: /\d+/, optional: true, description: 'How many pet boxes to make' },
-        { name: 'refill', regex: /refill/, optional: true, description: 'Refills your lockpick ring' }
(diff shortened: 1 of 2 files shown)


pick.lic
```diff
@@ -112,19 +112,28 @@ class LockPicker
...
     args = parse_args(arg_definitions, true)
 
     @make_pets = args.pets
     @pet_goal = args.count.to_i
+    force = args.force
     @pet_count = 0
 
     @settings = get_settings(args.flex)
+
+    if @make_pets and not force
```
rpherbig: 💡 `!force` instead of `not force`

pick.lic
```diff
@@ -112,19 +112,28 @@ class LockPicker
...
+      box_count = get_boxes(@settings.picking_box_source).count
+      @pet_goal = @settings.pet_boxes_on_hand - box_count
+      exit unless @pet_goal > 0
```
rpherbig: 💡 I prefer to use 'if' when possible. In this case, that would mean inverting the conditional, something like `exit if @pet_goal <= 0` (or possibly `exit if @pet_goal.zero?`)

PR #3000, opened 2018-07-29
Added Segoltha climbing bank in Crossing

mwest152:
Added this as an alternative to the default climbing of the branch by the gondola because characters without stealth skill cannot do that spot.

### DIFF
--- a/data/base-athletics.yaml
+++ b/data/base-athletics.yaml
@@ -76,7 +76,7 @@ practice_options:
       id:  2245
       target: branch
       hide: true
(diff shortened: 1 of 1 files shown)


Sarvatt:
oh the bank teaches at 350? that makes sense then

Sarvatt:
looks good to me 👍

mwest152:
I noticed this climbing spot was actually already in the base-athletics under therengia.  with a different name so probably isn't an issue.  Deleting it would just mess up someone that might already have it i guess.

PR #3002, opened 2018-07-30
Base-recipes: remove duplicate entries

jandersson:
Some of the robe items were listed twice. I left in the ones with workorders turned on


jandersson:
Checking for duplicate recipe names would be a good test

PR #3003, opened 2018-07-30
Combat-Trainer: wear moonblade instead of breaking it

jandersson:
Moonblades can be worn, no need to destroy them

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1352,7 +1352,11 @@ class SpellProcess
 
   def cast_ritual(data, game_state)
     if summoned = game_state.summoned_info(game_state.weapon_skill)
-      break_summoned_weapon(game_state.weapon_name)
+      if DRStats.moon_mage?
+        bput('wear moon', 'telekinetic')
+      else
+        break_summoned_weapon(game_state.weapon_name)
+      end
     else
       @equipment_manager.stow_weapon(game_state.weapon_name)
     end


combat-trainer.lic
```diff
@@ -1352,7 +1352,11 @@ class SpellProcess
 
   def cast_ritual(data, game_state)
     if summoned = game_state.summoned_info(game_state.weapon_skill)
-      break_summoned_weapon(game_state.weapon_name)
+      if DRStats.moon_mage?
+        bput('wear moon', 'telekinetic')
```
rpherbig: ❓ Does line 1371 need a conditional to 'get' the moonblade?

jandersson: Yes it does

jandersson: It also looks like a get and wield moonblade command might be good targets for common methods in combat-trainer as I see they are used extensively

jandersson: Thoughts on creating common-moonmage.lic? This could also house astrology functions. Combat astrology could use methods commonized from astrology.lic

rpherbig: Sounds like a good idea 👍

jandersson:
Yikes, this dropped off my radar. Thanks for the ping. I will implement the suggested changes soon.

jandersson:
I will reopen this when I have time to look at the specific changes that need to be made. Combat trainer is a sensitive script and making changes without really testing them is going to affect a lot of users.

PR #3004, opened 2018-07-30
 Add astrology prediction targets.

BastionWT:
Negative predictions can sometimes ruin hunting. These changes keep the core logic in the base.yaml in case people don't want to mess with it, but allow for the overriding of the prediction targets to skills that the user determined as necessary.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -32,6 +32,11 @@ class Astrology
     @ap_destination = settings.astral_plane_training['train_destination']
     @prediction_pool_target = settings.astrology_use_full_pools ? 10 : settings.astrology_pool_target
     @rtr_data = nil
+    @astrology_prediction_skills_magic = settings.astrology_prediction_skills['magic']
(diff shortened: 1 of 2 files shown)


jandersson:
This looks OK to me! As a small suggestion I think the astrology prediction targets could get compressed into a hash, e.g. ```ruby
    @astrology_prediction_skills = settings.astrology_prediction_skills

# Later on 

  def predict_all(pools)
    skillset_to_pool = {
      'offensive combat' => 'offense',
      'defensive combat' => 'defense',
      'magic' => 'magic',
      'survival' => 'survival',
      'lore' => 'lore',
      'offensive combat' => @astrology_prediction_skills['offense'],
      'defensive combat' => @astrology_prediction_skills['defense'],
      'magic' => @astrology_prediction_skills['magic'],
      'survival' => @astrology_prediction_skills['survival'],
      'lore' => @astrology_prediction_skills['lore'],
      'future events' => 'future events'
    }

PR #3012, opened 2018-07-31
skill-recorder.lic: Various fixes.

Sarvatt:
Move the thief tunnel checks to script startup and outside of the loop.
Only record the scouting uservar if you are a ranger.
up the time of the loop, checking the skills to update the db every 2 seconds is excessive.


Sarvatt:
the 30 seconds was pulled out of a hat, writing to the db every 2 seconds was excessive though. happy to change it to anything else if there are worries about that.

rpherbig:
❓ Why were the Thief checks moved? It seems reasonable enough that your circle or Athletics rank could change within a single session. And one of the checks uses `UserVars.athletics` which isn't populated (or updated) until the loop (i.e. after the check).

PR #3014, opened 2018-08-02
Attempt to fix aim/load loop

dr-leustyin:


rpherbig:
❓ Is this the same as line 2435?

PR #3015, opened 2018-08-02
base-crafting updates for Muspar'i Societies

Tarhish:
All disciplines tested through multiple ;workorders after dropping all restockables to demonstrate pulling oil and yarn and whatever.  Forging and Outfitting only support 1 workroom of each type due to identical room descriptions in their sewing, weaving, anvil and crucible rooms.

### DIFF
--- a/data/base-crafting.yaml
+++ b/data/base-crafting.yaml
@@ -56,6 +56,40 @@ blacksmithing:
     idle-room: 8775
   Leth Deriel:
     << : *zoluren_blacksmithing
+  Muspar'i:
+    npc: Master
+    npc_last_name: Master
+    npc-rooms:
+     - 11260
+     - 11261
+     - 11263
+     - 11264
+     - 11262
+     - 11265
+     - 11266
+     - 11267
+     - 14348
+    anvils:
+      - 14349
+    crucibles:
+      - 14350
+    grindstones:
+      - 14349
+    logbook: forging
+    pattern-book: blacksmithing
+    finisher: oil
+    finisher-full: flask of oil
+    finisher-room: 11263
+    finisher-number: 6
+    repair-room: 11267
+    repair-npc: Rokumru
+    stock-volume: 5
+    stock-room: 11264
+    stock-number: 11
+    stock-name: bronze ingot
+    trash-room: 14350
+    idle-room: 11260
+    part-room: 11263
   Riverhaven:
     npc: Fereldrin
     npc_last_name: Fereldrin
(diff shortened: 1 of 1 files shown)


Sarvatt:
This looks good, the only bad part is 1 room for some of the rooms like the sewing rooms. those are going to need to be mapped with a peer tag to get unique room numbers to get around the crappy way DR makes the rooms identical. usually they have different door materials you could match in the next room to make it unique. Then again it's not like many people are crafting in muspar'i and it's no different than how fang cove is set up, anything is better than not having support at all. Thanks for adding this!

Tarhish:
I saw odd tagging of rooms in some of the other societies, but I didn't understand it.  One of my characters is still up in Muspar'i and can definitely make some changes if I figure it out though.  The only problem is that while the outfitting society has 6 identical rooms behind 6 different curtains, the forging society is all NSEW plain rooms, and I had to snip all entrances to the identical parts or else go2 got confused.

I'll look at the map data for the other rooms that are like this and see if I can piece it out.

Sarvatt:
that is one of the hardest things to do to update the map, I'm guessing the curtains each had a unique look? here is an example of a room using a peer tag which makes you peer into the next room to determine the room you are in to avoid repeating rooms.

```   <room id="10463">
      <title>[[The Baby Metal Arachnid, Cephalothorax]]</title>
      <description>Thin steel plates are bolted together to form the floor and ceiling, arcing in a tangle of wires and gears high overhead.  Steaming pipes run along the walls, fitted with cogs and whirring clockwork parts that drive the machinery.  A metal grate blocks off a huge engine that chugs away with a steady rhythm, set up with levers that run to the left legs of the spider.</description>
      <paths>Obvious exits: east, west.</paths>
      <tag>peer east =~ /Several electrical burn marks mar a set of steel plates, warped slightly by whatever accident caused them./</tag>
      <image name="HolloweveBabySpider.gif" coords="253,93,268,107" />
      <exit target="10473" type="String" cost="0.2">east</exit>
      <exit target="10683" type="String" cost="0.2">west</exit>
   </room>``` there were 2 rooms there with identical descriptions, its using the peer tag to determine which of the two it is in. that could be used to look at the unique curtain descriptions but it will be hard to do.

Tarhish:
Oh wow, okay, that's a neat trick, so I've tentatively marked up my local map copy of the forging society with tags so that when I do an 'ids' it peers and assigns the room id based on what it sees in that direction.  Using your advice it looks like I can get all of them like that.  Once I test to make sure it's all stable and clean I'll look into finalizing that and adding it to the repository.

The curtains... yeah, definitely a little trickier considering they're invisible from the inside and each set of three lead to the same two rooms, but...  I'll think about that once I've actually fiddled for a while.

Thanks!

Tarhish:
Question.  I was able to work through the peer tags in the forging society and get all 6 rooms working.  I want to commit those rooms, but if I can get the outfitting done at the same time I'd like to.

Except I'm not sure I can.  Each of the three sewing rooms is identical and is connected to the same origin room, 11255, so if they PEER OUT they'll all see the same room as the others.  The weaving rooms are set up the same way.  I can't think of any set of actions that would let me distinguish between them while I'm inside one.

Unless someone has an idea I believe I'll have to just add the forging rooms and settle for that.

Tarhish:
I almost thought I had it.  I found ;e echo Room.new and guessed what the proper arguments were, and made (on my local map) two identical copies of 11257 before altering the waytos and timetos.  As a result I could go2 all three identical rooms individually and go2 out of them without problems, but because workorders uses a function that checks to make sure it reached the room successfully it fails right before crafting due to being unable to match itself to the room id it tried to enter.

While I GUESS it would be possible to change the scripts to accept any of the rooms in the sewing_rooms list as an ending point, it's sounding like it's not worth the potential side effects.

On the plus side I learned a lot about the map.

PR #3016, opened 2018-08-02
Added T2 trainer script

dr-leustyin:


t2.lic
```diff
@@ -0,0 +1,63 @@
+=begin
```
rpherbig: 💡 Do you plan to document this on elanthipedia?

PR #3017, opened 2018-08-02
Updated to not start performance if player uses an instrument

dr-leustyin:


first-aid.lic
```diff
@@ -30,8 +30,10 @@ class FirstAid
   end
 
   def compendium_charts
-    pause 2 # to give performance time to complete before_dying if stopped from the previous script
-    start_script('performance', ['noclean']) unless Script.running?('performance')
+    unless @settings.instrument
```
rpherbig: ❓ This is the same as `if ! @settings.instrument`. I thought the intent was to only start `performance` if the user _does_ have an instrument defined...?

dr-leustyin: We only want to start performance if the user does not have an instrument.  Doing first-aid and playing an instrument is not possible, so performance doesn't train.

I thought the convention was to use `unless` instead of `if not`.  I can get this updated if that's not the case

rpherbig: `unless` is fine, I was just trying to clarify the intent. Thank you.

PR #3025, opened 2018-08-03
base-theurgy.yaml: Add holy water source to shard

Sarvatt:


Sarvatt:
there is no natural source of holy water in p5 so making hometown: Hib people go there is the best option. It's not far with ice skates.

PR #3029, opened 2018-08-05
Updated the logic to be more accurate and useful. 

jayRyan24:
Ways to pick:

1. `;pick pets`: will automatically disarm the number of pets up to your `pet_boxes_on_hand:` limit
2. `;pick pets #`: will disarm up to the number of pets entered

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -128,9 +128,8 @@ class LockPicker
 
     @settings = get_settings(args.flex)
 
-    if @make_pets && !force
-      box_count = get_boxes(@settings.picking_box_source).count
-      @pet_goal = @settings.pet_boxes_on_hand - box_count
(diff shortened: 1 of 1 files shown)


Sarvatt:
would you mind lowering pet_boxes_on_hand: in base.yaml back to 1 with this change? the echo isn't really needed since it only triggers if someone literally puts `pet_boxes_on_hand:` in their yaml cancelling out the base, I'm guessing you hit it because you didn't update your own base.yaml though. looks good to me!

jayRyan24:
If I put it to 1, it means they will be required to use the setting, with it set to 0, they will continue on as normal. It's kind of a backward compatible thing. Not sure what to do here, maybe others can give input.

jayRyan24:
@rpherbig I'm ready.

PR #3031, opened 2018-08-07
Update ;pick to allow storing harvested trap components

Tarhish:
If component_container: is defined in the yaml, then when a picker is harvesting components it will attempt to place them inside the container.  If it can't, it will drop/dispose of them as normal.

Also a fix that prevents a rare problem I've run into.  There is a situation where stowing pliable equipment (eg. removing and stowing all the hindering equipment before beginning picking) can fill picking_box_source to the point that it can't stow a box it pulled out of it.  It would happen like:

  - Thief removes box from backpack
  - Thief fails to disarm box.
  - Thief tries to stow box but can't fit it inside the backpack anymore.
  - Thief grabs the next box and has two in his/her hands.

In rarer situations this can cause a picker to try and pick an un-disarmed box.  

With this change if it fails to stow the box it will drop it and continue picking.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -137,6 +137,7 @@ class LockPicker
     end
 
     @harvest_traps = !@make_pets && @settings.harvest_traps
+    @component_container = @settings.component_container
     @stop_pick_on_mindlock = !@make_pets && @settings.stop_pick_on_mindlock
     @loot_nouns = @settings.lootables
     echo "Loot nouns: #{@loot_nouns}" if UserVars.lockpick_debug
@@ -221,8 +222,11 @@ class LockPicker
     bput("get my #{box} from my #{@settings.picking_box_source}", 'You get a .* from inside ')
 
     unless disarm?(box)
-      if @settings.picking_box_storage
-        bput("put my #{box} in my #{@settings.picking_box_storage}", 'You put your .* in ') if right_hand
+      if @settings.picking_box_storage && right_hand
+		case bput("put my #{box} in my #{@settings.picking_box_storage}", 'You put your .* in ', "You just can\'t get", "There isn\'t any more")
+		when "You just can\'t get", "There isn\'t any more"
+			dispose_trash(box)
+		end
       elsif right_hand
         dispose_trash(box)
       end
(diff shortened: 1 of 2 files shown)


Sarvatt:
`component_container:` should be added to profiles/base.yaml

Sarvatt:
still waiting on an fput to bput update i mentioned in the review but thats easy enough to fix after.

Sarvatt:
hmm did that not go through? guess I'm not familiar with reviews here. I was asking if line 457 could be a bput instead, even if its bput("put my #{left_hand} in my #{@component_container}", 'You put')

Tarhish:
fputs no longer desired?  Can do, and will keep that in mind going forward.

pick.lic
```diff
@@ -452,6 +456,11 @@ class LockPicker
   end
 
   def harvest(box)
+  	patterns = /glass reservoir|steel striker|black cube|chitinous leg|
```
jayRyan24: Minor indenting issue here.

pick.lic
```diff
@@ -463,7 +472,8 @@ class LockPicker
       harvest(box)
     when 'Roundtime'
       waitrt?
-      dispose_trash(left_hand) # only dispose https://elanthipedia.play.net/Locksmithing_skill#Box_Traps items.
+	  bput("put my #{left_hand} in my #{@component_container}", 'You put your', "You just can\'t get", "There isn\'t any more") if (@component_container != nil && left_hand =~ patterns)
```
jayRyan24: I think you can just do 

`if @component_container && left_hand =~ patterns`

or 

`if @component_container && (left_hand =~ patterns)`

Also, there's the .nil? method. `if !@component_container.nil?`

pick.lic
```diff
@@ -221,8 +222,11 @@ class LockPicker
     bput("get my #{box} from my #{@settings.picking_box_source}", 'You get a .* from inside ')
 
     unless disarm?(box)
-      if @settings.picking_box_storage
-        bput("put my #{box} in my #{@settings.picking_box_storage}", 'You put your .* in ') if right_hand
+      if @settings.picking_box_storage && right_hand
+		case bput("put my #{box} in my #{@settings.picking_box_storage}", 'You put your .* in ', "You just can\'t get", "There isn\'t any more")
```
jayRyan24: Very large indentation here. I try to keep things to a 2 space tab. This is pretty standard in the industry.

jayRyan24:
fput isn't as safe as bput. fput will attempt 1 more time, and bput will keep trying.

Tarhish:
Right on all accounts.  Actually the indentation looks fine on my editor and matches everything else until it hits the repository so there's gotta be something about how I'm configured that's screwed up the formatting.  I already figured out the Linux EOL issue, but this is something else.  I'll fix this and resubmit so it doesn't keep happening.

Thanks for the feedback!

Tarhish:
Like a scrub I was letting the editor auto-add tabs where I didn't need them.  Shouldn't be a problem now, and I also fixed the needless nil comparison.  While I was looking at it I realized it was missing a step to open the component_container, so I added that.

PR #3046, opened 2018-08-08
combat-trainer.lic: enable dual load for thieves if khri steady is up.

Sarvatt:


Sarvatt:
added the ranger messaging here also because i noticed that check was missing.

PR #3047, opened 2018-08-08
athletics.lic: be extra careful about not doing the hardest climbs if…

Sarvatt:
… the athletics uservar is not set for some reason


Sarvatt:
I don't know how they got in the state, but someone with 35 athletics was sent through xalas, which means they did not have the uservar set somehow.

PR #3071, opened 2018-08-20
Added support to force careful disarming.

b-Y-d:
yaml setting `lockpick_force_disarm_careful: true`, but should default to false.

Will add it to base.yaml and PR that too. Too inexperienced with github to do them at the same time.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -392,10 +392,11 @@ class LockPicker
       disarm_speed?(box, 'careful')
     when *@disarm_quick
       disarm_speed?(box, 'careful') if @settings.lockpick_force_disarm_careful
-      disarm_speed?(box, 'quick') if !@settings.lockpick_force_disarm_careful
+      disarm_speed?(box, 'quick') unless @settings.lockpick_force_disarm_careful
     when *@disarm_normal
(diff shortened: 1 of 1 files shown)


pick.lic
```diff
@@ -391,10 +391,11 @@ class LockPicker
     when 'Roundtime', *@disarm_careful
       disarm_speed?(box, 'careful')
     when *@disarm_quick
-      disarm_speed?(box, 'quick')
+      disarm_speed?(box, 'careful') if @settings.lockpick_force_disarm_careful
+      disarm_speed?(box, 'quick') if !@settings.lockpick_force_disarm_careful
```
rpherbig: 💡 Consider using `unless` for the `if !...` cases

pick.lic
```diff
@@ -391,10 +391,11 @@ class LockPicker
...
     when *@disarm_normal
-      disarm_speed?(box, 'normal')
-    when *@disarm_identify_failed
```
rpherbig: ❓ Was this `when` supposed to be removed?

b-Y-d:
Fixed, and no that **when** was not supposed to be removed.

PR #3073, opened 2018-08-20
Reverting BPUT timeout timer to 15 seconds.

b-Y-d:
Some lag issues may be causing people to lose items in relation to bput timing out too fast. Safer to revert.


rpherbig:
Please update the tests too (or use git revert on the commit)

PR #3075, opened 2018-08-20
common.lic: Bump bput timeout up to 10 seconds

Sarvatt:
5 seconds is proving to be too little in some lag situations, but 15 seconds is too much. Lets try 10.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -66,7 +66,7 @@ module DRC
     clear
     put message
     timer = Time.now
-    while (response = get?) || (Time.now - timer < 5)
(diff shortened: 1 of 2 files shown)


rcuhljr:
Out of curiosity, why does this need to be shorter than it was originally? This is an error state.

Sarvatt:
I have no vested interest in arguing it :D

rcuhljr:
Nooooo I mean why was it changed to 5 in the first place. I hate you all.

PR #3077, opened 2018-08-20
circlecheck.lic: Fix up clerics again.

Sarvatt:
Note to self: Don't blindly trust the wiki. This one is verified against other circle calculators.


rpherbig:
I hope you fixed the wiki

PR #3078, opened 2018-08-20
combat-trainer.lic: Change combat_training_spells_max_threshold to a more telling name

Sarvatt:
This setting limits casting combat_training_spells to only when there are a certain amount of npcs in the room. I want to add an actual combat_training_spells_max_threshold that limits only combat_spell_training mindstates the same way magic_training_spells_max_threshold does because the magic version limits offensive casting too.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -857,8 +857,8 @@ class SpellProcess
     @training_spells = settings.combat_spell_training
     echo("  @training_spells: #{@training_spells}") if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


Sarvatt:
I agree and I'm happy to change it, I just felt like there was only one person in the world using it and I spent a few days explaining what it actually meant to people. Wish I reviewed it more before it was accepted.

Sarvatt:
don't care any more at this point, it would be easier to just add a comment in base.yaml about what it does

PR #3086, opened 2018-08-25
Combat-trainer support for regalia+starlight

Tarhish:
Added support for Trader Regalia, specifically for its use for full-set armor-swapping using Bespoke Regalia, though it also works for keeping a single set alive.  It handles refreshing Regalia if it starts to expire in combat.  If regalia fails due to lack of starlight it swaps back to your standard gearset.   The cycling uses existing cycle_armors logic with your set cycle_armors_time.  If only one regalia cycle set is defined it will stick with that one, refreshing it when it's about to expire.

As is, it lets me train all four complete armor sets without clown-suiting or carrying a set of plate, chain and brigandine around, then shift into my best type when it's all trained.

I would appreciate some inspection or testing beyond what I've done already, especially for someone without Bespoke Regalia to test this script by setting cycle_armors_regalia with just your highest armor skill.  I can't unlearn the metaspell for a week or two.

Usage:
Define a regalia: gearset that's missing any armor you want to replace.

gear_sets:
  standard:
  - quilted hauberk
  - mail gloves
  - ring balaclava
  - steel shield
  - parry stick
  regalia:
  - steel shield
  - parry stick

Define a cycle_armors_regalia: set that contains the item names that Regalia will be summoning for each skill.  If replacing everything it will look like below, but you could also include, say, just a mask for each set if your regalia set is only missing face armor.

cycle_armors_regalia:
  Brigandine:
    - scale hauberk
    - scale balaclava
    - scale gloves
  Light Armor:
    - carapace gloves
    - carapace balaclava
    - carapace hauberk
  Chain Armor:
    - chain hauberk
    - chain balaclava
    - chain gloves
  Plate Armor:
    - sallet
    - plate gauntlets
    - full plate

That's all you need to define.  Defining a regalia waggle, cycle_armors_time, and default_armor_type will improve its usage.  Defining a cycle_armors_regalia without a regalia gearset will generate a warning and cancel regalia functions.

Attempts to declare the items as gear didn't work out too well, since equipmanager wasn't really designed to handle items that can't be worn or removed.  Instead, it tracks these items by name just so it can bput remove them when they need to be destroyed.

If you have a default_armor_type: defined it will attempt to use that skill when it has no other armors to train, otherwise it will choose the highest-ranked on your list.  I can confirm this works exactly the way I want it for regalia, but I had to add it to the definitions in initialize, so perhaps someone else knows why it wasn't used yet.

If you have a regalia: waggle it will use the spell defined there, otherwise it will use the values in base_spells.

Known Limitations: 
It is not compatible with cycle_armors.  If both are declared it will generate a warning and nil the regalia swapping function before continuing.

If combat-trainer exits gracefully it puts the regalia away and swaps to your standard gearset before leaving.  If it's killed then you'll have old regalia pieces on you getting in the way.  This could cause a pileup if you kill combat-trainer and restart it right away.  Needs more development.

Also other minor changes,
starlight: attribute added to several spells in base-spells.  If combat-trainer detects a spell's failure due to lack of starlight it will delete all offensive_spells and buff_spells that have the starlight: true attribute.  There are probably more graceful ways to do this in the future, but it's better than it continuing to spam spells when it's out of starlight.  This can be overridden by declaring starlight: false in your spell list.

Also, added a 'Your concentration lapses' match to ct-spelllost.  In certain situations when I'm doing something weird and unnatural I might lose focus on a targeted spell without its target dying.

Also, a typo at the end of combat-trainer -- :balance_regen_threshold had an 'end' at the end of it

Please take a look -- I'd like other traders to try this out and I can make changes if we want to implement this a different way.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -39,7 +39,7 @@ class SetupProcess
       echo "ERROR - Regalia cycling and armorswap cycling at the same time not currently supported!  Removing Regalia from combat-training!"
       @cycle_regalia = nil
     end
-    if @cycle_regalia && settings.gear_sets['regalia'].empty?
+    if @cycle_regalia && (settings.gear_sets['regalia'].empty? || settings.gear_sets['regalia'].nil?)
       echo "ERROR - Regalia cycling requires a gear_set named 'regalia' that will be worn immediately before casting. Removing Regalia from combat-training!"
       @cycle_regalia = nil
     end
@@ -142,9 +142,10 @@ class SetupProcess
     return if armor_hysteresis?(game_state)
     return if Time.now - @last_cycle_time < @cycle_armors_time && !Flags['ct-regalia-expired']
     return if game_state.loaded
-    
-    @cycle_regalia = nil if Flags['ct-cancel-regalia']
-    
+    if game_state.regalia_cancel
+      regalia_shutdown(game_state) if Flags['ct-regalia-expired']
+      return
+    end
     armor_types = @cycle_regalia ? @cycle_regalia.map { |skill, _| skill } : @cycle_armors.map { |skill, _| skill }
     next_armor_type = armor_types.min_by { |skill| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }
     return if next_armor_type == @last_worn_type unless Flags['ct-regalia-expired']
@@ -282,6 +283,22 @@ class SetupProcess
     offhand_skill = temp_array.min_by { |skill| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }
     game_state.update_whirlwind_weapon_info(offhand_skill)
     game_state.wield_whirlwind_offhand
+  end
+  
+  def regalia_shutdown(game_state)
+    bput('retreat', 'You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat', 'retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing') if Flags['ct-engaged']
+    bput('retreat', 'You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat', 'retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing') if Flags['ct-engaged']
+    if game_state.last_regalia_type
+      @cycle_regalia[game_state.last_regalia_type].each do |item| 
+        bput("remove my #{item}", "into motes of silvery", "Remove what?", "You .*#{item}", 'The leather gauntlets slide')
+      end
+      game_state.last_regalia_type = nil
+      game_state.swap_regalia_type = nil
+      game_state.regalia_cancel = nil
+      @cycle_regalia = nil
+      Flags.reset('ct-regalia-expired')
+    end
+    @equipment_manager.wear_equipment_set?('standard')
   end
 end
 
@@ -965,7 +982,6 @@ class SpellProcess
     Flags.add('ct-need-bless', ' passes through the .* with no effect')
     Flags.add('ct-starlight-depleted', 'enough starlight') if DRStats.trader?
     Flags.add('ct-regalia-expired', 'Your .* glimmers weakly') if DRStats.trader?
-    Flags.add('ct-cancel-regalia')
 
     @offensive_spells
       .select { |spell| spell['expire'] }
@@ -985,6 +1001,8 @@ class SpellProcess
     @tk_spell = @offensive_spells.find { |spell| spell['abbrev'] =~ /tkt|tks/i }
     drop_tkt_ammo
     Flags.add('need-tkt-ammo', 'There is nothing here that can be thrown')
+    
+    @regalia_spell = get_data('spells').spell_data['Regalia'] if (@regalia_array && @regalia_spell.nil?)
   end
 
   def drop_tkt_ammo
@@ -1098,10 +1116,6 @@ class SpellProcess
     game_state.swap_regalia_type = nil if game_state.swap_regalia_type == game_state.last_regalia_type
     return if game_state.casting
     return unless game_state.swap_regalia_type
-    if @regalia_spell.nil?
-      echo "Couldn't find a regalia waggle.  Using default values for the spell."
-      @regalia_spell = get_data('spells').spell_data['Regalia']
-    end
     
     case game_state.swap_regalia_type
     when 'Light Armor'
@@ -1115,13 +1129,6 @@ class SpellProcess
     end
     @regalia_spell['cast'] = "cast #{armor_word} all"
     prepare_spell(@regalia_spell, game_state)
-    if game_state.last_regalia_type.nil? 
-      @equipment_manager.wear_equipment_set?('regalia')
-    else
-      @regalia_array[game_state.last_regalia_type].each do |item| 
-        bput("remove my #{item}", "into motes of silvery", "Remove what?", "You .*#{item}", 'The leather gauntlets slide')
-      end
-    end
   end
   
   def check_bless(game_state)
@@ -1203,7 +1210,19 @@ class SpellProcess
     if game_state.casting_sorcery
       @equipment_manager.stow_weapon(game_state.weapon_name)
     end
-
+    
+    if game_state.casting_regalia
+      bput('retreat', 'You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat', 'retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing')
+      bput('retreat', 'You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat', 'retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing')
+      if game_state.last_regalia_type.nil? 
+        @equipment_manager.wear_equipment_set?('regalia')
+      else
+        @regalia_array[game_state.last_regalia_type].each do |item| 
+          bput("remove my #{item}", "into motes of silvery", "Remove what?", "You .*#{item}", 'The leather gauntlets slide')
+        end
+      end
+    end
+    
     snapshot = DRSkill.getxp(@skill)
     success = cast?(@custom_cast, @symbiosis, @before, @after)
     if @symbiosis && @use_auto_mana
@@ -1360,22 +1379,22 @@ class SpellProcess
 
   def check_trader_magic(game_state)
     return unless DRStats.trader?
-    if Flags['ct-starlight-depleted'] then
+    if Flags['ct-starlight-depleted']
       echo "----OUT OF STARLIGHT - DELETING STARLIGHT MAGIC----"
-      @buff_spells.delete_if { |spell| @buff_spells[spell]['starlight'] }
-      @offensive_spells.delete_if do |spell|
-        if spell['starlight'] then true end
-      end
-      Flags['ct-cancel-regalia'] = true
+      @buff_spells.reject! { |spell| @buff_spells[spell]['starlight'] }
+      @offensive_spells.reject! { |spell| spell['starlight'] }
+      game_state.regalia_cancel = true
     end
     if game_state.casting_regalia
       if Flags['ct-starlight-depleted']  # In some conditions, regalia can run out of starlight partway through creation.
+        bput('retreat', 'You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat', 'retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing')
+        bput('retreat', 'You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat', 'retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing')
         @regalia_array[game_state.swap_regalia_type].each do |item| 
           bput("remove my #{item}", "into motes of silvery", "Remove what?", "You .*#{item}", 'The leather gauntlets slide')
         end
         @equipment_manager.wear_equipment_set?('standard')
       end
-      Flags['ct-starlight-depleted'] ? game_state.last_regalia_type = nil : game_state.last_regalia_type = game_state.swap_regalia_type
+      game_state.last_regalia_type = Flags['ct-starlight-depleted'] ? nil : game_state.swap_regalia_type
       Flags.reset('ct-regalia-expired')
       game_state.swap_regalia_type = nil
       game_state.casting_regalia = false
@@ -1521,7 +1540,7 @@ class SpellProcess
       end
     end
     
-    game_state.casting_regalia = true if data['abbrev'] == 'REGAL'
+    game_state.casting_regalia = data['abbrev'] == 'REGAL'
     
     prepare?(data['abbrev'], data['mana'], data['symbiosis'], command)
     game_state.casting = true
@@ -2721,7 +2740,7 @@ class GameState
   $ranged_skills = $thrown_skills + $aim_skills
   $non_dance_skills = $ranged_skills + ['Brawling', 'Offhand Weapon']
 
-  attr_accessor :mob_died, :last_weapon_skill, :danger, :parrying, :casting, :need_bundle, :cooldown_timers, :no_stab_current_mob, :loaded, :selected_maneuver, :cast_timer, :casting_moonblade, :casting_regalia, :casting_weapon_buff, :use_charged_maneuvers, :casting_consume, :prepare_consume, :casting_cfb, :prepare_cfb, :wounds, :blessed_room, :currently_whirlwinding, :whirlwind_trainables, :reset_stance, :charges_total, :casting_sorcery, :hide_on_cast, :starlight, :last_regalia_type, :swap_regalia_type
+  attr_accessor :mob_died, :last_weapon_skill, :danger, :parrying, :casting, :need_bundle, :cooldown_timers, :no_stab_current_mob, :loaded, :selected_maneuver, :cast_timer, :casting_moonblade, :casting_regalia, :casting_weapon_buff, :use_charged_maneuvers, :casting_consume, :prepare_consume, :casting_cfb, :prepare_cfb, :wounds, :blessed_room, :currently_whirlwinding, :whirlwind_trainables, :reset_stance, :charges_total, :casting_sorcery, :hide_on_cast, :regalia_cancel, :last_regalia_type, :swap_regalia_type
 
   def initialize(settings, equipment_manager)
     @equipment_manager = equipment_manager
@@ -2742,6 +2761,7 @@ class GameState
     @casting_sorcery = false
     @hide_on_cast = false
     @casting_regalia = false
+    @regalia_cancel = false
     @last_regalia_type = nil
     @swap_regalia_type = nil
     @casting_weapon_buff = false
@@ -3441,7 +3461,6 @@ before_dying do
   Flags.delete('last-stance')
   Flags.delete('ct-regalia-expired')
   Flags.delete('ct-starlight-depleted')
-  Flags.delete('ct-cancel-regalia')
 end
 
 $COMBAT_TRAINER = CombatTrainer.new
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1362,7 +1362,7 @@ class SpellProcess
       echo "----OUT OF STARLIGHT - DELETING STARLIGHT MAGIC----"
       @buff_spells.delete_if { |spell| @buff_spells[spell]['starlight'] }
       @offensive_spells.delete_if do |spell|
-        if spell['ct-starlight-depleted'] then true end
+        if spell['starlight'] then true end
       end
     end
     if game_state.casting_regalia
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -31,18 +31,13 @@ class SetupProcess
     echo("  @cycle_regalia: #{@cycle_regalia}") if $debug_mode_ct
     @default_armor = settings.default_armor_type
     echo("  @default_armor: #{@default_armor}") if $debug_mode_ct
+    
+    @gearsets = settings.gear_sets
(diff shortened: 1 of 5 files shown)


combat-trainer.lic
```diff
@@ -27,11 +27,23 @@ class SetupProcess
     @last_cycle_time = Time.now - @cycle_armors_time
     @combat_training_abilities_target = settings.combat_training_abilities_target
     echo("  @combat_training_abilities_target: #{@combat_training_abilities_target}") if $debug_mode_ct
-
+    @cycle_regalia = settings.cycle_armors_regalia
+    echo("  @cycle_regalia: #{@cycle_regalia}") if $debug_mode_ct
+    @default_armor = settings.default_armor_type
```
rpherbig: ❓ Where is `default_armor_type` used?

combat-trainer.lic
```diff
@@ -926,13 +950,22 @@ class SpellProcess
...
 
     @symbiosis_learning_threshold = settings.symbiosis_learning_threshold
     echo(" @symbiosis_learning_threshold: #{@symbiosis_learning_threshold}") if $debug_mode_ct
 
-    Flags.add('ct-spelllost', 'Your pattern dissipates with the loss of your target')
+    Flags.add('ct-spelllost', 'Your pattern dissipates with the loss of your target', 'Your concentration lapses')
     Flags.add('ct-need-bless', ' passes through the .* with no effect')
+    Flags.add('ct-starlight-depleted', 'enough starlight') if DRStats.trader?
+    Flags.add('ct-regalia-expired', 'Your .* glimmers weakly') if DRStats.trader?
+    Flags.add('ct-cancel-regalia')
```
rpherbig: ❓ What is the matching text for this Flag?

combat-trainer.lic
```diff
@@ -1287,6 +1358,30 @@ class SpellProcess
     end
   end
 
+  def check_trader_magic(game_state)
+    return unless DRStats.trader?
+    if Flags['ct-starlight-depleted'] then
```
rpherbig: 💡 We typically omit the `then` from if-blocks

combat-trainer.lic
```diff
@@ -1287,6 +1358,30 @@ class SpellProcess
...
+      echo "----OUT OF STARLIGHT - DELETING STARLIGHT MAGIC----"
+      @buff_spells.delete_if { |spell| @buff_spells[spell]['starlight'] }
```
rpherbig: 💡 We typically use `reject!` over `delete_if`

combat-trainer.lic
```diff
@@ -1287,6 +1358,30 @@ class SpellProcess
...
+      @offensive_spells.delete_if do |spell|
+        if spell['starlight'] then true end
```
rpherbig: 💡 Could this be `@offensive_spells.delete_if { |spell| spell['starlight'] }`?

combat-trainer.lic
```diff
@@ -2623,7 +2721,7 @@ class GameState
   $ranged_skills = $thrown_skills + $aim_skills
   $non_dance_skills = $ranged_skills + ['Brawling', 'Offhand Weapon']
 
-  attr_accessor :mob_died, :last_weapon_skill, :danger, :parrying, :casting, :need_bundle, :cooldown_timers, :no_stab_current_mob, :loaded, :selected_maneuver, :cast_timer, :casting_moonblade, :casting_weapon_buff, :use_charged_maneuvers, :casting_consume, :prepare_consume, :casting_cfb, :prepare_cfb, :wounds, :blessed_room, :currently_whirlwinding, :whirlwind_trainables, :reset_stance, :charges_total, :casting_sorcery, :hide_on_cast
+  attr_accessor :mob_died, :last_weapon_skill, :danger, :parrying, :casting, :need_bundle, :cooldown_timers, :no_stab_current_mob, :loaded, :selected_maneuver, :cast_timer, :casting_moonblade, :casting_regalia, :casting_weapon_buff, :use_charged_maneuvers, :casting_consume, :prepare_consume, :casting_cfb, :prepare_cfb, :wounds, :blessed_room, :currently_whirlwinding, :whirlwind_trainables, :reset_stance, :charges_total, :casting_sorcery, :hide_on_cast, :starlight, :last_regalia_type, :swap_regalia_type
```
rpherbig: ❓ Where is `starlight` used?

combat-trainer.lic
```diff
@@ -1425,6 +1520,9 @@ class SpellProcess
         data['cast'] = "cast #{moon}"
       end
     end
+    
+    game_state.casting_regalia = true if data['abbrev'] == 'REGAL'
```
rpherbig: 💡 Could this be `game_state.casting_regalia = data['abbrev'] == 'REGAL'`?

combat-trainer.lic
```diff
@@ -1287,6 +1358,30 @@ class SpellProcess
...
+      Flags['ct-cancel-regalia'] = true
+    end
+    if game_state.casting_regalia
+      if Flags['ct-starlight-depleted']  # In some conditions, regalia can run out of starlight partway through creation.
+        @regalia_array[game_state.swap_regalia_type].each do |item| 
+          bput("remove my #{item}", "into motes of silvery", "Remove what?", "You .*#{item}", 'The leather gauntlets slide')
+        end
+        @equipment_manager.wear_equipment_set?('standard')
+      end
+      Flags['ct-starlight-depleted'] ? game_state.last_regalia_type = nil : game_state.last_regalia_type = game_state.swap_regalia_type
```
rpherbig: 💡 Could this be `game_state.last_regalia_type = Flags['ct-starlight-depleted'] ? nil : game_state.swap_regalia_type`?

combat-trainer.lic
```diff
@@ -1055,6 +1094,36 @@ class SpellProcess
     game_state.casting = false
   end
 
+  def check_regalia(game_state)
+    game_state.swap_regalia_type = nil if game_state.swap_regalia_type == game_state.last_regalia_type
+    return if game_state.casting
+    return unless game_state.swap_regalia_type
+    if @regalia_spell.nil?
```
rpherbig: ❓ What do you think about moving this so it only happens once? Or moving it to `validate.lic`? It feels pretty similar to the checks on line 38 (though it should belong to a different Process)

combat-trainer.lic
```diff
@@ -1055,6 +1094,36 @@ class SpellProcess
...
+      echo "Couldn't find a regalia waggle.  Using default values for the spell."
+      @regalia_spell = get_data('spells').spell_data['Regalia']
+    end
+    
+    case game_state.swap_regalia_type
+    when 'Light Armor'
```
rpherbig: 💡 This would make another good error-check (whether in this Process or in `validate.lic`)

combat-trainer.lic
```diff
@@ -1287,6 +1358,30 @@ class SpellProcess
...
+        if spell['starlight'] then true end
+      end
+      Flags['ct-cancel-regalia'] = true
+    end
+    if game_state.casting_regalia
+      if Flags['ct-starlight-depleted']  # In some conditions, regalia can run out of starlight partway through creation.
+        @regalia_array[game_state.swap_regalia_type].each do |item| 
+          bput("remove my #{item}", "into motes of silvery", "Remove what?", "You .*#{item}", 'The leather gauntlets slide')
+        end
+        @equipment_manager.wear_equipment_set?('standard')
```
rpherbig: ❓ Should we retreat first? I thought wearing armor had roundtime

combat-trainer.lic
Tarhish: Thanks for the reviews!

default_armor_type was already called in armor_hysteresis when I started looking at combat-trainer.  To my eyes its intent was to define a skill for the NORMAL armor cycling functions that it would revert to when all the armors had been trained to intended levels.  Eg. once all four skills are above the threshold in armor_hysteresis it swaps to your previously defined 'light armor' set or whatever.

HOWEVER, although the logic was already in place it was not initialized.  There was no declaration for @default_armor = settings.default_armor_type or anything, so it didn't do anything, and as a result I don't think armor_hysteresis was working exactly as intended.  I added the missing declaration into the initialize portion, added it to base.yaml, and tested it.  Seems to work for normal armor swapping and definitely works for regalia.  My default armor type is Brigandine, which I have crafting techs for.  So once it trains all four up to max it then swaps to my brig regalia set and squats on it until one of the others drop below the training threshold.

I have no idea why it wasn't declared when I got there.  I suspect it was just a little mistake, since it seems to work just fine now that I've turned it on.  If nobody declares it then armor_hysteresis will continue doing what it did before.

combat-trainer.lic
Tarhish: I struggled with this one, and now I see a better way to do it.  The existing logic for armor-swapping (and thus also regalia) is up in SetupProcess.  So when SpellProcess detects that you've run out of starlight and can't cast regalia anymore, it needs to signal SetupProcess to stop swapping around and switch to your standard equipment set.  To do that I manually toggled a flag, which seems stupid in retrospect.

Probably better to add another attribute to gamestate and have the armor logic check for that.  I'll make a change, remove the flag and its deletion.

combat-trainer.lic
Tarhish: Existing swap logic doesn't have you retreat, so I didn't add it, though you're not the first person to ask about it...  From testing I've found that retreating in a full set of plate usually means I just sit there cycling retreat until the regalia wears off five-ten minutes later, but then that's only with the ~80 ranks I gained testing with regalia.  Easy enough to add an option for both normal and regalia swapping.  Do you feel like it should be opt-in or opt-out?

combat-trainer.lic
Tarhish: I'll gladly add some stuff to validate.  I'd say most of your suggestions would probably work best over there.

combat-trainer.lic
rpherbig: That makes sense, thank you

combat-trainer.lic
rpherbig: 👍

combat-trainer.lic
Tarhish: It wasn't, and it has been removed.  Thanks for the catch.

combat-trainer.lic
Tarhish: Apparently it can be!

combat-trainer.lic
Tarhish: Changed since that makes more sense.

combat-trainer.lic
Tarhish: Now only checks once during initialize.  Also removed the warning and moved a check for it into validate.lic

combat-trainer.lic
Tarhish: Added a check to validate to throw an error if it contains a non-armor-skill.

combat-trainer.lic
Tarhish: I have it retreat twice immediately before any time it does the swap and it's behaving fine now.

combat-trainer.lic
```diff
@@ -27,11 +27,23 @@ class SetupProcess
...
-
+    @cycle_regalia = settings.cycle_armors_regalia
+    echo("  @cycle_regalia: #{@cycle_regalia}") if $debug_mode_ct
+    @default_armor = settings.default_armor_type
+    echo("  @default_armor: #{@default_armor}") if $debug_mode_ct
     @warhorn = settings.warhorn
     echo("  @warhorn: #{@warhorn}") if $debug_mode_ct
     @last_warhorn = Time.now - 305
 
+    if @cycle_regalia && @cycle_armors
```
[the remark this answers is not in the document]
Tarhish: I added a check to validate like you suggested, but I also kept a check and warning here.  I'm split, but I'd like to generate some feedback before playing with armor and  totally ignoring regalia settings.

combat-trainer.lic
Tarhish: Done and done.  Now uses game_state.regalia_cancel to trigger regalia_shutdown, which will remove everything, swap back to normal gear, and remove regalia from training.

combat-trainer.lic
Tarhish: Nice catch

combat-trainer.lic
Tarhish: Sort of.  It's surprisingly different from buff_spells for whatever reason.  Switched it over to reject but kept the format it needed to do its thing.

combat-trainer.lic
Tarhish: Looked into it and I see why -- change made

Tarhish:
In addition to comments addressed above, a couple more changes.  Some adjustments to how it handled regalia expiration and starlight depletion, and it feels a lot more sane about it.

Also, Trabe Chalice's default recast in base-spells set to -1.  TRC can't be recast when it's up unless you put more power into it than you started with.  It also sometimes fails to cast due to being cast right before full prep (always auto-fails unless fully prepared) but that's an issue for another time.

Sarvatt:
there is a pull request open to handle people lagging vs the server more than 600ms that could work for the trc thing, but it needs some help.

combat-trainer.lic
```diff
@@ -259,6 +284,22 @@ class SetupProcess
     game_state.update_whirlwind_weapon_info(offhand_skill)
     game_state.wield_whirlwind_offhand
   end
+  
+  def regalia_shutdown(game_state)
+    bput('retreat', 'You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat', 'retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing') if Flags['ct-engaged']
```
rpherbig: ❌ Please use `DRCT.retreat` in place of this `bput` (and the others in this file)

combat-trainer.lic
```diff
@@ -952,6 +1001,8 @@ class SpellProcess
     @tk_spell = @offensive_spells.find { |spell| spell['abbrev'] =~ /tkt|tks/i }
     drop_tkt_ammo
     Flags.add('need-tkt-ammo', 'There is nothing here that can be thrown')
+    
+    @regalia_spell = get_data('spells').spell_data['Regalia'] if (@regalia_array && @regalia_spell.nil?)
```
rpherbig: ❓ Why not do this when `@regalia_spell` and `@regalia_array` are initialized (around line 975). Then this only has to run once.

Tarhish: I put it down there since defining a regalia_spell was performing a similar role as the part of initialize defining a TK spell, which happens right next to it.  I can move it, but although it might not look like it on review it IS still inside SpellProcess' initialize so it only happens once anyway.

combat-trainer.lic
Tarhish: What you're seeing there is actually the bput match from DRCT.retreat.

As far as I can see, DR common-travel's retreat retreats until it fully leaves combat.  Problem is,  trying this with plate sticks me in a retreat loop that lasts for 10-15 minutes until regalia expires naturally, even back-training against mobs half my rank.  Brigandine has slightly less severe problems, but can still lose a couple minutes against mobs at my level. I was unable to find a common function that was more forgiving.

I'm on board with adding a gamestate helper function for the switchoff to reduce the number of times this is repeated, but if it's considered bad form to do a retreat that accepts failure then I may either remove retreating entirely or add a new setting for retreating before swapping.  I'll think about it a bit first unless someone knows of an existing common function.  

Maybe have that function retreat twice for brig and plate but use DRCT.retreat for light and chain?  Is that too messy?

rpherbig: As far as I'm aware, if you want to retreat, you generally want to fully retreat.

I'm curious about this plate/brig situation. Are you saying the first retreat is easier than the second in plate?

Tarhish: So my exact problem is this:  Picture me with ~120 ranks in just about anything defensive/offensive/armors.  I have light encumbrance from the plate, but a lot of hindrance.  I'm testing this in, say endrus_serpents, so I'm very overpowered for the scenario.  I'm wearing full plate and it's time for CT to swap to, say, training Light armor.  The plate has 8 minutes left on it, so I need to retreat and remove the existing Regalia, before I can cast the new Regalia in its place.

At that moment I start retreating and I get stuck in retreat -> 'You try to retreat but are unable to get away.'  I never do anything else except spam retreat every 3 seconds, and I don't make it out past pole range for the next 8 minutes until the plate expires and disappears, at which point I dodge, buck naked, right out of combat and switch to the next regalia.  It's a poor and inefficient way to spend a hunt.  To be fair, I think if Traders were meant to wear full-plate they'd have made us Paladins, but the allure of TDPs is too darn high.

With the 'just try twice' method, I get out of combat wearing chain or light armor, but with brigandine I may only get out to pole range.  It's fine though, there's no RT for removing Regalia pieces, so it just spams through them, and the cast time for Regalia is only like 6 seconds or something.

Keep in mind, preexisting armor-swapping in combat-trainer doesn't retreat at all.  Even if you're swapping out everything on your body it just sits in combat while it gets naked.

I can solve this in two obvious ways.  One is to destroy the armor before retreating (this has no RT, so can spammed without any delay and immediately followed by retreat), then zip out of combat naked with no hindrance, and prepare+cast the regalia spell (~6 seconds?) before resuming the fight. The other way is to handle retreating like I'm doing now, (try it a couple times then give up) or by making a few adjustments to handle plate specifically.  The first option is neater so I'm leaning towards that.  I will poll some other traders to see what they want -- I just don't want to do something stupid in the code.

rpherbig: I can see you've done your due diligence and I trust you. Talk to the other traders and let me know what you feel is best.

Tarhish: Thanks Robert, and this has been very helpful.

After looking closer I was able to determine I was wrong about the order of events, and swapping cast orders was one of the commits I made mid-update.  None of the regalia swaps/refreshes cause any RT, so there's only one case where there's any RT between swapping sets, at the very start when it puts on your regalia gearset.  I've adjusted that step to use DRCT retreat and the others have had the retreats removed and all call a common function to break their regalia.  I think that's the best of both worlds.

I want to run it through a couple hunting cycles as per usual, but after that I'll submit the changes again.  Thanks again for the advice.

Tarhish:
Pause this -- I just got feedback from someone without Bespoke Regalia and the syntax doesn't work without the meta-spell.  I want to send it out with that working.

Tarhish:
After the most recent changes and some back and forth testing with Maltaiche I think this is actually finally ready to go.  Non-bespoke casting is handled now, and a problem I never ran into where transitioning between gear and regalia while wielding a worn weapon caused the weapon to get worn and not wielded again was fixed.

I'll keep looking into this, cause I can think of a couple improvements, but I'm now satisfied that the kinks have been hammered out.  I ran the combat-trainer with a couple of other classes and didn't see any oddities either.

Tarhish:
Alright, changes made and tested in all the edge cases we had problems with -- armor swapping without regalia still works, which I test everytime I modify that logic -- We've run it through everything I can think of, and the flag I added should catch a lot of edge problems we didn't run into.  This should be ready for real this time.

combat-trainer.lic
```diff
@@ -1395,13 +1395,14 @@ class SpellProcess
...
       if Flags['ct-starlight-depleted']  # In some conditions, regalia can run out of starlight partway through creation.
-        game_state.shatter_regalia(@regalia_array[game_state.swap_regalia_type])
+        DRCA.shatter_regalia(@regalia_array[game_state.swap_regalia_type])
         @equipment_manager.wear_equipment_set?('standard')
         @equipment_manager.wield_weapon(game_state.weapon_name, game_state.weapon_skill)
       end
       if Flags['ct-regalia-succeeded']
         game_state.last_regalia_type = Flags['ct-starlight-depleted'] ? nil : game_state.swap_regalia_type
         Flags.reset('ct-regalia-expired')
+        $worn_regalia = @regalia_array[game_state.swap_regalia_type] # Regalia doesn't show up in Active Spells, so it must be tracked to be destroyed before_dying
```
rpherbig: ❓ Can you `release` the spell instead of removing each piece of armor?

Tarhish:
No, that would be WAY too easy. Strangely, it doesn't show up on Perc or Active Spells, it can't be released, and the only way to get info on it is to FOCUS each piece with an RT.  And casting the spell again doesn't refresh it either, it just fails.

The way I'm handling shattering the armor before-dying feels a bit sloppy, but something need to do it.  I think the next big change would be making regalia removal dynamic rather than predefined.  Matching all the regalia pieces in inv combat and breaking them without any input.

Supposedly we're going to get another tester tonight which may be illuminating.

Tarhish:
Okay, making regalia handling dynamic was a lot easier than I thought it would be.  I can completely get rid of defining the individual pieces in cycle_armors_regalia, turn it into an array of just armor skills you want to use, get rid of that dumb global variable, support prebuffing with regalia and simplify the code. On hold again.

Sarvatt:
thank you so much for reviewing the code after people tested it and not just getting stuck in your initial way to do it. That's what I do assuming I'd fix it later.

Tarhish:
So I have significant changes to this.  I have a parse_regalia which gives the list of currently worn regalia nouns for armor beginning with 'rough-cut crystal', 'faceted crystal' or 'resplendent crystal' (thanks Aurayn).  shatter_regalia? parses your armor and breaks any regalia you're wearing, and this lets me accept a regalia on combat-trainer startup and break it down on exit.  I'm putting these in DRCA because I have hopes I'll be able to use this elsewhere.

So as it is, the setup has been reduced down to a regalia gearset and one new setting.
```
cycle_armors_regalia:
  - Light Armor
  - Chain Armor
  - Brigandine
  - Plate Armor
```

I also separated the function of swap_armors into the original swap_armors and swap_regalia, and split armor_hysteresis (also got default_armor_type working properly for cycle_armors).  Having them together made sense to me at first, but it just needlessly complicates things imo.  It's easier to show no unintended impacts this way.

I've tested with other guilds, with regalia on and off, with cycle_armors instead of regalia, and I don't see anything odd.  No other guilds see any regalia-related messaging.

I wanted to submit it this way, but since I've undone a lot of changes and base-spells and base.yaml have changed since I started, for ease of review should we close this request and start a new one?  It'll be a heck of a lot easier to parse.

rpherbig:
I think it would be easier to review with a new PR. I would also _highly_ recommend _two_ new PRs. One for the `common-*` changes and one for any `*.lic` changes.

There's a subtlety with how changes to `common-*` files are picked up by people running multiple characters that means we really to let `common-*` changes get merged in and wait 2-3 days before merging in `*.lic` changes that require them.

PR #3093, opened 2018-08-26
bescort.lic: Fix up mriss balloon.

Sarvatt:


Sarvatt:
I changed the map to use bescort and this is an obvious fix so it would be appreciated if it could be fast tracked in :)

PR #3096, opened 2018-08-27
dependency.lic: Add ruby patchlevel to the output and move the yaml c…

Sarvatt:
…ontents above the profile and data listings.


Sarvatt:
example of it

https://pastebin.com/kPs5kEMM

PR #3057, opened 2018-08-11
This is my start to adding observations and predictions in combat

aruthas:
Assumptions:
- PG is up
- training process Astro: 220

I took existing astro training process and edited it

1) having issues when hurt cant observe so spams
2) needs flag for when able to observe and combine with Astro-- No reason to do a observation, when it doesnt help
3) need a better coder, to help clean it up??

PR #3097, opened 2018-08-28
Combat astrology and walking astrology support

BastionWT:
This pull request includes two things:
1. Changes to combat-trainer.lic
2. Adding walkingastro.lic

Combat-trainer changes include repurposing the training_ability Astro: XXX to observe either the Sun (day) or the Heart constellation (night) on that timer. It will continue to do so until the Survival prediction pool is full (observation messaging), at which point it will check the Astrology skill's exp. If under 25, it will align to the Survival skill designated in the user's setup.yaml (astrology_prediction_skills: -> Survival), or default to "survival" as the base.yaml setting and predict the future.

Wakingastro does the same things as mentioned above, but outside of combat, similar to the newly included ;play script. It is meant to be run as a background process, and will time out intentionally when certain scripts are in-process.

Both scripts assume the usage of a telescope. Both scripts assume PG is up. Walkingastro assumes the usage of a telescope case. The latter will be adjusted in a subsequent pull.

Wanted to get this out there for further testing. In the last day that this has been running, I've had to run ;astrology a single time. combat-trainer and walkingastro have kept Astrology moving constantly.


BastionWT:
Big kudos to Aruthas, Jonas, Byd, and Leustyin for their assistance in this. Aruthas put together the framework in his other pull request on this topic, I just repurposed it. Byd helped nail down the tiny syntax issues on why things weren't working here/there. Leus did the same, and offered scripting advice along the way. Thanks guys!

rpherbig:
Is this intended to replace #3057 ?

cc @aruthas

BastionWT:
An alternate for consideration, @rcuhljr. This is a fair bit more streamlined with less roundtime, less flagging, and less overall management while serving the same purpose...keeping astrology going at all times.

Sarvatt:
Aruthas agrees this iis a bettter one to persue after talking to him and he is not upset this is just overruling all the work he put into the other pull request. @BastionWT please remove all of the echo's, or make them hidden behind `if $debug_mode_ct`, like `echo "Getting telescope" if $debug_mode_ct`

Sarvatt:
for future reference though, this pull request should have just been walkingastro.lic, the other PR could have gone in and you could have adjusted the combat-trainer part of that after the fact.

BastionWT:
Will definitely remove the echoes after the testing phase is done. Msaia and others are still finding issues. :)

combat-trainer.lic
```diff
@@ -1998,6 +2013,177 @@ class TrainerProcess
...
+      quick_retry
+    when "You are already holding that."
+      determine_time
+    end
+  end
+
+  def determine_time
+    echo "Determining time"
+    case DRC.bput("time","it is dawn", "morning", "midday", "noon", "afternoon", "dusk", "sunset", "evening", "midnight", "night", "almost dawn", "approaching sunrise")
+    when "it is dawn"
```
rpherbig: 💡 You could merge a lot of these `whens`:
```ruby
case ...
when "it is dawn", "morning"
    day_center
when "sunset", "evening"
...
```

combat-trainer.lic
```diff
@@ -1998,6 +2013,177 @@ class TrainerProcess
...
   end
 
+  def check_heavens(game_state)
+    @equipment_manager.stow_weapon(game_state.weapon_name)
+    echo "Getting telescope"
+    case DRC.bput("get my telescope","You get a", "You need a free hand to pick that up.", "You are already holding that.")
+    when "You get a"
+      determine_time
+    when "You need a free hand to pick that up."
+      quick_retry
```
rpherbig: ❓ Where is `quick_retry` defined?

rpherbig:
❌ Could you review the comments I made on #3057 ? I see a lot of the same issues here.

combat-trainer.lic
BastionWT: quick_retry should not have been included in the combat-trainer version. That's left over from walkingastro.lic. Will fix.

BastionWT:
> Based on the logs we've seen, I think you need to put this in a loop of some sort and handle re-retreating.

Retreating with the observation, peering, and aligning is not needed. Retreating to predict is. Adding that loop in.

> This can't use DRC.get_settings because the user could have passed args in to load YAML other than the default. It needs to use settings

This setting is going away. It is no longer necessary.

>  Is this Flag used anywhere?

This flag has been removed. It was not being used.

>  I believe delete is case-sensitive. This may need to be 'Astro'.

Not present in this version of the script.

The latest should be committed. I'm running through testing on the latest set of changes now.

jandersson:
You need to remove the combat astrology from this. That is unacceptable.

PR #3098, opened 2018-08-28
sanowet-crystal.lic: Make using it in combat configurable.

Sarvatt:
Some people who use lots rituals do not want to use it, other people like thieves do so make that configurable.

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -13,6 +13,7 @@ class SanowretCrystal
     ]
     @invalid_rooms = []
     @worn_crystal = false
+    @sanowret_during_combat: get_settngs.use_sanowret_during_combat
     args = parse_args(arg_definitions)
     check_crystal if args.run && !hiding? && !invisible?
     passive unless args.run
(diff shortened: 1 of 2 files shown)


sanowret-crystal.lic
```diff
@@ -13,6 +13,7 @@ class SanowretCrystal
     ]
     @invalid_rooms = []
     @worn_crystal = false
+    @sanowret_during_combat: get_settngs.use_sanowret_during_combat
```
rpherbig: ❌ `get_settings`

sanowret-crystal.lic
```diff
@@ -37,7 +38,8 @@ class SanowretCrystal
     return if DRStats.concentration < 100
     return if DRSkill.getxp('Arcana') >= 25
     return if hiding? || invisible?
-    return if Script.running?('sew') || Script.running?('carve') || Script.running?('forge') || Script.running?('remedy') || Script.running?('shape') || Script.running?('outdoorsmanship') || Script.running?('combat-trainer')
+    return unless @sanowret_during_combat && Script.running?('combat-trainer')
```
rpherbig: ❓ Are you sure this is the right logic? If I did my math right, this only runs the crystal logic if `sanowret_during_combat ` is true _and_ `combat-trainer` is running.

I expected something like `return if !@sanowret_during_combat && Script.running?('combat-trainer')` instead

Sarvatt:
note to self: don't commit things before coffee :)

PR #3099, opened 2018-08-28
smartlisten.lic: Allow listening to any magic type unless you are a b…

Sarvatt:
…arb or thief, it teaches sorcery.

### DIFF
--- a/smartlisten.lic
+++ b/smartlisten.lic
@@ -10,23 +10,25 @@ no_kill_all
 
 listen_skills = get_settings.listen_skills
 
+listen_skills += ['Life Magic', 'Elemental Magic', 'Lunar Magic', 'Holy Magic'] unless DRStats.barbarian? || DRStats.thief?
+
 if DRStats.barbarian?
   listen_skills += ['Expertise', 'Inner Fire']
-  listen_skills -= ['Arcana', 'Life Magic', 'Elemental Magic', 'Lunar Magic', 'Holy Magic']
+  listen_skills -= ['Arcana']
 elsif DRStats.bard?
-  listen_skills += ['Bardic Lore', 'Elemental Magic']
+  listen_skills += ['Bardic Lore']
 elsif DRStats.cleric?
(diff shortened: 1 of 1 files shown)


smartlisten.lic
```diff
@@ -10,23 +10,25 @@ no_kill_all
...
 elsif DRStats.necromancer?
   listen_skills += ['Thanatology', 'Arcane Magic']
 elsif DRStats.paladin?
-  listen_skills += ['Endurance', 'Holy Magic']
+  listen_skills += ['Endurance']
 elsif DRStats.ranger?
-  listen_skills += ['Scouting', 'Life Magic']
+  listen_skills += ['Scouting']
 elsif DRStats.thief?
   listen_skills += ['Backstab', 'Inner Magic']
```
rpherbig: ❓ Is this still called Backstab? I thought it was renamed

Sarvatt: sure is, its referred to in game as blindside but the skill is backstab still

smartlisten.lic
```diff
@@ -10,23 +10,25 @@ no_kill_all
 
 listen_skills = get_settings.listen_skills
 
+listen_skills += ['Life Magic', 'Elemental Magic', 'Lunar Magic', 'Holy Magic'] unless DRStats.barbarian? || DRStats.thief?
```
rpherbig: ❓ Does listening to a cross-magic-type, which teaches Sorcery, cause problems in town?

Sarvatt: no, only if you listen to a sorcery class. listening to elemental magic as a life magic user teaches sorcery instead of elemental magic without consequence

Sarvatt: I have been doing this for over a year inadvertently since life magic is included in the default listen_skills: in base.yaml for everyone, that's why I was surprised it was only life magic being listened to.

rpherbig: 💡 I think we should probably remove Life Magic from `base.yaml`

PR #3100, opened 2018-08-28
dependency.lic: make help_me pastes expire and use yaml formatting.

Sarvatt:
example: https://pastebin.com/H5MXKHbi

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -528,7 +528,7 @@ class ScriptManager
     http.verify_mode = OpenSSL::SSL::VERIFY_NONE
 
     request = Net::HTTP::Post.new(uri.request_uri)
-    request.set_form_data(api_dev_key: @paste_bin_token, api_paste_code: message_body, api_paste_private: '0', api_paste_format: 'yaml', api_paste_expire_date: '1W',  api_option: 'paste')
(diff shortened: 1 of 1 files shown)


dependency.lic
```diff
@@ -528,7 +528,7 @@ class ScriptManager
     http.verify_mode = OpenSSL::SSL::VERIFY_NONE
 
     request = Net::HTTP::Post.new(uri.request_uri)
-    request.set_form_data(api_dev_key: @paste_bin_token, api_paste_code: message_body, api_paste_private: '0', api_option: 'paste')
+    request.set_form_data(api_dev_key: @paste_bin_token, api_paste_code: message_body, api_paste_private: '1', api_paste_format: 'yaml', api_paste_expire_date: '1D',  api_option: 'paste')
```
rpherbig: ❓  1 day seems a bit short since people often link them in Issues/PRs... thoughts?

Sarvatt: who links them besides me? other option is a week, that works for me too. someone just expressed that they'd prefer a day when i asked. Just bumped it back up to a week :)

Sarvatt:
mainly I don't want our ;e help_me stuff lingering forever which may have private info in it.

Sarvatt:
since you can search pastebin to find out peoples info even if they do a private ;e help_me('person')

PR #3107, opened 2018-08-30
Walkingastro

BastionWT:
;walkingastro.lic is designed to observe either the Sun (day) or the Heart constellation (night) on a 215 second timer. It will continue to do so until the Survival prediction pool is full (observation messaging), at which point it will check the Astrology skill's exp. If under 25, it will align to the Survival skill designated in the user's setup.yaml (astrology_prediction_skills: -> Survival), or default to "survival" as the base.yaml setting and predict the future.

It is designed for out of combat use, similar to the newly included ;play script. It is meant to be run as a background process, and will time out intentionally when certain scripts are in-process.

This script assume the usage of a telescope and that PG is up.


BastionWT:
Any action on this one? Aruthas, Msaia, and I have been running it without issues for the past ~4 days since the bput changes went in.

walkingastro.lic
```diff
@@ -0,0 +1,210 @@
+custom_require.call %w[events]
```
rpherbig: 💡 Needs a header that links to elanthipedia

walkingastro.lic
```diff
@@ -0,0 +1,210 @@
...
+        day_center
+      end
+      observe
+      return
+    end
+  end
+
+  def night_center
+    echo "Night center"
+    fput("center telescope on heart")
```
rpherbig: ❔ Why 'heart' in specific?

walkingastro.lic
```diff
@@ -0,0 +1,210 @@
...
+      waitrt?
+      pause 1
+      fput("predict future")
+      waitrt?
+      pause 1
+    end
+  end
+
+  def can_observe?
+    return !['buff', 'steal', 'combat-trainer', 'pick', 'craft', 'shape', 'sew','bescort', 'remedy', 'forge', 'carve', 'performance', 'theurgy', 'hlctheurgy', 'astrology', 'astrology2', 'study-art', 'mech-lore', 'first-aid'].any? { |script| Script.running?(script) }
```
rpherbig: I'm not a fan of this approach... this list doesn't seem exhaustive nor easy to maintain

rpherbig:
Looks ok to me. @jandersson can you glance it over?

rpherbig:
> Any action on this one? Aruthas, Msaia, and I have been running it without issues for the past ~4 days since the bput changes went in.

Sorry, the holiday weekend kept me busy.

walkingastro.lic
BastionWT: Righto, will do.

walkingastro.lic
BastionWT: Heart is the only Survival night time constellation available all year, with no rising/setting during the night at unspecified times. It, like the sun during the day, is always available at night. As an added bonus, it fills the same pool that the sun does, which allows for the pool to fill in a timely manner for predictions.

walkingastro.lic
BastionWT: Can you recommend an alternative? I literally harvested from the ;play script that is being used. Basically, I want walkingastro to not run during any script that can fill your hands with 1 or 2 objects, as a telescope needs both hands to center and peer.

walkingastro.lic
rpherbig: 👍 thank you for the explanation

walkingastro.lic
rpherbig: I'll have to ponder it. Since this is a new script, people are effectively opting-in to its approach, so this is fine for now.

PR #3112, opened 2018-09-01
Town/Creature organization

ServerFailure:
Sorted creatures by town. (Theren only so far)
Tagged each creature with adjacent town.
Added anchors to not break existing yaml files.


ServerFailure:
Don't use this.

PR #3119, opened 2018-09-05
Walkingastro updates

BastionWT:
Update removes the echoes and adds documentation link in the header.


walkingastro.lic
```diff
@@ -1,3 +1,7 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#walkingastro
```
rpherbig: 💡 It's a personal preference, but I prefer to break up words when possible, e.g. `walking-astro`

PR #3132, opened 2018-09-16
base-hunting.yaml: Fix up cave trolls after the map was updated.

Sarvatt:
19203 doesn't exist anymore, and 19196 was missing.


Sarvatt:
whoops, didn't see the other PR.

PR #3140, opened 2018-09-20
Add 'swimming' gearset support for faldesu river 140 - 300 athletics

bmcneff:


Sarvatt:
is this ok if you don't have a swimming gear_set? We really need to just make a variable of EquipmentManager.new at one point in this script with how much it is used. Not saying that in relation to these changes, just at how much everything is making new instances of it.

bmcneff:
It works perfectly fine without a 'swimming' gearset. My thief has 294 athletics, is based in the Crossing, but has a hunt in goblin reivers. He makes the swim just fine without a swimming gearset.

PR #3149, opened 2018-09-21
Redirect Hibarnhvidar non-forgers to Shard in base-crafting

Tarhish:
Added a reference to redirect anyone trying to do remedies, tailoring, shaping, or carving in Hibarnhvidar over to Shard since there is only a Forging Society there.

### DIFF
--- a/data/base-crafting.yaml
+++ b/data/base-crafting.yaml
@@ -225,7 +225,7 @@ blacksmithing:
     stock-name: bronze ingot
     trash-room: 10767
     idle-room: 10765
-  Hibarnhvidar:
+  Hibarnhvidar: &forfedhar_blacksmithing
     npc: Juln
     npc_last_name: Juln
     npc-rooms:
@@ -310,7 +310,8 @@ blacksmithing:
     stock-number: 11
     trash-room: 8813
     idle-room: 8808
-
+  Boar Clan:
+    << : *forfedhar_blacksmithing
 tailoring:
   Crossing: &zoluren_tailoring
     npc: Milline
@@ -485,6 +486,8 @@ tailoring:
     idle-room: 11251
   Hibarnhvidar:
     << : *ilithi_tailoring
+  Boar Clan:
+    << : *ilithi_tailoring
 shaping:
   Crossing: &zoluren_shaping
     npc: Talia
@@ -626,6 +629,8 @@ shaping:
     idle-room: 11268
   Hibarnhvidar:
     << : *ilithi_shaping
+  Boar Clan:
+    << : *ilithi_shaping
 carving:
   Crossing: &zoluren_carving
     npc: Talia
@@ -731,7 +736,9 @@ carving:
     pattern-book: carving
     part-room: 11272
   Hibarnhvidar:
-    << : &ilithi_carving
+    << : *ilithi_carving
+  Boar Clan:
+    << : *ilithi_carving
 remedies:
   Crossing: &zoluren_remedies
     npc: Lanshado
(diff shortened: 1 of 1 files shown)


Sarvatt:
looks good to me, all references are after the declarations.

Sarvatt:
what do you think about making Boar Clan use them also?

Sarvatt:
people do use hometown: Boar Clan for some crazy reason.

Sarvatt:
i've tried to alias their crazy over to hib if they did in the other data files

jandersson:
I really like this idea

jandersson:
At some point we should do this more automatically.. e.g. just find the closest society via mapdb and look up the necessary stuff.

Tarhish:
That IS weird.  It's starting to get a bit silly, but I don't see why we wouldn't let them.  I can make a &forfedhar_forging for Hibarnhvidar and alias them to that.  Then send the rest of them over to Shard.

Tarhish:
Definitely some examples of automatically handling that sort of thing in trade.lic

jandersson:
<3

jandersson:
safe-room.lic does it too, to find the closest PC empath to you. Its all just leveraging `sort_destinatinations` which is in common-travel (i believe) 
```
  def find_closest_id(entity)
    rooms = @town_data.to_h.values.select { |data| data[entity] }
                      .map { |data| data[entity]['id'] }
    sort_destinations(rooms).first
  end
```

Sarvatt:
it really would be good to do that generically, instead of relying on your hometown: setting go to the closest place. this is just spitballing though, nothing related to this PR which is 100% ok as it is to go in.

Tarhish:
Caught a mistake on the re-review, bleh.

Also, I just looked this over and if I'm reading it right I think any recipe that uses parts wouldn't have a part-room for aliased towns.  @recipe_parts[part][@hometown] would just pull from a nonexistent entry. 

That's easy to add for Leth, Boar, and Hib, but another point in favor of eventually looking over making it generic

Tarhish:
Yeah, Leth already does this.  Okay, good to know -- I'll add those in for Hib and Boar Clan

Tarhish:
So a *ilithi alias had been called with a & and that's fixed.

I gave my best effort at putting the id for each part room in the city it's most likely to be used in -- commonly used forging parts go to Hib, stuff mostly used in outfitting/eng points to Shard, and I commented whenever a part has you travel out-of-city for later reference.  Ultimately there's gonna come a time when someone tries to forge something in Hib and takes a brisk stroll down to Shard for a part, and that's just going to be the price they pay for a non-standard configuration.

Sarvatt I'm really seeing what you were talking about re: Forfedhar and weirdoes.

Tarhish:
Also Shard was missing a couple part-rooms.  They got added in as well.

jandersson:
Just a general question: Is YAML documentation? base.yaml describes settings as if it were. This does not. Most people dont know what what <<:*hello_world means. Is that something worth addressing?

Tarhish:
A certain someone got themselves into such a mess with aliasing yesterday that I basically rewrote their yaml for them.  I'm not sure if the answer is to explain it or just not tell anyone about it until it's clear they get what it's there for.

Counter-question -- If I ever feel like adding MORE documentation to base.yaml would that be okay, or should we add it to, like, a tutorial page?

jandersson:
You wont hear a peep from me for adding documentation to anything. I love it. I just wanted to point out how unreadable YAML files become once we start adding anchors to them. Not at a stab at your contribution, just pontificating on how its weird we have documentation in so many places. Pull requests probably arent the place for pontificating....

Looks good! 👍

rpherbig:
I love everything about this PR.

PR #3155, opened 2018-09-24
Add functions to common-arcana to support regalia

Tarhish:
Adding two functions to common-arcana.

parse_regalia takes inv armor and returns the nouns for those that match T1, T2, or T3 regalia pieces.

shatter_regalia? takes a list of regalia nouns to remove or gets its own from parse_regalia.  Returns true if it found and destroyed a regalia, false otherwise.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -234,6 +234,34 @@ module DRCA
       .map { |_name, properties| properties['abbrev'] }
       .each { |abbrev| fput("release #{abbrev}") }
   end
+  
+  def parse_regalia  # generates an array of currently-worn regalia armor nouns
+    return unless DRStats.trader?
+    DRC.bput('inv armor', 'Type INVENTORY HELP for more options')
+    snapshot = reget(40)
+    if snapshot.grep(/All of your armor|You aren't wearing anything like that/).any? && snapshot.grep(/Type INVENTORY HELP/).any?
+      result = snapshot
+        .map(&:strip)
+        .reverse
+        .take_while { |item| !['All of your armor:', "You aren't wearing anything like that."].include?(item) }
+        .drop_while { |item| item != '[Type INVENTORY HELP for more options]' }
+        .drop(1)
+        .select { |item| item.include?('rough-cut crystal') || item.include?('faceted crystal') || item.include?('resplendent crystal')}
+    else
+      parse_regalia
+    end
+    result.each_with_index { |item, index| result[index] = DRC.get_noun(item)}
+  end
+  
+  def shatter_regalia?(worn_armor = nil) # takes an array of armor nouns to remove or gets its own from parse_regalia
+    return false unless DRStats.trader? 
+    worn_armor = parse_regalia unless worn_armor
+    return false if worn_armor.empty?
+    worn_armor.each do |item| 
+       DRC.bput("remove my #{item}", "into motes of silvery", "Remove what?", "You .*#{item}")
+    end
+    return true
+  end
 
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -234,6 +234,34 @@ module DRCA
...
+        .map(&:strip)
+        .reverse
+        .take_while { |item| !['All of your armor:', "You aren't wearing anything like that."].include?(item) }
+        .drop_while { |item| item != '[Type INVENTORY HELP for more options]' }
+        .drop(1)
+        .select { |item| item.include?('rough-cut crystal') || item.include?('faceted crystal') || item.include?('resplendent crystal')}
+    else
+      parse_regalia
+    end
+    result.each_with_index { |item, index| result[index] = DRC.get_noun(item)}
```
rpherbig: 💡 Is this the same as `result.map { |item| DRC.get_noun(item) }`?

common-arcana.lic
```diff
@@ -234,6 +234,34 @@ module DRCA
...
+  end
+  
+  def shatter_regalia?(worn_armor = nil) # takes an array of armor nouns to remove or gets its own from parse_regalia
+    return false unless DRStats.trader? 
+    worn_armor = parse_regalia unless worn_armor
```
rpherbig: 💡 Is `worn_regalia` a more descriptive name?

common-arcana.lic
```diff
@@ -234,6 +234,34 @@ module DRCA
...
+    if snapshot.grep(/All of your armor|You aren't wearing anything like that/).any? && snapshot.grep(/Type INVENTORY HELP/).any?
+      result = snapshot
+        .map(&:strip)
+        .reverse
+        .take_while { |item| !['All of your armor:', "You aren't wearing anything like that."].include?(item) }
+        .drop_while { |item| item != '[Type INVENTORY HELP for more options]' }
+        .drop(1)
+        .select { |item| item.include?('rough-cut crystal') || item.include?('faceted crystal') || item.include?('resplendent crystal')}
+    else
+      parse_regalia
```
rpherbig: ❓ Should this be `return parse_regalia`? As currently written, it looks like it will call `parse_regalia` but then exit the `else` and return the value of `result.each_...` on the next line.

rpherbig:
I'm going to merge despite the ❓ because it looks like an edge-case, but I'm still concerned about it.

common-arcana.lic
Tarhish: Yep, I think so.  I cribbed most of this from get_combat_items in equipmanager, though I wanted it to accomplish different things.  I see what I added later undermines that catch.

If I understand correctly I could change this easily when I'm making the other regalia changes, since it won't be changing the method declaration or what is returns.  No one will be inconvenienced if their common-arcana stays un-updated for a few days.

rpherbig: 👍

common-arcana.lic
Tarhish: Thanks, I knew there had to be a better way to do that.  I can move that into the if to become
```      
result = snapshot
        .map(&:strip)
        .reverse
        .take_while { |item| !['All of your armor:', "You aren't wearing anything like that."].include?(item) }
        .drop_while { |item| item != '[Type INVENTORY HELP for more options]' }
        .drop(1)
        .select { |item| item.include?('rough-cut crystal') || item.include?('faceted crystal') || item.include?('resplendent crystal')}
        .map { |item| DRC.get_noun(item)}
```

PR #3157, opened 2018-09-25
Trader starlight development + regalia support in combat-trainer

Tarhish:
Changes meant to support trader starlight, regalia armor swapping, and future development of trader magic.

COMMON-ARCANA:
Comments incorporated.  After removing the result.each_with it shouldn't require the 'return parse_regalia' if I understand it right.

starlight_messages added to perc_aura. Message added to the aura-growing flag to capture when your aura is still growing less than optimally.  Eventually I want to use all of these things in CT.  Right now this is not used.

BASE.YAML
values added for regalia variable and default_armor_type.
Should something be in base-empty?

BASE_SPELLS:
previously-planned starlight attribute replaced with starlight_threshold.  Currently this doesn't do anything beyond marking the spell for deletion when you run out of starlight, but it's being built for the future.  Meant to interact with DRCA.perc_aura with the idea being it'll update your aura occasionally and let you tag certain spells to only cast when you have an aura of that level or higher.

A value of -1 means it won't even be removed if you run out of starlight.

Also, Enrichment and its cast message added.

COMBAT-TRAINER:
IMO the critical points are line 91 and 2682.

2682 is what makes players wear their standard gear at the very start of combat.  Supporting starting combat-trainer with a prebuffed regalia required skipping this if you have regalia stuff set or you're wearing regalia.  My thief and cleric don't see any messages from this.

default_armor_type seems to be working now that it's declared up above.  Added a catch to make sure your default_armor_type is actually in your cycle_armors.

New functions in ct:
check_regalia_swap mimics check_armor_swap but tags game_state.swap_regalia_type for SpellProcess to cast regalia.

regalia_hysteresis mimics armor_hysteresis

validate_regalia runs once to catch a couple fatal errors that would make ct throw a fit, and captures a pre-existing regalia on start.

regalia_shutdown totally ends all regalia function and swaps to the standard gearset

check_regalia catches a game_state.swap_regalia or an expired regalia and calls prepare_spell to replace it.

check_trader_magic handles all starlight-related and trader magic stuff after a cast, including handling exhausted starlight and resetting regalia stuff.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1420,7 +1420,6 @@ class SpellProcess
       if Flags['ct-regalia-succeeded']
         Flags.reset('ct-regalia-expired')
         Flags.reset('ct-regalia-succeeded')
-        #game_state.last_regalia_type = Flags['ct-starlight-depleted'] ? nil : game_state.swap_regalia_type
         game_state.last_regalia_type = game_state.swap_regalia_type
         game_state.swap_regalia_type = nil
       end
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -27,11 +27,18 @@ class SetupProcess
     @last_cycle_time = Time.now - @cycle_armors_time
     @combat_training_abilities_target = settings.combat_training_abilities_target
     echo("  @combat_training_abilities_target: #{@combat_training_abilities_target}") if $debug_mode_ct
+    @cycle_regalia = settings.cycle_armors_regalia
+    echo("  @cycle_regalia: #{@cycle_regalia}") if $debug_mode_ct
+    @default_armor = settings.default_armor_type
+    echo("  @default_armor: #{@default_armor}") if $debug_mode_ct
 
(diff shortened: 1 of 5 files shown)


Tarhish:
The preexisting common-arcana changes will make it a heck of a lot easier for others to try things out, since the other files don't hang around in the background.  We can definitely take some time to review this.

Tarhish:
;validate.  I knew I was missing something.  I'll replace the ;validate checks tomorrow and add a new one for the starlight_threshold.

Tarhish:
I'm definitely confident in the testing, but I'm in no hurry.  I'll let some of the people on slack take a gander for another day or two and then give the signal, if that's alright.

rpherbig:
That sounds great

Sarvatt:
its specific to using non default settings so not many people wil use it
and hes on tech support, +1 from me :)

Tarhish:
I've got nothing left to check here, so whenever you're willing I'm thinking this can go live.

PR #3164, opened 2018-10-04
Doing sanowret correctly.

rcuhljr:


Sarvatt:
Working fine, added - play to the list and started sano and it didn't use it until play exited
```
>;play
--- Lich: play active.
[play]>play scales halt on zills
You fumble slightly as you begin a halting ruff on your copper zills.
>
--- Lich: sanowret-crystal active.
You continue playing on your copper zills.
>
You lose track of the time signature as you continue to play a ruff on your copper zills, causing it to be clumsy and ill wrought.
>
You continue playing on your copper zills.
>;k play
--- Lich: play has exited.
[sanowret-crystal]>tap my sanowret crystal
>```

Issue #3143, opened 2018-09-20
Combat trainer: Face next or analyze issue in warklins?


lbacker88:
Been running into this issue in warklins since I moved here.  Not sure if you need more information. 

Also - Had to remove whirlwind completely because it gets stuck in a broken loop trying to whirl with no weapons out after this bug occurs.

[combat-trainer]>loot
You search the armored warklin.
The warklin was carrying a huge covellite nugget!
>
[combat-trainer]>analyze flame
>
Analyze what?
>
[combat-trainer]>face next
You turn to face an armored warklin.
>
[combat-trainer]>
A shift in the breeze accompanies the arrival of an agitated warklin mauler.
A warklin mauler carefully picks its way into view.
>
The warklin mauler begins to advance on you!
The warklin mauler advances from nearby and is closing steadily.
>
* As if fumbling muscle flab were natural, an armored warklin lunges with its razorlike mandibles at you.  You repulse some of the mandibles with a steel bastard sword.
[You're winded, nimbly balanced and in good position.]
>
An armored warklin's body falls apart.
>
Lord General Reverence just arrived.
>
Lord General Reverence goes south.
>
The warklin mauler closes to melee range on you!
>
* Awkwardly, an armored warklin lunges with its razorlike mandibles at you.  You slap away some of the mandibles with a steel bastard sword.
[You're adeptly balanced and in very strong position.]
>
* Apparently without direction or thought, an armored warklin thrusts its deadly mandibles at you.  You knock aside most of the mandibles with a steel bastard sword.
[You're adeptly balanced and in strong position.]
>
* Looking as if this were a bad idea, a warklin mauler twists its torso wildly, thrusting its impalers at you.  You knock aside most of the impaler with a steel bastard sword.
[You're adeptly balanced and in better position.]
>

You feel fully rested.
>l
[Abandoned Mine, Tunnel]
Carts still filled with ore, rich veins in the rock face, pick axes and shovels all provide further evidence of a hasty departure by the miners that plied their trade here.  You also see an armored warklin, an armored warklin, a warklin mauler and a huge covellite nugget.
Obvious exits: south.
Room Number: 7277
>
* As if fumbling muscle flab were natural, an armored warklin snaps its mandibles at you.  You evade, stepping out of harm's way.
[You're nimbly balanced and in better position.]
>
* As if fumbling muscle flab were natural, an armored warklin lunges with its razorlike mandibles at you.  You beat off some of the mandibles with a steel bastard sword.
[...]

PR #3170, opened 2018-10-07
fix problem with empty analyze array on mob death

Tarhish:
This is related to issue https://github.com/rpherbig/dr-scripts/issues/3143

A condition can trigger whenever a mob is killed at the same time as a barbarian's analyze array is depleted and not replenished by perform_analyze? that causes an empty melee_attack_verb to be generated.

I don't believe there's an easy fix to this 'in-method' for perform_analyze?, since it would need to handle the states of being out of melee and needing to advance or all mobs being dead, and do it before a melee_attack_verb is decided.  So instead whereever update_analyze_array is used it will just default to 'attack' if it returns with an empty array.

Also, for somewhat more reliability, the face next in perform_analyze does try once more to turn on an enemy in melee, which should prevent this most of the time.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3069,7 +3069,7 @@ class GameState
   def determine_whirlwind_action
     if @use_weak_attacks && Flags['ct-accuracy-ready']
       update_analyze_array(true) if @analyze_combo_array.empty?
-      @analyze_combo_array.shift
+      @analyze_combo_array.empty? ? 'attack' : @analyze_combo_array.shift
     elsif (npcs.length > 2) && !balance_low?
       'whirlwind'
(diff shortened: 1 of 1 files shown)


Tarhish:
I'd like it if some actual barbarian tried this out and gave it an okay.

rpherbig:
I've not been experiencing this issue, but I'll run it on multiple characters (including a barb) for regression purposes.

xtremenerdrage:
Adding this to my combat-trainer and testing

xtremenerdrage:
Been running this for about 5 hours and its working great, I am getting the analyze what? message and it is either facing next or continuing the combo

PR #3174, opened 2018-10-08
Change Barbarian meditation activation messages

### LINKED ISSUE #3171: Barb Meditations
In base-spells.yaml the Barb Meditations have their activated messages of when you start the meditation. Without the Yogi Mastery, it still takes a few seconds to complete after the round time before you can stand up. So, can we change the Meditations from;

`activated_message: ^You .* to meditate`

to

`activated_message: feel a jolt as your vision snaps shut`

I've edited them all in the attached base-spells.yaml but I'm not quite sure how to get it added to the "base" so my file doesn't keep getting updated locally.
[base-spells.zip](https://github.com/rpherbig/dr-scripts/files/2454006/base-spells.zip)

Thanks!

rpherbig:
The Barbarian meditations previously had their activated messages of when you start the meditation. Without the Yogi Mastery, it still takes a few seconds to complete after the round time before you can stand up. Instead we should use the 'meditation active' message.

Closes #3171

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -301,27 +301,27 @@ barb_abilities:
   Contemplation:
     type: meditation
     start_command: med contemplation
-    activated_message: ^You .* to meditate
+    activated_message: feel a jolt as your vision snaps shut
     expired_message: contemplate enhanced defensive strategies drifts
   Bastion:
     type: meditation
     start_command: med bastion
-    activated_message: ^You .* to meditate
+    activated_message: feel a jolt as your vision snaps shut
     expired_message: bastion of strength slips from your mind
   Tenacity:
     type: meditation
     start_command: med tenacity
-    activated_message: ^You .* to meditate
+    activated_message: feel a jolt as your vision snaps shut
     expired_message: leaving you vulnerable to physical harm
   Serenity:
     type: meditation
     start_command: med serenity
-    activated_message: ^You .* to meditate
+    activated_message: feel a jolt as your vision snaps shut
     expired_message: leaving you vulnerable to magic
   Focus:
     type: meditation
     start_command: med focus
-    activated_message: ^You .* to meditate
+    activated_message: feel a jolt as your vision snaps shut
     expired_message: Focus meditation slips away from your mind
 
 spell_data:


Sarvatt:
Looks good to me, also we are missing 7 meditations here apparently

Sarvatt:
unrelated to this pull request and not blocking at all but

seek
staunch
prediction
unyielding
flame
power
dispel

are all missing from here, should they be added?

RyanEOD:
I would say please can we add them? I don't have most of those yet so I haven't tried using them, but eventually I'll be there.

rpherbig:
@RyanEOD did you find a Yogi barb to test this?

RyanEOD:
@rpherbig  - I have someone, just trying to fit into their schedule to get it tested out.

rpherbig:
@RyanEOD bump

BinuDR:
I looked at the code a bit more thoroughly and I think that flag I tried to add to the other method would work in the activate_barb_buff method.

`  def activate_barb_buff(name)
    Flags.add('med-done', /You feel a jolt as your vision snaps shut/)
    case DRC.bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand', 'You must be sitting')
    when 'You must be sitting'
      DRC.retreat
      DRC.bput('sit', 'You sit', 'You are already', 'You rise')
      activate_barb_buff(name)
    when 'You should stand'
      fix_standing
      activate_barb_buff(name)
    when 'Your inner fire lacks', 'find yourself lacking the inner fire'
      do something
    else
      if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
        pause 0.5 while !Flags['med-done']
      end
    end
    fix_standing
    Flags.delete('med-done')
  end`

BinuDR:
I would need barbarians to test it though, both with and without that Yogi mastery.  I'm going to be rolling up a barb to do some tests on my own as well

jayRyan24:
From my experience, you can feel a jolt, stand, and still mess up your meditation.

jayRyan24:
It actually used the message and then a separate meditation timer, because it doesn't seem to be an exact science.

jayRyan24:
#3581

PR #3177, opened 2018-10-10
add barb combo vars

xtremenerdrage:
barb_accuracy: true
barb_damage: true
barb_balance: true
barb_flame: true

if none are present it uses the current system

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2929,10 +2929,10 @@ class GameState
     @barb_damage = settings.barb_damage
     echo("  @barb_damage: #{@barb_damage}") if $debug_mode_ct
 
-	@barb_balance = settings.barb_balance
+    @barb_balance = settings.barb_balance
     echo("  @barb_balance: #{@barb_balance}") if $debug_mode_ct
 
-	@barb_flame = settings.barb_flame
+    @barb_flame = settings.barb_flame
     echo("  @barb_flame: #{@barb_flame}") if $debug_mode_ct
 
     @skip_last_kill = settings.skip_last_kill
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2925,13 +2925,13 @@ class GameState
 
     @barb_accuracy = settings.barb_accuracy
     echo("  @barb_accuracy: #{@barb_accuracy}") if $debug_mode_ct
-	
+
     @barb_damage = settings.barb_damage
     echo("  @barb_damage: #{@barb_damage}") if $debug_mode_ct
-	
+
 	@barb_balance = settings.barb_balance
     echo("  @barb_balance: #{@barb_balance}") if $debug_mode_ct
-	
+
 	@barb_flame = settings.barb_flame
     echo("  @barb_flame: #{@barb_flame}") if $debug_mode_ct
 
@@ -3485,10 +3485,10 @@ class GameState
     if Flags["ct-balance-ready"] && @barb_balance
       perform_analyze?('balance', 600)
       ret = true
-	end
-	if @barb_flame
-	  perform_analyze?('flame', 0)
-	  ret = true
+    end
+    if @barb_flame
+      perform_analyze?('flame', 0)
+      ret = true
     end
     return ret
   end
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3081,7 +3081,7 @@ class GameState
   def determine_whirlwind_action
     if @use_weak_attacks && Flags['ct-accuracy-ready']
       update_analyze_array(true) if @analyze_combo_array.empty?
-      @analyze_combo_array.empty? ? 'attack' : @analyze_combo_array.shift
+      @analyze_combo_array.shift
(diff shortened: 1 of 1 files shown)


xtremenerdrage:
This also has Tarhish's changes in it, I was testing them and they do work for the Analyze what? issue so I left them in

xtremenerdrage:
weird I have it running I guess the tabs and spaces came from me copy and pasting sections together, thx for the catch.

It could probably be a better hash and would just replace the current system all together, I dont really know how to set up the hash stuff that's why I did them as independent vars

xtremenerdrage:
fixed the tabs, if anyone wants to change the vars into a list of options and replace the barb_combos: to just use what is listed that would be cool. In the meantime this should work :)

combat-trainer.lic
```diff
@@ -3069,7 +3081,7 @@ class GameState
   def determine_whirlwind_action
     if @use_weak_attacks && Flags['ct-accuracy-ready']
       update_analyze_array(true) if @analyze_combo_array.empty?
-      @analyze_combo_array.shift
+      @analyze_combo_array.empty? ? 'attack' : @analyze_combo_array.shift
```
rpherbig: ❓ Is it necessary to include code from the other PR? I'd prefer to make each change separately if possible.

combat-trainer.lic
```diff
@@ -3449,6 +3472,27 @@ class GameState
     true
   end
 
+  def update_analyze_array_combo
+    ret = false
+    if Flags["ct-accuracy-ready"] && @barb_accuracy
```
rpherbig: ❓ If, for example, two analyze types are ready, won't this try both?

combat-trainer.lic
```diff
@@ -3449,6 +3472,27 @@ class GameState
...
+      perform_analyze?('accuracy', 50)
```
rpherbig: ❓ Should this do something with the return value of `perform_analyze?`?

combat-trainer.lic
xtremenerdrage: I'll remove the other PR stuff out

combat-trainer.lic
```diff
@@ -3360,16 +3372,24 @@ class GameState
     Flags.reset('ct-damage-ready')
   end
 
+  def set_barb_balance_flag
```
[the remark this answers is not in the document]
rpherbig: Follow-up question: are any of these `set*flag` functions needed due to this line? https://github.com/rpherbig/dr-scripts/blob/master/combat-trainer.lic#L3439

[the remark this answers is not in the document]
xtremenerdrage: I added def set_barb_balance_flag because there were defs for the other flags, I dont actually see where its called though.

  def set_barb_damage_flag
    Flags.reset('ct-damage-ready')
  end

  def set_barb_balance_flag
    Flags.reset('ct-balance-ready')
  end

combat-trainer.lic
```diff
@@ -3360,16 +3372,24 @@ class GameState
...
+  
   def melee_attack_verb
     if currently_whirlwinding
       determine_whirlwind_action
     elsif attack_override
       attack_override
     elsif @use_weak_attacks
       brawling? ? 'gouge' : 'jab'
-    elsif @use_barb_combos && !offhand? && !brawling?
+    elsif @use_barb_combos && !offhand? && !brawling? && (@barb_accuracy || @barb_balance || @barb_damage || @barb_flame)
```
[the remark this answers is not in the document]
xtremenerdrage: The first one would be if you have any of the specific barb vars added, if this was to replace barb_combos entirely the second elsif would not be necessary

combat-trainer.lic
xtremenerdrage: if two analyzes are ready it will indeed do both and it wont repeat any analyze combos until one of the flags is ready again

currently it will run through each analyze repeatedly every time one is ready and if none are ready it will nonstop analyze flame

combat-trainer.lic
xtremenerdrage: This sends the perform_analyze values in update_analyze_array_combo instead of update_analyze_array, just doing each analyze specifically rather than shifting to each one

jayRyan24:
This is how I originally implemented it. To have a list of skills you can set. This was turned down and I made it into a list which auto triggered. What is the reasoning behind this update? Just for my edification?

combat-trainer.lic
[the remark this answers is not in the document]
jayRyan24: This appears two be now two separate systems which handle the same thing. I think we just need to choose 1 and go with it, what would be the purpose of having this?

jayRyan24:
OK, so this seems to be that people don't want the system to:

a) always do the barb combos if they can. An example would be that you don't want it to put up the damage buff, because you aren't trying to do more damage?
b) You don't want to always analyze flame?

Some back story is that originally I had a list, but this was changed due to feedback from others. I can definitely fix these issues. The way it's being handled in this PR isn't very elegant, and I can fix that if that is what the 2 issues are.

PR #3178, opened 2018-10-10
Spell spelling check in ;validate

Tarhish:
A number of tech problems have been coming up due to misspelled/miscapitalized spells in buff-spells or offensive_spells.  Especially problematic with buffs and cyclics.  Adding a check to validate to catch that.

I wasn't really sure whether this should be a warning or error but I tentatively went with warning.

```
>
[validate: WARNING:< Spell not found in base_spells 'Trabe chalice'! Check its spelling and capitalization.  >]
  WARNINGS:1 ERRORS:0
  All done!
```

### DIFF
--- a/validate.lic
+++ b/validate.lic
@@ -678,6 +678,18 @@ class DRYamlValidator
             .any? { |skill| error("Skill name in cycle_armors_regalia is not valid: #{skill}") }
   end
 
+  def assert_that_non_training_spells_are_in_base_spells(settings)
+    return unless settings.offensive_spells || settings.buff_spells
+
+    spell_names = []
+    settings.buff_spells.each { |spell| spell_names.push(spell.first) }
+    settings.offensive_spells.each { |spell| spell_names.push(spell['name']) }
+    spell_data = get_data('spells').spell_data
+
+    spell_names.reject { |name| spell_data.keys.include?(name) }
+               .each { |name| warn("Spell not found in base_spells '#{name}'! Check its spelling and capitalization.") }
+  end
+
   private
 
   def warn(message)


rpherbig:
The validate tests are failing. It looks like you need to add a `spells` entry to https://github.com/rpherbig/dr-scripts/blob/master/test/test_validate.rb#L12

validate.lic
```diff
@@ -678,6 +678,18 @@ class DRYamlValidator
...
+  def assert_that_non_training_spells_are_in_base_spells(settings)
+    return unless settings.offensive_spells || settings.buff_spells
+
+    spell_names = []
+    settings.buff_spells.each { |spell| spell_names.push(spell.first) }
+    settings.offensive_spells.each { |spell| spell_names.push(spell['name']) }
+    spell_data = get_data('spells').spell_data
+
+    spell_names.reject { |name| spell_data.keys.include?(name) }
+               .each { |name| warn("Spell not found in base_spells '#{name}'! Check its spelling and capitalization.") }
```
rpherbig: 💡 What do you think about also outputting in the warning the name of the setting that needs fixed? e.g. `buff_spells` or `offensive_spells`

validate.lic
```diff
@@ -678,6 +678,18 @@ class DRYamlValidator
             .any? { |skill| error("Skill name in cycle_armors_regalia is not valid: #{skill}") }
   end
 
+  def assert_that_non_training_spells_are_in_base_spells(settings)
+    return unless settings.offensive_spells || settings.buff_spells
+
+    spell_names = []
+    settings.buff_spells.each { |spell| spell_names.push(spell.first) }
```
rpherbig: ❓ What do you think about also checking the other 'spell' settings? I'm looking at https://github.com/rpherbig/dr-scripts/blob/master/dependency.lic#L858-L901 for an example

rpherbig:
Warning seems fine since it _could_ still work (if the user specified enough of the spell's data by hand, so to speak).

validate.lic
Tarhish: I think that's a great idea for waggles, I just want to make sure I don't pull in khris and whatever.  Shouldn't be a problem.  I'm a little less sure about the need for it in all the training spells, since I don't even really think it harms anything there.  AFAIK crafting_training_spells shouldn't even care if the name is spelled wrong?

validate.lic
Tarhish: Sure, more info more better.

validate.lic
rpherbig: No worries, I didn't have a chance to look at each setting in detail. These two are better than what we had before, so I'm fine with keeping just these two.

Tarhish: No, I guess I'm thinking of it from the wrong angle.  If the name is screwed up then they probably won't have all the other stuff defined either and it won't pull from the defaults.  Yeah, I'll go back to the original and include everything.

Tarhish:
I'll look at this a little further to compare how those other testvalidate checks are being used and make it pass.

For future reference, though, the current output looks like:

```
--- Lich: validate active.
  Checking 77 different potential errors
[validate: WARNING:< training_spells contains the following spell abbreviations not found in data/base_spells, ["noum", "finesse", "nonchalance"] Check spelling and capitalization.  >]
[validate: WARNING:< crafting_training_spells contains the following spell abbreviations not found in data/base_spells, ["NOUm"] Check spelling and capitalization.  >]
[validate: WARNING:< waggle_sets contains the following spell names not found in data/base_spells, ["Normena", "Membrach's Tweed"] Check spelling and capitalization.  >]
  WARNINGS:3 ERRORS:0
  All done!
--- Lich: validate has exited.
```

rpherbig:
Neat, running the validate test caught a typo in base-spells. Thanks @Tarhish

Tarhish:
Hahah, we can crowdsource spelling like stealing data.  Everyone just gives their best guess as to how to spell Y'ntrel Sechra and we take the average.

rpherbig:
Haha ❤️

PR #3179, opened 2018-10-10
Use Gem::Version to check versions

jandersson:
Uses Gem::Version to simplify version checking. 

Check for non-fork versions of lich as well. Anyone on Ruby > 3.2 and not on the lich fork will have the correct $MODERN_LICH and $MODERN_RUBY vars set

### DIFF
--- a/bootstrap.lic
+++ b/bootstrap.lic
@@ -25,10 +25,13 @@ def remove_constant(symbol)
   end
 end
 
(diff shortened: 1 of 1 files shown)


jandersson:
Don't think so. As far as I understand this is just leveraging Gem version checking by making arbitrary version numbers and then comparing them.  http://ruby-doc.org/stdlib-2.0.0/libdoc/rubygems/rdoc/Gem/Version.html

PR #3187, opened 2018-10-19
dependency.lic: remove yaml formattting from help_me posts.

Sarvatt:
This has been throwing me off for the past few weeks, yaml formatting on pastebin sometimes gets whitespace wrong and its making valid yamls that look fine locally invalid on pastebin.


Sarvatt:
https://pastebin.com/FuA9jsQ3 is an example of a bad ;e help_me, it is deleting spaces before lines such as - d0 in it but it does work.

PR #3208, opened 2018-11-19
Attempt to grab ammo after shooting, as well as 'at foot' items

dr-leustyin:


dr-leustyin:
Probably should be in base-items.  I'm just trying to get something out the door to fix the issue.  More cleanup can be done later

dr-leustyin:
Is there a way to re-run the CI tests?  The only difference between the Pass and Fail test was the addition of equipmentmanager to ;afk, which I can't see as causing the failures in the test

rcuhljr:
Adding equipmanager will break the tests like that. Due to the fact that they're having to run in threads we can't tell why the child thread died (equipmanager doesn't exist because custom_require doesn't load anything during tests) so we just see no results from all the afk tests even though they're likely blowing up because `EquipManager.new` is blowing up because it doesn't exist.

dr-leustyin:
Any advice on how to fix these tests to not bomb?  Also, is there a way for me to execute the test locally?  I get errors why trying to kick off the test script:

```analyst@Development MINGW64 /c/lich-4.6.33/lich/dr-scripts (ammo)
$ ruby test/test_afk.rb
C:/Ruby200/lib/ruby/2.0.0/minitest/unit.rb:28:in `const_missing': uninitialized constant MiniTest::Test (NameError)
        from test/test_afk.rb:20:in `<main>'

analyst@Development MINGW64 /c/lich-4.6.33/lich/dr-scripts (ammo)
$ cd test/

analyst@Development MINGW64 /c/lich-4.6.33/lich/dr-scripts/test (ammo)
$ ruby test_afk.rb
test_afk.rb:4:in `load': cannot load such file -- test/test_harness.rb (LoadError)
        from test_afk.rb:4:in `<main>'
```

rcuhljr:
`rake` is the command to run the tests (From the root of dr-scripts).  The easiest fix is probably to define a dummy EquipmentManager class in the test and put empty methods in that are required by the new code in afk, which looks like is just `initialize` and `empty_hands`

rpherbig:
> Probably should be in base-items. I'm just trying to get something out the door to fix the issue. More cleanup can be done later

That's fine

rpherbig:
@dr-leustyin is this an emergency fix that can't wait on the unit test to be fixed?

PR #3215, opened 2018-11-19
sanowret.lic: Handle newer sanowret crystals

### LINKED ISSUE #3212: Sanowret-crystal doesn't check for "You tap an"
Some sanowret crystals use the 'You tap an' instead of 'You tap a' when tapping, and causes the script to miss it.

I have tested locally by changing all matches from "You tap a" to "You tap" and it works perfectly with both cases where it says 'You tap a' and 'You tap an.'

Sarvatt:
Closes #3212

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -44,8 +44,8 @@ class SanowretCrystal
     if @worn_crystal
       use_crystal
     else
-      case DRC.bput('tap my sanowret crystal', /^You tap (?:a|an) .*sanowret crystal inside your .*.$/, /^You tap a .*sanowret crystal.* that you are wearing.$/, /^You tap (?:a|an) .*sanowret crystal that you are holding.$/, 'I could not find what you were referring to.', 'Not here, ', 'a wave of Corruption magic')
-      when /You tap a .*sanowret crystal inside your/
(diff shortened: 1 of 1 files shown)


sanowret-crystal.lic
```diff
@@ -44,15 +44,15 @@ class SanowretCrystal
     if @worn_crystal
       use_crystal
     else
-      case DRC.bput('tap my sanowret crystal', /^You tap a .*sanowret crystal inside your .*.$/, /^You tap a .*sanowret crystal.* that you are wearing.$/, /^You tap a .*sanowret crystal that you are holding.$/, 'I could not find what you were referring to.', 'Not here, ', 'a wave of Corruption magic')
+      case DRC.bput('tap my sanowret crystal', /^You tap (?:a|an) .*sanowret crystal inside your .*.$/, /^You tap a .*sanowret crystal.* that you are wearing.$/, /^You tap (?:a|an) .*sanowret crystal that you are holding.$/, 'I could not find what you were referring to.', 'Not here, ', 'a wave of Corruption magic')
```
rpherbig: 💡 It looks like the `that you are wearing` case needs updated

sanowret-crystal.lic
```diff
@@ -44,15 +44,15 @@ class SanowretCrystal
...
       when /You tap a .*sanowret crystal inside your/
```
rpherbig: 💡 It looks like this case also needs updated

Sarvatt:
Thanks for catching those!

PR #3216, opened 2018-11-19
combat-trainer.lic: Add HE egg usage to warhorn logic and fix up cust…

### LINKED ISSUE #3214: warhorn settings copied or altered for HE 428 gift
This would be nice to have, since it works similar to the warhorn.

With the gift being the same for everyone, it could just be a true/false setting to enable it.

Another option, that would take more work, is to make the current warhorn settings more flexible. The changes i suspect would be, keep the noun setting, and add a response and action setting. I would think a worn setting would help, but for the response, it could just be a straight copy/paste or a partial grab of what it's trying to match. The action setting would be the action to take, since the egg you invoke it, while warhorns you blow the horn.

Sarvatt:
…om blown warhorn nouns

usage: warhorn: egg in your yaml.
Closes #3214

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -98,16 +98,20 @@ class SetupProcess
     return unless @warhorn
     return if Time.now - @last_warhorn < 300
 
-    game_state.sheath_whirlwind_offhand
+    game_state.sheath_whirlwind_offhand if DRStats.barbarian?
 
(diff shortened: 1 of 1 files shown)


Sarvatt:
Tested with an egg by doing ;combat-trainer d4 with warhorn: egg in my yaml

Sarvatt:
Also I made a few changes to the custom warhorn noun handling in here, warhorn was hardcoded in a few places even though it is customizable and the echo was echoing the wrong variable.

PR #3225, opened 2018-11-20
Maybe fix for initial startup race conditions?

rcuhljr:


rpherbig:
How do we know if this works?

rcuhljr:
We have some people who can reproduce it it seems, so we're gonna have them test a bit. It's a pretty negligible hit even if it doesn't work so we might pull it into the baseline just to be safe.

PR #3227, opened 2018-11-20
performance.lic: pause 30 seconds if stamina is too low to play

### LINKED ISSUE #3213: Crossing-training performance section fails to start and hangs if stamina is too low.
Noticed this while in Aesry after climbing around all the stairs. When the performance section starts and attempts to play something, the following message appears:

"You can't muster the strength to perform."

It then errors out with the list of expected return strings and hangs indefinitely.

Sarvatt:
Closes #3213 for real this time.

### DIFF
--- a/performance.lic
+++ b/performance.lic
@@ -44,6 +44,8 @@ class Performance
       line = get
       pause 0.05 unless line
       case line
+      when /muster the strength to perform/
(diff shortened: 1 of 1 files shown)


rpherbig:
Nice

PR #3228, opened 2018-11-21
Update workorders.lic - read logbook instead of log

BinuDR:
Changed it to read my logbook instead of read my log, to prevent it from trying to read my epistemic journal.


Sarvatt:
looks good to me.

PR #3229, opened 2018-11-21
Update combat-trainer.lic - Added snipe ammo pickup

BinuDR:
Line 2483, added snipe to the list.  It wasn't picking up ammo when sniping.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2475,7 +2475,7 @@ class AttackProcess
   end
 
   def shoot_aimed(command, game_state)
-    case bput(command, 'isn\'t loaded', 'There is nothing', 'But your', /you (fire|poach) a .* (arrow|bolt|shard|rock|sphere|clump|coral|fist|holder|lump|patella|pellet|pulzone|quadrello|quill|stone|stopper|verretto) at/i, 'I could not find', 'with no effect and falls to the ground', 'Face what', 'How can you (poach|snipe)', 'your moving to fire went unobserved', 'notices your attempt to remain hidden', 'you don\'t feel like fighting right now', 'That weapon must be in your right hand to fire')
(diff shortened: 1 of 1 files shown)


BinuDR:
There's still 1 case where it fails.  If you get noticed while sniping, it tries to shoot again and then goes straight to loading.  And combat-trainer assumes you always succeed at loading so it goes straight to aiming, and then works the aiming trainables indefinitely.

[combat-trainer]>snipe
The s'sugi malchata notices your attempt to remain hidden while firing your sling!
The s'sugi malchata discovers you, ruining your hiding place!
< Moving with powerful grace, you snipe a quartzite stone shard at a s'sugi malchata.  A s'sugi malchata attempts to dodge.  The shard lands a hard hit (5/23) that raps it sharply across the knuckles.
The stone shard falls to the ground!
[You're adeptly balanced and in dominating position.]
[Roundtime 1 sec.]
R>
[combat-trainer]>shoot
But your sling isn't loaded!
>
[combat-trainer]>load
You don't have the proper ammunition readily available for your sling.

BinuDR:
Sniping is problematic, especially for slings.  It's even worse if you only have 1 shard.

With bows and crossbows, you'll eventually kill the mob and be able to loot your ammo because it's in the room, but sniping with slings is a problem because it never matches the case where it will try to pick up your ammo.

It matches these instead:
'notices your attempt to remain hidden' 
'your moving to fire went unobserved' 

Those two cases don't match the ammo you fire.  The 1st one will try to shoot after sniping, and then moves on to trying to load and aim again.  The 2nd moves straight to the next load/aim cycle.

I fixed it in my local copy by removing those 2 matches entirely, but I'm not sure if there are any edge cases where they are needed.

PR #3230, opened 2018-11-21
combat-trainer.lic: add combat almanac support

Sarvatt:


rpherbig:
❓ What does the almanac do?

I'm thinking if this affects a skill, it might be better as a training ability.

Sarvatt:
the almanac gives you +17/34 mindstates in a random skill every 10 minutes, I would rather it be a training_ability but you cant map it to a specific skill. I was considering mapping it to enchanting that cant be trained so it always triggers on your timer though but thought that was too hacky :)

rpherbig:
We have some examples of training abilities _not_ tied to a skill. Check out 'Favor Orb': https://github.com/rpherbig/dr-scripts/blob/master/combat-trainer.lic#L1880

Sarvatt:
Ohh thanks for that, I had to do a lot of crap to shove it in where I did because I thought training_abilities wouldn't work, I'll update this tomorrow.

Sarvatt:
closing this to push a new branch

PR #3235, opened 2018-11-22
Update workorders.lic - Fix for deed packets in wyvern packs and hitm…

BinuDR:
…an's backpacks

Since you can't tap items in a wyvern hide hunting pack or a hitman's backpack, I changed the line that checks for the deed packet to use look instead.
Oh, and I did check the packet with only 1 deed left.  It still says claim forms and not form.

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -699,7 +699,7 @@ class WorkOrders
     end
 
     return unless @deed_own_ingot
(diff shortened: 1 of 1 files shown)


rpherbig:
> Oh, and I did check the packet with only 1 deed left. It still says claim forms and not form.

Nice catch!

PR #3237, opened 2018-11-22
 combat-trainer.lic: Add Almanac usage to training_abilities.

Sarvatt:


combat-trainer.lic
```diff
@@ -1937,6 +1937,9 @@ class TrainerProcess
     @combat_training_abilities_target = settings.combat_training_abilities_target
     echo("  @combat_training_abilities_target: #{@combat_training_abilities_target}") if $debug_mode_ct
 
+    @combat_almanac = settings.combat_almanac
```
rpherbig: ❓ Can this setting be removed?

combat-trainer.lic
```diff
@@ -2032,6 +2038,19 @@ class TrainerProcess
...
   private
 
+  def use_almanac(game_state)
+    return unless Time.now - UserVars.almanac_last_use >= 600
+    return if left_hand && !game_state.currently_whirlwinding
+
+    game_state.sheath_whirlwind_offhand if game_state.currently_whirlwinding
+    bput('get my almanac', 'You get', 'What were')
+    bput('study my almanac', 'You set about', 'gleaned all the insight you can', 'Study what')
+    waitrt?
```
rpherbig: ❓ Since there's roundtime, should we retreat first?

profiles/Sarvatt-setup.yaml
```diff
@@ -254,6 +268,7 @@ training_abilities:
 #  Collect: 200
   Analyze: 121
   Locks: 70
+  Almanac: 600
```
rpherbig: ❓ Should this value be lower so it checks if the timer is up more frequently? I'm envisioning a worst-case scenario in which the timer is a few seconds from being up, the training ability is checked (and skipped since the timer is not quite up yet), then you wait an additional 10 minutes before using the almanac.

Sarvatt: indeed, wont hurt to lower the timer since it checks the uservar

combat-trainer.lic
Sarvatt: sure can, shouldn't have added that

combat-trainer.lic
Sarvatt: added a retreat and engage

PR #3243, opened 2018-11-23
Added failure message

Glimmereyes:
Added a failure message when pushing the boulder into Gate of Souls

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -830,8 +830,8 @@ class Bescort
   def push_boulder
     fix_standing
     fput('push boulder')
-    case waitfor(/^At the bottom of the hollow/, /^You stop pushing/, /blocking all access to a low tunnel\.$/)
-    when /You stop pushing/, /blocking all access to a low tunnel/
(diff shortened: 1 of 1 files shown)


bescort.lic
```diff
@@ -831,18 +831,18 @@ class Bescort
...
     case waitfor(/^At the bottom of the hollow/, /^You stop pushing/, /blocking all access to a low tunnel\.$/)
-    when /You stop pushing/, /blocking all access to a low tunnel/
+    when /^You stop pushing/, /blocking all access to a low tunnel\.$/
       push_boulder
     end
   end
   
   def gos_tunnel
-    fput('kneel')
+    fput('kneel', 'You kneel down upon the ground')
```
rpherbig: 💡 Can this use `bput` instead?

Sarvatt: aka please consider switching it over in the future even though its merged.

Sarvatt: you might have meant to switch it to bput there, fput doesnt take args like that

rpherbig: `fput` does take args, similar to `bput`, it's just less reliable.

rpherbig: https://github.com/rcuhljr/dr-lich/blob/master/lich.rbw#L5700

PR #3249, opened 2018-11-24
Update smith.lic - replace inv search for oil

BinuDR:
I changed the check_oil method to look at the oil instead of using inv search (which takes like 14+ seconds for me with all of my junk).
I looked up all the different oils on elanthipedia and got their look messaging and put that in the bput matches.
If you have other oils as your 1st-10th oils, it will get the forging oil and make it the top item in your container.
If it's already the top item, it continues as normal.
If it's not found, it will go buy it as usual.

### DIFF
--- a/smith.lic
+++ b/smith.lic
@@ -109,19 +109,7 @@ class Smith
 
   def stamp_item(recipe)
     fput('get my stamp')
-    fput("mark  def check_oil(discipline)
-    $ORDINALS.each do |ord|
-	  case bput("look my #{ord} oil", /A thick and syrupy substance that easily coats and lubricates metal/, /You see nothing unusual/, /I could not find what you were referring to/, /There appears to be something written/, /(coal|lamp|fish) oil/, /The oil bottle is full/)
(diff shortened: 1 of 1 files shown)


smith.lic
```diff
@@ -99,7 +109,19 @@ class Smith
...
+	  when /A thick and syrupy substance that easily coats and lubricates metal/
+	    return if ord == 'first'
+	    bput("get my #{ord} oil", /You get/)
+		bput("stow oil", /You put/)
+		return
+	  when /I could not find what you were referring to/
+	    break
+	  end
+    end
+ my #{recipe['noun']} with my stamp")
```
rpherbig: Something doesn't look right here. Copy paste into the middle of a line?

BinuDR: Damn, it was fine on my end but when I pasted it, it looked wrong.  So I fixed it.  I guess it messed it up instead.

BinuDR:
Another benefit of this change is that it will work with any of the special oils from fests since they all have the same look descriptoion.  They don't show up with 'inv search flask of oil' though.

PR #3250, opened 2018-11-24
almanac.lic: Add more scripts to not run the loop during.

Sarvatt:


profiles/base.yaml
```diff
@@ -1127,3 +1127,13 @@ sanowret_no_use_scripts:
 - shape
 - outdoorsmanship
 - combat-trainer
+
+almanac_no_use_scripts:
```
rpherbig: 💡  Should `bescort` be in this list? I.e. it was in the original list, but not this new one.

Issue #3258, opened 2018-11-26
combat-trainer is not working with plain ammo


Sarvatt:
```[combat-trainer]>shoot
 
Your attempt is slightly hindered by your small shield.
< With strength, you fire an arrow at a young wyvern.  A young wyvern attempts to evade.  
The arrow lands a light hit (1/23) to the wyvern's chest.
The arrow falls to the ground!
[You're bruised, solidly balanced and in better position.]
[Roundtime 1 sec.]
R>
You feel fully rested.
You feel fully attuned to the mana streams again.
>
A young wyvern lurches forward to stab its tail stinger at you!
The young wyvern's tail lashes at you.  You block the tail with your small shield.
[You're bruised, solidly balanced and in better position.]
>
* In a weak and directionless display of aggression, a young wyvern lunges foward and snaps its fangs at you.  You evade, twisting to one side.  
[You're bruised, solidly balanced and have slight advantage.]
>
Brisknite just arrived.
 
>
Brisknite goes southwest.
 
>
* With pathetic timing, a young wyvern lunges foward and snaps its fangs at you.  You partially block with a small shield.  
[You're bruised, solidly balanced and opponent has slight advantage.]
>
* With pathetic timing, a young wyvern slashes its razor-sharp hind talons at you.  You dodge, managing to get out of the way with a minimum of effort.  
[You're bruised, solidly balanced with no advantage.]
>
* Moving in like a timid schoolboy, a young wyvern lunges foward and snaps its fangs at you.  You dodge, leaping aside.  
[You're bruised, solidly balanced and opponent has slight advantage.]
>
You feel fully prepared to cast your spell.
>
[combat-trainer: *** No match was found after 15 seconds, dumping info]
 
[combat-trainer: messages seen length: 31]
 
[combat-trainer: message: You feel fully prepared to cast your spell.]
 
[combat-trainer: message: [You're bruised, solidly balanced and opponent has slight advantage.]]
 
[combat-trainer: message: * Moving in like a timid schoolboy, a young wyvern lunges foward and snaps its fangs at you.  You dodge, leaping aside.  ]
 
[combat-trainer: message: [You're bruised, solidly balanced with no advantage.]]
 
[combat-trainer: message: * With pathetic timing, a young wyvern slashes its razor-sharp hind talons at you.  You dodge, managing to get out of the way with a minimum of effort.  ]
 
[combat-trainer: message: [You're bruised, solidly balanced and opponent has slight advantage.]]
 
[...]

PR #3264, opened 2018-11-27
fix for plain ammo

Glimmereyes:
Fix for issue #3258 

an?
Added "N" and a quantifier to account for situations where the next word starts with a vowel.  While not necessarily part of the reported issue, may be an issue at other points

(?:.*\s)?
Non-capture group to avoid complications for dealing with which group the next line should be using
\s (white space) is included in the group to avoid checking against double spaces in situations where an adjective is not present
? Quantifier at the end to allow for instances where there is no adjective (I.E. the reported problem)

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2493,12 +2493,12 @@ class AttackProcess
   end
 
   def shoot_aimed(command, game_state)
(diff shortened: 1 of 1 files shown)


rpherbig:
Thank you for the explanatory PR description and this contribution!

PR #3265, opened 2018-11-27
Updte afk.lic - now gets 99% of items at feet!

BinuDR:
Updated the section for at feet items to find the noun for items that don't have the noun as the last word.
This fix also requires an update to common.lic.  I will add that in a moment.

I didn't see the need for this line and removed it.  If the original author has a specific case in mind that needs it, please let me know.
    match = match.sub('.', '')

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -57,10 +57,9 @@ loop do
 
   if line =~ /You notice a (.*) at your feet/
     match = Regexp.last_match(1)
-    match = match.sub('.', '')
-    item = match.split(' ').last
-    DRC.bput("get my #{item}", 'You get', 'You pick up', 'But that is already in your inventory')
-    equipment_manager.empty_hands
+    item = DRC.get_noun(match)
+    DRC.bput("get my #{item}", 'You get', 'You pick up', 'But that is already in your inventory', /What were you referring to/)
+	equipment_manager.empty_hands
     stop_script('go2') if Script.running?('go2')
   end
 
--- a/afk.lic
+++ b/afk.lic
@@ -55,7 +55,7 @@ loop do
     end
   end
 
-  if line =~ /You notice a (.*) at your feet/
+  if line =~ /You notice (.*) at your feet/
     match = Regexp.last_match(1)
     item = DRC.get_noun(match)
     DRC.bput("get my #{item}", 'You get', 'You pick up', 'But that is already in your inventory', /What were you referring to/)


Glimmereyes:
Might want to replace line 58 with this
if line =~ /You notice (?:some|a|an) (.*) at your feet/

That would trigger for things like "some bundling rope" or "an arrow".  The "?:" makes it a non-capture group so no need to alter "Regexp.last_match(1)"

Sarvatt:
good catch

BinuDR:
We can just remove the article entirely so it's part of the regex match.
Is there a reason one way would be better than the other?
`/You notice (.*) at your feet/`

Glimmereyes:
> Is there a reason one way would be better than the other?

Because I didn't think of it?

Managed to work in a Loop that seems to work, but I haven't explored this issue much so I may be overlooking something.  This would help deal with multiple ammo being left at the feet.

&nbsp;&nbsp;if line =~ /You notice (.*) at your feet/
&nbsp;&nbsp;&nbsp;&nbsp;stop_script('go2') if Script.running?('go2')
&nbsp;&nbsp;&nbsp;&nbsp;match = Regexp.last_match(1)
&nbsp;&nbsp;&nbsp;&nbsp;match = match.sub('.', '')
&nbsp;&nbsp;&nbsp;&nbsp;item = match.split(' ').last
&nbsp;&nbsp;&nbsp;&nbsp;loop do
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;case DRC.bput("get my #{item}", 'You get', 'You pick up', 'But that is already in your inventory')
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;when 'You pick up'
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;equipment_manager.empty_hands
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;when 'You get', 'But that is already in your inventory'
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;if DRC.right_hand || DRC.left_hand 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;equipment_manager.empty_hands
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;end
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;break
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;end
&nbsp;&nbsp;&nbsp;&nbsp;end
&nbsp;&nbsp;end

Alternatively, could use the Inventory command to generate a list of everything at your feet and then stow from that, but that's starting to get beyond my skill level.  Assuming that works you could pause go2 at the start of the section, clear your feet and then unpause go2 to continue on your way.  

Pausing go2 works with the posted snippit, but if you have a bunch of different nouns you will eventually trigger go2's rerouting

dr-leustyin:
Pretty sure that eliminates the period at the end of the line so you don't try to 'get hammer.', which would fail

dr-leustyin:
No, that's not it.  I don't remember why that sub is there.

BinuDR:
Here's the line we're matching against:
You notice a scuffed leather shoe with a bent silver buckle at your feet, and do not wish to leave it behind.
You notice a dagger with a rearing wolf spider pommel at your feet, and do not wish to leave it behind.
You notice a thick book titled ~101 Rational Reasons Why I am Always Right~ at your feet, and do not wish to leave it behind.

These are the three I have that don't get picked up with the current afk code.

PR #3267, opened 2018-11-27
common-foo: Move DRCT.retreat to DRC to fix circular dependency probl…

Sarvatt:
…em when auto stating afk.lic after equipmanager was added to its custom_require

### DIFF
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -75,7 +75,7 @@ module DRCT
   def retreat(ignored_npcs = [])
     return if (DRRoom.npcs - ignored_npcs).empty?
 
-    DRC.retreat
+    DRC.retreat(ignored_npcs)
   end
 
   def find_empty_room(search_rooms, idle_room, predicate = nil, min_mana = 0, strict_mana = false)
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -75,10 +75,7 @@ module DRCT
   def retreat(ignored_npcs = [])
     return if (DRRoom.npcs - ignored_npcs).empty?
 
-    escape_messages = ['You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat']
-    until escape_messages.include?(DRC.bput('retreat', *(escape_messages + ['retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing'])))
-      DRC.fix_standing
-    end
+    DRC.retreat
   end
 
   def find_empty_room(search_rooms, idle_room, predicate = nil, min_mana = 0, strict_mana = false)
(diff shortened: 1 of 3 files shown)


Sarvatt:
both of these scripts get loaded at login so it should be ok to have a flag day moving it like this without a transition

Sarvatt:
hmm, it may be worth holding off on the equipmanager change at least until the common files propagate, since a lot of scripts do EquipManager.new? Or am I misunderstanding.

common-travel.lic
```diff
@@ -75,10 +75,7 @@ module DRCT
   def retreat(ignored_npcs = [])
     return if (DRRoom.npcs - ignored_npcs).empty?
```
rpherbig: ❓ Why wasn't this moved?

rpherbig:
❓ What's the circular dependency here?

rcuhljr:
common-arcana requires common-travel and common-travel requires common-arcana. SHAME ON YOU SHELT.

rpherbig:
> hmm, it may be worth holding off on the equipmanager change at least until the common files propagate, since a lot of scripts do EquipManager.new? Or am I misunderstanding.

Yeah, I think what we want to do is add `retreat` to `DRC`, have the `DRCT` version do nothing but call the new one, and point `DRCA` at the new one (removing the dependency on `DRCT` in the process).

Then a second PR updates scripts to point to `DRC` and removes the `DRCT` function.

common-travel.lic
Sarvatt: do we want to add drinfomon to custom_require in common.lic?

rcuhljr: Probably the best choice.

Sarvatt: I'm an idiot, was thinking adding drinfomon to common would add another circular dependency but drinfomon doesn't include common, and common already includes drinfomon. fixed.

common-travel.lic
```diff
@@ -75,10 +75,7 @@ module DRCT
...
 
-    escape_messages = ['You are already as far away as you can get', 'You retreat from combat', 'You sneak back out of combat']
-    until escape_messages.include?(DRC.bput('retreat', *(escape_messages + ['retreat', 'You try to back ', 'sneak', 'grip remains solid', 'You must stand first', 'grip on you', 'You stop advancing'])))
-      DRC.fix_standing
-    end
+    DRC.retreat
```
Sarvatt: should this be `retreat(ignored_npcs = [])`?

rcuhljr: it should be `DRC.retreat(ignored_npcs)`

common-travel.lic
rpherbig: 💡 You can delete this guard clause since it's duplicated in `DRC`

PR #3275, opened 2018-11-28
boostrap.lic: Clean up script loading

Sarvatt:


Sarvatt:
E added these changes when we were debugging the circular dependency problem with common-arcana and said he'd like it pushed, I've tested them to no negative side effect.

PR #3276, opened 2018-11-29
crossing-training.lic: Fix performance with too low stamina

### LINKED ISSUE #3272: [Crossing-Training] Performance stamina check possibly missing a message?
[crossing-training]>play polka

You can't muster the strength to perform.
> 
[crossing-training: *** No match was found after 15 seconds, dumping info]

[crossing-training: messages seen length: 3]

[crossing-training: message: Seer's Sense  (Fading)]

[crossing-training: message: Shadows  (19 roisaen)]

[crossing-training: message: You can't muster the strength to perform.]

[crossing-training: checked against [/dirtiness may affect your performance/i, /slightest hint of difficulty/i, /You begin a/i, /You struggle to begin/i, /You're already playing a song/i, /You effortlessly begin/i, /You begin some/i, /You cannot play/i, /Play on what instrument/i, /now isn't the best time to be playing/i, /Perhaps you should find somewhere drier before trying to play/i]]

[crossing-training: for command play polka]

Sarvatt:
This was fixed in ;athletics but crossing-training still had the problem.
Closes #3272

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -779,7 +779,7 @@ class CrossingTraining
 
     UserVars.song = @song_list.first.first unless UserVars.song
     @did_play = true
-    case bput("play #{UserVars.song}", 'dirtiness may affect your performance', 'slightest hint of difficulty', 'You begin a', 'You struggle to begin', 'You\'re already playing a song', 'You effortlessly begin', 'You begin some', 'You cannot play', 'Play on what instrument', 'now isn\'t the best time to be playing', 'Perhaps you should find somewhere drier before trying to play')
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -799,6 +799,8 @@ class CrossingTraining
       stop_play
       UserVars.song = @song_list[UserVars.song] || @song_list.first.first
       return play_song?
+    when 'muster the strength to perform'
+      pause 30
```
rpherbig: ❓ Do we really want to pause 30? Unlike performance, crossing-training is often playing while trying to train something else.

Sarvatt: good point

PR #3282, opened 2018-11-29
refactor valid item check

rcuhljr:
There's a small chance that this should actually be `item_data.has_key?(x)` but I don't see why you want to allow nulls in for these fields, it seems like you actually want to check that they have values.

Someone should verify this first.

### DIFF
--- a/restock.lic
+++ b/restock.lic
@@ -100,13 +100,7 @@ class Restock
   end
 
   def valid_item_data?(item_data)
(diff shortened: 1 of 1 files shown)


rpherbig:
The code looks ok. Needs testing.

Sarvatt:
This has been working fine for me.

rpherbig:
Thank you for testing it @Sarvatt

PR #3297, opened 2018-12-02
Update healme.lic

aruthas:
Made adjustments for Heal, cure disease, and flush poison with the adaptive cure

### DIFF
--- a/healme.lic
+++ b/healme.lic
@@ -33,18 +33,26 @@ class HealMe
   def severity(wounds)
     wounds.map { |wound| wound['level'] }.inject(&:+)
   end
-
   def spam_heal(base, camb)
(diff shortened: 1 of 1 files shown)


healme.lic
```diff
@@ -80,24 +88,42 @@ class HealMe
...
+		    end
+      end
+	  end  
   end
 
   def heal_self(skip_parts)
     return unless DRStats.empath?
-
+    health = check_health
+    cure_disease(spells, health)
```
rpherbig: ❓ Are diseases and poisons more urgent to heal than wounds?

aruthas: depends on level, if you are taking damage you may never get out of healing until the disease or poison where off

aruthas: played with it with critical poison and it kept in a healing loop because the poison was doing damage

PR #3305, opened 2018-12-03
added strength and capacity levels for Osrel Meraud

marshall:
Messaging taken from https://elanthipedia.play.net/Osrel_Meraud and verified with Rao (may want other clerics to verify as well):

```
>foc orb
You graze your fingertips over the surface of the shadowy orb, feeling the crackling energy coursing through your hands.

The orb resonates with a cache of holy mana.  You perceive a slight tugging sensation, and realize that it is receptive to magical energy.
The pervasiveness of the sensation evokes the image of a lake in your mind's eye. (4/11)
The strength of the sensation evokes the image of a blazing torch. (6/10)
Roundtime: 6 sec.
```

### DIFF
--- a/textsubs.lic
+++ b/textsubs.lic
@@ -497,6 +497,32 @@ TextSubs.add('^After a moment, you see a vision of your god who calls to you by 
 TextSubs.add('^After a moment, you see a crystal-clear vision of your god who speaks slowly and deliberately, "Your unwavering faith and devotion pleases me greatly, (\w+)\.\s+Go forth and continue your works, and you shall only attain a greater level of purity\."', 'After a moment, you see a crystal-clear vision of your god who speaks slowly and deliberately, "Your unwavering faith and devotion pleases me greatly, \1.  Go forth, and continue your works and you shall only attain a greater level of purity." (15/16)')
 TextSubs.add("^After a moment, you feel a clear presence like a warm blanket covering you beneath the shade of a giant sana'ati tree\.", "After a moment, you feel a clear presence like a warm blanket covering you beneath the shade of a giant Sana'ati tree. (16/16)")
 
+# Osrel Meraud power / capacity messaging for Clerics (seen when you 'focus orb')
+om_str = 'The strength of the sensation evokes the image of'
+TextSubs.add("^#{om_str} a pile of cold ashes\\.", "#{om_str} a pile of cold ashes. (1/10)")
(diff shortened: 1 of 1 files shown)


textsubs.lic
```diff
@@ -497,6 +497,32 @@ TextSubs.add('^After a moment, you see a vision of your god who calls to you by
 TextSubs.add('^After a moment, you see a crystal-clear vision of your god who speaks slowly and deliberately, "Your unwavering faith and devotion pleases me greatly, (\w+)\.\s+Go forth and continue your works, and you shall only attain a greater level of purity\."', 'After a moment, you see a crystal-clear vision of your god who speaks slowly and deliberately, "Your unwavering faith and devotion pleases me greatly, \1.  Go forth, and continue your works and you shall only attain a greater level of purity." (15/16)')
 TextSubs.add("^After a moment, you feel a clear presence like a warm blanket covering you beneath the shade of a giant sana'ati tree\.", "After a moment, you feel a clear presence like a warm blanket covering you beneath the shade of a giant Sana'ati tree. (16/16)")
 
+# Osrel Meraud power / capacity messaging for Clerics (seen when you 'focus orb')
+om_str = 'The strength of the sensation evokes the image of'
+TextSubs.add("^#{om_str} a pile of cold ashes\\.", "#{om_str} a pile of cold ashes. (1/10)")
```
rpherbig: ❓ Won't `\\.` escape the `\` and then interpret the `.` as 'any character', rather than the literal `.`?

marshall: the other way around I think? the remaining `\` tells the regex compiler to use a literal `.` and not 'any character'. 

notice this is a double quoted string, where escaping rules are different. Try this in IRB:
```
> '\.'
 => "\\."
```

Sarvatt: I don't think it works that way, these are passed to gsub. Look directly above the textsub line that works

`TextSubs.add("^After a moment, you feel a clear presence like a warm blanket covering you beneath the shade of a giant sana'ati tree\.", "After a moment, you feel a clear presence like a warm blanket covering you beneath the shade of a giant Sana'ati tree. (16/16)")`

the string is passed along to gsub after the variable interpolation as the regex in your change, you are not using regex directly here. `\.` should be correct.

Sarvatt: otherwise you are literally sending this to the regex parser as-is
```
/^The strength of the sensation evokes the image of a pile of cold ashes\\./

Sarvatt:
I've tested the latest change and it's working properly on my cleric.

PR #3317, opened 2018-12-12
Safe room tip fix for DRCM.wealth

Tarhish:
It always makes me suspicious when I seem to be the only one having a particular bug, but this looks pretty straightforward. 'wealth' on line 231 refers to DRCM.wealth, which accepts one arg, your hometown. Until I made this change it errored out every time it went to tip, and now everything works smoothly.


Sarvatt:
this makes sense, you are probably the only person that tips a PC empath after darts DRCM changes

PR #3318, opened 2018-12-12
base-spells.yaml: Release innocence before recasting, you can't refre…

Sarvatt:
…sh it.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -1167,7 +1167,7 @@ spell_data:
     skill: Utility
     abbrev: INNOCENCE
     mana: 5
-    recast: 2
+    recast: 0
     before:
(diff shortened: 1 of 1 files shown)


data/base-spells.yaml
```diff
@@ -1167,7 +1167,12 @@ spell_data:
     skill: Utility
     abbrev: INNOCENCE
     mana: 5
-    recast: -1
+    recast: 0
+    before:
+    - message: release innocence
+    matches:
```
rpherbig: ❓ Does `matches` need to be indented one more level? It looks like it's part of the spell definition when it should be part of the `before`.

Sarvatt:
glad you caught that! it was a bad cut and paste, thank you

PR #3319, opened 2018-12-13
Support safe-room declaration by hometown

Tarhish:
This is mucking with dependency so consider this more of a 'Hey, what does everyone think of this?'  Take a look, and this could go back to the drawing board.

This adds some entries to transform_settings to handle declaring various rooms as a hash, for example:
```
safe_room:
  Crossing: XXXX
  Shard: YYYY
lockpick_room_id:
  Crossing: XXXX
  Shard: YYYY
engineering_room:
  Crossing: XXXX
  Shard: YYYY
outfitting_room:
  Crossing: XXXX
  Shard: YYYY
```
Transform_settings will ensure any script sees the entry corresponding to the hometown you've set, or not do anything at all if you've just set it normally. As such, this shouldn't impact any existing yamls and should be opaque to any scripts.  I've been running with it for a couple weeks now swapping between Shard and Crossing, and it has saved me time, but mostly I just like the idea of setting room ids once and not having to update them whenever I change hunting grounds.

Anyone have any thoughts about this, or ideas on improvement?  Think this'll cause trouble?

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -920,6 +920,15 @@ class SetupFiles
     s.bankbot_room_id ||= UserVars.bankbot_room_id
     s.prehunt_buffs ||= UserVars.prehunt_buffs
 
+    # Pull town-specific room_ids iff defined
+    s.safe_room = s.safe_room[s.hometown] if s.safe_room.is_a?(Hash) && s.safe_room[s.hometown]
+    s.compost_room = s.compost_room[s.hometown] if s.compost_room.is_a?(Hash) && s.compost_room[s.hometown]
+    s.lockpick_room_id = s.lockpick_room_id[s.hometown] if s.lockpick_room_id.is_a?(Hash) && s.lockpick_room_id[s.hometown]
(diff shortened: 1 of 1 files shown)


Tarhish:
Sure, I don't see why not for bank bots, though I'm not sure how often they'll move towns.

As for training rooms I'd have to check.  I don't use it myself, but scanning how it's used it looks like a set of several rooms that you'd use randomly as your safe-room while training?  If that's true then I'd definitely want to test that.  I think it would work though.

Sarvatt:
only suggested the bankbot and theurgy_prayer_mat_room because its a 1 line cut and paste each that's easy, the prayer mat room change would be completely safe to add. dont worry about training_rooms :)

Tarhish:
Yeah, it just feels like a waste to do all that then have people still have to update their training_rooms when they move towns like some caveman hitting their yamls with sticks and rocks.  :D

Sarvatt:
no joke, this is a good change to add as-is, no reason to hold it up to make it perfect.

Sarvatt:
other people can add on to it if they need it, i can send a 2 line change immediately for my review things

rpherbig:
Sorry for the delay, I missed this PR somehow.

PR #3330, opened 2018-12-23
 walkingastro.lic: make the excluded scripts configurable and increase observation speed.

Sarvatt:


walkingastro.lic
```diff
@@ -162,22 +144,19 @@ class WalkingAstro
...
   end
 
-  def can_observe?
-    ['buff', 'steal', 'combat-trainer', 'pick', 'craft', 'shape', 'sew', 'bescort', 'remedy', 'forge', 'carve', 'performance', 'theurgy', 'hlctheurgy', 'astrology', 'astrology2', 'study-art', 'mech-lore', 'first-aid'].none? { |script| Script.running?(script) }
-  end
-
   def should_observe?
-    !xp_highenough? && can_observe?
-  end
+    return if @no_use_scripts.any? { |name| Script.running?(name) }
```
rpherbig: 💡 This would be clearer with an explicit `return false`

Issue #3329, opened 2018-12-23
Flee Command stalling combat-trainer


BL4CKM4GICK:
I have had this issues happen twice now.. 

[combat-trainer]>flee bluff
You assess your combat situation and realize you don't see anything engaged with you.

Combat-trainer just hangs after that.

PR #3332, opened 2018-12-24
Fix flee hang on empty room

Tarhish:
Appears to fix issue https://github.com/rpherbig/dr-scripts/issues/3329

This way it won't waitfor if it gets the 'nothing engaged to you' message


Sarvatt:
changing it to just Obvious would work then? cant say im surprised i didnt think of that :)

Tarhish:
Yeah, that works.  I was rummaging for examples of mulitple waitfors, but that's better.  Fails safer anyway.  I'll get that while I'm here.

PR #3334, opened 2018-12-24
Add common-moonmage

jandersson:
A module for common methods related to moon mages. This will help cut down
on code duplication for astrology related functionality.

Ive taken a look through walkingastro, astrology, and the PR by @aruthas which adds more astrology functionality. A lot of these scripts can be helped by refactoring. In the case of astrology, there are a lot of methods which need to be broken up into smaller methods (some of which can then be commonized).

### DIFF
--- a/bootstrap.lic
+++ b/bootstrap.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#bootstrap
 =end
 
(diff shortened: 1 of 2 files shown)


jandersson:
As scripts are refactored, we can update DRCMM. At the moment its difficult to move methods wholesale into DRCMM, the methods in astrology need to be broken up so the common bits stand on their own

PR #3335, opened 2018-12-24
Refactor walkingastro

jandersson:
* Use DRCMM
* General refactoring

Hopefully doesn't conflict with #3330 too much

### DIFF
--- a/walkingastro.lic
+++ b/walkingastro.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#walkingastro
 =end
 
-custom_require.call %w[common drinfomon]
+custom_require.call %w[common common-moonmage drinfomon]
 
 class WalkingAstro
   def initialize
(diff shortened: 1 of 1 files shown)


jandersson:
This depends on common-moonmage being live.

Sarvatt:
did you mean to submit common-moonmage.lic too here?

walkingastro.lic
```diff
@@ -14,16 +14,13 @@ class WalkingAstro
   end
 
   def get_telescope
-    fput('stop play')
-    case DRC.bput('get my telescope', 'You get a', 'You need a free hand to pick that up.', 'You are already holding that.', 'You should stop practicing your Athletics skill before you do that')
-    when 'You get a'
+    case DRCMM.get_telescope
```
rpherbig: 💡 The way a few of the DRCMM helpers are used makes me wonder if they should be predicates. I won't have a strong opinion until I see more than one usage, though.

PR #3339, opened 2018-12-31
Combat-Trainer: Use skill attr to determine sorcery spell

jandersson:
If the magic skill shows the spell to be sorcery then stow/get weapons. This lets you use a sorcery spell to train a non-sorcery skill with the safety of storing weapons before casting.

Realize this can be a bit confusing, so here is an example: 

```
combat_spell_training:                                    
  Warding:                                                
    name: Ghost Shroud                                    
    mana: 21                                              
    cyclic: true                                          
    skill: Sorcery                                        
    day: true # Keeps from conflicting with SLS           
```
A moon mage uses a cleric (sorcery) spell to train. The spell is warding, sorcery, and cyclic. It trains warding slower than sorcery. So I want combat-trainer to trigger the casting of it based on warding experience. I list the skill as sorcery to use the stow/get weapon mechanics so i dont lose a weapon when casting.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1335,11 +1335,18 @@ class SpellProcess
 
     game_state.hide_on_cast = true if data['use_stealth']
 
-    if needs_training == 'Sorcery' || @training_spells[needs_training]['skill'] == 'Sorcery'
+    if spell_is_sorcery?(data)
       echo 'Preparing a sorcery spell, held items will be stowed to prevent item loss' if $debug_mode_ct
       game_state.casting_sorcery = true
     end
     prepare_spell(data, game_state)
+  end
+
+  def spell_is_sorcery?(spell_data)
+    return true  if spell_data['sorcery']
+    return false if DRStats.native_mana.casecmp(spell_data['mana_type']).zero?
+    return false if spell_data['mana_type'].casecmp('ap').zero?
+    true
   end
 
   def check_buffs(game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1335,7 +1335,7 @@ class SpellProcess
 
     game_state.hide_on_cast = true if data['use_stealth']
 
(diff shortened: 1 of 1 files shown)


Sarvatt:
could we do this automatically based on mana_type: not being native to make it more intuitive? just spitballing, this looks good as-is.

jandersson:
Thats a fantastic idea.

rpherbig:
I like the idea of doing this automatically based on native mana type.

If we still want a flag, I wonder if this should be something like `sorcery: true` instead of `skill: Sorcery`. Is this overly extensible (i.e. when would you want a value _other than_ 'Sorcery' for the skill)?

Sarvatt:
how should we implement that? a mapping of guilds and native mana types in drinfomon maybe?

rpherbig:
That seems reasonable. The `DRStats` class seems like a good place for such a helper.

jandersson:
Solid ideas, ill set it up

combat-trainer.lic
```diff
@@ -1335,13 +1335,20 @@ class SpellProcess
...
       echo 'Preparing a sorcery spell, held items will be stowed to prevent item loss' if $debug_mode_ct
       game_state.casting_sorcery = true
     end
     prepare_spell(data, game_state)
   end
 
+  def spell_is_sorcery?(spell_data)
+    return true  if spell_data['sorcery']
+    return false if DRStats.native_mana.casecmp(spell_data['mana_type']).zero?
+    return false if spell_data['mana_type'].casecmp('ap').zero?
```
rpherbig: 💡 What is `ap` here?

Sarvatt:
analagous patterns I'm guessing

Sarvatt:
yeah ```  Ease Burden:
    skill: Augmentation
    abbrev: EASE
    mana: 1
    mana_type: ap

PR #3342, opened 2019-01-03
base-crafting.yaml: Make lang and theren hometowns use riverhaven soc…

### LINKED ISSUE #3341: NOT an issue...Just a change?
I was having trouble with workorders being skipped, since my hometown was set to langenfirth and no forge is in the area.  I solved this problem by splitting hometown: into hometown: and crafttown: in my yaml and modifying a copy or workorders.lic and smith.lic to read crafttown instead of hometown from yaml.  (i call the modified workorders.lic from a before: ) works like a charm, i can keep my training manager doing its thing in lang or theren and choose what town i want it to run to for crafting.  I can show the changes i made if anyone is interested.

Sarvatt:
…ieties"

Closes #3341

### DIFF
--- a/data/base-crafting.yaml
+++ b/data/base-crafting.yaml
@@ -96,7 +96,7 @@ blacksmithing:
     trash-room: 14350
     idle-room: 11260
     part-room: 11263
-  Riverhaven:
+  Riverhaven: &theren_blacksmithing
     npc: Fereldrin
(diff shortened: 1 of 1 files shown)


Sarvatt:
i plan on deduplicating all the part entries in base-crafting later but that's a project for another day :D

PR #3343, opened 2019-01-03
Simple script to deal with go2 hanging

dr-leustyin:


get2.lic
```diff
@@ -0,0 +1,3 @@
+custom_require.call(%w[common-travel])
+
+DRCT.walk_to(variable[1])
```
rpherbig: 💡 I recommend the `args` style

PR #3346, opened 2019-01-04
common-arcana.lic: Add persistent option for cyclics to skip releasin…

Sarvatt:
…g when combat-trainer exits.

Use case was someone requesting to not release regen. 
Add persistent: true to your cyclic spell.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -229,6 +229,7 @@ module DRCA
   def release_cyclics
     get_data('spells')
       .spell_data
+      .reject { |_name, properties| properties['persistent'] }
       .select { |_name, properties| properties['cyclic'] }
       .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
       .map { |_name, properties| properties['abbrev'] }
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -1,4 +1,4 @@
-Wperr---
+---
 charge_messages:
 - You'll have to hold it
 - You are in no condition to do that
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -1,4 +1,4 @@
----
+Wperr---
 charge_messages:
 - You'll have to hold it
 - You are in no condition to do that
@@ -332,6 +332,7 @@ spell_data:
     abbrev: ABAN
     prep_type: prepare
     cyclic: true
+    persistent: false
     mana: 7
   Absolution:
     skill: Utility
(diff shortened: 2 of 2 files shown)


Sarvatt:
please don't pull this yet, I'm using this to test fixing travis-ci

common-arcana.lic
```diff
@@ -229,6 +229,7 @@ module DRCA
   def release_cyclics
     get_data('spells')
       .spell_data
+      .reject { |_name, properties| properties['persistent'] }
```
rpherbig: ❓ Are you also planning on marking spells as `persistent` in this PR?

Sarvatt:
I pushed the CI changes to another PR so this is ready again. No actually, I think it's harmful to mark anything as persistent by default, this is just for people who opt in to wanting to be healed always instead of trying to learn skills. Marking anything persistent like this will stop it from being recycled every 5 minutes for training purposes also

rpherbig:
My concern is that this pulls the `persistent` flag from base-spells, which wasn't really intended to be overridden by a user. Are the users that want the `persistent` flag in dev mode?

Sarvatt:
I'm not understanding, are you saying maybe all spells should have persistent: false added in base-spells? I can do that.

common-arcana.lic
Sarvatt: marked all spells as persistent: false in base-spells.yaml

Sarvatt:
oh no, this just rejects releasing any spells where persistent: true exists in their yamls for that spell. ive tested this with hyh, my first thought was to reject releasing any hamless: true spell but that would leave up debils you might always want running in combat so i thought a new persistent: true option might make sense for the very rare spell someone might want to skip releasing after combat. i tested that his works by adding persistent: true to my hyh spell and killing combat trainer after it had cast it, it wouldnt release the spell afterwards. there were a few people that wanted to never drop off the regen cyclic after combat-trainer was killed. all it takes is adding persistent: true to your offensive_spells or buff_spells to make release_cyclics skip releasing it in drca.release_cyclics

rpherbig:
I guess I'm still confused as to how `get_data('spells').spell_data` (which is what `release_cyclics` does) pulls any info from your YAML.

If you could retest running `combat-trainer` once with `persistent: true` (and kill combat-trainer and watch the cyclic remain) then change it to `persistent: false`, run combat-trainer, kill combat-trainer, and watch the cyclic get dropped, that would alleviate my concern.

PR #3347, opened 2019-01-05
.travis.yaml: Fix CI builds after jan 3rd travis change

Sarvatt:


Sarvatt:
Ruby <2.3 stopped being supported by the latest bundler updates, so force it back to a lower version.

rpherbig:
Thank you for fixing this!

PR #3349, opened 2019-01-06
Change ;craft recipe that depends on parts

Tarhish:
Changing wood pendants from 100-175 to wood amulets.

Users reported trouble working with ;craft for wooden pendants.  Apparently that required a short leather cord, which craft doesn't grab.  wood amulets appear to be nearly identical in skill_requirements, and the users seem to be in the right area.  Tested myself by forcing myself into that skill zone and got a completed amulet out of it.

I didn't see any other recipes that seemed to need parts.


Sarvatt:
must have missed this when i switched everything using parts over, thanks for catching it!

Tarhish:
Thanks!

PR #3353, opened 2019-01-07
Add mana_type all spells

jandersson:
We'll need all spells to have a mana type for #3339


rpherbig:
Does this mean a spell like `Aethrolysis` won't be detected as Sorcery for WMs?

Sarvatt:
true, all high sorcery should be marked arcane

PR #3357, opened 2019-01-08
base-spells.yaml: Mark high sorcery as arcane mana_type

Sarvatt:


Sarvatt:
these are always sorcery and always guild specific, other guilds cant learn them so it wont interfere with a necro.

Issue #3314, opened 2018-12-08
Trabe Chalice recasting before it expires


Tarhish:
I mean to look at this myself at some point, but first I have to confront the pain of regexes in spellmonitor.

Combat-trainer and buff get sort of uncertain about whether or not Trabe Chalice is up or not.  In general, it'll try to recast Trabe Chalice one extra time during each cast's duration, which fails since the spell can't be overwritten without higher mana.  

It doesn't get stuck recasting it, it just tries once.  For example, if I prebuff TRC then when combat trainer starts up it'll try buffing Trabe Chalice, fail, then move on.  It won't try again until it fades.  I'm guessing it's something weird with the way it presents in ActiveSpells.

```
Trabe Chalice  (intact, 8 roisaen)
Trabe Chalice  (intact, 1 roisan)
```

jandersson:
TRC was a source of pain for me as well. It might be worth having spellmonitor strip everything after the first comma (plus the whitespace) to get it reading as a regular spell.

Sarvatt:
does the spell ever show if its not intact?

jandersson:
Yea, it has various stages of breakage

Tarhish:
Yeah, it's got descriptions like cracked, shattered, smashed, all the way
down.  It couldn't give a percentage, Noooooo.



On Jan 8, 2019 3:35 PM, "Robert Hooker" <notifications@github.com> wrote:

does the spell ever show if its not intact?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rpherbig/dr-scripts/issues/3314#issuecomment-452442344>,
or mute the thread

Sarvatt:
ok took Cran's suggestion and ignored everything before the comma inside the parenthesis

Sarvatt:
ok no, i think i found the real problem here, there is something else wrong.. spellmonitor does catch it fine as long as its the first spell in activespells, everything but the active spell gets added to the next active spell, what I pushed doesn't do anything different.

https://rubular.com/r/OIIbeY5Y4HoA7i

Tarhish:
Ahah, so that might be the answer to the mystery.  I couldn't explain on the surface of it why it wouldn't see TRC, cast it and fail, but then not cast it again.  Maybe that reset it to first position, which made it happy?

I'll have to see if that makes sense once I get home or if I'm just talking crap.

Tarhish:
[...]

PR #3361, opened 2019-01-08
spellmonitor.lic: handle trabe chalice

Sarvatt:
Closes: #3314


Sarvatt:
https://rubular.com/r/avecAQ4eszsbuE

Sarvatt:
Changed the regex to ignore anything before the comma since it has various states of breakage

https://rubular.com/r/THS3t5Yb97MuKm

Tarhish:
I insist that Rubular leaves my mother out of this, but I'll check to see if I notice any odd behavior when I get home :-D

Sarvatt:
ahh this is only handling the case where trabe chalice is the first spell in active spells, will fix

PR #3367, opened 2019-01-09
journal.lic: Use a uservar to track time instead of attempting every run

Sarvatt:
They are on a 24 hour timer.

### DIFF
--- a/journal.lic
+++ b/journal.lic
@@ -23,7 +23,10 @@ class WriteJournal
       echo 'No journal, exiting.'
       exit
     end
-    bput("write my #{@journal_noun}", 'You write', 'Flipping to a blank page', 'Having recently written', 'As you open your', "You can't do that")
+    case bput("write my #{@journal_noun}", 'You write', 'Flipping to a blank page', 'Having recently written', 'As you open your', "You can't do that")
+    when 'You write'
+        UserVars.journal_last_use = Time.now
+    end
     bput("wear my #{@journal_noun}", 'You attach', 'Wear what')
   end
 end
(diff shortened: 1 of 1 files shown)


Sarvatt:
Sorry, just rereading it after it got merged made me want to make a change :D

journal.lic
```diff
@@ -10,11 +10,14 @@ class WriteJournal
...
     settings = get_settings
     @journal_noun = settings.journal_noun
+    UserVars.journal_last_use ||= Time.now - 86400
 
     write_journal
   end
 
   def write_journal
+    return if Time.now - UserVars.journal_last_use < 86400
+
```
rpherbig: ❓ Do you need to set the UserVar to Time.now here?

Sarvatt: whoops

journal.lic
```diff
@@ -10,17 +10,23 @@ class WriteJournal
...
     case bput("remove my #{@journal_noun}", 'You remove', 'You detach', 'What were', 'Remove what')
     when 'Remove what', 'What were'
       echo 'No journal, exiting.'
       exit
     end
-    bput("write my #{@journal_noun}", 'You write', 'Flipping to a blank page', 'Having recently written', 'As you open your', "You can't do that")
+    case bput("write my #{@journal_noun}", 'You write', 'Flipping to a blank page', 'Having recently written', 'As you open your', "You can't do that")
+    when 'You write'
```
Sarvatt: did it this way just in case for some reason your journal isn't available you wont reset the timer

PR #3372, opened 2019-01-12
Add coordinator

jandersson:
This is a training manager made by @charleyc-dr that I use

Since he's not active at the moment id like to push this into the script library and maintain. Crannach's setup files already contain usage examples.

https://github.com/rpherbig/dr-scripts/blob/master/profiles/Crannach-setup.yaml#L703-L834

### DIFF
--- a/coordinator.lic
+++ b/coordinator.lic
@@ -1,0 +1,478 @@
+custom_require.call(%w[common drinfomon common-travel common-items common-arcana common-money])
+
+class Coordinator
+  include DRC
+  include DRCT
+  include DRCA
+  include DRCM
+
+  def initialize(debug)
+    UserVars.coordinator_timers ||= {}
+    
+    @debug = debug
+    @settings = get_settings
+    @hunting_tasks = @settings.coordinator_hunting_tasks
+    @hunting_tasks.each { |task| task[:type] = :hunting }
+    @town_tasks = @settings.coordinator_town_tasks
+    @town_tasks.each { |task| task[:type] = :town }
+    @cleanup_tasks = @settings.coordinator_hunting_cleanup
+    @cleanup_tasks.each { |task| task[:type] = :cleanup }
+
+    @song_list = get_data('perform').perform_options
+
+    prepare_tasks
+    
+    Flags.add('coord-song', 'you finish playing')
+    release_cyclics
+    loop do
+      break unless run_next_task
+    end
+  end
+
+  def run_next_task
+    return unless task = get_next_task
+    echo("Found task to execute: #{task}") if @debug
+    (task[:type] == :hunting) ? run_hunting_task(task) : run_town_task(task)
+    update_predicates_for_completed_task(task)
+    true
+  end
+
+  def get_next_task
+    [@cleanup_tasks, @hunting_tasks, @town_tasks].each do |task_list|
+      task_list.reject! { |task| task[:deleted] }
+    end
+    cleanup_task || town_task || hunting_task || default_hunting_task
+  end
+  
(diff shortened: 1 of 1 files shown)


coordinator.lic
```diff
@@ -0,0 +1,478 @@
+custom_require.call(%w[common drinfomon common-travel common-items common-arcana common-money])
```
rpherbig: 💡 Please add an elanthipedia entry and link here

coordinator.lic
```diff
@@ -0,0 +1,478 @@
...
+
+class Coordinator
+  include DRC
```
rpherbig: 💡 Why is `common-items` in the `custom_require` list but not the `include` list? Does it need to be here or removed from above?

coordinator.lic
```diff
@@ -0,0 +1,478 @@
...
+    DRSkill.getxp(skill) < target
+  end
+
+  def skill_greater_than?(skill, target)
+    echo("Skill level is #{DRSkill.getxp(skill)}") if @debug
+    DRSkill.getxp(skill) > target
+  end
+
+  def update_box_count
+    @num_boxes = (get_boxes(@settings.picking_box_source) || []).size
```
rpherbig: 💡 When would `get_boxes` return nil? The `|| []` looks suspicious here (and the next line)...

coordinator.lic
```diff
@@ -0,0 +1,478 @@
...
+        walk_to(safe_room)
+      else
+        walk_to(task[:walk_to])
+      end
+    end
+    if task[:buff]
+      wait_for_script_to_complete('buff', [task[:buff]])
+    end
+    if task[:town] && DRStats.necromancer?
+      wait_for_script_to_complete('release-necro')
```
rpherbig: 💡 This script isn't in this repository. What will to a user that triggers this case?

coordinator.lic
```diff
@@ -0,0 +1,478 @@
...
+  def infuse_osrel(task)
+    if @settings.osrel_amount && DRSpells.active_spells['Osrel Meraud']
+      infuse_om(!@settings.osrel_no_harness, @settings.osrel_amount)
+    else
+      echo '***OM IS DOWN OR NO osrel_amount***'
+      task[:deleted] = true
+    end
+  end
+
+  def check_favors(task)
```
rpherbig: 💡 A lot of this function looks to be identical to `training-manager`, a good candidate for a common file.

coordinator.lic
```diff
@@ -0,0 +1,478 @@
...
+  def train_performance(task)
+    return if play_song?(true)
+    
+    echo '***UNABLE TO TRAIN PERFORMANCE, REMOVING IT FROM THE TRAINING LIST***'
+    task[:deleted] = true
+  end
+
+  def train_outfitting(task)
+    return unless ensure_copper_on_hand(5000, @settings.hometown)
+    wait_for_script_to_complete('workorders', ['Tailoring'])
```
rpherbig: 💡 Is this still the best way to train outfitting since workorders don't pay out for f2p?

coordinator.lic
```diff
@@ -0,0 +1,478 @@
...
+      target_skills.reject! { |skill| DRSkill.getxp(skill) >= 28 }
+      $CROSSING_TRAINER.stop if target_skills.empty?
+      pause 5
+    end
+  end
+
+  # Harvested wholesale from crossing-training
+  
+  def play_song?(blocking = false)
+    return false if @no_instrument
```
rpherbig: 💡 Should this call `performance.lic`?

PR #3373, opened 2019-01-13
Add guild-reqs to textsubs

Tarhish:
The intent of this is to reduce the number of regex matches being performed on general stuff by splitting the textsubs by guild.  

A few warrior_mage textsubs moved to colocate with others of their guild. There might be some more textsubs that we could isolate further. It also might be worth double-checking to make sure I mapped guilds correctly, especially the bard stuff.

Also, added trader starlight levels, matching precedent in DRCA.perc_aura.


Tarhish:
I only ever come up with ideas after starting the PR.  I think I could move positional mana perception (east/west, etc.) to non-lunar guilds only.  There are probably a few more that could be changed up, but I'll wait to see if I need any other changes.

rpherbig:
> The intent of this is to reduce the number of regex matches being performed on general stuff by splitting the textsubs by guild.

Is the large number of regex matches causing problems?

Tarhish:
I'm going to have to summon @Sarvatt to the stand.  I've had some weird slowdowns on certain computers while using go2, but he's the one who gave me the idea.  He mentioned he'd profiled it and it was spending massive amounts of time doing regex parsing.

Sarvatt:
I dont have numbers anymore because I did this a year ago when I used a raspberry pi but I profiled it in perf to see that it was textsubs causing the overhead, most of the time spent is in ruby's C regex parser when textsubs is loaded and it had a significant cpu overhead increase when using higher typeahead amounts

Tarhish:
I think, if I'm reading it right, it's adding them onto the downstreamhook on startup, then removing them on dying.  So it should only be checking your guild on startup, adding all the ones relevant to you, then sitting there until it's killed.

BinuDR:
We should probably also have go2 pause textsubs while it's running.

Tarhish:
Yeah, I just tested it, and pausing it doesn't do anything since those entries are already hooked in to DownstreamHook in lich.rbw.  Killing it would unhook everything, but I'm not yet sure if that's a net positive, or if the overhead from stopping it and restarting it would be more desirable.  I'll have to test it on a potato.

BinuDR:
Yeah. I was thinking pause at first because it could be implemented in textsubs, but it needs to be killed.

rcuhljr:
Instead of pausing you can do something like this in go2.

```
subaction = DownstreamHook.remove('textsub')
ALL YOUR GO2 CODE STUFF
DownstreamHook.add('textsub', subaction)
```

Since you really don't want to lose the action you may want to put the later part in a before dying.

Alternatively you can refactor `TextSubs` to do the same thing internally with another class variable and just expose helper pause methods so anyone else can go `TextSubs.pause` `TextSubs.resume`.

Tarhish:
Thanks, this could be very useful, depending on the circumstances.

Changing something like walk-to to include that unhook should be trivially easy, but I hesitate to meddle with go2, given its Lich default status.

FWIW, I did test this process by locking some of the appraisal matches behind being not-my-guild, and everything continued working except that specific appraisal stuff, so I know the process in general works.  And the guild checking shouldn't add any overhead, since they only run on startup.

So I think this really depends on the prevailing opinion of whether this is worth doing in the first place.  This was maybe fifteen minutes of effort and I learned some interesting things doing it, so it's no loss if we want to scrap it.  :-)

Sarvatt:
Looks good to me, let's leave the go2 changing for a feature request. :)

Tarhish:
Hey, can barbarians or thieves recognize spell preps?

PR #3374, opened 2019-01-13
bescort.lic: Add moon mage messaging to study reed in crocs

Sarvatt:


Sarvatt:
```
You study the sky but fail to find anything like what you're looking for in the heavens.

PR #3375, opened 2019-01-14
go2.lic: kill textsubs while moving

Sarvatt:


Sarvatt:
I went the even easier way from https://github.com/rpherbig/dr-scripts/pull/3375 :)

Tarhish:
So if this works correctly, then would we even need the other PR?  I can't think anything else is going to be doing as much regexing as go2.

Sarvatt:
I'd like the other fix to go in anyway, good to remove any bloat we can :)

rpherbig:
Looks good to me. @Sarvatt - feel like this is well tested?

Sarvatt:
ive been using it for 2 days and am confident about it. its pretty straight forward.

PR #3381, opened 2019-01-16
Support using starlight_threshold in combat-trainer

Tarhish:
Second phase of implementation of trader starlight aura handling.  Might take a week or two to run this by other traders first, so no hurry.

Usage:
A user sets aura_frequency: to a time period in seconds.  At startup and at each aura_frequency interval afterwards SpellProcess runs DRCA.perc_aura and records the information in game_state.starlight_values.  

Then, for any buff_spell or offensive_spell that has starlight_threshold: defined, it will only select it if the current starlight level is equal to or above the defined threshold.  This lets me set, say:

```
- skill: Debilitation
  name: Fluoresce
  abbrev: flu
  mana: 15
  harmless: true
  min_threshold: 3
  starlight_threshold: 1
  cast_only_to_train: true
  cambrinth:
  - 10
  - 10
  
- skill: Targeted Magic
  name: Arbiter's Stylus
  abbrev: ars
  cyclic: true
  starlight_threshold: 2
  mana: 9

- skill: Targeted Magic
  name: Starcrash
  abbrev: star
  mana: 30
  starlight_threshold: 2
  min_threshold: 3
  
- skill: Targeted Magic
  name: Crystal Dart
  abbrev: crd
  starlight_threshold: 1
  mana: 18

- skill: Targeted Magic
  name: Strange Arrow
  abbrev: stra
  mana: 18
```
In other words, Starcrash and Arbiter's Stylus will only cast if my aura is at level 2 or higher, otherwise it will use Crystal Dart instead and continue to cast Fluoresce.  At level 0 (minimum) it will only use Strange Arrow, and reserve the remaining starlight for defensive spells like Blur and Trabe Chalice (threshold 0).  

All spells that use starlight are already marked in base-spells at starlight_threshold: 0 (except Enrichment, which is default 3 due to being uncastable below certain levels).  Also, if aura_frequency is not set, then all casting is identical to before -- it casts all spells until you run out of starlight, at which point it gets rid of them.

This has been tested on my trader and other non-trader magic users that I own.  I'll hunt around for other testing and feedback first.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -996,7 +996,11 @@ class SpellProcess
     @regalia_spell = settings.waggle_sets['regalia']['Regalia']
     echo("  @regalia_spell: #{@regalia_spell}") if $debug_mode_ct
 
+    @aura_frequency = settings.aura_frequency
+    echo("  @aura_frequency: #{@aura_frequency}") if $debug_mode_ct
+
     @perc_health_timer = Time.now
+    @aura_timer = 0
 
     @symbiosis_learning_threshold = settings.symbiosis_learning_threshold
     echo(" @symbiosis_learning_threshold: #{@symbiosis_learning_threshold}") if $debug_mode_ct
@@ -1073,6 +1077,7 @@ class SpellProcess
     check_ignite(game_state)
     check_rutilors_edge(game_state)
     check_health(game_state)
+    check_starlight(game_state)
     check_buffs(game_state)
     check_training(game_state)
     check_offensive(game_state)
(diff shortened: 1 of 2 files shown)


Tarhish:
As an aside, if there is some weirdo out there using trader starlight magic without being a trader, enough_starlight? will return on   
```
  return UserVars.sun['night'] unless DRStats.trader?
```
They can theoretically be cast on clear nights.  No guarantee it's going to be a CLEAR night from that vars entry, but that'll do for now.

Sarvatt:
should we add weather to moonwatch? 😆

Tarhish:
I'll have to feel a bit more masochistic than I am right now to do that.

Sarvatt:
Not being familiar with trader spells, would it make any sense to set a default starlight_threshold for any of the spells so its seamlessly used when its added? Do any of them require a set level?

Sarvatt:
hmm, maybe not, that would make them never cast looking at the code

Tarhish:
You mean a sort of default threshold level unless defined elsewhere?  That sounds very useful, but I don't think it could happen piecemeal since it gets merged in transform_settings.  I'll think about if there's an easy way to do that.

Sarvatt:
sorry to get you off on a tangent, thanks for double checking!

Tarhish:
That's what this is here for, yeah?  Thanks a lot for looking at it!

rpherbig:
Code looks fine to me. Let me know when you're happy with your testing.

Tarhish:
No word of any troubles with this, and I'm still not having any trouble with my trader or non-traders so I think this is good to go!

PR #3387, opened 2019-01-18
;sew return to room upon buying, and dampen pin-tapping

Tarhish:
Lots of little things today.

Two changes.  One is changing ;sew to return to the room you started in if you have to leave to go buy pins or thread.  Otherwise you continue crafting inside the crafting society.

Second is moving check_pins into the pins section so you don't tap your pins in between every sewing command.  When it gets the pins message, it'll check to see if you've got a pin collection, buy some if you don't, then come back.

I've been running this for about a month and had it run through pins and thread a number of times in that period.

### DIFF
--- a/sew.lic
+++ b/sew.lic
@@ -89,9 +89,11 @@ class Sew
 
   def check_thread
     return if exists?('cotton thread')
+    room = Room.current.id
     ensure_copper_on_hand(1000, @settings)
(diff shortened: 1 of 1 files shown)


Sarvatt:
Thanks for this! Two things that irk me that I never got around to fixing because I just stopped outfitting :) I don't see anything wrong with the code.

PR #3391, opened 2019-01-21
Fix tinker crafting-spell-training

Tarhish:
First, fixing my obvious if/unless mistake in tinker spell logic on crafting_training_spells.empty? caused by a persistent failure with my brain.

Second, I'm changing the structure of magic_routine here a bit, and I'd like to see if we want to do it to the others.

Basically, there is a case when magic_routine is entered with an already prepped training spell and nearly maxed magic mindstates -- this happens about every other cast once you do enough crafting.  So, magic_routine gets past needs_training, casts successfully, but then prepares a new spell even though your mind states are now too high to cast it the next time this is called.  The next time magic_routine would be called it would return on needs_training being nil before casting.

Normally this just means you let the prepped spell time out, but symbiosis stays up longer, and if you have a mix of some spells with symbiosis and some without then this could cause you to cast the wrong spell with symbioisis still up.  Rare, but it did occasionally give me some nerve damage when I was making the transition into being able to cast all of my schools with symbiosis.

Instead, this will store the spell being prepped in the preparing variable rather than making it true/false, then check to see if it needs to be cast before doing a needs_training check.

### DIFF
--- a/tinker.lic
+++ b/tinker.lic
@@ -281,29 +281,26 @@ class Tinker
   end
 
   def magic_routine
-    return unless @training_spells.empty?
+    return if @training_spells.empty?
+
+    if @preparing && Flags['tinker-magic-ready'] && mana > 40
+      crafting_cast_spell(@preparing, @settings)
+      Flags.reset('tinker-magic-ready')
+      @preparing = nil
+    end
 
     needs_training = %w[Warding Utility Augmentation]
                      .select { |skill| @training_spells[skill] }
                      .select { |skill| DRSkill.getxp(skill) < 31 }
                      .sort_by { |skill| DRSkill.getxp(skill) }.first
     return unless needs_training
-
-    data = @training_spells[needs_training]
-
-    if Flags['tinker-magic-ready'] && mana > 40
-      crafting_cast_spell(data, @settings)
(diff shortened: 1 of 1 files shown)


Tarhish:
@Sarvatt, got any opinions on this?  crafting_training_spells: have been waaay too useful to me, so I've made this change to everything I have that uses it and am going to see if anything snaps long-term.

I've been running with it for a few days now, but you might have a better idea if this might encounter some other problem down the line.

Sarvatt:
Thank you for spending more time figuring out the optimal way to do crafting_training_spells, it literally was just a one day thing i invented to help a trader. this looks good to me. i was just incapable of moving magic_routine to a common file and was commited to editing every instance of it if i changed one, there are a lot of them to fix up above and beyond this.

Tarhish:
As someone who keeps 3 crafting disciplines moving, among other things, yeah, I was super glad that you put this in.  I was curious though -- I haven't really put a lot of thought into commonizing it, but I was wondering if you think we should go in that direction or if you stopped because there was some roadblock along the way?

I noticed you made all the flags unique at some point, but I didn't know what that was meant to fix so I've just been following suit.

Tarhish:
Thanks for this!

PR #3394, opened 2019-01-25
combat-trainer.lic: Only apprase your gem_pouch_adjective pouches.

Sarvatt:
Closes #3389y

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2008,7 +2008,7 @@ class TrainerProcess
       appraise(game_state, 'careful')
     when 'App Pouch'
       retreat
-      bput('app my pouch quick', 'You sort', 'Appraise what', 'You cannot appraise')
+      bput("app my #{@gem_pouch_adjective} pouch quick", 'You sort', 'Appraise what', 'You cannot appraise')
     when 'Tactics'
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2008,7 +2008,7 @@ class TrainerProcess
       appraise(game_state, 'careful')
     when 'App Pouch'
       retreat
-      bput('app my pouch quick', 'You sort', 'Appraise what', 'You cannot appraise')
+      bput("app my #{@gem_pouch_adjective} pouch quick", 'You sort', 'Appraise what', 'You cannot appraise')
```
rpherbig: ❓ Does this process have access to `@gem_pouch_adjective`? Or does it need to be pulled from settings in the constructor?

Sarvatt: https://github.com/rpherbig/dr-scripts/blob/878da6a5899a3088fac1205a4c7888d2466625b8/combat-trainer.lic#L406 is in LootProcess, the same as this, it should

Sarvatt: no, I wrongly thought it was a different process, I will fix this.

combat-trainer.lic
```diff
@@ -416,6 +416,9 @@ class LootProcess
     @loot_specials = settings.loot_specials
     echo("  @loot_specials: #{@loot_specials}") if $debug_mode_ct
 
+    @gem_pouch_adjective = settings.gem_pouch_adjective
```
rpherbig: ❓ Is this a duplicate of line 406?

It looks like this adds the setting to `LootProcess` (for a second time), but the new use of the setting is in `TrainerProcess`. I would have expected to see these two lines added around line 1989 instead.

Issue #3406, opened 2019-02-04
DRinfomon lnet chat bug


trall541:
If someone posts "Log-on system converted -1% of your character's field experience into earned rank." or something similar DRinfomon does a exp full...

rpherbig:
#3407

PR #3407, opened 2019-02-04
Anchor drinfomon message triggered by lnet

Tarhish:
Due to REXP discussion we found a global lnet trigger for making everyone do an EXP ALL.

Not exactly critical, but this change no longer triggers on chat messages but does trigger on login messages.


Tarhish:
related to https://github.com/rpherbig/dr-scripts/issues/3406

PR #3415, opened 2019-02-07
Support resuming knitted goods and fix DR caravan bug

Tarhish:
Added resume_knit that triggers if called as 'sew resume needles', or with knitting needles in hand.  Analyzing a pair of knitting needles with something on it gives like 40 lines of scroll and TWO separate RTs from one action, which screws up bput entirely.

Tested on knitted items resumed at each stage (push/knit/turn/cast) and on empty needles.

Also, encountered a bug of some sort with the PUSH command in caravans.  PUSH MY Anything inside a caravan while someone is online with a name beginning with 'My' always results in 'Myfoo is not here!'

If it runs into that case it subs out 'my ' and just becomes 'push needles'.

### DIFF
--- a/sew.lic
+++ b/sew.lic
@@ -200,8 +200,30 @@ class Sew
     do_seal !wax
   end
 
+  def resume_knit
+    bput("analyze my knitting needles",
+         /You analyze both the needles and an? unfinished knitted .+ (.+) on them/,
+         'This appears to be a crafting tool',
+         'Roundtime')
+    pause 1
+    results = reget(20)
+    unless results.to_s =~ /You analyze both the needles and an? unfinished knitted .+ (.+) on them/
+      echo "Nothing found on the needles.  Exiting"
+      exit
+    end
+    @noun = Regexp.last_match(1).to_s
+    echo "Found an unfinished #{@noun} on the needles."
+    do_knit('push my knitting needles') if results.to_s.include?('purl stitching')
+    do_knit('knit my knitting needles') if results.to_s.include?('more knitting')
+    do_knit('turn my knitting needles') if results.to_s.include?('need to be turned')
+    cast_and_finish if results.to_s.include?('must be cast')
+    exit
+  end
+
   def resume
     waitrt?
+    resume_knit if @noun.include?('needle') || right_hand.include?('knitting needles') || left_hand.include?('knitting needles')
+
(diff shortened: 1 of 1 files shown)


Tarhish:
To be clear, the error (and fix) looks like:

[sew]>push my needles
Mysto is not here!
>
[sew]>push needles
You control the yarn and loop it with one hand and then throw it off onto your needles, adding perfect purl stitching along each row.
Roundtime: 14 sec.

Normally I doubt anyone would run into this, but I've got my trade script sewing on caravan trips, and this'll need to be fixed prior to submitting it.

sew.lic
```diff
@@ -360,7 +382,8 @@ class Sew
               'You are already knitting',
               'Roundtime',
               'That tool does not seem suitable for that task.',
-              'You need a larger amount of material')
+              'You need a larger amount of material',
+              'is not here!')
```
jandersson: Food for thought: This could be added to bput itself to make it auto-retry, much like it does with playing an instrument.

jandersson: Maybe that will just result in spam though?

Tarhish:
I've been thinking about it, but it's just such a weird issue.  This only seems to happen inside caravans, and only with PUSH, but the only time I'd know how to fix it is if you're using MY.  It's not a temporary thing, either, it always fails if one of these My- people is online.

I should check to see if PUSHing any other word fragments give this error, or if it's just 'my'.  I bet there are some other things in people's persons that you'd need to PUSH from time to time, but if that problem crops up for other words then I won't know what to replace it with.

sew.lic
Tarhish: I've been thinking about it, but it's just such a weird issue.  This only seems to happen inside caravans, and only with PUSH, but the only time I'd know how to fix it is if you're using MY.  It's not a temporary thing, either, it always fails if one of these My- people is online.

I should check to see if PUSHing any other word fragments give this error, or if it's just 'my'.  I bet there are some other things in people's persons that you'd need to PUSH from time to time, but if that problem crops up for other words then I won't know what to replace it with.

Tarhish: Oh, I see it now.  PUSH Person is syntax to kick someone out of the caravan.  Of course it looks at who's online.  So Push anything would probably give the same error.

Makes sense now.  Just a parser priority problem

PR #3417, opened 2019-02-07
Create Kyrtoth-setup.yaml

daedelis:
Was requested to add yaml to repo.


rpherbig:
Thank you for contributing!

Before I can merge this, you will need to update the test that checks that all settings are in `base.yaml`.

You can do this by adding `Kyrtoth` to this line: https://github.com/rpherbig/dr-scripts/blob/master/test/test_valid_yaml.rb#L19 (the list of YAMLs to exclude from the test)

daedelis:
Creating new pull request with requested udpates

PR #3419, opened 2019-02-09
Add snipe messaging to hunting-buddy roomcheck

Tarhish:
LNET suggestion - An arrow is as good as an 'excuse me', in my opinion.

If hunting-buddy sees snipe messaging it will move on, since it's unlikely the sniper will break cover in the time period allotted for room_check.


Sarvatt:
cant see how this would hurt anything and seems like a good idea to me.

PR #3427, opened 2019-02-13
Merge heal-herb into heal-remedy

andymanak:
This replaces any references to heal-herb with heal-remedy.  Heal-remedy now checks if herbs: is defined in the user-setup.yaml to determine which herb data file to use.


jandersson:
This is cool! thanks for the contribution

PR #3428, opened 2019-02-14
Provide mechanism to use divination tools other than bones.

lesharris:
# What is this?
Currently astrology.lic only provides support for bones as a divination tool. This pull adds a mechanism for using any divination tool currently in the game.

### Celestial Charts
```yaml
divination_tool:
  name: 'vellum.charts'
  container: 'case'
```
### Bones
```yaml
divination_tool:
  name: 'enruned.bones'
  container: 'belt'
  tied: true
```
### Mirror
```yaml
divination_tool:
  name: 'clouded mirror'
  container: 'cloak'
```
### Bowl
```yaml
divination_tool:
  name: 'clay bowl'
  container: 'backpack'
```
### Prism
```yaml
divination_tool:
  name: 'glittering prism'
  container: 'backpack'
```
## Current Bones Functionality
This pull leaves the current bone rolling functionality in place to not break users' existing configurations. Ideally it would be marked as deprecated and removed in a future version.  That said, it's not really hurting anything by still having it in there.

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -24,6 +24,9 @@ class Astrology
 
     @equipment_manager = EquipmentManager.new
     @constellations = get_data('constellations').constellations
+    @divination_tool = settings.divination_tool['name']
+    @divination_tool_container = settings.divination_tool['container']
+    @divination_tool_tied = settings.divination_tool['tied']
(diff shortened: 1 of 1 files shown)


jandersson:
fantastic, thank you

PR #3429, opened 2019-02-15
Update base-spells.yaml - Regalia cast messaging

BinuDR:
Regalia cast messaging changes depending on the available source of light.

I updated the match to this  /^Cupping your palms skyward, you bask in a cascade.*$/

Cupping your palms skyward, you bask in a cascade of moonlight.
The moonlight coalesces into a rough-cut crystal scale balaclava atop your head!
The moonlight coalesces into some rough-cut crystal scale gloves around your hands!
The moonlight coalesces into a rough-cut crystal scale hauberk around your body!
Roundtime: 2 sec.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -160,7 +160,7 @@ cast_messages:
 - ^Wisps of spectral flame whorl.*$
 - ^You cup your palms skyward to bask in.*$
 - ^Incited by the spell, your starlight aura
-- ^Cupping your palms skyward, you bask in a cascade of starlight.*$
+- ^Cupping your palms skyward, you bask in a cascade.*$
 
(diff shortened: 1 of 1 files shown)


Tarhish:
It... It does?!

BinuDR:
Apparently.  Limes sent me the original new messaging which I added, then yesterday he told me that it still wasn't recognizing his cast message so he sent me the new one and it was moonlight instead of starlight.

Tarhish:
I've cast it a thousand times and not seen that, but maaybe that's because I've never cast it without Noumena up.  I suppose moonlight can pierce clouds, but starlight can't.

Good to know, because I don't think I would have ever found that.

PR #3430, opened 2019-02-15
Update ;trade for training and multiple contracts

Tarhish:
And now for the huge, time-consuming update literally twos of people were asking for.

Previously ;trade went to an outpost, got a contract, then drove to that destination outpost, got another one, rinse and repeat.  This uses those methods to expand that, but it still only runs in Zoluren.  More to come.

Changes:
- Now goes around Zoluren visiting the closest nearest outpost it can either get a contract from or deliver a contract to. It only visits an outpost if it has business there, but it will accumulate multiple contracts in the process.
- Uses LEAD TO functions to train larger RT skills on the road.  If caravan_interior: true is set, it'll ride inside when useful.
- If enough contracts pile up, it will lock in on that destination outpost and head towards it, stopping at any useful outpost along the way (Primarily Leth and Dirge)
- ;trade room_id   - takes your caravan to a specified room_id
- ;trade outpost <town> - takes you to the trade outpost at that town
- ;trade closeup   - delivers all current outstanding contracts and returns the caravan to avoid fees.

- ;trade can ALWAYS be started in the field. Scans your contracts on startup, so you can restart after a DC, though you may have to leave the room and jump back in so drinfomon loads up your room_objs. The only requirement is that you start it in the same room as your caravan.
- Trains the following skills atm:
  #DONE WHILE WALKING THE CARAVAN - Only done every other step to avoid increasing travel time at all
  Utility: Uses crafting_training_spells
  Augmentation:  Uses crafting_training_spells
  Warding:  Uses crafting_training_spells
  Performance: Zills
  Appraisal: Runs ;appraisal while being led, or appraises a pouch on your person while walking
  Mechanical Lore: Braids grass from your feedbag - feedbag automatically filled up over time
  Attunement:  PERC MANAs while walking 

  #ONLY WHILE BEING LED ON A TRADE ROUTE OR INSIDE THE CARAVAN
  First Aid: Either a compendium or a textbook with textbook: true
  Outfitting:  Knits - need knitting needles - replenishes yarn every time you stop at Crossing up to yarn_quantity
  Engineering:  Shapes - needs shaping tools and lumber_quantity - replenishes each time you stop at Crossing
  Outdoorsmanship: Collects rocks while being led or riding in a caravan
  Perception: Collects rocks while being led or riding in a caravan
  Locksmithing: uses ;lockbox if you have those settings set, or pet boxes otherwise.

- offloads coins in excess of your caravan_coins_on_hand limit when stopping for business in Crossing
- Dumps expired contracts and crates.
- Refills feedbags over time
- trains on the Leth/Crossing ferry

Usage:
Rent a caravan, then run ;trade in the room with it. When you're done, kill ;trade and run ;trade closeup.

### DIFF
--- a/data/base-town.yaml
+++ b/data/base-town.yaml
@@ -31,10 +31,6 @@ Crossing:
   gemshop:
     id: 4652
     name: appraiser
-  pawnshop:
-    id: 8261
-    name: Cormyn
-    
   tannery:
     id: 8266
     name: Falken
@@ -140,6 +136,14 @@ Crossing:
     Warrior Mage:
       name: Gauthus
       id: 7887
+Darkling Wood:
+  currency: dokoras
+  shipment_clerk:
+    id: 14636
+  trade_minister:
+    id: 14636
+  trader_outpost:
+    id: 2702
 Dirge:
   trade_minister:
     id: 15175
@@ -148,6 +152,14 @@ Dirge:
     id: 8919
   trader_outpost:
     id: 1307
+Fayrin's Rest:
+  currency: dokoras
+  shipment_clerk:
+    id: 19278
+  trade_minister:
+    id: 19278
+  trader_outpost:
+    id: 2840
 Leth Deriel:
   currency: kronars
   leather_repair:
@@ -193,9 +205,6 @@ Shard:
     name:
     - Fatimi
     - Veilex
-  pawnshop:
-    id: 9759
-    name: Aelik
   tannery:
     id: 19340
     name: Tremagis
@@ -307,6 +316,19 @@ Shard:
     Warrior Mage:
       name: Melear
       id: 11197
+  shipment_clerk:
+    id: 8193
+  trade_minister:
+    id: 8197
+  trader_outpost:
+    id: 8191
+Steelclaw Clan:
+  shipment_clerk:
+    id: 19399
+  trade_minister:
+    id: 19399
+  trader_outpost:
+    id: 2874
 Stone Clan:
   currency: kronars
   shipment_clerk:
@@ -338,9 +360,6 @@ Riverhaven:
   gemshop:
     id: 7955
     name: Anthelorm
-  pawnshop:
-    id: 14337
-    name: Ioun
   guard_house:
     id: 7882
   tannery:
@@ -573,7 +592,7 @@ Langenfirth:
     - 12269
     - 12215
     - 12209
-  guild_leaders: 
+  guild_leaders:
     << : *therengia_guild_leaders
 Ratha:
   currency: lirums
@@ -807,6 +826,12 @@ Throne City:
   npc_empath: # None
 Hibarnhvidar:
   currency: dokoras
+  shipment_clerk:
+    id: 7412
+  trade_minister:
+    id: 7412
+  trader_outpost:
+    id: 3903
   leather_repair:
     id: 11206
     name: Ladar
@@ -895,8 +920,22 @@ Hibarnhvidar:
     Warrior Mage:
       name: Jorent
       id: 11498
+Raven's Point:
+  currency: dokoras
+  shipment_clerk:
+    id: 14642
+  trade_minister:
+    id: 14642
+  trader_outpost:
+    id: 4440
 Boar Clan:
   currency: dokoras
+  shipment_clerk:
+    id: 13675
+  trade_minister:
+    id: 13675
+  trader_outpost:
+    id: 4106
   leather_repair:
     id: 12174
     name: Tuzra
@@ -1009,11 +1048,21 @@ Muspar'i:
       id:
 Ain Ghazal:
   currency: dokoras
+  shipment_clerk:
+    id: 14638
+  trade_minister:
+    id: 14638
+  trader_outpost:
+    id: 11033
+  deposit:
+    id: 14640
+  exchange:
+    id: 14641
   locksmithing:
     id: 13190
   guild_leaders:
     Thief:
       name: Ivitha
       id: 13663
-  #Needs attunement_rooms, and premium bank/gem shop.
+  #Needs attunement_rooms
 
--- a/data/base-town.yaml
+++ b/data/base-town.yaml
@@ -31,6 +31,9 @@ Crossing:
   gemshop:
     id: 4652
     name: appraiser
+  pawnshop:
+    id: 8261
+    name: Cormyn
   tannery:
     id: 8266
     name: Falken
@@ -205,6 +208,9 @@ Shard:
     name:
     - Fatimi
     - Veilex
+  pawnshop:
+    id: 9759
+    name: Aelik
   tannery:
     id: 19340
     name: Tremagis
@@ -360,6 +366,9 @@ Riverhaven:
   gemshop:
     id: 7955
     name: Anthelorm
+  pawnshop:
+    id: 14337
+    name: Ioun
   guard_house:
     id: 7882
   tannery:
--- a/profiles/base-empty.yaml
+++ b/profiles/base-empty.yaml
@@ -62,4 +62,4 @@ empty_values:
   crafting_training_spells: {}
   restock_shop: {}
   theurgy_supply_levels: {}
-  caravan_training_skill: {}
+  caravan_training_skills: {}
--- a/profiles/base-empty.yaml
+++ b/profiles/base-empty.yaml
@@ -62,3 +62,4 @@ empty_values:
   crafting_training_spells: {}
   restock_shop: {}
   theurgy_supply_levels: {}
+  caravan_training_skill: {}
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -524,8 +524,9 @@ caravan_training_skills:
   Attunement: 100
   Outdoorsmanship: 5
   Locksmithing: 30
-  Perception: 15
-  Outfitting: 200
+  Perception: 45
+  First Aid: 120
+  # Outfitting: 200
   # Engineering: 200
 
 
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -488,6 +488,8 @@ crafting_training_spells:
 craft_max_mindstate: 31
 # amount of yarn to keep on hand for craft.lic, don't exceed 300
 yarn_quantity: 100
+# amount of balsa lumber to keep on hand in trade.lic.
+lumber_quantity: 15
 
 
 # Workorder settings
@@ -513,8 +515,9 @@ caravan_coins_on_hand: 9555
 caravan_interior: false
 # skills and cooldowns to train in ;trade.  See https://github.com/rpherbig/dr-scripts/wiki/Trader-Tutorials for valid skills and what they do
 caravan_training_skills:
-  Augmentation: 5 
-  Warding: 5
+  # Augmentation: 5 
+  # Warding: 5
+  # Utility: 5
   Performance: 120
   Appraisal: 60
   Mechanical Lore: 5
@@ -523,6 +526,7 @@ caravan_training_skills:
   Locksmithing: 30
   Perception: 15
   Outfitting: 200
+  # Engineering: 200
 
 
 # Lumber settings
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -504,6 +504,23 @@ workorder_max_items: 3
 workorder_recipes:
 workorder_cash_on_hand:
 
+#CARAVAN SETTINGS
+# container holding caravan contracts for ;trade
+trade_contract_container: backpack
+# number of copper to carry on-hand while running ;trade
+caravan_coins_on_hand: 15555
+# skills and cooldowns to train in ;trade.  See https://github.com/rpherbig/dr-scripts/wiki/Trader-Tutorials for valid skills and what they do
+caravan_training_skills:
+  Augmentation: 5 
+  Warding: 5
+  Performance: 120
+  Appraisal: 60
+  Mechanical Lore: 5
+  Attunement: 100
+  Outdoorsmanship: 5
+  Locksmithing: 30
+  Perception: 15
+  Outfitting: 200
 
 
 # Lumber settings
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -508,7 +508,9 @@ workorder_cash_on_hand:
 # container holding caravan contracts for ;trade
 trade_contract_container: backpack
 # number of copper to carry on-hand while running ;trade
-caravan_coins_on_hand: 15555
+caravan_coins_on_hand: 9555
+# Have you purchased a 750plat interior from the trade minister?
+caravan_interior: false
 # skills and cooldowns to train in ;trade.  See https://github.com/rpherbig/dr-scripts/wiki/Trader-Tutorials for valid skills and what they do
 caravan_training_skills:
   Augmentation: 5 
--- a/trade.lic
+++ b/trade.lic
@@ -124,6 +124,10 @@ class Trade
       echo("No contract container found. Define it in your settings as trade_contract_container: Exiting.")
       exit
     end
+    unless exists?(@craft_bag)
+      echo("No crafting container found for storing items.  This will cause problems if you're doing Outfitting or Engineering.")
+      pause 2
+    end
     @settings.storage_containers.each { |container| fput("open my #{container}") }
     @box = nil unless exists?(@box)
     if @caravan_training_skills.include?('Performance') && !exists?('zills')
@@ -225,10 +229,13 @@ class Trade
   end
 
   def setup_town_data # builds @trading_towns from town_data and will later handle multiple provinces
-    zoluren_towns = ['Wolf Clan', 'Tiger Clan', 'Crossing', 'Arthe Dale', 'Stone Clan', 'Dirge', 'Leth Deriel']
+    zoluren_towns = { 'Wolf Clan' => 2, 'Tiger Clan' => 2, 'Crossing' => 2, 'Arthe Dale' => 2, 'Stone Clan' => 2, 'Dirge' => 2, 'Leth Deriel' => 3 }
     @trading_towns = {}
-    zoluren_towns.each_with_index { |_info, index| @trading_towns[zoluren_towns[index]] = @town_data[zoluren_towns[index]] } #reduce town_data to only the towns in question
-    @trading_towns.each_with_index { |_info, index| @trading_towns[zoluren_towns[index]]['contracts_to'] = @trading_towns[zoluren_towns[index]]['contracts_from'] = 0 }
+    zoluren_towns.each do |town_name, threshold|
+      @trading_towns[town_name] = @town_data[town_name]
+      @trading_towns[town_name]['contracts_to'] = @trading_towns[town_name]['contracts_from'] = 0
+      @trading_towns[town_name]['contract_threshold'] = threshold
+    end
   end
 
   def mark_towns # records some information about contracts to and from towns
@@ -254,10 +261,10 @@ class Trade
     end
 	end
 
-  def check_for_emergency_delivery # if there are 4 or more pieces of business to do at a town, (contracts to deliver + contract to pick up) set it for emergency delivery
+  def check_for_emergency_delivery # go towards a town if you have a certain amount of business there
     echo "checking for emergency" if debugging?
     @trading_towns.find do |name, info|
-      if info['contracts_to'] - info['contracts_from'] >= 3
+      if info['contracts_to'] - info['contracts_from'] >= info['contract_threshold']
         @emergency_delivery_town = name
         echo "Emergency delivery needed to #{@emergency_delivery_town}"
       end
@@ -1417,6 +1424,7 @@ class Trade
   def outfitting_cleanup # sanitizes outfitting_routine
     return unless @training_token == 'Outfitting'
     stop_script('sew') if Script.running?('sew')
+    pause 1
     stow_item('knitting needle')
     bput('stow my wool yarn', 'You put', 'Stow what')
     be_outside_caravan
@@ -1464,9 +1472,9 @@ class Trade
     if left_hand.include?(@crafted_item.split.last) || right_hand.include?(@crafted_item.split.last)
       bput("drop my #{@crafted_item.split.last}", 'You drop', 'You place', 'What were', 'You spread') unless bput("tap my #{@crafted_item.split.last}", 'unfinished', 'You tap', 'I could not find') == 'unfinished'
     end
-    item_container = @settings.storage_containers.find { |container| bput("tap my #{@crafted_item.split.last} in my #{container}", 'unfinished', 'You tap', 'I could not find') == 'unfinished' }
-    if item_container
-      case bput("get my #{@crafted_item.split.last} from my #{item_container}", 'You get', 'What were', 'But that')
+
+    if bput("tap my #{@crafted_item.split.last} in my #{@craft_bag}", 'unfinished', 'You tap', 'I could not find') == 'unfinished'
+      case bput("get my #{@crafted_item.split.last} from my #{@craft_bag}", 'You get', 'What were', 'But that')
       when 'You get'
         echo "Engineering found an uncompleted item #{@crafted_item}"
         be_inside_caravan
@@ -1492,6 +1500,7 @@ class Trade
   def engineering_cleanup # sanitizes engineering
     return unless @training_token == 'Engineering'
     stop_script('shape') if Script.running?('shape')
+    pause 1
     bput("drop my #{@crafted_item.split.last}", 'You drop', 'You place', 'What were', 'You spread') unless bput("tap my #{@crafted_item.split.last}", 'unfinished', 'You tap', 'I could not find') == 'unfinished'
     stow_item(right_hand) if right_hand
     stow_item(left_hand) if left_hand
@@ -1549,7 +1558,7 @@ class Trade
     @cooldown_timers['Performance'] = Time.now
     waitrt?
   end
-  
+
   def check_wounds
     room = Room.current.id
     wait_for_script_to_complete('safe-room')
--- a/trade.lic
+++ b/trade.lic
@@ -117,8 +117,16 @@ class Trade
     end
     @settings.storage_containers.each { |container| fput("open my #{container}") }
     @box = nil unless exists?(@box)
-    unless exists?('zills')
+    if @caravan_training_skills.include?('Performance') && !exists?('zills')
+      echo "NO ZILLS FOUND.  REMOVING PERFORMANCE FROM TRAINING LIST."
       wipe_skill('Performance', @walk_training_skills)
+      wipe_skill('Performance', @lead_training_skills)
+    end
+    if @caravan_training_skills.include?('First Aid')
+      unless @settings.textbook ? exists?(@settings.textbook_type) : exists?('compendium')
+        echo "NO COMPENDIUM OR TEXTBOOK FOUND.  CHECK YOUR ;FIRST-AID SETTINGS"
+        wipe_skill('First Aid', @lead_training_skills) 
+      end
     end
   end
 
@@ -675,12 +683,17 @@ class Trade
     echo "Delivering Contracts"
     walk_to(clerk_id)
     clerk_messages = [
+      'What have you done with the goods',
       '^The \w+ clerk finds everything in order with your merchandise'
     ]
     contracts = find_contracts(@contract_container)
     current_town = find_town(Room.current.id).split.first
     until bput("get my #{current_town} contract from my #{@contract_container}", 'You get', 'What were you' ) == 'What were you'
-	    bput('give my contract to clerk', clerk_messages)
+      case bput('give my contract to clerk', clerk_messages)
+      when 'What have you done with the goods'
+        # Potential to do an immediate parse_contract and force_contract_from that town to 1 so you don't go back there for nothing.
+        dispose_trash('contract')
+      end
     end
   end
 
@@ -1476,8 +1489,8 @@ class Trade
   def performance_routine # runs ;performance for 2 minutes.  Will run ;first-aid too, if it hasn't been done very recently.
     return unless @training_token == 'Performance'
     echo "performance routine" if debugging?
-    if @caravan_being_led && check_ability?('First Aid', 180)
-      @training_token = 'First Aid' if @settings.textbook ? exists?(@settings.textbook_type) : exists?('compendium')
+    if @lead_training_skills.include?('First Aid') && @caravan_being_led && check_ability?('First Aid', 180) 
+      @training_token = 'First Aid'
       return
     end
     @cooldown_timers['perf-timer'] ||= Time.now
@@ -1515,6 +1528,7 @@ class Trade
     end
     if @caravan_training_skills.include?('Engineering') && @times_engineering > 4
       wait_for_script_to_complete('workorders', ['shaping', 'repair'])
+      @times_engineering = 0
     end
   end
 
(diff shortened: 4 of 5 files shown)


Tarhish:
I'm not expecting a pull for a while.  This is the first version I've run for multiple days without encountering any unintended behavior, so I'll try to get more people who may be interested in testing it first.

jandersson:
This is fantastic and the dream when i uploaded the original version :) Thanks for taking the time to make this greater than a proof of concept script

validate.lic
```diff
@@ -712,6 +712,22 @@ class DRYamlValidator
       warn("#{list} contains the following spell names not found in data/base_spells, #{name_warnings} Check spelling and capitalization.") unless name_warnings.empty?
     end
   end
+  
+  def assert_that_caravan_training_skills_are_valid(settings)
+    return unless settings.caravan_training_skills
```
jandersson: Check if its empty instead of if it exists. Since it is in base-empty, the empty hash will exist.
```
2.0.0-p648 :001 > test = {}
 => {}
2.0.0-p648 :002 > print "hello" if test
hello => nil
2.0.0-p648 :003 > puts "hello" if test
hello
 => nil
2.0.0-p648 :004 > test.empty?
 => true
```

Tarhish: Ah-hah, I did not try that after adding that into base-empty.  Thanks!

Tarhish:
Yeah, there's a lot I was only able to do because I followed along with the methods you already had in there.  Regardless of how useful it is or isn't, this was a very fun project, and it can definitely keep all the non-combat things a trader needs to circle moving.

Sarvatt:
you may want to override settings.appraisal_training so art isn't an option instead of assuming people will have removed that

trade.lic
```diff
@@ -1,78 +1,419 @@
...
     @equipment_manager = EquipmentManager.new
     @settings = get_settings
     @contract_container = @settings.trade_contract_container
-    @hometown = @settings.hometown
-    bput("open my #{@contract_container}", 'You open', 'That is already')
-    @equipment_manager.empty_hands
+    @hub_city = 'Crossing'
+    @emergency_delivery_town = nil
+    @current_caravan_destination = nil
+    @caravan_coins_on_hand = @settings.caravan_coins_on_hand.to_i || 25555
```
jandersson: Im not sure that the `25555` will ever be loaded, since this is a base setting. If base doesn't load then you have big problems.

Tarhish: Yep, didn't catch that one when I was removing temporary declarations

trade.lic
```diff
@@ -1,78 +1,419 @@
...
+    @caravan_interior = @settings.caravan_interior || false
```
jandersson: I would add this to base with the default as false.

Tarhish: Yep, that was the intention -- I didn't look closely enough at base when I was making them line up.  Thanks for catching these!  I'll do another similar run-through shortly

Tarhish:
The only weird behavior I've seen so far is that this has a tendency (~on average once every 10 hours of runtime?) to lock up the frontend then instant it steps out of the caravan.  It looks the same way that that ;roomnumbers bug does, where it just stops responding or taking input, except it happens even when roomnumbers is killed.

I find riding around in the caravan nice enough that I haven't turned off caravan_interior: true, but there's something funky going on in the back end with that weird transition of going from a 'not-room' to a 'room'.

trade.lic
jandersson: Sweet, just trying to give you feedback from a fresh pair of eyes perspective :)

Tarhish: I'm never displeased to see people catching my mistakes or suggesting a different way to do things.  Definitely a lot better than after it goes live.

jandersson:
Im loving this so ill reactivate my trader and give it a whirl

Sarvatt:
maybe just a return in lead_appraisal if appraisal_training contains art would be easiest, no need to tie this huge change up with making ;appraisal take args or something :) now i want to make a trader!

Tarhish:
I was thinking about art appraisal and maybe adding a warning not to include it to validate, but really I know what I want them to appraise, so overriding it sounds way better

Tarhish:
Oh, right, it just hit me.  I did go down this path in my head and realized I'd need to mess with ;appraisal.  I just forgot to go forward with either of those things.

Tarhish:
Yeah, actually, looking at ;appraisal, it's so compact.  Might be best to just rob and duplicate what I need instead of starting it as a script

jandersson:
```

**Comment by jandersson** [comment]
Doesnt seem to recognize non-caravan responses:
```
[trade]>prep NON 5                                                                                                                    
You raise your arms skyward, chanting the equation of the Nonchalance spell.                                                          
>                                                                                                                                     
You feel fully rested.                                                                                                                
>                                                                                                                                     
Your speckled bullock arrives from the southeast, following you.                                                                      
[trade]>tell bullock to wait                                                                                                          
You grab hold of your speckled bullock's harness and make it wait.                                                                    
>                                                                                                                                     
>                                                                                                                                     
[trade: *** No match was found after 15 seconds, dumping info]                                                                        
[trade: messages seen length: 11]                                                                                                     
[trade: message: You feel fully prepared to cast your spell.]                                                                         
[trade: message: Nonchalance  (10 roisaen)]                                                                                           
[trade: message: Finesse  (7 roisaen)]                                                                                                
[trade: message: Noumena  (10 roisaen)]                                                                                               
[trade: message: [DRPrime]-DR:Gigondas: "ha casaba."]                                                                                 
[trade: message:  * Unbelievable Lisae Valentyn just popped into existence.]                                                          
[trade: message: Nonchalance  (10 roisaen)]                                                                                           
[trade: message: Finesse  (7 roisaen)]                                                                                                
[trade: message: Noumena  (10 roisaen)]                                                                                               
[trade: message: [DRPrime]-DR:Kalinandra: "don't forget a nice casaba"]                                                               
[trade: message: You grab hold of your speckled bullock's harness and make it wait.]
[trade: checked against [/You pass on the order to wait to your driver, who makes sure your .* does your bidding/i]]                  
[trade: for command tell bullock to wait]                                                                                             
[trade]>release symb                                                                                                                  
You pause for a moment as the details of the Activate symbiosis fade from your mind.                                                  
You release the Activate symbiosis.                                                                                                   
[trade]>release spell                                                                                                                 
You let your concentration lapse and feel the spell's energies dissipate.                                                             
[trade]>release mana                                                                                                                  
```

**Comment by Tarhish** [comment]
I'll admit.  I didn't even touch non-caravan stuff considering the difference between the two was 200 kronar, but I should definitely handle it one way or another.  You already had a flag in there for catching when you'd need to upgrade to a real caravan, so it should be pretty trivial.

**Comment by jandersson** [comment]
Its just what i got when asking for a contract

**Comment by jandersson** [comment]
Getting stuck on the look part too
```
[trade]>look speckled bullock                                                          
>                                                                                      
>                                                                                      
You carefully go over the crates loaded on your speckled bullock.                      
  A sturdy crate is for a contract from The Crossing and destined for Arthe Dale.      
  A sturdy crate is for a contract from Wolf Clan and destined for Stone Clan.         
  Some wicker panniers to carry your items.                                            
The speckled bullock looks very hungry.                                                
>                                                                                      
```

**Comment by Tarhish** [comment]
I think this should handle non-caravan animals.  I've been running it on a second trader for a while now (Apparently once you buy a caravan interior you can't rent mules anymore).

I don't think running with a mule is ever reaaally a good idea, since the only time it'll be necessary to save you money is before circle ~5?  But you get such low trading experience around that time for some reason that it's probably better to wait to do caravans anyway.  And you can't be led anywhere by a mule, so you can't do higher training.  But at least now it handles it.

Next I'll retouch appraisal and a couple other things.  I think it should be possible to make it more self-sufficient, like, "Run trade for 2 hours then shut down, then move on to combat stuff when you're done"

**Comment by jandersson** [comment]
Running with a caravan this time. Havent run into any problems, seems good to me!

**Comment by Tarhish** [comment]
Nice.  I was sort of looking into expanding this into other provinces, which shouuuldn't be too hard.  But I think there's a little problem with map stringprocs that use athletics.  I seem to remember when I was trying things out around Haven it was trying to drive the caravan into the river, which implied he was passing the athletics check, but that was from months ago, prior to this complete rewrite.  I might have been screwing something else up.

Does your trader have what it takes to swim Segoltha?  If he does and he hasn't been trying to do so then it's probably fine.  Otherwise I'll bring my caravan up towards Haven and try crossing the river to see if it takes the ferry or not.

**Comment by jandersson** [comment]
One idea might be to temporarily set the athletics skill uservar to 0 so none of those routes get taken. Something like that.

**Comment by Tarhish** [comment]
Yeah, that Dijkstra method is down deep in lich itself isn't it?  But the way you had compute_path set up I don't think there'd be a problem with us kicking it down to 0 before it runs that algorithm.  I'll have to try that out, but definitely I'd think this wouldn't work as is for traders with higher Athletics.

File: profiles/base-empty.yaml [hunk 4]
```diff
@@ -62,3 +62,4 @@ empty_values:
   crafting_training_spells: {}
   restock_shop: {}
   theurgy_supply_levels: {}
+  caravan_training_skill: {}
```
rpherbig: ❓ Should this be `caravan_training_skills` (note the 's')?

validate.lic
```diff
@@ -712,6 +712,22 @@ class DRYamlValidator
       warn("#{list} contains the following spell names not found in data/base_spells, #{name_warnings} Check spelling and capitalization.") unless name_warnings.empty?
     end
   end
+  
+  def assert_that_caravan_training_skills_are_valid(settings)
+    return unless settings.caravan_training_skills.empty?
+    valid_caravan_skills = ['Augmentation', 'Utility', 'Warding', 'Attunement', 'Performance', 'Appraisal', 'Mechanical Lore',]
```
rpherbig: 💡 Can `valid_caravan_skills ` be removed?

rpherbig:
@jandersson @Sarvatt There's a ton of changes here and I don't know anything about Traders. Can you two be my reviewers/testers for this PR?

profiles/base.yaml
```diff
@@ -504,6 +504,23 @@ workorder_max_items: 3
...
+  Augmentation: 5 
+  Warding: 5
+  Performance: 120
+  Appraisal: 60
+  Mechanical Lore: 5
+  Attunement: 100
+  Outdoorsmanship: 5
+  Locksmithing: 30
+  Perception: 15
+  Outfitting: 200
```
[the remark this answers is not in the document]
Tarhish: It should definitely be false.  Jonas caught that but I haven't adjusted it yet

profiles/base-empty.yaml
Tarhish: Doh

validate.lic
Tarhish: Yep, I need to do a full review to remove a couple vestigial variables that have been called out and that's definitely one of them

profiles/base.yaml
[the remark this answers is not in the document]
Tarhish: I want to adjust these defaults, too.  I feel like it should be set to only base include skills that you can do without any tools, then you can add more later.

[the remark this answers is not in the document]
jandersson: nil/empty for the interior, right? Since interiors are only for you fancy folks who own your own personal caravans.

jandersson:
Its released my symbiosis research from doing too many `release symbiosis` commands. Is this something we can fix in this script or should it be delegated to a common-arcana issue?
```
 [go2: travel time: 0:00:00]                                                                                                
 --- Lich: go2 has exited.                                                                                                  
 [trade]>tap my contract                                                                                                    
 [trade]>recall caravan                                                                                                     
 You seem to recall that you left your two auroch caravan right behind you.                                                 
 [trade]>recall caravan                                                                                                     
 You seem to recall that you left your two auroch caravan right behind you.                                                 
 [trade]>tell caravan to follow                                                                                             
 You pass on the order to follow to your driver, who makes sure your two auroch caravan does your bidding.                  
 >                                                                                                                          
 [trade]>recall caravan                                                                                                     
 >                                                                                                                          
 You seem to recall that you left your two auroch caravan right behind you.                                                 
 [trade]>tell caravan to wait                                                                                               
 You pass on the order to wait to your driver, who makes sure your two auroch caravan does your bidding.                    
 [trade]>release symb                                                                                                       
 >                                                                                                                          
 You intentionally wipe the Activate symbiosis from your memory.                                                            
 >                                                                                                                          
 [trade: *** No match was found after 15 seconds, dumping info]                                                             
 [trade: messages seen length: 1]                                                                                           
 [trade: message: You intentionally wipe the Activate symbiosis from your memory.]                                          
 [trade: checked against [/But you haven't/i, /You release/i, /Repeat this command/i]]                                      
 [trade: for command release symb]       
```

Tarhish:
Yes, this could be made to avoid that, and it probably ties into how I wanted to look at communizing crafting_training_spells, if it could be done.  I think I need a flag to watch for symbiosis fading, or any successful cast, and have that track whether the symbioses needs to be cleared or not.

trade.lic
```diff
@@ -1,78 +1,425 @@
...
+    @full_pouch_container = @settings.full_pouch_container
+    @training_spells = @settings.crafting_training_spells
+    @step_toggle = -1
+    wipe_token
+    Flags.add('trading-magic-ready', 'You feel fully prepared to cast your spell.')
+
+    @equipment_manager = EquipmentManager.new(@settings)
+    @cooldown_timers = {}
+    @caravan_training_skills = @settings.caravan_training_skills
+    @lead_training_skills = @caravan_training_skills.select { |skill, _cooldown| !['Mechanical Lore', 'Augmentation', 'Utility', 'Warding'].include?(skill) }
```
jandersson: Is this the same as using .reject? without the `!`?

Tarhish: Oh, yeah, I suppose it is, isn't it.  I might as well change that.

trade.lic
```diff
@@ -1,78 +1,425 @@
...
+      [
+        { name: 'closeup', regex: /closeup/i, optional: false, description: 'Close up by delivering all remaining crates and accepting no new contracts' },
+        { name: 'exit', regex: /exit/i, optional: true, description: 'Exits the game after delivering all crates and returning the caravan' }
+      ],
+      [
+        { name: 'outpost', regex: /outpost/i, optional: false, description: 'Bring the caravan to the trade outpost of a town.' },
+        { name: 'town_name', regex: /\w+/i, variable: true, description: 'First word of the town name.' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
```
jandersson: This is nitpicking but arg parsing could be done first; if the users args dont make sense then there's no reason to do any of the above initialization steps..

Tarhish:
I see this as the best time to pick nits so that's no problem.  I've got a bunch of changes to push when I get home related to these and previous comments.  I'll add this to that and try to do another pass to catch any other idiosyncrasies.

jandersson:
For sure. I think seeing the args as one of the first items in a constructor is kind of what you expect too when reading something new

Tarhish:
Addressed a few of the comments -- items still on the list are a symbiosis fix and changing how appraisal training works so it doesn't freak out and do something unusual.

The biggest change this time was forcing UserVars.athletics to 1 during the route calculation process.  I was able to confirm that otherwise it will attempt to do things like run the caravan across the rivers.

Tarhish:
I've been testing the appraisal change for a while and it's definitely better.  I might add gear in, too, but it's just so much better with a few pouches in storage.

Also, mechlore should be a lot less noisy now.  I don't know why I assumed get braided grass wouldn't work without even trying it, but now it won't close and open the feedbag for no reason.

trade.lic
```diff
@@ -205,58 +662,851 @@ class Trade
...
+    @pouch_ordinal = 0 if @pouch_ordinal > 10
+    case bput("get #{$ORDINALS[@pouch_ordinal]} pouch from my #{@full_pouch_container}", '^You get ', '^What were you referring')
+    when /^You get /
+      bput('appraise my pouch', 'Roundtime')
+      waitrt?
+      bput("put my pouch in my #{@full_pouch_container}", 'You put')
+      @pouch_ordinal += 1
+      true
+    else
+      @appraisal_queue -= 'full-pouch' if @pouch_ordinal == 0
```
jandersson: Running into this error: 
```
 --- Lich: error: no implicit conversion of String into Array        
         trade:862:in `-'                                            
         trade:862:in `appraise_held_pouch?'                         
 --- Lich: trade has exited.                                         
```

I think this line needs to be `@appraisal_queue -= ['full-pouch'] if @pouch_ordinal == 0`

Tarhish: Yep, good catch.  I'll check the others.  That's what I get for doing it in a hurry.

jandersson:
I think ive got a contract but lost the goods due to negligence towards my pack animal... in any case im getting stuck in a contract delivery loop with the messaging: 
```
[trade]>give my contract to clerk                                                                                                                                                                                         
                                                                                                                                                                                                                          
The shipping clerk accepts your contract and peruses it.  "Ah, yes," he smiles.  "We did order this, didn't we?"  With a quick snap of his fingers, he calls for an assistant to help you unload your merchandise for his 
  inspection.                                                                                                                                                                                                             
                                                                                                                                                                                                                          
The shipping clerk looks perplexed.  "What have you done with the goods?  Is this a trick?"                                                                                                                               
```

Think the contract should just be trashed?

Tarhish:
Oh, holy cow, never seen that one.

Yeah, it should, and it shouldn't be too hard.  It handles expired contracts, and it even handles if you've lost a contract and the outpost still thinks you have it out, but I haven't ever run into that problem or message.

Tarhish:
Oh, wait, forget too hard.  I guess you could just drop it right there.  There's potentially some value to adding that to the caravan inventory check on startup, but for now I'll just add that message to make you drop the contract.

jandersson:
Sounds good.

I modified deliver_contracts like so and its back to normal:
```ruby
 def deliver_contract
   clerk_id = find_closest_id('shipment_clerk')
   echo "Delivering Contracts"
   walk_to(clerk_id)
   clerk_messages = [
     'What have you done with the goods',
     '^The \w+ clerk finds everything in order with your merchandise'
   ]
   contracts = find_contracts(@contract_container)
   current_town = find_town(Room.current.id).split.first
   until bput("get my #{current_town} contract from my #{@contract_container}", 'You get', 'What were you' ) == 'What were you'
     case bput('give my contract to clerk', clerk_messages)
     when 'What have you done with the goods'
       dispose_trash('contract')
     end
   end
 end
```

Tarhish:
Even better!

profiles/base.yaml
```diff
@@ -504,6 +504,25 @@ workorder_max_items: 3
...
 
+#CARAVAN SETTINGS
+# container holding caravan contracts for ;trade
+trade_contract_container: backpack
+# number of copper to carry on-hand while running ;trade
+caravan_coins_on_hand: 9555
+# Have you purchased a 750plat interior from the trade minister?
+caravan_interior: false
+# skills and cooldowns to train in ;trade.  See https://github.com/rpherbig/dr-scripts/wiki/Trader-Tutorials for valid skills and what they do
+caravan_training_skills:
```
jandersson: Noticed trade does first aid training, should this be included here?

trade.lic
```diff
@@ -205,58 +697,852 @@ class Trade
...
+    @cooldown_timers['First Aid'] = Time.now
+    @cooldown_timers['fa-timer'] = nil
+    waitrt?
+  end
+
+  def performance_routine # runs ;performance for 2 minutes.  Will run ;first-aid too, if it hasn't been done very recently.
+    return unless @training_token == 'Performance'
+    echo "performance routine" if debugging?
+    if @caravan_being_led && check_ability?('First Aid', 180)
+      @training_token = 'First Aid' if @settings.textbook ? exists?(@settings.textbook_type) : exists?('compendium')
```
jandersson: If I dont have either, this results in spam tapping:

```
 [trade]>tap my compendium                                                          
 >                                                                                  
 [trade: in lead training - training token is Performance]                          
 [trade: performance routine]                                                       
 [trade]>tap my compendium                                                          
 >                                                                                  
 [trade: in lead training - training token is Performance]                          
 [trade: performance routine]                                                       
 [trade]>tap my compendium                                                          
```

Tarhish: Yes, thanks for finding that.  I only wanted to put something in base if it was 1) likely for a young trader to have, and 2) not going to cause a script failure if they didn't have it.  But it's so easy to throw a check for compendiums or textbooks in startup validation that I think I'll just do that.

Tarhish: Among a couple other little bugs, there's a check for if you have a compendium/textbook on startup, and if you don't it'll wipe FA from your training list and not attempt to ever try it.

profiles/base.yaml
Tarhish: As below, there's now a check in there, so adding it in shouldn't hurt anyone. Before this goes live I'll do a full review of base skills and add anything in that has an escape clause.

Tarhish:
I think I fixed the excessive symbiosis release issue - A flag trips when you prep the chaos symb and a different one trips on spell success or when it fades from your mind.  It's working, and over a couple hours I haven't seen any unnecessary releases in my log or evidence that it's letting symbioses hang around too long.

All these things will need to be bundled up to common-arcana or other scripts at some point.  I've sort of pledged to do that myself, but first I want to get something that I know works.

Tarhish:
Expanded the arguments for the script and allowed it to go rent a caravan for you.  You can now just run ;trade and expect it to go to the clerk and get started.  Also a ;trade duration <time> <exit> which allows you to run it for X amount of minutes and then begin closing up, potentially exiting the game afterwards. 

Right now I have it dropping any remaining lumber after script completion -- I think it's too heavy to carry around when doing other things.

I think I'm getting close to calling this ready to go.  I want to adjust the base values, make sure the stables in Zoluren are all tagged and mapped, and maybe add a setting to let you adjust the threshold of how many contracts you want to build up before making an emergency delivery.  3 was most effective in my experience, but dropping it to 2 would mean less difficulty closing up.

Tarhish:
I tried running for a while without using my caravan interior and got shot by a champion, so I tied up my wound and added a safe-room check when it arrives at any outpost.

Tarhish:
Alright, I've got this roughly where I want it to be, and I've caught others running it over the last few days without griping.  I think this is capable of being put out there.  I've adjusted the base to where I think it should be -- Any skills that require tools or settings are commented out unless they've got an in-script check to wipe them clean if they don't work.  I'm going to update the trader tutorial and let this sit till tomorrow just to give the last-minute critical bugs time to emerge.

Thanks to everyone who was helping with this!

rpherbig:
Wow, that's a lot. Thank you so much @Tarhish !

@jandersson do you want to do any testing?

Tarhish:
I tried to stay on-task but got distracted trying to track down the weird lich crash that happens 1/500 times you exit a moving caravan.  I was dumping $_SERVERSTRING_s to files to see what the last thing that happened was, getting down and dirty into lich itself.  I learned a lot, but not what I wanted.

Still seems like after about 4-8 hours of running the script with caravan_interior: true it tends to crash for me as a result of that, but whatever. The actual functionality is where I want it to be for now, so I'll see if anyone else has found any more bugs then work on other provinces.

Tarhish:
Wait, I'm being an idiot.  I stow unfinished crafting goods, then look for them later in your storage_containers.  Why don't I just put them in your crafting_container?  Hold on...

Tarhish:
Alright, changed it to work from your crafting_container, which you'd need to have for shaping to work anyway.  I've been running around crafting this way for a few hours.

jandersson:
> Wow, that's a lot. Thank you so much @Tarhish !
> 
> @jandersson do you want to do any testing?

Ive tested this a good bit with my trader, but not the most recent updates. I dont think I can cover all the changes with testing. To say im happy with the changes is an understatement; its made so many improvements over the original trade. I would say its ready when Tarhish feels its ready.

jandersson:
One small thought: Can we echo a documentation link during start up? or in the argparse help messaging? (or both?)

Tarhish:
I feel confident the recent stuff is working better than it used to, and the scope is small enough I can test every one of the recent changes much more thoroughly than the full package I was dealing with at the start.

As for documentation, I'm all for it.  As soon as this is finalized I was planning to document usage at LEAST in the trader tutorial entry on the github wiki, and maybe have some of it make it to epedia if I can manage it, but mentioning it on startup sounds very good.

Also on my list in the near future is also adding a sample-trader yaml.

Tarhish:
We could change dependency's display_arg so it displays a certain given message if it reaches there from the HELP arg, but thinking about it I don't know if we'd need to.  It's probably alright to drop a big echo pointing to the wiki page right before arg processing.

Tarhish:
Just something like:
```

**Comment by jandersson** [comment]
I like it - You're only going to see it once.

**Comment by jandersson** [comment]
Just sayin, I get a lot of help text when starting programs, e.g.
```
jonas@flexo:~$ ipython
Python 2.7.12 (default, Nov 12 2018, 14:36:49)
Type "copyright", "credits" or "license" for more information.

IPython 2.4.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]:
```
It wouldn't be unusual if we did the same.

**Comment by Tarhish** [comment]
I added a message.  I'd like to go further, but I'll have to think about the best way to do it. 

The good news is I've got a version of the script working in Ilithi, and the other provinces should be just a matter of grabbing/mapping the right rooms and handling the ferries.  I don't want to do that yet though -- after changing that I'd probably want to do a lot more testing.

**Comment by Tarhish** [comment]
A change I made a while back to reduce scroll was incrementing forage even if it failed, which would cause mech lore to train more slowly than it could.

I'm just nitpicking now.  As far as I'm concerned this is ready. I'll add documentation onto the wiki and now just I'm testing the script in Forfedhar and Ilithi.

**Comment by jandersson** [comment]
Sweet! I would say the non-zoluren stuff can be safe as an alpha in this PR and leave it up to future PR's to adjust that?

**Comment by Tarhish** [comment]
Sure, I'm still in no hurry, and it... seems to be behaving itself.  I'll need to dump some base-town changes in, but let me make sure I haven't broken anything up in Zoluren first

**Comment by Tarhish** [comment]
I added new data to base-towns.  I've been running around in Ilithi and Forf with this data for a while now, so I doubt it could break anytihng, but I'd appreciate a second eye on that file regardless

**Comment by Tarhish** [comment]
Geez, trade routes are so fiddly.  I ran it in Forf and Ilithi for a long while with no problems, although Forf was built with too many lead routes and Ilithi was built with barely anything.  Nothing that can't be compensated for in the script later.

I went back to Zoluren and have been running this without complaint for a few hours now, so it's looking pretty alright.

**Comment by Tarhish** [comment]
Zoluren trading still works as well as before.  And as long as you have skates, Forfedhar works great as well.  Fewer interruptions, invasions, more time inside the caravans on lead routes -- it's maybe a little slower training trading but it still gets up there.

But since you have to go back to Shard for buying materials, skates are necessary.

**Comment by rpherbig** [comment]
@Tarhish @jandersson no hurry, just checking in. How do we feel about this PR now?

File: data/base-town.yaml [hunk 12]
```diff
@@ -31,10 +31,6 @@ Crossing:
   gemshop:
     id: 4652
     name: appraiser
-  pawnshop:
```
rpherbig: Why was this removed?

profiles/base.yaml
```diff
@@ -504,6 +506,28 @@ workorder_max_items: 3
...
```
rpherbig: ❌ This needs an entry in `base-empty.yaml`, something like `caravan_training_skills: {}`

data/base-town.yaml
Tarhish: :-/

Because I added those outpost rooms on a different computer and didn't pay close enough attention to the diff.  I'll roll those back tomorrow when I'm back home.

jandersson:
> @Tarhish @jandersson no hurry, just checking in. How do we feel about this PR now?

I havent had a chance to test anything after my last comments. I do still think its time to merge and add additional things in future PR's, this is already a humongous update.

data/base-town.yaml
Tarhish: The pawnshops have been replaced, and thank you for catching that.

Tarhish:
As far as I'm concerned this is good to go -- I've got a few tweaks that could improve efficiency especially down in Hibarnhvidar, but I don't want to overextend.  Nobody else is running into anything weird, so I'll wait to see if anyone in the general population can tickle any more problems out of it before continuing to work on it.

Tarhish:
FWIW, I'm not including it in this PR, but since then I've made a couple more efficiency changes and a fix to the caravan interior crash.  Turns out it was easy to keep a list of where each caravan stops on lead, and the way to the next room, which resolves the room id crash.

In addition I've added a 'hunt' arg that will check whenever you stop at capital cities (at most once per half hour) and perform a hunt defined in <playername>-caravanhunt.yaml.  A quick 10 minute jaunt into rock guardians or something bumps the combats I need up to ML, and then it returns to the caravan to continue.

Sarvatt:
This is a huge feature change that only affects one script barely anyone uses, it's seriously good to go and not pulling it is blocking additions to it :)

homebrewer12:
Has the crafting issue or spell casting issue with ;trade been fixed yet?

Tarhish:
I can think of a couple that were fixed?  What were you referring to?

homebrewer12:
Whenever it attempted to knit or shape, it would ask "send 1 for (item)"
"send 2 for (item)

homebrewer12:
And as far as casting, it would cast before the spell was
prepared...resulting in heavy nerve damage by the end of a long day of
trading

Tarhish:
Ah, yes, we talked about these.  Based on what you told me, it wasn't casting before it was fully prepared, but ;trade wasn't on the list of sanowret-no-use-scripts, so it was casting on the road when you had half-concentration, and backfiring if you weren't using low amounts of mana.  Other than that it's just using normal crafting_training_spells logic.

I don't want to put it on that list, so I added a concentration check so it won't cast below 80% concentration.  It seems safer now, but it could be pushed up later.

The crafting change went out a few days ago for both ;trade and ;craft, which had the same problem.  Thanks for identifying both of them!

Tarhish:
@rpherbig Sorry, would it be possible to merge this sometime soon?  I did another pass to make sure it doesn't contain anything that would brake anything else, and since I can't find anymore problems with it I'd like more people trying it out live.

After that the sample-trader PR should pass tests too, but we'll see.

homebrewer12:
Found another issue.  Started ;trade, got.first contract from crossing to
tiger clan.  Got to tiger clan, turned in contract, and it killed ;trade.

homebrewer12:
Where can I find the newest version with those things fixed?

Tarhish:
If you go up at the top of this page to Files Changed you'll see everything that was touched.  If it was mostly working beforehand then trade.lic is probably the only thing you'll need

homebrewer12:
Ok, I made the change for engineering, and its still having the same error:

[ntrade: Please select desired recipe ;send #]
    1: a wood cloak pin
    2: a wood cloak pin

Tarhish:
Oh flark, there are two identical wood cloak pin recipes in base-recipes

Tarhish:
The base-recipes data change will go in another PR.  I checked every other recipe used in trade and craft and there's no other sources of ambiguity to fix after that.  Again, good catch

Tarhish:
If you wanted to try it out now, go to scripts/data/base-recipes.yaml, find the two entries for 'cloak pin' and delete one entirely

homebrewer12:
I just changed the noun for cloak pin from 'pin' to 'cloak pin'.

There are two recipes, a wooden pin, and a wooden cloak pin, but both shared the same noun, 'pin'.

Tarhish:
There were actually two completely identical letter-for-letter entries for cloak pins which is why that was a problem.  It couldn't pick between them.  rpherbig merged that change a while ago so if you went looking you might have only seen the one.

I checked the others with a script and there wasn't any other ambiguous entry that would trigger that same problem.

PR #3436, opened 2019-02-21
herb-stock coin withdraw adjustment

andymanak:
To keep herb-stock from getting too many copper coins, now gold will be used when possible.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3670,7 +3670,6 @@ end
 before_dying do
   DRCA.release_cyclics
   DRCA.shatter_regalia?
-  fput('close fissure') if DRStats.warrior_mage? && DRRoom.room_objs.any? { |x| /fissure/ =~x }
   Flags.delete('using-corpse')
   Flags.delete('pouch-full')
   Flags.delete('container-full')
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3670,6 +3670,7 @@ end
 before_dying do
   DRCA.release_cyclics
   DRCA.shatter_regalia?
+  fput('close fissure') if DRStats.warrior_mage? && DRRoom.room_objs.any? { |x| /fissure/ =~x }
   Flags.delete('using-corpse')
   Flags.delete('pouch-full')
   Flags.delete('container-full')
--- a/herb-stock.lic
+++ b/herb-stock.lic
@@ -33,7 +33,8 @@ class Herbstock
       item['buy_num'] = buy_num
       items_to_restock.push(item)
     end
-    get_money_from_specific_bank?("#{coin_needed} copper #{town_currency(town)}", town) if coin_needed > 0
+    get_money_from_specific_bank?("#{coin_needed/1000} gold #{town_currency(town)}", town) if coin_needed > 0
+    get_money_from_specific_bank?("#{coin_needed%1000} copper #{town_currency(town)}", town) if coin_needed > 0
     items_to_restock.each do |item|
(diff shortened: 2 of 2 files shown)


Tarhish:
Can I make a suggestion?  I had to deal with this same issue myself recently, and it turned out minimize_coins was the way to go.

As an example - 
```
minimize_coins(withdraw_amt).each { |amount| get_money_from_bank(each_amount, settings)  }
```

Minimize coins takes a number (say 25555) and returns an array of the smallest number of coins needed to have that much money.  Then the .each withdraws each unit individually (2 plat, 5 gold, 5 silver, 5 bronze, 5 copper)

withdraw_from_specific_bank has a habit of going to the bank, then returning on each withdrawal, so it may not work well with this, but there are ways around that if you need to.  I did this in get_fare? in the recent bescort change.  You can change your @settings.hometown, walk to the bank you want, do get_money_from_bank, walk back, then change the hometown back to normal.

Sarvatt:
there is nothing wrong with this outside of not doing it the preferable way which can be fixed later. it would be preferable to do it right from the start but that shouldnt stop a new contributor from fixing things.

Tarhish:
Oh, yes, sorry.  I didn't realize -- a lot of scripts still use money this way so it's no big deal.  We could change it later

andymanak:
Thanks Tarhish, I'm glad there is a cleaner way to implement this.
Just for confirmation, in bescort is this where you are travelling to the bank before, this way it does not yo-yo in and out of the bank getting each denomination?
Line 1327:    manual_go2(get_data('town')[town]['deposit']['id'])

Tarhish:
Yes, but, in general whenever you want to have a script walk somewhere just use walk_to(roomid).  walk_to is defined in common-travel, I believe, and it's a go2 that handles several potential problems.  The reason bescort uses an internal manual_go2 is because walk_to gets confused if there's already a copy of go2 paused in the background (which is the case during bescort)

So if you want to move anywhere, just use walk_to(12345) or whatever.  You could also do something like:
```
current_room = Room.current.id
walk_to(wherever you want to go)
do_thing
walk_to(current_room)
```
Which will walk to a location, do the thing you want to do, then walk back to where you started.

Tarhish:
So you can walk to a specific town's bank with walk_to(get_data('town')['Crossing']['deposit']['id']) or even
```
@town_data = get_data('town')
walk_to(@town_data['Crossing']['deposit']['id'])

**Comment by Tarhish** [comment]
If I had to take a stab at it I think it would look like (get money from your hometown)
```
hometown = settings.hometown
town_data = get_data('town')
current_room = Room.current.id

if wealth(town_currency(hometown)) < coin_needed
  walk_to(town_data[hometown]['deposit']['id'])
  DRCM.minimize_coins(coin_needed).each { |each_amount| DRCM.withdraw_exact_amount?(coin_amount, settings) }
  walk_to(current_room)
end
```

Or if you want to send it to a particular town instead of your hometown:
```
settings.hometown = 'Shard' #or whatever
town_data = get_data('town')
current_room = Room.current.id

if wealth(town_currency(settings.hometown)) < coin_needed
  walk_to(town_data[settings.hometown]['deposit']['id'])
  DRCM.minimize_coins(coin_needed).each { |each_amount| DRCM.withdraw_exact_amount?(coin_amount, settings) }
  walk_to(current_room)
end
```

Tarhish:
Really I need to put something into common-money eventually that handles this without screwing around.

rpherbig:
Thank you for your contribution @andymanak !

PR #3442, opened 2019-02-24
combat-trainer.lic: Remove bob from whirlwind actions

Sarvatt:
With the new rested exp, some people want to minimize tactics training to conserve rested exp.


xtremenerdrage:
This would be great to remove the pulse group that has tactics, you still jab/feint to recover balance.

Issue #3022, opened 2018-08-03
combat-trainer getting stuck loading and aiming bows


Sarvatt:
deleted

Sarvatt:
it takes some time to trigger but once it's broken it'll forever be broken in that hunt unless it seems like.

Sarvatt:
scratch that, hit the bug again with that change.

xtremenerdrage:
I've noticed this alot, it does seem to only happen with a hunting buddy

xtremenerdrage:
Wouldn't it have to do with the line that makes the game_state false if a mob died?

  def attack_aimed(charged_maneuver, game_state)
    game_state.selected_maneuver = charged_maneuver unless game_state.loaded
    **_game_state.loaded = false if game_state.mob_died && !(game_state.casting_consume || game_state.prepare_consume)_**
    game_state.clear_aim_queue if Flags['ct-ranged-ready']

Sarvatt:
This seems to happen right after an armor swap happens from looking at the logs.

xtremenerdrage:
I actually haven't noticed the stuck loading issue in a long time

Sarvatt:
didn't you start using bandaid.lic as a during:?

xtremenerdrage:
Oh I haven't been hunting with people lately, that's probably why I haven't noticed it

Sarvatt:
it seems that if something happens on a timer in say trainerprocess it wont actually get passed along to the game if it's currently doing work in initialize (aka armor swapping or warhorns), I don't know why though.

PR #3443, opened 2019-02-24
combat-trainer.lic: work around a bug with cycling armors causing inf…

Sarvatt:
…inite load/aim loops.

Closes: #3022

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2763,14 +2763,14 @@ class AttackProcess
   end
 
   def aim(game_state)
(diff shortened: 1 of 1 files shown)


Sarvatt:
This still causes a 15 second stall for the first load after armor is cycled, but its better than being broken completely until we can fix the root cause.

Sarvatt:
it seems that if something happens in another process it wont actually get passed along to the game if it's currently doing work in initialize (aka armor swapping or warhorns), I don't know why though.

Sarvatt:
This is easy reproduceable by setting up cycle_armors: lowering the cycle_armors_timer: to something low and having a ranged only hunt btw if anyone else wants to look at it.

Issue #2519, opened 2018-02-14
Trader self rejuvinate - affects rezz script


yuoree:
Bitrad's memories are protected from loss by the Resumption spell.
His soul is bound to his body.
His body will decay in about 1 hour.
Bitrad has gained a substantial number of favors with his god.
Roundtime: 5 sec.

Overwrites rejuv when perceived - cant see rejuv color/etc, so script ends up stuck in a rejuv/fail to detect loop. Just needs to be able to treat the above as silver.

PR #3444, opened 2019-02-25
Add trader Resumption messaging to rezz

Tarhish:
Closes https://github.com/rpherbig/dr-scripts/issues/2519

This shows up on PERC for a corpse, and will avoid getting stuck in a rejuv loop.


Tarhish:
What the - is this failing because I'm a heathen?

PR #3447, opened 2019-02-25
Skip moonblade check if not summoning weapons

jandersson:
Give this a few days for testing; would like to avoid breaking anything

This is to avoid the 'glance moon' spam users dont like too much.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -192,7 +192,7 @@ class SetupProcess
 
     game_state.reset_action_count
 
-    if DRStats.moon_mage? && moon_used_to_summon_weapon.nil?
+    if DRStats.moon_mage? && game_state.summoned_weapons.empty? && moon_used_to_summon_weapon.nil?
       echo('skipping summoned weapons because no moonblade available') if $debug_mode_ct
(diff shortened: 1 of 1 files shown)


Sarvatt:
looks good to me, thanks for fixing this!

rpherbig:
:+1:

Sarvatt:
i dont see any reason to avoid merging this asap. its very straight forward and correct.

jandersson:
Ok, I finally sat down and tested this. I forgot to make the summon_weapons variable non-private, so its good that this didnt get merged.

With that fixed, ct will check glance moon only once at startup:
```
[combat-trainer]>glance moon
>
[combat-trainer: skipping summoned weapons because no moonblade available]
[combat-trainer: checking weapons as nil!=Large Edged]
[combat-trainer]>get my bastard.sword
```

jandersson:
It looks like it is still doing the glance, so will need to double check that logic.

jandersson:
@rpherbig This is ready to go. Tested with and without moonblade usage. Players wont 'glance moon' anymore if no summoned weapons defined

PR #3451, opened 2019-02-27
Use leather shield handle for leather shields

jandersson:
The other shield handle item refers to metal shield
handle and doesn't work with leather shields.

I noticed this when restock-shop broke when trying to restock gryphon shields

Staging this for review, there are other leather shields which need the switch. Ill give a mention when its ready.

### DIFF
--- a/data/base-crafting.yaml
+++ b/data/base-crafting.yaml
@@ -1255,6 +1255,10 @@ recipe_parts:
     Boar Clan: 
       part-room: 8892 # Hibarnhvidar
       part-number: 
+  leather shield handle:
+    Crossing:
+      part-room: 16667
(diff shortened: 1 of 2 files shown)


rpherbig:
👍

Sarvatt:
missing all the other towns but that can be fixed later :)

jandersson:
It doesn't seem to be a crisis, so I will take the time to do all the necessary updates. If anyone notices the issue, its ready to merge though.

Sarvatt:
working fine here

jandersson:
This is ready to merge. My work on this hasn't matched my ambition and it seems like nobody is crafting leather shields?

PR #3452, opened 2019-02-27
Fix sniping in combat-trainer for non-lodging ammo

Tarhish:
After the ammo-lands-at-foot patch, sniping doesn't enter the right WHEN in shoot_aimed, so it doesn't pick up fired ammo.  Instead, it enters the 'your moving to fire went unseen' or 'notices your attempt to remain hidden' WHENs, which don't do anything about the ammo.

However, the new WHEN we added does everything we wanted it to do when sniping in the first place, so testing suggests we should just remove it.

If someone stealthier has an opinion on special behavior we should have during sniping then I'm all ears, but at the moment this fixes it so that it will hide, snipe, then stow ammo.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2649,11 +2649,9 @@ class AttackProcess
   end
 
   def shoot_aimed(command, game_state)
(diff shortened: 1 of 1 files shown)


Tarhish:
This should only be a problem when using non-lodging ammunition.

PR #3457, opened 2019-03-02
sanowret-crystal.lic - added option for adjective

BinuDR:
I added a new setting sanowret_noun that can be used to specify the noun of your crystal if it's something other than sanowret.

There is no need to use this setting if your crystal responds to 'sanowret crystal', but someone mentioned on lnet last night that it didn't work with their hoarfrost white sanowret crystal.

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -14,6 +14,7 @@ class SanowretCrystal
     @invalid_rooms = []
     @worn_crystal = false
     args = parse_args(arg_definitions)
+    @noun = args.sanowret_noun ? args.sanowret_noun : 'sanowret'
     @no_use_scripts = get_settings.sanowret_no_use_scripts
     check_crystal if args.run && !hiding? && !invisible?
     passive unless args.run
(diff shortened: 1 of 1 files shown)


BinuDR:
It's an optional setting.  If it's not found in your yaml, it uses 'sanowret' as the noun.

sanowret-crystal.lic
```diff
@@ -14,6 +14,7 @@ class SanowretCrystal
     @invalid_rooms = []
     @worn_crystal = false
     args = parse_args(arg_definitions)
+    @noun = args.sanowret_noun ? args.sanowret_noun : 'sanowret'
```
rpherbig: ❌ This arg needs to be added to `arg_definitions` on line 9

sanowret-crystal.lic
```diff
@@ -45,17 +46,17 @@ class SanowretCrystal
...
     else
-      case DRC.bput('tap my sanowret crystal', /^You tap (?:a|an) .*sanowret crystal inside your .*.$/, /^You tap (?:a|an) .*sanowret crystal.* that you are wearing.$/, /^You tap (?:a|an) .*sanowret crystal that you are holding.$/, 'I could not find what you were referring to.', 'Not here, ', 'a wave of Corruption magic')
+      case DRC.bput("tap my #{@noun} crystal", /^You tap (?:a|an) .*sanowret crystal inside your .*.$/, /^You tap (?:a|an) .*sanowret crystal.* that you are wearing.$/, /^You tap (?:a|an) .*sanowret crystal that you are holding.$/, 'I could not find what you were referring to.', 'Not here, ', 'a wave of Corruption magic')
       when /You tap (?:a|an) .*sanowret crystal inside your/
-        DRC.bput('get my sanowret crystal', 'You get', 'What were you referring to')
+        DRC.bput("get my #{@noun} crystal", 'You get', 'What were you referring to')
         use_crystal
-        DRC.bput('stow my sanowret crystal', 'You put', 'Stow what')
+        DRC.bput("stow my #{@noun} crystal", 'You put', 'Stow what')
       when /^You tap (?:a|an) .*sanowret crystal.* that you are wearing\.$/
```
rpherbig: ❓ Does this line (and lines 50, 57, etc.) need to be updated to use the noun instead of `sanowret`?

sanowret-crystal.lic
rpherbig: Depending on the use case, this could be a setting in your YAML instead. How frequently/likely would this noun be to change over time?

sanowret-crystal.lic
BinuDR: Yes.  I'm not sure how many of the other sanowret crystals have this issue, but the hoarfrost white sanowret crystal uses white as the adjective.

I really must have been tired, I should have named the variable adjective not noun.

sanowret-crystal.lic
BinuDR: Oh, my mistake.  I was tired when I was doing this.

It was supposed to be from get_settings, not args.

BinuDR: Yes.  That was my intention.  I'll fix this up and submit the changes.

sanowret-crystal.lic
BinuDR: The match lines don't need to be updated because they have a .* in front of sanowret, so it'll match any adjective, or none.

sanowret-crystal.lic
```diff
@@ -14,6 +14,7 @@ class SanowretCrystal
     @invalid_rooms = []
     @worn_crystal = false
     args = parse_args(arg_definitions)
+    @adjective = get_settings.sanowret_adjective ? get_settings.sanowret_adjective : 'sanowret'
```
rpherbig: 💡 It's minor, but this script now loads settings from your YAML three times (twice on this line, once on line 18). I recommend saving it off to a local var.

rpherbig: ❌ Since this is a new setting, it needs to be in `base.yaml` with a default value. If you make the default 'sanowret' you could avoid the conditional on this line.

PR #3462, opened 2019-03-04
Add bone recipes and change container

Tarhish:
A certain someone is maxing out recipes in shaping.  Adding a few recipes he tested.

Also, a couple containers had their nouns adjusted to make them less likely to collide with worn containers

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -949,6 +949,13 @@ crafting_recipes:
   work_order: true
   chapter: 4
   material: stone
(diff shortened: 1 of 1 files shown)


Sarvatt:
Looks good to me. These were missing and needed to be added in and the cloth container changes are worth adding. I had an issue with making them workorder: true by default but talked to him about it and it should be fine.

rpherbig:
Thanks for looking it over @Sarvatt

PR #3432, opened 2019-02-16
Update definition of @gear_sets to flatten the gear list array

gerritlansing:
I would like to introduce this change to equipmanager to allow the use of anchors and references within gear_sets. For example, the following configuration allows cascading changes to each script with its own gear set. The flatten is harmless to existing configuration.
```
gear_sets:
  armor: &armor
  - quilted pants
  - quilted shirt
  - quilted gloves
  - quilted hood
  bow: &bow
  - battle longbow
  shield: &shield
  - target shield
  parry_stick: &parry_stick
  - parry stick
  brawling: &brawling
  - brass knuckles
  - elbow spikes
  standard: [ *armor, *bow, *shield, *brawling, *parry_stick ]
  stealing: [ *bow,  *parry_stick, *brawling ]
  nothing: []
```

Secondly, I see this as enabling the development of a manual equip/unequip gear set feature, which is convenient for interactive game play without needing to write separate scripts. (Forgive me if this exists, I haven't been able to find it but am new to Lich/dr-scripts.)

PR #3482, opened 2019-03-12
Clean up syntax of changes in PR #3432

gerritlansing:
Updated to the requested syntax. Thanks for the feedback. Testing yielded the same results.


rcuhljr:
Thank you @gerritlansing! If you're interested in what motivated the changes let me know and I can dive into some more detail that might not be interesting if you just want to get something working.

PR #3486, opened 2019-03-14
workorders.lic - removed knitting tools

BinuDR:
knitting_tools: was removed from base.  Removed it from workorders

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -92,7 +92,7 @@ class WorkOrders
         exit
       end
       @outfitting_room = @settings.outfitting_room
-      tools = (@settings.outfitting_tools + @settings.knitting_tools).uniq
+      tools = @settings.outfitting_tools
       @belt = @settings.outfitting_belt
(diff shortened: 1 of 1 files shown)


rpherbig:
We'll let this PR sit for a few days to make sure everyone using the old setting has had time to migrate over. Can you ping me if I haven't merged it by then?

BinuDR:
I made this because people are getting an error if they hadn't ever defined knitting_tools in their yaml.  It gives an error about trying to add to a nil class.

rpherbig:
Ah, gotcha! I'll merge it now then

Sarvatt:
ahh I didn't realize this would be a problem, now that base has changed it really should be pulled.

rpherbig:
Thanks @BinuDR

Sarvatt:
sorry about that and thanks for catching it

BinuDR:
Thanks

PR #3490, opened 2019-03-18
;craft 'wood' vs 'articulated wood' bracelet unique name

Tarhish:
Just had someone run into this in the new ;trade, where I stole ruthlessly from ;craft.

If you send recipe_lookup 'wood bracelet' it asks you to choose between 'a wood bracelet' and 'an articulated wood bracelet'.  If you send ;shape 'wood bracelet' it just chooses the articulated one instead, since it's higher up in the row.

I forced myself to this skill level and it crafted the right bracelet.  Already made the change to ;trade.

### DIFF
--- a/craft.lic
+++ b/craft.lic
@@ -76,7 +76,7 @@ class Craft
     if rank <= 25 # Tier 1  Extremely Easy
       shape(7, 'wood band', 'band')
     elsif rank <= 50 # Tier 2 Very Easy
-      shape(7, 'wood bracelet', 'bracelet')
+      shape(7, 'a wood bracelet', 'bracelet')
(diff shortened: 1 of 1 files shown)


Sarvatt:
thanks for this, probably wasn't noticed because you have to have a master's shaping book to hit it. then again now that I look I don't even see articulated wood bracelet on https://elanthipedia.play.net/Shaping_products ? is it new?

Tarhish:
It might be new, but yeah I forgot to mention that.  It's not on the wiki, but is in the book and base-recipes.  I think it's just startlingly rare that someone's running ;craft between 25-50 ranks of Engineering so we've never head anyone complain about it.

Issue #3494, opened 2019-03-22
performance needs wearing zills and stowing chamois cloth in its cleanup routine


proctophage:
https://pastebin.com/AMHGrFcn

even with performance_pause: 10 study-art fails to give enough time to performance for it to re-wear zills and stow the cloth -- resulting in full hands when the script is forcefully terminated by other scripts like study-art, athletics, and possibly first-aid. There needs to be a hand check in the cleanup routine for performance (where it stops playing before exiting).

The alternative is to stop weaving multiple scripts together like this because they cause a huge number of problems. These scripts that include performance have been the majority of the issues I've posted here. The best option would be to include performance.lic in every script that calls it instead of the current method of kludging them to run in tandem.

PR #3495, opened 2019-03-26
Item check on performance-end

Tarhish:
Closes https://github.com/rpherbig/dr-scripts/issues/3494

Stows zills or cloth if performance is killed during cleaning.

```
>;kill performance
[performance]>stop play
In the name of love?
>
[performance]>wear my zills
You slide a pair of copper zills onto your finger.
>
[performance]>stow my chamois cloth
You put your cloth in your cartographer's trunk.
```

### DIFF
--- a/performance.lic
+++ b/performance.lic
@@ -150,7 +150,10 @@ end
 
 before_dying do
   fput('stop play')
-
+  snapshot = "#{DRC.right_hand}#{DRC.left_hand}"
+  cloth = get_settings.cleaning_cloth
+  DRC.bput('wear my zills', 'You slide', 'You are already', 'Wear what') if snapshot.include?('zills')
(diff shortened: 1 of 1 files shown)


Sarvatt:
study-art could call performance with the noclean arg like all the other scripts do

Tarhish:
I was going to say this works out fine but Travis put a price out on my head, so that's that.  I guess I'll close this and go with that idea?

Sarvatt:
por que no los dos? :)

Sarvatt:
before_dying is really special, its hard to make it work. lich thinks the script is done when it’s in the before_dying routine. thats what needs ti be fixed

Tarhish:
From study-art - `start_script('performance', ['noclean']) unless Script.running?('performance') || Script.running?('crossing-training')`

Can you think of why this would have crossing-training in it?

Sarvatt:
crossing-training has its own internal performance logic in it, or at least it did

PR #3498, opened 2019-03-27
Update combat-trainer to replace direct references to @kneel_khri

gerritlansing:
I missed that combat-trainer references @kneel_khri (boolean) directly, so this PR addresses that by having combat-trainer call `kneel_for_khri?` as appropriate. This PR also includes a change to the Khri Prowess training ability so that it won't retreat and kneel (without activating prowess) if there are no mobs in the room.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1814,10 +1814,10 @@ class AbilityProcess
       fput action
       waitrt?
     end
-    return if game_state.danger && @kneel_khri
     @khri
       .map { |name| "Khri #{name}" }
       .each do |name|
+        next if game_state.danger && kneel_for_khri?(@kneel_khri, name)
         timer = game_state.cooldown_timers[name]
         next unless !timer || (Time.now - timer).to_i > 30
         game_state.cooldown_timers[name] = Time.now if activate_khri?(@kneel_khri, name)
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2030,14 +2030,16 @@ class TrainerProcess
...
-        fput('stand')
+      unless game_state.npcs.empty?
+        if kneel_for_khri?(@kneel_khri, 'khri prowess')
+          fput('retreat')
+          fput('kneel')
+        end
+        bput('khri prowess', 'Remembering the mantra of mind over matter', 'You\'re already using the Prowess meditation.', 'previous use of the Prowess', 'Your body is willing', 'Your mind and body are willing')
+        if kneel_for_khri?(@kneel_khri, 'khri prowess')
+          waitrt?
+          fput('stand')
```
rpherbig: I think there's a `fix_standing` helper that would be more appropriate

combat-trainer.lic
```diff
@@ -2030,14 +2030,16 @@ class TrainerProcess
...
-        fput('retreat')
-        fput('kneel')
-      end
-      bput('khri prowess', 'Remembering the mantra of mind over matter', 'You\'re already using the Prowess meditation.', 'previous use of the Prowess', 'Your body is willing', 'Your mind and body are willing') unless game_state.npcs.empty?
-      if @kneel_khri
-        waitrt?
-        fput('stand')
+      unless game_state.npcs.empty?
+        if kneel_for_khri?(@kneel_khri, 'khri prowess')
+          fput('retreat')
```
rpherbig: I think there's a `retreat?` helper that would be more appropriate

PR #3503, opened 2019-04-02
Introduce mefwatch.lic

gerritlansing:
mefwatch.lic is primarily useful to thieves, and uses two Mental Focus wands to keep Mental Focus up all the time.

Defaults assume that 'bloodwood branch' is the name of both wands, and the user is fine with mefwatch interrupting any other script. However, these can be changed in the user's setup.yaml. At present, mefwatch requires both MEF wands to be the same adjective & noun.

```
mefwatch:
  wand: bloodwood branch
  no_use_scripts: []
```

Debug may be enabled with `;vars add mefwatch_debug=true`.

### DIFF
--- a/mefwatch.lic
+++ b/mefwatch.lic
@@ -1,0 +1,104 @@
+=begin
+Alternates between mental focus wands to keep MEF up at all times
+Requires two of the same wand
+=end
+custom_require.call(%w[common drinfomon equipmanager])
+
+class MEFWatch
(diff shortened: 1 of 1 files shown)


Sarvatt:
What would you think about naming it something more generic, like cycle-magic-devices or something. We can't change the name of the script after its accepted and I could see this being refactored to work for more than just branches easily.

PR #3504, opened 2019-04-02
Introduce stabbity.lic

gerritlansing:
stabbity.lic is designed for thieves, and relies on backstab to kill things quickly. Stabbity is invoked as follows:

`stabbity [mode] [noloot]`

Where `mode` is:
  `arena`: Used when running stabbity in the Dusk Ruin arena.
  `single`: Will cause stabbity to kill all mobs in a room, and then terminate. Will terminate if no mobs are found at start.
  `equip`: Will wield the preferred weapon
  `cleanup`:  Will sheath the held weapon
 
And, `noloot` prevents the looting of corpses.

Configuration, in the user's setup.yaml, is as follows:
```
stabbity:
  weapons:
    preferred: glaes pasabas
    alternate: judge's gavel
    thrown: throwing club
  use_alternate_on:       # What mobs to use the alternate weapon on
    - archer
    - soldier
    - dryad
    - oshu
    - frostweaver
    - guardian
  use_thrown_on:          # What mobs to use a thrown weapon on
    - hawk
    - gryphon
  thrown_invoke: true   # Whether to use INVOKE to retrieve a bonded thrown weapon
  thrown_action: hurl    # What verb to use when throwing a weapon
```

### DIFF
--- a/stabbity.lic
+++ b/stabbity.lic
@@ -1,0 +1,253 @@
+=begin
+Stabbity, Stabbity, Stabbity. Repeat.
+=end
+custom_require.call(%w[common drinfomon equipmanager events])
+
+class Stabbity
+  include DRC
+
+  def initialize
+    unless DRStats.thief?
+      echo '*** But you\'re not a thief! ***'
+      exit
+    end
+    
+    arg_definitions = [
+      [
+        { name: 'mode', options: %w[arena single equip cleanup], optional: true, description: 'What mode to use' },
+        { name: 'noloot', options: %w[noloot], optional: true, description: 'Do not loot corpses' }
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+    @settings = get_settings
+    @debug = UserVars.stabbity_debug
+    @equipmanager = EquipmentManager.new
+    @preferred_weapon = @settings['stabbity']['weapons']['preferred']
+    @alternate_weapon = @settings['stabbity']['weapons']['alternate']
+    @thrown_weapon = @settings['stabbity']['weapons']['thrown']
+    @alternate_weapon_npcs = @settings['stabbity']['use_alternate_on']
+    @thrown_weapon_npcs = @settings['stabbity']['use_thrown_on']
+    @noloot = (args.mode == 'arena' || args.noloot) ? true : false
+    @mode = args.mode
+    @ignore_npcs = %w[Usdiwi Servant leopard construct zombie representative]
+    @current_weapon = ''
+    @arena_trap = nil
+    
+    cleanup if cleanup_mode?
+    use_weapon
+    exit if equip_mode?
+    hide?
+    combat_loop
+  end
+
+  def arena_mode?
+    @mode == 'arena'
+  end
+
+  def single_mode?
+    @mode == 'single'
+  end
+
+  def equip_mode?
+    @mode == 'equip'
+  end
+
+  def cleanup_mode?
+    @mode == 'cleanup'
+  end
+
+  def cleanup
+    echo "*** Cleaning up ***" if @debug
+    @equipmanager.stow_weapon
+    exit
+  end
+
+  def combat_loop
+    loop do
+      npcs = get_npcs
+      if npcs.count > 0
+        @target = npcs.first
+        @npc_is_alive = true
+        echo "*** Target is #{@target} ***" if @debug
+        watch_target if arena_mode?
+        if @alternate_weapon_npcs.include?(@target) || @thrown_weapon_npcs.include?(@target)
+          dodge_arena_trap if arena_mode?
+        end
+        case bput("advance #{@target}", 'You are already at melee', 'is already quite dead', 'You begin to', 'You spin around')
+        when 'is already quite dead'
+          pause 2
+        end
+        using_right_weapon?
+        if @thrown_weapon_npcs.include? @target
+          echo "*** Target #{@target} is in thrown_weapon_npcs ***"
+          kill_thrown
+          waitrt?
+          loot_mob unless @noloot
+          use_weapon 'preferred'
+          hide?
+        else
+          kill_stabbity
+          waitrt?
+          loot_mob unless @noloot
+          use_weapon 'preferred'
+          hide?
+        end
+      else
+        echo "*** No targets ***" if @debug
+        exit if single_mode?
+        pause 1
+      end
+    end
+  end
+
+  def kill_stabbity
+    while @npc_is_alive
+      fix_standing
+      case bput('backstab', 'You must be hidden to blindside', 
+                            'It would help if you were closer',
+                            'You\'ll need to stand up first',
+                            'flying out of reach',
+                            'to stop flying before',
+                            'Roundtime',
+                            /\[You\'re /,
+                            'There is nothing else')
+      when 'There is nothing else'
+        @npc_is_alive = false
+        return
+      when 'You must be hidden to blindside'
+        hide?
+        waitrt?
+      when 'It would help if you were closer'
+        fput('advance')
+        pause 3
+      when 'You\'ll need to stand up first'
+        fix_standing
+        hide?
+      when 'flying out of reach', 'to stop flying before'
+        echo "*** Switch to thrown -- add this MOB to your use_thrown_on list ***" if @debug
+        use_weapon 'thrown'
+        kill_thrown
+        return
+      when 'Roundtime', /\[You\'re /
+        return if npc_dead?
+        waitrt?
+        dodge_arena_trap if arena_mode?
+        hide?
+      end
+    end
+  end
+
+  def kill_thrown
+    action = @settings['stabbity']['thrown_action'].nil? ? 'throw' : @settings['stabbity']['thrown_action']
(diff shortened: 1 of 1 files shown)


rpherbig:
Thank you for providing an example of the YAML for this!

We already have some mechanisms in place that might reduce the configuration needed. For example, this script could wear a 'stabbity' `gear_set` at the start.

By using weapons declared in your gear list, you could use equipment manager to look up their properties ([example](https://github.com/rpherbig/dr-scripts/blob/3d79a7668d17640a64fba29c644e589cf29faabe/combat-trainer.lic#L3621)) and no longer need to define a `thrown_action` or `thrown_invoke`.

Thoughts?

stabbity.lic
```diff
@@ -0,0 +1,253 @@
+=begin
+Stabbity, Stabbity, Stabbity. Repeat.
```
rpherbig: Please link to elanthipedia and post relevant documentation there.

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+    @debug = UserVars.stabbity_debug
+    @equipmanager = EquipmentManager.new
+    @preferred_weapon = @settings['stabbity']['weapons']['preferred']
+    @alternate_weapon = @settings['stabbity']['weapons']['alternate']
+    @thrown_weapon = @settings['stabbity']['weapons']['thrown']
+    @alternate_weapon_npcs = @settings['stabbity']['use_alternate_on']
+    @thrown_weapon_npcs = @settings['stabbity']['use_thrown_on']
+    @noloot = (args.mode == 'arena' || args.noloot) ? true : false
+    @mode = args.mode
+    @ignore_npcs = %w[Usdiwi Servant leopard construct zombie representative]
```
rpherbig: We have an `ignored_npcs` setting that could be used here.

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+    if DRRoom.dead_npcs.include?(@target) || get_npcs.empty?
+      echo "*** Target #{@target} has died ***" if @debug
+      @npc_is_alive = false
+      return true
+    else
+      return false
+    end
+  end
+
+  def using_right_weapon?
```
rpherbig: This function name should not end in a `?` if you aren't going to return a boolean and use its value

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+        return if npc_dead?
+        waitrt?
+        dodge_arena_trap if arena_mode?
+        hide?
+      end
+    end
+  end
+
+  def kill_thrown
+    action = @settings['stabbity']['thrown_action'].nil? ? 'throw' : @settings['stabbity']['thrown_action']
```
rpherbig: Should the default throw action be to 'lob' to avoid lodging?

rpherbig: I _think_ you could simplify this: `action = @settings['stabbity']['thrown_action'] || 'throw'`

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+        hide?
+        waitrt?
+      when 'It would help if you were closer'
+        fput('advance')
+        pause 3
+      when 'You\'ll need to stand up first'
+        fix_standing
+        hide?
+      when 'flying out of reach', 'to stop flying before'
+        echo "*** Switch to thrown -- add this MOB to your use_thrown_on list ***" if @debug
```
rpherbig: Instead of 'this MOB' could you output `@target`?

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+      echo "*** Switching to thrown weapon ***" if @debug
+      use_weapon 'thrown'
+    elsif @current_weapon != 'preferred'
+      echo "*** Switching to preferred weapon" if @debug
+      use_weapon
+      hide?
+    end
+  end
+
+  def use_weapon(type = 'preferred')
```
rpherbig: I recommend not providing a default value

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+    echo "*** Already on right type ***" if @debug && @current_weapon == type
```
rpherbig: Would 'weapon' or 'type of weapon' be clearer than 'type'?

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+      @current_weapon = 'thrown'
+      return if right_hand =~ /#{@thrown_weapon}/
+      @equipmanager.stow_weapon
+      @equipmanager.wield_weapon(@thrown_weapon)
+    end
+    echo "*** Changed weapon type to #{type} ***" if @debug
+  end
+
+  def watch_target
+    case bput("watch #{@target}", 'you could try to pedal',
```
rpherbig: You could use a regex to simplify this

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+      @arena_trap = 'lean'
+    when 'you could try to cower'
+      @arena_trap = 'cower'
+    end
+  end
+
+  def dodge_arena_trap
+    if @arena_trap
+      echo "*** Performing arena action #{@arena_trap} ***" if @debug
+      bput(@arena_trap, '.*')
```
rpherbig: I _think_ you can omit the `'.*'` here (since it doesn't really provide much value). I recommend trying to find the appropriate text matches, though.

stabbity.lic
```diff
@@ -0,0 +1,253 @@
...
+      @arena_trap = nil
+    end
+  end
+
+  def loot_mob
+    bput('loot', '.*')
+    pause 1
+    #todo/maybe -- pick up loot
```
rpherbig: You might want to look at combat-trainer and see if you can steal any looting logic there.

As a user of this script, I would except it to loot the same way.

rpherbig:
Merging this due to the fact it won't break anything. I'd still like to have the discussion on my comments @gerritlansing .

gerritlansing:
@rpherbig Absolutely. Been tied up with things but will review them this weekend.

stabbity.lic
gerritlansing: Will use `ignored_npcs` setting in next release.

stabbity.lic
gerritlansing: Renamed to `switch_weapon` in next release.

stabbity.lic
gerritlansing: Yes. Next release will use EquipmentManager to read the lodging/bound parameters and copy `combat-trainer`'s `thrown_attack_verb` and `thrown_retrieve_verb` methods.

stabbity.lic
gerritlansing: Next release will output `@target` in this message.

stabbity.lic
gerritlansing: Next release will eliminate the default value.

stabbity.lic
gerritlansing: Renamed variable to `weapon_type` in next release

stabbity.lic
gerritlansing: This code was removed as part of the implementation of `thrown_attack_verb`

stabbity.lic
gerritlansing: Stole that code from someone else. Looking at it, I believe it could be simplified to one line, but would appreciate your thoughts on this approach:
```
  def watch_target
    # Actions as of DuskRuin 2019: pedal, bob, duck, jump, lean, cower
    @arena_trap = bput("watch #{@target}", /(?:pedal|bob|duck|jump|lean|cower)/)
  end
```

stabbity.lic
gerritlansing: Looked over some logs from Dusk Ruin and all the actions share the same base message. Next version will match that.

stabbity.lic
gerritlansing: Documentation added to elanthipedia and link added to script header in next version: https://elanthipedia.play.net/Lich_script_repository#stabbity.

stabbity.lic
gerritlansing: I thought about working to untangle LootProcess from combat-trainer, but that also requires untangling the whole base GameState too, which just seems... unattractive. So, I could implement a more limited form of looting (coins, boxes, gems, and skins) which in the context of the intended purpose of this script seems ok. Thoughts?

gerritlansing: @rpherbig ^

stabbity.lic
gerritlansing: @rpherbig ^

rpherbig: Looks good to me

stabbity.lic
rpherbig: New baby incoming, can't process that. @Sarvatt @jandersson (or anyone else) can you provide some guidance?

jandersson: Theres been discussion about breaking out the classes into individual files in which case you could possibly import the lootprocess class and gamestate class and make use of it?

gerritlansing: @jandersson I was considering that approach, but then I also need to implement the same model `execute` loop as combat trainer for the process to work, which requires a lot more rework, and it seems like it will introduce far more complexity than desired to just add loot support.

PR #3508, opened 2019-04-08
update base.yaml - add other outfit tools

BinuDR:
Added more outfitting tools to the base list.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -460,6 +460,10 @@ tinkering_tools:
 - carving knife
 outfitting_tools:
 - knitting needle
+- sewing needle
+- awl
+- yardstick
+- scissors
 shaping_tools:
 - carving knife
 - shaper


Sarvatt:
is this ok if you only having knitting needles? it was only knitting on purpose for the first 401 ranks, making everyone buy the full set and repair them by default doesn't seem right

Sarvatt:
which leads to people missing tools when they repair in another craft script oddly enough

BinuDR:
If you only have knitting needles, just override this in your yaml with
outfitting_tools:
  - knitting needles

I'm not sure how it would behave if you try to repair when you only have needles without overriding it.

Sarvatt:
1-401 ranks is the vast majority of the user base imo, defaults that work out of the box for them is what i was concerned about

BinuDR:
I'll test it and see what happens.

Sarvatt:
;workorder outfitting repair with only knitting needles should be all it needs to test

PR #3509, opened 2019-04-08
register.lic - updated get logic

BinuDR:
Changed the logic for getting the register to try again if it isn't in your crafting container.  Now it tries your crafting container first, and then tries to just get your deed register.  If that fails, it messages as such exits.  Otherwise it continues to search the register and then puts it in the right container.

### DIFF
--- a/register.lic
+++ b/register.lic
@@ -22,13 +22,14 @@ class Register
     container = get_settings.crafting_container
 
     stow_hands
-    if get_item('deed register', container)
-      bput('turn my register to contents', 'You flip your deed register to the contents page.', 'But a deed register is already at the table of contents!')
-      search(args.query)
-    else
-      echo "Can\'t find your deed register."
-      exit
+    unless get_item('deed register', container)
+      unless bput('get my deed register', /You get/, /What were you referring to/) == 'You get'
+        echo "Can\'t find your deed register."
+        exit
+      end
     end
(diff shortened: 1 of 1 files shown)


register.lic
```diff
@@ -22,13 +22,14 @@ class Register
...
 
     stow_hands
-    if get_item('deed register', container)
-      bput('turn my register to contents', 'You flip your deed register to the contents page.', 'But a deed register is already at the table of contents!')
-      search(args.query)
-    else
-      echo "Can\'t find your deed register."
-      exit
+    unless get_item('deed register', container)
+      unless bput('get my deed register', /You get/, /What were you referring to/) == 'You get'
```
rpherbig: 💡 This my read a little clearer as `if bput(...) != 'You get'`

BinuDR: True.  I was using an if at first but switched to unless later.  I can always change it if it would be better for the overall style we're aiming for.

Sarvatt:
double unless is kind of confusing to people starting out. it was to me at
least. just ruby preferring unless was the hard part. if there is any way
you can clear it up the logic for future contributers it would be
appreciated.

BinuDR:
I'll take care of it soon.  I hope other people actually use this script.  I love it

Sarvatt:
i love that recommending you use this script as a template starting out was a thing :)

PR #3516, opened 2019-04-11
Common crafting change to prevent script exit on repair

Tarhish:
The default behavior for get_crafting_item works for gathering tools to use in a craft, but often fails for repairing tools due to duplicated entries in various places.

A week from now we can change ;workorders to use that arg for repairs.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -117,7 +117,7 @@ module DRCC
     Regexp.last_match(1)
   end
 
-  def get_crafting_item(name, bag, bag_items, belt)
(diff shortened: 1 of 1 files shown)


Sarvatt:
is this a fix or https://github.com/rpherbig/dr-scripts/pull/3508?

Sarvatt:
this is the exact problem I was worried about it adding

Tarhish:
I think we were doomed no matter what we did.  This problem existed long before the outfitting change.  It'll just be nice to have a standard 'get crafting item but no big deal if it doesn't exist' method.

PR #3519, opened 2019-04-12
;combat-trainer worn box-noun support

Tarhish:
As someone who wears an enlarged 'cartographer's trunk', the parser gets confused when it tries to do STOW TRUNK to a dropped mob box.  STOW OTHER TRUNK does not resolve the issue, but 'STOW STEEL TRUNK' or 'deobar trunk' or whatever does.

If it gets 'already in your inventory' it searches the room for the item's full description (code pulled from stow_lootables), then pulls the last two words to stow.

This has been working for me for a few months now of fighting box-dropping mobs, so I feel pretty good about it.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -481,6 +481,9 @@ class LootProcess
     when 'already in your inventory'
       if @gem_nouns.include?(item)
         bput('stow gem', 'You pick up', 'You get', 'You need a free hand', 'You just can\'t', 'push you over the item limit', 'You stop as you realize the .* is not yours', 'Stow what', 'already in your inventory')
+      elsif @box_nouns.include?(item)
+        item_reg = item.split.join('.*')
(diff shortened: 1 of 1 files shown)


Sarvatt:
thank you from someone who wears a meat locker box and just worked around it by droppping box from the loot list completely :D

Sarvatt:
pick will still be broken though, it'll try to get trunk and get stuck?

Tarhish:
Huh - I've cycled through some trunks with ;pick but I wasn't paying attention to what exactly it was doing.  I think the `get my #{box} from my #{source}` will get around that in a way that STOW BOX didn't, but I'll keep an eye out to see if it's been leaving them on the ground or something.

Sarvatt:
oh good point, i didn't look at pick and you're right it should get the box from the container, sorry for the noise!

PR #3521, opened 2019-04-12
base.yaml: Switch outfitting_tools back to only knitting needles.

Sarvatt:
Having the others breaks people who only care about knitting, (aka everyone under 401 ranks) and have not bought all of the sewing tools which you only need for 401+ ranks when using workorders. Non workorder sewing using craft.lic uses knitting only until 650 ranks also. This reverts 6f2ac50613345546dd6538d823859fd58b0a77e1


rpherbig:
@BinuDR FYI

PR #3529, opened 2019-04-19
Add more hometown control to dependency

Tarhish:
Another change I'm poking around for thoughts on.

Allowing a user to change hometown by UserVars like other settings here seems like a no-brainer, especially since you can now set up a safe-room hash for each town.  I've got a safe-room setup I cut paste to all my alts and I haven't needed to adjust the room numbers for months, so I can adjust my hometown even without having to touch my yaml.

The big change here would be that a script can now set a $DYNAMIC_HOMETOWN global variable to override the hometown setting everywhere.  I wanted to use a UserVars initially, but I got frightened by the thought of trying to troubleshoot someone that lost power in a way that before_dying didn't get a chance to nil it...  What a nightmare that would be.

My first use-case is in ;trade.  When I arrive at Muspar'i I can drop off my caravan at the guild then set $DYNAMIC_HOMETOWN = "Muspar'i" and trust that ;safe-room, ;workorders, buying yarn/lumber, or any bank-stuff are all going to do their business in-town and not try to barge back towards my yaml hometown.  When I'm done, $DYNAMIC_HOMETOWN = nil.

As an example, (where closest_town returns the closest town with a trader_outpost in base-town and time_to_room() gives the map timeto distance between two rooms)

```
  def check_forging
    return unless @caravan_training_skills.include?('Forging')
    return unless check_ability?('Forging', @caravan_training_skills['Forging'])
    return unless @crafting_data['blacksmithing'][closest_town]
    return if time_to_room(Room.current.id, @crafting_data['blacksmithing'][closest_town]['stock-room']) > 60
    $DYNAMIC_HOMETOWN = closest_town
    wait_for_script_to_complete('workorders', ['blacksmithing'])
    $DYNAMIC_HOMETOWN = nil
    @cooldown_timers['Forging'] = Time.now
  end
```

Or, in other words, if the closest town has a blacksmithing entry, then check to see if you can get there in under 60 seconds (looking at you, Langenfirth -> Riverhaven crafting shunt) and if so set $DYNAMIC_HOMETOWN and start workorders.  This means whenever you call check_forging in/near a city that has a forge it'll go there, do a work order, do a repair, then nil $DYNAMIC_HOMETOWN and get back to business.

It might also eventually be useful for stealing in other towns (set hometown right before moving onto another town, then unset it when done or before_dying), or making crafting less location-specific like above.

I've run ;trade up in Therengia for about 60 hours with this going and an order to let me know if it ever starts running somewhere more than a minute away,  Nothing unusual happened.  This should be safe enough on its own as long as any script that implements it is tested thoroughly enough, but I'm going to keep thinking about unintended side-effects for a while.  I'm not above scrapping it either.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -919,6 +919,8 @@ class SetupFiles
     s.bankbot_name ||= UserVars.bankbot_name
     s.bankbot_room_id ||= UserVars.bankbot_room_id
     s.prehunt_buffs ||= UserVars.prehunt_buffs
+    s.hometown ||= UserVars.hometown
+    s.hometown = $DYNAMIC_HOMETOWN if $DYNAMIC_HOMETOWN
 
     # Pull town-specific room_ids iff defined
     s.safe_room = s.safe_room[s.hometown] if s.safe_room.is_a?(Hash) && s.safe_room[s.hometown]
--- a/dependency.lic
+++ b/dependency.lic
@@ -920,7 +920,7 @@ class SetupFiles
     s.bankbot_room_id ||= UserVars.bankbot_room_id
     s.prehunt_buffs ||= UserVars.prehunt_buffs
     s.hometown ||= UserVars.hometown
-    s.hometown = $DYNAMIC_HOMETOWN if $DYNAMIC_HOMETOWN
+    s.hometown = $HOMETOWN if $HOMETOWN
 
     # Pull town-specific room_ids iff defined
     s.safe_room = s.safe_room[s.hometown] if s.safe_room.is_a?(Hash) && s.safe_room[s.hometown]
@@ -1263,6 +1263,14 @@ end
 
 def update_moon_data(moon, data)
   $moon_watch.queue_update(moon, data)
+end
+
+def shift_hometown(town_name)
+  $HOMETOWN = town_name
+end
+
+def clear_hometown
+  $HOMETOWN = nil
 end
 
 full_install if install


rpherbig:
@Sarvatt @jandersson thoughts?

Sarvatt:
This will work. I wonder if it would be possible to just plain set anchor points in the map and make this easier. aka if you are closest to the center of leth you can do things in leth because its faster without the scripts knowing.

Tarhish:
I'm thinking of moving some of the methods jonas or I put in trade into common travel. Specifically find_closest_id which lets you give a key like 'deposit' or 'trader_outpost, and it retrieves the closest room id for that entity in base town or base crafting or whatever you give it.  And the time-between-rooms method.

A 'find closest room with X tag' method would be cool too, except I'm thinking I already saw something like that in lich.rbw when I was poking around.  I'll take a look later to see if I'm imagining it or if it's usable for our purposes.

rpherbig:
> A 'find closest room with X tag' method would be cool too

This might be a good starting point: https://github.com/rpherbig/dr-scripts/blob/1d747ef6b354a1d3f425cb4f8e72e602199d89fe/common-travel.lic#L115

Tarhish:
Yeah, jandersson made good use of that in the original trade and I stole it.

I poked around to find what I was thinking of and found these in Class Map in lich.rbw
```
   def find_nearest_by_tag(tag_name)
      target_list = Array.new
      @@list.each { |room| target_list.push(room.id) if room.tags.include?(tag_name) }
      previous, shortest_distances = Map.dijkstra(@id, target_list)
      if target_list.include?(@id)
         @id
      else
         target_list.delete_if { |room_num| shortest_distances[room_num].nil? }
         target_list.sort { |a,b| shortest_distances[a] <=> shortest_distances[b] }.first
      end
   end
   def find_all_nearest_by_tag(tag_name)
      target_list = Array.new
      @@list.each { |room| target_list.push(room.id) if room.tags.include?(tag_name) }
      previous, shortest_distances = Map.dijkstra(@id)
      target_list.delete_if { |room_num| shortest_distances[room_num].nil? }
      target_list.sort { |a,b| shortest_distances[a] <=> shortest_distances[b] }
   end
```

I have _NO_ idea if these work.

Sarvatt:
Do you still think this is the proper way to do it instead of making a tag in each town with the name instead and checking that tag via find_all_nearest_by_tag? using globals seems dirty but if you're using it and it works it can always be iterated on later if anyone cares to fix it, working and making life better is more important.

Sarvatt:
asking people to rewrite things in reviews they do like this is a detriment to progression, this is clean enough as it is that we can accept it and if anyone wants to change it it can be changed, there is nothing wrong with this. I'm sorry for asking questions that might have detracted you from contributing.

Tarhish:
Oh without a doubt each script could be changed to work with the closest hometown using some methods.  I use some stuff in ;trade to find the closest town, and there's no reason we couldn't make, say, crossing-repair accept an arg to use a specific town, or improve common money's specific town handling, etc. etc.  That's probably more correct?

It's just sooooo convenient this way.  Not only do I not have to change other scripts to make them work, but if I'm up in Muspar'i or something and I want to repair I can just do

···
;e $DYNAMIC_HOMETOWN="Muspar'i"
;crossing-repair
;e $DYNAMIC_HOMETOWN=nil
```
and I don't even have to change my yaml!

No, but seriously, I've been thinking about other options.  The 'add args to other scripts to use local towns' idea sounds like it could be very difficult, but I think I know enough about dependency now to solve it another way.

If there are any downsides to this then I'm not aware of them, but that's mainly because I have very narrow programming experience, not because I feel this is definitely right.  It's just sooo easy.

Tarhish:
For the record, I haven't once felt discouraged here.  I like that this place has some actual standards, and I've had an opportunity to learn a great deal in an interesting context.

That's on all of you, so thanks for that.

rpherbig:
I apparently missed the update on this PR.

@Tarhish @Sarvatt where are we on this one?

Tarhish:
@rpherbig Whoops, sorry, I have been busy and wasn't following up with this stuff.

So now this introduces a new dependency method to set the town you're running in.  It's pretty much necessary for advanced stuff in ;trade, especially in Therengia where returning to your yaml hometown from, say, Muspar'i can result in a terrific failure, but any script would have to be edited to use it.  Since it's just running dependency methods I don't think it'll trip over itself, but I'd immediately add some of its functionality to ;trade.

PR #3536, opened 2019-04-24
Shaping cash on hand is 5 gold

jandersson:
This was the only workorder task requiring 1 plat and it was also the oldest commit in this file ensuring cash on hand before doing the workorder. This will bring it in line with the other workorder disciplines; no sense in running to the bank for 1 plat.


jandersson:
Annnnd its gone

PR #3537, opened 2019-04-25
afk.lic - deal with hand objects and climbing

BinuDR:
There is an issue with the cliff at asketi's mount where it will try to climb the cliff but will fail because something's in the hands, and then it just keeps trying and failing.  This addresses that.

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -57,6 +57,7 @@ loop do
 
   if line =~ /^You should empty your hands first/
     equipment_manager.empty_hands
+    stop_script('go2') if Script.running?('go2')
   end
   
   if line =~ /^You notice (.*) at your feet/
--- a/afk.lic
+++ b/afk.lic
@@ -55,6 +55,10 @@ loop do
     end
   end
 
+  if line =~ /^You should empty your hands first/
+    equipment_manager.empty_hands
+  end
+  
   if line =~ /^You notice (.*) at your feet/
     match = Regexp.last_match(1)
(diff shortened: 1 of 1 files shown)


jandersson:
I feel like this belongs in common-travel::walk_to https://github.com/rpherbig/dr-scripts/blob/master/common-travel.lic#L11

BinuDR:
It needs to be part of go2 really.  The items in hand or at feet issue effect go2, and by extension walk_to.

If we fix go2 it'll take care of walk_to on its own.

When I have some time to dig into go2 more deeply, I'll see if I can fix it up in there.

jandersson:
Its a solution to a travel related problem in a script to handle afk problems. Its adding a weird dependency.

BinuDR:
I agree.  I didn't like the at feet stuff being handled in afk either.  We'll put it on the to do list to move these things to go2.

BinuDR:
And while we're tinkering with go2, let's add the ability to have it release spells that get you in trouble in towns.

jandersson:
It doesn't belong in afk.

jandersson:
walk_to is meant to handle travel related problems that go2 does not address. this would be where the handling of at feet items should probably be as well. i am trying to push against adding more things to afk as a catchall.

rpherbig:
I dislike this being in `afk`, but if we need it, we need it. I hope someone can look into a better solution, though.

PR #3539, opened 2019-04-25
Update levelup.lic to include shard thief guild

DRWafflesworth:
[levelup]>ask Saishla about circle

"Congratulations Dyler!" Saishla exclaims, "You are ready to train for your next level."

Saishla adds, "You may be qualified to learn another ability, now."
She continues, "The guild is watching your progress carefully, and we are very pleased with what we see. Continue in this manner, and who knows, someday you might sit in my chair."

[levelup: *** No match was found after 15 seconds, dumping info]


rpherbig:
Thank you for contributing @DRWafflesworth !

PR #3540, opened 2019-04-25
;performance support for cowbells

Tarhish:
Could do a little more than this, but for now I added a worn_instrument variable that defaults to zills.


Tarhish:
Oh, I didn't even realize this disappointed Travis.  Looks like a bput thing again.  I'll try reopening it.

PR #3541, opened 2019-04-26
Update base-recipes to include throwing dagger

DRWafflesworth:
Throwing dagger was missing


SekmehtDR:
Can you please also add the "short cord" part to the handclaws. Pretty please.

- name: some metal hand claws
  noun: claws
  volume: 4
  type: weaponsmithing
  work_order: true
  chapter: 8
  part:
  - short cord

PR #3542, opened 2019-04-26
;performance support for things other than zills

Tarhish:
Could do a little more than this, but for now I added a worn_instrument variable that defaults to zills.

Resubmit to make Travis happier.

### DIFF
--- a/performance.lic
+++ b/performance.lic
@@ -26,6 +26,7 @@ class Performance
     fput('release ecry') if DRSpells.active_spells["Eillie's Cry"].to_i > 0
 
     instrument = @settings.instrument
+    @worn_instrument = @settings.worn_instrument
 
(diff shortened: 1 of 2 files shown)


rpherbig:
Do you plan to also update `athletics.lic`, `coordinator.lic`, or `crossing-training.lic`?

Tarhish:
Yeah, might as well do that now while I've got it going.  The only thing I can't touch is ;play

Tarhish:
I was busy the last couple days, but I actually want to commonize the cleaning and playing since I'm touching this.  Shouldn't be too long.

Tarhish:
Yep, commonizing it worked better than expected.  I'm going to close this and submit the common changes after a few more tests, probably tomorrow.

PR #3550, opened 2019-05-06
Fixed the quest where it locks up

jayRyan24:
Fixed the quest where it locks up on your human form to just your, matching line 151.


Sarvatt:
this is a good change i swear i already submitted when i hit it.

PR #3551, opened 2019-05-06
Fixed possible hang due to missing therugy_container

OldAssCleric:
Theurgy.lic was using only "Stow" to put away holy_water instead of putting it in theurgy container. This caused Theurgy to hang the next time it tried to get  holy_water from the theurgy_container as stow may not be the same container. 

added -  put_in_container(@water_holder) - at line 303 just above stow_hands.


Tarhish:
This looks pretty open and shut to me.  All the other places that use holy water actually do this already -- this looks like the odd one out.

Sarvatt:
+1 from me too, it will be a bit before the change gets accepted though. bossman is off having a kid :)

PR #3552, opened 2019-05-12
Coordinator: Pass settings to ensure_copper_on_hand

jandersson:
Small change that fixes usage of ensure_copper_on hand

### DIFF
--- a/coordinator.lic
+++ b/coordinator.lic
@@ -343,19 +343,19 @@ class Coordinator
   end
 
   def train_outfitting(_task)
-    return unless ensure_copper_on_hand(5000, @settings.hometown)
+    return unless ensure_copper_on_hand(5000, @settings)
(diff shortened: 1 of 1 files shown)


gerritlansing:
Looks good to me.

Sarvatt:
are these lines even required actually? workorders checks for money itself and its different amounts than here, it might have just been cribbled out of crossing-training when it did non workorders also

Sarvatt:
its potentially one more crafting room to society trip here, this is requiring 5 gold for shaping and workorders goes and gets a plat

Sarvatt:
i could be wrong, on a phone and cant look it up but im almost positive workorders handles money on its own and this ensure copper on handle was just in crossig-training to handle he non workorders use case back in the day

jandersson:
I checked - You are right, these lines should just be removed instead

PR #3555, opened 2019-05-14
Added some extra attributes to DRRoom

dr-leustyin:


drinfomon.lic
```diff
@@ -694,9 +721,14 @@ status_action = proc do |server_string|
     DRRoom.room_objs = []
   elsif server_string =~ /^you \w+ (west|east|north|south|northeast|northwest|southeast|southwest|out)\./i
     DRRoom.group_members = []
+  elsif server_string =~ /Obvious exits:/ or server_string =~ /Obvious paths:/
```
Sarvatt: would you mind changing this to || instead of or? ruby is really weird about or, || is for boolean logic and ‘or’ is for control flow and lower presedence

Sarvatt: rubocop might catch that actually

jandersson: yea, sarvatts pinned it basically... `or` `and` etc are banned 
```
The and and or keywords are banned. The minimal added readability is just not worth the high probability of introducing subtle bugs. For boolean expressions, always use && and || instead. For flow control, use if and unless; && and || are also acceptable but less clear. [link]
```

dr-leustyin: Found a better method

Sarvatt: that or basically translates to elsif server_string yadda unless server_string foobar, so easy to have a subtle bug there

Sarvatt:
looks good to me. i am curious why you are adding it though :) this wont kill drinfomon if you walk through a dark room with no exits right?

dr-leustyin:
Just looking for some QOL improvements to get data from unknown rooms.  I prefer to work with DRRoom vs XMLData

dr-leustyin:
Just found that dark rooms, when you have a light source, send the XML as [].  I think it makes more sense to go back to parsing Obvious exits|paths so that the data is correct even when the server sends bad data.

Thoughts?

jandersson:
The Room class has these attributes

Room.current.paths
Room.current.description
etc...

You could have DRRoom inherit Room or use Room itself or copy the logic for Room.paths

jandersson:
It sounds like what youre after is like DRRoom is inheriting from Room? Like you have two room classes and you just want to use one... 

I havent tried that, im not sure why it wouldn't work - but it sounds interesting. I could be totally off.

jandersson:
Ah wait, Im forgetting that Room is basically a shortcut for the map database rooms. Nevermind. Nothing to see here!

rpherbig:
Looks good to me. Still working on it?

dr-leustyin:
All set now.  This one is ready

PR #3561, opened 2019-05-18
almanac.lic: Add support for the droghtmans almanacs

### LINKED ISSUE #3559: New Almanacs
Greetings folks,

I have acquired one of the new almanacs and am excited to use it. According to a very knowledgeable person, I was told that you were able to use the almanac to keep 3 primary, 5 secondary, or 6 tertiary skills. Ideally having a set of variables to cycle through (maybe 3?) would be super.

The current almanac script doesn't work as each time that you go to study the new almanac you have to TURN TO SKILL (e.g. forging, astrology, athletics, etc). It works on a 10 min timer much like the original almanac. Let me know how I can assist with creating this script if you're willing to take it on. Thanks!

Sarvatt:
These can be turned to individual skills. This adds support for a yaml setting to pick which skills to use, and it will pick whichever skill is at the lowest mindstate to study. If none of the skills are under 18/34 it will loop every 20 seconds checking to use it when one is.

Skills to use can be customized like

```
almanac_skills:
- Mechanical Lore
- Outdoorsmanship
- Scouting

```
Fixes #3559

### DIFF
--- a/almanac.lic
+++ b/almanac.lic
@@ -33,7 +33,7 @@ class Almanac
     Script.running.find_all { |s| !s.paused? && !s.no_pause_all && s.name != 'almanac' }.each(&:pause)
     waitrt?
     bput('stow left', 'Stow what', 'You put') if checkleft
-    case bput('get my almanac', 'You get', 'What were')
+    case bput('get my almanac', 'You get', 'What were', 'You are already')
     when 'What were'
       echo('Almanac not found, exiting.')
       Script.running.find_all { |s| s.paused? && !s.no_pause_all && s.name != 'almanac' }.each(&:unpause)
--- a/almanac.lic
+++ b/almanac.lic
@@ -8,8 +8,10 @@ class Almanac
   include DRC
 
   def initialize
+    settings = get_settings
     UserVars.almanac_last_use ||= Time.now - 600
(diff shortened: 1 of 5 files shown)


profiles/base.yaml
```diff
@@ -1190,6 +1190,8 @@ forge_container:
 spell_cycle:
 
 restock_shop:
+# For almanacs that let you pick which skill to use
+almanac_skills:
```
rpherbig: What do you think about adding a validator for this? I.e. that everything in the array is a valid skill

almanac.lic
```diff
@@ -20,6 +22,14 @@ class Almanac
     return if XMLData.room_title.include? 'Carousel Chamber'
     return if hidden?
 
+    if @almanac_skills
+      training_skill = @almanac_skills
+                       .select { |skill| @almanac_skills }
```
rpherbig: What does this line do?

Sarvatt: I don't know besides it working, was a cut and paste from another place :) Is there a more proper way to test all the skills in almanac_skills: for < 18 mindstates and use that skill?

profiles/base.yaml
Sarvatt: added a test to validate

xtremenerdrage:
Been running this for a week in and out of combat with no issues.

PR #3564, opened 2019-05-19
Update to afk.lic for grabbing at foot ammo

DRWafflesworth:
The get step is not needed and due to a bug with the new Droughtman's blunt-tip arrows, afk will not grab them. Individual arrows show up as "arrows" instead of "arrow" and if you "get my arrows" it will grab the arrows on your person and get stuck in a loop.

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -65,14 +65,12 @@ loop do
     item = DRC.get_noun(match)
     case DRC.bput('inv', /Lying at your feet is/, /Lying at your feet are (.*)\./)
     when /Lying at your feet is/
-      DRC.bput("get my #{item}", 'You get', 'You pick up', 'But that is already in your inventory', /What were you referring to/)
-      equipment_manager.empty_hands
+      DRC.bput("stow my #{item}", 'You get', 'You pick up', 'But that is already in your inventory', /What were you referring to/)
     when /Lying at your feet are (.*)\./
       at_feet = Regexp.last_match(1)
       items = DRC.list_to_nouns(at_feet)
       items.each do |thing|
-        DRC.bput("get my #{thing}", 'You get', 'You pick up', 'But that is already in your inventory', /What were you referring to/)
-        equipment_manager.empty_hands
+        DRC.bput("stow my #{thing}", 'You get', 'You pick up', 'But that is already in your inventory', /What were you referring to/)
       end
     end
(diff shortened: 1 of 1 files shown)


Tarhish:
Definitely a problem, but I was running through why this would've been written like that in my head.  While I think this change would work better for ammo, there's a different behavior for thrown weapons.

Specifically, empty_hands handles gear items, so a thrown weapon that's tied or sheathed or whatever would get put away per its declared attributes.  This would just STOW it. 

There might be some weird cases where it causes a problem?  I'll think about it some more but maybe there's an easy fix.

simiydr:
@Tarhish Perhaps if `item_by_desc` returns an item, then get it from the ground and use `stow_weapon`, otherwise just use stow?

Tarhish:
Looks fine to me at least.  I followed those methods but didn't really see anything that would be a problem.  It'll grab the item by noun on the floor, but then return_held_items will read the short description with the adjective, so... I don't see why not?

I think it could be done just like you say,  if item_by_desc(item) or if item_by_desc(thing) then work as normal, else do it per this change.

jandersson:
I think Tarhish makes a good point with it handling worn items, but since the stow occurs first it makes me believe that its intended as a fall-back incase you get a "that doesn't fit" message. Its likely that you picked up a thrown weapon that doesn't fit in the container. @BinuDR thoughts?

DRWafflesworth:
Is it possible for combat trainer to exit with thrown weapons on the ground? Personally I have only ever seen ammo. Would it ever have a case where you don't pickup a thrown weapon after lobbing? I think this PR can be cancelled as the only reason for this change is a quick fix for a bug with the new Droughtman's arrows.

[go2]>west
You notice a blunt-tipped arrows with chevron stripe at your feet, and do not wish to leave it behind.
>
[afk]>inv
Lying at your feet are a blunt-tipped arrows with chevron stripe and a blunt-tipped arrows with chevron stripe.

[Type INVENTORY HELP for more options]
>
[afk]>get my arrows
You get some blunt-tipped arrows with chevron stripes from inside your thigh quiver.
>
[afk]>stow right
You put your arrows in your thigh quiver.
>
[afk]>get my arrows
You get some blunt-tipped arrows with chevron stripes from inside your thigh quiver.
>
[afk]>stow right
You put your arrows in your thigh quiver.

PR #3569, opened 2019-05-23
AFK: Add check for user and script input timer

jandersson:
This helps avoid cases where the game logs a user out without sending an idle message

We use $_IDLE and $_SCRIPTIDLE to see the last time a user sent a command and the last time a script sent a command. If its more than 6 minutes then we send the same message we would when receiving a "you are idle" message.

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -15,6 +15,12 @@ def exit_game(message)
   fput('exit')
 end
 
+def idle?(line)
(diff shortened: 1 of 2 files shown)


jandersson:
I suspect Travis CI doesn't like this because we dont have $_IDLETIMESTAMP_ and $_SCRIPTIDLETIMESTAMP_  defined in our test harness. I will try adding those

PR #3573, opened 2019-05-27
Fix bug with using global idle timers before initialized

jandersson:
The lich global idle timers are not initialized immediately after
startup, but rather a few seconds after. This would cause afk.lic
to crash when autostarted. This fix alleviates that by skip their
check if falsey.

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -17,7 +17,9 @@ end
 
 def idle?(line)
   return true if line =~ /you have been idle too long/i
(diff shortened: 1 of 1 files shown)


jandersson:
Hi @sheltim sorry about the immediate bug fix. This is going to address an issue with afk where the global timers that were used are not initialized by Lich immediately. This causes afk.lic to crash if on autostart. The global variables are at some point initialized shortly after autostarts finish.

jandersson:
@rpherbig

jandersson:
Thanks for the quick merge! Sorry about the trouble

PR #3574, opened 2019-05-27
Added the ability to set a discern timer.

jayRyan24:
Really helps lower circle players. Set to default to 24, which is what the hard-coded value was.


profiles/base.yaml
```diff
@@ -294,6 +294,8 @@ cambrinth_cap: 32
 stored_cambrinth: false
 # used with use_auto_mana. Multiplies your discern limit by this number to get a safe mana amount.
 prep_scaling_factor: 0.85
+# used with use_auto_mana. How often it will check discerns. Lower circle/ranks should check more often.
```
rpherbig: 💡 it would nice to note the units here, i.e. 'hours'.

PR #3577, opened 2019-05-28
Added fix for blunt-tipped arrows

dr-leustyin:


jayRyan24:
Just adding this for clarity, it's a server-side error in the naming convention:

```
Moving with the precision of a mongoose, you fire a blunt-tipped arrows with chevron stripe at a dour forager goblin.
```

rpherbig:
Thank you for the explanation @jayRyan24

PR #3578, opened 2019-05-28
Updated collecting dual-load arrows to CT

jayRyan24:
This lets you pick up both ammo if dual loading a bow.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2664,7 +2664,7 @@ class AttackProcess
       game_state.action_taken
       ammo = Regexp.last_match(2)
       bput("stow my #{ammo}", 'You pick up', 'Stow what')
-      if game_state.dual_load? && game_state.current_weapon_skill.eql?('Bows')
+      if game_state.dual_load? && game_state.weapon_skill.eql?('Bows')
         ammo = Regexp.last_match(2)
         bput("stow my #{ammo}", 'You pick up', 'Stow what') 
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2664,6 +2664,10 @@ class AttackProcess
       game_state.action_taken
       ammo = Regexp.last_match(2)
       bput("stow my #{ammo}", 'You pick up', 'Stow what')
+      if game_state.dual_load? && game_state.weapon_skill.eql?('Bows')
```
rpherbig: ❓  The `dual_load?` function checks that weapon skill is 'Bow', which is different than 'Bows' here (note the plural). Why the difference?

jayRyan24: Was an error, I just have dual_load? in mine, and I dunno why I added in the skill check, also Bows should be Bow. I just didn't translate it over. Removed altogether.

combat-trainer.lic
```diff
@@ -2664,6 +2664,10 @@ class AttackProcess
       game_state.action_taken
       ammo = Regexp.last_match(2)
       bput("stow my #{ammo}", 'You pick up', 'Stow what')
+      if game_state.dual_load?
+        ammo = Regexp.last_match(2)
```
rpherbig: 💡 I think `ammo` here will be the same as the previous assignment, so this could be removed?

Issue #3576, opened 2019-05-28
Add units to the check_discern_timer yaml option.


Sarvatt:
See https://github.com/rpherbig/dr-scripts/pull/3574#discussion_r287898714

PR #3582, opened 2019-05-31
Added clarity to the comments for check_discern_timer.

jayRyan24:


jayRyan24:
Resolves: #3576

jandersson:
I think its a lot clearer if the unit is part of the variable
e.g. 
```yaml
time_to_self_destruction: 360
```
vs
```yaml
minutes_to_self_destruction: 360
```

jayRyan24:
I can do that, but then so many other settings need to be reworked.  I think the explanation in the comment is probably fine and sufficient. 

Just from a quick scan:
```
cycle_armors_time
favor_orb_rub_frequency
repair_timer
aura_frequency
performance_pause
```

jandersson:
I think a find and replace would work pretty well since this is a new feature

jayRyan24:
What?

jandersson:
find and replace check_discern_timer: with check_discern_timer_hours:

jayRyan24:
I mean, I understand how to change a variable name? LOL, wtf. I meant that it hadn't been implemented into user's yamls due to it being new so it shouldn't really have a large affect on the user base.

jandersson:
I was just responding to the 'what?'

jayRyan24:
Yeah, that's why I added the second part to my last comment. I think there was a disconnect. No worries,

rcuhljr:
Way to cross wires noobs. Jay was indicating that if we change this one we should probably change the multitude other time specific settings to include their units (which I'm a fan of hungarian notation in this case just because there's a clear separation between user space and software space and we don't have a good editor or notation system in place for settings.

jayRyan24:
So you want it to be `iCheckDiscernTimer` or `hrsCheckDiscernTimer`. Not really familiar with how weakly typed languages handle naming conventions, specifically because a variable can change type during runtime, even by accident, which could get really confusing.

jandersson:
Ah ok then I definitely misunderstood - I dont think previous settings should be setting precedence. I'm a fan of making all future ones clear

jayRyan24:
I honestly get that, and got it, I am probably being a "let's just follow convention asshole" because we're kind of changing the paradigm on the fly.

Sarvatt:
the difference is, all of the others are in seconds and this is not. Chuno just delights in breaking peoples yamls :D

jayRyan24:
Anything to create more work for me, that's the goal typically.

rcuhljr:
Hungarian was in specific a joke, but the concept of having the unit in time yamls makes sense. Just because we did something poorly in the past is no reason to continue to make the same mistake in the future with new settings.

jayRyan24:
Yes it is, I can't be an asshole if not.

jandersson:
> I honestly get that, and got it, I am probably being a "let's just follow convention asshole" because we're kind of changing the paradigm on the fly.

No worries, and I am splitting hairs to be sure :P was just trying to make a push for the variable change

jayRyan24:
To be quite frank, it was when he gave me examples that made my inner asshole appear. "Add the unit to the variable name." is sufficient. Sorry, I'm grumpy. Don't be like me.

Also, the reason I didn't from the start was mainly because I felt like "well now we'll break yamls that have implemented this change", etc. Again, don't be like me.

jandersson:
I think my favorite quote from the founder of stack overflow went something like "If you spend your day working with assholes, what do you become? An asshole. Well the biggest asshole you can ever work with is a computer, so its no wonder that most people who work with them are assholes."

That being said, I think you should leave the old setting in base.yaml and mark it to be deleted. So in the case where anyone is using the setting doesn't experience breakage.

jayRyan24:
@rpherbig

rpherbig:
I'm ok with this if you add a validator

validate.lic
```diff
@@ -478,6 +478,12 @@ class DRYamlValidator
       .each { |skill| error("Magic skill in crossing_training does not have a training_nonspells defined: #{skill}") }
   end
 
+  def assert_that_check_discern_timer_is_deprecated(settings)
+    return unless settings.check_discern_timer
+
+    warn('The check_discern_timer setting is deprecated. It is now check_discern_timer_in_hours for clarity. Please adjust accordingly.')
```
rpherbig: 💡 I would expect this to be an `error` given the old setting has been removed entirely.

jayRyan24: Aaah, was changing it, lol.

jandersson:
This new setting needs to be added to base-settings, else it causes an error: 
```
[DRPrime]-DR:Erons: "anyone know what this is"                                           
[DRPrime]-DR:Erons: "--- Lich: error: comparison of Float with nil failed"               
[DRPrime]-DR:Erons: "common-arcana:498:in `>'"
[DRPrime]-DR:Erons: "common-arcana:498:in `check_discern'"
```

jandersson:
@jayRyan24

jayRyan24:
ok

jayRyan24:
It's set in base.yaml, where else does it need to be?

jandersson:
sorry i goofed and missed it when scrolling through the diff. I'm not sure what caused it tbh.

PR #3586, opened 2019-06-02
update rezz.lic - Infuse amount based on mana pool

BinuDR:
I added a method to use to determine how much to infuse for each loop. `def infuse_amount`
Set `infuse_mana` variable inside `find_soul` infuse loop using that method.  I did this so that I can also update the new variable @total_infused for the output at the end of the script.

### DIFF
--- a/rezz.lic
+++ b/rezz.lic
@@ -45,7 +45,7 @@ class Rezz
                echo "Using default values for REJUV, make a 'rezz'=>Rejuvenation waggle set to avoid this."
                get_data('spells').spell_data['Rejuvenation']
              end
-
+    @total_infused = 0
     Flags.add('no-favor', /has no favor with h.. god!/)
     Flags.add('rejuv-silver', 'A thin silver nimbus surrounds', 'As you concentrate, a thin silver nimbus flickers into view around', 'memories are protected from loss by the Resumption spell')
     Flags.add('dead', /H.. soul (is|is not) bound to h.. body./)
@@ -71,7 +71,7 @@ class Rezz
   def rejuv(player, quick = false)
     spell = quick ? { 'abbrev' => 'rejuv', 'prep_time' => 5, 'cast' => "cast #{player}", 'mana' => 10 } : @rejuv
     until Flags['rejuv-silver'] || Flags['necro']
-      pause 5 while DRStats.mana < 40
+      pause 5 while DRStats.mana < 30
       cast_spell(spell, @settings)
     end
   end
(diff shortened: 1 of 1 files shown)


rezz.lic
```diff
@@ -80,14 +80,28 @@ class Rezz
     @vigil ||= { 'abbrev' => 'vigil', 'prep_time' => 10, 'cast' => "cast #{player}", 'mana' => 20 }
     vigil(player) unless cast_spell(@vigil, @settings)
   end
+  
+  def infuse_amount
+    if DRStats.mana < 20
+      return (@rezz['mana'] / 5)
+    elsif DRStats.mana < 40
+      return (@rezz['mana'] / 4)
+    elsif DRStats.mana < 61
```
jayRyan24: Why 61 and not 60?

rezz.lic
```diff
@@ -80,14 +80,28 @@ class Rezz
...
     Flags.add('rotting', /#{player}.s body'?s? (grows paler|seems to|appears to|appearance takes)/)
     15.times do
-      pause 5 while DRStats.mana < 40
-      @settings.osrel_no_harness ? nil : harness_mana([@rezz['mana']])
-      match = bput("infuse rezz #{@rezz['mana']}", 'You sense a spirit nearby, but you are unable to make it out clearly', /spirit of #{player} in the Void/)
+      infuse_mana = infuse_amount
+      pause 5 while DRStats.mana < 20
+      @settings.osrel_no_harness ? nil : harness_mana([infuse_mana])
+      match = bput("infuse rezz #{infuse_mana}", 'You sense a spirit nearby, but you are unable to make it out clearly', /spirit of #{player} in the Void/)
       @infuse_attempts += 1
```
jayRyan24: So, you have this code:
```
    until find_soul?(player)
      unless keep_going?	      unless keep_going?
        bput("whisper #{player} I can't find your soul.  You need more favors!", 'You whisper', 'Who are you trying to whisper to?')	        bput("whisper #{player} I can't find your soul.  You need more favors!", 'You whisper', 'Who are you trying to whisper to?')
        return	        return
      end	      end
    end	    end
```
You ask every infuse for the player to put `;send yes/no`? This seems less automated than I would personally expect. Why not give the player the ability to set the attempts they want before asking? Or maybe, ask if under 30% mana only, etc. Something like that? Every iteration seems a bit excessive. Nothing to do with this PR, but I wanted to make note of it seeing how the attempts are still being used.

rezz.lic
```diff
@@ -45,7 +45,7 @@ class Rezz
                echo "Using default values for REJUV, make a 'rezz'=>Rejuvenation waggle set to avoid this."
                get_data('spells').spell_data['Rejuvenation']
              end
-
+    @total_infused = 0
```
jayRyan24: This could be put above under settings, etc. It's a little out of place, most variables should be initialized at the top of a method for clarity. I know @rezz and @rejuv are also variables being initialized, but this is a single one that can go with the others.

PR #3592, opened 2019-06-06
Added ability to override prep type for rituals (specifically for tattoos)

dr-leustyin:
…tos)

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -115,7 +115,11 @@ module DRCA
     DRC.release_invisibility
     DRC.bput('stance set 100 0 80', 'Setting your')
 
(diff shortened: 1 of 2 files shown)


common-arcana.lic
```diff
@@ -115,7 +115,11 @@ module DRCA
     DRC.release_invisibility
     DRC.bput('stance set 100 0 80', 'Setting your')
 
-    return unless prepare?(spell['abbrev'], spell['mana'], spell['symbiosis'])
+    command = 'prepare'
+    command = spell['prep'] if spell['prep']
+    command = spell['prep_type'] if spell['prep_type']
```
rpherbig: 💡 Could this be `command = spell['prep_type'] || spell['prep'] || 'prepare'`? I'm not sure that would be any better, though.

PR #3593, opened 2019-06-06
Fix handling of no game response to `load`

simiydr:
Occasionally the game will not respond to the `load` command, resulting in a hang and endless loop in aiming_trainables. Testing against an empty string (as returned by `bput` with a timeout) will allow combat-trainer to retry the load command, which in testing has been successful.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2739,12 +2739,12 @@ class AttackProcess
       end
     else
       if game_state.dual_load?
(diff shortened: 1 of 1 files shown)


Sarvatt:
awesome! I was scratching my head on how to fix it properly.

rpherbig:
@Sarvatt I'm interpreting your comment as approval of this method of fixing the issue :)

Sarvatt:
yep! thanks @rpherbig

PR #3595, opened 2019-06-07
steal.lic: Run performance if you don't hide to steal.

Sarvatt:


simiydr:
As a general question (and applied to this case) would it be better to start using the `play_song` method introduced in common in PR #3546?

Tarhish:
> As a general question (and applied to this case) would it be better to start using the `play_song` method introduced in common in PR #3546?

Nooo, regretfully it's not done yet.  It works in ;performance if I replace methods with calls to common, but I need to make a couple changes to make sure it works right.  It needs a noclean itself, or possibly a way to decouple playing from cleaning neatly and without wasted actions.

steal.lic
```diff
@@ -52,7 +52,7 @@ class Steal
 
     settings = get_settings
     setup(settings)
-
+    start_script('performance', ['noclean']) unless settings.hide_to_steal
```
simiydr: I think we probably want to move starting performance down to below the checks that might cause the script to exit quickly. So, perhaps after line 73?

simiydr:
@Tarhish Understood, thanks!

steal.lic
Tarhish: Yeah, good catch.  There's also buffs in there, which I think means spellcasters with stealing buffs will start playing, try to buff, stop playing, buff, then begin stealing.  73 looks like good to me?

@Sarvatt got any thoughts?

Sarvatt: bput handles stopping playing and the script will start again between buffs so it wouldn't hurt anything, but I moved it anyway just to be extra safe :D

simiydr:
Thanks @Sarvatt! Looks good to me.

PR #3596, opened 2019-06-07
Fix bug in choose-a-skill almanac support

simiydr:
Adding support for the new choose-a-skill almanacs broke older almanacs (the block to select a skill was always entered into). This update fixes older almanacs. 

**NOTE:** I was unable to find someone to test the newer almanacs against this.

### DIFF
--- a/almanac.lic
+++ b/almanac.lic
@@ -22,9 +22,8 @@ class Almanac
     return if XMLData.room_title.include? 'Carousel Chamber'
     return if hidden?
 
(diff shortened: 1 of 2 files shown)


rpherbig:
@sarvatt can you test this?

Sarvatt:
works fine, I just did this independently not realizing there was already a PR up. Of course I didn't have the base-empty.yaml entry in when I was testing it!

PR #3599, opened 2019-06-08
initial commit of the message method to common.lic

jayRyan24:
This method I found in Bigshot.lic from Gemstone 4. It puts up a nice monsterbold text, formatted nicely, and can be used without dependency. From here we can begin using it within the system, and even converting a lot of things over to it that we find useful or could be better [presented](https://imgur.com/nvw6fQs). 

[Here](https://imgur.com/q2696qk) is how it appears in Stormfront, using BigShot.

[Here's ](https://imgur.com/jrn7V3A) an example of it in Stormfront while playing DR.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -465,4 +465,16 @@ module DRC
     bput("stow my #{cloth}", 'You put')
     true
   end
+
+  def message(text)
+    string = ''
+    if $fake_stormfront then string.concat("\034GSL\r\n ") else string.concat("<pushBold\/>") end
+    if( text.index('\n') )
+        text.split('\n').each { |line| string.concat("| #{line}") }
+    else
+        string.concat('| ' + text)
+    end
+    if $fake_stormfront then string.concat("\034GSM\r\n ") else string.concat("<popBold\/>") end
+    _respond string
+  end
 end


jayRyan24:
I am going to look into adding this into lich.rbw's echo() method.

Sarvatt:
did you decide on doing it globally in lich or a new method here? not really a fan of having all my script text prepended with "| " to be honest.

jayRyan24:
It's not all text, it's echos from the system which already begin with [.

Sarvatt:
mostly what I was asking is if you wanted to go ahead with this way and change scripts, or do it in dr-lich because this is still open. There is nothing wrong with adding this if you want to use it other places :)

Sarvatt:
using it in debug in combat-trainer would be a good use case for this.

jayRyan24:
Ah ok, well I think it should just be a method we can apply wherever we want. It's pretty useful. Then we can go into the code and update scripts with it as we see fit.

Sarvatt:
There is no harm in adding this, approved from me :)

jayRyan24:
Trying to solve the conflict, but it won't even show up for me, just fails to load. :(

jandersson:
I think this is a cool addition, but what is `$fake_stormfront`?

jayRyan24:
> I think this is a cool addition, but what is `$fake_stormfront`?

```
   # backward compatibility
   if $frontend =~ /^(?:wizard|avalon)$/
      $fake_stormfront = true
   else
      $fake_stormfront = false
   end
```

jayRyan24:
Closing and reopening. Not going to merge 100+ conflicts. :)

PR #3602, opened 2019-06-11
Add pause_all and unpause_all to common

simiydr:
There are a handful of scripts that seek to pause/unpause all other running scripts. These methods standardize the way we handle pausing/unpausing, and introduce a few improvements:

* Won't unpause scripts that were already paused; and,
* One script will hold a lock on pause_all functions until it unpauses, to prevent two "pause all" scripts from pausing each other.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -465,4 +465,37 @@ module DRC
     bput("stow my #{cloth}", 'You put')
     true
   end
+
+  def pause_all
+    return false if $pause_all_lock
+    $pause_all_lock = caller[0][/^[a-z0-9_-]+/i]
+    @pause_all_no_unpause = []
+
+    Script.running.find_all { |script| script.paused? }.each do |script| 
+       @pause_all_no_unpause << script.name
+    end
+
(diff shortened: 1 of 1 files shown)


rcuhljr:
Technically this isn't thread safe with the lock as is, how terrified would you be if I said add a [semaphore](https://ruby-doc.org/core-2.5.0/Mutex.html) around the lock?

simiydr:
Not super terrified. I wasn't sure if we were concerned with thread safety in the lich context. I'll work on that and resubmit.

rcuhljr:
This is somewhat more of an exception because it's designed explicitly to handle two different scripts calling the same resource at the same time.

rcuhljr:
Some discussing about using Script.current instead of going through the call stack, also difference between determining scripts by name vs instance. Looks good though when Simiy is happy merge this.

simiydr:
Updated to use Script.current and test against the object instead of the name. @rcuhljr Should be good to go.

PR #3603, opened 2019-06-12
Update scripts to use common pause_all and unpause_all methods

simiydr:
With PR #3602 merged, this PR updates the scripts that were using the previous approach to pause/unpause all other scripts to use the new methods.

### DIFF
--- a/almanac.lic
+++ b/almanac.lic
@@ -29,13 +29,13 @@ class Almanac
       return unless training_skill
     end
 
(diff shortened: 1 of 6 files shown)


rcuhljr:
Looks good. We might want to wait a few days on this so we break fewer people with long running lich instances.

simiydr:
@rcuhljr I think we should be good to go? The reboot for Enchanting should have forced everyone not in dev mode to update.

Sarvatt:
yep that's true, everyone was force restarted yesterday so it's safe to pull

PR #3605, opened 2019-06-12
combat-trainer.lic: Add an option to always harness instead of using …

Sarvatt:
…cambrinth for non-buff spells.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -981,7 +981,7 @@ class SpellProcess
     @cambrinth_invoke_exact_amount = settings.cambrinth_invoke_exact_amount
     echo("  @cambrinth_invoke_exact_amount: #{@cambrinth_invoke_exact_amount}") if $debug_mode_ct
 
-    @harness_for_attunement = settings.use_harness_when_arcana_locked
+    @harness_for_attunement = settings.use_harness_when_arcana_locked || settings.combat_trainer_always_harness
     echo(" @harness_for_attunement: #{@harness_for_attunement}") if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -981,7 +981,11 @@ class SpellProcess
     @cambrinth_invoke_exact_amount = settings.cambrinth_invoke_exact_amount
     echo("  @cambrinth_invoke_exact_amount: #{@cambrinth_invoke_exact_amount}") if $debug_mode_ct
 
-    @harness_for_attunement = settings.use_harness_when_arcana_locked
+    if settings.combat_trainer_always_harness
```
rpherbig: 💡 Could this be simplified to `@harness_for_attunement = settings.combat_trainer_always_harness || settings.use_harness_when_arcana_locked`?

profiles/base.yaml
```diff
@@ -288,6 +288,8 @@ safe_room_empaths:
 # Spellcasting (and Khri) settings
 use_harness_when_arcana_locked: false
 combat_trainer_buffs_force_cambrinth: true
+# always skip cambrinth and use harness for non-buffs in combat-trainer.
```
rpherbig: ❓ Is this description correct? If I'm reading this right, as written this won't 'always skip cambrinth and use harness'. It will force `@harness_for_attunement` to `true`, but that still gets passed into `check_to_harness` which will still check Attunement vs Arcana exp.

combat-trainer.lic
Sarvatt: yes it can, I'm an idiot sometimes :D

Issue #3609, opened 2019-06-14
Smelt: need to specify stirring rod due to enchantment's imbue rod


dakre18:
If you do not use the Imbue spell to enchant, then you have to buy an imbuement rod. The Smelt script just gets a rod, and it will fail to run with the imbuement rod in hand.

Changing the script to get a stirring rod instead, will keep that from happening.

PR #3610, opened 2019-06-14
smelt.lic: Specify stirring rod instead of just rod.

Sarvatt:
Fixes: #3609

Double checked that just rod wasn't used in any of the other forging scripts and it's ok.

### DIFF
--- a/smelt.lic
+++ b/smelt.lic
@@ -24,13 +24,13 @@ class Smelt
 
     args = parse_args(arg_definitions)
 
-    get_crafting_item('rod', @bag, @bag_items, @belt)
+    get_crafting_item('stirring rod', @bag, @bag_items, @belt)
     if args.refine
       refine
     else
       stir
     end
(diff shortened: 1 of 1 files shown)


andymanak:
This broke smelt for me:

>tap stirring rod 
>tap stout rod
You tap a serpentine stout stirring rod that you are holding.

rpherbig:
@sarvatt

Sarvatt:
Not sure how to fix that. Enchanting rods will go in a crafting_container: at some point, just getting a rod from it isn't ok. Should I add a new setting for the rod adjective, or just special case this particular rod? I'm thinking the latter, most people will be using crafted rods.

rpherbig:
A setting for the adjective seems straightforward, but it's a weak preference.

Sarvatt:
ideally we'd get the rod adjective and noun from forging_tools: but I don't know how to do that

PR #3614, opened 2019-06-14
mech-lore: Change exp check to Engineering

simiydr:
Update the mech-lore script to use Engineering instead of Mechanical Lore.

### DIFF
--- a/mech-lore.lic
+++ b/mech-lore.lic
@@ -42,7 +42,7 @@ class MechLore
       case bput("braid my #{item}", 'You need to have more', 'Roundtime', 'You need both hands to do that', 'You can\'t braid the .* into your braided', 'is already as long as you can make it', 'You are in no condition')
       when 'Roundtime'
         rt = reget(10, 'Roundtime').last.scan(/\d+/).first.to_i
-        break if rt >= 8 || DRSkill.getxp('Mechanical Lore') >= 30
+        break if rt >= 8 || DRSkill.getxp('Engineering') >= 30
       when 'is already as long as you can make it'
         waitrt?
         break


mech-lore.lic
```diff
@@ -42,7 +42,7 @@ class MechLore
       case bput("braid my #{item}", 'You need to have more', 'Roundtime', 'You need both hands to do that', 'You can\'t braid the .* into your braided', 'is already as long as you can make it', 'You are in no condition')
       when 'Roundtime'
         rt = reget(10, 'Roundtime').last.scan(/\d+/).first.to_i
-        break if rt >= 8 || DRSkill.getxp('Mechanical Lore') >= 30
+        break if rt >= 8 || DRSkill.getxp('Engineering') >= 30
```
Sarvatt: I wish we could rename this to braid.lic but yeah I've mad the same change locally and forgot to push it. It needs changing.

simiydr: We could rename this braid.lic and then just put a stub of a mech-lore.lic script that just starts braid.lic instead.

Sarvatt: that's what I recommended to do to switch to town-training.lic instead of crossing-training.lic but it got pushback :D

Sarvatt: that would be a big +1 from me though.

Sarvatt: this script is essentially useless now, doesn't teach past like 50 engineering

simiydr: I'm using it on a brand new toon and will track how it goes for me. Early ranks are going like mech lore used to.

PR #3619, opened 2019-06-14
Fixed the thief spellmon issue.

jayRyan24:
Emergency fix.


Sarvatt:
This is needed ASAP, they changed how the xml comes through so khri aren't working for anyone.

PR #3620, opened 2019-06-15
shape.lic - magic routine fix for ethereal fissure

BinuDR:
Fixes any spell that has to be cast as  `cast <option>`.

Before it would load up the cast data from a different spell sometimes and try to just cast, now it always uses the data from the correct spell when it casts.


BinuDR:
I need to edit all of the scripts that use the magic routine.  I'll do that in my fork and pull it all in as 1 commit.

PR #3623, opened 2019-06-18
Added enchant script

dr-leustyin:


jandersson:
Youre a champ for getting this out so soon after the release

enchant.lic
```diff
@@ -0,0 +1,176 @@
...
+  include DRC
+  include DRCC
+  include DRCA
+
+  def initialize
+    @settings = get_settings
+    @book_type = "artificing"
+    @bag = @settings.crafting_container
+    @bag_items = @settings.crafting_items_in_container
+    @belt = @settings.enchanting_belt
```
Sarvatt: enchanting_belt: should be added to profiles/base.yaml, not there are any :)

enchant.lic
```diff
@@ -0,0 +1,176 @@
...
+    @secondary_sigils = []
+    
+    @equipment_manager = EquipmentManager.new
+    @equipment_manager.empty_hands
+
+    if @use_own_brazier
+      DRCT.walk_to(@settings.enchanting_room) if @settings.enchanting_room
+    end
+
+    study_recpie
```
Sarvatt: study_recipe

enchant.lic
```diff
@@ -0,0 +1,176 @@
...
+    get_item('fount')
+    bput("wave fount at #{@item} on #{@brazier}",'You slowly wave')
+    
+    imbue
+    
+    get_item('burin')
+    scribe
+  end
+
+  def study_recpie
```
Sarvatt: def study_recipe?

enchant.lic
dr-leustyin: Done!

enchant.lic
```diff
@@ -0,0 +1,181 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#enchant
+=end
+
+custom_require.call(%w[common common-arcana events common-crafting events])
```
rpherbig: 💡 `events` is in here twice. I don't think that will cause any problems, though.

enchant.lic
```diff
@@ -0,0 +1,181 @@
...
+    @bag_items = @settings.crafting_items_in_container
+    @belt = @settings.enchanting_belt
+    arg_definitions = [
+      [
+        { name: 'chapter', regex: /\d+/i, variable: true, description: 'Chapter containing the item.' },
+        { name: 'recipe', display: 'recipe name', regex: /^[A-z\s\-']+$/i, variable: true, description: 'Name of the recipe, wrap in double quotes if this is multiple words.' },
+        { name: 'noun', regex: /\w+/i, variable: true }
+      ]
+    ]
+    #/'  Fix my syntax highlighting
```
rpherbig: ❔

dr-leustyin: Notepad++ highlighting bug.  I needed that line to get syntax highlighting working below line 25.  Meant to remove it

enchant.lic
```diff
@@ -0,0 +1,181 @@
...
+    Flags.add('enchant-meditate', 'The traced sigil pattern blurs before your eyes')
+    Flags.add('enchant-imbue', 'Once finished you sense an imbue spell will be required to continue enchanting')
+    Flags.add('enchant-push', 'You notice many of the scribed sigils are slowly merging back')
+    Flags.add('enchant-sigil', /You need another ([\w ]*)(primary|secondary) sigil to continue the enchanting process/)
+    Flags.add('enchant-complete', 'With the enchanting process completed, you believe it is safe to collect your things once more.')
+    Flags.add('imbue-failed', 'The streams collide, rending the space before you and disrupting the enchantment')
+    
+    @brazier = "brazier"
+    @use_own_brazier=true
+    DRRoom.room_objs.each do | obj |
```
rpherbig: 💡 I wonder if this is any simpler?

```ruby
if DRRoom.room_objs.any? { |obj| obj.include?("enchanter's brazier") }
        @brazier = "enchanter's brazier" 
        @use_own_brazier = false
end
```

enchant.lic
rpherbig: ❌ This needs `common-travel`

enchant.lic
```diff
@@ -0,0 +1,181 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#enchant
+=end
+
+custom_require.call(%w[common common-arcana events common-crafting events])
+
+class Enchant
+  include DRC
+  include DRCC
+  include DRCA
```
rpherbig: ❌ This needs `DRCT`

enchant.lic
```diff
@@ -0,0 +1,181 @@
...
+    @chapter = args.chapter
+    @recipe = args.recipe
+    @primary_sigils = []
+    @secondary_sigils = []
+    
+    @equipment_manager = EquipmentManager.new
+    @equipment_manager.empty_hands
+
+    if @use_own_brazier
+      DRCT.walk_to(@settings.enchanting_room) if @settings.enchanting_room
```
rpherbig: 💡 You can omit the conditional here (or combine it with the one on line 55) - `walk_to` will safely handle `nil`

enchant.lic
```diff
@@ -0,0 +1,181 @@
...
+    case bput("wave rod at #{@item} on #{@brazier}",'Roundtime','The streams collide, rending the space before you and disrupting the enchantment')
+    when 'The streams collide, rending the space before you and disrupting the enchantment'
+      imbue
+    else
+      stow_item('imbue rod') if left_hand.include?('rod')
+      Flags.reset('enchant-imbue')
+    end
+  end
+  
+  def cleanbrazier
```
rpherbig: 💡 What do you think about naming this `clean_brazier` instead?

enchant.lic
Sarvatt: another one that needs addiig to base.yaml, enchanting_room:

enchant.lic
rpherbig: And `equipmanager`

enchant.lic
Sarvatt: no need for DRCT either if its included

dr-leustyin: Is that just convention? It works if it's not set, either in the -setup or base

rpherbig: Nice catch!

Sarvatt:
yeah so its a documented setting

enchant.lic
Sarvatt: yeah so its a documented setting

rpherbig: Yeah, we've gone back on forth on whether or not to include the module when we don't have to.

I was originally of the opinion we should omit it whenever possible so make the code simpler, but I've since changed my mind. I think always including the 'prefix' makes it easier for people new to the ecosystem to read the scripts and get a feel for how the shared modules are organized.

Anyway, mine is a weak preference, so I haven't been pushing either way.

Sarvatt:
how about using get_crafting_item and stow_crafting_item so belts and crafting_containers are respected? would just need to add this like the other crafting scripts

```
  def get_item(name)
    get_crafting_item(name, @bag, @bag_items, @belt)
  end

  def stow_item(name)
    @bag = "other #{@bag}" unless stow_crafting_item(name, @bag, @belt)
  end

PR #3624, opened 2019-06-18
Coordinator: Add easy alchemy training option

jandersson:
Adds an option for `train_alchemy` to the list of actions

Updated profiles as well

### DIFF
--- a/coordinator.lic
+++ b/coordinator.lic
@@ -297,6 +297,7 @@ class Coordinator
       train_outfitting
       train_engineering
       train_forging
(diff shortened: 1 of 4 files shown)


coordinator.lic
```diff
@@ -297,6 +297,7 @@ class Coordinator
       train_outfitting
       train_engineering
       train_forging
+      train_alchemy
```
Sarvatt: this all looks straight forward, just a copy of the other training methods.

PR #3626, opened 2019-06-18
circle-check.lic: Remove mech lore from circle requirements

Sarvatt:
There is no longer any way to train this skill. It can be used if you already have it people need to convert it to a craft to keep circling.

### DIFF
--- a/circlecheck.lic
+++ b/circlecheck.lic
@@ -25,7 +25,6 @@ $Skills = [
   { name: 'Scholarship', type: 'Lore' },
   { name: 'Appraisal', type: 'Lore' },
   { name: 'Tactics', type: 'Lore' },
-  { name: 'Mechanical Lore', type: 'Lore' },
   { name: 'Performance', type: 'Lore' },
   { name: 'Empathy', type: 'Lore' },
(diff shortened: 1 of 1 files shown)


Sarvatt:
yeah they can and I'm on the fence about it. I could see someone coming back not knowing mech can't be learned anymore thinking they're set for 20 circles, then hitting a brick wall when they could have been training a craft. That's a big hypothetical though, I wouldn't mind if this doesn't get accepted at all :D

jandersson:
Yea I can see its kind of an edge case - _most_ people will or have already converted. But it seems like it is still valid to have mech lore at the moment. Up to you or if anyone else has any strong opinion about it, just wanted to point it out. 

It would be funny if it did a "Convert your mech lore, you slob!" message like bescort does.

jandersson:
Nooo re-open it, im not trying to convince you to close this

Sarvatt:
I'm going to fix it up to mention converting mech lore, thats a better idea :D or possibly have an arg to ignore mech

rpherbig:
Can f2p convert mech lore now?

PR #3627, opened 2019-06-19
base-stealing.yaml: Remove non-existant brown feather haversack.

Sarvatt:
this is the cleric shop that doesn't sell haversacks


rpherbig:
Can you check the shop again? This item is not a haversack, it's a `brown feather` _in_ a haversack.

rpherbig:
It's possible the `item_in` attribute should be `in haversack`. That would also be worth double checking.

PR #3629, opened 2019-06-19
combat-trainer: invoke MY egg

Sarvatt:
Ran into a room with eggshells on the ground in headspliters, breaking egg usage.


jandersson:
LGTM

PR #3634, opened 2019-06-21
combat-trainer.lic: lower MY weapon in pickbox in case your weapon ha…

Sarvatt:
…s the same noun as a mob.

This just happened for someone fighting warklin maulers trying to lower a maul.


rpherbig:
❓ I don't think it's quite this simple. See #2927 - I don't know what problem that PR was trying to solve.

Maybe we need to try one way, catch an error, then try the other?

Sarvatt:
oof, good catch

Sarvatt:
so I had the person hitting this problem check it and lower my weapon works just fine, I can only guess that the person who submitted that change had My in their name so it didn't work and I'd rather optimize for things people will hit than someones bad character name choice.

PR #3642, opened 2019-06-25
Remedy updates

dakre18:
Updated remedy.lic to fix multiple bugs, and add functionality to handle special containers. Added match message for herb containers to workorders.lic and common-crafting.lic, and container settings in base.yaml.

### DIFF
--- a/remedy.lic
+++ b/remedy.lic
@@ -12,6 +12,7 @@ class Remedy
   include DRCC
   include DRCI
   include DRCA
+  include DRCT
 
   def initialize
     @settings = get_settings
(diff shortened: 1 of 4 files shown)


remedy.lic
```diff
@@ -12,6 +12,7 @@ class Remedy
   include DRCC
   include DRCI
   include DRCA
+  include DRCT
```
rpherbig: ❌ You'll need to add `common-travel` on line 8

rpherbig:
@Sarvatt @jandersson could you review/test this, or ask around in Slack for a volunteer to do so?

remedy.lic
dakre18: is there a way to add that while it's in a pull request?

Sarvatt: edit it in your branch aka https://github.com/dakre18/dr-scripts/tree/remedy-updates and it’ll update here

dakre18: Done, let me know if i need to do anything else.

PR #3644, opened 2019-06-28
Added imbue waggle set

dr-leustyin:


enchant.lic
```diff
@@ -77,20 +78,35 @@ class Enchant
...
 
   def imbue
-    get_item('imbue rod') unless left_hand.include?('rod')
-    case bput("wave rod at #{@item} on #{@brazier}", 'Roundtime', 'The streams collide, rending the space before you and disrupting the enchantment')
-    when 'The streams collide, rending the space before you and disrupting the enchantment'
-      imbue
+    imbue_data = @settings['waggle_sets']['imbue']['Imbue']
+    if imbue_data
+      crafting_prepare_spell(imbue_data,@settings)
+      castrt = checkcastrt - (imbue_data['cambrinth'].size * 3)
```
rpherbig: ❓ would it be easier to use `DRCA.crafting_cast_spell`?

rpherbig: or maybe something like:
```ruby
imbue_spells = @settings['waggle_sets']['imbue']
cast_spells(imbue_spells, settings, settings.waggle_force_cambrinth)
```

enchant.lic
```diff
@@ -77,20 +78,35 @@ class Enchant
     clean_brazier
     bput("get my #{@item}", 'You get a')
     2.times do
-      bput("put my #{@item} on #{@brazier}", 'You glance down', 'With a flick')
+      case bput("put my #{@item} on #{@brazier}", 'You glance down', 'With a flick', 'You must first clean')
+      when 'You must first clean'
+        echo "Something is still on the brazier.  Cannot continue"
```
rpherbig: ❓ have you run into this case, where `clean_brazier` didn't clean it sufficiently?

enchant.lic
dr-leustyin: common-arcana doesn't have a method to cast at a particular target, which is what's required in this case

enchant.lic
dr-leustyin: I just did in testing because the script wasn't completing correctly and leaving items on the brazier.  We'll need a better mechanism to deal with that edge case.  It's on my list as a future activity

enchant.lic
Sarvatt: true, that is a combat-trainer only thing

enchant.lic
rpherbig: Sounds good

enchant.lic
dr-leustyin: Updated with the override to imbue_data['cast'].  Not sure if this is what you were thinking, but it worked in my test

PR #3646, opened 2019-06-28
remedy.lic - Updated to use hometown settings

dakre18:
Changed buy_item to use hometown settings when determining where to buy ingredients from.


remedy.lic
```diff
@@ -332,24 +334,18 @@ class Remedy
 
   def buy_item(name)
     # walk to bank
-    walk_to('1900')
-    fput("withdraw 4 silver")
+    walk_to(@bank_room)
```
rpherbig: ❓ Isn't there a DRCM helper for this?

dakre18: i may have overlooked it, i'll check

PR #3652, opened 2019-07-03
Added optional container setting for equipment

dr-leustyin:


rpherbig:
Doesn't `wield`/`sheathe` get and store the item in the same container?

dr-leustyin:
Wield only works for weapons.  This particular fix is to allow people that put crafting tools in a crafting_container to be able to repair with Randal/Catrox and make sure the tools get back in the correct container. 

I can also see a use case where someone would want their clown suit armor pieces to go into a specific (not default stow container), although no one has specifically asked for that.

PR #3653, opened 2019-07-03
Add held items to athletics

jayRyan24:
I did not use the equipment manage because I don't always want to list the item I'm using in equipment, my bolts and arrows, for example.

```
held_athletics_items:
  - bolts
  - arrows
```

```
held_athletics_items:
  - club
  - sword
```

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -19,7 +19,7 @@ class Athletics
     have_climbing_rope = get_settings.have_climbing_rope
     @outdoorsmanship_rooms = get_settings.athletics_outdoorsmanship_rooms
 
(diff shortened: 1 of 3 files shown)


jayRyan24:
image

rpherbig:
What happens if something is in your hands when the script starts?

jayRyan24:
Then another script has really screwed up? I can just call `stow_hands`, but it seems like if that's going to be an issue, then you've already messed up somewhere else?

rpherbig:
I guess that problem would have existed before these changes, too.

PR #3664, opened 2019-07-07
combat-trainer.lic - Fix regex, additional skinning knife message, Allow non barbarians to do tactical combos

Hazado:


rpherbig:
The first two commits look fine.

combat-trainer.lic
```diff
@@ -2613,6 +2613,10 @@ class AttackProcess
 
       bput(command, 'Wouldn\'t it be better if you used a melee weapon?', 'Face What?', 'Roundtime', "You aren't close enough to attack", 'It would help if you were closer', 'There is nothing else to face!', 'You must be hidden', 'flying too high for you to attack', 'Bumbling, you slip')
     end
+    
+    if reget(5, 'Wouldn\'t it be better if you used a melee weapon?')
```
jandersson: Not a big deal, but I think doing a `case when` on bput makes more sense here instead of reget

jandersson:
Looks good to me

combat-trainer.lic
```diff
@@ -3581,7 +3585,7 @@ class GameState
       attack_override
     elsif @use_weak_attacks
       brawling? ? 'gouge' : 'jab'
-    elsif @use_barb_combos && !offhand? && !brawling?
+    elsif @use_barb_combos
```
jayRyan24: This really should be renamed as it is now just doing analyze combos and the name doesn't make sense, it will be confusing in the future. This should be `@perform_analyze_combos` probably. 

I am not terribly sure why I included !offhand in the check here, but there was a reason. !brawling is there so that you don't attempt to do a brawling attack while holding a weapon, there's issues there. I do realize you added a reget 5 to determine if this is the case and then to just attack. I did this when I first implemented it, but I was told to change it from the reget to just simply not allowing analyzing with brawling. You will, with this setup you have, a lot of times where you will have an array filled with brawling commands, but then you are using a melee weapon. You are going to be double inputting actions while the array doesn't match. This is going to be super annoying an obnoxious spam. 

How well was this tested? It seems like you just thought you would add it randomly to an already existing PR which didn't have anything to do with the current subject. Maybe next time we should break up the PRs.

combat-trainer.lic
```diff
@@ -3642,16 +3646,20 @@ class GameState
   end
 
   def perform_analyze?(combo_type, expertise_requirement)
-    return false if combo_type != 'flame' && !Flags["ct-#{combo_type}-ready"]
+    return false if combo_type != 'flame' && !Flags["ct-#{combo_type}-ready"] && DRStats.barbarian?
     return false if DRSkill.getrank('Expertise') < expertise_requirement
```
jayRyan24: Does this check work? Does the `getrank` method return something that an equation can be formed on for a non-barb?

combat-trainer.lic
jayRyan24: You can make it:

`elsif @use_barb_combos || @perform_analyze_combos`

Just for past compatibility, and have a warning that barb_combos will be removed, or something.

combat-trainer.lic
jandersson: I checked this on my moonmage, and drskills says his expertise is 0

jayRyan24: OK great, thank you.

combat-trainer.lic
Hazado: I have been using it for three days with my new empathy, it does do double command occasionally when it switches from using a weapon to brawling. After that creature dies, it no more double commands. Other thing I can think of doing if the double commands is not acceptable, is to do a face next and analyze new mob.

Hazado: Yes next time I will break up the pull requests. My first time doing one.

rpherbig: I am a fan of smaller, more focused, PRs.

rpherbig: Which is to say, thank you for your contribution @Hazado !

combat-trainer.lic
Hazado: Yes, made sense to me but the rest of this section also used reget. So I stuck with the current implementation.

combat-trainer.lic
jayRyan24: Yeah, definitely thank you. I would say just don't do Brawling at all. It's 1 skill out of 15, it makes everything more difficult. Swapping targets isn't always the best either. I know when I ran an Empath, I would focus the target my pet was purposely. I just think the best course of action is to not worry about 1 skill when using analyze combos. This alleviates the need for a reget or any other type of action, for 1 skill that doesn't really need to worry about it. Any other thoughts from anyone else?

combat-trainer.lic
jayRyan24: Yup, which is what I did at first, it's weird, because those have multiple lines which come out and sometimes fail to get the proper text. The command you're using should pull it well enough, but definitely if you make the change watch it carefully, it could fail. But if we simply don't do brawling, then we don't need to worry about this at all.

combat-trainer.lic
Hazado: I have no issue with it other than people that only use brawling and want to do combos. Might be better to just add a flag when we get the use a weapon message and have it clear when the creature dies. That way you get the message once and it goes back to attacking normally.

jayRyan24: I thought this, but I think there's an issue with clearing it. I did this and it just gives you back the previous array. Try it, but iirc, that was the issue.

jayRyan24: And yeah, I do realize people only use brawling, but that's just really Empaths. Are you using Ssarek's YAML as a guide? Also, I think that doing the combos really isn't even needed for empaths, you have nothing to do besides Analyze or whatnot, every once in a while. I feel like this is trying to force something that provides only a very marginal impact.

jayRyan24: Again, any other thoughts is good on this subject.

jayRyan24: Also, I will be honest, the optimal way to train any character is through weak attacks, this is pretty well known and anyone not using weak attacks is putting themselves at a large disadvantage. I know you don't when brawling, but yeah, 

Also, as I think of it, when I used brawling on my Empath, I didn't do any type of attacks. Are you using IZ with brawling?

Hazado: new empath, no iz yet. if you use construct mode it attacks normally, you also need to type BRAWL to allow normal brawling. didnt know about the weak attacks training

jayRyan24: I'm not sure we understand each other here. What I mean is, if a person is only training brawling, then that's a very one off case that we don't need to cater an entire system around. I would say, just simply don't use analyze if the user is brawling. This is the easiest solution, and won't affect 99.99% of all players. What I meant by brawling fr the empath, is that when you are brawling, you are training TM with a GS up and you probably aren't attacking, so analyze combos won't even be a thing for you. I can't think of any other situation where a player is only attacking with brawling to train, and if they are, it's probably back training or something like that where they don't want to worry about Tactics.

andymanak: This line broke offhand for me as a barbarian using combos, now if I am training offhand it will analyze and find brawling commands only.

jayRyan24: Yes, I just remembered why I kept in offhand. I will fix this. 

@Hazado doesn't use a lot of the combat system, so his changes didn't actually hit any of the bugs in his code. !offhand is there because if you are training an offhand weapon, you will have nothing in your main hand. This will break, or invoke the "attack" default and not train the offhand, something like that. The issue with the brawling I already stated.

Hazado: Let's revert the offhand and brawling check. Just avoid the issues with offhand and brawling.

PR #3667, opened 2019-07-08
Add researcher script

jandersson:
Adding a stand alone research script. This is a work-in-progress while more research topics are added (and the commit history is cleaned up).

### DIFF
--- a/profiles/Crannach-back_weapons.yaml
+++ b/profiles/Crannach-back_weapons.yaml
@@ -1,11 +1,11 @@
 hunting_info:
-- :zone: zombie_headsplitters
+- :zone: xalas_1
   :duration: 60
   boxes: no
-  before:
-  - mm fm safe_room
-  after:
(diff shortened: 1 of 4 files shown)


rpherbig:
> the commit history is cleaned up

Use the force push, Luke!

jandersson:
Think that bput timeout test needs fixing

khrispyb:
When running this it is using minimum Mana for GAF even though I have it as automana true in waggle for that spell. When researching in crossing training it uses my automana setting. Something I need to do differently to get it to use auto mana scale?

Separately, could we add research plane (astrology) to the list pls?

rpherbig:
@jandersson any updates on this?

jandersson:
I will close this until I feel like i want to focus on developing it or another stand alone research script gets merged.

The branch will still be there if someone wants to cherry-pick something from it.

PR #3673, opened 2019-07-09
afk.lic - simplified stowing feet items

BinuDR:
Now it uses stow feet and will deal with all of the weird items like this:
`a thick book titled ~101 Rational Reasons Why I am Always Right~`


rpherbig:
Neat!

PR #3674, opened 2019-07-10
Reverted some analyze changes, and made setting.

jayRyan24:
I am submitting this to change some of the issues with the PR #3664 .

1.)  Naming convention needs to change to be more accurate.
2.)  If you let the user use brawling analyzes, it will continue over (if you don't finish the array) and break with other weapons. This is an issue which is why I simply just didn't let people use brawling combos. There's no need. It adds a code and processes for a one-off situation where it's beneficial.
3.) If you don't include while !offhand, the user will be training offhand and get errors with the commands not lining up. You will then be issuing "attack", offhand will not go up, and therefore be trained again, this endless cycle of failure.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2614,10 +2614,6 @@ class AttackProcess
       bput(command, 'Wouldn\'t it be better if you used a melee weapon?', 'Face What?', 'Roundtime', "You aren't close enough to attack", 'It would help if you were closer', 'There is nothing else to face!', 'You must be hidden', 'flying too high for you to attack', 'Bumbling, you slip')
     end
     
-    if reget(5, 'Wouldn\'t it be better if you used a melee weapon?')
-      bput('attack', 'Wouldn\'t it be better if you used a melee weapon?', 'Face What?', 'Roundtime', "You aren't close enough to attack", 'It would help if you were closer', 'There is nothing else to face!', 'You must be hidden', 'flying too high for you to attack', 'Bumbling, you slip')
-    end
-
     pause
     waitrt?
 
@@ -3136,6 +3132,9 @@ class GameState
     @use_barb_combos = settings.use_barb_combos
     echo("  @use_barb_combos: #{@use_barb_combos}") if $debug_mode_ct
 
+    @use_analyze_combos = settings.use_analyze_combos
+    echo(" @use_analyze_combos: #{@use_analyze_combos}") if $debug_mode_ct 
+
     @skip_last_kill = settings.skip_last_kill
     echo("  @skip_last_kill: #{@skip_last_kill}") if $debug_mode_ct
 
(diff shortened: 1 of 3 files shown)


combat-trainer.lic
```diff
@@ -3585,7 +3584,11 @@ class GameState
       attack_override
     elsif @use_weak_attacks
       brawling? ? 'gouge' : 'jab'
-    elsif @use_barb_combos
+    # @use_barb_combos needs to be removed given people time to switch
+    # !offhand is there because you can't combo while training offhand weapon without main hand
+    # !brawling is there because you can't use a weapon for brawling commands
+    # Both of these issues will cause problems with the combo array system
+    elsif @use_barb_combos || @use_analyze_combos && !offhand? && !brawling?
```
rpherbig: :x: This needs parens: `(@use_barb_combos || @use_analyze_combos) && ... && ...`


```ruby
a = true
b = false
c = false
d = false
puts a || b && c && d
```

results in `true`

jayRyan24: Thought parens weren't a thing in ruby, lol, wasn't sure, thanks

jayRyan24: Ruby isn't magic? Got it.

profiles/base.yaml
```diff
@@ -186,8 +186,10 @@ dump_item_count: 9
 # as an example, chuno currently runs it at 6 being circle 195
 # defaulting to 25, as 20 for lower skilled barbs was explained to be common
 meditation_pause_timer: 25
-# use the analyze combos to train
+# use the analyze combos to train for the barb - is deprecated!
```
rpherbig: 💡 This looks like a good candidate for a validator

jayRyan24: Yeah, adding that now actually

combat-trainer.lic
```diff
@@ -3136,6 +3132,9 @@ class GameState
     @use_barb_combos = settings.use_barb_combos
     echo("  @use_barb_combos: #{@use_barb_combos}") if $debug_mode_ct
 
+    @use_analyze_combos = settings.use_analyze_combos
```
rpherbig: 💡 You could (maybe) simplify these changes with `@use_analyze_combos = settings.use_analyze_combos || @use_barb_combos`

Then you only have one thing to remove later once `@use_barb_combos` can be removed. And move the "# @use_barb_combos needs to be removed given people time to switch" comment up here.

rpherbig:
Looks good to me. Let me know when you're ready for me to merge it.

combat-trainer.lic
jayRyan24: Yup good call.

jayRyan24:
```
[Private]-DR:Garnn: "any idea why my barbarian isn't analyzing flame to build inner fire?  Need to fix my yaml" [8:51 pm]
```

jayRyan24:
This should be good.

jayRyan24:
One sec @rpherbig , one more change coming.

jayRyan24:
OK this is ready, I knew something looked fishy to me and I fixed it. Dunno why this is the commit message, but it is, lol. C:/Users/Dad

jayRyan24:
oooh, copy pasted my commit message and pressed enter and pushed before I noticed it's a ~... nice.

jayRyan24:
Last commit message: dca3426: 

`$ git commit -m "Removed redundant checks to !offhand and !brawling, as those are resolved in the method which calls this method."`

Hazado:
combat_trainer.lic crashes.

needs an end at line 368

--- Lich: error: combat-trainer:3680: syntax error, unexpected keyword_else, expecting keyword_end

jayRyan24:
@hazado yeah it got fixed a bit before in #3676

PR #3679, opened 2019-07-13
common-arcana - Update to ritual spell method

BinuDR:
It now uses your default_stance instead of being hard coded to 100 80 0.

I also changed the second argument in the method to settings from only ignored_npcs so that I could pull default_stance from it, and for forward compatibility for any other possible additions from settings.

There will be another pull request for combat trainer to update its use of this method.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -110,10 +110,10 @@ module DRCA
     match
   end
 
-  def ritual(spell, ignored_npcs)
-    DRC.retreat(ignored_npcs) unless spell['skip_retreat']
+  def ritual(spell, settings)
+    DRC.retreat(settings.ignored_npcs) unless spell['skip_retreat']
     DRC.release_invisibility
-    DRC.bput('stance set 100 0 80', 'Setting your')
+    DRC.bput("stance set #{settings.default_stance}", 'Setting your')
 
     command = 'prepare'
     command = spell['prep'] if spell['prep']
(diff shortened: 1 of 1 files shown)


BinuDR:
Why is the Travis-CI check returning this error?

```  1) Failure:
TestDRCA#test_ritual_with_skip_retreat [/home/travis/build/rpherbig/dr-scripts/test/test_common-arcana.rb:77]:
Expected script to send messages.
Expected ["stance set 100 0 80", "prepare SPELL 1", "remove my staff", "invoke my staff", "wear my staff", "cast"] to be empty.
```

common-arcana.lic
```diff
@@ -110,10 +110,10 @@ module DRCA
     match
   end
 
-  def ritual(spell, ignored_npcs)
-    DRC.retreat(ignored_npcs) unless spell['skip_retreat']
+  def ritual(spell, settings)
```
rpherbig: ❓ Won't this break for anyone that gets the updated `common-arcana` before we merge in `combat-trainer` changes?

I.e. their `combat-trainer` will be sending in `ignored_npcs` where this function will be expecting `settings` and get `nil` when it tries to pull out `ignored_npcs` or `default_stance`.

I'd rather see a third argument here.

BinuDR: Yes it will and I talked to a few people in slack about it.  They suggested that this was a better method of implementing the change and more inline with the way the rest of the common files implement methods.

BinuDR: There was also talk recently about finding some way of forcing common method updates to propagate to people.

BinuDR: A third argument will also break combat trainer because the number of arguments won't match.  That happened with the last change I made to combat trainer and equipmanager.

https://github.com/rpherbig/dr-scripts/pull/3661
https://github.com/rpherbig/dr-scripts/pull/3662

PR #3682, opened 2019-07-15
carve.lic - added resume function

BinuDR:
You can now resume carving an item!

Syntax:`;carve resume <bone/stone> <noun>`

Please test and if you find any analyze lines that don't get detected by this, let me know.

### DIFF
--- a/carve.lic
+++ b/carve.lic
@@ -23,6 +23,11 @@ class Carve
         { name: 'material', regex: /\w+/i, variable: true, description: 'Type of stone/bone to use.' },
         { name: 'type', options: %w[stack rock stone pebble boulder deed], description: 'What material noun you are using.' },
         { name: 'noun', regex: /\w+/i, variable: true }
+      ],
+      [
+        { name: 'resume', regex: /resume/i },
+        { name: 'type', options: %w[bone stone], description: 'Which material type the item is.' },
+        { name: 'noun', regex: /\w+/i, variable: true, description: 'Noun of item to resume.' }
       ]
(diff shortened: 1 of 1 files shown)


carve.lic
```diff
@@ -23,6 +23,11 @@ class Carve
         { name: 'material', regex: /\w+/i, variable: true, description: 'Type of stone/bone to use.' },
         { name: 'type', options: %w[stack rock stone pebble boulder deed], description: 'What material noun you are using.' },
         { name: 'noun', regex: /\w+/i, variable: true }
+      ],
+      [
+        { name: 'resume', regex: /resume/i },
+        { name: 'type', options: %w[bone stone], description: 'Which material type the item is.' },
```
rpherbig: 💡 I'm not a fan of overloading the meaning of 'type', but I don't have a better name to suggest.

BinuDR: I thought about using the yaml option `carving_workorder_material_type:` but there is no guarantee that someone is always going to be trying to resume the same material type as that.

Material was the other name I was going to use but it was also already used.

The other option I was looking at was to have the script determine what material the item is.  It  would require either 2 analyzes or I would need to change the whole section in resume to use a reget where I could pull the data.  It's messier either way, and it also ends up requiring duplicate code to set the main tool.

Currently it gets set during the initialize portion and I was able to work the type arg into it.

```    @main_tool = if @type == 'stack' || @type == 'bone'
                   'saw'
                 else
                   'chisel'
                 end```

PR #3686, opened 2019-07-16
Combat-Trainer.lic - Empath hunting improvements

Hazado:
Basically, I made changes so when you hunt undead/constructs that can also have live npcs in the room, you will try to manipulate the live npcs and only attack the ones you want to attack.

Added a flag for when your link breaks with manipulate
changed manipulate construct flagging to also work for undead
changed manipulate to work for entire room instead of the filtered npcs
Changed engagement check to try to engage or face when you first start the script and when something dies

May need more check for the various shock inducing attack to prevent empaths from getting shocked accidentally


combat-trainer.lic
```diff
@@ -1853,12 +1853,14 @@ class ManipulateProcess
...
 
+    Flags.add('ct-manip-broke', 'You feel your link with .* suddenly break', 'You feel your link suddenly break', 'You feel your link with .* weakening with distance.')
+
     @last_manip = Time.now - 200
   end
 
   def execute(game_state)
     return if game_state.danger || @threshold.nil? || game_state.construct_mode?
-    @filtered_npcs = game_state.npcs
+    @filtered_npcs = DRRoom.npcs#game_state.npcs
```
rpherbig: ❓ This doesn't look right to me. Why was this changed?

rpherbig:
Would it be possible for you to update your fork and use a fresh branch so there isn't so much noise in the commit history?

I would also prefer one PR per logical change. It keeps the PRs focused, easy to review, and they can be merged independently of each other (i.e. a problem in one doesn't hold up the rest).

combat-trainer.lic
```diff
@@ -2645,6 +2658,7 @@ class AttackProcess
   end
 
   def attack_thrown(game_state)
+    return if DRRoom.dead_npcs.size > 0
```
rpherbig: ❓ Won't this abort throwing if there is a dead NPC and a living NPC? Why was this change made?

combat-trainer.lic
Hazado: if your hunting constructs or undead in a room that has live npcs, you need to put the live npcs in the ignore npc list. Without this change you won't manipulate the live npcs.

combat-trainer.lic
Hazado: Yes, helps to prevent empaths from attacking a live npc if their current target dies before the loot check.

Hazado:
> 
> 
> Would it be possible for you to update your fork and use a fresh branch so there isn't so much noise in the commit history?
> 
> I would also prefer one PR per logical change. It keeps the PRs focused, easy to review, and they can be merged independently of each other (i.e. a problem in one doesn't hold up the rest).

Yes, Im still learning how to use github.
The sorter.lic commit was done in the master before I learned about forks.
Going forward all my changes will be done in forks.

combat-trainer.lic
Sarvatt: maybe make the new return ```return if DRRoom.dead_npcs.size > 0 && DRStats.empath?``` to limit any potential breakage to empaths.

Hazado: Might be better to call the check face instead if dead Npc and empath

Hazado:
resubmitting

PR #3687, opened 2019-07-18
Sorter.lic

Hazado:
Resubmitting for better commit history.


common.lic
```diff
@@ -220,7 +220,13 @@ module DRC
...
...r-finished|mottled|multiple woods|naphtha|oak|oblong sanguine|one|onyx buttons|opposing images|overlapping|pale cerulean|pallid links|pastel-hued|pins|pitted (?:black iron|steel)|plush velvet|polished (?:bronze|hemlock|steel)|raccoon tails|ram's horns|rat pelts|raw|red and blue|rich (?:purple|golden)|riveted bindings|roughened|rowan|sanguine thornweave|scattered star|scorch marks|sculpted|shadows|shark cartilage|shifting (?:celadon|shades)|shipboard|(?:braided |cobalt |deep black |desert-tan |dusky red Taisidon |ebony |exquisite spider|fine leaf-green |flowing night|glimmering ebony |heavy |marigold |pale gold marquisette and virid |rich copper |spiral-braided |steel|unadorned black Musparan )?silk(?:cress)?|(?:coiled |shimmering )?silver(?:steel| and |y)?|sirese blue spun glitter|six crossed|slender|small bones|smoothly interlocking|snow leopard|soft brushed|somber black|sprawled|sun-bleached|steel links|stones|strips of|sunny yellow|teardrop plates|telothian|the|tiny (?:golden|indurium|scales|skull)|tightly braided|tomiek|torn|twists|two|undyed|vibrant multicolored|viscous|waves of|weighted|well-cured|white ironwood|windstorm gossamer|wintry faeweave|woven diamondwood))\b.*/, '')
```
rpherbig: Was this going to be moved to `base-sorting.yaml`?

Hazado: I tried to move the large block with no success. Yaml would fail to read it.

rpherbig:
Looks ok to me

ping @Sarvatt @jandersson @jayRyan24  for one last chance to jump in with comments

jandersson:
Go for it

common.lic
jayRyan24: Thank you Dragonrealms, for your great flavor!

data/base-sorting.yaml
```diff
@@ -0,0 +1,483 @@
+cambrinth:
+- .*cambrinth.*
```
jayRyan24: Some items can use cambrinth as a decoration. This is minimal concern, but it's another one of those "gotchas" by the flavor text team... :D

jayRyan24:
@rpherbig @jandersson @Sarvatt @Hazado Works well, and I really appreciate the effort on this one. One of the issues I have though is that it will `sort` a simple `look charName`, which can be really obnoxious. Do we want to ignore that, or should that be adjusted?

jayRyan24:
Also, what will break due to the failed checks?

Hazado:
> @rpherbig @jandersson @Sarvatt @Hazado Works well, and I really appreciate the effort on this one. One of the issues I have though is that it will `sort` a simple `look charName`, which can be really obnoxious. Do we want to ignore that, or should that be adjusted?
> 
> I haven't tested the recent ones, so it could already have been fixed!


Look <person>
Look item

> @rpherbig @jandersson @Sarvatt @Hazado Works well, and I really appreciate the effort on this one. One of the issues I have though is that it will `sort` a simple `look charName`, which can be really obnoxious. Do we want to ignore that, or should that be adjusted?
> 
> I haven't tested the recent ones, so it could already have been fixed!

Toggle options to sort
Mute old inventory
Inventory
Look <person>
Look item

data/base-sorting.yaml
Hazado: Yeah, but what can we do? It's either remove cambrinth category or live with the gotchas

Hazado: The look charname can be set to false by default if you want.

jayRyan24: Yeah, my comment was more of a "you're going to see this come up" type of thing. I think sorting it is just fine, otherwise you'd need to then parse the string for type of cambrinth, would be messy.

jayRyan24:
Oh god I'm blind as all fucking hell. My bad. Thanks.

jayRyan24:
I've been running the changes for a few days without hiccup. The sorter works great with it too, don't think there's a reason to not push it through and let people get in there and break stuff!

common.lic
KatoakDR: Bless you all for writing this method. This and `get_noun` are quite handy and making my life much easier at the moment.

PR #3688, opened 2019-07-18
combat-trainer.lic - Empath hunting improvements

Hazado:
Resubmitted for better commit history

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -989,7 +989,7 @@ class SpellProcess
 
     @ignored_npcs = settings.ignored_npcs
     echo(" @ignored_npcs: #{@ignored_npcs}") if $debug_mode_ct
-    
+
     @hide_type = settings.hide_type
     echo("  @hide_type: #{@hide_type}") if $debug_mode_ct
 
@@ -1607,7 +1607,7 @@ class SpellProcess
     end
 
     game_state.casting_regalia = data['abbrev'].casecmp('REGAL').zero?
-    
+
     check_face(game_state)
     prepare?(data['abbrev'], data['mana'], data['symbiosis'], command)
     game_state.casting = true
@@ -1629,7 +1629,7 @@ class SpellProcess
     @use_auto_mana = data['use_auto_mana']
     game_state.action_taken if (@skill == 'Targeted Magic' || @prep == 'target') && game_state.weapon_skill == 'Targeted Magic'
   end
-  
+
   def check_face(game_state)
     return unless Flags['ct-face-what'] || (DRRoom.dead_npcs.size > 0 && DRStats.empath?)
     if (game_state.npcs.length > 0)
@@ -2644,7 +2644,7 @@ class AttackProcess
 
       bput(command, 'Wouldn\'t it be better if you used a melee weapon?', 'Face What?', 'Roundtime', "You aren't close enough to attack", 'It would help if you were closer', 'There is nothing else to face!', 'You must be hidden', 'flying too high for you to attack', 'Bumbling, you slip')
     end
-    
+
     pause
     waitrt?
 
@@ -3135,7 +3135,7 @@ class GameState
 
     @ignored_npcs = settings.ignored_npcs
     echo("  @ignored_npcs: #{@ignored_npcs}") if $debug_mode_ct
-    
+
     @manipulate_npcs = settings.manipulate_npcs
     echo(" @manipulate_npcs: #{@manipulate_npcs}") if $debug_mode_ct
 
@@ -3567,7 +3567,7 @@ class GameState
   def npcs
     DRRoom.npcs - @ignored_npcs
   end
-  
+
   def manip_npcs
     DRRoom.npcs - (@ignored_npcs - @manipulate_npcs)
   end
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -989,6 +989,9 @@ class SpellProcess
 
     @ignored_npcs = settings.ignored_npcs
     echo(" @ignored_npcs: #{@ignored_npcs}") if $debug_mode_ct
+    
+    @manipulate_npcs = settings.manipulate_npcs
+    echo(" @manipulate_npcs: #{@manipulate_npcs}") if $debug_mode_ct
 
     @hide_type = settings.hide_type
     echo("  @hide_type: #{@hide_type}") if $debug_mode_ct
@@ -1262,7 +1265,7 @@ class SpellProcess
     end
 
     snapshot = DRSkill.getxp(@skill)
-    check_face(game_state) if DRRoom.dead_npcs.size > 0 && DRStats.empath?
+    check_face(game_state)
     success = cast?(@custom_cast, @symbiosis, @before, @after)
     if @symbiosis && @use_auto_mana
       if !success
@@ -1630,7 +1633,7 @@ class SpellProcess
   end
   
   def check_face(game_state)
-    return unless Flags['ct-face-what'] || DRRoom.dead_npcs.size > 0
+    return unless Flags['ct-face-what'] || (DRRoom.dead_npcs.size > 0 && DRStats.empath?)
     if (game_state.npcs.length > 0)
         case bput("eng #{game_state.npcs.first}", 'You will have to retreat from your current melee first', 'You are already at melee', 'You begin to advance', 'You spin around to face', 'is already quite dead')
         when /You will have to retreat from your current melee first/
@@ -1870,14 +1873,14 @@ class ManipulateProcess
     @manip_to_train = settings.manipulate_only_to_train
     echo("  @manip_to_train: #{@manip_to_train}") if $debug_mode_ct
 
-    Flags.add('ct-manip-broke', 'You feel your link with .* suddenly break', 'You feel your link suddenly break', 'You feel your link with .* weakening with distance.')
+    Flags.add('ct-manip-broke', 'You feel your link.*(?:suddenly break|weakening with distance)')
 
     @last_manip = Time.now - 200
   end
 
   def execute(game_state)
     return if game_state.danger || @threshold.nil? || game_state.construct_mode?
-    @filtered_npcs = DRRoom.npcs
+    @filtered_npcs = game_state.manip_npcs
     manipulate(game_state) if should_manipulate?
   end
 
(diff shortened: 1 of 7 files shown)


combat-trainer.lic
```diff
@@ -1262,6 +1262,7 @@ class SpellProcess
     end
 
     snapshot = DRSkill.getxp(@skill)
+    check_face(game_state) if DRRoom.dead_npcs.size > 0 && DRStats.empath?
```
rpherbig: 💡 You have the `DRRoom.dead_npcs.size > 0` check here and inside the function. I suggest consolidating all of the conditionals inside the function.

Hazado: That one is inside the spells section. Wasnt able to figure out how to call check face from combat class inside spell class.
I will change it so the check is inside the function instead though.

combat-trainer.lic
```diff
@@ -1627,6 +1628,22 @@ class SpellProcess
     @use_auto_mana = data['use_auto_mana']
     game_state.action_taken if (@skill == 'Targeted Magic' || @prep == 'target') && game_state.weapon_skill == 'Targeted Magic'
   end
+  
+  def check_face(game_state)
+    return unless Flags['ct-face-what'] || DRRoom.dead_npcs.size > 0
+    if (game_state.npcs.length > 0)
+        case bput("eng #{game_state.npcs.first}", 'You will have to retreat from your current melee first', 'You are already at melee', 'You begin to advance', 'You spin around to face', 'is already quite dead')
```
jayRyan24: The internals of this if statement are tabbed instead of 2 spaced for alignment.

combat-trainer.lic
```diff
@@ -1853,12 +1870,14 @@ class ManipulateProcess
...
 
+    Flags.add('ct-manip-broke', 'You feel your link with .* suddenly break', 'You feel your link suddenly break', 'You feel your link with .* weakening with distance.')
+
     @last_manip = Time.now - 200
   end
 
   def execute(game_state)
     return if game_state.danger || @threshold.nil? || game_state.construct_mode?
-    @filtered_npcs = game_state.npcs
+    @filtered_npcs = DRRoom.npcs
```
jayRyan24: Why is this variable named `filtered_npcs`? Is the `game_state` keeping track of things differently than the simple `DRRoom.npcs`? Maybe there are some `ingored_npcs` or something we're then not accounting for? Something to consider.

rcuhljr: Filtered npcs is the list with ignored npcs removed.

rcuhljr: Filtered npcs is the list with ignored npcs removed.

Hazado: game_state.npcs is drroom.npcs - ignored_npcs
ignored_npcs are npcs we dont wanr empaths making attacks on

only other way i can think to do it is make a list of construct and undead npcs and filter the room npcs through that

combat-trainer.lic
```diff
@@ -1853,12 +1870,14 @@ class ManipulateProcess
     @manip_to_train = settings.manipulate_only_to_train
     echo("  @manip_to_train: #{@manip_to_train}") if $debug_mode_ct
 
+    Flags.add('ct-manip-broke', 'You feel your link with .* suddenly break', 'You feel your link suddenly break', 'You feel your link with .* weakening with distance.')
```
jayRyan24: These returns could probably be condensed into a regex. `/^You feel your link (with)?.* (suddenly|weakening)?/` something like that. I'm not the best regexer.

https://rubular.com/r/rDPAFupLyToSJl

combat-trainer.lic
```diff
@@ -1883,8 +1902,8 @@ class ManipulateProcess
       index = npc_occurrences[npc]
       ordinal_string = $ORDINALS[index]
 
-      case bput("manipulate friendship #{ordinal_string} #{npc}", 'You\'re already manipulating', 'beyond your ken', 'You attempt to empathically manipulate', 'You strain', 'does not seem to have a life essence')
-      when 'does not seem to have a life essence' then
+      case bput("manipulate friendship #{ordinal_string} #{npc}", 'You\'re already manipulating', 'beyond your ken', 'You strain', 'does not seem to have a life essence', 'is definitely attracted to you', 'Roundtime', 'You have certainly attracted', 'Manipulate what\?')
+      when 'does not seem to have a life essence', 'is definitely attracted to you', 'You have certainly attracted' then
```
jayRyan24: The old code had this, but I don't think the `then` is required here.

combat-trainer.lic
jayRyan24: Is it the case that you don't mind manipulating the ignored_npcs? I know I use those for player-created entities like familiars and watchsouls, or just NPC's that wander some hunting areas.

combat-trainer.lic
Hazado: Will work on it

combat-trainer.lic
Hazado: Maybe better implementation would be to have a second list for npcs we want to manipulate? So manipulate npcs list will add npcs back in to filtered list

Hazado: So far in my short hunting experience, every area I've tried has had npcs I can't attack but can manipulate. Haven't had the issue where I tried to manipulate a familiar etc. I would think we just need to capture the message for that and add it to the construct check. You will try to manipulate them once a hunt then it will ignore them again

jayRyan24:
@Hazado Who are you in game again? And are you on right now?

combat-trainer.lic
Hiinky: You also can't manipulate undead.  Familiars are automatically in the ignore_npc: list in base.  If not then you will try to manipulate/app them.  I can get you any messaging you want to see.

jayRyan24: Don't believe undead are in the ignore_npc list, not manipulating undead is on the player. The familiars being in the base ignore list is nice then, it wasn't before, thanks for that.

jayRyan24: Just confused at what the benefit is of using the DRRoom.npcs over the game_state.npcs? You will encounter familiar problems eventually, once WM's join you, and a lot of HLC empaths group, so that would be a consistent problem.

Hiinky: For whatever its worth, empath warriors, watchsouls, ranger spawn, and familiars are definitely annoying.  And you don't always notice right away they're messing with you.  I don't use the various pets on my necro but I imagine they would also be on that list though more of an edge case as empaths aren't often hunting with necro.

Hazado:
> @Hazado Who are you in game again? And are you on right now?

Helf and not on right now

combat-trainer.lic
Hazado: fixed

combat-trainer.lic
Hazado: Added a new line to base.yaml for adding back in NPCs that you want to manipulate.
going to test it now on my empath.

combat-trainer.lic
Hazado: done.

combat-trainer.lic
Hazado: removed the then
testing now

combat-trainer.lic
jayRyan24: Sorry, can you explain the reason behind just not using game_state.npcs? You're adding a second list now?

Hazado:
Did an hour of hunting reavers and wind hounds without getting any shock

jayRyan24:
Oh, so this is just so you can hunt a non-manip creature with a manip one?

combat-trainer.lic
```diff
@@ -3104,6 +3146,9 @@ class GameState
 
     @ignored_npcs = settings.ignored_npcs
     echo("  @ignored_npcs: #{@ignored_npcs}") if $debug_mode_ct
+    
+    @manipulate_npcs = settings.manipulate_npcs
```
jayRyan24: Is there a need for this to also be here if it's included in SpellProcess?

Hazado:
The second list is so we keep the ignore npcs and have npcs we can manip but not attack.

Yes basically it's changes so you can use manip and hunt non manip npcs at the same time.

combat-trainer.lic
```diff
@@ -989,6 +989,9 @@ class SpellProcess
 
     @ignored_npcs = settings.ignored_npcs
     echo(" @ignored_npcs: #{@ignored_npcs}") if $debug_mode_ct
+    
+    @manipulate_npcs = settings.manipulate_npcs
```
jayRyan24: This be removed it's unused and being supported through GameState.

combat-trainer.lic
```diff
@@ -2584,14 +2609,24 @@ class AttackProcess
...
+          fput("face #{game_state.npcs.last}")
+        end
+      end
+      Flags.reset('ct-face-what')
+    end
   end
 
   def attack_melee(charged_maneuver, game_state)
     waitrt?
+    return if DRRoom.dead_npcs.size > 0 && DRStats.empath?
```
jayRyan24: What is the purpose of these returns throughout the actions? This seems like an issue that would only arise with the Empath attacking with weapons while also manipulating? When does this actually come up? I think a lot of these changes are being put through to support hunting manipulatable critters with non. This is a fairly uncommon issue I found only once (and that's where you mentioned with the reavers and hounds). Not sure it's worth always checking this for a one-off?  

Also, this could be `return if DRRoom.dead_npcs.empty? && DRStats.empath?` which seems clearer to me when reading it using the method, but either way is logically the same.

jayRyan24: FWIW, I was going to do this when, but then realized it only occurs once while hunting, so never seemed an issue worth fixing. I don't think this is a bad fix though and as long as it doesn't add terrible overhead, is probably not a big deal.

jayRyan24:
@rpherbig @Hazado I would probably have moved the CheckFace changes to another PR as it would be easier to track, but no worries.

combat-trainer.lic
Hazado: The returns are to help prevent shock when hunting in undead/constructs with live npcs around.
If a creature dies and you try do an attack before getting to loot the dead creatures, you turn to face the closest creature, which is sometimes a live npc. And you get stunned and a bit of shock as an empath because of it.

combat-trainer.lic
Hazado: Second list allows us to keep ignoring npcs we dont want to attack, and manipulate npcs we do.

combat-trainer.lic
jayRyan24: Yeah, but this is a check every single action for every single characters based on something that really only occurs once in the entire ladder, and you really only hunt these two critters for a short amount of time. Maybe a day? I just think I need more feedback from others, I only experienced this once, in the same place you did, from circle 0-165, and that's when I sold.

jayRyan24:
FWIW, I checked with another few HLC empaths, and this does only occur here and one other ~100 rank place.

combat-trainer.lic
jayRyan24: OK, this needs to be handled differently, you can probably add the return of this to a single place within AttackProcess rather than every method inside. This would be a better solution.

profiles/base.yaml
```diff
@@ -91,6 +91,8 @@ combat_analyze_retry_count: 3
 manipulate_threshold:
 # manipulate only to train empathy, or always manipulate when above the threshold.
 manipulate_only_to_train: false
+# NPCs added here will be manipulated even if they are included on the ignore_npcs list
+manipulate_npcs:
```
jandersson, line 95: If this is an array, it should be in base-empty as well as `manipulate_npcs:  []`

PR #3691, opened 2019-07-18
Adds more carving recipes from apprentice book

Hazado:


Sarvatt:
are you absolutely sure the ones you set to workorders: true work in workorders?

Sarvatt:
usually the items that need parts that haven't been tested before are broken at the start

Hazado:
actually, base-crafting.yaml needs updating to correct the volume size of alabaster

Hazado:
I didn't add items that need parts. The ones I added and have set to workorders: true are ones that use volume 3 stone and down.

Sarvatt:
medallions don't need parts? the wiki doesnt list the ingredients but for other crafts neck things do is why i was asking

Hazado:
Nope

`        -=   Chapter 4, Page 11: Instructions for crafting a stone medallion    =-

  A stone medallion is a craftable item in the Engineering society under the Carving crafting discipline.  This is considered to be a simple piece to make, though knowledge of the Decorative Stone Design technique will be beneficial to the crafter.

  This item is listed as a "finished stone body jewelry" ingredient type and is created using a chisel set.  A crafter may also find it helpful to have a rasp, a riffler set and a jar of polish on hand.

 A list of ingredients is provided:

   (1) raw stone material (stone)`

Sarvatt:
oh cool, thanks for double checking!

PR #3694, opened 2019-07-18
setupaliases.lic - change to as alias

Hazado:
change sets it so that it accepts, then stows whatever hand the item went to instead

Lets people accept and stow when they have something in their right hand already

### DIFF
--- a/setupaliases.lic
+++ b/setupaliases.lic
@@ -4,7 +4,7 @@
 
 [
   ['cc', "#{$clean_lich_char}circlecheck"],
-  ['as', 'accept \r stow \?'],
+  ['as', "#{$clean_lich_char}e GameObj.right_hand.name == 'Empty' ? (fput('accept');fput('stow right')) : (fput('accept');fput('stow left'))"],
   ['gs', 'get \? \r stow \?'],
 
   # toggles crossing-training on or off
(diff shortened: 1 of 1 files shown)


setupaliases.lic
```diff
@@ -4,7 +4,7 @@
 
 [
   ['cc', "#{$clean_lich_char}circlecheck"],
-  ['as', 'accept \r stow \?'],
+  ['as', "#{$clean_lich_char}e GameObj.right_hand.name == 'Empty' ? (fput('accept');fput('stow right')) : (fput('accept');fput('stow left'))"],
```
jayRyan24: There are some issues with GameObj and DR. We have these two methods for cleaning the issues away:
```
  def fix_dr_bullshit(string)
    return string if string.split.length <= 2
    string =~ /(\w+) .* (\w+)/
    "#{Regexp.last_match(1)} #{Regexp.last_match(2)}"
  end

  def left_hand
    GameObj.left_hand.name == 'Empty' ? nil : fix_dr_bullshit(GameObj.left_hand.name)
  end

  def right_hand
    GameObj.right_hand.name == 'Empty' ? nil : fix_dr_bullshit(GameObj.right_hand.name)
  end
```
I believe we should use the DRC.right_hand, etc, for this for consistency at a minimum. (Though I see places where this isn't followed.) Good addition/fix.

Hazado: Setup to use DRC.right_hand now.

PR #3700, opened 2019-07-20
test_common-arcana - Fix for travis ci check failure

BinuDR:
This should fix the issue with Travis CI's failing tests.  This code is from @rcuhljr so if it needs troubleshooting we're going to need his assistance.  I'll be here to watch and learn though.


BinuDR:
I need to add this to the other PR for common-arcana.

PR #3704, opened 2019-07-24
Look at all shelves, racks, etc in vault

Hazado:
Code to change current look on first container to look on all containers in vault

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -815,16 +815,14 @@ while line = script.gets
         .select { |_town, titles| titles.include?(checkroom) }
         .each do |_town, _titles|
           contents = Regexp.last_match(1)
-          if line =~ /a .+?\s?(?:(\w+ (?:shelf|rack|box|case|tree))) with some stuff (on|in) it/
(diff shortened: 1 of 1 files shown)


drinfomon.lic
```diff
@@ -815,11 +815,13 @@ while line = script.gets
         .select { |_town, titles| titles.include?(checkroom) }
         .each do |_town, _titles|
           contents = Regexp.last_match(1)
-          if line =~ /a (\w+ shelf|rack|box|case|tree) with some stuff (on|in) it/
-            fput 'look ' + Regexp.last_match(2) + ' ' + Regexp.last_match(1)
-            if get =~ /(?:On|In) the \w+ (?:shelf|rack|box|case|tree) you see(.*)\./
-              contents += Regexp.last_match(1)
-            end
+          if line =~ /a .+?\s?(?:(\w+ (?:shelf|rack|box|case|tree))) with some stuff (on|in) it/
+            line.scan(/a .+?\s?(?:(\w+ (?:shelf|rack|box|case|tree))) with some stuff (on|in) it/).each{|noun,location|
```
rpherbig: 💡 I think you no longer need the `if`, since `scan` will return an empty array if nothing matches.

Hazado:
Should be working now

PR #3711, opened 2019-07-27
make looting gleaming chest more specific

Hazado:


Sarvatt:
straightforward fix, looks good.

PR #3715, opened 2019-07-28
Common-arcana.lic - location casting

Hazado:
adding a location variable allows the ability to cast heal wounds and heal scars

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -344,7 +344,7 @@ module DRCA
     aura
   end
 
-  def cast_spells(spells, settings, force_cambrinth = false, location = nil)
+  def cast_spells(spells, settings, force_cambrinth = false)
     infuse_om(!settings.osrel_no_harness, settings.osrel_amount)
     spells.each do |name, data|
       next if DRSpells.active_spells[name] && (data['recast'].nil? || DRSpells.active_spells[name].to_i > data['recast'])
@@ -353,7 +353,7 @@ module DRCA
         pause 15
       end
 
-      cast_spell(data, settings, force_cambrinth, location)
+      cast_spell(data, settings, force_cambrinth)
     end
   end
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -344,7 +344,7 @@ module DRCA
     aura
   end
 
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -344,7 +344,7 @@ module DRCA
     aura
   end
 
-  def cast_spells(spells, settings, force_cambrinth = false)
+  def cast_spells(spells, settings, force_cambrinth = false, location = nil)
```
rpherbig: ❓ This will use the location for every spell in `spells`. Is that desirable? Location feels to me like it would/should be specific to a single spell.

Hazado: Not sure how many spells need something other than "cast".
This could be used for any of them. "cast player", "cast left arm internal reverse", etc

Hazado: https://elanthipedia.play.net/Cast_command
There are quite a few different casting options.

rpherbig: This function is used to cast multiple spells. As written, this will use location for all of them.

Is that intended?

common-arcana.lic
```diff
@@ -426,7 +426,7 @@ module DRCA
     pause 0.5 while UserVars.moons.empty?
   end
 
-  def cast_spell(data, settings, force_cambrinth = false)
+  def cast_spell(data, settings, force_cambrinth = false, location = nil)
```
rpherbig: ❓ How do you plan to use this? I ask because the consumer can already set `data['cast']` before calling `cast_spell`, which doesn't require a change to `common-arcana`.

BinuDR:
You can edit the data['cast'] variable before sending data to this method.
Something like this:
body_parts = ['head', 'neck', 'chest', 'abdomen', 'back']
body_parts.each{ |bp| data['cast'] = "cast #{bp}"; cast_spell(data, settings) }

common-arcana.lic
Hazado: working on updating healme.lic to use waggle

Hazado:
That might be a better implementation on healme.lic than modifying common-arcana.lic

common-arcana.lic
Hazado: your right, we can close this. Will implement that into healme.lic instead.

PR #3716, opened 2019-07-28
healme.lic - use waggle for hw and hs

Hazado:


BinuDR:
This already uses a waggle set.
```waggle_sets:
  healing:
    Heal Wounds:
      mana: 5
      cambrinth:
      - 5
    Heal Scars:
      mana: 8
      cambrinth:
      - 8```

Hazado:
No it uses empath_healing not waggle_sets for heal wounds and heal scars

BinuDR:
Really?  I wonder which script I have that waggle set made for.

BinuDR:
```    using_waggle = false
    spells = if @settings.waggle_sets['healing']
               using_waggle = true
               @settings.waggle_sets['healing']
             else
               @settings.empath_healing
             end

Hazado:
Yes thats for regenerate and heal. For heal wounds and heal scars it doesnt use it

BinuDR:
Ah, yeah I have both ways setup in my yaml.  That must be for a different script then.  My mistake.

PR #3719, opened 2019-07-28
Workorder.lic improvements

Hazado:
Add @retain_crafting_materials to shaping, sewing, and knitting
Changes to stone carving to allow use of left over pieces after carving

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -387,13 +387,13 @@ class WorkOrders
   def sew_items(info, recipe, quantity)
     ensure_copper_on_hand(@cash_on_hand || 5000, @settings)
 
-    existing = if bput("get cloth from my #{@bag}", 'What were', 'You get') == 'What were'
+    existing = if bput("get #{info['sew-stock-name']} cloth from my #{@bag}", 'What were', 'You get') == 'What were'
                  0
                else
-                 bput('count my cloth', 'You count out \d+ yards').scan(/\d+/).first.to_i
+                 bput("count my #{info['sew-stock-name']} cloth", 'You count out \d+ yards').scan(/\d+/).first.to_i
                end
     stock_needed = ((quantity * recipe['volume'] - existing) / 10.0).ceil
-    order_fabric(info['stock-room'], stock_needed, info['sew-stock-number'], 'cloth')
+    order_fabric(info['stock-room'], stock_needed, info['sew-stock-number'], "#{info['sew-stock-name']} cloth")
 
     order_parts(recipe['part'], quantity) if recipe['part']
 
--- a/workorders.lic
+++ b/workorders.lic
@@ -221,7 +221,8 @@ class WorkOrders
   def carve_items(info, item, quantity)
     ensure_copper_on_hand(@cash_on_hand || 5000, @settings)
     recipe, items_per_stock, spare_stock, scrap = find_recipe(info, item, quantity)
-
+    rock_noun = %w[deed pebble stone rock rock boulder]
+    rock_count = 0
     bone_carving = recipe['material'] == 'bone'
 
     if bone_carving
@@ -233,7 +234,7 @@ class WorkOrders
     end
 
     if search?('surface polish')
-      fput('get my surface polish')
+      bput('get my surface polish', 'You get')
       /(\d+)/ =~ bput('count my polish', 'The surface polish has \d+ uses remaining')
       if Regexp.last_match(1).to_i < 3
         stow_hands
@@ -246,6 +247,9 @@ class WorkOrders
     stow_hands
 
     quantity.times do |count|
+      if (info['stock-volume'] - (count * recipe['volume'])) > 0
+        rock_count = info['stock-volume'] - (count * recipe['volume'])
+      end
       if items_per_stock.zero? || (count % items_per_stock).zero?
         if count > 0
           go_door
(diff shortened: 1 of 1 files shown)


workorders.lic
```diff
@@ -221,7 +221,8 @@ class WorkOrders
   def carve_items(info, item, quantity)
     ensure_copper_on_hand(@cash_on_hand || 5000, @settings)
     recipe, items_per_stock, spare_stock, scrap = find_recipe(info, item, quantity)
-
+    rock_noun = %w[deed pebble stone rock rock boulder]
```
Sarvatt: an extra rock here

Hazado: Yes intentional
Deed, pebble, stone, small rock, large rock, boulder

workorders.lic
```diff
@@ -246,6 +247,9 @@ class WorkOrders
     stow_hands
 
     quantity.times do |count|
+      if (info['stock-volume'] - (count * recipe['volume'])) > 0
```
rpherbig: 💡 I'd like to see `(info['stock-volume'] - (count * recipe['volume'])` extracted into a well named var to give more context to the next dev that comes along

workorders.lic
```diff
@@ -371,7 +385,9 @@ class WorkOrders
       end
     end
     leftover = (quantity * recipe['volume']) % 10 != 0
-    dispose("#{info['sew-stock-name']} cloth") if leftover
+    dispose("#{info['sew-stock-name']} cloth") if leftover &&  @retain_crafting_materials
```
rpherbig: ❓ Should this be `!@retain_crafting_materials`?

workorders.lic
```diff
@@ -246,6 +247,9 @@ class WorkOrders
     stow_hands
 
     quantity.times do |count|
+      if (info['stock-volume'] - (count * recipe['volume'])) > 0
+        rock_count = info['stock-volume'] - (count * recipe['volume'])
```
rpherbig: 💡 `rock_noun[rock_count]` could also use a well named var

workorders.lic
rpherbig: ❓ I'm not clear on how `rock_noun` and `rock_count` work. Isn't `stock-volume` assuming you get a `stone`?

If I'm reading the loop on line 249 correctly, the first time through the loop `count` is zero, so `rock_count = info['stock-volume']` (i.e. buy a `rock`). How does that account for when `recipe['volume']` requires something bigger than a rock (or allows for something smaller like a pebble`?

Hazado: No currently in crafting yaml its a small rock,
Volume 3. Currently no workorders in recipes are larger than 3 volume.

workorders.lic
```diff
@@ -246,6 +247,7 @@ class WorkOrders
     stow_hands
 
     quantity.times do |count|
+      dispose("#{info['stock-name']} #{material_noun[material_volume+recipe['volume']]}") if (count > 0 && spare_stock)
```
rpherbig: Thank you for renaming `material_[noun|volume]`, it makes this clearer.

Could you extract `#{material_noun[material_volume+recipe['volume']]}` into a well-named variable to help provide some context?

workorders.lic
```diff
@@ -258,32 +260,40 @@ class WorkOrders
...
       else
         find_shaping_room(@hometown, @engineering_room)
       end
+      bput('swap', 'You move') if right_hand =~ /#{@noun}/i
 
-      wait_for_script_to_complete('carve', [recipe['chapter'], recipe['name'], info[stock_name], bone_carving ? 'stack' : 'rock', recipe['noun']])
+      wait_for_script_to_complete('carve', [recipe['chapter'], recipe['name'], info[stock_name], bone_carving ? 'stack' : material_noun[material_volume], recipe['noun']])
+      
+      material_volume = info['stock-volume'] if material_volume == 0
+      material_volume -= recipe['volume']
```
rpherbig: Why subtract `recipe['volume']` here just to add it again on line 296?

workorders.lic
Hazado: Its hard to do that because of the loop its in

workorders.lic
Hazado: its not needed actually...the -=recipe['volume'] takes care of the issue.

workorders.lic
Hazado: Its not needed anymore.

Hazado:
Any more issues with this anyone can see or have encountered?

Tarhish:
Sorry, just poking around -- this is cool, but I had to run my crafter through its paces.

Knitting: seems to work as advertised.  Keeps yarn around and uses it.  I like this.

Sewing: does save the cloth but there's no logic to combine/count it, so it just stacks up burlap cloths in your bags.  We should probably either add counting/combining or remove retention for that discipline for now.

Carving stone:  Seems to work.  I can complete shaping and carving recipes, and although I don't have a good way to test work orders using pebbles to see if it uses their leftovers I think it looks right.

Shaping: it works, but when it scrapes lumber the lumber is dropped at your feet.  Previously, dispose() would handle picking up the lumber and dropping it again as trash, but since we're skipping dispose now it just sits there and relies on ;afk to pick it up before leaving the room.  Since ;afk is really more of a bandaid for a problem we intend to solve at some point, we should definitely handle stowing that lumber away.

Hazado:
@Tarhish Please test again, should have fixed what you found.
Let me know how shaping goes, think the coding will work, but could use some improvement or fixes.

Tarhish:
I can take a look at shaping -- I think it'll work better now.

For sewing, though, two things.  One, I would suggest adding stock name to the cloth whenever you use it.  There are going to be a lot of people out there with a chamois cloth in the same bag they use to craft, and that'll break everything.  Two, I think this will work to get the length of preexisting cloth just fine, but it doesn't combine them together.

I think just a get my burlap cloth and get my other burlap cloth, then combine and stow in the crafting bag will work.  The game response is "You combine the bolts of cloth together."  I'll try it out and see what I see.

Tarhish:
Never mind, I'm totally wrong.  order_fabric will force a combine if your hands are full, so it should work.  Let me try it out as is

Tarhish:
Sewing and Shaping both seem to do what's expected now.  I turned off afk to make sure I wouldn't miss some sleight of hand for foot-wood, and the cloths are being combined just fine.  I still would suggest adding the fabric type to the gets and counts though.  Most people will be running around with a cloth on their person, and it's best to avoid any possible ambiguity.

Hazado:
Okay, made it more specific with the cloth section for sewing.
Try it one more time with new code and no issues, I think this is ready to be pushed.

Tarhish:
Yeah, I don't see any problems with it.  I kept poking, but everything seems to be behaving itself.

rpherbig:
Thank you @Hazado and @Tarhish !

PR #3723, opened 2019-07-29
Test overhaul

rcuhljr:
Work in progress.


rcuhljr:
No longer a WIP, someone can feel free to look it over and merge it.

test/test_items.rb
```diff
@@ -2,27 +2,35 @@
 load 'test/test_harness.rb'
 
 class TestItems < Minitest::Test
-  def setup
-    load('common.lic')
+  def teardown
+    @test.join if @test
```
jandersson: Is the idea with this to build up the tests then execute them all at once?

jandersson: Or what is happening here?

rcuhljr: It prevents the test from thinking it's done before the thread that the actual test is in has finished running. Otherwise it can just move on and say "All good, no failed assertions" even though the script is still running in another thread.

rcuhljr:
Most of our tests (There's still some offenders) run the code in question in their own thread to prevent contamination between tests. Because of that we grab on to the thread the test is running in via @test and then the teardown will call join which basically tells ruby to let @test finish fully then resume with what it was doing.

PR #3725, opened 2019-07-30
healme.lic - min mana to heal wounds/scars

Hazado:
Changed use_auto_mana for waggle to use the minimum amount of mana necessary to heal wounds completely when using "Heal Wounds" and "Heal Scars" spells

### DIFF
--- a/healme.lic
+++ b/healme.lic
@@ -74,19 +74,21 @@ class HealMe
       data = check_discern(spells['Heal Wounds'], @settings) if spells['Heal Wounds']['use_auto_mana']
       if data
         maxmana = part.sum{ |x| if x['type'] =~ /Fresh/i;x['level'];else;0;end;} * 2
-        spells['Heal Wounds']['cambrinth'] = []
-        for i in 0...@settings.cambrinth_items.size do
-          if maxmana > 1
-            if @settings.cambrinth_items[i]['cap'] >= maxmana
-              spells['Heal Wounds']['cambrinth'] << [maxmana - 1]
-              maxmana = 1
-            else
-              spells['Heal Wounds']['cambrinth'] << [@settings.cambrinth_items[i]['cap']]
-              maxmana -= @settings.cambrinth_items[i]['cap']
+        if data['cambrinth'].flatten.sum >= maxmana
+          spells['Heal Wounds']['cambrinth'] = []
+          for i in 0...@settings.cambrinth_items.size do
+            if maxmana > 1
+              if @settings.cambrinth_items[i]['cap'] >= maxmana
+                spells['Heal Wounds']['cambrinth'] << [maxmana - 1]
+                maxmana = 1
+              else
+                spells['Heal Wounds']['cambrinth'] << [@settings.cambrinth_items[i]['cap']]
+                maxmana -= @settings.cambrinth_items[i]['cap']
+              end
             end
           end
+          spells['Heal Wounds']['mana'] = maxmana
         end
-        spells['Heal Wounds']['mana'] = maxmana
       end
       loop do
         Flags.reset('hm-partial-heal')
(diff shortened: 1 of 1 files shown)


rpherbig:
Could you elaborate on what problem you're solving with this change, and how this code solves it?

Hazado:
Well when you use auto mana it uses the maximum amount of mana you can. Thats not needed to heal level 1 wounds. You quickly start running out of mana after doing a few max casts. This saves you mana basically by using the minimum amount of mana needed to heal the wound.

healme.lic
```diff
@@ -69,32 +73,44 @@ class HealMe
...
   def heal_wound(spells, part, internal)
     if @using_waggle
-      spells['Heal Wounds']['cast'] = "cast #{part}"
-      spells['Heal Wounds']['cast'] = "cast #{part} internal" if internal
-      check_discern(spells['Heal Wounds'], @settings) if spells['Heal Wounds']['use_auto_mana']
+      spells['Heal Wounds']['cast'] = "cast #{part.first['part']}"
+      spells['Heal Wounds']['cast'] = "cast #{part.first['part']} internal" if internal
+      data = check_discern(spells['Heal Wounds'], @settings) if spells['Heal Wounds']['use_auto_mana']
+      if data
+        maxmana = part.sum{ |x| if x['type'] =~ /Fresh/i;x['level'];else;0;end;} * 2
```
rpherbig: ❌ Please convert this to a multi-line block for readability (and consider `max_mana` as the var name).

healme.lic
```diff
@@ -69,32 +73,44 @@ class HealMe
 
   def heal_wound(spells, part, internal)
     if @using_waggle
-      spells['Heal Wounds']['cast'] = "cast #{part}"
-      spells['Heal Wounds']['cast'] = "cast #{part} internal" if internal
-      check_discern(spells['Heal Wounds'], @settings) if spells['Heal Wounds']['use_auto_mana']
+      spells['Heal Wounds']['cast'] = "cast #{part.first['part']}"
+      spells['Heal Wounds']['cast'] = "cast #{part.first['part']} internal" if internal
```
rpherbig: 💡  What do you think about `spells['Heal Wounds']['cast'] += " internal" if internal`?

healme.lic
```diff
@@ -35,15 +35,19 @@ class HealMe
     wounds.map { |wound| wound['level'] }.inject(&:+)
   end
 
-  def spam_heal(base, camb)
+  def spam_heal(base, camb, spells)
```
rpherbig: ❓ What do you think about passing in `spells['Heal']` here (instead of `spells`)?

healme.lic
```diff
@@ -69,32 +73,44 @@ class HealMe
 
   def heal_wound(spells, part, internal)
```
rpherbig: ❓ What do you think about passing in `spells['Heal Wounds']` here (instead of `spells`)?

healme.lic
```diff
@@ -69,32 +73,44 @@ class HealMe
...
-      pause 0.5 while cast_spell('HW', @settings.empath_healing['HW'], part, internal)
+      pause 0.5 while cast_spell('HW', @settings.empath_healing['HW'], part.first['part'], internal)
     end
   end
 
   def heal_scar(spells, part)
     if @using_waggle
-      spells['Heal Scars']['cast'] = "cast #{part}"
-      check_discern(spells['Heal Scars'], @settings) if spells['Heal Scars']['use_auto_mana']
+      spells['Heal Scars']['cast'] = "cast #{part.first['part']}"
```
rpherbig: ❓ If I'm reading this right, you're only casting on the first part, but calculating mana based on all parts? That's not what I would expect, could you help me understand the mana calculation here?

Hazado: When you heal the wound first, it causes a scar. So wound external is 1, scar external is 1. Heal the external, scar external goes to 2.  So calcing across entire value, lets you heal for correct amount. Could multiply by 4 and only grab scars, but was less code to just grab all parts

healme.lic
Hazado: we can't without eliminating the empath_healing integration from yaml

healme.lic
Hazado: we can't without eliminating the empath_healing integration from yaml

Hazado:
Noticed that using calculate_mana() redid the discern values.
Had to bring that code over to make it so it doesn't alter the discern stuff.

PR #3726, opened 2019-07-30
common-arcana.lic - check for find/stow cambrinth

Hazado:
Adds a check to see if your already hold your cambrinth before trying to get it
Adds a check to see if you have already stowed your cambrinth before you try to stow it again.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -200,18 +200,22 @@ module DRCA
   end
 
   def find_cambrinth(cambrinth, stored_cambrinth, cambrinth_cap)
-    if stored_cambrinth
-      DRC.bput("get my #{cambrinth}", 'You get')
-    elsif DRSkill.getrank('Arcana').to_i < cambrinth_cap * 2 + 100
-      DRC.bput("remove my #{cambrinth}", 'You remove', 'You slide', 'You sling', 'You take', 'You detach')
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -200,18 +200,22 @@ module DRCA
   end
 
   def find_cambrinth(cambrinth, stored_cambrinth, cambrinth_cap)
-    if stored_cambrinth
-      DRC.bput("get my #{cambrinth}", 'You get')
-    elsif DRSkill.getrank('Arcana').to_i < cambrinth_cap * 2 + 100
-      DRC.bput("remove my #{cambrinth}", 'You remove', 'You slide', 'You sling', 'You take', 'You detach')
+    if (DRC.right_hand || DRC.left_hand) !~ /#{cambrinth}/i
```
rpherbig: 💡 I think both of these would read better as guard clauses, something like 
```ruby
def find_cambrinth(cambrinth, stored_cambrinth, cambrinth_cap)
    return if ...

    if stored_cambrinth
    ...
```

Hazado: I can do that instead if you want.

rpherbig: It's something to keep in mind for the future, e.g. the next time you're in this file. Doesn't need a whole PR just for that change.

PR #3728, opened 2019-07-30
Ulfhara.lic - added regen and heal

Hazado:
Downcased the food and drink section, so no more Grog and grog :(
added catch for when you try to give supplies but they already are happy
added flags.delete when script exits


ulfhara.lic
```diff
@@ -122,6 +123,23 @@ class UlfHara
         combo
       when /help any of our wounded you encounter/
         fput("accept task")
+        if DRStats.empath? && !DRSpells.active_spells.include?('Regenerate') && @settings.waggle_sets['healing']['Regenerate']
+          data = @settings.waggle_sets['healing']['Regenerate']
```
jandersson: You will be surprised at how many folks have 'regenerate' vs 'Regenerate'. This isn't a call to action, but a note that at some point the keys in the Hash should be downcased before comparing.

ulfhara.lic
```diff
@@ -122,6 +123,23 @@ class UlfHara
...
+          check_discern(data, @settings) if data['use_auto_mana']
```
jandersson: should we assign data to the return from this? Since `check_discern` returns the modified data

ulfhara.lic
```diff
@@ -122,6 +123,23 @@ class UlfHara
...
+          release_cyclics
+          DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell") unless checkprep == 'None'
```
jandersson: We do this in several places. Consider adding as a method to DRCA.

ulfhara.lic
```diff
@@ -122,6 +123,23 @@ class UlfHara
...
+          DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
```
jandersson: We do this in several places. Consider adding as a method to DRCA.

ulfhara.lic
Hazado: the only reason to assign the return is if you haven't don't have is set already. Which I did just above.

ulfhara.lic
Hazado: would make it much easier to do

ulfhara.lic
```diff
@@ -122,6 +123,23 @@ class UlfHara
...
+          Flags.reset('spell-ready')
+          prepare?(data['abbrev'], data['mana'], data['symbiosis'], 'prep')
+        elsif DRStats.empath? && !DRSpells.active_spells.include?('Heal') && @settings.waggle_sets['healing']['Heal']
+          data = @settings.waggle_sets['healing']['Heal']
+          check_discern(data, @settings) if data['use_auto_mana']
```
jandersson: check_discern returns modified data

ulfhara.lic
jandersson: Consider it for a future PR, if you're feeling like it :)

ulfhara.lic
Hazado: yes, but it directly modifies it, not a local modification.

jandersson: `data = check_discern(data, @settings) if data['use_auto_mana']` makes more sense, since we'll be updating the data with the results from `check_discern`

ulfhara.lic
Hazado: changed.

ulfhara.lic
Hazado: changed.

jandersson: nevermind! TIL that hashes are modified in place

jandersson:
Its an event script, Id merge it ASAP

There's some recreating the wheel with the healing casting which could be integrated with common-arcana/healing/empath some time

PR #3731, opened 2019-08-01
common.lic - Additional flavor added

Hazado:
Added a few adjectives and other items


Hazado:
really need to convert this to be in the sorting yaml

PR #3739, opened 2019-08-07
WIP: Add trader bonus calculator

jandersson:
Just a code golf implementation of https://elanthipedia.play.net/Trader_gem_selling_bonus#Primary_Formula

### DIFF
--- a/setupaliases.lic
+++ b/setupaliases.lic
@@ -43,7 +43,9 @@
   # List all escort zones from base-hunting alphabetically by name.
   ['ez', "#{$clean_lich_char}e get_data('hunting').escort_zones.sort.each{ |x| echo x }"],
   # Search for a hunting zone, e.g. 'fz moth' will find all zones with moth in it
-  ['fz', "#{$clean_lich_char}e get_data('hunting').hunting_zones.find_all { |k, v| k =~ /\\?/i}.each { |zone| respond \"\#\{zone[0]\}:  \#\{zone[1].join(', ')\}\" }"]
+  ['fz', "#{$clean_lich_char}e get_data('hunting').hunting_zones.find_all { |k, v| k =~ /\\?/i}.each { |zone| respond \"\#\{zone[0]\}:  \#\{zone[1].join(', ')\}\" }"],
+  # Calculate your unofficial trader bonus according to elanthipedia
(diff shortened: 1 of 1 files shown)


rcuhljr:
`case` supports `else` instead of needing to rely on float.

jandersson:
Yea thats probably the more reasonable approach.

jandersson:
It also needs to be capped and probably check if you are a trader

rpherbig:
This feels to me like it would be cleaner as a small script

rcuhljr:
I think it just returns zero instead of failing if you aren't a trader? Running it on a non trader I'm not really worried if it returns nonsense as long as it doesn't boom.

rcuhljr:
In the spirit with which it was written I'm fine with it as an alias instead of a script.

Hazado:
It returns your trader bonus even if your not a trader.
Just got a 12 as an empath.

rcuhljr:
I meant the call to DRSkills.getrank for a skill you don't have will return 0. Anyone who runs it on a non trader isn't someone worth worrying about beyond it not erroring.

jandersson:
Im kind of with rpherbig.

 A generic `;calc` script might be nice and more extendable. But definitely more work than I intended ;)

But, I think a script to do just this is kind of silly. So here I will try to over complicate things:

If I make this a small script. Should I generalize it enough to calculate any formula?

e.g. 
```
;calc trader-bonus
;calc total-tdps-earned
;calc bonus-from-melee-mastery
```

rcuhljr:
Fine then build a generic rules engine and add support for variable templating so users can submit formulas like `$cha+$trading*$appraisal` :)

jandersson:
Thats not what i meant, someone would have to write these. The input would be like ;bescort

rcuhljr:
Sure I was one upping the 'lets overly complicate this' direction :P

rpherbig:
I'm a fan of starting small with what you need right now.

E.g. a script named `traderbonus` that looks something like:
```ruby
custom_require.call(%w[drinfomon])

if !DRStats.trader?
    echo "Not a trader"
    exit
end

x = DRStats.charisma*(DRSkill.getrank('Trading') + DRSkill.getrank('Appraisal')).to_f
echo case x
when 0..2100
    (x/210).floor
when 2101..8400
    ((x/350)+1).floor
else
    (((x+444)/2211)+21).floor
end
```

rcuhljr:
Let's be aware that this information is never needed except when selling pouches at which time you get the value exactly by selling the pouch :)

jandersson:
Its definitely just a curiosity/navel gazing 99% of the time output.

PR #3740, opened 2019-08-09
Move more stuff into common play

Tarhish:
Adding a few more universal items from performance, as well as skip_clean.

Right now, my ;performance has an entry to remove hand items, calls play_song?, then goes into the watch loop until it's done.  I bet that behavior could be put into other scripts with a flag instead, and I'll try it in ;trade.

After this goes live I can at least add the performance changes so it'll support zills and cowbells.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -392,8 +392,10 @@ module DRC
     g
   end
 
-  def play_song?(settings, song_list, worn = true)
+  def play_song?(settings, song_list, worn = true, skip_clean = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
+    UserVars.song = song_list.first.first unless UserVars.song
+    fput('release ecry') if DRSpells.active_spells["Eillie's Cry"].to_i > 0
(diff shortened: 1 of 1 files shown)


common.lic
```diff
@@ -392,8 +392,10 @@ module DRC
     g
   end
 
-  def play_song?(settings, song_list, worn = true)
+  def play_song?(settings, song_list, worn = true, skip_clean = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
+    UserVars.song = song_list.first.first unless UserVars.song
+    fput('release ecry') if DRSpells.active_spells["Eillie's Cry"].to_i > 0
```
rpherbig: 💡 I suspect this could be simplified to `if DRSpells.active_spells["Eillie's Cry"]`. Does EC show up even if its inactive?

rpherbig:
> After this goes live I can at least add the performance changes so it'll support zills and cowbells.

I will merge any PR in the pursuit of more cowbell

common.lic
Tarhish: I don't even get why it's in there at all, now that I look at it.  But I think it'll just show up as nil if it's not active, 0 if it's fading, and > 0 otherwise.  Just `if DRSpells.active_spells["Eillie's Cry"]` should work?

rpherbig: That's my understanding, but I'd definitely test it with a Bard to be sure

PR #3742, opened 2019-08-12
Updated enchant to pull movement out

dr-leustyin:


craft.lic
```diff
@@ -144,6 +144,7 @@ class Craft
       order_item(crafting_data['artificing'][@hometown]['stock-room'], 2)
       order_item(crafting_data['artificing'][@hometown]['stock-room'], 6)
       stow_hands
+      walk_to(@training_room)
```
Sarvatt: Why remove this from craft.lic? You only want to move the movement logic into workorders right? Craft does movement  on its own for every other craft

enchant.lic
```diff
@@ -59,10 +59,6 @@ class Enchant
     @equipment_manager = EquipmentManager.new
     @equipment_manager.empty_hands
 
-    if @use_own_brazier
```
rpherbig: 💡 Can we remove the `DRCT` references from this script now?

craft.lic
rpherbig: 💡 It really feels like we ought to have a helper method for this

rpherbig:
@dr-leustyin my only concern is that someone might be using `enchant` in their YAML when now they should be using `craft`. Are you cool with monitoring LNet for a bit after this is merged?

enchant.lic
dr-leustyin: Done!

craft.lic
dr-leustyin: Any suggestions?  walk_to in itself is a helper method

Issue #3693, opened 2019-07-18
Riverhaven's Alchemy Society + go2


Kaesken:
So apparently Riverhaven's alchemy society used to be just plain "Alchemy Society" and now it's "Riverhaven Alchemy Society".  The go2 can get me in there but using go2 to get out, I keep getting:

--- Lich: go2 active.

[go2: error: your current room was not found in the map database]

--- Lich: go2 has exited.

--
Edit; definitely busts the entire thing if you have set to workorders and hometown is Riverhaven

[workorders: In an unknown room, manually attempting to navigate to 19041]

[workorders: failed to find a matching room]

[workorders: In an unknown room, manually attempting to navigate to 19055]

[workorders: failed to find a matching room]

[workorders: In an unknown room, manually attempting to navigate to 19043]

[workorders: failed to find a matching room]

[workorders: In an unknown room, manually attempting to navigate to 19059]

[workorders: failed to find a matching room]

[workorders: In an unknown room, manually attempting to navigate to 19045]

[workorders: failed to find a matching room]

[workorders: In an unknown room, manually attempting to navigate to 19047]

dakre18:
i noticed this same issue, only in TF, in multiple society buildings in The Crossing.

Kaesken:
I haven't experienced it in Crossing.  Just Riverhaven since they added the town name into the societies title which is probably what's causing the go2 issue.  I already fixed my Genie's maps for the change but I don't know how to do the same for Lich's go2.

Tarhish:
I believe I fixed the Haven Alchemy and Engineering societies now, but keep an eye out for more.  I'm glad they have the town names in the society buildings now but boy that change made it weird.

Hazado:
Fixed Fang Cove earlier today

Kaesken:
Yeah, same here but I don't think they have Crossing's alchemy society with the town name in it.  Probably why they changed Riverhaven's due to both buildings having the same ID name.  I'll keep an eye out for anymore!

Kazmous:
Engineering seems to work now with the following potential small exception if you don't have an engineering_room: set . After ordering the materials, it just immediately gives [workorders: ***Failed to find an empty room, stopping the search***] and starts crafting in the resource depot room (ID 9110).

[...]

PR #3744, opened 2019-08-12
Fix riverhaven crafting rooms

Hazado:
Fixed the crafting rooms for riverhaven.


Hazado:
Should fix the issues for #3693

PR #3756, opened 2019-08-14
 More specific killing of scripts 

Hazado:
prevent killing of wrong script if multiple are running

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -125,7 +125,7 @@ module DRC
     verify_script(name)
     script_handle = start_script(name, args.map { |arg| arg =~ /\s/ ? "\"#{arg}\"" : arg }, flags)
     if script_handle
-      pause 0.5
+      pause 2
       pause 0.5 while Script.running.include?(script_handle)
     end
     script_handle
--- a/common.lic
+++ b/common.lic
@@ -123,13 +123,12 @@ module DRC
 
   def wait_for_script_to_complete(name, args = [], flags = {})
     verify_script(name)
-    before = Script.running.find_all { |s| s.name =~ /#{name}/i }.length
-    success = start_script(name, args.map { |arg| arg =~ /\s/ ? "\"#{arg}\"" : arg }, flags)
-    if success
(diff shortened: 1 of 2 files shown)


common.lic
```diff
@@ -123,13 +123,12 @@ module DRC
 
   def wait_for_script_to_complete(name, args = [], flags = {})
     verify_script(name)
-    before = Script.running.find_all { |s| s.name =~ /#{name}/i }.length
-    success = start_script(name, args.map { |arg| arg =~ /\s/ ? "\"#{arg}\"" : arg }, flags)
-    if success
-      pause 2
-      pause 1 while before < Script.running.find_all { |s| s.name =~ /#{name}/i }.length
+    script_handle = start_script(name, args.map { |arg| arg =~ /\s/ ? "\"#{arg}\"" : arg }, flags)
+    if script_handle
```
Sarvatt: shortening the delay here is scary. when a script is in before_dying it is considered script.running == false even its still running and we’ve run into lots of serialization problems with wait_for_script_to_complete before because of it. I cant test his but just want to give a heads up that it might be preferred to pad this time a bit if there are issues later.

rcuhljr: I would just put the delay after the loop then since that's when it would be needed, how certain are we about this though?

https://github.com/rcuhljr/dr-lich/blob/master/lich.rbw#L2962-L2964

That really looks like it should run the before_dying before it removes itself from the running list.

Sarvatt: it could just be something specific performance.lic is doing in before_dying tripping it up but we had lots of problems moving it out to that script where it would try to start_script performance while it was still in before_dying which would fail. that wasn’t wait_for_script_to_complete, sorry

Sarvatt: all the calling scripts use   start_script('performance', ['noclean']) unless Script.running?('performance') and if performance was in before_dying it would try to start it again and fail and hang.

Hazado:
2 seconds pause is back

PR #3158, opened 2018-09-27
Improve status monitor flexibility, setup framework for distance analysis

rcuhljr:
A simple change to trigger status monitor on high frequency arrival of unseen messages, even if they are not identical. Also set up for future change to use levenshtein(sp) distance to determine similarity of messages to make a more robust uniqueness filter. Also move settings to base.yaml for custom per player settings.

PR #3758, opened 2019-08-15
Add messages to base-filters

Tarhish:
Added some messages to base-filters that frequently contain different variables and would trip status-mon during PR https://github.com/rpherbig/dr-scripts/pull/3158.

Of note:
`- '^\s+\w+.*:\s+\d+\s+\d+%.*$'`
is intended to capture EXP window output.  I realize there's already a component id=exp in the non-useful tags section, but at some point Simu seems to have changed the xml for EXP (like they did active spells).  Sometimes it's still component_id=exp, but other times it seems to declare a mono output block and just dump it as-is, in which case an EXP will trigger status-mon almost every time, since all those numbers are variable.

I'd add this new style of output to status-mon, but I'm not sure what else ignoring that would catch in its net.  For now that regex seems like it should catch what we need, but maybe other front-ends present it differently...


rcuhljr:
Will exp actually show up in status monitor? It should be blanket ignoring everything inside most xml tags.

Tarhish:
Yeah, everything I put in there was actually tripping spam alerts.  The ones that were actually inside XML tags WERE still getting ignored as planned, but there's a different style of exp presentation where they dropped a mono output block open tag, then the entire exp window, then another line to end the block.  Since status-mon is just checking line-by-line it doesn't see the EXP output as being inside a tag.

Tarhish:
I don't have my notes on what caused it in front of me.  The component_id='exp tag IS still doing things, but I think it was EXP ALL that caused a big butt-dump without individually-tagged lines and spammed status-mon with 30ish alerts all at once.

rcuhljr:
K. Most of the rest look fine, although two of them seem questionable

'^.+You believe you can assemble the two ingredients together once you acquire them.$'
'^Glue should now be applied to the \w+ and assembly with the backing material can begin.$'

The first one has no variability and neither seems like a message you'd see frequently enough to actually trigger things? For those cases I prefer to let them fall through to the seen cache since it is far faster to check than against a regex. Not that speed is a huge concern but as the list of ignores grows it's something I try to be mindful of.

Tarhish:
Maybe?  With a unique unseen threshold of 4 every crafting session with a new object starts up, fills the buffer, then starts punting each new message into spam_alert as it comes in.  It'll learn it, then start up again the next time you craft a necklace instead of a longbow or something.  With a few of those messages in there it doesn't fill up the buffer fast enough to start generating alerts, even with new items.

rpherbig:
Looks good to me. @rcuhljr ?

Issue #3765, opened 2019-08-17
cleaning_cloth in Crossing-training


RyanEOD:
I have my worn_instrument: set to cowbell, I also tried just setting my instrument: to it as well. With a cleaning cloth set, it keeps trying to remove my zills and clean them, when my cowbell needs cleaning. I'm assuming it isn't looking for the variable when trying to clean - 

Looks like the problem is here - 

  def clean_zills
    cloth = @settings.cleaning_cloth

    case bput("get my #{cloth}", 'You get', 'What were you')
    when 'What were you'
      echo('You have no chamois cloth, removing Performance from training')
      @settings.crossing_training.delete('Performance')
      return
    end

    bput('remove my zills', 'You slide')

    loop do
      case bput("wipe my zills with my #{cloth}", 'Roundtime', 'not in need of drying', 'You should be sitting up')
      when 'not in need of drying'
        break
      when 'You should be sitting up'
        fix_standing
        next
      end
      pause 1
      waitrt?

      until /you wring a dry/i =~ bput("wring my #{cloth}", 'You wring a dry', 'You wring out')
        pause 1
        waitrt?
      end
    end

    until /not in need of cleaning/i =~ bput("clean my zills with my #{cloth}", 'Roundtime', 'not in need of cleaning')
      pause 1
      waitrt?
    end

    bput('wear my zills', 'You slide')
    bput("stow my #{cloth}", 'You put')
  end
end

Tarhish:
Thanks for pointing this out.  I see crossing training doing its own thing, so I might as well move everything over to the common method.

Tarhish:
I think I've got the changes worked out -- I'm just going to test them until dirt is achieved.

PR #3767, opened 2019-08-18
Commonize performance in crossing-training

Tarhish:
Crossing training has its own implementation of play_song, but it can use common's instead.  This should close https://github.com/rpherbig/dr-scripts/issues/3765

Everything should be standardized with this now.  play_song? returns true when it did play and false when it didn't, which lets crossing-training decide what to do afterwards.  I kept its user-error check for a missing instrument, but flagged it.

I've tested this with zills and cowbells, but my cowbell is so frustratingly spotless that I haven't seen it clean itself yet.  Still, I don't see any reason for it not to work considering it works in ;performance, and we're using the same method now.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -58,6 +58,7 @@ class CrossingTraining
     release_cyclics
 
     Flags.add('ct-song', 'you finish playing')
+    Flags.add('ct-no-instrument', 'Play on what instrument')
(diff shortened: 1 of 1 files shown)


Tarhish:
I finally witnessed it actually clean dirty instruments, so that's good.  It also plays nice with the sub-methods that call play_song while doing other training.

rpherbig:
Nice!

Issue #3768, opened 2019-08-18
Premium toggle in YAML?


RyanEOD:
Now that the Fang Cove portals are in the go2 mapping, can we do a premium toggle? Setting "force_healer_town: Fang Cove" in my yaml is still taking the mammoth even though I have a variable set in Genie for "premium = true". I believe it should use the portals but for some reason it doesn't believe I'm premie. Setting this in the YAML would be pretty easy. Thoughts?

Hazado:
This should be fixed in the latest update to drinfomon.lic
Should be automatically set now.

rpherbig:
Fixed in #3769

PR #3769, opened 2019-08-18
drinfomon.lic - add check for premium

Hazado:
sets UserVars.premium to true if your a premium/plat subscriber.
This is used for FC portals with go2.

should fix issue #3768

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -770,29 +770,25 @@ unless dead?
 end
 
 def premiumcheck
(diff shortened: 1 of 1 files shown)


rcuhljr:
I'm not seeing any reason to have the outer silence proc wrapper?

rcuhljr:
What's the pause 3 for?

Hazado:
Your right, doesn't need that. Fixed

PR #3770, opened 2019-08-19
drinfomon.lic - fix issue with unintialized DRC

Hazado:


drinfomon.lic
```diff
@@ -770,26 +771,29 @@ unless dead?
...
-  }
-  DownstreamHook.add("premiumcheck_silence", silence)
-  before_dying { DownstreamHook.remove("premiumcheck_silence") }
-  case DRC.bput("ltb info", "You are not currently a Premium or Platinum subscriber.", "Your premium service has been")
-  when "Your premium service has been"
-    UserVars.premium = "true"
-  else
-    UserVars.premium = nil
   end
   DownstreamHook.remove("premiumcheck_silence")
```
rpherbig: Can this `.remove` be removed?

rcuhljr:
Adding a new item with the same hook functions the same as removing the old one I believe if anyone cares to dig through lich.rbw so remove is redundant. Otherwise I'm happy with this.

Hazado:
Removed the .remove

Issue #3169, opened 2018-10-06
Hunt_only_with_moons


Saxblue42:
Hey guys, absolutely love lich, great work, revived the game for me. Anyway, a thought I had for moon mages is a command that would allow only going into combat trainer when a moon was up. This would allow exclusive use of moon weapons and slivers for TKT/TKS, which would be awesome. 
Thanks for all the hard work.

drgabellia:
This is a GREAT idea!  Would it be hard to code?  Seems like just a check to hunting_buddy to see if a moon was up would work?  Just hoping.......

MahtraDR:
@rpherbig can close this. Katoak added this to t2 in https://github.com/rpherbig/dr-scripts/pull/5410

PR #3772, opened 2019-08-19
Moon Mage hunting updates

andymanak:
* Added option to end/skip hunt if no moons available -  #3169 
`- :zone: rats`
`  moons: true`
* Updated a casting message to work with either left or right hand
* Updated cambrinth wearing messages to work with empowered moonblade

### DIFF
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -12,7 +12,7 @@ class CrossingRepair
   def initialize
     settings = setup
 
-    start_script('performance', ['noclean'])
+    start_script('performance', ['noclean']) unless settings.instrument
     @repair_info.each do |repairer, items|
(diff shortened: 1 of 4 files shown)


rpherbig:
In the future, please consider making independent changes separate PRs.

crossing-repair.lic
```diff
@@ -12,7 +12,7 @@ class CrossingRepair
   def initialize
     settings = setup
 
-    start_script('performance', ['noclean'])
+    start_script('performance', ['noclean']) unless settings.instrument
```
rpherbig: Could you elaborate on why this check was added?

andymanak: Added this after I submitted this PR, was intending to do it on the next one.  Sorry for the mistake.

Issue #3766, opened 2019-08-18
Jail-buddy/Pay-debt


blackheart217:
Was arrested in Shard, hometown set to Hib.  

Jail buddy active, found guilty.  

pay-debt began, found debts to both Illithi and Forfedhdar.   It took me all the way down to Hib, where it withdrew the amount for the shard debt, went to Hib justice office and paid the shard amount, then timed out.

Guessing its a hometown issue? but also had debts to both places.

jayRyan24:
@blackheart217 Can you close this issue, it seems that @Hazado fixed it already.

blackheart217:
Sure thing

PR #3785, opened 2019-08-20
jail-buddy.lic - fix for issue #3766

Hazado:
pay based on closest town tag when you need to pay debt
this should fix issue #3766


rpherbig:
@Hazado could you find someone in Slack to test this?

PR #3789, opened 2019-08-21
Changed stow_hands to stow_item for tier 7 and 8 under enchanting.

Dartellum:


rpherbig:
Thanks for catching that

PR #3790, opened 2019-08-21
Shape update

Dartellum:
Okay, fixed the files as they look in the main repo now and with my proposed changes. Again, this also resolves issues 3613 and 3692.


workorders.lic
```diff
@@ -872,7 +893,6 @@ class WorkOrders
           fput("drop my #{left_hand}")
         end
         fput('get logbook') unless [left_hand, right_hand].grep(/logbook/i).any?
-      end
```
Hazado: Should this end be removed?
Looks like it will cause issues.

profiles/base.yaml
```diff
@@ -98,7 +98,7 @@ manipulate_only_to_train: false
 # See combat-trainer documentation above for some examples.
 training_abilities:
   Hunt: 80
-# how high in mind-states to train training_abilities before skipping them
+# how high in mind-states to train training_abilities before skipping them
```
rpherbig: ❓ What's with the whitespace changes in this file?

Dartellum: I do not know. I will try to fix. Since I was in dev mode working on enchanting, many changes went in. I have tried to make sure the files were updated to match existing and then my changes.

Hazado: Think a commit I did just before this one trimmed all the trailing whitespaces.
There is a menu option in notepad++ that lets you do that very easily.

workorders.lic
Dartellum: Looks like it was removed by mistake. Correcting. Good catch.

data/base-crafting.yaml
```diff
@@ -971,6 +995,10 @@ artificing:
     - 14760
     tool-room: 14754
     fount: 3
+    sigil-book: 5
+    blank-scrolls: 8
+    repair-room: 19209
```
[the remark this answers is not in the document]
Dartellum: Fixed.

profiles/base.yaml
Dartellum: I believe I did it.

rpherbig:
Looks good to me

@Hazado any more review comments?

Hazado:
Looks good to me as well.

PR #3793, opened 2019-08-21
steal.lic - Allow steal to use auto mana

Hazado:


steal.lic
```diff
@@ -111,7 +111,10 @@ class Steal
   end
 
   def do_buff(buffs, settings)
-    buffs['spells'].each { |spell| cast_spell(spell, settings) }
+    buffs['spells'].each { |spell|
```
rpherbig: 💡 for multi-line blocks I prefer the `do ... end` syntax over `{ ... }`

PR #3796, opened 2019-08-22
setupalias.lic - missing a # for traderbonus

Hazado:


jandersson:
Thank you! must have been thinking in python

PR #3816, opened 2019-08-27
New script to train scholarship by perceiving sigils

dr-leustyin:


khrispyb:
this is a great idea thanks guys

PR #3829, opened 2019-09-03
Large update to ;trade. Therengia support, hunting, and forging

Tarhish:
This is a biggish update, but it'll be nice to have it out here to test it and get input.

The big changes:
- Therengia support.  It turns out to work fairly well up there.
- Users can add a Schmuckatelli-caravanhunt.yaml file that will be called with hunting-buddy periodically when Trading has EXP to drain and you've stopped in your hometown.  Get boxes, train weapons, do whatever.  This is activated by the <hunt> arg when starting ;trade.
- Forging works now.  Add Forging: 1200 to do a forging workorder every 1200 seconds (or whatever) whenever you stop in a town with a forge.
- Updated for the new LEAD route changes.
- Fixed the intermittent lich crash when stepping out of a caravan at the end of a lead route.

Minor changes:
- Improved caravan recognition for all these weird guildfest caravans
- Mech Lore removed, since that finally happened.
- Complies a bit more with common crafting_training_spells changes.
- Finally got more cowbell
- Ability training timers standardized

The Questions:
1.) This is the first use of dependency's shift_hometown and clear_hometown functions.  When stopping at a town it will try to adjust your hometown to whatever nearby town it needs to be in (ie. if it stops in Muspar'i and wants to do a forging workorder it can't just call ;workorders.  It shifts hometown to Muspar'i, does the workorder there, then clears it, reverting to your yaml-defined hometown.)  Works fine for me, but it's a change.

2.) This script edits the map to prevent users from trying to take caravans over the Therengia rope bridge.  Just snips the timeto at both ends, which it does everytime you start moving the caravan, and undoes anytime you start traveling independently or end the script.  I'm going back and forth about adding a warning if automap starts up, but I'm really not sure how many people are going to take a breather from running caravans to muck with the map.  Besides me.

I can explain any parts of these changes, though I've run with this version of the script for hundreds of hours by this point, and tested it in all provinces.  This will be where I point other volunteers.

### DIFF
--- a/trade.lic
+++ b/trade.lic
@@ -247,7 +247,6 @@ class Trade
   end
 
   def closest_town
-    echo find_town(find_closest_id('trader_outpost'))
     find_town(find_closest_id('trader_outpost'))
   end
 
@@ -542,7 +541,7 @@ class Trade
     return unless @caravan_being_led
     return if @no_ride_rooms.any? { |name| XMLData.room_title.include?(name) } # Don't do this on ferrys - you can't tell when you've arrived.
     track_caravan
-    bput('go caravan', 'You open the door')
+    bput("go #{@caravan.adjective} #{@caravan.noun}", 'You open the door')
   end
 
   def be_outside_caravan # call whenever you need to be outside the caravan
@@ -587,10 +586,9 @@ class Trade
       true
     when /You recall that your (?<description>.*) (?<name>.*) should be located in/
       /You recall that your (?<description>.*) (?<name>.*) should be located in/ =~ response
-      @caravan.noun = name
+      @caravan.noun ||= name
       echo "@caravan.noun: #{@caravan.noun}" if debugging?
-      @caravan.adjective = description
-      
+      @caravan.adjective ||= description
       echo "@caravan.adjective #{@caravan.adjective}" if debugging?
       false
     when /You are far too occupied/
@@ -970,7 +968,8 @@ class Trade
       /The minister reminds you that you still owe (?<dues_amt>.*) (?<dues_currency>Kronars|Lirums|Dokoras) in dues to (?:The )?(?<outpost>.*) and asks that you pay the clerk before s?he issues you a new contract\./ =~ response
       echo "Need to pay #{dues_amt} #{dues_currency} to #{outpost}" if debugging?
       if wealth(@settings.hometown) < dues_amt.to_i
-        room = Room.current.id  # POTENTIAL TO CHANGE HOMETOWN?
+        # If you've run out of money, it will run to your hometown to get more.  Potential to pull from drinfomon for nearest savings account?
+        room = Room.current.id
         ensure_copper_on_hand(@caravan_coins_on_hand + dues_amt.to_i, @settings)
         walk_to(room)
       end
@@ -1111,10 +1110,9 @@ class Trade
     Flags.delete('caravan-upgrade')
   end
 
-  def deposit_coins(withdraw_amt, return_to_id)
+  def deposit_coins(withdraw_amt)
     return if wealth(@settings.hometown) < withdraw_amt
     near_town = closest_town
-    echo "deposit coins #{near_town} and #{@trading_towns['near_town']['deposit']}"
 
     return unless @bank_towns.include?(near_town)
     return unless @trading_towns[near_town]['deposit']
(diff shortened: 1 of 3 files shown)


Tarhish:
User testing revealed a condition where Zoluren trading would get hung up on the leth ferry due to trade route changes.  Also, added a lower threshold for starting a forging workorder, since the normal threshold could easily make it keep working through mindlock.

trade.lic
```diff
@@ -27,7 +27,11 @@ class Trade
       [
         { name: 'duration', regex: /duration/i, optional: false, description: 'Run for a certain amount of time plus the time required to closeup.' },
         { name: 'minutes', regex: /\d+/i, optional: false, variable: true, description: 'Duration of the run in minutes.' },
+        { name: 'hunt', regex: /hunt/i, optional: true, description: 'If Trading xp is above half, hunt occasionally using hunting_info: defined in <name>-caravanhunt.yaml' },
         { name: 'exit', regex: /exit/i, optional: true, description: 'Exits the game after delivering all crates and returning the caravan' }
+      ],
+      [
+        { name: 'hunt', regex: /hunt/i, optional: false, description: 'If Trading xp is above half, hunt occasionally using hunting_info: defined in <name>-caravanhunt.yaml' }
```
jayRyan24: Does this argument set work the same as the `hunt` above, but without the exit? The exit is an optional setting, so it seems like it would function the same? Sorry, I'm just trying to understand why there's two of the same argument sets, but with a caveat of a single optional exit. Not that this isn't right, I just simply don't follow. :D

jayRyan24:
This is a lot of work @Tarhish, thanks a bunch!

trade.lic
Tarhish: For the third argset, <duration> and <minutes> are not optional.  So, functionally, your options to use it are:

;trade no args : just runs caravans forever until it's killed.
;trade closeup (exit optional) : runs existing contracts without accepting new ones then ends, quitting if <exit>
;trade hunt : runs caravans forever, but also does hunts occasionally.
;trade duration xminutes <exit> <hunt>: runs trades for Xminutes before closing up and maybe exiting the game.  If hunt is specified it runs hunts during that time period.

So, for example, last night I wanted to test performance in Zoluren, so I set up a hunt file then ran ;trade duration 180 hunt exit.  Which means "Run for 3 hours, hunting occasionally whenever you stop at your hometown, and after 3 hours have passed stop accepting new contracts and exit the game when you're done."

rpherbig:
@Tarhish thank you for doing this!

Could you ask around in Slack for a Trader to test this?

Tarhish:
Yeah, definitely.  I've got some other interested parties running around with it, but I don't have nearly as much communication with them.

rpherbig:
@Tarhish  any update on this?

Tarhish:
I got some more help testing this and it's working out -- considering the script doesn't really work at all right now since the caravan route changes, I think this is fine to merge.  There can't really be anyone using this.

I'll change my sample yaml and update the wiki, then keep an eye out for problems.

Tarhish:
@rpherbig I'd say this is good to go as is whenever you're ready.

homebrewer12:
Quilleth, you're going to kill me, but I have an amazing idea to integrate into ;trade.  I think I might actualy have to learn how to write ruby to help with it...

Tarhish:
Well, this definitely isn't done, that's for sure.  First priority is getting it pulled so that the script works with the new trade route changes, then I'm working off of things I either know we need or ideas I've been given.

- Some handling of storage boxes for storing lumber/resources during a hunt
- Enchanting and Alchemy
- Non-knitting outfitting
- Sorcery training
- Some way of saving crafted goods to turn in for work orders when you stop at a town with a society
- Commodities (perhaps)

Beyond that, I'd hear any other suggestions, though if I'm the one trying to make the change I'll prioritize things as I think they're useful/easy-to-implement.  If you wanna try your hand at making a change, send me a message and I can unravel how the script is built.

homebrewer12:
actually....it was Commodities.  I'm working on spreadsheets showing their price ranges.

Tarhish:
That's the kind of data that really would be necessary to get that going.  Commodities has such wonky logic that some data on price ranges would be the most useful thing we could have.

Issue #3830, opened 2019-09-04
Appraise focus <object>


Windamere:
Would like have implemented with Ruby maybe into ;appraisal or as stand alone. ;Appraise focus (object)
Preferable a stand alone because not everyone is going to even want to use it due to the 10-15 minute research time.

Appraise focus requires 200 ranks of Appraisal or an item https://elanthipedia.play.net/Item:Clockwork_shark

Appraise focus does train appraisal, it also gives small absorption bonus to the <object> focused, and DOES NOT stack with an RPA AWARD. But it's better than not having a bonus at all.

It works just like research..you Appraise focus (object) wait until Breakthrough! then start using the (object) the bonus lasts about 20-30 minutes at 450 appraisal...may last longer with higher ranks...Objects that can be appraised can be found here- https://elanthipedia.play.net/Appraisal_skill#Compatible_Skills

...with notes on how to focus some items. example for stealth, You appraise focus (Any cardinal direction from that room while hidden)

I wrote it in lich using standard lich scripting, and works for me before buffing for every hunt. Since i'm not good at Ruby or even know what half that stuff means in those ruby scripts. I've posted it here so someone better than me can implement it.
Thanks! 

PS. Just learned something, anything in (less than)(greater than brackets) doesn't show up here when posted.

PR #3831, opened 2019-09-04
Initial commit of focus appraisal.

jayRyan24:
Adding changes to complete #3830.

`;appr focus <item>`

### DIFF
--- a/appraisal.lic
+++ b/appraisal.lic
@@ -13,21 +13,34 @@ class Appraisal
     settings = get_settings
     train_list = settings.appraisal_training
 
-    train_list.each do |task|
-      break if DRSkill.getxp('Appraisal') >= 30
-      case task
-      when 'zills'
-        assess_zills
-      when 'pouches'
-        train_appraisal_with_pouches(settings.full_pouch_container, settings.gem_pouch_adjective)
-      when 'gear'
-        train_appraisal_with_gear
-      when 'bundle'
-        train_appraisal_with_bundle
-      when 'art'
-        train_appraisal_with_art
-      else
-        echo "#{task} is not a valid appraisal_training setting"
+    arg_definitions = [
+      [
+        { name: 'focus', regex: /focus/i, optional: true, description: 'Perform appraise focus on an item.' },
+        { name: 'item', regex: /\w+/i, optional: true, description: 'Item to use appraise focus with.' }
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+
+    if args.focus
+      appraise_focus(args.item)
+    else
+      train_list.each do |task|
+        break if DRSkill.getxp('Appraisal') >= 30
+        case task
+        when 'zills'
+          assess_zills
+        when 'pouches'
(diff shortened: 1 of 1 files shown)


appraisal.lic
```diff
@@ -13,21 +13,39 @@ class Appraisal
...
-      when 'gear'
-        train_appraisal_with_gear
-      when 'bundle'
-        train_appraisal_with_bundle
-      when 'art'
-        train_appraisal_with_art
-      else
-        echo "#{task} is not a valid appraisal_training setting"
+    arg_definitions = [
+      [
```
rpherbig: I _think_ that this will allow `;appr focus` or `;appr broadsword` since both args are optional.

I _think_ this will prevent this problem:

```
arg_definitions = [
      [
	        { name: 'focus', regex: /focus/i, description: 'Perform appraise focus on an item.' },
	        { name: 'item', regex: /\w+/i, description: 'Item to use appraise focus with.' }
       ],
       []
]
```

❓ Could you test these two cases?

appraisal.lic
```diff
@@ -95,6 +113,38 @@ class Appraisal
     waitrt?
     bput('wear my zill', 'You slide')
   end
+
+  def appraise_focus(item)
+    case bput("get my #{item}", "^You get.*#{item}", '^What were you referring', 'But that is already', 'You are already')
+    when 'But that is already', '^What were you referring'
+      echo "Cannot get the item you are trying to focus upon. Exiting..."
```
rpherbig: 💡 Suggestion: include `item` in the echo, something like `"Cannot get item '#{item}'.....`

appraisal.lic
```diff
@@ -95,6 +113,38 @@ class Appraisal
...
+      exit
+    end
+
+    focus(item)
+    waitrt?
+    pause 0.5
```
rpherbig: ❓ Is appraise focus quick enough that we only want to wait a half second? Is that not too spammy?

appraisal.lic
```diff
@@ -95,6 +113,38 @@ class Appraisal
     waitrt?
     bput('wear my zill', 'You slide')
   end
+
+  def appraise_focus(item)
+    case bput("get my #{item}", "^You get.*#{item}", '^What were you referring', 'But that is already', 'You are already')
+    when 'But that is already', '^What were you referring'
```
rpherbig: 💡 Does `But that is already` mean you already have the item out? If so, could we proceed instead of exiting?

appraisal.lic
```diff
@@ -13,21 +13,39 @@ class Appraisal
     settings = get_settings
```
rpherbig: 💡 Minor: I _think_ these three declarations could be moved below the arg parsing.

rpherbig:
After thinking about it more, since the args are all optional this shouldn't break anything for existing users and should be ok to merge as-is. I'd still appreciate considering changes per my comments in a follow-up PR.

appraisal.lic
jayRyan24: This would be an issue with it as an item in your inventory that isn't held, I believe. This is just one of the cases that was listed, I couldn't get it to trigger, or rather didn't even try.

PR #3833, opened 2019-09-04
Addition an OM creating script.

jayRyan24:
I don't think I need to give extra info on waggle_sets, they can look up how those work separately.

```
waggle_sets:
  create_om:
    << : *om
  buff_om:
    << : *om
    << : *pom
    << : *pfe
    << : *bene
    << : *mapp
    << : *center
```

### DIFF
--- a/om.lic
+++ b/om.lic
@@ -12,7 +12,7 @@ class CreateOsrelMeraud
     end
       
     if DRSpells.active_spells.include?('Osrel Meraud')
-      echo ("OM already active!")
+      echo "OM already active!"
       exit
     end
 
@@ -48,6 +48,7 @@ class CreateOsrelMeraud
 
   def create_om
     echo "Creating orb!"
+
     walk_to(@settings.safe_room)
 
     case bput("get my #{@settings.osrel_favor_god} orb", 'You get', 'What were you')
--- a/om.lic
+++ b/om.lic
@@ -25,35 +25,34 @@ class CreateOsrelMeraud
 
     echo "No OM found!"
     
-    buff_om if create_om if get_orb
+    buff_om if create_om? if get_orb?
   end
 
-  def get_orb
-    case bput("tap my #{@settings.osrel_favor_god} orb", 'I could not find', 'The orb is delicate')
-    when 'The orb is delicate'
+  def get_orb?
+    if DRCI.exists?("#{@settings.osrel_favor_god} orb")
       echo "Already have an orb for #{@settings.osrel_favor_god}!"
       return true
-    when 'I could not find'
-      echo "Getting orb for #{@settings.osrel_favor_god}!"
-      wait_for_script_to_complete('favor', [@settings.osrel_favor_god])
     end
 
-    case bput("tap my #{@settings.osrel_favor_god} orb", 'I could not find', 'The orb is delicate')
-    when 'I could not find'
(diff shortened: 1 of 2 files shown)


om.lic
```diff
@@ -0,0 +1,82 @@
...
+
+    echo "No OM found!"
+    
+    buff_om if create_om if get_orb
+  end
+
+  def get_orb
+    case bput("tap my #{@settings.osrel_favor_god} orb", 'I could not find', 'The orb is delicate')
+    when 'The orb is delicate'
+      echo "Already have an orb for #{@settings.osrel_favor_god}!"
```
Sarvatt: could we just reuse @settings.imortal_aspect for this instead of making a new setting? that settings only purpose was to carve beads for om orbs in theurgy.lic.

jayRyan24: Well, the issue we have with that is that immortal_aspect is asking for the creature, not the god. In this instance, I am wanting to keep all things together with the same set of names, `osrel_`. 

```
osrel_favor_god: Firulf
osrel_no_harness: true
osrel_amount: 15
```
I think in this game, what should be done is that theurgy should use the `osrel_favor_god` and then use the base-theurgy to find the creature/aspect based on the god's name.

What do you think?

Sarvatt: I mean, that's fine, it's just more settings when we already have a setting for it and I thought maybe we could piggy back on whats already there instead of adding more settings. This script could just do the conversion alternately the same way favor.lic does, the conversion is available in base-theurgy already. I'm not invested in either option just wanted to bring it up :)

om.lic
```diff
@@ -0,0 +1,82 @@
...
+      echo "This script uses waggle_sets, you need a create_om. It will also buff with a buff_om set."
+      exit
+    end
+
+    echo "No OM found!"
+    
+    buff_om if create_om if get_orb
+  end
+
+  def get_orb
```
rpherbig: 💡 Ruby convention is for predicate functions to be named `foo?` (e.g. `get_orb?`)

om.lic
```diff
@@ -0,0 +1,82 @@
...
+    case bput("tap my #{@settings.osrel_favor_god} orb", 'I could not find', 'The orb is delicate')
```
rpherbig: 💡 Consider using `DRCI.exists?` instead of `bput("tap....`

rpherbig:
Looks good to me, let me know when you're ready for it to be merged

om.lic
jayRyan24: I actually definitely agree with this, but I think it should be the reverse of what you are suggesting, or it shouldn't be implemented at all. Going from creature to immortal name just seems backwards to me within this type of situation, and in the future would be one of those "why did I code it this way" type of things. And then anyone using it would be like "it's trying to find a firfulf orb, why am I putting in owl? Again, I completely agree and I thought of this, but it's just too far backwards for me.

om.lic
jayRyan24: Good call, done!

om.lic
jayRyan24: Good call, done!

jayRyan24:
I've used it a dozen times without an issue. I'm fine with it being included.

om.lic
```diff
@@ -0,0 +1,81 @@
+
```
rpherbig: 💡 comment block

PR #3842, opened 2019-09-07
Quick update to fix a glaring bad code choice

jayRyan24:


jayRyan24:
@rpherbig Real quick change, was an obvious one I simply used early and and didn't fix when I changed how it worked.

PR #3843, opened 2019-09-07
Propagate skip_clean in performance

Tarhish:
skip_clean is sent to play_song? initially, however if the first play attempt fails it isn't propagated to subsequent calls.


rpherbig:
Good catch

PR #3845, opened 2019-09-08
[athletics] Add cowbell support

Ry7n:
Removes hard-coded references to zills and adds messaging hooks for
cowbells. Fixes an issue where cowbells would play but not be cleaned by
athletics.

Leverages the `instrument` field of setup.yaml that crossing-training
already uses.

n.b.: I don't have a cowbell to test this with, but didn't notice any regressions with my zills

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -14,7 +14,7 @@ class Athletics
     @athletics_options = get_data('athletics').athletics_options
     @performance_pause = @settings.performance_pause
     # maintains compatibility with those who had instrument unset
-    @instrument = @settings.instrument || 'zills'
+    @instrument = @settings.worn_instrument || 'zills'
 
     @climbing_target = get_data('athletics').practice_options[@settings.climbing_target]
     @swimming_target = get_data('athletics').swimming_options[@settings.swimming_target]
--- a/athletics.lic
+++ b/athletics.lic
@@ -13,6 +13,8 @@ class Athletics
     @settings = get_settings
     @athletics_options = get_data('athletics').athletics_options
     @performance_pause = @settings.performance_pause
+    # maintains compatibility with those who had instrument unset
+    @instrument = @settings.instrument || 'zills'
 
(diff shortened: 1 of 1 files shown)


rpherbig:
Rather than trying to make the performance logic in athletics more robust, what do you think about using some of the performance helpers in `common`?

E.g. https://github.com/rpherbig/dr-scripts/blob/348514202d6b40c294561e12ebf65c18d7ace670/common.lic#L439

There's a distinction between the `instrument` setting and the `worn_instrument` that isn't handled in athletics and overall duplication that I don't think is necessary.

Tarhish:
I'm poking my head in.  I was going to say it'd be easy to replace the existing functions with just a call to DRC.play_song?, but I don't understand why there's a different song for climbing than normal play.  Is it harder to play songs while climbing?  Does anyone even use that?

Ry7n:
I don't have a strong preference; was just trying to quickly help out some cowbell folks that have to manually clean.

`;first-aid` does something similar, I could move the logic from there, which doesn't mention zills specifically.

Tarhish:
I don't have a problem with making a change like this to increase support in the interim.  I can figure out the commonization in the long run, but yeah there are probably some clonkers with cowbells that could use this.

Slight thing to adjust though, due to archaic terminology, `instrument:` is the setting for if you've got a non-worn instrument, like a tuba or something.  `worn_instrument:` is the base setting that defaults to zills, and what people will clonk their cowbells by.  `instrument:` can't be used in athletics because it requires hands.

So I think that first line could just be `@instrument = @settings.worn_instrument`.  The setting defaults to 'zills' in base, but in my yaml it's 'cowbell'.  Other than that I think this would work as is.

rpherbig:
Thanks for pointing that out @Tarhish 

@Ry7n if you use `worn_instrument` I'll merge this as-is. We can commonize stuff later.

Sarvatt:
mind changing line 17 to @instrument = @settings.worn_instrument to get this through @Ry7n? That defaults to zills in base, no need for the or

Issue #3847, opened 2019-09-09
smith/forge threw away tools


Akhorishaan:
https://pastebin.com/vaevTqQR
Ran into an issue where if the smith/forge scripts encountered an ingot with not enough material for the crafting project, it would throw your tools in a waste bucket.  This can happen sometimes if you have a few small ingots on you for custom crafting projects and they happen to be in an earlier container in your inventory than your stow container, and stow custom ingot isn't set.

Quick rundown of what happened:
Smith script started, bought a bronze ingot, put the ingot in my backpack, then went to an anvil.  Then it studied the book, put an ingot FROM MY SATCHEL on the anvil, tried to do the first pound and found out there wasn't enough material, exited the forge script, went to the forging supplies room, tried to "get my bronze ingot" again, had a no-match error, then throw my tools in the nearest waste bucket.

I had a couple small (1-3 volumes) ingots in my satchel (which is my crafting_container) for some custom projects I was doing, which made it grab one of those and go into a subroutine to get another ingot, but the hands were full since  the forge script didn't put the tools away after finding out there wasn't enough material to work with.

jayRyan24:
This looks like a definite issue within forge.lic. The first thing to be done is that the scripts should handle errors by putting the gear away first, and not just exit. I think at this stage and to get this error, it's always at the beginning, so there needs to be a stow_item for each of the starting items, but this is definitely a bug that should be fixed. Another issue is that craft.lic should handle thing better as well, and instead of just disposing of the items in hand, it should handle them as possible items that belong to the player. 

I think the biggest issue, especially one found in Dependency, is that there's a mindset that it should be done a certain way and if it isn't, then it's being done wrong. I'm not sure how others haven't run into this issue all the time before, including myself, especially the big crafter/Traders. Possibly it's because they use deed packets, and my extra ingots were always kept inside containers in other containers, but regardless, this is an issue.

This is probably where the item cleanup should go for forge. Something big went wrong, but it can at least handle putting the crafting items away.
```
    when 'You need a larger volume of metal'
      echo '***You need a larger ingot to forge this item.***'
      magic_cleanup
      exit
```

jayRyan24:
Usually the first item in the inv list to get is the one last touched. So, this thing should make the ingot, cut it, and then use it. Did it do anything else between the time it cut it, or did you do anything with inventory before that, to reorder?

jayRyan24:
Looks like pound is always hammer and tongs, nm.

[...]

PR #3849, opened 2019-09-09
More fixes to ensure we're using the proper crafting containers.

jayRyan24:
"More fixes to ensure we're using the proper crafting containers for smithing.

Further fixes for this issue #3847.

### DIFF
--- a/smith.lic
+++ b/smith.lic
@@ -73,13 +73,14 @@ class Smith
       else
         buy_item(data['part-room'], part)
       end
-      bput("stow #{part}", 'You put')
+      bput("put #{part} in my #{@container}", 'You put')
     end
   end
 
   def buy_ingot(discipline)
     ensure_copper_on_hand(700, @settings)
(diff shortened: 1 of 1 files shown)


jayRyan24:
I left stow_hands in just in case something was leftover.

smith.lic
```diff
@@ -104,6 +105,7 @@ class Smith
 
     ensure_copper_on_hand(1000, @settings)
     order_item(discipline['finisher-room'], discipline['finisher-number'])
+    bput("put oil in my #{@container}", 'You put')
```
jayRyan24: In this case, the oil is what is bought:
```
>order 6
The attendant says, "You can purchase a flask of oil for 1,000 Kronars.  Just order it again and we'll see it done!"
>order 6
The attendant takes some coins from you and hands you a flask of oil.
>put my oil in my back
You put your oil in your pink backpack.
```

PR #3851, opened 2019-09-10
Adding updated common-moonmage and walkingastro

jayRyan24:
walkingastro uses DRCMM which would be broken if fixes weren't added because they are required by DRCMM.

### DIFF
--- a/walkingastro.lic
+++ b/walkingastro.lic
@@ -9,13 +9,14 @@ class WalkingAstro
     settings = get_settings
     @astrology_prediction_skills_survival = settings.astrology_prediction_skills['survival']
     @no_use_scripts = settings.walkingastro_no_use_scripts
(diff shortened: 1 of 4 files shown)


walkingastro.lic
```diff
@@ -9,13 +9,14 @@ class WalkingAstro
     settings = get_settings
     @astrology_prediction_skills_survival = settings.astrology_prediction_skills['survival']
     @no_use_scripts = settings.walkingastro_no_use_scripts
+    @tele_storage = settings.telescope_storage
```
rpherbig: ❌ The `base` and `base-empty` changes need to be in this PR

jayRyan24: Yup, getting to that. Need to do them one at a time or I'll simply forget, lol. Thank you!

PR #3856, opened 2019-09-11
Update common moonmage methods and astrology

jayRyan24:
Reduce, reuse, recycle. Reducing bputs that do the same thing, will look for this more expansively as well.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2407,7 +2407,7 @@ class TrainerProcess
   def center(target)
     case DRCMM.center_telescope(target)
     when "That's a bit tough to do when you can't see the sky", 'Your search for'
-      DRCMM.store_telescope@telescope_storage)
+      DRCMM.store_telescope(@telescope_storage)
     else
       observe
     end
@@ -2417,15 +2417,15 @@ class TrainerProcess
     case DRCMM.peer_telescope
     when "You believe you've learned all that you can about survival", 'Too many futures cloud your mind - you learn nothing.'
       waitrt?
-      DRCMM.store_telescope@telescope_storage)
+      DRCMM.store_telescope(@telescope_storage)
       check_predict
     when 'The pain is too much'
       waitrt?
-      DRCMM.store_telescope@telescope_storage)
+      DRCMM.store_telescope(@telescope_storage)
       @training_abilities.delete('Astro')
     else
       waitrt?
-      DRCMM.store_telescope@telescope_storage)
+      DRCMM.store_telescope(@telescope_storage)
     end
   end
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1019,7 +1019,6 @@ class SpellProcess
 
     @offensive_spells
       .select { |spell| spell['expire'] }
-      .select { |spell| spell['expire'] }
       .each { |spell| add_spell_flag(spell['abbrev'], spell['expire']) }
     @buff_spells
       .values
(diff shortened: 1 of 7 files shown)


jayRyan24:
@Tarhish I did add those settings to this file, but then realized why I hadn't originally. So, if you add those settings to the hash:
```divination_bones_storage:
  container:
  tied:```

Then that hash is now filled, even though those are empty. So, you cannot do things based on `if hash`. A lot of the code in the system, and it's a cool thing about Ruby, is that you can: `if @hash_setting`, and if it is empty, you do ignore it, etc. So, the div bones above include those in the base yaml on top of another setting that is "well I have div bones"... which the hash could tell you both without the extra setting.  I removed them from my hash in the base to reduce redundant settings.

**Comment by jayRyan24** [comment]
Log of current changes: https://pastebin.com/njQz9M0t

I don't have a way to test the bones or tools, but they are basically copy and pasted.

File: combat-trainer.lic [hunk 1]
```diff
@@ -1977,20 +1976,11 @@ class TrainerProcess
...
-    echo("  @divination_bones_tied: #{@divination_bones_tied}") if $debug_mode_ct
-
-    @have_telescope = settings.have_telescope
-    echo("  @have_telescope: #{@have_telescope}") if $debug_mode_ct
+    @telescope_storage = settings.telescope_storage
+    echo("  @telescope_storage: #{@telescope_storage}") if $debug_mode_ct
 
     @astrology_prediction_skills_survival = settings.astrology_prediction_skills['survival']
-    echo("  @astrology_prediction_skills_survival: #{@have_telescope}") if $debug_mode_ct
+    echo("  @astrology_prediction_skills_survival: #{@astrology_prediction_skills_survival}") if $debug_mode_ct
```
rpherbig: Nice catch ✔️

rpherbig:
Looks good to me

jayRyan24:
> Looks good to me

Yeah overall it's simple changes. I think the only thing that could be an issue would be if a return isn't known and put in, but I went through all the bputs and took all the returns and put them together. I'll be here if anything breaks and I can speed fix them. @rpherbig, thanks.

nro2:
Hey I just opened an issue but astrology is totally broken as of today.  I am guessing it has to do with this PR.

Tarhish:
It looks like part of this is because astrology expects the peer_telescope to return with either a fruitless search or Roundtime, but the new method returns all sorts of messages, so it continues observing forever.

I guess it'd have to be fixed either in astrology so it looks for more than just those two things, or change peer_telescope to return true if it succeeded and false if it didn't, then change other things to use that...

jayRyan24:
Sort of but not really, the issue is that the `open it` return which it expected in astrology is the same thing in the common file but longer, so the `open it` didn't match the `open it blah blah`. I merged a lot of different use cases for random bput peer telescopes all over the code, and removed the smaller one. It's not really that big of an issue to fix. Also, returning true or false is pointless since we use case statements based on said return values to do different things.

Tarhish:
Sorry, I'm referring to one small part of this --  in check_heavens

```
      things_to_try.find do |data|
        result = nil
        DRCMM.get_telescope(@tele_storage) if @have_telescope
        until ['Roundtime', 'You peer aimlessly through your telescope'].include?(result)
          return check_heavens if Flags['bad-search'] == 'is foiled by the daylight'
          next if Flags['bad-search'] == 'turns up fruitless'
          result = observe_routine(data['name'])
        end
        DRCMM.store_telescope(@tele_storage) if @have_telescope
        result == 'Roundtime'
      end
```
Apparently we handled observing with telescopes slightly differently everywhere we used it, and there it came back to bite us.  It saves observe_routine as result then later compares it to see if it's == 'Roundtime'.  Except observe_routine just forwards the result of DRCMM.peer_telescope, so it will NEVER return 'Roundtime'.  It'll just keep checking the skies forever.  But you're right, I thought this was a bigger issue, but it looks like that part's just a small thing.

Tarhish:
I'm regretting shelving my moonmage.  I can't test any of this, but I can guess for this part we'd just want to pull the acceptable messages from DRCMM.peer_telescope and compare them instead of the ones we have here.

Tarhish:
Oh, nm, you solved it already.

PR #3863, opened 2019-09-14
Expansion of Smoke script.

jayRyan24:
Added a setting for smoke, renamed to smoker, added ability for flint/blade lighting, added the ability to just smoke a cigar without an image.

### DIFF
--- a/profiles/Shagium-setup.yaml
+++ b/profiles/Shagium-setup.yaml
@@ -483,4 +483,12 @@ sorter:
   sort_inv_command: false
   sort_look_others: false
   sort_look_items_command: true
-  ignore_categories: lootables|trash  
+  ignore_categories: lootables|trash
+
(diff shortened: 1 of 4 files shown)


jayRyan24:
https://pastebin.com/54iKbqhh

Sarvatt:
love it, lol

jayRyan24:
I like smoking while sitting around and talking and stuff.

Sarvatt:
i am not a fan of subkeys instead of explicit options though that is the new du jour lately, its hard to document in a way people can see and there are many checks in ruby not obvious that you need to check for .empty? instead of nil which comes up a lot. i guess we could add example usages in comments in base-empty for that. dont mind me, i’m not around just cracking up someone is updating smoke.lic that i laughed my ass off making :)

PR #3868, opened 2019-09-16
Fixed astrology bugs.

### LINKED ISSUE #3867: Astrology isn't working right.
Starting today, when training astrology, my moon mage is centering his telescope on Dawgolesh and peering, over and over and over.  Used to only train with RTR, now it isn't using it at all.

jayRyan24:
fixed the bput return for peer_telescope to match the returns that use it within astrology.lic, fixed the storage has issue with the bones.

Fixes #3867.

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -45,7 +45,7 @@ module DRCMM
   end
 
   def peer_telescope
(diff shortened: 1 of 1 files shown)


jayRyan24:
This `open it` return was the only instance of it being used in any case statement, so I just changed the one in common instead of astrology.

PR #3879, opened 2019-09-17
Adding an additional spell prep message

bmcneff:


bmcneff:
Rolled into https://github.com/rpherbig/dr-scripts/pull/3878

PR #3880, opened 2019-09-17
Fix for endless peering bug.

jayRyan24:
So, it appears that the issue is ultimately that the bput returns were too thorough, so the "Roundtime" which it wanted never was received. So, added some success and failures to the list and based the finish on it. Thanks to @Tarhish  who provided the fix.

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -45,7 +45,18 @@ module DRCMM
   end
 
   def peer_telescope
-    DRC.bput('peer telescope', 'The pain is too much', 'You learned something useful from your observation', 'Clouds obscure the sky', "While the sighting wasn't quite", 'You peer aimlessly through your telescope', 'You have not pondered your last observation sufficiently', 'You see nothing regarding the future', "You believe you've learned all that you can about", 'Too many futures cloud your mind - you learn nothing.', 'open it', 'You see nothing regarding the future', 'you still learned more', 'Roundtime')
+    DRC.bput('peer telescope', 'The pain is too much',\
+                               'You learned something useful from your observation',\
+                               'Clouds obscure the sky',\
+                               "While the sighting wasn't quite",\
+                               'You peer aimlessly through your telescope',\
+                               'You have not pondered your last observation sufficiently',\
+                               'You see nothing regarding the future',\
+                               "You believe you've learned all that you can about",\
+                               'Too many futures cloud your mind - you learn nothing.',\
+                               'open it',\
+                               'you still learned more',\
+                               'Roundtime')
   end
 
   def center_telescope(target)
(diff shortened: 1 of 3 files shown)


jayRyan24:
We need testing.

Aurayn:
Thanks to everyone for their efforts. I will try and test today as well.

jayRyan24:
OK, I added more to it because it was still failing for some. That seems to have fixed it for them.

Tarhish:
Oh, very good catch.  Thank you for looking at this.

jayRyan24:
I moved the success and finished arrays to base-constelations and I rearranged the common file's peer returns to be easier to read, also removed a double return.

jayRyan24:
Nothing but positive feedback from this change. Will continue to monitor.

common-moonmage.lic
```diff
@@ -45,7 +45,18 @@ module DRCMM
   end
 
   def peer_telescope
-    DRC.bput('peer telescope', 'The pain is too much', 'You learned something useful from your observation', 'Clouds obscure the sky', "While the sighting wasn't quite", 'You peer aimlessly through your telescope', 'You have not pondered your last observation sufficiently', 'You see nothing regarding the future', "You believe you've learned all that you can about", 'Too many futures cloud your mind - you learn nothing.', 'open it', 'You see nothing regarding the future', 'you still learned more', 'Roundtime')
+    DRC.bput('peer telescope', 'The pain is too much',\
+                               'You learned something useful from your observation',\
```
rpherbig: 💡 You could pull many of these messages from base-constellations instead of duplicating them

rpherbig:
✔️ Let me know when you're ready for this to be merged

common-moonmage.lic
jayRyan24: 👍

jayRyan24:
There are new bones, `divination bones` aren't the only ones. Changed to support all bones.

Obsolete:
```
  def roll_bones
    if @divination_bones_tied
      bput("untie bones from my #{@divination_bones_container}", 'divination bones')
    else
      bput("get bones from my #{@divination_bones_container}", 'divination bones', 'you get')
    end
    bput('roll my bones', 'roundtime')
    waitrt?
    if @divination_bones_tied
      bput("tie bones to my #{@divination_bones_container}", 'divination bones')
    else
      bput("put bones in my #{@divination_bones_container}", 'divination bones', 'you put')
    end
  end
```

jayRyan24:
Moved the redundant bputs and used the new info from base-constellations which I added.

rpherbig:
✔️ Let me know when you're ready for this to be merged

jayRyan24:
@rpherbig It's OK to merge, I'll just fix if anything pops up unexpectedly.

PR #3894, opened 2019-09-20
Common-crafting update.

Dartellum:


common-crafting.lic
```diff
@@ -181,4 +181,38 @@ module DRCC
       DRC.bput("stow my #{logbook} logbook", 'You put', 'What were you referring to')
     end
   end
+
+  def order_enchant(stock_room, stock_needed, stock_number)
+    stock_needed.times do
+      order_item(stock_room, stock_number)
+      stow_crafting_item(left_hand, @bag, @belt)
```
rpherbig: ❓ Will `@bag` and `@belt` resolve correctly? I think you'll need to make them function args

common-crafting.lic
```diff
@@ -181,4 +181,38 @@ module DRCC
...
+      stow_crafting_item(right_hand, @bag, @belt)
+      next unless left_hand && right_hand
+    end
+  end
+
+  def fount(stock_room, stock_needed, stock_number, quantity)
```
rpherbig: 💡 Could add a verb to this function name? What is it doing with a fount?

common-crafting.lic
```diff
@@ -181,4 +181,38 @@ module DRCC
...
+    case bput('tap my fount', /You tap .* inside your .*/, /You tap .*your .*/, /I could not find what you were referring to./)
```
rpherbig: ❓ Would using `DRCI.exists?` or `DRCI.inside?` simplify this?

common-crafting.lic
Dartellum: The resolve for me. stow_crafting_item is a function in common-crafting.

common-crafting.lic
Dartellum: This function checks for the fount, either on a belt, in a container, or on a brazier. Then, the fount count is found and will discard if below a certain threshold and purchase a new one.

common-crafting.lic
Dartellum: I would need to test.

common-crafting.lic
rpherbig: I think you've lucked into some state in which those happen to be populated. This is probably due to some nuance of how import works and we shouldn't rely on that. Using function args is safer.

common-crafting.lic
```diff
@@ -181,4 +181,38 @@ module DRCC
...
+
+  def fount(stock_room, stock_needed, stock_number, quantity, bag, belt)
+    case bput('tap my fount', /You tap .* inside your .*/, /You tap .*your .*/, /I could not find what you were referring to./)
+    when /You tap (.*) inside your (.*)/, /You tap (.*) your (.*)/
+      /(\d+)/ =~ bput('analyze my fount', 'This appears to be a crafting tool and it has approximately \d+ uses remaining')
+      if Regexp.last_match(1).to_i < (quantity + 1)
+        get_tool('fount')
+        dispose('fount')
+        stow_hands
+        order_enchant(stock_room, stock_needed, stock_number)
```
rpherbig: ❌ These three calls to `order_enchant` also need `bag` and `belt`

Dartellum: Added.

Issue #3910, opened 2019-09-25
Improve status-monitor's seen message file handling


rcuhljr:
Modify status-monitor to make a `.bak` backup of the file (replacing the existing one) before attempting to save each time. Add a rescue block in initialization where if marshalling fails it will attempt to restore the backup file and marshal that instead.

PR #3915, opened 2019-09-27
Status-monitor changes - backups, distance checking, startup error fix

Tarhish:
So there are three big changes here which I'd like feedback on.  Other ideas on implementation or material to test it is welcome.

First, occasionally on startup/login statusmon will start triggering on random things.  You have to kill and restart it.  I was able to track that down to a failure to load base-filters with get_data.  For whatever reason, rarely get_data fails to load (only on startup), so now it will request them until base-filters actually fills in with something.  Now, ~20% of the time I'll see the script request the data twice before being satisfied.

Second, backups.  Related to issue https://github.com/rpherbig/dr-scripts/issues/3910.  There's probably a better way to do this, but this works?  When loading status-mon, if it loads up the marshal data successfully it writes a .bak file to the lich/backups folder.  If it fails, it rescues and reads the bak file and rewrites it to seen_messages.  I tested this a number of times by corrupting my messages file and starting status-mon.

Third, distance-checking.  So the previous status-mon changes were meant to catch the weird vaguely-similar spam messages, but we only were measuring unique messages in a specific time-frame.  I'm still running into several false alarms per hour when I turn it on, so I wanted to implement the suggestion @rcuhljr had in the original issue and narrow the field. 

I heard the words 'levenshtein distance checking' and went down to code-mart to steal a method.  It still does repeated line checking as normal (x of the same unseen message in a short time) but for unseen messages it also requires that the messages be similar to ones its seen recently.

To demonstrate with examples of unseen messages...

```
>echo Never seen this message before.
Never seen this message before.
>echo Nor this message neither.
Nor this message neither.
>echo Something something about unknown messages
Something something about unknown messages
>echo Just filling up the buffers.
Just filling up the buffers.
>echo This message shouldn't trigger the alert
This message shouldn't trigger the alert
>echo This one should also be fine
This one should also be fine
```

With similar messages...
```
>echo New test with similar messages.  This is the first message.
New test with similar messages.  This is the first message.
>echo New test with similar messages.  This is the second message.
New test with similar messages.  This is the second message.
>echo New test with similar messages.  This is the third message.
New test with similar messages.  This is the third message.
>echo New test with similar messages.  This is the fourth message.
New test with similar messages.  This is the fourth message.
>echo New test with similar messages.  This is the fifth message.
New test with similar messages.  This is the fifth message.
>
[status-monitor: Passed freq_buffer size]
[status-monitor: ]
[status-monitor: ]
[status-monitor: ]
[status-monitor: New test with similar messages.  This is the fifth message.]
```
hits the alarm

### DIFF
--- a/status-monitor.lic
+++ b/status-monitor.lic
@@ -149,7 +149,7 @@ class GameFilter
   def spam_check(line)
     return if @recent_seen.include?(line)
     @recent_seen << line
-    @recent_seen = @recent_seen[1..50] if @recent_seen.length > 50
+    @recent_seen = @recent_seen[1..20] if @recent_seen.length > 20
     counts = @recent_seen.each_with_object(Hash.new(0)) { |phrase, count| count[phrase] += 1 }
     if counts.values.max > @unique_line_threshold
       echo "Passed unique line threshold #{@recent_seen}"
@@ -181,6 +181,7 @@ class GameFilter
     @spam_line = line
     pause 2
     @recent_seen = []
+    @frequency_buffer = []
   end
   
   def check_distance?(line)
--- a/status-monitor.lic
+++ b/status-monitor.lic
@@ -157,7 +157,7 @@ class GameFilter
     end
 
     return unless check_distance?(line)
-    @frequency_buffer.delete_if{|x| (Time.now - x) > 60}
+    @frequency_buffer.delete_if{|x| (Time.now - x) > 90}
     @frequency_buffer << Time.now
     if @frequency_buffer.size >= @line_frequency_threshold
       echo "Passed freq_buffer size"
@@ -191,28 +191,32 @@ class GameFilter
     false
   end
   
-  def levenshtein_distance(s, t)
-    m = s.length
-    n = t.length
-    return m if n == 0
-    return n if m == 0
-    d = Array.new(m+1) {Array.new(n+1)}
+  def levenshtein_distance(source, compare)
+    # Calculates minimum required edits to change one string into another
+    # For an understanding of the concept, look at http://devnull.absolventa.de/2015/11/24/exploring-levenshtein-algorithm-with-ruby/ 
+    source_length = source.length
+    compare_length = compare.length
+    return source_length if compare_length == 0
+    return compare_length if source_length == 0                                                                                      
+    lev_array = Array.new(source_length + 1) { Array.new(compare_length + 1) } # create a 2D array with the dimensions of both string lengths
 
-    (0..m).each {|i| d[i][0] = i}
-    (0..n).each {|j| d[0][j] = j}
-    (1..n).each do |j|
-      (1..m).each do |i|
-        d[i][j] = if s[i-1] == t[j-1]  # adjust index into string
-                    d[i-1][j-1]       # no operation required
+    (0..source_length).each { |i| lev_array[i][0] = i }
+    (0..compare_length).each { |j| lev_array[0][j] = j }
+
+    # Work through the two strings to determine the various ways to adjust each string by propogating costs forward until it reaches the final element.
+    (1..compare_length).each do |j|
+      (1..source_length).each do |i|
+        lev_array[i][j] = if source[i-1] == compare[j-1]  # What's the minimum cost to adjust nearby characters
+                    lev_array[i-1][j-1]       # The same. No operation required
                   else
-                    [ d[i-1][j]+1,    # deletion
-                      d[i][j-1]+1,    # insertion
-                      d[i-1][j-1]+1,  # substitution
+                    [ lev_array[i-1][j]+1,    # Look left in the matrix. This would be a "deletion". It costs + 1 more than d[i-1][j]]
+                      lev_array[i][j-1]+1,    # Look directly above the current entry in matrix. This would correspond to an insertion which costs +1 additionally
+                      lev_array[i-1][j-1]+1,  # Completely substitute the character, this also costs +1 more  than d[j-1, i-1]
                     ].min
                   end
       end
     end
-    d[m][n]
+    lev_array[source_length][compare_length]
   end
 
 
--- a/status-monitor.lic
+++ b/status-monitor.lic
@@ -260,11 +260,9 @@ class GameFilter
 
   def similarity_scrub(line)
     # pulls items from the line that are likely to be variable, and unlikely to be indicative of a hit
-    # currently, numerals, currency names, npc names.
+    # currently, numerals, currency names
     sub_array = [ /[0-9]+/, /(kronars|lirums|dokoras)/i ]
     sub_array.each { |reg_ex| line = line.gsub(reg_ex, '') }
-
-    UserVars.npcs.find { |name| line = line.gsub(name, '') }
     line
   end
 end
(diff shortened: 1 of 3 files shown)


rcuhljr:
I'm super curious how filter strings was failing to populate but also very lazy, so I guess I'll never know. This looks pretty solid and was what I had in mind with both of my suggestions. Probably rope one other person into testing it for you for a few days and it should be good to go.

One quibble, since we're stealing someones code snippet for calculating string distances and they wrote it in an aggressively unreadable style I'd consider putting a link to the source page you found it at as a comment of that function in case anyone needs to tinker with it or wants to find more information.

Tarhish:
I tried to track filter strings through the process but there wasn't a lot to grab onto.  I don't think I've ever seen any other get_data request fail, even the ones that happen at startup.

Tarhish:
Lemme see if I can unwrap that method a bit.

rpherbig:
✔️ Looks good to me

Tarhish:
It's still behaving, and false alarms have gone way down.  I saw one come in while trading contracts and doing a gear repair run at the same time, which both have a lot of unseen-yet-similar messages, and it seemed to match what I expected.  I was looking for some actual logs of past events to use as test fodder, but so far no luck.

Barring someone dredging something like that out from the archives, I'd say this is fine for now.

rpherbig:
Could you get someone from Slack to test this? I'll merge it once you're confident it's been tested.

Dartellum:
Testing and working so far. My alarm rate went down.

Tarhish:
Thanks a lot -- I saw your mention of the aoe messages that were triggering it and I'm going to make a couple little tweaks, but so far it's still working for me and what others I've seen try it out.  By tomorrow I think I'll give my thumbs up.

Tarhish:
Pulled back the buffer size to 20, since when I bumped it to 50 it was reaching VERY far back.  Also wiping the freq_buffer when it triggers just like the recent_seen buffer, so it won't keep triggering on new messages after an alarm.

Also, does anyone have an opinion on removing numbers [0-9] from the lines the same way clean removes <[^>]+> and replaces them with ''?  Most unseen entries are just "That'll be 43,509 elanthiabucks, please." That would probably eliminate all remaining false alarms from this side of things once a couple messages slip in that cover the remaining commas.

Tarhish:
Oh boy, I tried that out and it did _not_ work.

rcuhljr:
Removing numbers seems reasonable, but I maybe we could snip out `[0-9,]+ (kronars|lirums|dokora)` to be more specific.

Tarhish:
I just realized it was working, I just misinterpreted what that regex was doing in the first place.  Now I've got it acting reasonably after it ran through a couple "gold,  silver,  bronze, and  copper Kronars ( copper Kronars)." along with lirums and dokoras.

So yeah, there might be something to snipping currency as well.

Tarhish:
Actually, I was expecting to have to muck with commas, but I was wrong.  I've gotten so used to textsubs adding them in that I forgot DR doesn't use them naturally, and since status-mon doesn't read the subs it's all invisible.

rcuhljr:
Fair enough, my memory has grown foggy.

Tarhish:
Yeah, I want to do a little more with this.  It's close to where I want it.

Dartellum:
Testing the latest.

status-monitor.lic
```diff
@@ -22,15 +22,23 @@ class GameFilter
 
   def initialize
     @seen_cache_name = "seen_messages_#{checkname}.dat"
+    @seen_backup_name = "backup/seen_messages_#{checkname}.bak"
```
rpherbig: 💡 Why store this in a subdirectory?

Tarhish: The entire reason is that I saw my lich install had a lich\backup directory with nothing in it and I thought, "Huh, maybe this was supposed to be something once upon a time.  Guess I'll put the backups in here."

Unless that was something _I_ made and forgot about.  More than possible.

Tarhish:
Okay, I've been running with this change for most of today, and it seems stable enough to talk about at least.

For repeat unseen messages (old status-mon behavior) it checks the unedited line as before.
```
Testing an unseen message
>echo Testing an unseen message
Testing an unseen message
>echo Testing an unseen message
Testing an unseen message
>echo Testing an unseen message
Testing an unseen message
>echo Testing an unseen message
>
Testing an unseen message
>
[status-monitor: Passed unique line threshold {"     PvP has strict rules!  See NEWS  , NEWS  , NEWS  !"=>1, "      Scripting is ok, but you must be responsive!  NEWS  !"=>1, "           Vulgarity is not allowed in public!  NEWS  !"=>1, "\"Testing an unseen message"=>1, "Testing an unseen message"=>6}]
[status-monitor: ]
>
[status-monitor: ]
[status-monitor: ]
[status-monitor: Testing an unseen message]

**Comment by Tarhish** [comment]
I've been running with this for most of today now, and it's been stable, so I feel like it's at least worth talking about.

New lines come in and are run through the filters (some filters use numerals in the regex, so either way we want to keep that.)
Then the line is stripped of numerals and currency (other things could be added in later).
Then it's checked to see if we've seen it before in seen_messages (this means that there will be a little bit of relearning, but seen_messages will take the stripped line, so it'll grow much more slowly going forward.)
If not, the edited line is passed to check to see if >x copies of that same message have come through recently.
If not, the edited line is compared against other recent similar unseen messages.
eventually these messages are migrated to the seen_messages db.

```
>echo Testing similar messages with random 39403 numerals in it.
Testing similar messages with random 39403 numerals in it.
>echo Testing similar messages with random 43941 numbers in it.
Testing similar messages with random 43941 numbers in it.
>echo Testing similar messages with random 929 digits in it.
Testing similar messages with random 929 digits in it.
>echo Testing similar messages with random 11111111 tallies in it.
Testing similar messages with random 11111111 tallies in it.
>echo Testing similar messages with random 121212 onestwos in it.
Testing similar messages with random 121212 onestwos in it.
>
[status-monitor: Passed freq_buffer size]
[status-monitor: ]
[status-monitor: ]
[status-monitor: ]
[status-monitor: Testing similar messages with random  onestwos in it.]
```

Tarhish:
I've adjusted it so it doesn't consider similar messages if they've come in within half a second of each other.  This significantly reduces alarms due to menus, such as shops, or help menus, or FLAGS, etc.  I spent a while with an echo watching what it was filtering out and was pleased with it.  I don't think this should hinder its intended behavior, and its old x identical messages behavior is untouched.

Tarhish:
Revisiting this after a bit -- I adjusted a couple variables to where I think they should be by default -- the only real change is saying Yes, Hm, Oh? is no longer default.  I kind of feel like that's suspicious at the start when you haven't trained your seen_messages.

I tested the original behavior of seeing new messages and saving them to seen_messages after 10 minutes, and that's still working.  It still alarms on x repeats of an unseen message and then saves it later.  My backups are still getting refreshed, and new similar messages are being processed on top of the initial behavior.  I've only found a couple people willing to test it, but it's all working with them.

As far as I'm concerned this is good to go.  There may be improvements to be had in the additional 'similar message tracking' side of things, but false alarms are way down and the original critical behavior is still there.

rcuhljr:
Works for me, really the short phrases were just a filler idea to make it easy to see if the thing was triggering since most people have conversation windows. They could be completely snipped out and I don't think anyone would miss them.

Tarhish:
Thanks!

Issue #3923, opened 2019-09-30
Walkingastro breaks Astrology


jayRyan24:
Now that we only get the telescope once when doing astrology observations, walkingastro tends to break it with the perfect timing. 

```
R> 
[astrology]>get my telescope

[walkingastro]>stow my telescope

You get a maple simple telescope from inside your duffel bag.
> 
You put your telescope in your duffel bag.
> 
[astrology]>center telescope on Dawgolesh

Center what?
```

PR #3924, opened 2019-09-30
Added check for center what to common and a case to observe.

jayRyan24:
To fix the issue #3923:
```
R> 
[astrology]>get my telescope

[walkingastro]>stow my telescope

You get a maple simple telescope from inside your duffel bag.
> 
You put your telescope in your duffel bag.
> 
[astrology]>center telescope on Dawgolesh

Center what?
```

### DIFF
--- a/astrology.lic
+++ b/astrology.lic
@@ -9,8 +9,6 @@ class Astrology
   include DRCA
   include DRCT
 
-
-
   def initialize
     exit unless DRStats.moon_mage?
 
--- a/astrology.lic
+++ b/astrology.lic
@@ -279,7 +279,7 @@ class Astrology
 
     if @have_telescope
       case DRCMM.center_telescope(body)
-      when /Center what?/
+      when /Center what/
         DRCMM.get_telescope(@tele_storage)
         return observe_routine(body)
       when /open it/
--- a/astrology.lic
+++ b/astrology.lic
@@ -9,6 +9,8 @@ class Astrology
   include DRCA
   include DRCT
 
+
+
   def initialize
     exit unless DRStats.moon_mage?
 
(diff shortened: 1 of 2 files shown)


astrology.lic
```diff
@@ -279,6 +279,9 @@ class Astrology
 
     if @have_telescope
       case DRCMM.center_telescope(body)
+      when /Center what?/
```
rpherbig: ❓ Does this work without a 15 second delay? I don't see `Center what?` in the list of matching messages in `DRCMM.center_telescope`

common-moonmage.lic
```diff
@@ -54,7 +54,7 @@ module DRCMM
   end
 
   def center_telescope(target)
-    DRC.bput("center telescope on #{target}", 'You put your eye', 'open it to make any use of it', 'The pain is too much', "That's a bit tough to do when you can't see the sky", 'Your search for')
+    DRC.bput("center telescope on #{target}", 'Center what?', 'You put your eye', 'open it to make any use of it', 'The pain is too much', "That's a bit tough to do when you can't see the sky", 'Your search for')
```
jayRyan24: Center what?

jayRyan24: @rpherbig

astrology.lic
jayRyan24: Should be down below.

common-moonmage.lic
rpherbig: Did you test that this works?

For some reason I have it in my head that `bput` interprets the match message as a regex, so the `?` in the message gives me pause. Will it be interpreted as the literal `?` or as "Match between 0 and 1 of the preceding token."?

jayRyan24: I just removed it, no reason to include it anyway.

jayRyan24: Good call.

PR #3934, opened 2019-10-02
Seems pointless

jayRyan24:
Need feedback on this, I do this with my own hunting-buddy, just wanting to know if it is a good idea. I see no reason to proceed with _befores_ for a hunt I will not partake in!


Tarhish:
I dunno, makes sense to me.  I can't think of any reason not to make the change.

rpherbig:
Makes sense to me

jayRyan24:
OK, gtg then, I think it's good!

PR #3938, opened 2019-10-03
Moved tactics above overrides

jayRyan24:
Should do tactics first!


Sarvatt:
+1, this change is needed to fix tactics in aiming_trainables:, thanks

PR #3935, opened 2019-10-02
Lets you use an array of optional rooms.

jayRyan24:
```hunting_info:
- :zone:
  - void_moths
  - cabalist
  - archons
  args:
  - d0
  full_waiting_room:
  - 90210
  - 12345
  - 501
  stop_on:
  - Heavy Thrown
  stop_on_low:
  - Locksmithing
  - Athletics

PR #3943, opened 2019-10-05
Adjust room sampling in common-travel

Tarhish:
Someone reported that common-travel throws a fit if idle_room is defined:
`idle_room: 12345` instead of
```
idle_room:
- 12345
```
This pulls the sample only if it's working with an array, otherwise uses it as-is.

### DIFF
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -102,7 +102,8 @@ module DRCT
       (check_mana = min_mana > 0) && !check_mana
 
       if idle_room
-        walk_to(idle_room.sample)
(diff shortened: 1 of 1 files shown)


Tarhish:
I jacked my gryphons entry down to a single room with someone hunting in it and tested full_waiting_room: with both types of entries.

rpherbig:
@jayRyan24 was this the thing from #3935 ?

rpherbig:
In general, I'm not a fan of this approach, but if it's tested I'm ok with it. It's usually a sign we need to change something at the design-level external to this helper.

PR #3947, opened 2019-10-07
Adjust play_song to support climbing songs

Tarhish:
I purchased a climbing rope and changed athletics to use the common play_song logic and support cowbells.  Step 1 would be to change common play_song? to support it for later.

I can show the tests for athletics' climbing song stuff in a later PR, but at the moment as for this change, running performance worked and trained up to max skill, and cleaned the cowbell as expected.  I deleted my song variable and watched it climb the ranks from ditty up to psalm.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -391,42 +391,50 @@ module DRC
     g
   end
 
-  def play_song?(settings, song_list, worn = true, skip_clean = false)
+  def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
     UserVars.song = song_list.first.first unless UserVars.song
+    UserVars.climbing_song = @song_list.first.first unless UserVars.climbing_song || !climbing
+    song_to_play = climbing ? UserVars.climbing_song : UserVars.song
     fput('release ecry') if DRSpells.active_spells["Eillie's Cry"].to_i > 0
-    result = bput("play #{UserVars.song}", 'dirtiness may affect your performance', 'slightest hint of difficulty', 'fumble slightly', /Your .+ is submerged in the water/, 'You begin a', 'You struggle to begin', 'You\'re already playing a song', 'You effortlessly begin', 'You begin some', 'You cannot play', 'Play on what instrument', 'now isn\'t the best time to be playing', 'Perhaps you should find somewhere drier before trying to play', 'You should stop practicing')
+    result = bput("play #{song_to_play}", 'dirtiness may affect your performance', 'slightest hint of difficulty', 'fumble slightly', /Your .+ is submerged in the water/, 'You begin a', 'You struggle to begin', 'You\'re already playing a song', 'You effortlessly begin', 'You begin some', 'You cannot play', 'Play on what instrument', 'now isn\'t the best time to be playing', 'Perhaps you should find somewhere drier before trying to play', 'You should stop practicing')
     case result
     when 'Play on what instrument'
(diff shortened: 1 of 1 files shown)


common.lic
```diff
@@ -391,42 +391,50 @@ module DRC
     g
   end
 
-  def play_song?(settings, song_list, worn = true, skip_clean = false)
+  def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
     UserVars.song = song_list.first.first unless UserVars.song
+    UserVars.climbing_song = @song_list.first.first unless UserVars.climbing_song || !climbing
```
rpherbig: ❓ Should this be `song_list` and not `@song_list`?

rpherbig: ❓ Why does this have `|| !climbing`? It seems to me like it should be sufficient to just check if the UserVar is set

common.lic
```diff
@@ -391,42 +391,50 @@ module DRC
...
       return true if DRSkill.getrank('Performance') < 20
       return true if skip_clean
       return true unless clean_instrument(settings, worn)
-      play_song?(settings, song_list, worn, skip_clean)
+      play_song?(settings, song_list, worn, skip_clean, climbing)
     when 'slightest hint of difficulty', 'fumble slightly'
       true
     when 'You begin a', 'You effortlessly begin', 'You begin some'
-      return true if UserVars.song == 'concerto masterful'
+      return true if song_to_play == 'concerto masterful'
```
rpherbig: 💡 Should this check something like `song_list.last.last` instead of a hardcoded string?

common.lic
Tarhish: Thank you.  Yes, that variable being wrong causes an extra cycle through the play_song to get that set

Tarhish: When I started this I was thinking none of the climbing logic should trigger unless <climbing> was true.  Now that I understand it better I don't think it matters.  Since that unless is really confusing I'll remove it prior to another couple tests.

common.lic
Tarhish: Uh, perhaps.  base-perform is a hash of arrays, and for some mindblowing reason this Ruby allows hash.first but not hash.last.   So it'll be song_list.to_a.last.last, which is slightly weirder.

rpherbig: Not a big deal, you don't need to change this one.

Tarhish: I did check that song_list.to_a.last.last returns concerto masterful, so I just did it anyway.

rpherbig: That's fine too!

Tarhish:
I made the changes to athletics.lic and then adjusted trade.lic to use athletics.lic to run climbing ropes in and around my caravan.  Since then I've run through just about every aspect of this method on the climbing and non-climbing side.  Performing, adjusting songs, cleaning instruments, and then all the same with the climbing rope up.

So far so good.  I think this should work as-is.

rpherbig:
I am unclear: are you asking for this to be merged as-is, or are you going to make a commit with the changes you mentioned in the review comments?

Tarhish:
You're understandably confused, because I forgot to push the changes I made.  I blame my brain.

Tarhish:
Alright, this time it got pushed.  It could wait till tomorrow when I'm around to watch out for it though.

rpherbig:
✔️ Looks good to me.

Let me know when you're ready for me to merge it

Tarhish:
Thanks a lot, man.  Yeah, this should be good to go

PR #3951, opened 2019-10-11
athletics - remove duplicate logic - add cheap-rope support

Tarhish:
This shouldn't be merged yet due to the recent common change.  Also, I want to find someone with a permanent climbing rope to test this, if possible.

Three changes, first, moved stop_play, clean_instrument, and play_song? out to common.  Second, added a flag to kill the training and quit out if your climbing rope runs out of minutes.  Third, added an arg to just stay in your room and not move with a climbing rope.  Mainly this is so I can climb my rope inside or behind my caravan.

The @instrument var up top isn't used anymore.

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -13,8 +13,6 @@ class Athletics
     @settings = get_settings
     @athletics_options = get_data('athletics').athletics_options
     @performance_pause = @settings.performance_pause
-    # maintains compatibility with those who had instrument unset
-    @instrument = @settings.worn_instrument || 'zills'
 
     @climbing_target = get_data('athletics').practice_options[@settings.climbing_target]
     @swimming_target = get_data('athletics').swimming_options[@settings.swimming_target]
@@ -30,7 +28,8 @@ class Athletics
       [
         { name: 'wyvern', regex: /^wy.*/i, optional: true, description: 'Climb both undergondola and wyvern cliffs' },
         { name: 'undergondola', regex: /^un.*/i, optional: true, description: 'Climb branch etc undergondola' },
-        { name: 'xalas', regex: /^xala.*/i, optional: true, description: 'Climb xalas in zoluren in instead of undergondola' }
+        { name: 'xalas', regex: /^xala.*/i, optional: true, description: 'Climb xalas in zoluren in instead of undergondola' },
+        { name: 'stationary', regex: /^stat.*/i, optional: true, description: 'Stand still and use a climbing rope' }
       ]
     ]
 
@@ -44,7 +43,7 @@ class Athletics
     elsif args.xalas
       climb_xalas
     elsif have_climbing_rope
-      train_with_rope
+      train_with_rope(args.stationary)
     elsif @swimming_target
       swim_loop(@swimming_target['rooms'])
     elsif @climbing_target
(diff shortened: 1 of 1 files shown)


Sarvatt:
the song logic has always been wrong for climbing ropes in athletics.lic and pretty much everyone has to manually set their uservar for it to be useful unless athletics and performance are at the same ranks btw. you want to set the song based on athletics skill instead of performance and only back down if you have to keep restarting the song.

Tarhish:
I think someone may have fixed it at some point.  My normal performance on cowbell is 'psalm', and if I delete climbing_song or set it to psalm manually I get this.

```

[athletics]>study cornhusk rope
You're certain you can CLIMB, TAP, PULL, and WAVE a twisted cornhusk rope.
You believe you can use it for 12 minutes per day, and 11 minutes and 45 seconds more today.
[athletics]>play psalm
You believe it will reset in 23 hours 59 minutes 31 seconds.
>
You begin a quiet psalm on your sungold cowbell with only the slightest hint of difficulty.
>
[athletics]>climb practice cornhusk rope
Directing your attention toward your rope, the ends immediately begin twitching.  The rope suddenly unfurls itself completely, moving in pace with the music.  You hang on tightly as a twisted cornhusk rope finishes its transformation, and reaching up with your free hand you begin to climb.
>
The Riverhaven Warden strides in, looking around carefully for signs of trouble.
>
You continue playing on your sungold cowbell.
>
This climb is too difficult, so you stop practicing.  Your cornhusk rope coils back up neatly in your hand.
>
[athletics]>stop play
You stop playing your song.
>
[athletics]>play psalm halt
You begin a halting psalm on your sungold cowbell.
>
[athletics]>climb practice cornhusk rope
Directing your attention toward your rope, the ends immediately begin twitching.  The rope suddenly unfurls itself completely, moving in pace with the music.  You hang on tightly as a twisted cornhusk rope finishes its transformation, and reaching up with your free hand you begin to climb.
>
The Riverhaven Warden scans the area with a watchful eye.
>
Your body is still soaked.
>
The Riverhaven Warden scans the area with a watchful eye.
>
You continue playing on your sungold cowbell.
>
Leaning one way then another while you climb, you manage to keep a twisted cornhusk rope from dancing away with you on it.
```

Tarhish:
I imagine if the skills are far apart in that translation it'd take quite a number of adjustments to get it right though.

Sarvatt:
maybe the limited ropes work differently then? Noone's fixed athletlics.lic from what I can see

Tarhish:
Yeah, this is definitely why I need to find someone with the real thing.

Sarvatt:
I had to force concerto with 200's in performance and 400's in athletics back in the day, we had xp gain tracking in the debug at the time before it was added and it was crazy slower to play it at the defaults if there was a skill gap but at the capped 17 mindstates/6 minutes climb practice has for harder things, hmm.

Tarhish:
Oh, yeah, I can imagine that. If there's a difference in athletics gain between the lower end of acceptable and higher end then it'd climb up the ranks and hit the bottom, which would train horribly.  I imagine you couldn't play a concerto successfully, but did the rope care?

Might be able to define a few acceptable starting points for that.

Tarhish:
The values are wild guesses and I think they'll lean towards the high-end and work down a bit before settling, but anyone can just set their var manually like before if they want to avoid that.

athletics.lic
```diff
@@ -234,11 +164,40 @@ class Athletics
...
     Flags.delete('climbing-too-hard')
+    Flags.delete('climbing-dead-rope')
     bput('stop climb', 'You stop practicing your climbing skills.', "You weren't practicing your climbing skills anyway.")
+    DRC.stop_playing
     fix_standing
     stow_hands
   end
 
+  def check_rope?
+    fput("study #{@settings.climbing_rope_adjective} rope")
```
rpherbig: ❓ Is it feasible to convert this to `bput`?

rpherbig:
✔️ Looks good to me. Can you get someone from Slack to test it?

athletics.lic
Tarhish: Yeah, now I can.  Apparently the permanent ones have the same first line of study as the temp ones.

halprind:
I tested this with a unlimited climbing rope, and doesn't appear to have any issues.

PR #3953, opened 2019-10-12
[combat-trainer] Allow wearing bundles on other locations than shoulders

Ry7n:
The default location for wearing a lumpy bundle is the shoulder, but
it's a fairly in-demand slot. This change adds a new skinning option to
`PULL MY BUNDLE` a number of times, allowing it to be worn on the
shoulders, back or waist as well.

Example:

```yaml
skinning:
  skin: true
  pull_count: 1
```

This will pull the bundle once, draping across the shoulders.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -352,6 +352,9 @@ class LootProcess
     @tie_bundle = skinning['tie_bundle'] || false
     echo("  @tie_bundle: #{@tie_bundle}") if $debug_mode_ct
 
+    @pull_bundle_count = skinning['pull_count'] || 0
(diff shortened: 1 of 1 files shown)


rpherbig:
Could you update base.yaml?

https://github.com/rpherbig/dr-scripts/blob/master/profiles/base.yaml#L59

Hiinky:
Just my two cents,  TOGGLE on Simu side is more reliable for this function and requires less actions.

Ry7n:
@Hiinky I was unaware of TOGGLE. I think that's definitely a better solution. I'm going to close this.

PR #3954, opened 2019-10-12
Updated to only drop items that you try to steal

dr-leustyin:


steal.lic
```diff
@@ -94,7 +94,7 @@ class Steal
     count.times do
       break unless DRRoom.npcs.include?(npc)
       echo("***STATUS*** Stealing from: #{npc}") if UserVars.thievery_debug
-      hide? if @hide_to_steal
+      hide? if @hide_to_steal and DRSkill.getxp("Stealth") < 30
```
rpherbig: ❓ Is hiding _only_ for exp, or does it do anything else like make it easier or safer to steal?

dr-leustyin: Not 100% sure.  I would expect hiding to reduce the chance of being caught. This is one of the optimizations I've been running in my fork to speed up stealing runs for characters that use ;steal to train stealth.

Sarvatt: how does not hiding over 30 ranks help training stealth? surely someone that only wants to hide for 30 ranks can set the yaml setting.. i still learn stealth over 1100 ranks on stealing runs and wouldnt be able to here. also, need to use && not and in ruby unless you explicitly mean to use a lower priority operator, operator precedence is weird with and or or.

Sarvatt: oh wait, getxp is the mindstate one isnt it, can tell i’ve been away a while :D

rpherbig: @dr-leustyin I'm worried that some people might be impacted negatively by this change. Could you poll some other Thieves and ask around how they expect this to behave?

Sarvatt: just remove the exp change ` and DRSkill.getxp("Stealth") < 30` and this will be good to pull, that change can be debated separately and is holding it up.

Sarvatt:
sent you a PR on your branch that removes the mindstate check, that's a separate change to the initial goal that can be submitted after and the only thing blocking this good fix.

Issue #3955, opened 2019-10-12
setting up Buff upkeep during mining


aemenash:
Would there be a way to have mining manager or mining buddy set up to use mining buffs and upkeep the buffs during mining? buffs like wolf scent spell or even a line to use the trader speculate luck ability as cooldown permits?

rpherbig:
It shouldn't be hard to have one of the mining scripts call `buff` and look for a spell set.

aemenash:
I’m really really new to DR and Lich but thought it would be a cool thing, I have NO idea how easy or hard it would be to get that sort of thing added in or not but thought the idea worth posting to the geniuses that are making these scripts lol 😊

Sent from Mail for Windows 10

rpherbig:
Huh, it looks like this may have already been done in #2672

Have you tried adding a buff set called `mining`?

Tarhish:
There's a nuance to it and I forgot about the fact I wanted to fix this.  The mining waggle is called when you start mining, but only at the start of the process.  Buffs wear off as you move room to room, especially when you might spend 15 minutes in one of them.

I was annoyed by this back when I needed the buff to mine successfully, but I didn't do much updating then, and forgot about it.  I can make a change for this.

rpherbig:
Fixed in #3958

PR #3958, opened 2019-10-13
Buff persistence in mining/forestry

Tarhish:
Closes issue https://github.com/rpherbig/dr-scripts/issues/3955

Currently mining/forestry-buddy calls buff at script start, but in any successful prospect some will fall off.  If mine/chop actually runs in a room then it now runs ;buff again.

Also adding in Trader Spec Luck to these.  I was considering adding a section to ;buff and common-arcana for speculates, except as far as I know this is the only one remaining, and they're moving away from speculates in favor of magic?

### DIFF
--- a/mining-buddy.lic
+++ b/mining-buddy.lic
@@ -28,6 +28,7 @@ class MiningBuddy
       end
     end
     wait_for_script_to_complete('buff', ['mining'])
+    bput('speculate luck', '^You focus your mind on the world around you', '^You are already focusing on the world around you.') if DRStats.trader? && DRStats.circle >= 65
     @areas.each { |area_name| mine_rooms(@area_list[area_name]) }
   end
 
@@ -95,7 +96,11 @@ class MiningBuddy
   def mine_rooms(rooms)
     rooms.each do |room|
       wait_for_script_to_complete('safe-room') if bleeding?
-      check_repair if mine?(room)
+      if mine?(room)
+        check_repair
+        wait_for_script_to_complete('buff', ['mining'])
+        bput('speculate luck', '^You focus your mind on the world around you', '^You are already focusing on the world around you.') if DRStats.trader? && DRStats.circle >= 65
+      end
     end
   end
 
--- a/mining-buddy.lic
+++ b/mining-buddy.lic
@@ -27,7 +27,7 @@ class MiningBuddy
         fput('stow my packet')
       end
     end
-    wait_for_script_to_complete('buff', ['mining']) if @settings.waggle_sets['mining']
+    DRCA.do_buffs(@settings, 'mining')
     bput('speculate luck', '^You focus your mind on the world around you', '^You are already focusing on the world around you.', '^Your pattern-matching skills are still exhausted') if DRStats.trader? && DRStats.circle >= 65
     @areas.each { |area_name| mine_rooms(@area_list[area_name]) }
   end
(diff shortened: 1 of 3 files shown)


Sarvatt:
can we make the wait_for_script_to_completes have an unless the waggle set exists? just running buff even without a waggle set wastes a few seconds every call, was thinking about doing that everywhere

Tarhish:
Oh yeah, you did mention that and I thought it was a good idea.

mining-buddy.lic
```diff
@@ -95,7 +96,11 @@ class MiningBuddy
   def mine_rooms(rooms)
     rooms.each do |room|
       wait_for_script_to_complete('safe-room') if bleeding?
-      check_repair if mine?(room)
+      if mine?(room)
+        check_repair
+        wait_for_script_to_complete('buff', ['mining']) if @settings.waggle_sets['mining']
+        bput('speculate luck', '^You focus your mind on the world around you', '^You are already focusing on the world around you.', '^Your pattern-matching skills are still exhausted') if DRStats.trader? && DRStats.circle >= 65
```
rpherbig: 💡 This might be a good candidate for `DRCA`

mining-buddy.lic
```diff
@@ -95,7 +96,11 @@ class MiningBuddy
   def mine_rooms(rooms)
     rooms.each do |room|
       wait_for_script_to_complete('safe-room') if bleeding?
-      check_repair if mine?(room)
+      if mine?(room)
+        check_repair
+        wait_for_script_to_complete('buff', ['mining']) if @settings.waggle_sets['mining']
```
rpherbig: 💡 Maybe another DRCA helper? Is this pattern common enough that it would be useful?

```ruby
def foo(settings, buff_set_name)
    return unless settings.waggle_sets[buff_set_name]
    wait_for_script_to_complete('buff', [buff_set_name])
end
```

mining-buddy.lic
rpherbig: I can't remember for sure if DRCA can access `DRStats`, but I'm pretty sure it can.

mining-buddy.lic
Sarvatt: it’s SUPER common in the scripts, that would make it easier converting them all

Sarvatt: wasnt there some reason not to use settings in common files though? i guess not since common-money is using it now

Tarhish: I'm game.  We'd want to use this the exact same way for every waggle call so why not

Tarhish: I was thinking about this a bit more.

Can we just migrate all of ;buff into common-arcana?

```
  def cast_waggle(settings, waggle_sets_name = 'default')
    return unless settings.waggle_sets[waggle_sets_name]
    
    spells = settings.waggle_sets[waggle_sets_name]
    return unless spells

    if DRStats.barbarian?
      start_barb_abilities(spells, settings)
    elsif DRStats.thief?
      start_khris(spells, settings)
    else
      spells.each_value do |value|
        next unless value['use_auto_mana']
        check_discern(value, settings)
      end

      cast_spells(spells, settings, settings.waggle_force_cambrinth)
    end
  end
```

That's the entirety of the script.  So in-script things would just call cast_waggle, and players could just cast ;buff.

Tarhish: edited to remove some redundancy.

rpherbig: It's still best to have a standalone `buff` script since some people use it before/after hunting. However, the entirety of the `buff` script could be a call to the new DRCA helper.

Tarhish: Yep, it seems to work

Tarhish:
Okay, I made the change to ;buff in here too, and I can still call ;buff to any waggle or without arg and get my default waggle.

PR #3961, opened 2019-10-14
Add cast_waggle function to DRCA

Tarhish:
The method takes settings and a waggle name, then casts it in an identical manner to ;buff.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -640,4 +640,23 @@ module DRCA
 
     crafting_prepare_spell(training_spells[needs_training], settings)
   end
+
+  def cast_waggle(settings, waggle_sets_name = 'default')
+    return unless settings.waggle_sets[waggle_sets_name]
+
+    spells = settings.waggle_sets[waggle_sets_name]
+
+    if DRStats.barbarian?
+      start_barb_abilities(spells, settings)
+    elsif DRStats.thief?
+      start_khris(spells, settings)
+    else
+      spells.each_value do |value|
+        next unless value['use_auto_mana']
+        check_discern(value, settings)
+      end
+
+      cast_spells(spells, settings, settings.waggle_force_cambrinth)
+    end
+  end
 end
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -641,21 +641,20 @@ module DRCA
     crafting_prepare_spell(training_spells[needs_training], settings)
   end
 
-  def cast_waggle(settings, waggle_sets_name = 'default')
-    return unless settings.waggle_sets[waggle_sets_name]
+  def do_buffs(settings, set_name)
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -640,4 +640,23 @@ module DRCA
 
     crafting_prepare_spell(training_spells[needs_training], settings)
   end
+
+  def cast_waggle(settings, waggle_sets_name = 'default')
```
rpherbig: 💡 I don't like 'waggle' and I think it's overall been more confusing than helpful.

common-arcana.lic
```diff
@@ -640,4 +640,23 @@ module DRCA
...
+    return unless settings.waggle_sets[waggle_sets_name]
+
+    spells = settings.waggle_sets[waggle_sets_name]
+
+    if DRStats.barbarian?
```
rpherbig: ❓ Did you confirm that `DRStats` works here in a common-* helper?

common-arcana.lic
rpherbig: 💡 I'm not a fan of the default value. I think `buff` should be responsible for figuring out how to interpret the user's args. Thoughts?

common-arcana.lic
```diff
@@ -640,4 +640,23 @@ module DRCA
...
+      start_barb_abilities(spells, settings)
+    elsif DRStats.thief?
+      start_khris(spells, settings)
+    else
+      spells.each_value do |value|
```
rpherbig: 💡 Could we do something like this?

```ruby
spells.values
.select { |spell| spell['use_auto_mana'] }
.each { |spell| check_discern(spell, settings) }
```

common-arcana.lic
Tarhish: I think you're right,  and new lichers have no idea what waggles are.  They sound like something irrelevant just by their name.  But if that is their name I'd like to stick to it, unless this is the time we add a transform to dependency that maps wagglesets to a different name and begin the long process of transitioning away from it?  I'd do it, but that's a bigger codebase change. 

Alternately I could make this more generic and accept any hash of spells.

common-arcana.lic
Tarhish: I think it should work -- I'll try it out.

common-arcana.lic
rpherbig: I think the only place we have to use 'waggle' is in the setting name itself, since we haven't fixed that yet.

Tarhish: Done and done

Tarhish: I got distracted by other things, but I'm back to this.  Still, I'll look at the wider scale some other time maybe.

Tarhish:
I ran through those changes to make sure they work the same way, and it's been successful.

PR #3967, opened 2019-10-19
Create safe_room.lic

cfrisby77:
For issue issue 3966

Fixes issue with using Fang Cove Healer

Added  'do anything more for you now', to line 45


Tarhish:
For what it's worth, I know some of the healers out in the lesser-traveled areas have some missing messages, so this is a great change, and it _looks_ fine at my first glance.  You've just got the file called safe_room instead of safe-room, so it thinks it's a new file.

PR #3972, opened 2019-10-21
[clean-leather] Allow careful scraping and crafting bags

Ry7n:
This allows the caller to specify the scraping speed for leather
cleaning (quick, normal, slow), and honors the `craft_item_in_container`
setting if it exists, and prevents the script from trying to pour tanning
lotion on bones.

### DIFF
--- a/clean-leather.lic
+++ b/clean-leather.lic
@@ -27,7 +27,7 @@ class CleanLeather
 
     ensure_copper_on_hand(2000, @settings)
 
-    @preservative = args.noun == 'bone' ? 'bleaching solution' : 'tanning lotion'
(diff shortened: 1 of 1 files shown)


clean-leather.lic
```diff
@@ -26,23 +27,36 @@ class CleanLeather
 
     ensure_copper_on_hand(2000, @settings)
 
-    @preservative = args.noun == 'bone' ? 'bleaching solution' : 'tanning lotion'
+    @preservative = args.noun =~ /^bones?$/i ? 'bleaching solution' : 'tanning lotion'
```
Ry7n: Uncleaned bones are usually "bones" not "bone", and using the actual noun caused the script to fail.

rpherbig: ❓ "usually"... should the 's' be optional in the regex?

PR #3978, opened 2019-10-22
Add drinfomon to buff requires

Tarhish:
Sorry for all these little changes recently. 

A thief reported seeing an error with the recent buff change that triggered because DRStats.thief? returned false.

I can see DRStats.thief? on mine, and restarting drinfomon fixed the problem for him.  I'm guessing no custom_require for drinfomon did this somehow?

### DIFF
--- a/buff.lic
+++ b/buff.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#buff
 =end
 
-custom_require.call(%w[common common-arcana events spellmonitor])
+custom_require.call(%w[common drinfomon common-arcana events spellmonitor])
 
 class Waggle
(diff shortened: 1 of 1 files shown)


rpherbig:
I don't think this change is needed. `buff` doesn't use anything from `drinfomon` and `common-arcana` already has it as a custom_require

Tarhish:
This is a weird coincidence and I'm being paranoid.  The error he got, `undefined method `each_value' for ["Focus", "Sight", "Safe", "Hasten"]:Array` couldn't have come from the new version, since we don't use each_value anymore per your earlier suggestion.  He was on an old version of _something_ one way or another, regardless of what DRStats did or did not see.

Issue #810, opened 2016-09-16
Create Holy Weapon charge script for Paladins in Training-Manager


wscherer420:
I'd think before hunting-buddy activates, that the character could check their weapon and make sure it is fully charged.  Similar to how it checks favor orbs prior to getting started.  

Here is a SF script i use to charge my holy weapon manually.

echo from outside temple
if_1 goto po broadsword
move n
move go oak door
move go widder
move go widder
pause
move go widder
move go widder
move go stair
pause
waitfor Obvious paths:
move go widder
move go widder
move go stair
pause
waitfor Obvious paths:
move go widder
move go widder
pause
move go jade arch
move go jade plat

put get pol broadsword
put look at pol broadsword
pause
put stow l
pause
put stow r
pause

put pray chadatru
echo WAITING 5 minutes
pause 120
echo Waiting 3 minutes
match sword After clearing your thoughts,
matchwait

sword:
put get pol broad
pause
put glyph renew pol broad
pause
put look pol broad

echo *
echo *
echo *
echo \* Start 1 hour timer ********
pause 5
getgoing:
echo leaving

out:
put stand
move up
move go jade arch
move go clock
move go clock
pause
move go stair
waitfor Obvious paths:
pause
move go clock
move go clock
pause
move climb stair
waitfor Obvious paths:
move go clock
move go clock
pause
move go clock
pause
move go clock
move go oak door
pause
move s
move go gate
pause
put remove badge
pause
put pray badge
pause
put wear badge
echo done

wscherer420:
or maybe better if it happens after hunting buddy in case it has drained too far and requires the 1 hour break between the 1st and 2nd glyph renewal attempts.

PR #3983, opened 2019-10-24
New stealing items

rpherbig:
I need someone to review these item additions. They don't look right to me, but I don't have a Thief right now.

For example, for the first new item in that PR, the item_in is wrong, it should be 'in cata' or 'on cata' (I assume cata means catalog...).

The second item clearly is wrong, I don't think the noun was supposed to be 'rugged' and the container 'gloves'.

Some could also be duplicates of items already in the stealing data.

### DIFF
--- a/data/base-stealing.yaml
+++ b/data/base-stealing.yaml
@@ -8623,11 +8623,3 @@
   pawnable: false
   room: 12453
   town: Muspar'i
-- acceptable_max: 1111
-  acceptable_min: 1111
-  id: '810'
-  item: ironwood zurna
-  item_in: on table
-  pawnable: false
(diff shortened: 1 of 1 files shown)


Tarhish:
Yeah, and the water is already in base-stealing in that shop.  In vat, not in backpack.  I think all the crossing ones are wrong.

Tarhish:
792 doesn't even have a town or... in fact, a name at all.

rpherbig:
I removed those.

aruthas:
id #810 needs to go, it has it mapped to room 61 that I cant find
all the muspari stuff is good

PR #3984, opened 2019-10-24
Starting Avtalia Array Support

Tarhish:
Okay, this is a biggish one this time.  We can take time for this, though I think I've worked all the kinks out already.

Trader Avtalia Array currently isn't very useful for lich and this is the start of fixing it.  I have all of this working in combat-trainer, but before that all the common stuff needs to go in.  AVTA spins mana into worn cambrinth while taking in starlight, and avtalia.lic is intended to monitor it.  avtalia.lic starts up and runs in the background, watching any cambrinth defined in avtalia_array:

```
avtalia_array:
 - name: cambrinth bracer
   cap: 32
 - name: watersilk bag
   cap: 108
 - name: cambrinth anklet
   cap: 12
 - name: tattered robe
   cap: 50
 - name: cambrinth chains
   cap: 50
```

Any time AVTA brings the camb up to max mana it generates an atmospheric message that makes avtalia.lic set its mana to full, and anytime you FOCUS one of them or do FOCUS CAMBRINTH, it updates its mana level and the last-seen time for that value.

Every 10 seconds avtalia.lic looks at the relevant flags and processes them, setting mana and time values in your UserVars.

 avtalia:  {"cambrinth bracer"=>{"mana"=>32, "cap"=>32, "time_seen"=>2019-10-23 20:12:33 -0400}, "watersilk bag"=>{"mana"=>96, "cap"=>108, "time_seen"=>2019-10-23 20:08:03 -0400}, "cambrinth anklet"=>{"mana"=>12, "cap"=>12, "time_seen"=>2019-10-23 20:12:33 -0400}, "tattered robe"=>{"mana"=>36, "cap"=>50, "time_seen"=>2019-10-23 20:08:03 -0400}, "cambrinth chains"=>{"mana"=>26, "cap"=>50, "time_seen"=>2019-10-23 20:08:03 -0400}}

DRCA.charge_avtalia can be used to add mana to one of them, and invoke_avtalia can be used to pull mana from them.  DRCA.choose_avtalia picks one of them based on mana levels, time seen, and the relative percentage of charge it's looking for, so any script can choose whether it wants a cambrinth that's mostly full or if it doesn't mind either way.  Similarly, any script can charge to and invoke from the avtalia, and call DRCA.update_cambrinth to refresh all their data.

So, in my version of combat-trainer, it'll get down to 60% mana, try to buff Last Gift of Vithwok IV with a cambrinth charge of -25 and -25.  It asks DRCA.choose_avtalia( 50, 0 ) and choose_avtalia returns the watersilk bag, which has 96 mana in it, and has been seen in the last ten minutes. It invokes 50 from it instead of charging something else.

### DIFF
--- a/avtalia.lic
+++ b/avtalia.lic
@@ -1,3 +1,7 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#avtalia
+=end
+
 custom_require.call(%w[common common-arcana])
 
 no_kill_all
--- a/avtalia.lic
+++ b/avtalia.lic
@@ -1,0 +1,34 @@
+custom_require.call(%w[common common-arcana])
+
+no_kill_all
+no_pause_all
+
+### Avtalia defines an UserVars listing items in your avtalia_array.
+### Each cambrinth has a value representing how much mana is available
+### as well as how fresh the reading is.  It can be used by other scripts as-needed
+
+
+class Avtalia
+  include DRC
+  include DRCA
+
+  def initialize
+    @settings = get_settings
+    
+    DRCA.add_avtalia_flags(@settings)
+    execute
+  end
+  
+  def execute
+    loop do
+      pause 10
+      DRCA.process_avtalia(@settings)
+    end
+  end
+end
+
+before_dying do
+  DRCA.remove_avtalia_flags(get_settings)
+end
+
+Avtalia.new
--- a/avtalia.lic
+++ b/avtalia.lic
@@ -18,21 +18,57 @@ class Avtalia
 
   def initialize
     @settings = get_settings
-    
-    DRCA.add_avtalia_flags(@settings)
+    @avtalia_array = @settings.avtalia_array
+    exit if @avtalia_array.empty?
+    add_avtalia_flags
     execute
   end
-  
+
   def execute
     loop do
       pause 10
-      DRCA.process_avtalia(@settings)
+      process_avtalia
+    end
+  end
+
+  def add_avtalia_flags
+    UserVars.avtalia = {}
+
+    @settings.avtalia_array.each do |camb|
+      first = camb['name'].split.first
+      second = camb['name'].split.last
+      short_reg = first == second ? /\b#{first}/i : /#{first}.*\b#{second}/i
+      Flags.add("avtalia-full-#{camb['name']}", /^Your.* #{short_reg} pulses brightly with Lunar energy/)
+      Flags.add("avtalia-focus-#{camb['name']}", /A.* #{short_reg}.* pulses? .+ (\d+)/)
+      UserVars.avtalia[camb['name']] = {'mana' => 0, 'cap' => camb['cap'], 'time_seen' => Time.now}
+    end
+  end
+
+  def process_avtalia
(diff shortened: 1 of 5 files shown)


avtalia.lic
```diff
@@ -0,0 +1,34 @@
+custom_require.call(%w[common common-arcana])
```
rpherbig: 💡 Please add the documentation header block

Tarhish: I'll have to put something out there to refer to later, but it's done.

rpherbig:
❓ Why put those helpers in `DRCA` (which is already pretty big) if they're only used by one script? Do you plan to use them in other scripts later?

common-arcana.lic
```diff
@@ -658,4 +658,89 @@ module DRCA
       cast_spells(spells, settings, settings.waggle_force_cambrinth)
     end
   end
+
+  def add_avtalia_flags(settings)
```
rpherbig: ❓ Why put these helpers in `DRCA` (which is already pretty big) if they're only used by one script? Do you plan to use them in other scripts later?

validate.lic
```diff
@@ -766,6 +766,20 @@ class DRYamlValidator
...
 
+  def assert_that_avtalia_array_is_valid(settings)
+    return if settings.avtalia_array.empty?
+
+    settings.avtalia_array.each do |camb|
+      first = camb['name'].split.first
+      second = camb['name'].split.last
+      short_reg = first == second ? /\b#{first}/i : /#{first}.*\b#{second}/i
+      error("avtalia_array appears to include your normal casting cambrinth.  This could break it horribly.") if short_reg =~ settings.cambrinth
+      if settings.avtalia_array.select { |camb| short_reg =~ camb['name'] }.to_a.length > 1
```
rpherbig: ❓ At a glance, it looks like this would always trigger because it would find its own entry. Can you confirm this does not trigger in the unique case and does trigger in the duplicate name case?

common-arcana.lic
Tarhish: You're right.  After building a few things with this, the flag add, remove, and process methods will only be used in avtalia.lic and nothing else should call them.  The other three will need to stay in DRCA to be called by others as needed.  I'll change them around.

validate.lic
Tarhish: Yes, you're right, except for the .to_a.length > 1.  My entries or a blank entry generates no error, but adding an entry for, say, 'tatt rob' generates an error re: uniqueness conflict with 'tattered robe.'

Tarhish:
Nobody else I've talked to have had any changes to suggest at this stage -- all that would come later.  At some point I will want to adjust invoke_avtalia to catch actual mana invoked, since that can vary at lower levels of Arcana, but that could also be changed once live.  

Since this is going to be built to have no impact to anyone not running avtalia.lic, and right now none of these changes are even used in any script, I'd say for now this is safe from my POV.

One pleasant upside discovered during HE is that with Enrichment running I literally have more mana rushing into the avtalia cambrinth than I can use.

Tarhish:
Looks like a conflict in base-empty arose in the interim.  I'll reopen this when I get home.

PR #3992, opened 2019-10-26
initial brocket deer commits. 

jayRyan24:
Adds the bescort methods and base-hunting data.

I put it into three separate methods because we need to know where we start within the land to know which way to go first, else we could be wandering west or east. I did this to accommodate the already implemented way things are handled. 

https://pastebin.com/eN7Jx4WT

Need testers.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -284,7 +284,8 @@ class Bescort
   private
 
   def brocket_1(mode)
-    move 'climb fence' if mode.include?('enter') 
+    move 'climb fence' if mode.include?('enter')
+
     if mode.include?('enter')
       find_room_list(%w[w w w])
     else
--- a/bescort.lic
+++ b/bescort.lic
@@ -201,15 +201,15 @@ class Bescort
         { name: 'hvaral_passport', regex: /hvaral_passport/i, description: 'Handles the gate between Hvaral and Muspari' }
       ],
       [
-        { name: 'brocket_1', regex: /brocket_1/i, description: 'Enter the young brocket deer hunting area.' },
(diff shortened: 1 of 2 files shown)


jayRyan24:
Just had a successful full hunt in youngs.

End of it:  https://pastebin.com/EWaY0vJ6

bescort.lic
```diff
@@ -199,14 +199,32 @@ class Bescort
       ],
       [
         { name: 'hvaral_passport', regex: /hvaral_passport/i, description: 'Handles the gate between Hvaral and Muspari' }
+      ],
+      [
+        { name: 'brocket_1', regex: /brocket_1/i, description: 'Enter the young brocket deer hunting area.' },
```
rpherbig: ❓ What do you think about `young_brocket`, etc. here instead of 1/2/3?

bescort.lic
```diff
@@ -265,6 +283,51 @@ class Bescort
 
   private
 
+  def brocket_1(mode)
+    move 'climb fence' if mode.include?('enter')
```
rpherbig: ❓ Should this do a check that you're in the right room to start?

bescort.lic
jayRyan24: I was trying to match the rest of the hunting areas and how we do them.

bescort.lic
jayRyan24: I can add that, sure.

bescort.lic
jayRyan24: lol, we actually have a mix of it. I changed it to young, mid, and elder.

Sarvatt:
this looks good to me, thanks for doing his long requested hunting area

PR #3997, opened 2019-11-02
dragon's breath won't go up

Hiinky:
I don't' know what heavy: true does but its only when I remove it that ;buff  or ;combat will maintain it.  for mechanical purposes DB is considered a buff.  Tested many ways.  this is the only way it works.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -795,8 +795,9 @@ spell_data:
     mana_type: holy
   Dragon's Breath:
     skill: Targeted Magic
-    heavy: true
-    mana: 15
+    abbrev: DB
+    mana: 30
+    mana_type: elemental
   Drums of the Snake:
     skill: Augmentation
     mana: 15
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -796,7 +796,7 @@ spell_data:
   Dragon's Breath:
     skill: Targeted Magic
     abbrev: DB
-    mana: 30
+    mana: 15
     mana_type: elemental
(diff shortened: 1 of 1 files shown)


Sarvatt:
the only time it shouldnt recast is when there no mobs in the room as far as i can see

Hiinky:
so heavy: true might be causing some problems but now DB is up and BG won't fire.  its making me bonkers.

rpherbig:
❌ The `heavy: true` is because DB is a [heavy offensive spell](https://elanthipedia.play.net/Category:Heavy_offensive_abilities) and needs to remain

I don't know why DB and BG won't be up at the same time. Have you tried having both up at the same time by manually casting them?

Hiinky:
I'm trying to put it up with ;buffs and maintain it up at all times.  With heavy: true only casting with mobs....   Mechanically the spell itself its works like a buff such as Heal or MAF.  It goes up with a duration and just needs to be recast: 1.  It is the SPIT/EXHALE action that actually triggers the TM offensive and can "use up" the spell.  There isn't really a reason that you should wait for mobs to CAST the spell, only wait for multiples to SPIT/EXHALE at them.

Manually BG and DB go up fine.  But turns out it was all a moot point anyway. I've since figured out it is the lack of a recast: timer that was stopping the works so I forced recast: 1 in the the specific yaml and it's working as I had hoped.

Tarhish:
Sounds like it might be a good idea to add a recast: to it in base, if that's what it needs to work, but I don't know WMs.

PR #4000, opened 2019-11-02
Find darkbox update

Dartellum:


Tarhish:
I like it.  Gives me another idea for later, too.

PR #4005, opened 2019-11-10
Add a cornmaze script

Tarhish:
Adding a cornmaze script stolen from task-loot and task-noloot.  We should be able to control it better here.  This behaves almost exactly as the old versions did except with a bit of failure-proofing.

1.) If you accidentally enter the maze with an external task, this will cancel it instead of getting stuck forever in a loop. (tested)
2.) args for noloot (search but don't loot non-pass incidentals) and 'restart' (don't search at all - tested)
3.) adds support for cornmaze_junk (tested)
4.) Will STOW incidentals as normal, but if you define cornmaze_containers: then it'll cycle through them when one gets full, or it fails to stuff something in there (tested this).
5.) Disabled fast movement for genie, which should theoretically let them use this.
6.) Some fputs converted to bputs.  I couldn't get them all.
7.) It's been turned into a class with actual includes and all that jazz.

I've run this through 4 maze loops so far with success.


corn-maze.lic
```diff
@@ -0,0 +1,632 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#corn-maze
+  This script runs non-combat tasks for sleeping dragon maze. 
+=end
+
+custom_require.call(%w(common drinfomon textsubs common-travel events common-items))
```
Sarvatt: is textsubs needed here?

Tarhish: why the hell?

Uh, I'm not sure.  All I can say is that is was in task-loot originally.  Why though?

Tarhish: I can't think of a single reason that should be there.  I'll unload textsubs and run it again tomorrow though, before I say definitively.

Sarvatt: yeah there's no reason, it's safe to remove. just using this excuse to actually look over the script now that you're submitting it and noticing anything stupid that was already there :D

corn-maze.lic
```diff
@@ -0,0 +1,632 @@
...
+    @done = false
+    find_start
+    @done = false
+    wait_for_command
+  end
+
+  def find_start
+    wait_on_done_move
+    unless XMLData.room_title.include?('Labyrinth') || XMLData.room_title.include?('Corn Maze, Entrance')
+      DRC.message("The script can't find its way back from here!  Make your way back towards the northwest Labyrinth using elanthipedia's Corn Maze entry, such as https://elanthipedia.play.net/Sleeping_Dragon_Corn_Maze_432 and then unpause the script!")
```
Sarvatt: What would you think about saying starting narost right clicking and picking the corn maze map here instead?

Tarhish: Yeah, I didn't know that was an option, but it's a lot easier.

corn-maze.lic
```diff
@@ -0,0 +1,632 @@
...
+  Flags.delete('weeds')
+  Flags.delete('traps')
+  Flags.delete('grasshoppers')
+  Flags.delete('scream')
+  Flags.delete('landmarks')
+  Flags.delete('tokens')
+  Flags.delete('corn')
+  Flags.delete('halflings')
+  Flags.delete('loot')
+end
```
Sarvatt: maybe add maze_done and Task_done, poisoned, and already_built here also, it's pedantic because they'd only be left over if you killed the script but deleting anything that has a Flags.add is the goal.

Sarvatt:
Looks good to me! Thanks for fixing this up

corn-maze.lic
```diff
@@ -0,0 +1,632 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#corn-maze
+  This script runs non-combat tasks for sleeping dragon maze.
```
rpherbig: 💡 Maybe 'This script is intended to replace loot-corn-maze.lic`?

corn-maze.lic
```diff
@@ -0,0 +1,632 @@
...
+      if !Flags['Task_done']
+        case DRC.bput('search', 'You don\'t think there are any traps here', 'tiny black dots are jumping around inside', 'You notice', 'already looked for traps around here')
+        when 'You notice'
+          waitrt?
+          fput 'disarm trap'
+          while checkstunned || checkwebbed
+            pause 1
+          end
+          waitrt?
+          while !checkstanding
```
rpherbig: 💡 `DRC.fix_standing`

corn-maze.lic
```diff
@@ -0,0 +1,632 @@
...
+
+  def grasshopper_task
+    @mice_path.each do |x|
+      exit if Flags['maze_done']
+      if !Flags['Task_done']
+        fput 'search'
+        pause 1
+        waitrt?
+        DRC.fix_standing
+        custom_move "#{x}"
```
rpherbig: 💡 I'm pretty sure all of these `custom_move` calls don't need the string interpolation. `x` should already be a string

corn-maze.lic
```diff
@@ -0,0 +1,632 @@
...
+      else
+        custom_move("#{x}",true)
+      end
+    end
+    finish_task
+  end
+
+  def touch_room
+    targets = DRRoom.room_objs.grep(/rake|stones|hay|wood|basket|boulder|wheel|spiderweb|straw|statue|pail|fence|wagon|burrow|hut|scarecrow/)
```
rpherbig: 💡 This would be good to extract out to the top of the file into a named var (like the directions)

corn-maze.lic
```diff
@@ -0,0 +1,632 @@
...
+      exits = find_exits.match(/^Obvious paths: (.*)\./).to_a[1].split(', ')
+      if (exits.include? 'north')
+        custom_move 'north'
+      elsif (exits.include? 'east')
+        custom_move 'east'
+      elsif (exits.include? 'south')
+        custom_move 'south'
+        custom_move 'south'
+        custom_move 'east'
+      elsif exits.include?('west') && !exits.include?('southwest') && !exits.include?('east') && !exits.include?('south') && !exits.include?('north') && !exits.include?('southeast') && !exits.include?('northwest') && !exits.include?('northeast')
```
rpherbig: 💡 You could use intersection to see if any elements in one array are in another:
```ruby
shared = [1,2,3,'A'] & (1..9).to_a
shared     # => [1, 2, 3]
```

In our case, `(exits & ['west', 'southwest', ...]).empty?`. Do you think that is any clearer?

Or `elsif ['west', 'southwest', ...].none? { |dir| exits.include?(dir) }`
```

File: corn-maze.lic [hunk 9]
```diff
@@ -0,0 +1,632 @@
...
+    Flags.reset('corn')
+    Flags.reset('halflings')
+    Flags.reset('loot')
+  end
+
+  def custom_move(dir, fast=false)
+    if fast && $frontend != 'genie'
+      put dir
+      @current_type_ahead += 1
+      if @type_ahead_size < @current_type_ahead
```
rpherbig: 💡 Why is this conditional in here if `@current_type_ahead` always starts at 0 and `@type_ahead_size` is always 1?

profiles/base-empty.yaml
```diff
@@ -72,3 +72,4 @@ empty_values:
   smoke: {}
   divination_tool: {}
   divination_bones_storage: {}
+  cornmaze_containers: {}
```
rpherbig: 💡 Please add `maze_junk` too

rpherbig:
New script, can be merged as-is. @Sarvatt were any of your comments blocking?

profiles/base.yaml
```diff
@@ -1037,6 +1037,31 @@ boggle_junk:
...
+- farmer hat
+- armure tunic
+- pipe
+- blossom
+- muffin
+- brocade pouch
+- socks
+
+# Define a list of containers like you would for storage_containers.  All the bags you could shove stuff into for the cornmaze 
+cornmaze_containers:
```
Dartellum: Would you remove this from base as you have it in base-empty?

Sarvatt:
none of my comments were blocking, no

corn-maze.lic
Sarvatt: loot-corn-maze is just the looting part ripped out of this script, its still useful if you!re manually running around the maze

corn-maze.lic
Tarhish: Looks like once upon a time they intended to make that adjust itself.  I don't think we'd ever need it to be more than 1 extra though, so I suppose the extra steps could be eliminated

corn-maze.lic
Tarhish: It's really not.  It's just the same thing in fput instead of bput, and it spams it needlessly.  I'll remove that.

corn-maze.lic
Tarhish: Okay, I never saw loot-corn-maze somehow.  It looks useful, but also like it's totally obsoleted by this... although it'd be pretty trivial to add a 'manual' arg to this that does the same thing.

I guess the question is, should this replace loot-corn-maze entirely?  I like corn-maze.lic more than loot-corn-maze, but that's just like my opinion, man.

corn-maze.lic
Dartellum: Sarvatt, in the before_dying section or in def reset_flags? Or both?

Tarhish:
Okay, these are the last feature changes/reorgs I'm doing.  A number of people learned I was doing this and offered problems they had had, mostly related to getting desynced in the maze.

The pickup off the ground feature didn't work right, and I decided to scrap it for now.  It can go in later.  The checkstunned stuff in the original script was totally useless -- we were able to confirm the maze stuns/webs don't trigger the XML icon data, which is why they were spamming STAND, but some people had their script moving on because they ended up stunned while standing.  I replaced the checkstunned stuff with a method that tries to stand until it gets the 'You are already standing' message, which it'll only do if it isn't stunned/webbed.  This will probably make fix_standing redundant, but I trust fix_standing more.  No harm to keeping it there.

Also, custom_move will DRC.retreat if it's been > 8 seconds since entering the room.  DRC.retreat only triggers if there's an npc around, so it only retreats in certain situations.  I tested this and was happy with it.

Added a manual arg, which loots the rooms as you walk through them.  I ran through the maze and was happy with it too.  Also added profanity to the list of good FEs.

Beyond that, just addressing various comments.

Don't quite pull this yet -- I wanted to test the stunned change, but I have to go back into test.  I went an entire 14 tasks without the poke task or disarm task.

Sarvatt:
maybe upload to ;repos in the interim for more testing?

Tarhish:
I'm removing loot-corn-maze.  This does everything it did.

Also, I added a 'slow' arg for the people who just can't handle the typeahead even with running stormfront.

Also also, I'm running this on Test over and over again and I STILL haven't gotten a poke task or tripped a trap.  Like 50-80 tasks in and nothing.  This is getting silly.

Tarhish:
Alright, whining about it worked, and I got what I was looking for minutes later.  Aside from a missed 'You stand back up' bput catch it did manage the stuns just fine.  So this should now be a working script.

corn-maze.lic
Tarhish: I'll look into changing how it was done there.

profiles/base.yaml
Sarvatt: nope it should still be here too

PR #4011, opened 2019-11-12
Almanac nouns in combat-trainer

Hiinky:


Tarhish:
I think the previous version had the almanac setting in base.yaml, yeah?

Hiinky:
Its three separate now.  I gave up on being consolidated and perfect

Tarhish:
You might be handling this already then, but if this alone is pulled then no one besides you will have almanac: defined so they'll all crash out when they try to use the almanacs in combat-trainer

rpherbig:
In the future please try to keep like changes packed up.

Hiinky:
I so tried to make it one.  I need github desktop lessons....

updated for almanac_noun

PR #4015, opened 2019-11-12
Resubmit Avtalia array script

Tarhish:
Resubmitting due to base-empty conflict that emerged.  See PR https://github.com/rpherbig/dr-scripts/pull/3984


rpherbig:
Sorry for the delay, not sure how I missed seeing this

PR #4018, opened 2019-11-13
modified combat-trainer to allow astrology without telescope

andymanak:
Use existing setting for have_telescope within combat_trainer to determine if telescope should be used or not when training Astrology.


andymanak:
Not sure why it lists the older commit related to cambrinth.

PR #4030, opened 2019-11-18
Update base-hunting.yaml

n1neb4ll:
Blood Wolves west of crossing near dire bears. Don't hit me as often as buccas, and train nicely.

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -553,6 +553,13 @@ hunting_zones:
   - 14123
   - 19225
   - 14124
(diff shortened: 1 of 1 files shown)


Hiinky:
Might need to add a warning that they can sometimes see a wandering silver-back bear which are a bit tougher.  https://elanthipedia.play.net/Silver-backed_bear

n1neb4ll:
I don't think they will encounter them in this area. You would have to go up again from the last room, to run into the Dire-bears and silver-backs. If you stay "down" in these few rooms, the bears won't come this way. (as far as I have seen).

PR #4036, opened 2019-11-24
Tarantula script.

Dartellum:


profiles/base.yaml
```diff
@@ -1144,6 +1144,12 @@ grave_junk:
 - millipede
 - cockroach
 
+#HE tarantula gift, list the skills you want the tarantula to use
+spider_noun: bio tarantula
+crafting_skills:
```
rpherbig: ❓ `*_skills` is really generic. What do you think about `spider_*_skills`?

tarantula.lic
```diff
@@ -0,0 +1,52 @@
...
+  crafting_skills += settings.crafting_skills
+  hunting_skills += settings.hunting_skills
+  outdoor_skills += settings.outdoor_skills
+
+  loop do
+    if Script.running?('workorders') || Script.running?('craft')
+      crafting_skills.each { |e| DRC.bput("turn my bio tarantula to #{e}", /You fiddle/, /With a/)
+      case DRC.bput("rub my #{spider_noun}", /The tarantula comes alive/, /It is mere/, /you currently aren\'t/, "It needs approximately \d+ (?:roisan|roisaen)", /You try, but it does nothing./)
+      when /It needs approximately (\d+) (?:roisan|roisaen)/
+        ptime = Regexp.last_match(1).to_i
```
rpherbig: Could you make this (and the next) variable names clearer?

Dartellum: Sure - ptime = pause time.

profiles/base.yaml
Dartellum: Sure can. I am testing some other pieces and why I have lag.

DryIce41:
Thank you for putting this together Dart.

I set it up and tried it a bit. 
During hunting-buddy it would occasionally use my set noun of Seven-legged tarantula. And sometimes it would try to use "bio tarantula"  

Also you set it up so you can choose scripts that will pause it. Athletics for example.

Dartellum:
> Thank you for putting this together Dart.
> 
> I set it up and tried it a bit.
> During hunting-buddy it would occasionally use my set noun of Seven-legged tarantula. And sometimes it would try to use "bio tarantula"
> 
> Also you set it up so you can choose scripts that will pause it. Athletics for example. I called that as a before.

I had not changed a piece of code for the noun. I have that staged. As for the pausing, I did not add anything to pause other scripts, unless you are asking for that.

DryIce41:
Not asking it to pause scripts, but something similar to walkingastro. Where you can designate scripts for it to not be used during.  Does it only cycle between the first two variable skills you tell it to?

Dartellum:
Closing as another version is getting put up.

PR #4046, opened 2019-11-28
Two more chipper ammo

Hiinky:


Tarhish:
This failed, but it looks like nothing is wrong, just another bput test timeout error when you look at the test details.  Resubmitting it should fix the issue.  Wish I knew how to stop it from doing this.

PR #4049, opened 2019-11-28
Added before to Absolution to drop GoL.

Dartellum:


rpherbig:
Should GOL have a similar before to drop Absolution?

PR #4055, opened 2019-12-01
Mitigate peer tag crashes

Tarhish:
Infrequently (~5-10% of the time), when pulling a roomnumber for a room with a peer tag, lich will suffer a fatal hang and eventually crash.  I did a great deal of hunting for the bug and failed miserably, so instead this PR does 3 things.

1.) Gates afk's room id check by seeing the room title, which prevents it from triggering a roomnumber check when walking into every room.

2.) go2 kills roomnumbers if the calculated path includes a room with a peer tag in it.  Normally it will restart on go2 exit, but if it ends directly in a peer tag room then it'll just leave it stopped.

3.) Moves the textsub kill to after estimated destination time has been calculated, and only kill it if it'll be > 5 seconds.  This isn't directly related to the issue, just adjacent to it.

This is worth sitting on and testing for awhile, but it shouldn't add any new dependencies to go2.

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -75,7 +75,7 @@ loop do
     stop_script('go2') if Script.running?('go2')
   end
 
-  if Room.current.id == 9610
+  if XMLData.room_title.include?('Skeletal Claw') && Room.current.id == 9610
     DRC.fix_standing
     move('out')
   end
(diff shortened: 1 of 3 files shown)


Tarhish:
Phew, I guessed correctly about the test failure.

rpherbig:
✔️ Looks good to me. Let me know when you feel it's been sufficiently tested.

Tarhish:
I've had a couple other people try this, as well as a couple of my own characters, and surprisingly haven't run into anything weird or any slowdowns.  As far as I'm concerned this is good to go.

PR #4060, opened 2019-12-02
Avtalia Phase 2 - combat-trainer

Tarhish:
Again, biggish change so I'll take my time and get some other people to test it, but this is the combat-trainer changes for traders.

First, it adds the option to use the tessera to train trading after the recent trader changes.  Seems to give somewhere between 5-7 Trading exp depending, and it's a good way to use up aura in a final backhunt.  Useful for people who aren't running caravans?

Second, anyone with avtalia.lic running and an avtalia_array defined will start pulling spare mana from their cambrinth for spells and cyclics (cyclics only if they have dedicated_camb_use: spell set)

Every five minutes it'll do a focus cambrinth to update the avtalia_array.  If it has spare mana to use and you're getting low on mana, it'll pull it from the array first and only charge what's left if needed.  It also doesn't mess with the breaks in the cambrinth charges.

In other words if your spell is
```
cambrinth:
- 15
- 15
- 15
```
And you have a cambrinth chain in the array with 50 free mana, it'll INVOKE CAMBRINTH CHAIN 45 and won't charge anything.

If you have 15 free mana in the array, it'll INVOKE CAMBRINTH CHAIN 15, CHARGE NORMAL CAMBRINTH 15, CHARGE NORMAL CAMBRINTH 15, INVOKE NORMAL CAMBRINTH 30.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1933,7 +1933,7 @@ class TrainerProcess
                    'Stealth' => 'Stealth', 'Ambush Stun' => settings.stun_skill, 'Ambush Choke' => 'Debilitation', 'App Pouch' => 'Appraisal',
                    'Favor Orb' => 'Anything', 'Charged Maneuver' => 'Expertise', 'Meraud' => 'Theurgy', 'Recall' => 'Scholarship',
                    'Barb Research Warding' => 'Warding', 'Barb Research Augmentation' => 'Augmentation',
-                   'Barb Research Debilitation' => 'Debilitation', 'Collect' => 'Outdoorsmanship', 'Smite' => 'Conviction', 'Locks' => 'Locksmithing', 'Tessera' => 'Trading' }
+                   'Barb Research Debilitation' => 'Debilitation', 'Collect' => 'Outdoorsmanship', 'Smite' => 'Conviction', 'Locks' => 'Locksmithing' }
 
     @skill_map['Hunt'] = %w[Perception Scouting] if DRStats.ranger?
     @favor_god = settings.favor_god
@@ -2115,8 +2115,6 @@ class TrainerProcess
       else
         wait_for_script_to_complete('heal-remedy', %w[quick nohands])
       end
-    when 'Tessera'
-      invoke_tessera(game_state)
     end
     waitrt?
     false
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2466,6 +2486,19 @@ class TrainerProcess
...
     end
   end
+  
+  def invoke_tessera(game_state)
+    return unless DRStats.trader?
+    return if game_state.retreating?
+    return if game_state.loaded || game_state.offhand?
+    DRC.retreat
+    return unless /inside your (.*)/ =~ DRC.bput("get my tessera", 'You get a .+ tessera from inside your (.*).', 'What were you')
+    container = Regexp.last_match(1).split
```
rpherbig: 💡 Would `DRC.get_noun` simplify the `Regexp.last_match(1).split.last` logic?

combat-trainer.lic
```diff
@@ -3511,18 +3552,64 @@ class GameState
       return false
     end
 
-    find_cambrinth(@cambrinth, @stored_cambrinth, @cambrinth_cap)
-
     next_charge = @charges.pop
     echo("check_charging?: #{next_charge}") if $debug_mode_ct
```
rpherbig: ❓ I'm not a fan of assigning `next_charge`, echoing its value, then possibly changing it. What do you think about moving the echo or having a second echo?

rpherbig: Alternatively, could `@charges.pop` be moved into `avtalia_charge`? I see a similar `.pop` on line 3579 with the recursive call.

combat-trainer.lic
```diff
@@ -3511,18 +3552,64 @@ class GameState
...
   end
 
+  def avtalia_charge(charge_num) # sources mana from filled avtalia cambrinth, then returns the amount left uncharged
+    return charge_num unless DRStats.trader?
+    return charge_num unless @cambrinth_invoke_exact_amount
+    return charge_num if @avtalia_array.empty?
+    return charge_num unless DRStats.mana < 80
+
+    remaining_charge = @charges.inject(0, :+) + charge_num
+    @avtalia_cambrinth = find_avtalia(remaining_charge) if @avtalia_cambrinth.empty?
```
rpherbig: ❓ Why is `@avtalia_cambrinth` stored in an instance var? Couldn't it change each time we need to charge? And it's cleared out (to `{}`) most of the time.

combat-trainer.lic
Tarhish: Can definitely do the former or something like it to make sure the debug remains consistent.  I'd rather avtalia_charge? continue to do absolutely nothing for anyone who isn't using it, just for the sake of sanity.

combat-trainer.lic
Tarhish: Yep,  A relic from when I had all the common logic in combat-trainer.  I thought it'd be sitting around doing more in gamestate, but it turned out not to be necessary.

combat-trainer.lic
Tarhish: I've kept the echo for check_charging? to mean 'how much is check_charging offloading onto your cambrinth?'  If avtalia_charge does anything on its own it reports how much it charged to what avtalia_cambrinth.  Anything left for check_charging will get echoed with the value it's actually using.

combat-trainer.lic
Tarhish: I remembered why it's stored as an instance variable.  avtalia_charge is a little bit recursive as it unwraps bites of the @charges arrays, and I keep it instanced so that I can avoid calling DRCA.choose_avtalia if avtalia_cambrinth is already set and potentially changing ships mid-stream.  See:
```
    # unwrap the charges for the spell. Keep pulling in more cambrinth mana entries until it either hits the camb cap or gets all of it.
    return avtalia_charge(charge_num + @charges.pop) if charge_num < [@avtalia_cambrinth.last['mana'], remaining_charge].min
```

I can still do it another way, but there's value to this if I want anything else to be aware that I've got a lien on that piece of jewelry while I'm prepping to invoke it.  I need to think about that to see if that's a real situation I'd ever run into, or just fantasy.

Issue #4062, opened 2019-12-02
Increase efficiency of athletics.lic


vtcifer:
For climbing targets that support climbing over and back, athletics should climb them before practice.  This should decrease the time spent practicing, and increase experience.

MahtraDR:
This is a very old issue now. Closing. Would be happy to revisit this request as fresh if it comes up again.

PR #4064, opened 2019-12-02
athletics.lic add support for more efficient athletics learning #4062 

vtcifer:
Updated athletics.lic and base-athletics.yaml for climbing over and back prior to starting climb practice. https://github.com/rpherbig/dr-scripts/issues/4062

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -77,7 +77,7 @@ class Athletics
 
   def override_location_and_practice(place)
     @climbing_target = get_data('athletics').practice_options[place]
-    climb_practice(@climbing_target['id'], @climbing_target['target'], @climbing_target['target2'], @climbing_target['hide'])
+    climb_practice(@climbing_target['id'], @climbing_target['target'], @climbing_target['hide'])
   end
 
   def swim_loop(rooms)
@@ -236,7 +236,20 @@ class Athletics
   end
 
   def hib_athletics
-    swim_liirewsag
+    until done_training?
+    swim_archers_ford
+=begin
+      if UserVars.athletics <= 45
+        echo 'Nothing to train below 45 ranks in athletics'
+      elsif UserVars.athletics <= 80
+        swim_archers_ford
+      elsif UserVars.athletics <= 150
+        override_location_and_practice('boar_peregan_tree')
+      elsif UserVars.athletics
+        swim_liirewsag
+      end
+=end
+    end
   end
 
   def ratha_athletics
@@ -286,7 +299,11 @@ class Athletics
     waitrt?
   end
 
-  def climb_practice(room, target, target2, to_hide = false)
+  def swim_archers_ford
+    swim_loop(get_data('athletics').swimming_options['archers_ford']['rooms']) until done_training?
+  end
+
+  def climb_practice(room, target, to_hide = false)
     return unless target
     walk_to(room)
     until done_training?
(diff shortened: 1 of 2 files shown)


Tarhish:
I love the love for Ratha, and this looks reasonable.  Looking at this, though, makes me think we could probably just add the extra lines to climb_practice and skip adding a new climb_practice2, like up above you'd call

```
climb_practice(@climbing_target['id'], @climbing_target['target'], @climbing_target['target2'], @climbing_target['hide'])
```

And then you'd add the argument to climb_practice, ` def climb_practice(room, target, target2, to_hide = false)` so that down below it would have:
```
move("climb #{target}") if target2
move("climb #{target2}") if target2
```

If there is a target2 then it'll get sent along to climb_practice, but if not then it'll remain nil so it won't do the moves.  Do you think that'd work?

I'll gladly try this out though.

vtcifer:
Not a ruby expert by any stretch, so I'm not 100% on what works/what doesn't.  Most of what I'll be contributing will be filtered through learning from stackoverflow. My original thought on this one was to preserve backwards compatibility the only way I knew.  I can give this a shot and test.

Tarhish:
If it's a problem we can look at condensing it later, too. I'd be willing to help with that.

vtcifer:
Tested OK.  
I will add the strong caveat that the target2 option should only be added when you can make the climb both directions, so careful updates to base-athletics.yaml will be required to make use of this.

athletics.lic
```diff
@@ -290,6 +302,8 @@ class Athletics
         start_script('performance', ['noclean']) unless Script.running?('performance')
         pause @performance_pause # make sure you are playing before starting climb practice
       end
+      move("climb #{target}") if target2
```
jayRyan24: Possibly make this an array and just `move climb` for each if this is the best solution.

athletics.lic
```diff
@@ -47,7 +47,7 @@ class Athletics
     elsif @swimming_target
       swim_loop(@swimming_target['rooms'])
     elsif @climbing_target
-      climb_practice(@climbing_target['id'], @climbing_target['target'], @climbing_target['hide'])
+      climb_practice(@climbing_target['id'], @climbing_target['target'], @climbing_target['target2'], @climbing_target['hide'])
```
jayRyan24: If anything here, I'd consider making `target` an array and then doing the above for each of them. A good thing to consider is why this is being done. Is it stopping too early I take it? So, it climbs for 15 ranks but not 30? Something like that?

jayRyan24:
Thanks for contributing to the system! It really does still need a lot of work! 

For this review, overall this is very kludge, and maybe not the best solution? I guess the main question is what is the issue? And then we should decide on the best way of fixing it, not just doing things multiple times within the code.

vtcifer:
> I guess the main question is what is the issue?

This is the issue:
```
[Script finished (In 720.0249101 seconds): climb.cmd(19)]
[Script finished (In 722.0445302 seconds): climb.cmd(19)]
[Script finished (In 579.6748853 seconds): climb2.cmd(106)]
```

Climb JUST does climb practice. (I ran it twice, on each side of a climb to validate there wasn't a difference depending on direction).
Climb2 climbs over and back, then does climb practice until finished, then repeats. 
Climbing over an obstacle is essentially instant, and provides ~ 1/8 of the bits of a full climb practice, in near 0 time.  Climbing over and back provides about ~1/4 the bits of a full climb practice in near 0 time. 
By doing this, you gain an approx 20% efficiency. 

>And then we should decide on the best way of fixing it, not just doing things multiple times within the code.

I'm not sure what is being talked about here.  The double move? The tracking of multiple objects (not all climbs use the same object for back and forth)?

As I pointed out in my original reply, not a ruby expert by any stretch so more than happy to know what would work better with ruby.

PR #4073, opened 2019-12-05
Updating Starcrash prep type to PREPARE

izawwlgood:
Starcrash is not TARGETed, but PREPAREd.


Tarhish:
Bput test has failed again for some reason so this will probably need to be  resubmitted, but this is a valid change. STARCRASH cannot be targeted.

PR #4074, opened 2019-12-06
remedy.lic - fixed some issues and added comments

dakre18:
Continue arg will now get the pestle/mixing stick before starting.
Added match message to handle herb containers.
Changed fput to get_item, in case stow_item ties it.
Added comments to clarify some parts of the script.

### DIFF
--- a/remedy.lic
+++ b/remedy.lic
@@ -122,6 +122,7 @@ class Remedy
       @verb = 'mix'
     end
 
+    get_item(@pestle) if args.continue
     work("analyze my #{@noun}") if args.continue
 
     create_item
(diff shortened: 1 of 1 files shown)


Sarvatt:
using get_item instead is fine, but it wont untie items. that's get_crafting_item

dakre18:
get_item calls get_crafting_item in remedy.lic. it works slightly different depending on some conditions.

remedy.lic
```diff
@@ -187,17 +188,19 @@ class Remedy
...
+          # stow then get, to prevent getting herbs from container (i.e. mortar)
           stow_item(@container)
           bput("get my #{@herb1}", 'You get', 'What were you referring to?', 'You are already holding that.') # using bput instead of get_item to keep script running when it cannot get the herb
-          fput("get my #{@container}")
+          get_item(@container)
           bput("lower #{@container} ground", 'You lower')
           bput("get #{@herb1} from #{@container}", 'You get', 'What were you referring to?')
           if bput("combine my #{@herb1}", 'You combine', 'too large to add', 'You must be holding both') == 'You must be holding both'
-            @count = 25
+            @count = 25 # 25 used to break loops
```
rpherbig: 💡 If this is really just used to break the loop, then `break` is the better way to do this

dakre18: There are 2 separate loops using count, to handle some inventory issues. I think you have a point, so i looked at it again. At that point, it's meant to break both loops, because there are no more herbs to get.

I think i can optimize that section more, and replace that line with a break (i have to make a few changes for that to work). I've only programmed for personal projects, never professionally, so I still have plenty to learn.

PR #4075, opened 2019-12-06
Create tarantula.lic

BinuDR:
Tarantula script!  Yaml settings below:

This one is only necessary if you want to set it to true.  The script defaults to false.
`tarantula_debug: false`

This one is only necessary if you got yours altered to be something other than a tarantula, or you want it to use something other than the full name "biomechanical tarantula" when using it.
`tarantula_noun: tarantula`

This one is to tell the script which skills to use per skillset, in and out of combat.  You can specify the same skill both in combat and non_combat.
```
tarantula:
  Armor:
    combat:
    - Chain Armor
    - Defending
    - Shield Usage
  Weapon:
    combat:
    - Melee Mastery
    - Missile Mastery
  Magic:
    combat:
    - Utility
    non_combat:
    - Sorcery
  Survival:
    combat:
    - Evasion
    - Skinning
    non_combat:
    - Athletics
  Lore:
    non_combat:
    - Appraisal
    - Performance
    - Scholarship
```

This is for skills that you NEVER want the tarantula to use.
The script will try other skills if nothing is available from the ones you specify in the above setting.
```
tarantula_excluded_skills:
  - Forging
  - Outfitting
  - Alchemy
  - Enchanting
  - Engineering
```

### DIFF
--- a/tarantula.lic
+++ b/tarantula.lic
@@ -1,0 +1,140 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#tarantula
+=end
+
+custom_require.call(%w(common))
+
+class Tarantula
+  include DRC
+  
+  def initialize
+    settings = get_settings
+    @tarantula = settings.tarantula
+    @tarantula_noun = settings.tarantula_noun.nil? ? 'biomechanical tarantula' : settings.tarantula_noun
+    @exclude = settings.tarantula_excluded_skills
+    UserVars.tarantula_last_use ||= Time.now - 600
+    @last = UserVars.tarantula_last_skillset
+    @last ||= check_last
+    @no_use_scripts = settings.tarantula_no_use_scripts
+    @debug = settings.tarantula_debug.nil? ? false : settings.tarantula_debug
+    @all_skills = { 
+      "Armor" => ['Shield Usage', 'Light Armor', 'Chain Armor', 'Brigandine', 'Plate Armor', 'Defending', 'Conviction'],
+      "Weapon" => ['Parry Ability', 'Small Edged', 'Large Edged', 'Twohanded Edged', 'Small Blunt', 'Large Blunt', 'Twohanded Blunt', 'Slings', 'Bow', 'Crossbow', 'Staves', 'Polearms', 'Light Thrown', 'Heavy Thrown', 'Brawling', 'Offhand Weapon', 'Melee Mastery', 'Missile Mastery', 'Expertise'],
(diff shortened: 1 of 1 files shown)


tarantula.lic
```diff
@@ -0,0 +1,140 @@
...
+      "Magic" => ['Lunar Magic', 'Holy Magic', 'Life Magic', 'Elemental Magic', 'Arcane Magic', 'Inner Magic', 'Inner Fire', 'Attunement', 'Arcana', 'Targeted Magic', 'Augmentation', 'Debilitation', 'Utility', 'Warding', 'Sorcery', 'Astrology', 'Summoning', 'Theurgy'],
+      "Survival" =>  ['Evasion', 'Athletics', 'Perception', 'Scouting', 'Stealth', 'Locksmithing', 'Thievery', 'First Aid', 'Outdoorsmanship', 'Skinning', 'Backstab', 'Thanatology'],
+      "Lore" => ['Forging', 'Engineering', 'Outfitting', 'Alchemy', 'Enchanting', 'Scholarship', 'Appraisal', 'Bardic Lore', 'Trading', 'Performance', 'Tactics', 'Empathy'] }
+    
+    passive_loop
+  end
+  
+  def check_last
+    _respond( bold("*** Checking last skillset used ***") ) if @debug
+    bput("study #{@tarantula_noun}", /knowledge of (.*) techniques/, /current set to consume .* experience\./) =~ /knowledge of (.*) techniques/
```
rpherbig: 💡 Is it 'current set to consume' or 'currently set to consume'?

tarantula.lic
```diff
@@ -0,0 +1,140 @@
...
+    end
+    skill = skills.flatten
+                  .reject{ |s| DRSkill.getrank(s) == 1750 }
+                  .reject{ |s| DRSkill.getrank(s) == 0 }
+                  .reject{ |s| DRSkill.getxp(s) < 27 }
+    return skill.sample
+  end
+  
+  def choose_alternate
+    selected = []
```
rpherbig: 💡 You could write this in a bit more idiomatic Ruby, I _think_  like this:

```
def choose_alternate
    skill = @all_skills
                    .reject{ |k,_| k == @last }
                    .values
                    .flatten
                    .reject{ |s| DRSkill.getrank(s) == 1750 }
                    .reject{ |s| DRSkill.getrank(s) == 0 }
                    .reject{ |s| DRSkill.getxp(s) < 30 }
    skill = skill - @exclude if @exclude # this could probably be in a .reject block by checking include? but I'm out of time to tinker with this
    return skill.sample
end
```

tarantula.lic
```diff
@@ -0,0 +1,140 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#tarantula
+=end
+
+custom_require.call(%w(common))
```
rpherbig: ❌ Needs `drinfomon`

tarantula.lic
BinuDR: It's current set.  I have bugged the typo in game.

PR #4079, opened 2019-12-07
update tarantula.lic - alternate skill bug fix

BinuDR:
This fixes the bug where Lunar/Holy/Elemental/etc magic gets chosen as the alternate skill.  Now it will turn to Magic if it's a priority choice or an alternate choice.


Tarhish:
I tried this out and it was a success, so 👍

PR #4082, opened 2019-12-08
hangs when wielding implement

minor-fixes-only:
When there's a tree explosion, the script runs safe-room then returns to the original location and resumes chopping.  The implement (axe) is never put away, but when the script returns to chop wood again, it tried to re-equip it.  For some reason there is no logic for already having the axe out in the equipping method, so the script stops.  The easy fix here is to remove line 34 and not re-equip since the implement is never put away.  May not be the most robust fix, however.

### DIFF
--- a/chop-wood.lic
+++ b/chop-wood.lic
@@ -31,7 +31,7 @@ class ChopWood
       snapshot = Room.current.id
       wait_for_script_to_complete('safe-room')
       walk_to(snapshot)
-      @equipment_manager.wield_weapon(@lumber_implement)
+     # @equipment_manager.wield_weapon(@lumber_implement)
     end
 
     return unless Flags['chop-danger']


rpherbig:
I'm not a fan of this. Necromancers empty their hands before healing. Can a tree explosion damage your hands?

Could you post a log of the failure?

minor-fixes-only:
Agree, easy but not good.  The issue seems to hinge around the stowing/getting of the axe.  See this log.  Script recovers when you stow/get the axe manually.
[chop-wood_error.txt](https://github.com/rpherbig/dr-scripts/files/3937190/chop-wood_error.txt)

chop-wood.lic
```diff
@@ -31,7 +31,7 @@ class ChopWood
       snapshot = Room.current.id
       wait_for_script_to_complete('safe-room')
       walk_to(snapshot)
-      @equipment_manager.wield_weapon(@lumber_implement)
+     # @equipment_manager.wield_weapon(@lumber_implement)
```
jandersson, line 34: :x: Just remove it - no need to comment out.

jandersson: This also doesn't seem like the right solution. I dont see the connection between the log you provided and removing this line, can you explain?

Dartellum: How about we stow the cutting implement before `wait_for_script_to_complete('safe-room')` then this would continue as expected?

chop-wood.lic
```diff
@@ -31,7 +31,7 @@ class ChopWood
       snapshot = Room.current.id
       wait_for_script_to_complete('safe-room')
```
Dartellum, line 32: @equipment_manager.empty_hands

chop-wood.lic
```diff
@@ -31,7 +31,7 @@ class ChopWood
       snapshot = Room.current.id
```
Dartellum, line 31: @equipment_manager.empty_hands

jandersson:
Closing this as it is over a year old, you can still re-open it.

Issue #4087, opened 2019-12-11
Base recipes needs some parts


TegCross:
Looks like a few recipes in weaponsmithing need parts assigned

- metal dire mace needs 
part:
- haft

a metal duraka skefne
part:
- long pole

metal duraka skefne
part:
- long pole

That's all I noticed so far. Thank you for your help!!!

PR #4098, opened 2019-12-18
Wounds when tying tools to belts

Tarhish:
Closes issue https://github.com/rpherbig/dr-scripts/issues/4087

There's a section in get_crafting_item for when you can't get an item because you're too wounded, but apparently it can happen when you're tying an item at different wound thresholds.  This can lead to gumming up the works, so I copied that logic over.

Also, 'What do you'/'What were you' are semi-common bput results, and if enough stuff get's thrown around when multiple scripts are running, I can sometimes see a crafting script exit fatally because it tried 'get shaper from my trunk' 3 times, and the second one succeeded, but the 3rd one failed.

Then you're stuck with items in your hands.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -137,13 +137,15 @@ module DRCC
         craft_room = Room.current.id
         wait_for_script_to_complete('safe-room', ['force'])
         DRCT.walk_to(craft_room)
-        get_crafting_item(name, bag, bag_items, belt)
+        return get_crafting_item(name, bag, bag_items, belt)
       end
     end
     command = "get my #{name}"
     command += " from my #{bag}" if bag_items && bag_items.include?(name)
     case DRC.bput(command, '^You get', '^You are already', '^What do you', '^What were you', 'You pick up', "can't quite lift it")
     when 'What do you', 'What were you'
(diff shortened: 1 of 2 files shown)


common-crafting.lic
```diff
@@ -157,7 +159,15 @@ module DRCC
 
     waitrt?
     if belt && belt['items'].find { |item| /\b#{name}/i =~ item || /\b#{item}/i =~ name }
-      DRC.bput("tie my #{name} to my #{belt['name']}", 'you attach')
+      case DRC.bput("tie my #{name} to my #{belt['name']}", 'you attach', 'Your wounds hinder')
+      when 'Your wounds hinder'
+        craft_room = Room.current.id
+        wait_for_script_to_complete('safe-room', ['force'])
+        DRCT.walk_to(craft_room)
+        return stow_crafting_item(name, bag, bag_items, belt)
```
rpherbig: ❓ Why does this line have four arguments but the method takes three?

Tarhish: Yeah, you're absolutely right -- there is different arg handling between get_item and stow_item.  Changed that in a commit.

PR #4105, opened 2020-01-02
Stow feet after hunt complete

Tarhish:
I can't see any reason this wouldn't work, and my initial tests seem like they are handled just fine.  After combat-trainer ends on any given hunt, but before afters like go2 are called, do what afk does and Stow feet until you can't anymore.

Anyone see a problem with this?  I'll ask around in a bit to see if anyone who routinely ends up with stuff at their feet minds trying it.

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -128,7 +128,9 @@ class HuntingBuddy
       end
 
       stop_actions(during_actions)
-
+      DRCI.stow_hand('left') unless !DRC.right_hand || !DRC.left_hand
+      while DRC.bput('stow feet', /You pick up/, /Stow what/) =~ /You pick up/
+      end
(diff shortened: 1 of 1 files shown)


hunting-buddy.lic
```diff
@@ -128,7 +128,9 @@ class HuntingBuddy
       end
 
       stop_actions(during_actions)
-
+      DRCI.stow_hand('left') unless !DRC.right_hand || !DRC.left_hand
```
rpherbig: ❓ Why check `right_hand` before stowing the left?

Tarhish: Primarily because I've never heard of ;afk causing a problem this way, so I kept the logic unchanged, but the notion of 'stow your left hand unless either your right/left hand is free' makes intuitive sense to me at least.

PR #4108, opened 2020-01-04
Added Crossing Estate Holder Gargoyles

Hiinky:
Should not be approved until the map upload with a stringproc for room 1629 is live.  Room data does not transition from 1629 to 10732 but does transition from 10732 to 1629.  Room 1629 appears to be the problem as the entire room is nothing but procs.
Once approved, this is a super yummy new empath-safe hunting spot we are very excited to be using!!!

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -620,6 +620,17 @@ hunting_zones:
   - 12991
   - 12992
   - 12996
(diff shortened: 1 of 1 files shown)


rpherbig:
✔️ Looks good, let me know when to merge it

Hiinky:
Map update appears live!  Just updated and reloaded and I was able to go2 into the zone!!

PR #4126, opened 2020-01-18
added training_ability: Ret Stealth

Hiinky:
I get asked fairly often about retreating before hiding and this seemed the least intrusive way to add it.  Was actually working quite well for my cleric's stealth hunt during testing when she can just barely hide on the current mob.  - open to different names.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2065,27 +2065,11 @@ class TrainerProcess
           fix_standing
         end
       end
+     when 'Stealth'
+      hide_action(game_state)
     when 'Ret Stealth'
       retreat
-      if hide?(@hide_type) && !@dont_stalk && !game_state.npcs.empty? && @hide_type == 'hide'
-        stalked = bput('stalk', 'Stalk what',
-                       'Try being out of sight',
-                       'You move into position',
-                       'already stalking',
-                       'discovers you, ruining your hiding place') == ('You move into position' || 'already stalking')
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2065,6 +2065,17 @@ class TrainerProcess
           fix_standing
         end
       end
+    when 'Ret Stealth'
+      retreat
```
cconnors23: I had to backtrain stealth on my Barb and I was toying with something similar by extending Stealth to take an optional yaml parameter for "hide_range" with a default of melee. And if you set pole or missile, then it would ensure you were retreated to the appropriate range before hiding.

combat-trainer.lic
```diff
@@ -2065,6 +2065,17 @@ class TrainerProcess
           fix_standing
         end
       end
+    when 'Ret Stealth'
```
rpherbig: 💡 'Stealth With Retreat'?

combat-trainer.lic
```diff
@@ -2065,6 +2065,17 @@ class TrainerProcess
...
+      retreat
+      if hide?(@hide_type) && !@dont_stalk && !game_state.npcs.empty? && @hide_type == 'hide'
```
rpherbig: ❓ Is this section a copy of the 'Stealth' block? If so, you should extract this to a re-used helper function.

Hiinky: updated!  Thanks binu for some nudging.

PR #4129, opened 2020-01-20
Adjust pick buffing/worn-container handling

Tarhish:
There exists a failure state when picking pet boxes with a worn container that shares a noun with a box.  Let's say a 'trunk' and a worn 'cartographer's trunk'.  If you get to Locksmithing > 30 mindstates, pick puts the trunk away, then attempts to loot your trunk.  At that point it looks in the cartographer's trunk, which parses first, and empties everything inside it onto the ground.  This orders some of the logic about so you only loot/dismantle if you actually opened the box.

Also, I'm removing lockpick_buffs: after a couple years and moving us over to DRCA.do_buffs unless someone else has an objection.  I ran this on my thief and it Khri'd up, picked some boxes, then stopped my khris and closed down.

Finally, an insignificant change to DRCI.in_hands?  My original testing never ran into this somehow, but adding strings together crashes if one is nil.  Doing it this way instead works in Ruby.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -101,7 +101,7 @@ module DRCI
 
   def in_hands?(item)
     item_regex = item.split.size > 1 ? /.*\b#{item.split.first}.*\b#{item.split.last}/i : /\b#{item}/i
-    item_regex =~ DRC.left_hand + DRC.right_hand
+    item_regex =~ "#{DRC.right_hand}#{DRC.left_hand}"
   end
 
   def exists?(description)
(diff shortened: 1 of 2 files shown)


rpherbig:
> Also, I'm removing lockpick_buffs: after a couple years and moving us over to DRCA.do_buffs unless someone else has an objection. I ran this on my thief and it Khri'd up, picked some boxes, then stopped my khris and closed down.

I'm a bit worried that this could surprise people. At the least, please announce it on lnet a few times and add a validator that `lockpick_buffs` is deprecated.

Tarhish:
I can definitely do that.  I was thinking that two years straight of people getting nagged by ;pick to swap to waggles would have weeded out anyone still using lockpick_buffs, but I can see now that's only mostly true.

Tarhish:
People had far more varied opinions on picking than I expected!  I think I'm going to look into that settings nagger/transformer idea and kill two birds with one stone.  For now, I'm undoing that buff stuff except for changing it to a DRCA call.

PR #4148, opened 2020-02-13
Add Sample Empath file

Hiinky:
Highly commented Empath yaml


profiles/SampleEmpath-setup.yaml
```diff
@@ -0,0 +1,695 @@
...
+full_pouch_container: lootsack
+#separate location for spare empty pouches
+spare_gem_pouch_container: lotusweave.pouch
+
+#true to save pouches for trader to sell for better profit
+# false to sell yourself
+tie_gem_pouches: true
+
+#pick a specific gem pouch adjective.  
+#DO NOT USE "GEM"! It will eventually break and you'll en dup in the junk yard
```
rpherbig: 💡 'end up'

profiles/SampleEmpath-setup.yaml
```diff
@@ -0,0 +1,695 @@
...
+#     - Appraisal
+#     - Performance
+
+# #skills to never ever "eat"
+# tarantula_excluded_skills:
+#   - Empathy
+
+# don't say anything when entering a hunting room
+empty_hunting_room_messages:
+- /
```
rpherbig: ❔ Does this say `/` in your room?

rpherbig:
Something in this YAML file is breaking the 'valid YAML' test:

```
  1) Error:
TestValidYaml#test_all_settings_in_base_yaml:
TypeError: no implicit conversion of String into Hash
    /home/travis/build/rpherbig/dr-scripts/test/test_valid_yaml.rb:27:in `merge!'
    /home/travis/build/rpherbig/dr-scripts/test/test_valid_yaml.rb:27:in `block in test_all_settings_in_base_yaml'
    /home/travis/build/rpherbig/dr-scripts/test/test_valid_yaml.rb:27:in `each'
    /home/travis/build/rpherbig/dr-scripts/test/test_valid_yaml.rb:27:in `each_with_object'
    /home/travis/build/rpherbig/dr-scripts/test/test_valid_yaml.rb:27:in `test_all_settings_in_base_yaml'
```

Can you debug this?

profiles/SampleEmpath-setup.yaml
BinuDR: No, I use that too.  It does this:

>'/
>
You must specifiy an emote.

Hiinky: Many of us use this to avoid the telltale "anyone here?"  It's definitely safe to use.

profiles/SampleEmpath-setup.yaml
Hiinky: Fixed

Hiinky:
I honestly have no idea what is causing the problem.  the file validates and runs so I'm not sure where to even look for a fail?

Tarhish:
I know I haven't been around a bit but I poked around a bit and figured this out.  The tester's freaking out because of the two --s in the first line.  none is fine and three is fine, but two is apparently right out.

Hiinky:
Hi Q!!  Trying that  now.

Tarhish:
I think based on the travis method, it doesn't see attack_overrides in base.  Which really only means that when it was added to combat-trainer it didn't get added to base.yaml and base-empty: {}

Tarhish:
At some point we'll have to add that in, but if you wanted to make sure this went through you could remove attack_overrides from this sample.

Hiinky:
its really only here for educational purposes.  If this goes through, i'll add the other two so it goes next time.

Hiinky:
<3 Q  thank!

profiles/SampleEmpath-setup.yaml
rpherbig: Makes sense, thanks

profiles/base-empty.yaml
```diff
@@ -53,6 +53,7 @@ empty_values:
   classes_to_teach: []
   whirlwind_trainables: []
   aiming_trainables: []
+  attack_overrides: []
```
rpherbig: ❓ Shouldn't this be a hash (and not an array)?

Hiinky: Fixed!

PR #4149, opened 2020-02-13
Create include directive

vtcifer:
Add new supported directive to yamls: include

Example:
```yaml
include:
  - foo
  - bar
```

This will attempt to load files located in profiles called 'include-foo.yaml' and 'include-bar.yaml'. 
This allows some settings to be shared, and only edited once to update for multiple characters.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -75,7 +75,7 @@ module DRCA
         activate_barb_buff(name)
         waitrt?
         pause settings.meditation_pause_timer.to_i if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
-        fix_standing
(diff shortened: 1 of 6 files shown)


rpherbig:
What would you think about requiring these file names to be of the form `CharacterName-include-foo.yaml`?

vtcifer:
> What would you think about requiring these file names to be of the form `CharacterName-include-foo.yaml`?

I think that defeats the reasons for this change.  
This allows you to easily have some settings that are common across characters, without having to update individual character yamls.

rpherbig:
Sorry, I missed 'for multiple characters' and I thought it was a way to manage multiple YAMLs for a single character.

PR #4150, opened 2020-02-16
New script burgle.lic

vtcifer:
Support for Breaking and Entering, released to live today.

### DIFF
--- a/bne.lic
+++ b/bne.lic
@@ -1,13 +1,8 @@
 =begin
 
-Very simple BnE script.  Use at your own risk - there are VERY high fines if you get caught
+Very simple Breaking and Entering script.  Use at your own risk - there are VERY high fines if you get caught
 
-#yaml settings examples:
-#bne_settings:
-#  room: nnnn            # room id that bne should be done in.  Guards should never come to this room, so pick carefully
-#  entry_type: loxkpick  # entry method - rope or lockpick.  must have a heavy rope if using rope.  if using lockpick, it will use use_lockpick_ring/lockpick_container settings
-#  search_count: 0       # futue develpment, not supported right now
-#  dispose: false        # futue develpment, not supported right now , when supproted will use stealing_bag for goods
+Documentation: https://elanthipedia.play.net/Lich_script_repository#bne
 
 =end
 
@@ -22,24 +17,46 @@ class BNE
     start_script('jail-buddy') unless Script.running?('jail-buddy')
     @settings = get_settings
     @bne_settings = @settings.bne_settings
+    if !valid_bne_settings?
+      message("Invalid bne_settings.  Chcek the documenatation before running, this is a dangerous script to run if it's not understood.")
+      message("Very high fines, and loss of all your items if you can't pay your debt are possible.")
+      pause 10
+      exit
+    end
+
     burgle
+  end
+
+  def valid_bne_settings?
+    return false if @bne_settings.empty?
+    return false if @bne_settings['room'] == nil or @bne_settings['room'] =~ /\D/
+    return false if @bne_settings['entry_type'] == nil or @bne_settings['entry_type'] !~ /lockpick|rope/i
+    if @bne_settings['search_count'] > 0
+      return false if /(That is already open|You open your)/ !~ bput("open my #{@settings.stealing_bag}", 'That is already open', 'You open your', 'Please rephrase that command', 'What were you referring')
+    end
+
+    return true
   end
 
   def burgle
     #check to make sure you aren't under cool down
-    return unless /The heat has died down from your last caper/ =~ DRC.bput('burgle recall', '^You should wait at least \d+ roisaen for the heat to die down.', '^The heat has died down from your last caper')
+    return unless /The heat has died down from your last caper/ =~ bput('burgle recall', '^You should wait at least \d+ roisaen for the heat to die down', '^The heat has died down from your last caper')
 
     #get to your defined room:
     walk_to(@bne_settings['room'])
 
+    #stow any items in hands before begining - entry method must be in right hand, or a worn lockpick ring
+    #right hand takes priority
+    stow_hands
+
     #check for entry method
     case @bne_settings['entry_type']
-    when /rope/
+    when /rope/i
       if bput('get my heavy rope', 'You get', 'You are already holding' ,'What were you') =~ /What were you/
         message("Couldn't find entry item: heavy rope")
         return
       end
-    when /lockpick/
+    when /lockpick/i
       if @settings.use_lockpick_ring
         if !exists?("lockpick #{@settings.lockpick_container}")
           message("Couldn't find entry item: lockpick #{@settings.lockpick_container}")
@@ -53,6 +70,7 @@ class BNE
       message("Unknown entry method: #{@bne_settings['entry_type']}")
       return
     end
+
     #setup your buffs for bne
     wait_for_script_to_complete('buff', ['bne']) if @settings.waggle_sets['bne']
 
@@ -67,10 +85,67 @@ class BNE
         end
       end
     end
+
     bput('burgle', 'You make short work of the lock', 'You scale up the side of a wall')
-      #for each room_list  - tbd
-    fput('go window')
+
+    #stow entry method if needed once inside - easier loot managment
+    stow_hands
+
+    #TODO:  Define list of room types, and searchable objects to search for loot (https://elanthipedia.play.net/Breaking_and_Entering#Rooms)
+    #       Write movement and get out routines
+    if @bne_settings['search_count'] > 0
+      search_for_loot('counter')
+    end
+
+    bput('go window', 'You take a moment to reflect on the caper')
+
+  end
+
+  def search_for_loot(target)
+    if /It looks valuable/ =~ bput("search #{target}",'It looks valuable','Roundtime','I could not')
+      if @bne_settings['dispose']
+        drop_item("#{DRC.right_hand}")
+      else
+        put_item?("#{DRC.right_hand}")
+      end
+    end
+  end
+
+  def put_item?(item)
+    case bput("put my #{item} in my #{@settings.stealing_bag}", 'What were you', 'You put', "You can't do that", 'no matter how you arrange it', 'The .* is *.* too \w+ to fit in', 'There isn\'t any more room', 'perhaps try doing that again')
+    when 'perhaps try doing that again'
+      return put_item?(item)
+    when 'You put'
+      return true
+    when 'What were you'
+      handheld = held_item(item)
+      drop_item(held_item(item)) if handheld
+      return false
+    end
+
+    false
+  end
+
+  def drop_item(item)
+    case bput("drop my #{item}", 'You drop', 'would damage it', 'smashing it to bits', 'Something appears different about', 'What were you')
+    when 'would damage it', 'Something appears different about'
+      drop_item(item)
+    when 'What were you'
+      handheld = held_item(item)
+      return drop_item(held_item(item)) if handheld
+    end
+  end
+
+  def held_item(item)
+    [DRC.right_hand, DRC.left_hand].each do |hand_item|
+      hand_item.split.each do |item_word|
+        return hand_item if item.include?(item_word)
+      end
+    end
+
+    return nil
   end
 
 end
+
 BNE.new
--- a/bne.lic
+++ b/bne.lic
@@ -2,13 +2,12 @@
 
 Very simple BnE script.  Use at your own risk - there are VERY high fines if you get caught
 
-yaml settings examples:
-
-bne_settings:
-  room: nnnn            # room id that bne should be done in.  Guards should never come to this room, so pick carefully
-  entry_type: loxkpick  # entry method - rope or lockpick.  must have a heavy rope if using rope.  if using lockpick, it will use use_lockpick_ring/lockpick_container settings
-  search_count: 0       # futue develpment, not supported right now
-  dispose: false        # futue develpment, not supported right now , when supproted will use stealing_bag for goods
+#yaml settings examples:
+#bne_settings:
+#  room: nnnn            # room id that bne should be done in.  Guards should never come to this room, so pick carefully
+#  entry_type: loxkpick  # entry method - rope or lockpick.  must have a heavy rope if using rope.  if using lockpick, it will use use_lockpick_ring/lockpick_container settings
+#  search_count: 0       # futue develpment, not supported right now
+#  dispose: false        # futue develpment, not supported right now , when supproted will use stealing_bag for goods
 
 =end
 
--- a/bne.lic
+++ b/bne.lic
@@ -1,0 +1,77 @@
+=begin
+
+Very simple BnE script.  Use at your own risk - there are VERY high fines if you get caught
+
+yaml settings examples:
+
+bne_settings:
+  room: nnnn            # room id that bne should be done in.  Guards should never come to this room, so pick carefully
+  entry_type: loxkpick  # entry method - rope or lockpick.  must have a heavy rope if using rope.  if using lockpick, it will use use_lockpick_ring/lockpick_container settings
+  search_count: 0       # futue develpment, not supported right now
+  dispose: false        # futue develpment, not supported right now , when supproted will use stealing_bag for goods
+
+=end
+
+custom_require.call(%w[common common-travel common-items])
+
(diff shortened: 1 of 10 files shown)


bne.lic
```diff
@@ -0,0 +1,76 @@
+=begin
+
+Very simple BnE script.  Use at your own risk - there are VERY high fines if you get caught
```
rpherbig: 💡 I have a minor preference for this documentation going on elanthipedia

bne.lic
```diff
@@ -0,0 +1,76 @@
...
+
+class BNE
+  include DRC
+  include DRCT
+  include DRCI
+
+  def initialize
+    start_script('jail-buddy') unless Script.running?('jail-buddy')
+    @settings = get_settings
+    @bne_settings = @settings.bne_settings
```
rpherbig: ❌ `bne_settings` needs to be in `base.yaml` and `base-empty.yaml`

bne.lic
```diff
@@ -0,0 +1,76 @@
+=begin
+
+Very simple BnE script.  Use at your own risk - there are VERY high fines if you get caught
+
+#yaml settings examples:
+#bne_settings:
+#  room: nnnn            # room id that bne should be done in.  Guards should never come to this room, so pick carefully
+#  entry_type: loxkpick  # entry method - rope or lockpick.  must have a heavy rope if using rope.  if using lockpick, it will use use_lockpick_ring/lockpick_container settings
+#  search_count: 0       # futue develpment, not supported right now
+#  dispose: false        # futue develpment, not supported right now , when supproted will use stealing_bag for goods
```
rpherbig: 💡 'supproted' -> 'supported'

bne.lic
```diff
@@ -0,0 +1,76 @@
+=begin
+
+Very simple BnE script.  Use at your own risk - there are VERY high fines if you get caught
+
+#yaml settings examples:
+#bne_settings:
+#  room: nnnn            # room id that bne should be done in.  Guards should never come to this room, so pick carefully
+#  entry_type: loxkpick  # entry method - rope or lockpick.  must have a heavy rope if using rope.  if using lockpick, it will use use_lockpick_ring/lockpick_container settings
```
rpherbig: 💡 'loxkpick' -> 'lockpick'

bne.lic
```diff
@@ -0,0 +1,76 @@
...
+      elsif bput('get my lockpick', 'You get', 'You are already holding' ,'What were you') =~ /What were you/
+        message("Couldn't find entry item: lockpick")
+        return
+      end
+    else
+      message("Unknown entry method: #{@bne_settings['entry_type']}")
+      return
+    end
+    #setup your buffs for bne
+    wait_for_script_to_complete('buff', ['bne']) if @settings.waggle_sets['bne']
```
rpherbig: ❔ Might be worth mentioning in the documentation that it supports a buff set?

bne.lic
```diff
@@ -0,0 +1,76 @@
...
+        hide_attemps -= 1
+        if hide_attemps <= 0
+          message("Couldn't hide.  Find a better room.")
+          return
+        end
+      end
+    end
+    bput('burgle', 'You make short work of the lock', 'You scale up the side of a wall')
+      #for each room_list  - tbd
+    fput('go window')
```
rpherbig: ❔ Can this be a `bput`? Don't worry if it can't

rpherbig:
A bunch of nice-to-have comments. The only thing stopping this from being merged are the settings in base and base-empty.

vtcifer:
All reviewed comments should be applied.

dr-leustyin:
Any chance we can change the name to ;burgle ?

;bne isn't a very intuitive name

bne.lic
```diff
@@ -67,10 +85,67 @@ class BNE
...
+    end
+
+    bput('go window', 'You take a moment to reflect on the caper')
+
+  end
+
+  def search_for_loot(target)
+    if /It looks valuable/ =~ bput("search #{target}",'It looks valuable','Roundtime','I could not')
+      if @bne_settings['dispose']
+        drop_item("#{DRC.right_hand}")
```
rpherbig: 💡 The string interpolation isn't necessary here (or line 109)

bne.lic
```diff
@@ -67,10 +85,67 @@ class BNE
...
+
+  def put_item?(item)
+    case bput("put my #{item} in my #{@settings.stealing_bag}", 'What were you', 'You put', "You can't do that", 'no matter how you arrange it', 'The .* is *.* too \w+ to fit in', 'There isn\'t any more room', 'perhaps try doing that again')
+    when 'perhaps try doing that again'
+      return put_item?(item)
+    when 'You put'
+      return true
+    when 'What were you'
+      handheld = held_item(item)
+      drop_item(held_item(item)) if handheld
```
rpherbig: 💡 Could this be `drop_item(handheld) if handheld`?

bne.lic
```diff
@@ -67,10 +85,67 @@ class BNE
...
+    if /It looks valuable/ =~ bput("search #{target}",'It looks valuable','Roundtime','I could not')
+      if @bne_settings['dispose']
+        drop_item("#{DRC.right_hand}")
+      else
+        put_item?("#{DRC.right_hand}")
+      end
+    end
+  end
+
+  def put_item?(item)
```
rpherbig: 💡 Why does `put_item?` return a boolean if its return value isn't used?

rpherbig:
@vtcifer looks good - I'm ready to merge it. How do you feel about the comment from @dr-leustyin ?

Hiinky:
Might be an edit for later as its more complicated due to the random exits but if we could add sneaking from room to room before searching and jumping out the window, it doubles your stealth mindstates.

Dartellum:
Could you add a section to pause scripts, such as sanowret-crystal or performance? In def initialize: 
    pause_script('performance') if Script.running?('performance')
    pause_script('sanowret-crystal') if Script.running?('sanowret-crystal')

then:
before_dying do
  unpause_script('performance') if Script.running?('performance')
  unpause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
end

etc.

BinuDR:
> 
> 
> Could you add a section to pause scripts, such as sanowret-crystal or performance? In def initialize:
> pause_script('performance') if Script.running?('performance')
> pause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> 
> then:
> before_dying do
> unpause_script('performance') if Script.running?('performance')
> unpause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> end
> 
> etc.

It might be a good idea to pause all scripts.  Any accidental interference could be highly problematic.

Dartellum:
> > Could you add a section to pause scripts, such as sanowret-crystal or performance? In def initialize:
> > pause_script('performance') if Script.running?('performance')
> > pause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> > then:
> > before_dying do
> > unpause_script('performance') if Script.running?('performance')
> > unpause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> > end
> > etc.
> 
> It might be a good idea to pause all scripts. Any accidental interference could be highly problematic.

put **pause_all** under the start jail-buddy line. :)

BinuDR:
and **unpause_all** in the **before_dying do** section.

Dartellum:
> and **unpause_all** in the **before_dying do** section.

That did not work for me. 
This did though: **Script.running.find_all { |s| s.paused? && !s.no_pause_all }.each(&:unpause)**

vtcifer:
> Any chance we can change the name to ;burgle ?
> 
> ;bne isn't a very intuitive name

IMO, BNE is a pretty common shorthand for breaking and entering (which is the system name technically, and the shorthand most are using already).  I'm a big fan of script names as short as possible. 

> Might be an edit for later as its more complicated due to the random exits but if we could add sneaking from room to room before searching and jumping out the window, it doubles your stealth mindstates.

Once more room searches are put in, I don't think that'll be necessary.  You get additional stealth exp for stealing too, get 1 or 2 items, and it's damn near 34/34 to begin with.  Sneaking uses RT which adds too much danger I think.  I'm open to it though, if it's really needed after searching more than 1 room is in.

> > Could you add a section to pause scripts, such as sanowret-crystal or performance? In def initialize:
> > pause_script('performance') if Script.running?('performance')
> > pause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> > then:
> > before_dying do
> > unpause_script('performance') if Script.running?('performance')
> > unpause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> > end
> > etc.
> 
> It might be a good idea to pause all scripts. Any accidental interference could be highly problematic.

Not a bad idea.  does pause-all impact jail-buddy?  Because that should remain running.

bne.lic
vtcifer: Ripped it out of steal, for future use with bining/pawning.

Dartellum:
> > Any chance we can change the name to ;burgle ?
> > ;bne isn't a very intuitive name
> 
> IMO, BNE is a pretty common shorthand for breaking and entering (which is the system name technically, and the shorthand most are using already). I'm a big fan of script names as short as possible.
> 
> > Might be an edit for later as its more complicated due to the random exits but if we could add sneaking from room to room before searching and jumping out the window, it doubles your stealth mindstates.
> 
> Once more room searches are put in, I don't think that'll be necessary. You get additional stealth exp for stealing too, get 1 or 2 items, and it's damn near 34/34 to begin with. Sneaking uses RT which adds too much danger I think. I'm open to it though, if it's really needed after searching more than 1 room is in.
> 
> > > Could you add a section to pause scripts, such as sanowret-crystal or performance? In def initialize:
> > > pause_script('performance') if Script.running?('performance')
> > > pause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> > > then:
> > > before_dying do
> > > unpause_script('performance') if Script.running?('performance')
> > > unpause_script('sanowret-crystal') if Script.running?('sanowret-crystal')
> > > end
> > > etc.
> > 
> > 
> > It might be a good idea to pause all scripts. Any accidental interference could be highly problematic.
> 
> Not a bad idea. does pause-all impact jail-buddy? Because that should remain running.

Jail-buddy stays running. 

I had to use **Script.running.find_all { |s| s.paused? && !s.no_pause_all }.each(&:unpause)** as unpause_all did not work. Trying to figure out why.

bne.lic
vtcifer: I think I just left that in from when I was using bput and didn't fully clear it when I ripped the code out of steal.  I can fix that with the pausing other scripts too.

vtcifer:
> Jail-buddy stays running.
> 
> I had to use **Script.running.find_all { |s| s.paused? && !s.no_pause_all }.each(&:unpause)** as unpause_all did not work. Trying to figure out why.

Actually, may be better to iterate and store all unpaused scripts to pause anyway, that way we don't unpause something paused for other reasons than this script.

Hiinky:
I vote name burgle.  I haven't used BNE anywhere but since its going to be mostly called by t2 its not terrible.

The only way I've been able to get 17/34 stealth as a survival second or tert is to sneak from room to room and search twice.  Otherwise its only 9/34 and I've not had any trouble getting out.

I would prefer jail-buddy to not work. It defaults to plead innocent while pleading guilty significantly lowers the fine if caught.

vtcifer:
> The only way I've been able to get 17/34 stealth as a survival second or tert is to sneak from room to room and search twice. Otherwise its only 9/34 and I've not had any trouble getting out.

I'll do some more testing with this.  I suspect it'll be quicker/safe to move->hide->search than sneak->search.

> I would prefer jail-buddy to not work. It defaults to plead innocent while pleading guilty significantly lowers the fine if caught.

I think jail-buddy would need to be enhanced in that case.  Due to the fines for getting caught ideally it should never be needed (we want to avoid getting caught 100% of the time), but given the risk of item loss I can't say leaving it out is a good idea.

vtcifer:
I think I've got a good idea for the movement, and want to put it into psuedocode before working on it.  A recursive function I think works best here.  


This goes right after the burgle, before the go window:
```
	if search_max > 0
		@search_counter = 0
		rob_the_place
	end

def rob_the_place(direction=nil)
	visited.add(direction) unless direction = nil
	loot(roomobj) if !blacklist.contains(roomname)
	if search_counter >= search_max return
	while directions.size > visited.size and search_counter < search_max
		pick random direction you haven't been -> newdir
		visited.add(newdir)
		reverse = reverse[newdir]
		move(newdir)
		rob_the_place(reverse)
		move(reverse)
	loop
end

def Move(direction)
	if hidden && search_counter < search_max
		sneak direction
	else
		direction
	end
end

def loot(obj)
	hide if !hidden && !invis
	search obj
	if succesful
		store hands if full
		return
	elsif not succesful && retry=true
		loot(obj)
	end
end
```

One question I have with lich/ruby.  Is there a way with flags to change a variable value when the string occurs (or something similar).

BinuDR:
Here's the list of rooms and their respective surfaces for searching.

https://elanthipedia.play.net/Breaking_and_Entering#Rooms

vtcifer:
I'm not against changing the name from bne to something else, but I'd like avoid burgle, as someone running it manually, and mistyping could result in doing it before the cooldown is up.

rpherbig:
> I'm not against changing the name from bne to something else, but I'd like avoid burgle, as someone running it manually, and mistyping could result in doing it before the cooldown is up.

We've solved this in two ways in other scripts:
1. Require an argument (e.g. `start`) or the script aborts, to prevent accidental executions
1. Check cooldown when the script runs and refuse to run unless an `override` arg is provided

rpherbig:
@vtcifer there has been a lot of great discussion in here. That said, I want to try to differentiate changes necessary to make this script usable, vs those to make it better. The former is all that is needed to merge this. Let me know when you feel like it's in a good place - you can always make it better in future PRs.

vtcifer:
> > I'm not against changing the name from bne to something else, but I'd like avoid burgle, as someone running it manually, and mistyping could result in doing it before the cooldown is up.
> 
> We've solved this in two ways in other scripts:
> 
> 1. Require an argument (e.g. `start`) or the script aborts, to prevent accidental executions
> 2. Check cooldown when the script runs and refuse to run unless an `override` arg is provided

My concern was if they typo without the script start start character. ex, instead of ;burgle, they typo burgle.  Given that in game burgle now defaults to burgle help if you type anything after that isn't help/recall, I think requiring a "Start" arg can put in.
#2 is already taken care of in the script.

> @vtcifer there has been a lot of great discussion in here. That said, I want to try to differentiate changes necessary to make this script usable, vs those to make it better. The former is all that is needed to merge this. Let me know when you feel like it's in a good place - you can always make it better in future PRs.

Excellent point.  I think the last bits before it's MVP, is the script pausing and the typo protection.  Next commit will have those.

Room Movement, blacklisting and bin/pawn support can be in subsequent PRs

vtcifer:
I want to call out for transparency, there is also an update to common.lic here to support ending khri silence with release_invisibility.

DRWafflesworth:
Could this be called with an arg to choose on the spot whether to use a rope or lockpick?

burgle.lic
```diff
@@ -0,0 +1,178 @@
...
+    ]
+
+    args = parse_args(arg_definitions)
+
+    start_script('jail-buddy') unless Script.running?('jail-buddy')
+    @settings = get_settings
+    @burgle_settings = @settings.burgle_settings
+
+    if !valid_burgle_settings?
+      message("Invalid burgle_settings.  Chcek the documenatation before running, this is a dangerous script to run if it's not understood.")
```
BinuDR: 💡  Chcek -> Check

rpherbig:
✔️ Looks good to me. @vtcifer are you ready for me to merge it?

vtcifer:
> ✔️ Looks good to me. @vtcifer are you ready for me to merge it?

GTG.  I'll update the message typos in the next PR,

Hiinky:
For sure something for a later version but I second the request to be able to send rope or lockpick as an arg if we want to rotate methods on the fly.

Dartellum:
You could change requiring start to requiring lockpick or rope.

Stormsocks:
I'm going to toss in a third for the entry method changing - something like necromancers use for their rituals would be incredible:

cycle, rope, or lockpick.   cycle would choose whichever skill has the lowest exp at the moment.

When additional rooms get added, it would be great if there was a way to set it to loot the special room before the kitchen/bedroom.

PR #4152, opened 2020-02-17
Change to allow for different types of lockpick stackers

BinuDR:
This will allow people to specify both the adjective and noun in the setting so that any type of stacker can be used.  For example:
lockpick_container: golden key
lockpick_container: lockpick ankle-cuff

I changed the default in base.yaml to:  lockpick_container: lockpick ring

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -151,7 +151,7 @@ module DRCI
   end
 
   def count_lockpick_container(container)
-    count = DRC.bput("app my lockpick #{container}", 'and it appears to be full', 'and it might hold an additional \d+', '\d+ lockpicks would probably fit').scan(/\d+/).first.to_i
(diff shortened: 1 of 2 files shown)


vtcifer:
One of the things I was thinking about updating was how it refills.

Since there are stackers out there that are 50 count, with the way it currently works it refills them when there are a lot of picks still on it. Would this be a good time to update that too?

rpherbig:
My only concern with this PR is that this could break for anyone who has overridden the setting and is relying on the existing behavior.

@BinuDR are you going to be around tonight? I'd like you to warn people in LNET a few times throughout the night  and be available to act as tech support if this breaks for anyone.

rpherbig:
> Since there are stackers out there that are 50 count, with the way it currently works it refills them when there are a lot of picks still on it. Would this be a good time to update that too?

This would be a great change, but should be in a separate PR.

Hiinky:
I told Binu I'd help announce the changes.  I know Dyler said he will too.  I'm around tons during the day too.

PR #4154, opened 2020-02-19
Quick change to let thieves farm with backstab even when mindlocked.

jayRyan24:
By request.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3136,6 +3136,7 @@ class GameState
     echo("  @ambush: #{@ambush}") if $debug_mode_ct
 
     @backstab = settings.backstab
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -3136,6 +3136,7 @@ class GameState
     echo("  @ambush: #{@ambush}") if $debug_mode_ct
 
     @backstab = settings.backstab
+    @backstab_past_mindlock = settings.backstab_past_mindlock
```
rpherbig: Please add a default value to `base.yaml`

jayRyan24: Sorry this took so long, added it.

PR #4159, opened 2020-02-22
burgle.lic - Don't try to open stealing bag if dropping loot

BinuDR:
Added another condition to the if statement so it doesn't try to open your stealing_bag if you are set to drop loot.

### DIFF
--- a/burgle.lic
+++ b/burgle.lic
@@ -43,7 +43,7 @@ class Burgle
     return false if @burgle_settings.empty?
     return false if @burgle_settings['room'] == nil or @burgle_settings['room'] =~ /\D/
     return false if @burgle_settings['entry_type'] == nil or @burgle_settings['entry_type'] !~ /lockpick|rope/i
-    if @burgle_settings['max_search_count'] > 0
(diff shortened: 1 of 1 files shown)


vtcifer:
I have a pending fix for this, because unfortunately due to some items with custom drop scripts, and how they interact with hidden/invisibility you will need to have a loot container set 100% of the time.

vtcifer:
```I>drop blanket
Trying to go unnoticed, are you?
I>empty right
Trying to go unnoticed, are you?

I>rele eotb
Your corruption fades, revealing you to the world once more.
>empty right
You spread a thick jade wool blanket embroidered with twining vines on the ground.

[Remember, if you leave the game or are disconnected, someone else may take your blanket.  If it's lost to a crash, we won't replace it.  And the janitor and shadowlings are acquisitive creatures, who are quite capable of stealing this from you -- leave it on the ground at your own risk.]```

rpherbig:
@vtcifer is this a short-term fix we can merge in, or do we need to wait for your PR?

BinuDR:
It doesn't need to be merged.

The wiki should be updated to mention that the stealing_bag setting is required though.

I made this change because the script was hanging trying to open a backpack I don't have since that's the default value in base.yaml

PR #4161, opened 2020-02-22
Burgle.lic add additional options

vtcifer:
- Added command line argument overrides for entry, room and loot_type (what to do with loot)
- Added support for multiple rooms
- Migrated reused settings (stealing_bag, use_lockpick_ring, lockpick_container) into burgle settings due to some confusion as to the location of those settings.  
  - If users want to reuse these settings, yaml anchors can provide that functionality, and if not they can set different settings. 
  - This is a non-breaking change as there is code to mitigate that and warn users of the change.
- Changed how dropping loot works.  It will always attempt to store it now, meaing you must always have a burgle_settings:loot_container set to work.
- Added flags to watch for footsteps, and if safe_mode setting is set, try to get out as  quick as possible.
- Add cycle option to pick from lowest of locksmithing/athletics learning rate, and lowest skilled if learning rate are the same.
- Add checks to getting to your burgle room so if it fails, it exits out.
- Added before:/after: so that custom scripts for things such as putting pets away/taking them out can be done

### DIFF
--- a/burgle.lic
+++ b/burgle.lic
@@ -26,10 +26,6 @@ class Burgle
 
     args = parse_args(arg_definitions)
 
-    @entry_type = args.entry if args.entry
-    @burgle_room = args.roomid if args.roomid
-    @loot_type = args.loot_type if args.loot_type
-
     #hash which maps directions to their reverse, for backing out of the house
     @reverse_direction_map = {
       'east' => 'west',
@@ -61,12 +57,12 @@ class Burgle
     @lockpick_container =  @burgle_settings['lockpick_container']
 
     #set yaml settings unless being overridden from the command line arguments
-    @entry_type = @burgle_settings['entry_type'] unless @entry_type
-    @burgle_room = @burgle_settings['room'] unless @burgle_room
-    @loot_type = @burgle_settings['loot'] unless @loot_type
+    @entry_type = args.entry || @burgle_settings['entry_type']
+    @burgle_room = args.roomid || @burgle_settings['room']
+    @loot_type = args.loot_type ||  @burgle_settings['loot']
 
-    $burgle_before_scripts = @burgle_settings['before']
-    $burgle_after_scripts = @burgle_settings['after']
+    @burgle_before_scripts = @burgle_settings['before']
+    @burgle_after_scripts = @burgle_settings['after']
 
     #check if still looks like using old settings, and don't break things for now
     if @loot_container == nil
@@ -98,6 +94,7 @@ class Burgle
       pause 10
       exit
     end
+
     burgle
   end
 
@@ -171,16 +168,17 @@ class Burgle
       exit
     end
 
-    execute_extra_scripts($burgle_before_scripts)
+    execute_extra_scripts(@burgle_before_scripts)
     get_entry(@entry_type)
     wait_for_script_to_complete('buff', ['burgle']) if @settings.waggle_sets['burgle']
 
+    fix_standing
     #ensure you're in hiding or invis before starting
     if !invisible?
-      hide_attemps = 3
+      hide_attempts = 3
       until hide?
-        hide_attemps -= 1
-        if hide_attemps <= 0
+        hide_attempts -= 1
+        if hide_attempts <= 0
           message("Couldn't hide.  Find a better room.")
           return
         end
@@ -207,7 +205,8 @@ class Burgle
     #make hands empty again, as they likely have loot/entry method in them
     store_loot
 
-    #do afters here?
+
+    execute_extra_scripts(@burgle_after_scripts)
 
     case @loot_type
     when 'drop'
@@ -329,15 +328,10 @@ class Burgle
     return if Flags['burgle-footsteps']
     #ensure you're in hiding or invis before looting -> only try once because of timer
     if !invisible?
-      hide_attemps = 1
-      until hide?
-        hide_attemps -= 1
-        if hide_attemps <= 0
-          message("Couldn't hide.  Searching to avoid delays.")
-          break
-        end
-      end
+      hide?
+      message("Couldn't hide.  Searching to avoid delays.") unless hidden?
     end
+
     #don't search if you've seen footsteps and are in safe_mode
     return if Flags['burgle-footsteps']
     result = bput("search #{target}",'It looks valuable','Roundtime','I could not')
@@ -426,22 +420,21 @@ class Burgle
     return nil
   end
 
-end
-
-#ripped out of hunting-buddy - not part of class definition so it can be called in before_dying
-def execute_extra_scripts(extra_scripts)
-  extra_scripts.each do |script|
-    DRC.message "***STATUS*** EXECUTE #{script}"
-    script_args = script.split(' ')
-    script_name = script_args.shift
-    DRC.wait_for_script_to_complete(script_name, script_args)
+  #ripped out of hunting-buddy
+  def execute_extra_scripts(extra_scripts)
+    extra_scripts.each do |script|
+      DRC.message "***STATUS*** EXECUTE #{script}"
+      script_args = script.split(' ')
+      script_name = script_args.shift
+      DRC.wait_for_script_to_complete(script_name, script_args)
+    end
   end
+
 end
 
 
 
 before_dying do
-  execute_extra_scripts($burgle_after_scripts)
   #resume the previously paused scripts
   Script.running.find_all { |s| s.paused? && !s.no_pause_all && $scripts_to_unpause.include?(s.name) }.each(&:unpause)
   Flags.delete('burgle-footsteps')
--- a/burgle.lic
+++ b/burgle.lic
@@ -13,77 +13,166 @@ class Burgle
   include DRCT
   include DRCI
 
-
   def initialize
 
     arg_definitions = [
       [
-        { name: 'start', regex: /start/i, description: 'Required: prevents accidentally running burgle and protects against burgle typo for starting script' },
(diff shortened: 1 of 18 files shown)


vtcifer:
New example yaml.  Will work on updating wiki next:

```yaml
burgle_settings:
  room: nnn                                 # pick your own room, no guards should ever come here
  entry_type: lockpick                      # lockpick | rope | cycle
  use_lockpick_ring: true                   # if you use a worn lockpick stacker
  lockpick_container: lockpick ring         # name of your lockpick stacker
  max_search_count: 2                       # max number of total searches to do including retries
  retry: false                              # retry if you don't find loot.  default: false
  loot: keep                                # drop | keep | bin | pawn   default: keep (bin and pawn are not supported currently)
  loot_container: duffel bag                # container to keep loot in while burglering
  safe_mode: true                           # stop searching and try to get out as quickly as possible when hearing footsteps.  Default: true
  room_blacklist:                           # array of rooms to ignore. Case sensitive.
    - Bedroom
  item_whitelist:                           # array of full item names (adj + noun) of items to keep no matter what.  Case Sensitive.  
    - worn book
    - memory orb
    - blunt-tipped bolts
    - blunt-tipped arrows
    - blunt stones
  before:                                   # array of scripts to run, similar to hunting-buddy, before the burgle (will be called right before buffing)
    - yourscripthere with args here
  after:                                    # array of scripts to run, similar to hunting-buddy, after the burgle, (will be called right before script ends)
    - anotherscript here
```

vtcifer:
So I've got a couple of reports of before/after causing problems, but not sure what difference in environments could cause this:

```
--- Lich: error: undefined method 'execute_extra_scripts' for #<Scripting::Burgle:0x2bc4ef8>
    burgle:171:in 'burgle'
    burgle:98:in 'initialize'
```

BinuDR:
Move the execute_extra_scripts method into the class and call it as the last item in initialize.

vtcifer:
> Move the execute_extra_scripts method into the class and call it as the last item in initialize.

If it's within initialize, it could not be called due to some path's just exiting , that's why it's specifically outside of the class definition so that it can get called in before_dying.  

Plus it works in my environment, so I'd like to know why it's failing for others so that it can be fixed to ideal behavior.

burgle.lic
```diff
@@ -13,77 +13,166 @@ class Burgle
...
+      message("This is a dangerous script to run if it's not understood.  If you get caught there are very high fines, and the loss of all your items is possible if you can't pay your debt.")
       pause 10
       exit
     end
-
-    fix_standing
     burgle
   end
 
   def valid_burgle_settings?
```
rpherbig: 💡 These would also make good validators

burgle.lic
```diff
@@ -13,77 +13,166 @@ class Burgle
...
     start_script('jail-buddy') unless Script.running?('jail-buddy')
+
     @settings = get_settings
     @burgle_settings = @settings.burgle_settings
+    @loot_container = @burgle_settings['loot_container']
+    @use_lockpick_ring =  @burgle_settings['use_lockpick_ring']
+    @lockpick_container =  @burgle_settings['lockpick_container']
+
+    #set yaml settings unless being overridden from the command line arguments
+    @entry_type = @burgle_settings['entry_type'] unless @entry_type
```
rpherbig: 💡 I _think_ you can simplify this line to `@entry_type |= @burgle_settings['entry_type']`.

I _think_ you could instead simplify line 29 to `@entry_type = args.entry || @burgle_settings['entry_type']` (assuming `@burgle_settings` has already been defined).

burgle.lic
```diff
@@ -13,77 +13,166 @@ class Burgle
...
     ]
 
     args = parse_args(arg_definitions)
 
+    @entry_type = args.entry if args.entry
+    @burgle_room = args.roomid if args.roomid
+    @loot_type = args.loot_type if args.loot_type
+
+    #hash which maps directions to their reverse, for backing out of the house
+    @reverse_direction_map = {
```
rpherbig: 💡 We do the same thing in https://github.com/rpherbig/dr-scripts/blob/4fcbc37f7eb85f733643ea43b102adb00c2bc1cc/dusk-labyrinth.lic#L22

We're not at the point I'm insisting on commonizing this, but it's something to keep an eye on.

burgle.lic
```diff
@@ -112,22 +201,193 @@ class Burgle
...
 
   def search_for_loot(target)
-    if /It looks valuable/ =~ bput("search #{target}",'It looks valuable','Roundtime','I could not')
-      if @burgle_settings['loot'] == "drop"
-        drop_item(DRC.right_hand)
+    #don't hide/search if you've seen footsteps and are in safe_mode
+    return if Flags['burgle-footsteps']
+    #ensure you're in hiding or invis before looting -> only try once because of timer
+    if !invisible?
+      hide_attemps = 1
```
rpherbig: 💡 Typo: `hide_attemps` -> `hide_attempts`

burgle.lic
```diff
@@ -112,22 +201,193 @@ class Burgle
...
+    end
   end
 
   def search_for_loot(target)
-    if /It looks valuable/ =~ bput("search #{target}",'It looks valuable','Roundtime','I could not')
-      if @burgle_settings['loot'] == "drop"
-        drop_item(DRC.right_hand)
+    #don't hide/search if you've seen footsteps and are in safe_mode
+    return if Flags['burgle-footsteps']
+    #ensure you're in hiding or invis before looting -> only try once because of timer
```
rpherbig: ❔ If you're only going to try once, why have a loop?

burgle.lic
```diff
@@ -151,6 +415,7 @@ class Burgle
     end
   end
 
+  #ripped out of steal.lic
```
rpherbig: ❔  Do we need a `common-thieving` or `common-stealing` (or you pick the name)?

vtcifer: While that comes out of steal.lic, I would think it would be better in common-items.  It's really just putting an item away or dropping it if unsuccessful.

burgle.lic
vtcifer: Legacy from just c/p from the same code prior to the burgle call.  I had thought about either putting in a value to let users pick how many times it tried, but realized that's probably a bad idea.  

I also thought about updating hide? to specify a number of times to try before giving up, but hide? is more complex than I originally thought.

burgle.lic
vtcifer: The first one doesn't seem to work when I tested.
I get an validation error (@entry_type gets set to a boolean).

```ruby
    @entry_type |= @burgle_settings['entry_type']
    echo @entry_type
```

```
,burgle start rope 
--- Lich: burgle active.
[burgle: true]
| Invalid burgle_settings:entry_type setting.
| It is very important that you check the documentation before running: https://elanthipedia.play.net/Lich_script_repository#burgle 
| This is a dangerous script to run if it's not understood.  If you get caught there are very high fines, and the loss of all your items is possible if you can't pay your debt.
```

The second does seem to work:
```ruby
    @entry_type = args.entry || @burgle_settings['entry_type']
    @burgle_room = args.roomid || @burgle_settings['room']
    @loot_type = args.loot_type ||  @burgle_settings['loot']
```

vtcifer:
> > Move the execute_extra_scripts method into the class and call it as the last item in initialize.
> 
> If it's within initialize, it could not be called due to some path's just exiting , that's why it's specifically outside of the class definition so that it can get called in before_dying.
> 
> Plus it works in my environment, so I'd like to know why it's failing for others so that it can be fixed to ideal behavior.

Is it possible something with how dependency interacts with scripts is causing this?

rpherbig:
Nothing comes to my mind. @rcuhljr any ideas?

burgle.lic
rpherbig: Sounds reasonable to me. Not necessary for this PR, mind you, just something to consider for the future.

rpherbig:
> I know when I've modified some of the common-* files, I've needed to kill dependency and restart it for my changes to take effect.

You should be able to just re-run the `common-*` file you changed (which loads it into memory), then re-run the script that uses the common file.

-----

✔️ Changes look good to me, let me know how the debugging goes and when you're ready for this to be merged.

vtcifer:
I moved the before/after into the class definition for now, that should catch it almost all the time, except where something odd happens like your entry method disappears between the check and actually burglarizing. 

Should be gtg. I’ll update wiki in a little while.

burgle.lic
```diff
@@ -136,13 +390,17 @@ class Burgle
...
+      return false
     end
 
     false
   end
 
+  #ripped out of steal.lic
   def drop_item(item)
-    case bput("drop my #{item}", 'You drop', 'would damage it', 'smashing it to bits', 'Something appears different about', 'What were you')
+    case bput("drop my #{item}", 'You drop', 'You wince', 'would damage it', 'smashing it to bits', 'Something appears different about', 'What were you')
```
KatoakDR, line 403: @vtcifer Do you recall the game message that led to adding `You wince` to the list of match strings?

Per https://github.com/rpherbig/dr-scripts/pull/4959/files#r634241388 and https://github.com/rpherbig/dr-scripts/pull/4161/files#r382959034, I'm working on updates and tests for `common-items` so that `burgle` and `stealing` can switch to `DRCI.dispose_trash(item)`. On that journey, I want to make sure that I've got all the use cases handled.

Thanks

cc @rpherbig

BinuDR: It's from when you drop something that breaks or gets damaged.
Instruments, gaethzen, telescopes.  One of those.

PR #4177, opened 2020-03-06
common-travel.lic - small fix for reverse_path()

BinuDR:
If it's sent a a direction as a string instead of an array, it would return an array of nils.  This takes care of that.

```
;e echo DRCT.reverse_path('east')
--- Lich: exec1 active.
[exec1: [nil, nil, nil, nil]]
--- Lich: exec1 has exited.
```

After change:
```
>;common-travel
;e echo DRCT.reverse_path('east')
--- Lich: exec1 active.
[exec1: ["west"]]
--- Lich: exec1 has exited.
```

### DIFF
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -161,6 +161,7 @@ module DRCT
   end
 
   def reverse_path(path)
+    path = path.to_a unless path.class == Array
     dir_to_prev_dir = {
       'northeast' => 'southwest',
       'southwest' => 'northeast',
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -161,7 +161,6 @@ module DRCT
   end
 
   def reverse_path(path)
-    path = path.to_a unless path.class == Array
     dir_to_prev_dir = {
       'northeast' => 'southwest',
       'southwest' => 'northeast',
@@ -177,6 +176,10 @@ module DRCT
     reverse_path = []
     path = path.reverse
     for i in 0..path.length-1
+      if dir_to_prev_dir[path[i]].nil?
+        DRC.message("Error: No reverse direction found for #{path[i]}.  Please use the full direction, northeast instead of ne, check your spelling, and make sure the path parameter is an array.")
+        exit
+      end
       reverse_path.push(dir_to_prev_dir[path[i]])
     end
     return reverse_path


rpherbig:
This returns the direction as a single-element array, not a string as I'd expect (since I passed in a string).

Does the consumer handle the array return type correctly?

BinuDR:
I was thinking of doing it another way, but I thought having a consistent return type might be better.

I was going to do this.
```
  def reverse_path(path)
    path = path.to_a unless path.class == Array
    dir_to_prev_dir = {
      'northeast' => 'southwest',
      'southwest' => 'northeast',
      'northwest' => 'southeast',
      'southeast' => 'northwest',
      'north' => 'south',
      'south' => 'north',
      'east' => 'west',
      'west' => 'east',
      'up' => 'down',
      'down' => 'up'
    }
    return dir_to_prev_dir[path] if path.class != Array
    reverse_path = []
    path = path.reverse
    for i in 0..path.length-1
      reverse_path.push(dir_to_prev_dir[path[i]])
    end
    return reverse_path
  end
end
```

BinuDR:
No scripts use this method yet as far as I know.

Do you think it would be better to always return an Array for consistency, or to have it return the same class it receives?

rpherbig:
I greatly prefer a consistent return type (Array).

Similarly, I don't think we should have special handling for a String. The caller can just pass in `[foo]` if they really only have one direction to reverse.

BinuDR:
If they do pass in a string, it needs to either return an error or nothing because currently it returns an array of nils and that's just confusing.

rpherbig:
Ruby's duck typing is the issue here, since a String responds to `reverse` and `index`. Because of that we reverse the char array representation and then lookup 'e', 'a', 's', and 't' consecutively, which are all `nil`.

I admit it's not a great development experience, and I'm not sure what the best answer is. I'm leaning toward rejecting Strings.

BinuDR:
How about this as the first line?
`return false unless path.class == Array`

rpherbig:
What do you think about raising an error inside the loop if we try to lookup a direction that doesn't exist?

BinuDR:
That would be fine with me.  How do you prefer the error being relayed?  Echo an error message and then return false?

BinuDR:
How about something like this?
```
    for i in 0..path.length-1
      if dir_to_prev_dir[path[i]].nil?
        DRC.message("No reverse direction found for #{path[i]}.  Path parameter should be an array.")
        break
      end
      reverse_path.push(dir_to_prev_dir[path[i]])
    end
```

rpherbig:
> Path parameter should be an array

Failing the lookup doesn't mean the input was a non-Array, it means it was something unexpected (the more general case). For example, 'ewst'.

-----

I'm inclined to `exit` instead of `break`. I don't think there's much recovery we can do in this case, and returning a partial path doesn't seem useful.

BinuDR:
Yeah, I just updated it for that too.

BinuDR:
Ok exit instead of break.  Right, that makes sense as a typo would be another possible error.

BinuDR:
How's this?
```
    for i in 0..path.length-1
      if dir_to_prev_dir[path[i]].nil?
        DRC.message("Error: No reverse direction found for #{path[i]}.  Please use the full direction, northeast instead of ne, check your spelling, and make sure the path parameter is an array.")
        exit
      end
      reverse_path.push(dir_to_prev_dir[path[i]])
    end
```

rpherbig:
✔️ Looks good to me

BinuDR:
Ready to merge

PR #4181, opened 2020-03-14
trade - support stables and non-knit outfitting

Tarhish:
After running with this for awhile, it looks like this should be working.  Two big changes and a number of changes to condense code or minimize certain calls.

1.) if trade_use_stables is set, it will attempt to store and recover caravans from stables, by either checking the last_seen_caravan uservar to see where it left it last run, or, if it's not there, checking the nearest stable besides that.  Otherwise it will pause and let the user find the darn thing.  This setting means it won't clean out the caravan's storage box either, since it's not necessary.  It also handles variable fees.

2.) if outfitting is above a certain rank (or if the user defines a caravan_recipes: setting without 'knitted' in the title) it will purchase and combine burlap cloth, check straight pins and thread inventories, and keep them stocked up for use with non-knitted outfitting.

I'll get some people to test the more recent changes with me.

### DIFF
--- a/trade.lic
+++ b/trade.lic
@@ -127,7 +127,7 @@ class Trade
     end
     
     if Room.current.id.nil?
-      DRC.message("The script doesn't know what room it's in!  Move one room away to fix the map!")
+      DRC.message("The script doesn't know what room it's in!  Manually move one room away and return, then everything should resume automatically!")
       pause 1 while Room.current.id.nil?
     end
 
@@ -140,13 +140,21 @@ class Trade
     @equipment_manager.empty_hands
 
     if args.destination
-      echo "Taking the caravan directly to room #{args.destination}"
+      unless caravan_exists?
+        DRC.message("No caravan found.  Call ;trade without a destination, or grab a caravan manually")
+        exit
+      end
+      DRC.message("Taking the caravan directly to room #{args.destination}")
       take_caravan_to(args.destination.to_i)
       exit
     elsif args.outpost
+      unless caravan_exists?
+        DRC.message("No caravan found.  Call ;trade without a destination, or grab a caravan manually")
+        exit
+      end
       town_dest = find_town_from_arg(args.town_name)
       unless town_dest
-        echo 'TOWN NOT FOUND OR NOT SUPPORTED.  CURRENTLY SUPPORTS ZOLUREN, ILITHI, FORFEDHAR ONLY'
+        DRC.message('TOWN NOT FOUND OR NOT SUPPORTED.  CURRENTLY SUPPORTS ZOLUREN, THERENGIA, ILITHI, FORFEDHAR ONLY')
         exit
       end
       @current_caravan_destination = outpost_from_name(town_dest)
@@ -202,7 +210,6 @@ class Trade
   end
 
   def start_up # should parse contracts, check mats, handle expireds, get the caravan to a useful outpost, deliver any contracts, then begin the loop.
-    caravan_exists?
     if wealth(@settings.hometown) < @caravan_coins_on_hand
       room = Room.current.id
       command_caravan?('wait') if caravan_exists?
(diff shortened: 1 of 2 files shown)


trade.lic
```diff
@@ -121,6 +125,11 @@ class Trade
       echo 'WAITING FOR BESCORT TO FINISH'
       pause 1 while Script.running?('bescort')
     end
+    
+    if Room.current.id.nil?
+      DRC.message("The script doesn't know what room it's in!  Move one room away to fix the map!")
```
rpherbig: 💡 'Manually move one room...'?

rpherbig: 💡 '... and the script will automatically resume'?

trade.lic
```diff
@@ -193,9 +202,18 @@ class Trade
   end
 
   def start_up # should parse contracts, check mats, handle expireds, get the caravan to a useful outpost, deliver any contracts, then begin the loop.
-    get_caravan unless caravan_exists?
-    fput('look')
+    caravan_exists?
```
rpherbig: ❔ What is this line doing?

trade.lic
```diff
@@ -1134,8 +1222,7 @@ class Trade
 
   def caravan_exists? # do you have a caravan at all?
     recall_caravan? unless @caravan.adjective && @caravan.noun
-    return recall_caravan? unless @caravan.adjective && @caravan.noun
-    true
+    @caravan.adjective && @caravan.noun ? true : false
```
rpherbig: 💡 You can simplify this to `return @caravan.adjective && @caravan.noun`

trade.lic
```diff
@@ -228,11 +264,33 @@ class Trade
...
+    turn_in_items
+    DRC.message('Heading out to find a stable for the night')
+    take_caravan_to(DRCT.tag_to_id('caravan_stable'))
+    store_item_in_box('balsa lumber') if @caravan_training_skills.include?('Engineering') && DRCI.exists?('balsa lumber')
+    store_item_in_box(@outfitting_material_type) if @caravan_training_skills.include?('Outfitting') && DRCI.exists?(@outfitting_material_type)
+    /(\d+) (Kronars|Lirums|Dokoras) for a caravan/ =~ bput('pay', 'The cost for stabling .*')
+    case bput("pay #{Regexp.last_match(1)}", 'The price for stabling and feeding .*', 'stable.+ who runs up and leads your caravan')
+    when /The price for stabling and feeding/
+      DRC.beep
+      DRC.message('Something has gone wrong!  Put in an issue at the github with a log of this event!')
```
rpherbig: 💡 '... at https://github.com/rpherbig/dr-scripts/issues/new ...'?

trade.lic
```diff
@@ -1558,126 +1646,233 @@ class Trade
...
+    order_item(crafting_data['stock-room'], 13)
     bput('combine my yarn', 'You combine', 'You must be holding both')
-    @current_yarn_count = bput('count my wool yarn', 'You count out \d+ yards').scan(/\d+/).first.to_i
-    buy_yarn(true) if @current_yarn_count < @yarn_quantity
+    count_materials('wool yarn')
+    buy_yarn(crafting_data) if @current_materials[@outfitting_material_type] < @outfitting_quantity
+    stow_hands
+  end
+
+  def buy_cloth(crafting_data)
```
rpherbig: ❓ Should any of this logic be in common-crafting?

rpherbig:
Done reviewing, let me know if you need it merged now or can look at some of the comments.

trade.lic
Tarhish: Absolutely nothing anymore.  Good catch.  It also alerted me to a failure if someone tried to do the 'direct to room or outpost travel' without a caravan.

trade.lic
Tarhish: Yeah, the plan is definitely to commonize some of this and make it work for ;craft.  But I don't yet know the scope of that.  I'll look at that later.

Tarhish:
I think this is good to go.  I had a couple people running this with no problems, and the only thing I really had to change in the last month was generalizing that stable regex a bit more.  Either way, I'll be around if someone has problems with it.  When it goes live I can update the documentation.

PR #4185, opened 2020-03-15
Update Rising Mists Expiration

joelim01:
The current message is:

The unnatural fog breaks apart, fading into barely perceptible wisps before vanishing altogether. 

This may only be when Zephyr is currently up as well, but I think we can be more general here.


joelim01:
Oops! Being addressed in https://github.com/rpherbig/dr-scripts/pull/4182

PR #4186, opened 2020-03-15
Allow spells that expire to also have recast_every

joelim01:
The way the logic was working, spells with expire messages would not be re-castable even if they had recast_every: or recast: arguments. This is especially confusing when spells have expiration messages in base_spells.

Proposed change:

Allow any type of trigger for re-casting if the spell has the arguments and meets the re-cast criteria. I've separated pet casting out because I don't know enough about it to be sure that putting it in with the other expiration conditions will not cause trouble.


joelim01:
Closing while I investigate if it makes sense to recast these spells before they expire...

PR #4195, opened 2020-03-20
Allow attack_overrides on MH thrown

Hiinky:
Allow attack_overrides on MH thrown

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3706,10 +3706,10 @@ class GameState
   end
 
   def thrown_attack_verb
-    if bound_weapon? && !@use_weak_attacks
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -3708,6 +3708,8 @@ class GameState
   def thrown_attack_verb
     if bound_weapon? && !@use_weak_attacks
       'hurl'
+    elsif attack_override
```
rpherbig: ❓ Why isn't `attack_override` the first check?

Hiinky: I was being lazy. :D  fixed now.

PR #4210, opened 2020-04-05
Update movement between burgle and pick

Hiinky:
Users mentioned they're picking boxes in the wrong room after ;burgle before ;pick if they don't have a lockpick_room_id selected.  Thievery,athletics, stealth do not need specific rooms.

### DIFF
--- a/crossing-training.lic
+++ b/crossing-training.lic
@@ -338,6 +338,7 @@ class CrossingTraining
     return if @researching
     wait_for_script_to_complete('burgle',['start']) if @settings.train_with_burgle
     return if DRSkill.getxp('Locksmithing') >= 30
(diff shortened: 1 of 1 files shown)


crossing-training.lic
```diff
@@ -338,6 +338,7 @@ class CrossingTraining
     return if @researching
     wait_for_script_to_complete('burgle',['start']) if @settings.train_with_burgle
     return if DRSkill.getxp('Locksmithing') >= 30
+    walk_to(@training_room) unless @settings.lockpick_room_id
```
rpherbig, line 341: 💡 I'd prefer to see this as an if/else block, it would be a bit clearer.

PR #4213, opened 2020-04-09
Added a firing timer for bow, crossbow, and sling.

Dartellum:
This also helps with the load/aim loop cycle as it only lets it get to the delay limit, in seconds, before shooting. This loop is caused when the flag, ct-ranged-ready, is not triggered. I even tried putting the period on the end to see if that helped but it did not. I put the time as 15 seconds for a fully aimed shot that will shoot the weapon if hit. I would like some others to test - Kali was testing on several characters. 

Also, added a loot_delay and/or for those of use that need the longer wait to loot due to slow decay of some creatures.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -411,7 +411,7 @@ class LootProcess
     @gem_pouch_adjective = settings.gem_pouch_adjective
     echo("  @gem_pouch_adjective: #{@gem_pouch_adjective}") if $debug_mode_ct
 
-    @loot_delay = 5 || settings.loot_delay # in seconds
+    @loot_delay = settings.loot_delay
     @loot_timer = Time.now - @loot_delay
     echo("  @loot_delay: #{@loot_delay}") if $debug_mode_ct
     echo("  @loot_timer: #{@loot_timer}") if $debug_mode_ct
@@ -2577,9 +2577,10 @@ class AttackProcess
     @use_overrides_for_aiming_trainables = settings.use_overrides_for_aiming_trainables
     echo(" @use_overrides_for_aiming_trainables: #{@use_overrides_for_aiming_trainables}") if $debug_mode_ct
 
-    @firing_delay = settings.firing_delay || 15 # Delay for shooting aimed weapons.
-    @firing_timer = Time.now - @firing_delay
+    @firing_delay = settings.firing_delay
     echo(" @firing_delay: #{@firing_delay}") if $debug_mode_ct
+    @firing_timer = Time.now - @firing_delay
+    echo("@firing_timer: #{@firing_timer}") if $debug_mode_ct
     @firing_check = 0
 
     Flags.add('ct-aim-failed', 'you stop aiming', 'stop concentrating on aiming')
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -411,8 +411,9 @@ class LootProcess
     @gem_pouch_adjective = settings.gem_pouch_adjective
     echo("  @gem_pouch_adjective: #{@gem_pouch_adjective}") if $debug_mode_ct
 
-    @loot_delay = 5 # in seconds
+    @loot_delay = 5 || settings.loot_delay # in seconds
     @loot_timer = Time.now - @loot_delay
+    echo("  @loot_delay: #{@loot_delay}") if $debug_mode_ct
     echo("  @loot_timer: #{@loot_timer}") if $debug_mode_ct
 
     @loot_bodies = settings.loot_bodies
@@ -2119,7 +2120,7 @@ class TrainerProcess
   end
 
   private
-  
+
   def hide_action(game_state)
     if hide?(@hide_type) && !@dont_stalk && !game_state.npcs.empty? && @hide_type == 'hide'
       stalked = bput('stalk', 'Stalk what',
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -2575,9 +2576,14 @@ class AttackProcess
 
     @use_overrides_for_aiming_trainables = settings.use_overrides_for_aiming_trainables
     echo(" @use_overrides_for_aiming_trainables: #{@use_overrides_for_aiming_trainables}") if $debug_mode_ct
-    
+
+    @firing_delay = settings.firing_delay || 15 # Delay for shooting aimed weapons.
+    @firing_timer = Time.now - @firing_delay
+    echo(" @firing_delay: #{@firing_delay}") if $debug_mode_ct
```
rpherbig, line 2582: ❌ Please also echo `firing_timer`

rpherbig, line 2582: ❌ Please also echo `firing_timer`

combat-trainer.lic
```diff
@@ -2753,7 +2763,7 @@ class AttackProcess
       end
       game_state.loaded = false
       waitrt?
-    elsif game_state.loaded && !game_state.aiming_trainables.empty?
+    elsif game_state.loaded && !game_state.aiming_trainables.empty? #&& (Time.now - @firing_timer).to_i < @firing_delay
```
rpherbig, line 2766: ❓ Please either uncomment this or remove the comment

combat-trainer.lic
```diff
@@ -2713,12 +2719,16 @@ class AttackProcess
...
+    echo("Time now: #{Time.now.to_i}") if $debug_mode_ct
+    echo("@firing_timer: #{@firing_timer}") if $debug_mode_ct
+    echo("@firing_delay: #{@firing_delay}") if $debug_mode_ct
     case bput(command, 'isn\'t loaded', 'There is nothing', 'But your', /you (fire|poach|snipe) an? (?:.*\s)?(arrows?|bolt|shard|rock|sphere|clump|coral|fist|holder|lump|patella|pellet|pulzone|quadrello|quill|stone|stopper|verretto|crumb|spine|mantrap spike|dragon|icicle|fang|scale|grey-black spike|bacon strip|page|naga|thorn).* at/i, 'I could not find', 'with no effect and falls to the ground', 'Face what', 'How can you (poach|snipe)', 'you don\'t feel like fighting right now', 'That weapon must be in your right hand to fire')
     when /How can you (poach|snipe)/
       shoot_aimed('shoot', game_state)
     when /you (fire|poach|snipe) an? (?:.*\s)?(arrows?|bolt|shard|rock|sphere|clump|coral|fist|holder|lump|patella|pellet|pulzone|quadrello|quill|stone|stopper|verretto|blowgun dart|crumb|spine|mantrap spike|dragon|icicle|fang|scale|grey-black spike|bacon strip|page|naga|thorn).* at/i
       game_state.action_taken
       ammo = Regexp.last_match(2)
+      @firing_check = 0
```
rpherbig, line 2731: ❓ I think you can get rid of `@firing_check` if you set `@firing_timer = Time.now` here (and get rid of line 2848).

What do you think?

combat-trainer.lic
```diff
@@ -411,8 +411,9 @@ class LootProcess
     @gem_pouch_adjective = settings.gem_pouch_adjective
     echo("  @gem_pouch_adjective: #{@gem_pouch_adjective}") if $debug_mode_ct
 
-    @loot_delay = 5 # in seconds
+    @loot_delay = 5 || settings.loot_delay # in seconds
```
rpherbig, line 414: ❓ Shouldn't this be `settings.loot_delay || 5`?

Regardless, this should instead be done by adding `loot_delay` to `base.yaml` with a default value of 5.

combat-trainer.lic
Dartellum: I was testing if it was needed. I do not believe it is and can remove.

combat-trainer.lic
```diff
@@ -2575,9 +2576,14 @@ class AttackProcess
 
     @use_overrides_for_aiming_trainables = settings.use_overrides_for_aiming_trainables
     echo(" @use_overrides_for_aiming_trainables: #{@use_overrides_for_aiming_trainables}") if $debug_mode_ct
-    
+
+    @firing_delay = settings.firing_delay || 15 # Delay for shooting aimed weapons.
```
rpherbig, line 2580: ❓ Is this the same case as `loot_delay` above? I.e. it should instead be in `base.yaml`.

combat-trainer.lic
Dartellum: I can do that. My thought was wold not need to add to base.yaml as the either/or would catch.

combat-trainer.lic
Dartellum: The `@firing_check` catches the loop because if it is the second time in we do not want `@firing_timer` to change.

combat-trainer.lic
Dartellum: Added.

combat-trainer.lic
rpherbig: Every setting needs to be in `base.yaml`.

combat-trainer.lic
rpherbig: How much time passes between this line and line 2849?

rpherbig:
@Dartellum Can you get a volunteer from Slack to test this?

Dartellum:
> @Dartellum Can you get a volunteer from Slack to test this?

@Hiinky did testing with several characters and may have passed the changes on to others already.

Hiinky:
I ran it for several days before I accidentally overrode the changes.  I had no snags and ranged was working nicely.

PR #4217, opened 2020-04-12
Update to support new burgle books

Hiinky:
support new newb compendiums with nouns guide/manual


rpherbig:
In hindsight, I think we should have gone with a shared setting for the noun. But changing a setting name is hard, so this is fine.

PR #4228, opened 2020-04-18
Expand/Fix Favor Support

Hiinky:
- Altar at 8128 was incorrectly named for Hodierna.  Updated to Phelim.
- Update base-town rooms for northern and southern puzzle rooms.
- Update favor to read hometown information and added bug handling for window puzzle.
- Removed Validate error messaging - no longer applies.

### DIFF
--- a/favor.lic
+++ b/favor.lic
@@ -20,6 +20,10 @@ class CrossingFavor
     args = parse_args(arg_definitions, true)
     @settings = get_settings
     @hometown = @settings.hometown
+    @town_data = get_data('town')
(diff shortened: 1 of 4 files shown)


favor.lic
```diff
@@ -135,6 +137,8 @@ class CrossingFavor
       leave_room
     when 'ancient window'
```
rpherbig, line 138: 💡 You could instead do this:

```ruby
    when 'ancient window', 'You cannot go through a closed window'
      window  
```

PR #4240, opened 2020-04-23
Added Effective Scholarship Chart Selection

Hiinky:
- add scholarship modifier to control charts selected
   - add control over scholarship modifier
- add control over number of charts selected
- add second attempt if first read unsuccessful
- added notes to the wiki to explain usage and theory: https://github.com/rpherbig/dr-scripts/wiki/First-Aid-Strategy

I've tested these charts extensively.  Years really.  The new scholarship scale will work for the vast majority of users and yaml settings will sort out the rest.

### DIFF
--- a/first-aid.lic
+++ b/first-aid.lic
@@ -26,7 +26,7 @@ class FirstAid
       @booktype = @settings.textbook_type
       textbook_charts
     else
-      @booktype = 'compendium'
+      @booktype = @settings.compendium_type
       compendium_charts
     end
   end
@@ -36,8 +36,8 @@ class FirstAid
       pause @performance_pause # to give performance time to complete before_dying if stopped from the previous script
       start_script('performance', ['noclean']) unless Script.running?('performance')
     end
-    bput('get my compendium', 'You get', 'You are already holding')
-    bput('look my compendium', '^The compendium lies open to the section on .* physiology')
+    bput("get my #{@booktype}", 'You get', 'You are already holding')
+    bput("look my #{@booktype}", "^The #{@booktype} lies open to the section on .* physiology")
     pause
     compendium_charts = reget(40).grep(/^   .+$/).map(&:strip)
     charts_to_read = @chart_data.select { |name, _info| compendium_charts.include?(name) }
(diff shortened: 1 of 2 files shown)


first-aid.lic
```diff
@@ -56,22 +57,56 @@ class FirstAid
...
-      .take(20)
+      .take(@num_charts)
       .each do |info|
-        case bput("turn my #{@booktype} to #{info['index']}", 'You turn', 'That section does not exist', 'Turn what?')
+        case bput("turn my #{@booktype} to #{info['index']}", 'You turn', 'That section does not exist', 'Turn what?','almost impossible to do')
         when 'You turn'
-          case bput("study my #{@booktype}", 'you begin to study', 'you begin studying', 'With a sudden moment of', 'In a sudden moment of clarity', '^Why ', 'you continue to study', 'you attempt', 'you continue studying', 'You need to be holding')
+          case bput("study my #{@booktype}", 'gradually absorbing', 'difficult time comprehending the advanced text', 'suddenly makes sense to you', '^Why ', 'You need to be holding', 'discerned all you can')
+          when 'gradually absorbing'
+            loop{ message = bput("study my #{@booktype}", /Roundtime/, /makes sense/, /discerned all you can/ ); break if message == 'makes sense' || 'discerned all you can'}
```
jandersson, line 69: I think there might be a risk for infinite loop here. Would something like `5.times.do` instead of `loop` make more sense here?

Hiinky: oddly it jumps out after just one retry. I've never had it even try a third but I don't see why we can't limit it just in case.

jandersson: Just being wary of something that can potentially loop forever given an edge case. Could be unfounded :)

jandersson:
This is working just fine for me and is teaching much better than previously. Good stuff

jandersson:
You might want to consider updating the documentation https://elanthipedia.play.net/Lich_script_repository#first-aid or linking to the wiki you made at the top of first aid. Just so the new settings are more discoverable

rpherbig:
Thank you @Hiinky and @jandersson !

PR #4241, opened 2020-04-23
Workorders repair setting

Dartellum:
I left in the ability to run _;workorders <craft> repair_ if one so desires.


jandersson:
Good stuff. I have a feeling setting this new setting to false will also avoid some pickpocketing.

jandersson:
Don't forget to update https://elanthipedia.play.net/Lich_script_repository#workorders with the new setting :)

Dartellum:
> Don't forget to update https://elanthipedia.play.net/Lich_script_repository#workorders with the new setting :)

Once it is approved/merged. :)

PR #4247, opened 2020-04-26
Updated locksmithing script for trainer boxes

Hiinky:
Updated options across the board for locksmith training boxes to be supported by crossing-training without interrupting anyone's current settings.  Can be used independently or in t2 also.

add support for the daily charge boxes and incidental trainers / keepsake boxes from burgle.  I opted to keep live boxes handled by pick.lic specifically because it's kind of a beast and needs updating but folks really want to use the trainer boxes now since pet boxes are dead.  Settings in new locksmithing.lic will send args to pick.lic as needed.

Handles in order:
- Live boxes based on existing ;pick settings - selected first to clear inventory weight
- Permanent trainer boxes - unlimited or daily charges - selected second to use up available charges before using up consumables.
- Consumable trainer boxes such as burgled keepsake boxes, incidental items

### DIFF
--- a/locksmithing.lic
+++ b/locksmithing.lic
@@ -1,0 +1,116 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#locksmithing
+=end
+
+custom_require.call(%w[common common-items drinfomon equipmanager])
+ 
+class Locksmithing
+  include DRC
+  include DRCI
+  def initialize
+    settings = get_settings
+    @equipment_manager = EquipmentManager.new(settings)
+    @daily_trainer ||= settings.have_training_box
+    @box = settings.picking_lockbox || 'training box'
+    @worn_lockbox = settings.picking_worn_lockbox
+    @box_locked = false
+    @liveboxes = settings.pick_live_boxes
+    @consumable_trainers = settings.consumable_lockboxes
+    @consumable_container = settings.burgle_settings['loot_container']
+
+    @equipment_manager.empty_hands
+
+    live_boxes if @liveboxes
+    daily_lockbox if @daily_trainer
+    consumable_lockbox if @consumable_trainers
+  end
+
+  def daily_lockbox
+    if DRSkill.getxp('Locksmithing') < 34
+      if @worn_lockbox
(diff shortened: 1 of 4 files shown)


locksmithing.lic
```diff
@@ -0,0 +1,116 @@
...
+
+custom_require.call(%w[common common-items drinfomon equipmanager])
+ 
+class Locksmithing
+  include DRC
+  include DRCI
+  def initialize
+    settings = get_settings
+    @equipment_manager = EquipmentManager.new(settings)
+    @daily_trainer ||= settings.have_training_box
```
rpherbig, line 13: 💡 Why the `||=`? This is the first time it's assigned to.

locksmithing.lic
```diff
@@ -0,0 +1,116 @@
...
+    @box = settings.picking_lockbox || 'training box'
```
rpherbig, line 14: 💡 Default value shouldn't be necessary since you updated `base.yaml`.

profiles/base.yaml
```diff
@@ -721,10 +721,21 @@ mining_implement: shovel
...
+#https://elanthipedia.play.net/Lich_script_repository#locksmithing
+# use ;pick and pick settings first
+pick_live_boxes: true
+# second option to a daily use lockbox trainer like a training box or harvest bag
+have_training_box: false
+picking_lockbox: training box
+picking_worn_lockbox: false
+# consumable boxes will be pulled from your burgle loot_container and should be kept separate from live boxes.
+# list as many as you want - though it will try from the top down.
+consumable_lockboxes:
```
rpherbig, line 734: ❌ This should have an entry in `base-empty.yaml`

locksmithing.lic
Hiinky: Ahh some of these were in the base script I started with and I didn't look too close. editing now.

profiles/base.yaml
Hiinky: added.

locksmithing.lic
Hiinky: Ahh some of these were in the base script I started with and I didn't look too close. editing now.

PR #4254, opened 2020-04-27
gosafe.lic Walk to safe_room:

cfrisby77:
walk to safe_room
Purpose: Use in Before: / After: sections of yamls where anchors do not seem to work

### DIFF
--- a/gosafe.lic
+++ b/gosafe.lic
@@ -1,7 +1,7 @@
 =begin
 =end
 
-custom_require.call(%w[common common-travel])
+custom_require.call(%w[common-travel])
 
 class GoSafe
   
--- a/gosafe.lic
+++ b/gosafe.lic
@@ -1,4 +1,6 @@
 =begin
+	walks to safe_room:
+	Usage: Put in before / after sections of yaml settings where anchors do not seem to work.
 =end
 
 custom_require.call(%w[common-travel])
--- a/gosafe.lic
+++ b/gosafe.lic
@@ -1,0 +1,17 @@
+=begin
+=end
+
+custom_require.call(%w[common common-travel])
+
+class Gosafe
+  include DRC
+  include DRCT
+  include DRCI
+
+	def initialize
+		@settings = get_settings
+		walk_to @settings.safe_room
+	end
+  
+end
+Gosafe.new
--- a/gosafe.lic
+++ b/gosafe.lic
@@ -7,8 +7,8 @@ class GoSafe
 include DRCT
 
 	def initialize
-		@settings = get_settings
-		walk_to @settings.safe_room
+		settings = get_settings
+		walk_to settings.safe_room
 	end
   
 end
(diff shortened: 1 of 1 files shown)


gosafe.lic
```diff
@@ -0,0 +1,17 @@
+=begin
+=end
+
+custom_require.call(%w[common common-travel])
+
+class Gosafe
+  include DRC
+  include DRCT
+  include DRCI
```
rpherbig, line 9: ❌ Not in `custom_require` - it should be removed here or added there

gosafe.lic
```diff
@@ -0,0 +1,17 @@
...
+
+custom_require.call(%w[common common-travel])
+
+class Gosafe
+  include DRC
+  include DRCT
+  include DRCI
+
+	def initialize
+		@settings = get_settings
```
rpherbig, line 12: 💡 Does not need to be an instance var, consider scoping it locally

gosafe.lic
```diff
@@ -0,0 +1,17 @@
+=begin
+=end
+
+custom_require.call(%w[common common-travel])
+
+class Gosafe
```
rpherbig, line 6: 💡 I'd prefer `GoSafe`, but not a big deal

gosafe.lic
cfrisby77: Removed includes

gosafe.lic
cfrisby77: Changed to GoSafe

gosafe.lic
cfrisby77: could you provide an example of your suggestion here. Thank you

gosafe.lic
rpherbig: Sorry I wasn't clear, I was only referring to `DRCI` (although `DRC` is also not necessary). I _think_ `include DRCT` is still necessary or I would expect to see an error related to Ruby not knowing which module `walk_to` is in. Can you test this for me?

gosafe.lic
rpherbig: In Ruby, class instance variables are prefixed with `@`. Since `@settings` is never used outside of its method, I was suggested you scope it to the method instead of the class. Code-wise, all it means is removing the `@`.

```ruby
    settings = get_settings
    walk_to settings.safe_room
```

gosafe.lic
rpherbig: Thank you 😄

gosafe.lic
cfrisby77: Build updated. it did need the DRCT. 

Oddly it was working without it before i closed and reopened the FE. is there a way to reload the lich without closing the restarting the FE

gosafe.lic
cfrisby77: Ok, i had thought the @ was just part of each scripter's naming convention. removed them

gosafe.lic
cfrisby77: Ok, i think this is ready put in purpose in script documentation.

PR #4256, opened 2020-04-28
Script that simplifies the go2 drag syntax

cfrisby77:
Usage: ;drag <location> <person>

### DIFF
--- a/drag.lic
+++ b/drag.lic
@@ -1,0 +1,18 @@
+=begin
+  simple script to start dragging with go2 
+  Usage: ;drag <location> <person>
+=end
+
+if script.vars[2]
+	goto "drag"
+end
+goto "usage"
+
+
+drag:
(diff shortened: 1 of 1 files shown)


drag.lic
```diff
@@ -0,0 +1,18 @@
+=begin
+  simple script to start dragging with go2 
+  Usage: ;drag <location> <person>
+=end
+
+if script.vars[2]
```
rpherbig, line 6: ❌ I have a strong preference not to use `goto`

drag.lic
```diff
@@ -0,0 +1,18 @@
+=begin
+  simple script to start dragging with go2 
+  Usage: ;drag <location> <person>
```
rpherbig, line 3: ❌  We have built-in support for handling script args in a cleaner way.

cfrisby77:
Ok, will rethink this.

PR #4260, opened 2020-04-30
add handling for sell_loot yaml setting.

Hiinky:


locksmithing.lic
```diff
@@ -77,7 +77,7 @@ class Locksmithing
       start_time = Time.now
       wait_for_script_to_complete('pick')
       if Time.now - start_time > 15
-        wait_for_script_to_complete('sell-loot')
+        wait_for_script_to_complete('sell-loot') if sell_loot
```
rpherbig, line 80: ❓ Should this be `settings.sell_loot`? I don't see where `sell_loot` is defined

Hiinky: Fixed!

rpherbig:
@Hiinky I merged this before realizing you merged master into your branch. I think your merge was backwards. It should no longer be `settings.foo` and instead `@foo`. Can you look into this and make a new PR if needed?

rpherbig:
Nevermind, I did some investigation. It looks like this was already addressed in #4265, so I reverted this change in #4292. `master` should be all good to go. Let me know if I messed something up.

PR #4304, opened 2020-05-16
Add play.lic

jandersson:
*/ play an instrument
always play an instrument
   abuses bput /*

This is essentially a script that loops playing your zills. 

Its meant to be ran like the sanowret-crystal scrip; run it in the background and accumulate performance experience.

### DIFF
--- a/play.lic
+++ b/play.lic
@@ -1,0 +1,35 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#play
+=end
+class Play
+  def initialize
+    @no_play_scripts = get_settings.sanowret_no_use_scripts
+    @song = UserVars.song
+    main_loop
+  end
+
+  def play_song
+    fput("play #{@song}")
+  end
+
+  def can_play?
+    @no_play_scripts.none? { |script| Script.running?(script) }
+  end
+
+  def should_play?
+    !xp_highenough? && can_play?
+  end
+
+  def xp_highenough?
+    DRSkill.getxp('Performance') > 32
+  end
+
+  def main_loop
+    loop do
+      play_song if should_play?
+      pause 60
+    end
+  end
+end
+
+Play.new


play.lic
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def check_flags
+    Flags.add('play-clean', 'Your .* dirtiness may affect your performance') # Could check if the flag exists, but playing an instrument tells you if it needs cleaning
+    
+  end
+  
+  def play_song
+    fput("play #{@song} on zills")
+  end
```
Hiinky, line 20: I noticed when using this previously that you can remove "on zills" without needing to say what instrument for those wearing cowbells or similar.

jandersson: Yea this is a super old script. I'll change that. I'll make a note in the comments too.

jandersson:
So this is a super old script I wrote as a one-off. There is probably a lot to change. Look at it skeptically when reviewing it. I'll make the changes.

play.lic
```diff
@@ -0,0 +1,43 @@
+custom_require.call %w[events]
+class Play
+  def initialize
+    @song ||= UserVars.song
+    check_flags
+    main_loop 
+  end
+
+  def clean?
```
jandersson, line 9: Useless

play.lic
```diff
@@ -0,0 +1,43 @@
+custom_require.call %w[events]
```
jandersson, line 1: Useless

play.lic
```diff
@@ -0,0 +1,43 @@
...
+    @song ||= UserVars.song
+    check_flags
+    main_loop 
+  end
+
+  def clean?
+    return !Flags['clean']
+  end
+
+  def check_flags
```
jandersson, line 13: Dont bother. Let other scripts clean instruments. Dirty instruments are, anecdotally, not impacting performance training.

play.lic
```diff
@@ -0,0 +1,43 @@
...
+    Flags.add('play-clean', 'Your .* dirtiness may affect your performance') # Could check if the flag exists, but playing an instrument tells you if it needs cleaning
+    
+  end
+  
+  def play_song
+    fput("play #{@song} on zills")
+  end
+
+  def can_play?
```
jandersson, line 22: Get from data folder. I think there is a list of scripts to disable to reuse.

play.lic
```diff
@@ -0,0 +1,43 @@
+custom_require.call %w[events]
+class Play
+  def initialize
+    @song ||= UserVars.song
+    check_flags
```
jandersson, line 5: Remove

play.lic
```diff
@@ -0,0 +1,43 @@
...
+    Flags.add('play-clean', 'Your .* dirtiness may affect your performance') # Could check if the flag exists, but playing an instrument tells you if it needs cleaning
+    
+  end
+  
+  def play_song
+    fput("play #{@song} on zills")
+  end
+
+  def can_play?
+    return !['combat-trainer', 'shape', 'forge', 'carve'].any? { |script| Script.running?(script) }
```
jandersson, line 23: Get from data folder. I think there is a list of scripts to disable to reuse.

play.lic
```diff
@@ -0,0 +1,35 @@
...
+
+  def can_play?
+    @no_play_scripts.none? { |script| Script.running?(script) }
+  end
+
+  def should_play?
+    !xp_highenough? && can_play?
+  end
+
+  def xp_highenough?
```
rpherbig, line 23: 💡 I'd invert this to `xp_low_enough?`

play.lic
```diff
@@ -0,0 +1,35 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#play
+=end
+class Play
+  def initialize
+    @no_play_scripts = get_settings.sanowret_no_use_scripts
```
rpherbig, line 6: ❓ Is the list of scripts always going to be the same as for sanowret or is that just a quick hack for now?

play.lic
```diff
@@ -0,0 +1,35 @@
...
+    @song = UserVars.song
+    main_loop
+  end
+
+  def play_song
+    fput("play #{@song}")
```
rpherbig, line 12: 💡 Can we use `bput` here?

jandersson:
Thanks for merging. Id like to address your feedback in another merge request

play.lic
jandersson: Its a quick hack. There should be a play_no_use_scripts

play.lic
jandersson: Makes sense.

play.lic
jandersson: I'd prefer not since we don't validate whether the action was a success or not. If we start doing checks like scaling the instrument difficulty then we can switch it there.

PR #4309, opened 2020-05-17
Update smoke.lic

Tyrranny:
Fixed a couple bugs with how this script works. 

1. Inhale MY cigar fixes issue I had.
2. Exhale Line <image> now looks for Roundtime instead of the dozens of possible return strings for different images and skills with those images. 
3. Sheath my Scimitar now looks for you sheath instead of just you put.


rpherbig:
Thank you for your pull request @Tyrranny !

PR #4311, opened 2020-05-18
Refactoring play.lic

jandersson:
* Uses DRC.play_song?

I have one reservation on using this common method: It looks like it can
run safe-room which I don't believe we want an 'always on' script to do.

* Invert xp predicate

* Add no_play scripts to base and base-empty

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -431,7 +431,7 @@ module DRC
     g
   end
 
-  def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
+  def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false, skip_heal = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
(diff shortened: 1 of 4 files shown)


jandersson:
Have marked this as a WIP/Draft while thinking about the safe-room issue with `DRC.play_song?`. May revert this back to fput.

rpherbig:
I'm fine either way. Have you considered adding _yet another_ optional arg to `play_song?`?

jandersson:
> I'm fine either way. Have you considered adding _yet another_ optional arg to `play_song?`?

Yep :) currently considering this now. I think at some point we possibly want `play_song?` broken up so the individual components are a bit more accessible.

jandersson:
Ill cherry-pick the common changes into a separate PR, then remove the draft marker on this one

jandersson:
I'll re-open this when I have an active account.

PR #4312, opened 2020-05-19
Issue #4298 Stow tk_ammo on manual stop (kill script)

### LINKED ISSUE #4298: Request stow tk_ammo to be added to combat-trainer.lic in the before dying section
It should be fput("stow #{@tk_ammo}")  but if falls outside the main script so it is not getting the settings. I tried a few things to make it work but it is still not getting the settings.

cfrisby77:
If you are a Moon Mage and tk_ammo is on the ground it will attempt to stow tk_ammo if script is manually stopped.
Fixes #4298

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3837,9 +3837,11 @@ class GameState
 end
 
 before_dying do
+  @tk_ammo = get_settings.tk_ammo
   DRCA.release_cyclics
   DRCA.shatter_regalia?
   fput('close fissure') if DRStats.warrior_mage? && DRRoom.room_objs.any? { |x| /fissure/ =~x }
+  fput("stow #{@tk_ammo}") if DRStats.moon_mage? && DRRoom.room_objs.any? { |x| /#{@tk_ammo}/ =~x }
   Flags.delete('using-corpse')
   Flags.delete('pouch-full')
   Flags.delete('container-full')


combat-trainer.lic
```diff
@@ -3837,9 +3837,11 @@ class GameState
 end
 
 before_dying do
+  @tk_ammo = get_settings.tk_ammo
```
rpherbig, line 3840: ❌ You can't use `get_settings` like this because it won't load additional YAML files, only the base.

See https://github.com/rpherbig/dr-scripts/blob/master/combat-trainer.lic#L1040 for an example of using the passed in settings.

I'm pretty sure you can't access `@tk_ammo` here since it's not in the `SpellProcess` class. I'm also pretty sure we can't pass anything into `before_dying` since, if I remember right, that's a Lich thing.

My suggestion is to make `tk_ammo` a global instead of instance var in `SpellProcess`. If you have another idea, I'm all ears. Slack may also have some suggestions.

cfrisby77: I am almost certain i tested the @tk_ammo = settings.tk_ammo and it didn't work. I will look at making it global. and test that.

rcuhljr:
I like to avoid polluting the global namespace when possible. Have we considered having the combat-trainer create a temp hash in UserVars to store information that would be needed out of scope. It's an option for cross script communication but it also works for dying procs that need info that isn't available anymore.

rpherbig:
I'm not generally a fan of using globals either. I hadn't considered UserVars, but those are even more 'permanent' than globals...

Do we need out of scope info more than just this once? I'm worried about investing time building a feature we'll only need once.

jayRyan24:
I don't really think this needs all that type of work. If you are manually stopping combat trainer, just manually type one more line and stow. I do this across all my chars, all the time inside DR and GS. It's just not a big deal, and opening up to globals for something that shouldn't even be an issue sounds pretty kludge.

jayRyan24:
The idea of UserVars as a temporary placeholder for "current" variables would work for what you're asking, but my biggest concern is if it's worth it.

jayRyan24:
Personally, if this was an issue I wanted to resolve for myself, I would just hardcode my items and leave it at that. I've done this in the past, because it's so marginally useful for others. I also tend to use cheap restock tk_ammo, so when I have forgotten to pick mine up, it's just a few copper and seconds to replace.

cfrisby77:
ok...

jayRyan24:
> ok...

We're just conversing here @cfrisby77 . I have been against things in the past and convinced otherwise. Just give your POV so it convinces me. You don't even need to convince me to get this through, I am just a person on the outside giving feedback, I don't hold a merge button's power.

combat-trainer.lic
jayRyan24: **settings.tk_ammo** will not work because settings is a class variable and **before_dying** is out of scope.

rcuhljr:
Just so I'm in scope on things here, TK damage is completely unaffected by what you're throwing, so using the cheapest disposable thing from a common store is generally the right move? Would managing your tk ammo through restock be fine since even if you left a club or whatever behind it could grab one before the hunt anyways.

cfrisby77:
I will just just the restock function and it should be sufficient. I have closed the issue and pull request.

Hiinky:
Really the only limitation is not using a common ammo noun like arrows or bolts so its not normally looted.  I'm fond of general store razors because they're extremely cheap, can be bought in pink with restock, and can't be destroyed by being tossed around.  I also make sure to restock 2 just in case I do leave one.  Not a perfect solution as I have definitely lost my share but its only a couple silver to replace my stash.

jayRyan24:
> Just so I'm in scope on things here, TK damage is completely unaffected by what you're throwing, so using the cheapest disposable thing from a common store is generally the right move? Would managing your tk ammo through restock be fine since even if you left a club or whatever behind it could grab one before the hunt anyways.

Correct yeah, I use just plain sold stuff from the town vendors. Can't be a 0/23 anymore, so stuff like flint isn't as good. Just use the razors or daggers, etc.. I leave my restock on those items as 3, personally.

jayRyan24:
https://github.com/rpherbig/dr-scripts/blob/master/profiles/Jardani-setup.yaml#L377-L388

PR #4360, opened 2020-07-04
Add fell_hogs to base-hunting.yaml

robbintt:
what's on the tin


data/base-hunting.yaml
```diff
@@ -399,6 +399,12 @@ hunting_zones:
   - 1446
   - 1447
   - 1448
+  fell_hogs:
```
rpherbig, line 402: Please also include a link to elanthipedia and approximate hunting range (see line 408 for an example)

PR #4361, opened 2020-07-05
Add optional tries parameter to forage?

robbintt:
Add tries and return false if tries are exhausted.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -160,7 +160,7 @@ module DRC
     script_handle
   end
 
-  def forage?(item, tries=nil)
+  def forage?(item, tries=5)
     snapshot = "#{right_hand}#{left_hand}"
     while snapshot == "#{right_hand}#{left_hand}"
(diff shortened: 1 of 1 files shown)


Tarhish:
Looking back at this method, I think it's actually better if the default would be failing false after a set number of tries.  It does have escapes set for if the area isn't forageable, but if you try to forage for, say, grass in a room with no grass then it'll never escape. I can't see anything in the code base that uses forage? that wouldn't be improved by the ability to abort after, say, twenty attempts instead of going on infinitely looking for an item in a place that it isn't seeded.

My personal vote would be that this method just defaults to something like ten or twenty attempts, then returns false if it runs out the clock.  If that's the case then both of these changes would become simpler too.

rpherbig:
I like that idea @Tarhish . What do you think, @robbintt ?

robbintt:
Nice data.  I've updated it, but I haven't used a ternary in ruby before. I assume it works this way? I can test it out.

rpherbig:
Looks good to me. Thanks again @robbintt and @Tarhish ! 🎉

PR #4365, opened 2020-07-05
Add focus concept to appraise focus.

robbintt:
Not sure if this works, will wait for others to report back.


Angel-DR:
It worked!  Thank you SO much for fixing it and adding appraisal focus defense.  You were so quick and efficient!

rpherbig:
Thank you for fixing this @robbintt !

PR #4373, opened 2020-07-08
Write startergear gatherer for zoluren.

robbintt:
Script for novices to gather gear. Costs about 10 plat, so they will need to beg.

Not yet implemented:
* Instrument
* first aid charts / compendium
* othermagictype runestone for sorcery
* pike/HT that is cheaper than 10 plat (weapon strap + spear, and maybe lob a different weapon?)
* Default yaml for all this?
* ???

### DIFF
--- a/startergear.lic
+++ b/startergear.lic
@@ -221,6 +221,12 @@ class BuyStarterGear
     fput("stow my yew shortbow", "You put")
     waitrt?
     pause
+    buy_fast("steel scimitar")
+    waitrt?
+    pause
+    fput("stow my steel scimitar", "You put")
+    waitrt?
+    pause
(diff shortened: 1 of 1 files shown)


rpherbig:
This looks really similar to https://github.com/rpherbig/dr-scripts/blob/master/new-character.lic - have you looked at that script?

robbintt:
No, I'll take a look, thanks.

robbintt:
This script for novices to kit themselves out in storebought gear without leaving zoluren.

It looks like the other one is more geared towards experienced players spinning up a templated new character.

I'm actually going to look into the other one more and see if it would be better to write an equipment manager profile for novice players.

PR #4391, opened 2020-07-21
Support divine charms

vtcifer:
Add support to safe-room.lic to use the divine charm that starter characters get, instead of going to an auto-healer.

Safely falls back to using regular auto-healers in case of loss of divine charm, and trying too soon again before cool down is up.


safe-room.lic
```diff
@@ -130,6 +134,23 @@ class SafeRoom
     bput("put my material in my #{settings.thanatology['harvest_container']}", 'You put', 'What were') if right_hand
   end
 
+  def use_divine_charm
```
rpherbig, line 137: Since this returns a bool, the convention is to name it with `?` at the end, i.e. `use_divine_charm?`

safe-room.lic
```diff
@@ -130,6 +134,23 @@ class SafeRoom
...
+    return false unless exists?("divine charm")   # charm gone (likely too high cirlce/skill), couldn't heal, return false, heal normal way
```
rpherbig, line 138: 'circle'

PR #4401, opened 2020-07-26
Adds currach support for Aesry to Halasa Temple

Nituki:
Boards currach if on aesry or halasa island and rows to target. Also added support to row to either target if already on the currach.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -219,7 +219,11 @@ class Bescort
       [
         { name: 'jolas', regex: /jolas/i, description: "Ride The Jolas between Hara'jaal and Mer'Kresh." },
         { name: 'mode', options: %w[harajaal merkresh], description: "harajaal or merkresh?  Can be started while aboard the Jolas." }
-      ]   
+      ],
+      [
+        { name: 'currach', regex: /currach/i, description: "Row a currach between Halasa Temple and Aesry Surlaenis'a." },
+        { name: 'mode', options: %w[halasa aesry], description: "halasa or aesry?" }
+      ]  
     ]
 
     args = parse_args(arg_definitions)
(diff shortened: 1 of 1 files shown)


bescort.lic
```diff
@@ -289,12 +293,44 @@ class Bescort
...
+      row_direction = 'pier'
+    elsif mode == 'halasa'
+      row_direction = 'rock'
+    else
+      echo("Unrecognized argument: #{mode}")
+      exit
+    end  
+
+    if (Room.current.id == 5555 || Room.current.id == 5557 || Room.current.id == 5558 || Room.current.id == 15863)
+      fput "go currach", "You climb aboard the", "You are aboard", "What are you referring"
```
rpherbig, line 318: How do you feel about using `bput` and `move` instead of `fput` for all of this new code?

bescort.lic
```diff
@@ -289,12 +293,44 @@ class Bescort
...
+      jolas(args.mode)
+    elsif args.currach
+      currach(args.mode)
     end
   end
 
   private
 
+  def currach(mode)
+    if (mode.include?('aesry') && (Room.current.id == 5555 || Room.current.id == 5557 || Room.current.id == 5558) || (mode.include?('halasa') && Room.current.id == 15863))
```
rpherbig, line 305: This conditional is a bit hard to read, in my opinion. What would you think about refactoring it?

A few ideas:

* two `if` blocks, one for aesry and one for halasa
* Using something like `[5555, 5557, 5558].include?(Room.current.id)`

bescort.lic
```diff
@@ -289,12 +293,44 @@ class Bescort
...
+
+    if (Room.current.id == 5555 || Room.current.id == 5557 || Room.current.id == 5558 || Room.current.id == 15863)
+      fput "go currach", "You climb aboard the", "You are aboard", "What are you referring"
+      fput "get oars"
+      fput "untie currach"
+    end
+
+    loop do
+      if bput("tie currach to #{row_direction}", "isn't close enough!", "You pull at the oars") == "You pull at the oars"
+        fput "stand"
```
rpherbig, line 325: What do you think about using `fix_standing` here?

bescort.lic
Nituki: Thanks submitted a cleaner version based off your comments.

bescort.lic
Nituki: Great idea!

bescort.lic
Nituki: Didn't even know about fix_standing. Updated on v2

PR #4404, opened 2020-07-27
Added oasis support for Hidisharon Desert

Nituki:
Added support for finding the oasis in the Hidisharon Desert (Armadillos) including an update to exit mode to go from oasis out of desert.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -610,8 +610,58 @@ class Bescort
     end
     case mode
     when /oasis/
-      echo 'Not implemented'
-      exit
+    loop do
+      desert_enter
+      path = %w[s ne ne s s ne e]
+      path.each do |movement|
+        if DRRoom.room_objs.include?('shimmering oasis')
+          move 'go oasis'
+          exit
+        else
+          move(movement)
+        end
+      end
+      move 'go trail'
+      path = %w[ne n ne n ne n n ne n ne n]
+      path.each do |movement|
+        if DRRoom.room_objs.include?('shimmering oasis')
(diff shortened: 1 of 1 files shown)


bescort.lic
```diff
@@ -610,8 +610,58 @@ class Bescort
     end
     case mode
     when /oasis/
-      echo 'Not implemented'
-      exit
+    loop do
+      desert_enter
+      path = %w[s ne ne s s ne e]
+      path.each do |movement|
```
rpherbig, line 616: What do you think about extracting this loop into a function to reduce duplicated code?

If you don't have time, I'm fine with merging this as-is, too.

bescort.lic
```diff
@@ -610,8 +610,58 @@ class Bescort
     end
     case mode
     when /oasis/
-      echo 'Not implemented'
-      exit
+    loop do
```
rpherbig, line 613: How confident are you that this will eventually terminate? I don't generally like unbounded `loop`s...

Nituki: I can't imagine it not terminating since it goes through every room until it finds the oasis and the oasis doesn't move too often, but I had it as a loop in the rare chance it moves while you are looking (happened during one of my test runs) but I ran it over 15 times and it found it and exited with the oasis in different locations. I could try and set it to only loop a few times. I just need to think of how to do that elegantly (still a new to ruby scripting).

rpherbig: Nah, this seems fine with your explanation. Thank you!

PR #4406, opened 2020-07-28
Add generic repeater loop.

robbintt:
Great inline in-game replacement for one offs.

### DIFF
--- a/repeat.lic
+++ b/repeat.lic
@@ -1,0 +1,14 @@
+=begin
+  Documentation:
+
+=end
+
+loop {
+  for n in 1..9
+    if variable[n]
+      pause
+      waitrt?
+      put "#{variable[n]}"
+    end
+  end
+}


repeat.lic
```diff
@@ -0,0 +1,14 @@
+=begin
+  Documentation:
```
robbintt, line 2: the repeated items must be double quoted e.g. `;repeat "get potato from my backpack" "put potato in urn"`

Hiinky:
This already exits.  ;repos download multi.  Accepts scripts also
syntax:  ;multi 3,do thing one, do thing two,;script 2

robbintt:
> This already exists. ;repos download multi. Accepts scripts also
> syntax: ;multi 3,do thing one, do thing two,;script 2

Thanks, where does that exist under source control?

jandersson:
Hello! Looks good to me. With respect to the duplicate concept, I don't see any issue with this repo maintaining its own version of it. If the original author wants to submit `multi` then it can be taken from there.

repeat.lic
robbintt: will add docs in separate PR

PR #4408, opened 2020-07-29
[script] [dusk-labyrinth] efficiently walk rooms

KatoakDR:
Builds an internal map of the maze to efficiently skip rooms that the player has already been to (e.g. back tracking to pick a new direction). This saves several minutes by not spending time waiting for a pet scurrying message when we know the pet won't be found in that room.

When maze reshuffles then script clears its knowledge of the maze and rebuilds it anew.

### DIFF
--- a/dusk-labyrinth.lic
+++ b/dusk-labyrinth.lic
@@ -150,12 +150,20 @@ class DuskLab
 
     move(@forward_move)
 
+    # Since rooms in the maze don't have ids, we build
+    # a bi-directional graph where each vertex (room)
+    # points to another vertex (room) and vice versa.
     @previous_room = @current_room
     @current_room = @current_room['exits'][@forward_move]['room'] || create_room
 
     @previous_room['exits'][@forward_move]['room'] = @current_room
     @current_room['exits'][@reverse_move]['room'] = @previous_room
 
+    # Increment the count of us having moved between these two rooms.
+    # In graph algorithms, this is known as a weighted graph so that
+    # when we pick the "next best move" to go down we choose paths
+    # with the least count (weight) thereby prioritizing paths we
+    # have not visited before spending time visiting rooms we've been to.
     @previous_room['exits'][@forward_move]['count'] = @previous_room['exits'][@forward_move]['count'] + 1
     @current_room['exits'][@reverse_move]['count'] = @current_room['exits'][@reverse_move]['count'] + 1
 
--- a/dusk-labyrinth.lic
+++ b/dusk-labyrinth.lic
@@ -7,74 +7,200 @@ custom_require.call %w[common common-items events]
 class DuskLab
   include DRC
   include DRCI
+
   def initialize
+
+    arg_definitions = [
+      [
+        { name: 'debug', regex: /debug/i, optional: true, description: 'Enable debug output' },
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+    $debug_mode_dl = UserVars.dusk_labyrinth_debug || args.debug || false
+
     @settings = get_settings
     @labyrinth_data = @settings.duskruin['labyrinth']
+
     loot_container = @labyrinth_data['loot_container']
     pet_noun = @labyrinth_data['pet_noun']
     redeem_scrip = @labyrinth_data['redeem_scrip']
+
+    # Map of directions to help know where we've come from and how to go back.
+    @reverse_direction_map = {
+      'e' => 'w',
+      'w' => 'e',
+      's' => 'n',
+      'n' => 's',
+      'ne' => 'sw',
+      'sw' => 'ne',
+      'nw' => 'se',
+      'se' => 'nw',
+      'east' => 'west',
+      'west' => 'east',
+      'south' => 'north',
+      'north' => 'south',
+      'northeast' => 'southwest',
+      'southwest' => 'northeast',
+      'northwest' => 'southeast',
+      'southeast' => 'northwest'
+    }
+
+    # Be notified when certain events happen, such as seeing a pet move or the maze reshuffle.
     Flags.add('five-minute-warning', 'You only have about 5 minutes left in the labyrinth!')
     Flags.add('done', 'A surly Dwarf stomps in and glowers at you in the dim gloom')
     Flags.add('no-pet', "A small #{pet_noun} scurries #{Regexp.new('(?<pet_dir>\w+)')}!")
     Flags.add('pet', "You notice a small #{pet_noun} scurrying around the area")
     Flags.add('caught', 'You pick them both up, claiming your new pet')
-    @prev_dir = nil
-    @dir_to_prev_dir = {
-      'northeast' => 'southwest',
-      'southwest' => 'northeast',
-      'northwest' => 'southeast',
-      'southeast' => 'northwest',
-      'north' => 'south',
-      'south' => 'north',
-      'east' => 'west',
-      'west' => 'east',
-      'up' => 'down', # up and down dont appear in the maze, but are used for testing
-      'down' => 'up'
-    }
+    Flags.add('maze-shuffled', 'no longer certain of your directions')
+
+    # Let's do this!
+    init_maze
     stow_hands
     main(loot_container) until done?
     redeem if redeem_scrip
   end
 
-  def redeem
-    while bput('get my bloodscrip', 'You get', 'What were') == 'You get'
-      bput('redeem my bloodscrip', 'You quickly pocket')
-    end
+  # Used at beginning of script and any time the maze shuffles.
+  def init_maze
+    echo("*** initializing map data") if $debug_mode_dl
+    # Maze data for the current room.
+    @current_room = nil
+    # Maze data for the previous room.
+    @previous_room = nil
+    # Direction we last moved to go from "previous_room" to "current_room".
+    @forward_move = nil
+    @reverse_move = nil
   end
 
+  # Main loop to check our room, search, or wander the maze.
+  # Repeatedly invoked until the maze event ends.
   def main(loot_container)
-    # search
-    @pause_timer = Time.now
-    pause 1 until Flags['no-pet'] || Flags['pet'] || Flags['caught']
+    if @previous_room && @previous_room['exits'][@forward_move]['count'] > 1 && !Flags['caught'] && !Flags['maze-shuffled'] && !Flags['five-minute-warning']
+      echo("*** Been here before, moving to another room") if $debug_mode_dl
+      wander unless done?
+      return
+    end
+    echo("Checking room...") if $debug_mode_dl
+    # Wait until we determine that this room either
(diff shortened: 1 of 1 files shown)


dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
 class DuskLab
   include DRC
   include DRCI
+
   def initialize
+
+    arg_definitions = [
+      [
+        { name: 'debug', regex: /debug/i, optional: true, description: 'Enable debug output' },
```
KatoakDR, line 15: Modeled after `combat-trainer`

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+    # Map of directions to help know where we've come from and how to go back.
+    @reverse_direction_map = {
+      'e' => 'w',
+      'w' => 'e',
+      's' => 'n',
+      'n' => 's',
+      'ne' => 'sw',
+      'sw' => 'ne',
+      'nw' => 'se',
+      'se' => 'nw',
```
KatoakDR, line 38: The `checkpaths` command returns the direction abbreviations. Original script used `XMLData.room_exits` which returned the full names. Keeping both to cover our bases.

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+    }
+
+    # Be notified when certain events happen, such as seeing a pet move or the maze reshuffle.
     Flags.add('five-minute-warning', 'You only have about 5 minutes left in the labyrinth!')
     Flags.add('done', 'A surly Dwarf stomps in and glowers at you in the dim gloom')
     Flags.add('no-pet', "A small #{pet_noun} scurries #{Regexp.new('(?<pet_dir>\w+)')}!")
     Flags.add('pet', "You notice a small #{pet_noun} scurrying around the area")
     Flags.add('caught', 'You pick them both up, claiming your new pet')
-    @prev_dir = nil
-    @dir_to_prev_dir = {
```
KatoakDR, line 22: Moved to line 30

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+    Flags.add('maze-shuffled', 'no longer certain of your directions')
+
+    # Let's do this!
+    init_maze
     stow_hands
     main(loot_container) until done?
     redeem if redeem_scrip
   end
 
-  def redeem
```
KatoakDR, line 39: Moved to line 235

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+    # Direction we last moved to go from "previous_room" to "current_room".
+    @forward_move = nil
+    @reverse_move = nil
   end
 
+  # Main loop to check our room, search, or wander the maze.
+  # Repeatedly invoked until the maze event ends.
   def main(loot_container)
-    # search
-    @pause_timer = Time.now
```
KatoakDR, line 47: Didn't see any other places where `@pause_timer` was used. I think it was left over from previous authors. I haven't noticed any issues without it.

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
-    pause 1 until Flags['no-pet'] || Flags['pet'] || Flags['caught']
+    if @previous_room && @previous_room['exits'][@forward_move]['count'] > 1 && !Flags['caught'] && !Flags['maze-shuffled'] && !Flags['five-minute-warning']
+      echo("*** Been here before, moving to another room") if $debug_mode_dl
+      wander unless done?
+      return
+    end
```
KatoakDR, line 83: What this is saying is if we came from a previous room, and the number of times we've traversed from "there to here" or "here to there" is more than once (meaning we are visiting this room for the second time), then we'll immediately pick another room to move to. There's no value in waiting in this room because we already know we won't find a pet here -- we've already been here and waited once before.

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+    echo("Checking room...") if $debug_mode_dl
+    # Wait until we determine that this room either
+    # doesn't have a pet, has a pet, we caught the pet, or the maze reshuffeld.
+    pause 1 until Flags['no-pet'] || Flags['pet'] || Flags['caught'] || Flags['maze-shuffled']
```
KatoakDR, line 87: I added the `|| Flags['maze-shuffled']` to this existing condition

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+    # In graph algorithms, this is known as a weighted graph so that
+    # when we pick the "next best move" to go down we choose paths
+    # with the least count (weight) thereby prioritizing paths we
+    # have not visited before spending time visiting rooms we've been to.
+    @previous_room['exits'][@forward_move]['count'] = @previous_room['exits'][@forward_move]['count'] + 1
+    @current_room['exits'][@reverse_move]['count'] = @current_room['exits'][@reverse_move]['count'] + 1
+
   end
 
+  def create_room
```
KatoakDR, line 172: In our in-memory graph of the maze, this function creates a new vertex (room) and initializes its internal state

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+      }
+    end
+    return new_room
+  end
+
+  # Returns a direction to move in.
+  # Prioritizes rooms we've not visited before.
+  def choose_next_move(exits)
+    # https://medium.com/@florenceliang/some-notes-about-using-hash-sort-by-in-ruby-f4b3a700fc33
+    sorted_exits = exits.sort_by { |k, v| v['count'] }
```
KatoakDR, line 201: Each "exit" in the array given is a "room" (vertex) in our graph. And the rooms (vertices) keep track of how many times we've moved from the the previous room to this room in the `count` property. So of the exits from this room, we want to pick a direction that we've not been down before, or if we've visited them all, begin backtracking rooms that have been visited the least.

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
+  # Used at beginning of script and any time the maze shuffles.
+  def init_maze
+    echo("*** initializing map data") if $debug_mode_dl
+    # Maze data for the current room.
+    @current_room = nil
+    # Maze data for the previous room.
+    @previous_room = nil
+    # Direction we last moved to go from "previous_room" to "current_room".
+    @forward_move = nil
+    @reverse_move = nil
```
rcuhljr, line 73: I realize this is a refactor and may not be your original script, so feel free to ignore comments about things that you just moved, I'm just giving some off the cuff thoughts. 

There's no need to initialize variables like this, they effectively exist as soon as the code references them anywhere, and are already set to nil. 

```
irb(main):001:0> class Test
irb(main):002:1>   def initialize
irb(main):003:2>     puts @not_mentioned.nil?
irb(main):004:2>   end
irb(main):005:1> end
=> :initialize
irb(main):006:0> Test.new
true
```

If you want a place to call out the variables you'll be using you could use [attribute accessors](https://www.rubyguides.com/2018/11/attr_accessor/) which also gives you some debug access to state if you hang onto the created object from this class.

dusk-labyrinth.lic
```diff
@@ -7,74 +7,208 @@ custom_require.call %w[common common-items events]
...
   end
 
+  # Main loop to check our room, search, or wander the maze.
+  # Repeatedly invoked until the maze event ends.
   def main(loot_container)
-    # search
-    @pause_timer = Time.now
-    pause 1 until Flags['no-pet'] || Flags['pet'] || Flags['caught']
+    if @previous_room && @previous_room['exits'][@forward_move]['count'] > 1 && !Flags['caught'] && !Flags['maze-shuffled'] && !Flags['five-minute-warning']
```
rcuhljr, line 79: Uh, not linked to this line, but this is what started my thinking. Would you be better off just having a boolean in your room nodes that tracks if you've visited the room before? It would be a lot more conceptually clear than asking 'have I traveled down the edge to this node from the previous node before. Also unless you're incrementing all neighbors from this room that travel to this room (or all their edges) it seems like this would consider a room untraveled if you arrived at it from a different neighbor than you had before? 

Honestly you could make new 'rooms' for every direction out of your current room that need them, then you don't need edge weights for graph transversal (all costs are identical anyways) and are just looking for the closest node that hasn't been visited yet.

dusk-labyrinth.lic
KatoakDR: Thanks for the insightful and helpful coaching :) I am learning Ruby _as I refactor_ Lich scripts, hehe

Good to know that variables don't have to be pre-defined, and thanks for the tip about attribute accessors.

In this very specific case, those variable assignments are part of the `init_maze` function defined at line 65 which is called at script start but also anytime the maze reshuffles. It's when the maze reshuffles that they needed to be set back to nil.

dusk-labyrinth.lic
KatoakDR: Very astute observations, and I agree with you. I got hung up on the idea of a "weighted" graph to influence which room to move to next when the core goal is just to track "have I been here before?" and if yes, immediately move to a not-yet-visited room rather than sit and redundantly wait for messaging that you won't find a pet here.

We're at the tail end of Duskruin, I won't have time to refactor and test before this year's event closes. How would you like to proceed -- leave this open and revisit next year or merge as-is or secret option 3?

Thanks again for taking time to provide great feedback; it's helping me a lot with learning Ruby and Lich scripting =)

dusk-labyrinth.lic
rcuhljr: Ah yeah, my brain saw init_maze and turned it into initialize, I thought this was inside the constructor, ignore the comment in general, but yes the call on 58 doesn't do anything and a more accurate name might be `reset_maze_data` or something similar.

dusk-labyrinth.lic
rcuhljr: I don't play anymore so I'm not up on the scheduling, if dusk ruin is basically over and you've had a chance to test this so it's at least not worse then the old logic as far as repeating rooms I wouldn't be opposed to @rpherbig merging this in with the hope that next year you're still contributing and can take a fresh look at it. Alternatively it also doesn't hurt to just leave this PR open until it can be finished next year. I don't feel strongly so I'll leave it to you two to sort out.

Welcome to Lich and glad to have a new set of interested hands helping out.

rpherbig: I'm inclined to merge this, then. I like better-but-not-perfect over waiting until next year.

douglascayers: Thanks!

KatoakDR: Thanks!

PR #4413, opened 2020-08-04
Add Support for FA/Skinning MT trainers

Hiinky:
supports noun variable
allows skill selection
can be used independently, with t2 or crossing-training/training-manager.

### DIFF
--- a/faskinner.lic
+++ b/faskinner.lic
@@ -32,7 +32,8 @@ class FaSkinner
     elsif @priority == 'Both' || @priority == 'both'
       until DRSkill.getxp('First Aid') > 32  && DRSkill.getxp('Skinning') > 32
         train_skills
-      end        
+      end 
+      echo ("Exiting because First Aid AND Skinning locked!")
     else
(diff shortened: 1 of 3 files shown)


faskinner.lic
```diff
@@ -0,0 +1,77 @@
...
+      until DRSkill.getxp('First Aid') > 32
+        train_skills
+      end
+      echo ("Exiting because First Aid locked!")
+    elsif @priority == 'Skinning' && DRSkill.getxp('Skinning') < 34
+      until DRSkill.getxp('Skinning') > 32
+        train_skills
+      end
+      echo ("Exiting because Skinning locked!")
+    elsif @priority == 'Both' || @priority == 'both'
```
rpherbig, line 34: I suggest something like `.downcase == 'both'` for all of these branches so you don't have to worry about casing.

rcuhljr: or `@priority =~ /both/i`. Although this is the only one that's case insensitive which fields weird (versus the other options, which I guess are proper skill names).

PR #4418, opened 2020-08-09
New Script - wand-watcher.lic

vtcifer:
Generalized version of mefwatch from repository. Should work with all kinds of wands, and bases its activation on cooldown timers, instead of needing to check if the spell/effect is still active.

Script runs in background and waits for the next time to use a wand.  It stores these times as UserVars.

Example settings:

```yaml
wands:
  # magic wand:                     #Adj + Noun of the wand.  Must be unique  [Mandatory]
  #   activation message: Foo Bar   #Spell activation message                 [Mandatory]
  #   activation verb: tap          #Verb used to activate the wand           [Optional - defaults to 'tap']
  #   container: backpack           #Container where the wand(s) are stored   [Mandatory]
  #   cooldown: 32                  #minutes between activation attempts      [Optional - defaults to 30]
  #   count: 2                      #number of wands you have                 [Optional - defaults to 2]
  bloodwood branch:
    activation message: The world around you seems to slow as the spell grips your mind.
    activation verb: tap
    container: rucksack
    cooldown: 32
    count: 2

wand_watcher_no_use_scripts:  # Don't try to grab any wands while these scripts are running.
  - burgle
  - go2

wand_watcher_startup_delay: 10  # Wait n seconds on startup before doing anything   [Optional - defaults to 10 s]
wand_watcher_passive_delay: 60  # How long to wait between checks for reusing wands [Optional - defaults to 60 s]
```


vtcifer:
Forgot to note, that 4417 needs to be accepted before this can, as this scripts makes use of new methods included in that PR.

rpherbig:
#4417 was merged. Remind me in a few days to check on this one

KatoakDR:
Thanks for this script, @vtcifer! If you don't mind, I'm going to add some comments for enhancements that I recently made to my local copy of `mefwatch` to support rooms where not to try and invoke a wand, like at a bank window or a vault. The concept is just like no_use_scripts, but for rooms :)

wand-watcher.lic
```diff
@@ -0,0 +1,153 @@
...
+
+class WandWatcher
+  include DRC
+  include DRCI
+
+  def initialize
+    settings = get_settings();
+    @wand_list = settings.wands
+    @queued_wands = []
+    @no_use_scripts = settings.wand_watcher_no_use_scripts || []
```
KatoakDR, line 18: Proposing list of rooms where not to try and use the wands, such as banks and vaults or other areas where magical devices are restricted.

```suggestion
    @no_use_scripts = settings.wand_watcher_no_use_scripts || []
    @no_use_rooms = settings.wand_watcher_no_use_rooms || []
```

wand-watcher.lic
```diff
@@ -0,0 +1,153 @@
...
+    @startup_delay = settings.wand_watcher_startup_delay || 10
+    @passive_delay = settings.wand_watcher_passive_delay || 60
+    @activation_failure_messages = [/^The .* remains inert/]
+    UserVars.wand_watcher_timers = {} unless UserVars.wand_watcher_timers
+
+    #don't allow users to make foolish mistakes, always prevent use during go2/burgle
+    @no_use_scripts << 'go2'
+    @no_use_scripts << 'burgle'
```
KatoakDR, line 26: If `go2` or `burgle` are already in the user's `wand_watcher_no_use_scripts ` yaml config, then the original lines 25-26 -- although very succinct -- would duplicate those entries. This tweak is my novice ruby mind's way of only adding the entries if they don't already exist.
```suggestion
    ['go2', 'burgle'].each { |script| @no_use_scripts.push(script) unless @no_use_scripts.include?(script) }
```

wand-watcher.lic
```diff
@@ -0,0 +1,153 @@
...
+    end
+  end
+
+  def check_timers  
+    #if any no_use_scripts are running, loop every 10 seconds checking until not
+    loop do
+      break if !(@no_use_scripts.any? { |name| Script.running?(name) })
+      pause 10
+    end
+
```
KatoakDR, line 68: Proposing list of rooms where not to try and use the wands, such as banks and vaults or other areas where magical devices are restricted. This loops and checks every 10 seconds if the user has moved to a different room before resuming the script just like waiting for no_use_scripts to finish.
```suggestion

    #if player is in any no_use_rooms, loop every 10 seconds checking until not
    loop do
      # if in a room without an id then Room.current is nil but DRRoom.title is reliable
      break if !(@no_use_rooms.any? { |name| /#{name}/ =~ DRRoom.title || name.to_s == Room.current.id.to_s })
      pause 10
    end

```

wand-watcher.lic
```diff
@@ -0,0 +1,153 @@
...
+    
+    #if both hands are full, store the left hand item in a variable for future use and lower the 
+    #left hand to the feet slot.  
+    #left causes less problems with combat (less likely to be a loaded weapon)
+    #feet slot causes less space/correct container issues
+    temp_left_item = nil
+    if DRC.right_hand && DRC.left_hand
+      temp_left_item = DRC.left_hand
+      bput("lower ground left",/^You lower/)
+    end
```
KatoakDR, line 96: Ingenius! I didn't think of this as a viable option for emptying hands without knowing _where_ they'd end up stored until I saw this. Went and tested and sure enough, more than one item can be in your "at feet" slot, pretty cool!

wand-watcher.lic
```diff
@@ -0,0 +1,153 @@
...
+      if !get_item("#{$ORDINALS[count]} #{wand}",container)
+        message "Could not find correct number of wands for #{wand} in #{container}."
+        message "Double check your wand name, count, an container settings." 
+        message "Removing this wand from the list."
+        @wand_list.delete(wand)
+        UserVars.wand_watcher_timers.delete(wand)
+        next
+      end
+
+      case bput("#{activation_verb} my #{wand}",activation_successful_message,@activation_failure_messages)
```
KatoakDR, line 126: Proposing list of rooms where not to try and use the wands, such as banks and vaults or other areas where magical devices are restricted. This detects if the current room prevents use of the wand.
```suggestion
      case bput("#{activation_verb} my #{wand}",activation_successful_message,@activation_failure_messages,@activation_blocked_messages)
```

wand-watcher.lic
```diff
@@ -0,0 +1,153 @@
...
+        UserVars.wand_watcher_timers[wand] = Time.now + cooldown
+      when *@activation_failure_messages
+        #if activation fails, could just be a sort issue due login.
+        #retry, until you succeed, or run out of wands
+        put_away_item?(wand,container)
+        redo_count += 1
+        redo unless redo_count > count
+        #try again in 1/2 the cooldown time
+        UserVars.wand_watcher_timers[wand] = Time.now + (cooldown / 2)
+      end
```
KatoakDR, line 138: Proposing list of rooms where not to try and use the wands, such as banks and vaults or other areas where magical devices are restricted. This notifies the user that they are in a room that prevents use of their wand. As a convenience, it adds the room to the no_use_rooms list for this script run. To permanently be remembered then the user needs to add this room to their yaml config. Note that we use `DRRoom.title` and not `Room.current.id` because some restricted rooms, like vaults, have no id in the map database so the room title is the most reliable data that I'm aware of.
```suggestion
      when *@activation_blocked_messages
        message "Could not use #{wand} in room #{DRRoom.title}."
        message "Consider adding this room to your wand_watcher_no_use_rooms settings."
        # Go ahead and remember to ignore this room while the script is running
        # This convenience won't persist between script runs though, go update your yaml
        @no_use_rooms.push(DRRoom.title) unless @no_use_rooms.include?(DRRoom.title)
      end
```

profiles/base.yaml
```diff
@@ -1432,3 +1432,20 @@ duskruin:
...
+  #   activation message: Foo Bar   # Spell activation message                 [Mandatory]
+  #   activation verb: tap          # Verb used to activate the wand           [Optional - defaults to 'tap']
+  #   container: backpack           # Container where the wand(s) are stored   [Mandatory]
+  #   cooldown: 32                  # minutes between activation attempts      [Optional - defaults to 30]
+  #   count: 2                      # number of wands you have                 [Optional - defaults to 2]
+
+wand_watcher_no_use_scripts:  # Don't try to grab any wands while these scripts are running.
+  - burgle
+  - go2
+
```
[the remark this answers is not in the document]
vtcifer: I don't think partial room name matches would be a good idea here, it should require full room titles to prevent over-matching.  

The number of no use rooms that need to be specified that way (rooms which you can't use it, _but can't be mapped_) should be minimal.

wand-watcher.lic
vtcifer: I think forcing this into a looping construct for such a simple thing is overkill.  
There is minimal impact to having duplicates of the these two scripts in the array, and adding the looping construct just makes the code more complex without extra benefit in my opinion.

profiles/base.yaml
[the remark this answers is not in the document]
KatoakDR: I hear you. Since this is the user's yaml, they will have the flexibility to put as much of the room's title here as they'd like. So to your point, the `base.yml` and examples should specify a full room's title as maybe the "best practice" and I like for the script to still support the flexibility of `=~` for folks (like me) that may want a broader net cast rather than listing each room's title or id individually.

wand-watcher.lic
KatoakDR: Yeah, I wasn't really a fan of how much extra code that turned into, either. If there's little performance impact to checking if a script is running multiple times then the duplicates may be ok then.

wand-watcher.lic
```diff
@@ -0,0 +1,153 @@
...
+  include DRCI
+
+  def initialize
+    settings = get_settings();
+    @wand_list = settings.wands
+    @queued_wands = []
+    @no_use_scripts = settings.wand_watcher_no_use_scripts || []
+    @startup_delay = settings.wand_watcher_startup_delay || 10
+    @passive_delay = settings.wand_watcher_passive_delay || 60
+    @activation_failure_messages = [/^The .* remains inert/]
```
KatoakDR, line 21: Proposing list of rooms where not to try and use the wands, such as banks and vaults or other areas where magical devices are restricted.
```suggestion
    @activation_failure_messages = [/^The .* remains inert/]
    @activation_blocked_messages = ["This is not a good place for that.", "You shouldn't disrupt the area", "Something in the area is interfering"]
```

KatoakDR:
Thanks for the no-use room updates, @vtcifer 👍

wand-watcher.lic
```diff
@@ -0,0 +1,179 @@
...
+
+    # redo_count prevents infinite loops when getting a failure message on activation
+    redo_count = 0
+
+    @queued_wands.each do |wand|
+      # Get wand settings
+      activation_successful_message = @wand_list[wand]['activation message']
+      activation_verb = @wand_list[wand]['activate verb'] || 'tap'
+      container = @wand_list[wand]['container'] 
+      cooldown = @wand_list[wand]['cooldown'] || 30
```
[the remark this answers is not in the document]
vtcifer: I think checking the presence of activation message and container are the only ones that really needs a validator?  The others two are just defaults that match most wands.  
But more validators are always good idea.

wand-watcher.lic
```diff
@@ -0,0 +1,179 @@
...
+class WandWatcher
+  include DRC
+  include DRCI
+
+  def initialize
+    settings = get_settings
+    @wand_list = settings.wands
+    @queued_wands = []
+    @no_use_scripts = settings.wand_watcher_no_use_scripts || []
+    @no_use_rooms = settings.wand_watcher_no_use_rooms || []
```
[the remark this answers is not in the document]
vtcifer: I always forget about base-empty.  Thanks!

vtcifer:
Switched to draft as I have encountered a (very rare) race condition in dealing with left hand items on wake up, that could have the potential to lead to item loss...

wand-watcher.lic
```diff
@@ -0,0 +1,179 @@
...
+      check_timers
+      pause @passive_delay
+    end
+  end
+
+  def check_timers  
+    # If any no_use_scripts are running, or you are in any of the no_use_rooms, 
+    # loop every 10 seconds checking until not
+    loop do
+      break if !(@no_use_scripts.any? { |name| Script.running?(name) }) && !( @no_use_rooms.any? { |room| room === DRRoom.title.to_s()[2..-3] || room == Room.current.id} )
```
KatoakDR, line 72: I may or may not have been skulking in the shadows when I was forced out of my hiding spot to use a wand, subsequently spooking my mark. 😁 

```suggestion
      break if !invisible? && !hidden? && !(@no_use_scripts.any? { |name| Script.running?(name) }) && !( @no_use_rooms.any? { |room| room === DRRoom.title.to_s()[2..-3] || room == Room.current.id} )
```

wand-watcher.lic
```diff
@@ -0,0 +1,179 @@
...
+    @queued_wands = []
+    @no_use_scripts = settings.wand_watcher_no_use_scripts || []
+    @no_use_rooms = settings.wand_watcher_no_use_rooms || []
+    @startup_delay = settings.wand_watcher_startup_delay || 10
+    @passive_delay = settings.wand_watcher_passive_delay || 60
+    @activation_failure_messages = [/^The .* remains inert/]
+    @activation_blocked_messages = [/^Something in the area is interfering/,
+                                    /^This is not a good place for that/,
+                                    /^You shouldn't disrupt the area/,
+                                    /^You really shouldn't be loitering in here/]
```
KatoakDR, line 26: Blocked when at the Item Registration Office
```suggestion
                                    /^You really shouldn't be loitering in here/,
                                    /^As you begin to do that an assistant scowls at you and motions for you to stop/]
```

wand-watcher.lic
vtcifer: I'm not convinced on this one tbh.  This could create a race condition where you never end up using it in combat if you're using stealth and the timers interact badly.  I feel like the better option is remembering to kill the script when you don't want it taking over for you.

douglascayers: Good point. I don't train stealth in combat and hadn't thought of that scenario. I can easily kill this or other scripts I don't want interfering with something else I'm doing. 👍

KatoakDR: (reposting reply as KatoakDR — trying to keep my GitHub profile on theme!)

Good point. I don't train stealth in combat and hadn't thought of that scenario. I can easily kill this or other scripts I don't want interfering with something else I'm doing. 👍

vtcifer:
This should be ready to merge... @rpherbig (Just a friendly ping).

rpherbig:
Thanks for the poke

PR #4424, opened 2020-08-09
Add support for forcing crafting hometown 

vtcifer:
This adds support for several new yaml settings to allow hometown overrides on where to do crafting, similar to how it is handled with force_healer_town and saferoom.

I've tested for shaping and remedies, and works perfect, but could probably stand to be tested with others as well.

### DIFF
--- a/remedy.lic
+++ b/remedy.lic
@@ -17,11 +17,14 @@ class Remedy
 
   def initialize
     @settings = get_settings
+    @hometown = @settings.hometown
+    @hometown = @settings.force_crafting_town if @settings.force_crafting_town
+    @hometown = @settings.force_remedies_crafting_town if @settings.force_remedies_crafting_town
     @bag = @settings.crafting_container
     @bag_items = @settings.crafting_items_in_container
     @belt = @settings.alchemy_belt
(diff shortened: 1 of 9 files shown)


remedy.lic
```diff
@@ -17,11 +17,14 @@ class Remedy
 
   def initialize
     @settings = get_settings
+    @hometown = @settings.hometown
```
rpherbig, line 20: What would you think about something like `@hometown = @settings.force_remedies_crafting_town || @settings.force_crafting_town || @settings.hometown`? Do you think that makes the order of precedence clearer?

vtcifer: I liked the symmetry with how the other scripts work (and in some cases HAS to work due to how smith works) to keep the code logic as close to the same, but am not opposed to switching it up if it's more in line with ruby style standards.

rpherbig: It's not a big deal

vtcifer:
After review, some additional scripts (helper scripts mostly) need to be updated.  Will try to get some work done on this in the next few weeks, but pulling back to draft for now.

vtcifer:
Stale forked repo - will redo at a later time

PR #4432, opened 2020-08-13
[script] [common-money] Ensure convert_to_copper returns numeric value

KatoakDR:
Fixes issue where `sell-loot` script at [line 92](https://github.com/rpherbig/dr-scripts/blob/master/sell-loot.lic#L92) will fail if the denomination of coins to keep on hand is not one of platinum, gold, silver, or bronze, or if the amount to keep on hand is less than 10 copper. For example, if you want to keep `5 copper` on hand.

### The Problem

The `convert_to_copper` function returns a String instead of an Integer in the default case.

```
--- Lich: error: undefined method `/' for "0":String
	common-money:16:in `block in minimize_coins'
	common-money:13:in `each'

--- Lich: sell-loot has exited.
```

### Steps to Reproduce Error

1. Define the following in your yaml:
```yaml
sell_loot_money_on_hand: 0 copper
```

2. Run `sell-loot` script

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -25,7 +25,7 @@ module DRCM
     return (amount.to_i * 1000) if 'gold' =~ /^#{denomination}/
     return (amount.to_i * 100) if 'silver' =~ /^#{denomination}/
     return (amount.to_i * 10) if 'bronze' =~ /^#{denomination}/
-    amount
+    amount.to_i
   end
 
   def convert_currency(amount, from, to, fee)


rpherbig:
My kingdom for a type system

common-money.lic
```diff
@@ -25,7 +25,7 @@ module DRCM
     return (amount.to_i * 1000) if 'gold' =~ /^#{denomination}/
     return (amount.to_i * 100) if 'silver' =~ /^#{denomination}/
     return (amount.to_i * 10) if 'bronze' =~ /^#{denomination}/
-    amount
+    amount.to_i
```
rpherbig, line 28: 💡 In hindsight (i.e. after I already merged this) I wonder if we should instead do `amount = amount.to_i` as the first thing in the function... oh well

KatoakDR: Thanks for the quick review and merge. I also thought about that but not as familiar with the why it was done the way it was so chose path of least resistance 😅

PR #4434, opened 2020-08-13
Update bescort.lic

bastionwt:
For people < 400 athletics, the ferry from Crossing to Leth can be a major time sink where little training is done. This update adds in first-aid to that downtime if a textbook or compendium is present, otherwise it defaults back to performance. No new settings are needed in the YAML, as both compendium_type and textbook_type are already present in base.yaml.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -1464,8 +1464,10 @@ class Bescort
 
     case bput('go ferry', 'The Captain gives you a little nod', 'You hand him', 'The ferry has just pulled away from the dock', 'There is no ferry here to go aboard', 'Come back when you can afford the fare')
     when 'You hand him', 'The Captain gives you a little nod'
-      # hide?
-      if get_settings.textbook
+      if hide?
+        waitfor 'reaches the dock and its crew ties the ferry off'
+        move 'go dock'
+      elsif get_settings.textbook
         case bput("tap my #{textbook_type}", 'You tap', 'I could not', 'What were')
         when /You tap/
           ferry_textbook
--- a/bescort.lic
+++ b/bescort.lic
@@ -1447,6 +1447,8 @@ class Bescort
   end
 
   def take_xing_ferry(mode)
+    compendium_type = get_settings.compendium_type
+    textbook_type = get_settings.textbook_type
     unless [1904, 957].include?(Room.current.id)
       echo 'You are not at the ferry docks'
       return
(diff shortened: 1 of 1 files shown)


bastionwt:
The build failing, I can't see details on what might need to be changed to allow that to succeed.

rpherbig:
A timeout failed. I'm rerunning the build now.

bescort.lic
```diff
@@ -1447,6 +1447,8 @@ class Bescort
   end
 
   def take_xing_ferry(mode)
+    compendium_type = get_settings.compendium_type
```
rpherbig, line 1450: I don't think we want to get calling `get_settings` this many times. Could you cache it?

bescort.lic
```diff
@@ -1462,9 +1464,22 @@ class Bescort
 
     case bput('go ferry', 'The Captain gives you a little nod', 'You hand him', 'The ferry has just pulled away from the dock', 'There is no ferry here to go aboard', 'Come back when you can afford the fare')
     when 'You hand him', 'The Captain gives you a little nod'
-      hide?
-      waitfor 'reaches the dock and its crew ties the ferry off'
-      move 'go dock'
+      # hide?
```
rpherbig, line 1467: Why was this removed?

bescort.lic
```diff
@@ -1462,9 +1464,22 @@ class Bescort
...
+      if get_settings.textbook
+        case bput("tap my #{textbook_type}", 'You tap', 'I could not', 'What were')
```
rpherbig, line 1469: Could `DRCI.exists?` simplify this code a bit?

bescort.lic
```diff
@@ -1474,6 +1489,30 @@ class Bescort
     end
   end
 
+  def ferry_textbook
+    start_script('first-aid')
```
rpherbig, line 1493: I'm a little worried about calling `first-aid` because it may try to move around. For example: https://github.com/rpherbig/dr-scripts/blob/master/first-aid.lic#L19

What do you think?

bescort.lic
```diff
@@ -1474,6 +1489,30 @@ class Bescort
...
+    waitfor 'reaches the dock and its crew ties the ferry off'
+    stop_script('first-aid') if Script.running?('first-aid')
+    waitrt?
+    move 'go dock'
+  end
+
+  def ferry_compendium
```
rpherbig, line 1500: What do you think about using a single function that takes the script name as an argument?

vtcifer:
General comment:

I feel like this should be an option, and not assumed one always wants to be out of hiding on the ferry.  It essentially overrides `bescort_hide: true`, unless I missed something.

bescort.lic
bastionwt: I tried a couple different methods that didn't work. Do you have suggestions on how to implement?

bescort.lic
bastionwt: I was unaware of the bleed bot setting. That's not something I've ever had enabled on this side. Suggestions?

bescort.lic
bastionwt: Adding this back in as an option if they have bescort_hide: true

bastionwt:
> General comment:
> 
> I feel like this should be an option, and not assumed one always wants to be out of hiding on the ferry. It essentially overrides `bescort_hide: true`, unless I missed something.

Yep. Updated to that affect, honoring the bescort option where present.

bescort.lic
rpherbig: A few options I can think of offhand:
1. Return if `settings.bleed_bot`
2.  Add a 'nomove' mode to `first-aid` (optional script argument, I would think)
3. Add a 'no-bleedbot' mode to `first-aid` (similar but slightly different intent to option 2)

bescort.lic
rpherbig: The simplest would be something like

```
settings = get_settings
compendium_type = settings.compendium_type
textbook_type = settings.textbook_type
use_textbook = settings.textbook
```

PR #4435, opened 2020-08-14
[script] [sanowret-crystal] Configurable rooms to ignore

KatoakDR:
* Handle rooms without ids that don't support crystals to prevent script from spamming inert actions
* Allow users to configure rooms where they don't want this script to run
* Added `burgle` and `go2` to base.yaml example of when not to run this script
* Removed duplicate checks that are already handled in `check_crystal` function

### DIFF
--- a/sanowret-crystal.lic
+++ b/sanowret-crystal.lic
@@ -11,30 +11,28 @@ class SanowretCrystal
         { name: 'run', regex: /run/i, description: 'Single run of the script', optional: true }
       ]
     ]
-    @invalid_rooms = []
     @worn_crystal = false
     args = parse_args(arg_definitions)
     settings = get_settings
-    @adjective = settings.sanowret_adjective
-    @no_use_scripts = settings.sanowret_no_use_scripts
-    check_crystal if args.run && !hiding? && !invisible?
+    @adjective = settings.sanowret_adjective || 'sanowret'
+    @no_use_scripts = settings.sanowret_no_use_scripts || []
+    @no_use_rooms = settings.sanowret_no_use_rooms || []
+    check_crystal if args.run
     passive unless args.run
   end
 
   def use_crystal
-    return if @invalid_rooms.include?(Room.current.id)
-    return if hiding? || invisible?
-    return if XMLData.room_title.include? 'Carousel Chamber'
-
     if DRSkill.getxp('Arcana') <= 10 && @worn_crystal
       response = DRC.bput("gaze my #{@adjective} crystal", /A soft light blossoms in the very center of the crystal, and begins to fill your mind with the knowledge of .*\./, 'However, nothing much else happens, as you lack the concentration to focus.', 'This is not a good place for that.', 'However, you realize that you\'re already gleaning knowledge from it.', 'That would be difficult to do while you attempt to keep the trail in sight.')
     elsif DRSkill.getxp('Arcana') <= 25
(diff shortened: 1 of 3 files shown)


sanowret-crystal.lic
```diff
@@ -11,38 +11,36 @@ class SanowretCrystal
...
+    @no_use_scripts = settings.sanowret_no_use_scripts || []
+    @no_use_rooms = settings.sanowret_no_use_rooms || []
+    check_crystal if args.run
     passive unless args.run
   end
 
   def use_crystal
-    return if @invalid_rooms.include?(Room.current.id)
-    return if hiding? || invisible?
-    return if XMLData.room_title.include? 'Carousel Chamber'
```
KatoakDR, line 27: Now handled via `base.yaml` or user's custom configuration of new `sanowret_no_use_rooms` property

https://github.com/rpherbig/dr-scripts/pull/4435/files#diff-251c8759d23aa692b5efd5efe2a3e30cR1339

sanowret-crystal.lic
```diff
@@ -11,38 +11,36 @@ class SanowretCrystal
...
+    @adjective = settings.sanowret_adjective || 'sanowret'
+    @no_use_scripts = settings.sanowret_no_use_scripts || []
+    @no_use_rooms = settings.sanowret_no_use_rooms || []
+    check_crystal if args.run
     passive unless args.run
   end
 
   def use_crystal
-    return if @invalid_rooms.include?(Room.current.id)
-    return if hiding? || invisible?
```
KatoakDR, line 26: Redundant of `check_crystal` 

https://github.com/rpherbig/dr-scripts/pull/4435/files#diff-193d0b1c6934695a17f92dc36130239cL40

sanowret-crystal.lic
```diff
@@ -11,38 +11,36 @@ class SanowretCrystal
...
     end
-
     return if response !~ /This is not a good place for that\./
-    return if @invalid_rooms.include?(Room.current.id)
-    @invalid_rooms.push(Room.current.id)
+    message "Could not use crystal in room #{DRRoom.title}."
+    message "Consider adding this room to your sanowret_no_use_rooms settings."
+    # Go ahead and remember to ignore this room while the script is running
+    # This convenience won't persist between script runs though, go update your yaml
+    @no_use_rooms.push(DRRoom.title) unless @no_use_rooms.include?(DRRoom.title)
```
KatoakDR, line 35: Handle rooms without ids

sanowret-crystal.lic
```diff
@@ -11,38 +11,36 @@ class SanowretCrystal
...
     ]
-    @invalid_rooms = []
     @worn_crystal = false
     args = parse_args(arg_definitions)
     settings = get_settings
-    @adjective = settings.sanowret_adjective
-    @no_use_scripts = settings.sanowret_no_use_scripts
-    check_crystal if args.run && !hiding? && !invisible?
+    @adjective = settings.sanowret_adjective || 'sanowret'
+    @no_use_scripts = settings.sanowret_no_use_scripts || []
```
rpherbig, line 18: If this setting is in `base-empty.yaml`, we don't need the default value. If it isn't in there, it should be (and then we still don't need the default value).

Same with the other two settings.

The idea here is to avoid having every script try and provide default values - that's what the two `base` YAML files are for.

rpherbig:
Looks good other than the default value comment ✔️

sanowret-crystal.lic
KatoakDR: Ahh, understood, thanks! -- I'll double-check the `base.yaml` and `base-empty.yaml` for those settings. It's finally clicked how those are used 😅

KatoakDR: Empty defaults added per https://github.com/rpherbig/dr-scripts/pull/4435/commits/b74a4c345c365b66c68db3127654a3945c50abf5

KatoakDR:
Thanks!

PR #4436, opened 2020-08-15
Spin.lic - Spins 400 yards of silk instead of 390

BinuDR:
The script will now cut 10 yards off the first set of fibers it buys and drop the excess 20 from that, then buy the other 390 so it ends up with 400 yards total.
This allows you to not waste any when converting to cloth.

### DIFF
--- a/spin.lic
+++ b/spin.lic
@@ -34,21 +34,21 @@ class Spin
     @belt = @settings.outfitting_belt
     ensure_copper_on_hand(data[:cost] * data[:count], @settings)
 
-    data[:count].times do |c|
+    data[:count].times do |index|
       stock_room = get_data('crafting')['tailoring'][@settings.hometown]['stock-room']
       order_item(stock_room, data[:index])
-      if c == 0 && data[:name] == 'silk'
-        fput('mark my fibers at 10')
+      if index == 0 && data[:name] == 'silk'
+        bput('mark my fibers at 10', /mark it for cutting/)
         get_item('scissors')
-        fput('cut my fibers with my scissors')
+        bput('cut my fibers with my scissors', /You carefully cut off/)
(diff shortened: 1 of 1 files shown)


rpherbig:
Could you have someone from Slack test this?

spin.lic
```diff
@@ -20,24 +20,36 @@ class Spin
...
-    ensure_copper_on_hand(data[:cost] * data[:count], settings)
+    @settings = get_settings
+    @bag = @settings.crafting_container
+    @bag_items = @settings.crafting_items_in_container
+    @belt = @settings.outfitting_belt
+    ensure_copper_on_hand(data[:cost] * data[:count], @settings)
 
-    data[:count].times do
-      stock_room = get_data('crafting')['tailoring'][settings.hometown]['stock-room']
+    data[:count].times do |c|
```
rpherbig, line 37: 💡 I prefer `index`

spin.lic
```diff
@@ -20,24 +20,36 @@ class Spin
...
+      stock_room = get_data('crafting')['tailoring'][@settings.hometown]['stock-room']
       order_item(stock_room, data[:index])
-      fput('combine my fiber')
+      if c == 0 && data[:name] == 'silk'
+        fput('mark my fibers at 10')
```
rpherbig, line 41: 💡 Could these use `bput`?

BinuDR:
> Could you have someone from Slack test this?

Totenus and I have tested it so far.

spin.lic
BinuDR: I'll take care of that.

spin.lic
BinuDR: Yes, I'll update it in a bit.

PR #4416, opened 2020-08-08
[script] [combat-trainer] Handle already worn items whose noun is "moon"

KatoakDR:
Honestly, I don't know what line 840 is trying to do but I'm a Moon Mage who wears "moonsilk shirt" and "moonsilk pants" armor and this line hangs because the command "wear moon" referenced something the script didn't anticipate.

If there's a better fix, as well as if someone could clarify what "wear moon" refers to, that'd be helpful :)

PR #4439, opened 2020-08-15
[script] [combat-trainer] [common-items] Be explicit when 'wear moon' to reduce conflicts with worn items

KatoakDR:
Response to this feedback: https://github.com/rpherbig/dr-scripts/pull/4416#issuecomment-671445603

* Be explicit about wearing what's in the right/left hand to minimize chances of conflicting with already worn items whose noun is "moon"
* Previous forked repo was deleted and GitHub lost track of the original branch for https://github.com/rpherbig/dr-scripts/pull/4416 so needed to create this new PR

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -266,9 +266,7 @@ class SetupProcess
     elsif !next_summoned && !DRStats.moon_mage?
       break_summoned_weapon(game_state.last_weapon_name)
     elsif !next_summoned && DRStats.moon_mage?
-      if right_hand =~ /moon/ || left_hand =~ /moon/
-        bput('wear moon', 'telekinetic')
-      end
+      wear_moon_weapon
     end
 
     # Prepare the next weapon
@@ -371,7 +369,7 @@ class LootProcess
 
     @rituals = get_data('spells').rituals
     echo("  @rituals: #{@rituals}") if $debug_mode_ct
-    
+
     @redeemed = settings.necro_redeemed
     echo("  @redeemed: #{@redeemed}") if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


common-items.lic
```diff
@@ -268,4 +268,18 @@ module DRCI
...
+  def wear_moon_weapon
+    moon_weapon_regex = /moon/
+    moon_wear_messages = ['telekinetic']
+    if left_hand =~ moon_weapon_regex
+      bput("wear #{left_hand}", *moon_wear_messages)
+    end
+    if right_hand =~ moon_weapon_regex
+      bput("wear #{right_hand}", *moon_wear_messages)
+    end
+  end
```
KatoakDR, line 284: ### Why add method to common-items?
This method exists in `common-items` mainly because the two spots in `combat-trainer` that needed to be updated with this logic exists in two different classes. I wasn't sure where to put this method in `combat-trainer` where it'd be accessible by both classes so here instead. It might have use for other scripts, too?

### moon_weapon_regex

There are multiple variants of a summoned moon weapon (moonblade, moonstaff) and many more if you have the [Shape Moonblade](https://elanthipedia.play.net/Shape_Moonblade) spell. The idea here is that if you have a "moon" noun in your hand and you're calling `wear_moon_weapon` then that moon noun is (should be) a summoned moon weapon.

To minimize conflict with already worn moon nouns, like "a trio of moons", then the wear command uses the exact <adjective> <noun> listed in your hand.

rpherbig:
Can you split this into two PRs, one with the `combat-trainer` changes and one with the `common-items` changes? They need to be merged separately.

combat-trainer.lic
```diff
@@ -266,9 +266,7 @@ class SetupProcess
     elsif !next_summoned && !DRStats.moon_mage?
       break_summoned_weapon(game_state.last_weapon_name)
     elsif !next_summoned && DRStats.moon_mage?
-      if right_hand =~ /moon/ || left_hand =~ /moon/
-        bput('wear moon', 'telekinetic')
-      end
+      wear_moon_weapon
```
rpherbig, line 269: I'd prefer to see the module prefix (`DRCI.wear_moon_weapon`), but it's not a big deal

KatoakDR: Sure thing

KatoakDR:
> Can you split this into two PRs, one with the `combat-trainer` changes and one with the `common-items` changes? They need to be merged separately.

Will do

Issue #798, opened 2016-09-14
Attacked another player through trainer


Endraig:
Attacked warchyld during my goblin hunt about 7/8ths through this document, not sure what happened.

[Wallet(6) (1).txt](https://github.com/rpherbig/dr-scripts/files/473193/Wallet.6.1.txt)

rpherbig:
This really makes no sense to me. On line 29604:
```
[combat-trainer]>aim

You begin to target a scavenger goblin.
```

Then a few bobs, then on line 29641:
```
[combat-trainer]>shoot

Your attempt is slightly hindered by your target shield.
< As if fumbling muscle flab were natural, you fire a boar-tusk arrow at Warchyld.  Warchyld dodges.  
The boar-tusk arrow lands nearby!
[You're adeptly balanced and in dominating position.]
[Roundtime 1 sec.]
```

Why would shoot be at a different target than the one at which you aimed?

rcuhljr:
Looks like DR broke, maybe that's what they restarted the servers for?

PR #4445, opened 2020-08-18
Create buffother.lic

bastionwt:
New helper script to use when wanting to buff a friend or alt. Has required args for who you buff (recipient) and which waggle set to call (waggle), with an optional room number arg (room) to move you there first before buffing. The script can be called manually, as an item in your t2 training_list, or as a before/after in a hunt file. Example usage:

;buffother Etreu buffsomeone 798

That will move you to room #798 and cycle through each of the spells in your buffsomeone waggle set, appending a cast of "cast Etreu" to each of them.

Example waggle setup shown below:

```
   buffsomeone:
    Shadows:
      abbrev: Shadows
      recast: 2
      mana: 20
      cambrinth:
      - 8
      - 8
      - 8
      - 8
    Clear Vision:
      abbrev: CV
      recast: 2
      mana: 20
      cambrinth:
      - 8
      - 8
      - 8
      - 8
    Psychic Shield:
      abbrev: PSY
      recast: 2
      mana: 20
      cambrinth:
      - 8
      - 8
      - 8
      - 8
```

### DIFF
--- a/buffother.lic
+++ b/buffother.lic
@@ -20,8 +20,9 @@ class Buffother
     @recipient = args.recipient.capitalize
     @waggle = args.waggle
     @room = args.room
+    
     move_to_room
-    cast_buff_spells
+    do_buffs
   end
 
   def move_to_room
@@ -29,16 +30,81 @@ class Buffother
     pause 1
   end
 
-  def cast_buff_spells
-    data = @settings.waggle_sets["#{@waggle}"].values.each do |data|
-      data['cast'] = "cast #{@recipient}"
-      DRCA.do_buffs(@settings, @waggle)
+  def do_buffs
+    # takes a waggle and sends it to the appropriate helper
+    return unless @settings.waggle_sets["#{@waggle}"]
+    spells = @settings.waggle_sets["#{@waggle}"]
+
+    spells.values
+          .select { |spell| spell['use_auto_mana'] }
+          .each { |spell| check_discern(spell, settings) }
+    cast_spells(spells, @settings, @settings.waggle_force_cambrinth)
+  end
+
+  def cast_spells(spells, settings, force_cambrinth = false)
+    spells.each do |name, data|
+      next if DRSpells.active_spells[name] && (data['recast'].nil? || DRSpells.active_spells[name].to_i > data['recast'])
+      while (mana < settings.waggle_spells_mana_threshold || DRStats.concentration < settings.waggle_spells_concentration_threshold)
+        echo("Waiting on mana over #{settings.waggle_spells_mana_threshold} or concentration over #{settings.waggle_spells_concentration_threshold}...")
+        pause 15
+      end
+
+      cast_spell(data, settings, force_cambrinth)
     end
   end
+
+def cast_spell(data, settings, force_cambrinth = false)
+    return unless data
+    return unless settings
+
+    data = update_astral_data(data)
+    data['cast'] = "cast #{@recipient}"
+    return unless data # update_astral_data returns nil on failure
+
+    release_cyclics if data['cyclic']
+    DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell") unless checkprep == 'None'
+    DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
+
+    if data['ritual']
+      ritual(data, settings)
+      return
+    end
+
+    command = 'prep'
+    command = data['prep'] if data['prep']
+    command = data['prep_type'] if data['prep_type']
+
+    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command)
+    prepare_time = Time.now
+
+    unless settings.cambrinth_items[0]['name']
+      settings.cambrinth_items = [{
+        'name' => settings.cambrinth,
+        'cap' => settings.cambrinth_cap,
+        'stored' => settings.stored_cambrinth
+      }]
+    end
+    settings.cambrinth_items.each_with_index do |item, index|
+      case data['cambrinth'].first
+      when Array
+        find_charge_invoke_stow(item['name'], item['stored'], item['cap'], settings.dedicated_camb_use, data['cambrinth'][index], settings.cambrinth_invoke_exact_amount)
+      when Fixnum, Integer
+        find_charge_invoke_stow(item['name'], item['stored'], item['cap'], settings.dedicated_camb_use, data['cambrinth'], settings.cambrinth_invoke_exact_amount)
+      end
+    end
+
+    if data['prep_time']
+      pause 0.1 until checkcastrt.zero? || Time.now - prepare_time >= data['prep_time']
+    else
+      waitcastrt?
+    end
+    cast?(data['cast'], data['symbiosis'], data['before'], data['after'])
+  end
+
 end
 
 before_dying do
-    fput("release mana")
+  fput("release mana")
 end
 
 Buffother.new
--- a/buffother.lic
+++ b/buffother.lic
@@ -16,24 +16,24 @@ class Buffother
       ]
     ]
     args = parse_args(arg_definitions)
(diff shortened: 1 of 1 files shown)


buffother.lic
```diff
@@ -0,0 +1,47 @@
...
+    @recipient = args.recipient.capitalize
+    @waggle = args.waggle
+    @room = args.room
+    move_to_room
+  end
+
+  def move_to_room
+    echo "Moving to room"
+    DRCT.walk_to(@room)
+    cast_buff_spells
```
KatoakDR, line 29: To decouple the responsibilities of the methods, I recommend moving line 29 (cast_buff_spells) to just after line 23. In this way, the logic for moving to a room does not take on the logic of what is done while in that room. The higher order function `initialize` then orchestrates the two behaviors.

buffother.lic
```diff
@@ -0,0 +1,47 @@
...
+    args = parse_args(arg_definitions)
+    @settings = get_settings
+    @recipient = args.recipient.capitalize
+    @waggle = args.waggle
+    @room = args.room
+    move_to_room
+  end
+
+  def move_to_room
+    echo "Moving to room"
```
KatoakDR, line 27: Consider controlling echo output via a `debug` flag. I've seen that technique used in some other scripts.

A simple example is in [dusk-labyrinth](https://github.com/rpherbig/dr-scripts/blob/master/dusk-labyrinth.lic).

The change would include adding another arg definition:
```
arg_definitions = [
  [
    ...
    { name: 'debug', regex: /debug/i, optional: true, description: 'Enable debug output' }
  ]
]
```

Then declare a debug variable that looks at UserVars first, then the script argument, then default value (usually false for off).
```
$debug_mode_buffother = UserVars.buffother_debug || args.debug || false
```

Then modify the echo statements to only output if the debug flag is set.
```
echo "Moving to room" if $debug_mode_buffother
```

buffother.lic
```diff
@@ -0,0 +1,47 @@
...
+    data = @settings.waggle_sets["#{@waggle}"].values.each do |data|
+      data['cast'] = "cast #{@recipient}"
+      DRCA.do_buffs(@settings, @waggle)
+    end
+    cleanup
+  end
+
+  def cleanup
+    fput("release mana")
+  end
```
KatoakDR, line 44: To ensure this happens even if the script exits prematurely, either on purpose (user stopping script) or by some other reason, I recommend adding a [before_dying](https://elanthipedia.play.net/Lich_scripting_reference#before_dying) method at the end of this. class that calls your `cleanup` method.

```
before_dying do
    cleanup
end
```

Additionally, since this ensures your cleanup method is called on script exit, you could remove line 39 as it would be redundant to the failsafe invocation of it.

buffother.lic
```diff
@@ -0,0 +1,47 @@
...
+    DRCT.walk_to(@room)
+    cast_buff_spells
+    pause 1
+  end
+
+  def cast_buff_spells
+    echo "In cast_buff_spells"
+    data = @settings.waggle_sets["#{@waggle}"].values.each do |data|
+      data['cast'] = "cast #{@recipient}"
+      DRCA.do_buffs(@settings, @waggle)
```
KatoakDR, line 37: Note, because of https://github.com/rpherbig/dr-scripts/blob/master/common-arcana.lic#L355, if this character who is casting the spells somehow already has the same spell active on themself, I don't think the buff will end up being cast on _the other character_.

KatoakDR:
@bastionwt neat use of hooking into the existing buff script to cast on another character

buffother.lic
bastionwt: Love it. Making that change.

buffother.lic
bastionwt: I didn't intend on committing the version with echoes included. Those will be removed. That was my own internal troubleshooting in putting the script together.

buffother.lic
bastionwt: Love it. Change made.

buffother.lic
bastionwt: They'll need to set the recast: value high in the waggle they're calling. i.e. recast: 100 instead of recast: 1 or recast: 2 which is more normal. As long as the buff on themself has a duration of less than that large value, it'll be fine.

bastionwt:
Hmm, latest change loops infinitely through the buffing. More work to be done.

bastionwt:
> Hmm, latest change loops infinitely through the buffing. More work to be done.

Actually, it'll loop infinitely as long as those buffs aren't up on the buffer when using do_buffs. Will need to pursue an alternate method to this. cast_spell was another method I'd used, but it doesn't take into account remaining mana, which is why I initially swapped over to do_buffs.

Back to the drawing board.

BinuDR:
You can use a line in the buffing section before it starts each new buff like this:

`pause 1 while DRStats.mana < 70`

bastionwt:
> You can use a line in the buffing section before it starts each new buff like this:
> 
> `pause 1 while DRStats.mana < 70`

The buffing section is being handled by common-arcana exclusively, so I'm not sure an easy way of doing that without making changes to the core script.

bastionwt:
Okay, latest update in.

- Solves the infinite loop. It will loop through each of the buffs a single time.
- Solves the mana issue. It properly pauses if mana drops below threshold values.
- And it reuses a ton of code stolen from common-arcana. I'm not sure I like that, but each of the functions needed some subtle manipulation (adds/removes) in order to accomplish the objective. I don't think that this is perfect, but it solves each of the problems previously encountered so far.

bastionwt:
Okay. Chatted some with Binu and he recommended a MUCH more elegant solution. Basically back to the original commit, with a small update to the data and mana usage. Thanks Binu!

KatoakDR:
Thanks @BinuDR for the insights to help keep this script as a small wrapper.

Thanks @bastionwt for taking all the feedback in stride!

rpherbig:
I ❤️ this PR so much

buffother.lic
```diff
@@ -0,0 +1,46 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#buffother
+=end
+custom_require.call(%w(common common-arcana events spellmonitor drinfomon common-travel))
+class Buffother
+  include DRC
```
rpherbig, line 6: 💡 Where is `common` used?

buffother.lic
```diff
@@ -0,0 +1,46 @@
...
+    cast_buff_spells
+  end
+
+  def move_to_room
+    DRCT.walk_to(@room)
+    pause 1
+  end
+
+  def cast_buff_spells
+    data = @settings.waggle_sets["#{@waggle}"].values.each do |data|
```
rpherbig, line 33: 💡 No need for string interpolation here

buffother.lic
```diff
@@ -0,0 +1,46 @@
...
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'recipient', regex: /\w+/, description: 'Person to buff' },
+        { name: 'waggle', regex: /\w+/, description: 'Waggle set to cast' },
+        { name: 'room', regex: /\d+/, optional: true, description: 'Room to move to' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
+    @settings = get_settings
```
rpherbig, line 19: 💡 I'd prefer to see these as non-instance vars and pass them as function args. What do you think?

bastionwt: While I don't actually understand the merits one way or another (noob here), I've made those adjustments.

buffother.lic
bastionwt: Fixed!

buffother.lic
bastionwt: It was at one point, but isn't used anymore. I'll remove it on the next commit.

bastionwt:
Updated in light of Sheltim's suggestions.

PR #4446, opened 2020-08-19
[script] [equipmanager] Updates "wear/remove" logic to handle tie_to, wield, container

KatoakDR:
### What's changing?

The `remove_item` and `wear_item` functions were out of date with new gear options for where/how an item should be stowed, such as tied to a strap, use "wield/sheath" mechanics, "transforming" items, or put it in a specified container.

A recent change to this file made some progress. This change continues the improvement effort.

The updates are largely based on what `return_held_gear` is doing.

### Why is this necessary?

An example where the existing logic fails is I wear a longbow that I want tied to a bag with straps. When trying to wear a gearset that includes the longbow, the existing logic tries to `get longbow` and fails with "You should untie the longbow from the bag first." With these changes, the logic respects the `tie_to`, `wield`, or `container` settings.

Similarly, if someone were to _not_ specify a gear item's `tie_to`, `wield`, or `container` in their yaml, then the default case when trying to remove an item and just "stow longbow" may fail with "The longbow is too long to fit in the backpack." With these changes, the character will re-wear the item and keep their hands empty as the original logic was doing.

FYI @vtcifer

### DIFF
--- a/equipmanager.lic
+++ b/equipmanager.lic
@@ -142,16 +142,12 @@ class EquipmentManager
       echo('failed to match an item, try turning on debugging with ;e UserVars.equipmanager_debug = true')
       return false
     end
-    case bput("get my #{item.short_name}", item.short_regex, '^You get ', '^What were you referring')
-    when item.short_regex
+    if get_item?(item)
       bput("wear my #{item.short_name}", 'You sling', 'You attach', 'You .* unload', 'You strap', 'You slide', 'You work your way into', 'You spin', 'You put', 'You carefully loop', 'slide effortlessly onto your', 'You slip', "^You .* #{item.short_regex}")
       waitrt?
       return true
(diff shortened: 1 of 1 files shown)


equipmanager.lic
```diff
@@ -142,16 +146,12 @@ class EquipmentManager
       echo('failed to match an item, try turning on debugging with ;e UserVars.equipmanager_debug = true')
       return false
     end
-    case bput("get my #{item.short_name}", item.short_regex, '^You get ', '^What were you referring')
-    when item.short_regex
+    if get_item?(item)
```
KatoakDR, line 149: Rather than assume "get my <item>" will work, now use the `get_item?` helper function that respects the `tie_to`, `wield`, and `container` options so knows when to do "untie my <item>".

PR #4447, opened 2020-08-21
Add forced recast timer to buffother.

jayRyan24:
- Added some structure to make it more readable. 
- Added protection logic to not sit and spam with errors if target not found. 
- Forced a recast time to eliminate the recommended be set high to avoid conflict with personal buff timers.
- Made variables class scope.
- Removed excess code.

### DIFF
--- a/buffother.lic
+++ b/buffother.lic
@@ -1,38 +1,50 @@
 =begin
   Documentation: https://elanthipedia.play.net/Lich_script_repository#buffother
 =end
-custom_require.call(%w(common-arcana events spellmonitor drinfomon common-travel))
-class Buffother
+
+custom_require.call(%w(common-arcana common-travel))
+
+class BuffOther
   include DRCA
   include DRCT
 
   def initialize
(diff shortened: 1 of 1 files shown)


jayRyan24:
I am wondering if this can be merged into buff.lic. I would think that we could make buff.lic take in an optional name and room number. We can then do all the action there. Not needed, but that's my thought. The other thought is that we can also make a new method within common-arcana to handle this. Buff.lic uses do_buffs. We could make a new method buff_other or do_buffs_other, or override do_buffs to take an optional name. This would do the basic work of what the script buff_other does. It seems we have other script(s) we could just refactor to do what this new script does. But as it is, we can add in a new method into common-arcana which does the prementioned things, so we can remove that logic out of BuffOther in case there's another place we'd want to handle this. As in, grouping and buffing others while hunting. spells_process can simple check if buffing_targets or some other dataset is present, and if so, buff them and then wait, etc. Just an idea.

PR #4451, opened 2020-08-23
[script] [walkingastro] observe only if can see sky, too

KatoakDR:
Depends on https://github.com/rpherbig/dr-scripts/pull/4450

### DIFF
--- a/walkingastro.lic
+++ b/walkingastro.lic
@@ -76,8 +76,18 @@ class WalkingAstro
     waitrt?
   end
 
+  def can_see_sky?
+    # If we didn't see the sky last time we checked
+    # and we're in the same room still then don't redundantly check
+    # so that we don't spam the game window with "That's a bit hard to do while inside".
+    return false if @last_room_check == DRRoom.title && @last_sky_check == false
+    @last_sky_check = DRC.can_see_sky?
+    @last_room_check = DRRoom.title
+    return @last_sky_check
+  end
+
   def should_observe?
-    @no_use_scripts.none? { |name| Script.running?(name) } && (!lefthand && !righthand) && DRC.can_see_sky?
+    @no_use_scripts.none? { |name| Script.running?(name) } && (!lefthand && !righthand) && can_see_sky?
   end
 
   def main_loop
--- a/walkingastro.lic
+++ b/walkingastro.lic
@@ -77,7 +77,7 @@ class WalkingAstro
   end
 
   def should_observe?
-    @no_use_scripts.none? { |name| Script.running?(name) } && (!lefthand && !righthand)
+    @no_use_scripts.none? { |name| Script.running?(name) } && (!lefthand && !righthand) && DRC.can_see_sky?
   end
 
   def main_loop
--- a/walkingastro.lic
+++ b/walkingastro.lic
@@ -6,7 +6,18 @@ custom_require.call %w[common common-moonmage drinfomon]
 
 class WalkingAstro
   include DRC
+
   def initialize
+
+    arg_definitions = [
+      [
+        { name: 'debug', regex: /debug/i, optional: true, description: 'Enable debug output' },
+      ]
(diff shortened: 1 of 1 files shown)


walkingastro.lic
```diff
@@ -76,8 +76,18 @@ class WalkingAstro
...
 
+  def can_see_sky?
+    # If we didn't see the sky last time we checked
+    # and we're in the same room still then don't redundantly check
+    # so that we don't spam the game window with "That's a bit hard to do while inside".
+    return false if @last_room_check == DRRoom.title && @last_sky_check == false
+    @last_sky_check = DRC.can_see_sky?
+    @last_room_check = DRRoom.title
+    return @last_sky_check
+  end
```
KatoakDR, line 87: Using the room title instead of the room id because not all rooms have ids, such as in a vault or maze.

walkingastro.lic
```diff
@@ -76,8 +76,18 @@ class WalkingAstro
     waitrt?
   end
 
+  def can_see_sky?
+    # If we didn't see the sky last time we checked
+    # and we're in the same room still then don't redundantly check
+    # so that we don't spam the game window with "That's a bit hard to do while inside".
+    return false if @last_room_check == DRRoom.title && @last_sky_check == false
```
rpherbig, line 83: If we're still in the same room and the last sky check was true, is there a reason to check the weather again?

I.e. could this line instead be `return @last_sky_check if @last_room_check == DRRoom.title`?

KatoakDR: Great point, yes!

walkingastro.lic
```diff
@@ -6,7 +6,18 @@ custom_require.call %w[common common-moonmage drinfomon]
...
   def initialize
+
+    arg_definitions = [
+      [
+        { name: 'debug', regex: /debug/i, optional: true, description: 'Enable debug output' },
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+    $debug_mode_walkingastro = UserVars.walkingastro_debug || args.debug || false
```
KatoakDR, line 19: Adds debug mode support

walkingastro.lic
```diff
@@ -69,15 +80,51 @@ class WalkingAstro
...
+    end
+    if $debug_mode_walkingastro
+      echo "--- room_id=#{Room.current.id}"
+      echo "--- room_title=#{DRRoom.title}"
+      echo "--- last_room_id=#{@last_room_id}"
+      echo "--- last_room_title=#{@last_room_title}"
+      echo "--- moved_rooms=#{moved_rooms}"
+    end
+    return moved_rooms
+  end
```
KatoakDR, line 109: Check if we've moved rooms, and if so then check if we can see the sky. If we haven't changed rooms then return the answer from the first time we checked.

Handles rooms that don't have ids by checking titles instead.

walkingastro.lic
```diff
@@ -69,15 +80,51 @@ class WalkingAstro
...
+    if moved_rooms?
+      echo "Moved rooms, checking if we can see sky in this room" if $debug_mode_walkingastro
+      @can_see_sky = DRC.can_see_sky?
+      @last_room_id = Room.current.id
+      @last_room_title = DRRoom.title
+    else
+      echo "In same room, returning previous answer if we can see the sky: #{@can_see_sky}"
+    end
+    return @can_see_sky
+  end
```
KatoakDR, line 124: If we've moved rooms since the last time we checked if we can see the sky, then check if we can see the sky. Otherwise, just return the previous answer.

walkingastro.lic
KatoakDR: Resolved in commit https://github.com/rpherbig/dr-scripts/pull/4451/commits/43594821a601f6cd3f371d2f89e58a8d0d525a91

KatoakDR:
Thanks!

PR #4453, opened 2020-08-25
Bescort - Edit to Segoltha section

BinuDR:
Added option to use the entrance in Crossing.  You need 1001 athletics or more to use it, and for now there's also a UserVar that needs to be set so that it can be tested.  The athletics requirement is currently higher than it really needs to be.  I'll update it when we figure out a better number.

Here is the var that needs to be set.
`segoltha_shortcut: true`


BinuDR:
In case anyone looks back at this, the athletics check was updated to needing 540 or more ranks.

PR #4442, opened 2020-08-17
[script] [combat-trainer] Safely wear summoned moon weapon

KatoakDR:
### Changes
* The utility method greatly reduces chances that the command `wear moon` would conflict with another moon noun item the character is wearing.
* Not all mages can snap cast the moonblade spell nor every attempt to break a summoned moonblade may work when trying to create slivers as telekinetic ammo. This change adds logic to retry casting the spell up to 3 times. Maintains the original "snap cast" prep time logic that was here before.

For more background context, see conversations on:
* https://github.com/rpherbig/dr-scripts/pull/4416
* https://github.com/rpherbig/dr-scripts/pull/4439

PR #4455, opened 2020-08-26
[script] [common-moonmage] Explicitly reference DRC module

KatoakDR:
In some contexts when `wear_moon_weapon` is called then `left_hand` and `right_hand` variables are not in scope. This ensures we reference the class variables explicitly via `DRC.left_hand` or `DRC.right_hand`.

```
--- Lich: error: undefined local variable or method `left_hand' for DRCMM:Module
Did you mean?  lefthand
               lefthand?
	common-moonmage:130:in `wear_moon_weapon'
	combat-trainer:839:in `check_skinning'

```

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -128,13 +128,14 @@ module DRCMM
   # https://elanthipedia.play.net/Shape_Moonblade
   def wear_moon_weapon?
     moon_weapon_regex = /moon[\w]+/
-    moon_wear_messages = ['telekinetic', 'already wearing', 'Wear what']
+    moon_wear_messages = ["telekinetic", "already wearing", "You can't wear", "Wear what"]
+    wore_it = false
     if DRC.left_hand =~ moon_weapon_regex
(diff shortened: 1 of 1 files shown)


common-moonmage.lic
```diff
@@ -123,15 +123,18 @@ module DRCMM
 
   # There are many variants of a summoned moon weapon (blade, staff, sword, etc)
   # This function checks if you're holding one then tries to wear it.
+  # Returns true if what is in your hands is a summoned moon weapon that becomes worn.
+  # Returns false if you're not holding a moon weapon, or you are but can't wear it.
   # https://elanthipedia.play.net/Shape_Moonblade
   def wear_moon_weapon
```
rpherbig, line 129: Can you also rename this to `wear_moon_weapon?` since its now a predicate?

KatoakDR: Will do. This change to return a boolean supports another spot I found in combat-trainer that was still doing the old “wear moon” action and script would hang because my character was wearing “trio of moons”.

https://github.com/rpherbig/dr-scripts/pull/4442/files#diff-772f69cb948ff35829a81342b2805fa8R839

PR #4462, opened 2020-08-28
Update Bard Sample to match the current batch

Hiinky:
Sample is 2 yrs old and didn't get the same update as the others.  Moonie file coming soon.


rpherbig:
Thank you!

PR #4464, opened 2020-08-28
Rough Draft of new docs

Hiinky:
This isn't really ready for posting as its short and only has a few fields but I wanted to see what people think.  Am I over thinking this?  Making it too complicated?

### DIFF
--- a/repository.md
+++ b/repository.md
@@ -1,0 +1,45 @@
+## dr-scripts variables list
+
+A Work in Progress
+
+### bare bones
+No matter if only use lich for go2 travel magic or have customized every possible option, these fields are universally helpful.
+script | yaml_field | explanation
+------------ | ------------ | -------------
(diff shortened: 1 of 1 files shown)


Hiinky:
@BinuDR @bastionwt @rcuhljr @rpherbig  @anyone else with an opinion  hehe

rcuhljr:
I mean start small, build on it. I'd probably make a welcome/landing page first and start breaking out each of these areas into smaller ones but getting anything up in place is step 1.

rpherbig:
I love the idea. I'm no expert in knowledge management, and that's really what this is.

What I do know about KM is there are two goals that sometimes compete:
1. Findability - can the user find what they're searching for?
1. Discoverability - if the user doesn't know what to search for, can they find it anyway?

I think that's a pretty good mental model to start with. There are lots of ways we could organize this - by script, by related settings, by "thing the user would want to do"... I don't know the right answer.

Can we start by trying to understand what questions people are asking that we think documentation can help with? That should point us in a direction to get started.

PR #4465, opened 2020-08-28
[script] [locksmithing] Wait to begin playing before begin picking

KatoakDR:
Depending on your config, the `locksmithing` and `performance` scripts may need to remove worn items (e.g. hand armor). 

Currently, the `locksmithing` script starts the `performance` script asynchronously and moves on. There is a race condition where one or both scripts are trying to remove worn items and get boxes/instruments and your hands become full and `bput` commands stall out.

## Key Changes
* Wait for the `performance` script to get the character playing the instrument _then_ resume picking.
* If `locksmithing` script ends abruptly (on purpose or otherwise) then stop `performance` script, too.

---

FYI @Hiinky

### DIFF
--- a/locksmithing.lic
+++ b/locksmithing.lic
@@ -22,10 +22,11 @@ class Locksmithing
     @equipment_manager.empty_hands
 
     live_boxes if @liveboxes
+    # Practice some music while use lock trainers
     start_script('performance', ['noclean']) unless Script.running?('performance')
+    waitfor('as you begin')
     daily_lockbox if @daily_trainer
     consumable_lockbox if @consumable_trainers
-    kill_script('performance') if running?('performance')
   end
 
   def daily_lockbox
@@ -115,6 +116,10 @@ class Locksmithing
         return
       end
     end
+  end
+
+  before_dying do
+    kill_script('performance') if running?('performance')
   end
 end
 


Hiinky:
I don't think I'm following.  Why would you be removing hand armor for training boxes or zills?  I do like the before dying update though.  Also, this was just an edit I added for someone else. 😁

rpherbig:
@KatoakDR ✔️ looks good to me. Let me if you and @Hiinky decide to change anything or if this is ready to be merged.

Hiinky:
I'd say go for it.  No reason not to.

KatoakDR:
Thanks for the quick turnaround and merge :)

To answer the question about why performance is removing armor... I'm not sure, but it does [here](https://github.com/rpherbig/dr-scripts/blob/5c952780da10e959a37c5ee6df8fc38d1aed1509/performance.lic#L31). And the `@settings.hand_armor` is defined in `base.yaml` [here](https://github.com/rpherbig/dr-scripts/blob/5c952780da10e959a37c5ee6df8fc38d1aed1509/profiles/base.yaml#L386) as "gloves", which coincidentally my hand armor is gloves so the script tried to remove them.

The two scripts then got into this state where one hand held a keepsake box, one hand had my gloves, and one script is trying to lock and pick, and the other script is trying to stow the gloves and pull out my instrument from my backpack... and they all got tangled up 😆

PR #4470, opened 2020-08-30
[script] [common-moonmage] Helper functions to wear/hold moon weapons

KatoakDR:
Bulk of the logic is to ensure the regex patterns match to actual moon weapons and not get confused by items whose nouns are or start with the word "moon", like "a trio of moons" or "moonsilk shirt". I'm personally invested in these changes because my moon mage wears both those items, haha.

This is a pre-requisite for https://github.com/rpherbig/dr-scripts/pull/4471

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -127,15 +127,46 @@ module DRCMM
   # Returns false if you're not holding a moon weapon, or you are but can't wear it.
   # https://elanthipedia.play.net/Shape_Moonblade
   def wear_moon_weapon?
-    moon_weapon_regex = /moon[\w]+/
-    moon_wear_messages = ["telekinetic", "already wearing", "You can't wear", "Wear what"]
+    moon_weapon_regex = /((black|red-hot|blue-white) moon[\w]+)/
+    moon_wear_messages = ["You're already", "You can't wear", "Wear what", "telekinetic"]
     wore_it = false
     if DRC.left_hand =~ moon_weapon_regex
-      wore_it = wore_it || bput("wear #{DRC.left_hand}", *moon_wear_messages) =~ /telekinetic/
+      wore_it = wore_it || DRC.bput("wear #{DRC.left_hand}", *moon_wear_messages) == "telekinetic"
     end
     if DRC.right_hand =~ moon_weapon_regex
-      wore_it = wore_it || bput("wear #{DRC.right_hand}", *moon_wear_messages) =~ /telekinetic/
+      wore_it = wore_it || DRC.bput("wear #{DRC.right_hand}", *moon_wear_messages) == "telekinetic"
     end
     return wore_it
   end
+
+  # Is a moon weapon in your hands?
+  def holding_moon_weapon?
+    moon_weapon_regex = /((black|red-hot|blue-white) moon[\w]+)/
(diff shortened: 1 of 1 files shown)


common-moonmage.lic
```diff
@@ -127,15 +127,46 @@ module DRCMM
   # Returns false if you're not holding a moon weapon, or you are but can't wear it.
   # https://elanthipedia.play.net/Shape_Moonblade
   def wear_moon_weapon?
-    moon_weapon_regex = /moon[\w]+/
-    moon_wear_messages = ["telekinetic", "already wearing", "You can't wear", "Wear what"]
+    moon_weapon_regex = /((black|red-hot|blue-white) moon[\w]+)/
+    moon_wear_messages = ["You're already", "You can't wear", "Wear what", "telekinetic"]
```
KatoakDR, line 131: Move "telekinetic" match to last because if you're already wearing a moon weapon and you have a second one in your hands and you try to wear it the game will respond with "You're already telekinetically supporting <moon weapon>". So the word "telekinetic" which is to match the success case was also matching the failure case.

Another consideration instead of matching on "telekinetic" for the success case is "You gently suspend" but I'm not experienced enough with moon weapons to know if that messaging ever changes due to weapon shape, strength of the spell, etc etc so I kept it as telekinetic.

common-moonmage.lic
```diff
@@ -127,15 +127,46 @@ module DRCMM
...
     wore_it = false
     if DRC.left_hand =~ moon_weapon_regex
-      wore_it = wore_it || bput("wear #{DRC.left_hand}", *moon_wear_messages) =~ /telekinetic/
+      wore_it = wore_it || DRC.bput("wear #{DRC.left_hand}", *moon_wear_messages) == "telekinetic"
     end
     if DRC.right_hand =~ moon_weapon_regex
-      wore_it = wore_it || bput("wear #{DRC.right_hand}", *moon_wear_messages) =~ /telekinetic/
+      wore_it = wore_it || DRC.bput("wear #{DRC.right_hand}", *moon_wear_messages) == "telekinetic"
```
KatoakDR, line 137: When testing this script one-off via `;e echo DRCMM.wear_moon_weapon?` Lich complained about `bput` function didn't exist so I fully qualified the method call.

Additionally, I wanted to ensure this function returned a proper boolean not just a "truthy/falsey" value so changed the `=~` to `==`.

common-moonmage.lic
```diff
@@ -127,15 +127,46 @@ module DRCMM
...
+    look_result = DRC.bput("look #{Char.name}", moon_floating_regex, "You are wearing")
+    # Now try to extract just the moon weapon name from what we saw.
+    scan_result = look_result.scan(moon_weapon_regex)
+    # If a match was found then grab it from the captured groups per the regex.
+    if scan_result.length > 0 && scan_result[0].length > 0
+      moon_weapon = scan_result[0][0]
+      held_it = DRC.bput("hold #{moon_weapon}", *moon_hold_messages) == "You grab"
+    end
+    return held_it
+  end
```
KatoakDR, line 170: Caveat, if you're wearing a black shadowsilk cloak then `LOOK`ing at your character won't see anything so it won't be able know if you're wearing one or not. The only alternative I'm aware of if this use case needs to be supported would be to brute force try to hold each possibility:
- hold moonblade
- hold moonstaff
- hold moonkatar
- ...

PR #4471, opened 2020-08-30
[script] [common-summoning] Update logic to hold moon weapons

KatoakDR:
* Changes ambiguous use of `get moon` with `DRCMM.hold_moon_weapon?` (depends on https://github.com/rpherbig/dr-scripts/pull/4470)
* Changes ambiguous use of `glance moon` with checking what's held in your hands

Bulk of the logic is to ensure the regex patterns match to actual moon weapons and not get confused by items whose nouns are or start with the word "moon", like "a trio of moons" or "moonsilk shirt". I'm personally invested in these changes because my moon mage wears both those items, haha.

PR #4472, opened 2020-08-30
Revert "[script] [common-summoning] Update logic to hold moon weapons"

rpherbig:
Reverts rpherbig/dr-scripts#4471


rpherbig:
Reverting because I missed that this required a new `common-moonmage` function

PR #4474, opened 2020-08-30
Add support for climbing target

nro2:
These changes allow 'climbing_target' to work with the climbing targets (xalas, undergondola, wyverns, cliffs) instead of just with practice.  Tested by running athletics with no arguments, as well as with CT.

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -37,13 +37,13 @@ class Athletics
     args = parse_args(arg_definitions)
 
     start_script('skill-recorder') unless Script.running?('skill-recorder')
-    if args.wyvern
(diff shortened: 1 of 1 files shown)


Hiinky:
Good good to me!  In most cases I think people just run a before/after/t2 for athletics to use the args but this could be great for folks using tm or who would prefer to use a  yaml setting.

nro2:
I agree with you on that.  I personally use a before/after for my athletics, but someone recently asked me if there was a way for CT to not only use hometown in determining where it climbed.  This is handy for people like him that want to have their hometown set somewhere like Crossing but are tired of getting killed climbing in xalas.

PR #4482, opened 2020-09-06
[script] [attunement] Loop until done training

KatoakDR:
* Rather than script stop once exhausted all rooms, repeats itself until at sufficient learning rate or run out of time
* Fixes issue where stationary training or lunar mages (moonies, traders) would only perform once perceive because only in one room
* Refactors code into their own methods

### Test Cases

I've used this to successfully lock Attunement skill on my Moon Mage (stands still) and Paladin (power walks in Crossing)

### DIFF
--- a/attunement.lic
+++ b/attunement.lic
@@ -52,11 +52,4 @@ class Attunement
 
 end
 
-before_dying do
-  # release any mana if preparing a spell
-  fput("release")
-  # put away any cambrinth items
-  EquipmentManager.new.empty_hands
-end
-
 Attunement.new
--- a/attunement.lic
+++ b/attunement.lic
@@ -14,30 +14,49 @@ class Attunement
     @stationary_skills_only = settings.crossing_training_stationary_skills_only
     @hometown = settings.hometown
     @attunement_rooms = settings.attunement_rooms
-    cast_spells(settings.waggle_sets['attunement'], settings)
-    train_attunement
+    @start_timer = Time.now
+    train_attunement(settings)
   end
 
-  def train_attunement
-    command = DRStats.moon_mage? || DRStats.trader? ? 'perc mana' : 'perc'
+  def get_perceive_command
+    DRStats.moon_mage? || DRStats.trader? ? 'perceive mana' : 'perceive'
+  end
 
-    room_list = if @stationary_skills_only || DRStats.moon_mage? || DRStats.trader?
-                  [Room.current.id]
-                elsif !@attunement_rooms.empty?
-                  @attunement_rooms
-                else
-                  get_data('town')[@hometown]['attunement_rooms']
-                end
-
-    start_timer = Time.now
-    room_list.each do |room_id|
-      walk_to(room_id)
-      bput(command, 'You reach out')
-      waitrt?
-      break if DRSkill.getxp('Attunement') >= 30 || start_timer - Time.now > 90
+  def get_rooms
+    if @stationary_skills_only || DRStats.moon_mage? || DRStats.trader?
+      [Room.current.id]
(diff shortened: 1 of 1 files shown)


attunement.lic
```diff
@@ -14,30 +14,42 @@ class Attunement
...
     @hometown = settings.hometown
     @attunement_rooms = settings.attunement_rooms
-    cast_spells(settings.waggle_sets['attunement'], settings)
-    train_attunement
+    @start_timer = Time.now
+    train_attunement(settings)
   end
 
-  def train_attunement
-    command = DRStats.moon_mage? || DRStats.trader? ? 'perc mana' : 'perc'
```
KatoakDR, line 22: Moved to new method `get_perceive_command`

attunement.lic
```diff
@@ -14,30 +14,42 @@ class Attunement
...
-
-    room_list = if @stationary_skills_only || DRStats.moon_mage? || DRStats.trader?
-                  [Room.current.id]
-                elsif !@attunement_rooms.empty?
-                  @attunement_rooms
-                else
-                  get_data('town')[@hometown]['attunement_rooms']
-                end
```
KatoakDR, line 30: Moved to new method `get_rooms`

attunement.lic
```diff
@@ -14,30 +14,42 @@ class Attunement
...
-
-    start_timer = Time.now
-    room_list.each do |room_id|
-      walk_to(room_id)
-      bput(command, 'You reach out')
-      waitrt?
-      break if DRSkill.getxp('Attunement') >= 30 || start_timer - Time.now > 90
```
KatoakDR, line 37: Replaced by new lines 42-49 to keep looping until training is done

rpherbig:
If I'm reading this right, it'll run through all rooms if the xp or timer thresholds aren't met. For example, if you're at 89.999 seconds at the end of room list, it'll run through all rooms again even though it would cross the threshold at the next room.

Why isn't that check done before moving to each room?

KatoakDR:
Line 48, after perceiving in a room, will exit the loop if you’re done training (learning rate or timer). 

I think that handles your concern?

There are two loops:

1. Outer loop that starts buffs and starts the rooms loop
2. Inner loop to walk each room and perceive

Each loop checks if we’re done training at each iteration to avoid the scenario you bring up: don’t walk all the rooms again needlessly. 

Please advise if you still see an issue, thanks!

rpherbig:
Ah, I missed that. Yep, that works. Looks good to me. ✔️

KatoakDR:
Thanks!

BinuDR (commented):
There is an error in this line as noted.  This script also causes moon mages and traders to just perceive mana repeatedly which, when combined with the previous error, causes the script to take a very long time to finish.

The script needs to handle lunar mages differently.  The previous behavior of perceiving once and then exiting was decent.  It could be improved with the various options for perceiving that are available (planets, the moons, spellbooks, etc).  Implementing more perceive methods for them can be a project for later though.

attunement.lic
```diff
@@ -14,30 +14,42 @@ class Attunement
...
+      [Room.current.id]
+    elsif !@attunement_rooms.empty?
+      @attunement_rooms
+    else
+      get_data('town')[@hometown]['attunement_rooms']
+    end
+  end
+
+  def done_training?
+    DRSkill.getxp('Attunement') >= 30 || @start_timer - Time.now >= 90
```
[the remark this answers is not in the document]
KatoakDR: Good catch on the timer bug! I didn’t notice it when I was making my changes. That was the original logic so didn’t think to modify it.

PR #4483, opened 2020-09-06
[script] [common-items] More options to check which hand holds an item

KatoakDR:
### New Features
* Is item in the left hand?
* Is item in the right hand?
* Is item in either hand?
* Does each hand hold one of the items?

These new methods supports both `String` or `DRC::Item` as the 'item' argument. Use of the `DRC::Item` class standardizes on regex used in `DRC::Item`'s `short_regex` method rather than duplicate that logic here and there. This also ensures this method picks up the updated regex if #4484 gets merged, or if the regex ever changes in the future.

### Use Cases

One use case is when deciding if to swap a weapon between hands to train offhand weapon skill. If the weapon is already in your left hand, you don't need to swap.

Or, perhaps an action requires the item in a particular hand, you can check if it's already there or not with these new utility functions.

### Test Cases

Hold a **vault book** in your left hand.

Hold a **bank book** in your right hand.

#### Assert can find 'vault book' in left hand or in either hand
```
> glance

You glance down to see a leather bank book in your right hand and a vault book in your left hand.

,e echo !!DRCI.in_hand?('vault book', 'left')

[exec1: true]

,e echo !!DRCI.in_hand?('vault book', 'right')

[exec1: false]

,e echo !!DRCI.in_hand?('vault book', 'either')

[exec1: true]

,e echo !!DRCI.in_hand?('vault book', 'both')

[exec1: false]

,e echo !!DRCI.in_hand?(DRC::Item.new(adjective:'vault', name:'book'), 'either')

[exec1: true]
```

#### Assert can find any 'book' in any hand.
```
> glance

You glance down to see a leather bank book in your right hand and a vault book in your left hand.

,e echo !!DRCI.in_hand?('book', 'left')

[exec1: true]

,e echo !!DRCI.in_hand?('book', 'right')

[exec1: true]

,e echo !!DRCI.in_hand?('book', 'either')

[exec1: true]

,e echo !!DRCI.in_hand?('book', 'both')

[exec1: true]

,e echo !!DRCI.in_hand?(DRC::Item.new(name:'book'), 'either')

[exec1: true]
```

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -137,7 +137,7 @@ module DRCI
   # Checks if the item is in one or more hands.
   # Hand options are: left, right, either, both.
   def in_hand?(item, whichHand = 'either')
-    item_regex = item.split.size > 1 ? /.*\b#{item.split.first}.*\b#{item.split.last}/i : /\b#{item}/i
+    item_regex = item.split.size > 1 ? /.*\b#{item.split.first}.*\b#{item.split.last}\b/i : /\b#{item}\b/i
     item_found = false
     case whichHand.downcase
     when 'left'
--- a/common-items.lic
+++ b/common-items.lic
@@ -119,17 +119,14 @@ module DRCI
     result =~ /inside/
   end
 
-  # Is the item in either hand?
   def in_hands?(item)
     in_hand?(item, 'either')
   end
 
-  # Is the item in your left hand?
   def in_left_hand?(item)
     in_hand?(item, 'left')
   end
 
-  # Is the item in your right hand?
   def in_right_hand?(item)
     in_hand?(item, 'right')
   end
@@ -141,16 +138,12 @@ module DRCI
     item_found = false
     case which_hand.downcase
     when 'left'
-      # Your holding the item in your left hand?
       item_found = ("#{DRC.left_hand}" =~ item_regex) != nil
     when 'right'
-      # Your holding the item in your right hand?
       item_found = ("#{DRC.right_hand}" =~ item_regex) != nil
     when 'either'
-      # At least one hand is holding the item?
       item_found = in_hand?(item, 'left') || in_hand?(item, 'right')
     when 'both'
-      # You're holding two of the same item, one in each hand?
       item_found = in_hand?(item, 'left') && in_hand?(item, 'right')
     else
       DRC.message("Unknown hand: #{which_hand}. Valid options are: left, right, either, both")
--- a/common-items.lic
+++ b/common-items.lic
@@ -141,9 +141,9 @@ module DRCI
     when 'right'
       DRC.right_hand =~ item_regex
     when 'either'
-      in_hand?(item, 'left') || in_hand?(item, 'right')
+      in_left_hand?(item) || in_right_hand?(item)
     when 'both'
-      in_hand?(item, 'left') && in_hand?(item, 'right')
+      in_left_hand?(item) && in_right_hand?(item)
     else
       DRC.message("Unknown hand: #{which_hand}. Valid options are: left, right, either, both")
       false
--- a/common-items.lic
+++ b/common-items.lic
@@ -136,10 +136,10 @@ module DRCI
 
   # Checks if the item is in one or more hands.
   # Hand options are: left, right, either, both.
-  def in_hand?(item, whichHand = 'either')
+  def in_hand?(item, which_hand = 'either')
     item_regex = item.split.size > 1 ? /#{item.split.first}.*\b#{item.split.last}/i : /\b#{item}/i
     item_found = false
-    case whichHand.downcase
+    case which_hand.downcase
     when 'left'
       # Your holding the item in your left hand?
       item_found = ("#{DRC.left_hand}" =~ item_regex) != nil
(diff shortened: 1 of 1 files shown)


common-items.lic
```diff
@@ -119,9 +119,44 @@ module DRCI
...
+  end
+
+  # Is the item in your right hand?
+  def in_right_hand?(item)
+    in_hand?(item, 'right')
+  end
+
+  # Checks if the item is in one or more hands.
+  # Hand options are: left, right, either, both.
+  def in_hand?(item, whichHand = 'either')
```
rcuhljr, line 139: Just some idiomatic ruby corrections. `whichHand` -> `which_hand`

common-items.lic
```diff
@@ -119,9 +119,44 @@ module DRCI
...
+      # At least one hand is holding the item?
+      item_found = in_hand?(item, 'left') || in_hand?(item, 'right')
+    when 'both'
+      # You're holding two of the same item, one in each hand?
+      item_found = in_hand?(item, 'left') && in_hand?(item, 'right')
+    else
+      DRC.message("Unknown hand: #{whichHand}. Valid options are: left, right, either, both")
+      item_found = false
+    end
+    return item_found
```
rcuhljr, line 159: You can get rid of item_found entirely and returns, ruby methods return the value of the last evaluated statement by default. e.g.

```
irb(main):001:0> def test(item)
irb(main):002:1>   case item
irb(main):003:2>   when 'bob'
irb(main):004:2>     true
irb(main):005:2>   when 'tim'
irb(main):006:2>     'tod'
irb(main):007:2>   else
irb(main):008:2>     false
irb(main):009:2>   end
irb(main):010:1> end
=> :test
irb(main):011:0> test('bob')
=> true
irb(main):012:0> test('tim')
=> "tod"
irb(main):013:0> test('joe')
=> false
```

common-items.lic
```diff
@@ -119,9 +119,44 @@ module DRCI
...
+  end
+
+  # Checks if the item is in one or more hands.
+  # Hand options are: left, right, either, both.
+  def in_hand?(item, whichHand = 'either')
+    item_regex = item.split.size > 1 ? /#{item.split.first}.*\b#{item.split.last}/i : /\b#{item}/i
+    item_found = false
+    case whichHand.downcase
+    when 'left'
+      # Your holding the item in your left hand?
```
rcuhljr, line 144: I think these individual case comments are probably overkill, the code is very clear about what it's doing here.

common-items.lic
```diff
@@ -119,9 +119,44 @@ module DRCI
...
+      item_found = ("#{DRC.left_hand}" =~ item_regex) != nil
```
rcuhljr, line 145: I don't feel strongly in favor of ruby lang convention of predicate methods returning true/false instead of truthy/falsey values, but if you want to conform to it the double bang approach is clean and idiomatic. Also I think DRC.left_hand is already returning a string so the interpolation `"#{}` is no longer needed since you aren't combining two strings together anymore.

`item_found = ("#{DRC.left_hand}" =~ item_regex) != nil`
combined with the other suggestion becomes
`!!(DRC.left_hand =~ item_regex)`

Which gets back to my minor preference for no strict true/false adherence because then you just get away with
`DRC.left_hand =~ item_regex`

rcuhljr (changes requested):
Just some minor idiomatic ruby stuffs.

common-items.lic
```diff
@@ -119,9 +119,44 @@ module DRCI
...
+    when 'right'
+      # Your holding the item in your right hand?
+      item_found = ("#{DRC.right_hand}" =~ item_regex) != nil
+    when 'either'
+      # At least one hand is holding the item?
+      item_found = in_hand?(item, 'left') || in_hand?(item, 'right')
```
rcuhljr, line 151: Oh, worlds most trivial comment I forgot. If you used your helper methods you made and later change the expected value of 'right' you'll have a few less places to change the hard coded string.

`item_found = in_right_hand?(item) || in_left_hand?(item)`

KatoakDR:
Thanks @rcuhljr for the feedback and idiomatic suggestions. It's a bit painfully obvious Ruby is not my primary programming language 😁 

I'll work on getting updates here this week.

KatoakDR:
@rpherbig I noticed a possible bug in a test that's run to validate the commits. The test is unrelated to this PR but I noticed it on [this check for a commit in this PR](https://github.com/rpherbig/dr-scripts/pull/4483/checks?sha=ce48b26522946d0c3d686df917dcec537901ccb0).

```
TestCommon#test_bput_returns_error_code_after_timeout [/home/travis/build/rpherbig/dr-scripts/test/test_bput.rb:53]:
Expected |15 - 16.12816047668457| (1.1281604766845703) to be <= 1.0.
```

Looks like a randomly generated number was too big?

rcuhljr:
Not a random number, the machine running the tests was probably just running particularly slow that day. It's a test that waits to see if bput times out properly after 15 seconds, it's not normally off by more than a second, but since we're using a helper library to accelerate time so the test doesn't actually take 15 seconds to run small delays can be somewhat magnified.

rcuhljr (approved):

PR #4484, opened 2020-09-07
[script] [common] short_regex to match on full words from config

KatoakDR:
### Problem
Without the leading `\b` in the regex, then words that end with the adjective match which may not be accurate. For example, "oak spear" is in your config but "redoak spear" would also match although it's a wrong match.

Without the trailing `\b` in the regex, then words that begin with the name match which may not be accurate. For example, "leather book" is in your config but "leather booklet" would also match although it's a wrong match.

### Change
Ensure `short_regex` uses `\b` to match on the exact word and not match on words that "start with" or "end with" the adjective/noun from the gear config.

### Test Cases

**Original regex without word boundaries (false positives)**
```ruby
> ,e echo /leather.*\bbook/i =~ 'redleather booklet' ? 'yes match' : 'no match'

[exec1: yes match]
```

**Updated regex (no false positives)**
```ruby
> ,e echo /\bleather\b.*\bbook\b/i =~ 'redleather booklet' ? 'yes match' : 'no match'

[exec1: no match]
```

**Regression Test, both regexes match the item in the happy path**
```ruby
,e echo /leather.*\bbook/i =~ 'leather book' ? 'yes match' : 'no match'

[exec1: yes match]

> ,e echo /\bleather\b.*\bbook\b/i =~ 'leather book' ? 'yes match' : 'no match'

[exec1: yes match]
```


KatoakDR:
Closing per feedback from @Hiinky in Discord at https://discordapp.com/channels/745675889622384681/745696628714897508/752411675751219322

This change could be a "breaking change" for players whose YAML was loosely configured on item names rather than the exact adjective and noun the game uses for an item in your hand.

For example, the current regular expression would match "oak-handled pilum" for `adjective: oak, noun: pil` however the proposed regular expression would require the config to be `adjective: oak-handled, noun: pilum`.

Although players _should_ have used the full words the game uses, up to this point Lich hasn't _required_ them too.

Moreover, since the game allows partial adjective/name matching so players don't have to spell the words out completely, they may have been accustomed to that and used shorthand adjective/noun in their YAML assuming it's correct because the game itself never had issues with that syntax.

Sadly, I'm withdrawing this change as the risk outweighs the gains.

Thanks for your consideration

KatoakDR:
Per more feedback from the group on this change in Discord at  https://discordapp.com/channels/745675889622384681/745696628714897508/752520463082717244, I've re-opened this as a draft.

The consensus is that if we make the regex match more strict then we need to give players ample time to be aware of the coming change, how it may impact them, and what changes they need to make to their YAML if we detect their current config would fail to match with the new logic.

I will submit a new pull request to alert players of this upcoming change and how to fix their YAMLs. Once the pre-PR is merged then it should be live for perhaps a month to give players ample time to make adjustments then bring this PR back up for review.

Thanks

jandersson:
First off, really nice merge description! 

Second, I understand the change, but I don't really understand what's being fixed? Could you motivate that a bit more? Why is it bad that I write the shortened name of an item?

KatoakDR:
Feedback from [Gorzon in Discord](https://discordapp.com/channels/745675889622384681/745696628714897508/752634021896519791):
> I have enough trouble making my yaml work the way I have it.. I'll probably just keep moving farther from mainline and doing weekly merges of things that fit.
in other words - if you have the problem with not using adjectives, then use adjectives.  This seems (to me) like a solution seeking a problem - and a lot more trouble than it's worth.

**Katoak's response:**

Thanks for your feedback, Gorzon. This change won't require you to use adjectives. It requires that the adjective, if provided, matches the adjective spelling as the game knows the item to be. Same with the name.

For example, if the game knows your weapon as "oak-hafted staff" the adjective is "oak-hafted" and the noun is "staff".

Your YAML can include (or not) the adjective part in your config. But if you do specify the adjective then it must be "oak-hafted" and not simply "oak". Same with the name, it must be exactly "staff".

Other scripts like burgle and scripts that dispose items to trash also have very strict item name matches. This will be bringing the Item class in line with those player protections.

If it's not there already, I will add to validate script a check that taps all the items in your gear config to make sure the adjective/noun are correct.

jandersson:
> Other scripts like burgle and scripts that dispose items to trash also have very strict item name matches. This will be bringing the Item class in line with those player protections.

This seems like a key point. i.e. not trashing items by mistake.

rpherbig:
> I will submit a new pull request to alert players of this upcoming change and how to fix their YAMLs. Once the pre-PR is merged then it should be live for perhaps a month to give players ample time to make adjustments then bring this PR back up for review.

> If it's not there already, I will add to validate script a check that taps all the items in your gear config to make sure the adjective/noun are correct.

I'm on board with this approach

rpherbig:
@KatoakDR what's the status of this? I see 65 commits, it's still a draft, but the actual diff appears to be limited to one line (and tests).

KatoakDR:
> @KatoakDR what's the status of this? I see 65 commits, it's still a draft, but the actual diff appears to be limited to one line (and tests).

Still on my todo list. The flood of commits was my attempt to back merge `master` into this branch to keep it current-ish. I was updating my fork and cleaning up branches. Sorry for the noise.

Hiinky:
Just circling back to this, if we're really going to commit this, we need at least a week or two of spam warnings and updates to people.  The breakage potential is very high.  While I do agree more specific matches are better in the long run, I also don't recall ever seeing this particular match/regex cause issues.  

```yaml
gear:
- :name: red
  :adjective: bola
```
`You tap a redoak bola.`
```
>;e echo /\bredl\b.*\bbola\b/i =~ 'redoak bola' ? 'yes match' : 'no match'
--- Lich: exec1 active.
[exec1: no match]
```

KatoakDR:
I've submitted https://github.com/rpherbig/dr-scripts/pull/4581 as the precursor to this PR. It introduces the "early warning system" to let player's know if their `gear:` config doesn't specify the exact adjective/noun of their gear.

KatoakDR:
Closing due to my own lack of activity on this topic. I'll re-open this if I loop back to it.

PR #4486, opened 2020-09-11
[script] [attunement] Feedback from Binu (#4485)

### LINKED ISSUE #4485: Attunement script errors
The recent change to this script has introduced some problems as noted in my review of the closed PR.

https://github.com/rpherbig/dr-scripts/pull/4482

KatoakDR:
### Changes
* Fixes #4485, feedback from @BinuDR
* Lunar mages will alternate between perceiving mana, moons, and planets
* Removed timer logic since it was wasn't working to begin with and replaced with new setting `attunement_target_increment` for how many mind states you want to increase by before script exits
* Made exp learning rate upper threshold configurable by looking at setting `crossing_training_max_threshold` (existing setting)
* Added validations for the settings
* Added `attunement_target_increment` to `base.yaml`

Inspired by the [outdoorsmanship](https://github.com/rpherbig/dr-scripts/blob/master/outdoorsmanship.lic#L15) script, the new setting `attunement_target_increment` lets players limit how long the script runs by specifying the number of mind states they want to increase by. A low number will run the script for less time than a high number.

### Examples

**Example 1**
```yaml
crossing_training_max_threshold: 30
attunement_target_increment: 6
```
Player's attunement is at 20. The player runs attunement script and it checks if skill(20) >= threshold(30). It's not so we proceed to next step to determine how long to run the script. The threshold(30) - skill(20) is 10 mind states to go. However, the target increment is 6 so once the skill(20) has increased to at least 26 then the script will stop.

**Example 2**
```yaml
crossing_training_max_threshold: 30
attunement_target_increment: 6
```
Player's attunement is at 28. The player runs attunement script and it checks if skill(28) >= threshold(30). It's not so we proceed to next step to determine how long to run the script. The threshold(30) - skill(28) is 2 mind states to go. Even though the target increment is 6 the script will stop once the skill is at 30 (28+2), not 34 (28+6).

**Example 3**
```yaml
crossing_training_max_threshold: 30
attunement_target_increment: 6
```
Player's attunement is at 34. The player runs attunement script and it checks if skill(34) >= threshold(30). It is so the script stops because the player has met or exceeded the max threshold.

### Summary
If players want the previous behavior for lunar mages to do one perceive then script exits then they only need set `attunement_target_increment` to a sufficiently low value.

If players want to ensure their lunar mage keeps training until attain a certain mind state then they only need set `attunement_target_increment` to a sufficiently high value.

For both players, ensure `crossing_training_max_threshold` is set to a desired threshold.

### DIFF
--- a/attunement.lic
+++ b/attunement.lic
@@ -19,7 +19,6 @@ class Attunement
     @exp_timer = settings.exp_timers['Attunement'] || 90
     @exp_max_threshold = settings.crossing_training_max_threshold
     @lunar_perceive_index = -1
-    @start_timer = Time.now
     train_attunement(settings)
   end
 
@@ -66,8 +65,13 @@ class Attunement
   end
 
   def train_attunement(settings)
+    # Buff first then start timer so that
+    # cast time doesn't count against you
+    DRCA.do_buffs(settings, 'attunement')
+    @start_timer = Time.now
     room_list = get_rooms
     until done_training?
+      # If we keep looping, ensure buffs stay active
       DRCA.do_buffs(settings, 'attunement')
       room_list.each do |room_id|
         walk_to(room_id)
--- a/attunement.lic
+++ b/attunement.lic
@@ -16,10 +16,20 @@ class Attunement
     @stationary_skills_only = settings.crossing_training_stationary_skills_only
     @hometown = settings.hometown
     @attunement_rooms = settings.attunement_rooms
-    @exp_timer = settings.exp_timers['Attunement'] || 90
     @exp_max_threshold = settings.crossing_training_max_threshold
+    @exp_target_increment = settings.attunement_target_increment
+    @exp_start_mindstate = DRSkill.getxp('Attunement')
+    @exp_stop_mindstate = get_target_mindstate(@exp_start_mindstate, @exp_target_increment, @exp_max_threshold)
     @lunar_perceive_index = -1
     train_attunement(settings)
+  end
+
+  # Determines the target mind state to train to
+  # based on the starting value of `mindstate` and
+  # a target goal of increasing by `increment` without
+  # exceeding a max mind state of `threshold`.
+  def get_target_mindstate(mindstate, increment, threshold)
+    mindstate + [increment, [threshold - mindstate, 0].max].min
   end
 
   def is_lunar_mage?
@@ -48,23 +58,11 @@ class Attunement
     end
   end
 
-  def reached_exp_learn_threshold?
-    DRSkill.getxp('Attunement') >= @exp_max_threshold
-  end
-
-  def reached_exp_timer_threshold?
-    Time.now - @start_timer >= @exp_timer
-  end
-
   def done_training?
-    reached_exp_learn_threshold? || reached_exp_timer_threshold?
+    DRSkill.getxp('Attunement') >= @exp_stop_mindstate
   end
 
   def train_attunement(settings)
-    # Buff first then start timer so that
-    # cast time doesn't count against you
-    DRCA.do_buffs(settings, 'attunement')
-    @start_timer = Time.now
     room_list = get_rooms
     until done_training?
       # If we keep looping, ensure buffs stay active
--- a/attunement.lic
+++ b/attunement.lic
@@ -9,21 +9,42 @@ class Attunement
   include DRCA
   include DRCT
 
+  LUNAR_PERCEIVE_COMMANDS = ['mana', 'moons', 'planets']
+
   def initialize
     settings = get_settings
     @stationary_skills_only = settings.crossing_training_stationary_skills_only
     @hometown = settings.hometown
     @attunement_rooms = settings.attunement_rooms
+    @exp_timer = settings.exp_timers['Attunement'] || 90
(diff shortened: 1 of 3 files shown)


attunement.lic
```diff
@@ -9,21 +9,41 @@ class Attunement
...
+    @lunar_perceive_index = -1
     train_attunement(settings)
   end
 
+  def is_lunar_mage?
+    DRStats.moon_mage? || DRStats.trader?
+  end
+
+  def get_next_lunar_perceive_index
+    @lunar_perceive_index = @lunar_perceive_index + 1
```
rpherbig, line 30: The modulus operator (`%`) could simplify this logic

```
@commands = ['a', 'b', 'c']
@index = 0

def increment_index()
  @index = (@index + 1) % @commands.length
end

puts @index
increment_index()
puts @index
increment_index()
puts @index
increment_index()
puts @index
increment_index()
puts @index
increment_index()
puts @index
increment_index()
puts @index
increment_index()
```

Gives:
```
> ruby main.rb
0
1
2
0
1
2
0
```

attunement.lic
```diff
@@ -9,21 +9,41 @@ class Attunement
...
+  LUNAR_PERCEIVE_COMMANDS = ['mana', 'moons', 'planets']
+
   def initialize
     settings = get_settings
     @stationary_skills_only = settings.crossing_training_stationary_skills_only
     @hometown = settings.hometown
     @attunement_rooms = settings.attunement_rooms
-    @start_timer = Time.now
+    @exp_timer = settings.exp_timers['Attunement'] || 90
+    @exp_max_threshold = settings.crossing_training_max_threshold
```
rpherbig, line 20: I'm not sure how I feel about this... this setting has been specific to the `crossing-training` script up until now...

attunement.lic
```diff
@@ -9,21 +9,41 @@ class Attunement
...
 
+  LUNAR_PERCEIVE_COMMANDS = ['mana', 'moons', 'planets']
+
   def initialize
     settings = get_settings
     @stationary_skills_only = settings.crossing_training_stationary_skills_only
     @hometown = settings.hometown
     @attunement_rooms = settings.attunement_rooms
-    @start_timer = Time.now
+    @exp_timer = settings.exp_timers['Attunement'] || 90
```
rpherbig, line 19: This could make a good validator, thoughts?

attunement.lic
KatoakDR: Sweet, thanks!

attunement.lic
rpherbig: I'm open to either. Can you (quickly) poll the user base and see if there's a strong opinion either way?

Hiinky: Without me digging through the whole works, if someone wants to give me a question or two to ask, i can put up a survey

KatoakDR: Thanks @Hiinky! 

We're wanting to know if people want to use `exp_timers` to limit how long this script will run (like a time upper limit) and `crossing_training_max_threshold` as the max learning rate limit.

Those are two existing settings used by crossing-trainer script. That script calls this attunement script. Does it make sense to re-use those crossing-training settings here or should we introduce new settings?

attunement.lic
KatoakDR: Yes, I'll hold off on writing the validator until we come to consensus on which settings will be used

attunement.lic
KatoakDR: @rpherbig @Hiinky @BinuDR 

To simplify things, I'm going with Option 2 unless ya'll think otherwise. I'll introduce `attunement_target_increment` setting and if set means how many mind states you want the script to increase your Attunement skill before stopping. Players can set a small value to run the script for short time, or a large value to ensure you get to a high mind state.

`crossing_training_max_threshold` used by crossing-training script is still an upper threshold. The training script won't call any of the sub-scripts like attunement if your mind state is at or above that threshold.

`exp_timers` will stay exclusive to crossing-training script. Rather than use a timer, the attunement script will rely on `crossing_training_max_threshold` and/or `attunement_target_increment`.

### Examples

**Example 1**
```yaml
crossing_training_max_threshold: 30
attunement_target_increment: 6
```
Player's attunement is at 20. The player runs attunement script and it checks if skill(20) >= threshold(30). It's not so we proceed to next step to determine how long to run the script. The threshold(30) - skill(20) is 10 mind states to go. However, the target increment is 6 so once the skill(20) has increased to at least 26 then the script will stop.

**Example 2**
```yaml
crossing_training_max_threshold: 30
attunement_target_increment: 6
```
Player's attunement is at 28. The player runs attunement script and it checks if skill(28) >= threshold(30). It's not so we proceed to next step to determine how long to run the script. The threshold(30) - skill(28) is 2 mind states to go. Even though the target increment is 6 the script will stop once the skill is at 30 (28+2), not 34 (28+6).

attunement.lic
KatoakDR: Validations updated per https://github.com/rpherbig/dr-scripts/pull/4486/commits/0d711f7579f8d0c5a9efcacb9a9f7cca8b36e8e5

PR #4488, opened 2020-09-13
[script] [faux-atmo] New script to put those verby items to work ✨

KatoakDR:
### Summary
_Bring your verby items to life!_

Some items perform [atmospheric](https://elanthipedia.play.net/Atmospheric_item) actions on their own where they emit some messaging to the game window periodically. For example a [robe strewn with twinkling points of starlight](https://elanthipedia.play.net/Item:Black_robe_strewn_with_twinkling_points_of_starlight) periodically emits "The points of starlight on your robe float languidly across the fabric before suddenly beginning to swirl and flicker in wild, hypnotic patterns."

Some items have verbs with fun messaging but the player must actively perform the verb for the messaging to appear, such as a [blue-scaled basilisk with firestained uthamar and ice sapphire eyes](https://elanthipedia.play.net/Item:Blue-scaled_basilisk_with_firestained_uthamar_and_ice_sapphire_eyes). These verby items are classically not atmo items. However, one can cause them to behave like an atmo item if verbs are performed on them periodically. For example, to `pet|rub|touch|push` a worn toy or `wave|rub|pat` a worn cloak.

This new `faux-atmo` script lets players configure which verbs they want performed on which items every N minutes.

### How it works
* The player provides a list of items via `faux_atmo_items`. Each entry in the list has a `name` (string) and a `verbs` (list) properties.
* Every N minutes, the script round-robin chooses one item and one verb and performs the action.

### Conditional execution
Like the `almanac` and `wand-watcher` scripts, players can customize when the script runs.
* Players can provide a `faux_atmo_no_use_scripts` to not perform an action if those scripts are running.
* Players can provide a `faux_atmo_no_use_rooms` to not perform an action if in one of those rooms.
* The script takes no action if the player is hidden or invisible.

### Example config
```yaml
# Number of minutes to wait between performing a verb on an item.
faux_atmo_interval: 15

# Items that don't have an inherent atmo effect
# but whose verbs if done periodically would.
faux_atmo_items:
  - name: blue basilisk
    verbs:
    - pet
    - pinch
    - poke
  - name: red wyvern
    verbs:
    - touch
    - push
    - tilt

# Don't perform verbs on items if these scripts are running.
faux_atmo_no_use_scripts:
- combat-trainer
- go2
- get2
- bescort
- steal
- burgle

# Don't perform verbs on items if in these rooms.
faux_atmo_no_use_rooms:
- Carousel Chamber     # Room with vault
- Carousel Booth       # Room just before vault
[...]

PR #4492, opened 2020-09-15
[script] [faux-atmo] Randomize item/verb selection

KatoakDR:
### Changes
* Per [feedback](https://github.com/rpherbig/dr-scripts/pull/4488#pullrequestreview-488143353), uncommented `faux_atmo_items` in `base.yaml`
* Per [feedback](https://github.com/rpherbig/dr-scripts/pull/4488#pullrequestreview-488144288), support random or round-robin selection of items/verbs via `faux_atmo_random`

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -1452,6 +1452,10 @@ tarantula_debug: false
 # Number of minutes to wait between performing a verb on an item.
 faux_atmo_interval: 15
 
+# Set to true to randomly select an item and verb to perform.
+# Set to false to round-robin select items and verbs to perform.
+faux_atmo_random: true
+
 # Items that don't have an inherent atmo effect
 # but whose verbs if done periodically would.
 # faux_atmo_items:
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -1458,7 +1458,7 @@ faux_atmo_random: true
 
 # Items that don't have an inherent atmo effect
 # but whose verbs if done periodically would.
-faux_atmo_items: []
+faux_atmo_items:
 #   - name: blue basilisk
 #     verbs:
 #     - pet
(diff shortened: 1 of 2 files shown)


profiles/base.yaml
```diff
@@ -1452,9 +1452,13 @@ tarantula_debug: false
...
 faux_atmo_interval: 15
 
+# Set to true to randomly select an item and verb to perform.
+# Set to false to round-robin select items and verbs to perform.
+faux_atmo_random: true
+
 # Items that don't have an inherent atmo effect
 # but whose verbs if done periodically would.
-# faux_atmo_items:
+faux_atmo_items: []
```
rpherbig, line 1461: I would not initialize this to an empty array. `base-empty` will take care of that. You can just initialize it to nil (e.g. `faux_atmo_items:`) to make the base case simpler for new users.

profiles/base.yaml
```diff
@@ -1452,9 +1452,13 @@ tarantula_debug: false
 # Number of minutes to wait between performing a verb on an item.
 faux_atmo_interval: 15
 
+# Set to true to randomly select an item and verb to perform.
```
rpherbig, line 1455: To clarify what I was suggesting: I meant to randomize the starting item, then round robin.

The current implementation always picks randomly. Mine is a weak preference, so let me know if you want to keep this (and I'll merge the PR) or change it.

profiles/base.yaml
KatoakDR: Ah, gotcha. I'll try that out

profiles/base.yaml
KatoakDR: What do you think of if `@random = true` then during init the items and their verbs are shuffled? From then on it's round-robin to avoid the nature of random-every-time having chance of you performing the same verb/item too soon or ignoring items.

profiles/base.yaml
KatoakDR: Initialized `faux_atmo_items:` to nil in https://github.com/rpherbig/dr-scripts/pull/4492/commits/6761f49278069403637aac8fa007b9f63f65f2ae

PR #4496, opened 2020-09-18
Add active mitigation for new critter mechanic

Hiinky:
New mechanic added with bronze leucro.  The release post made it sound like this may be a theme going forward so going forward we may need to revisit if their strings don't match but I'm hoping if that day comes it is a ways out.

```
You believe you could dip out of the way of the leucro, if you act quickly enough.
You believe you could jump out of the way of the leucro, if you act quickly enough.
You believe you could duck out of the way of the leucro, if you act quickly enough.
You believe you could lean out of the way of the leucro, if you act quickly enough.
You believe you could dive out of the way of the leucro, if you act quickly enough.
```

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -910,7 +910,7 @@ class SafetyProcess
 
   def active_mitigation
     return unless Flags['active-mitigation']
-    bput("#{Flags['active-mitigation'][1]} #{Flags['active-mitigation'][2]}" ,"You manage to #{Flags['active-mitigation'][1]}", "You've got to be joking",'Please rephrase that command','You jump back from',"You can't do that here")
+    bput("#{Flags['active-mitigation'][1]} #{Flags['active-mitigation'][2]}", 'You manage to', "You've got to", 'Please rephrase', 'You jump back', "You can't do")
     Flags.reset('active-mitigation')
   end
   def in_danger?(danger)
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -906,6 +908,11 @@ class SafetyProcess
     retreat
   end
 
+  def active_mitigation
+    return unless Flags['active-mitigation']
+    bput("#{Flags['active-mitigation'][1]} #{Flags['active-mitigation'][2]}" ,"You manage to #{Flags['active-mitigation'][1]}", "You've got to be joking",'Please rephrase that command','You jump back from',"You can't do that here")
```
KatoakDR, line 913: @Hiinky this logic looks good to me :)  One day one of my chars might be able to hunt them to try it out, hehe.

I only have one suggestion and it's purely cosmetic/aesthetic to break the one long line into multiples for readability. What are your thoughts on something like this:
```suggestion
    action = Flags['active-mitigation'][1]
    target = Flags['active-mitigation'][2]
    messages = [
      "You manage to #{action}",
      "You've got to be joking",
      "Please rephrase that command",
      "You jump back from",
      "You can't do that here"
    ]
    bput("#{action} #{target}", *messages)
```

rpherbig: I like this, but it won't hold up merging. @Hiinky let us know your thoughts.

rpherbig:
No additional comments ✔️

combat-trainer.lic
Hiinky: I'm of two minds.  

1) that method is much more tidy and I like it.

2) I also hesitate to add that information directly to the monster that is ;combat.  I ultimately went this route to keep it to a few lines so that if this mechanic is isolated to a single mob it's taking up minimal space.  In the event it becomes a prevalent mechanic it could very easily be updated to work more like your suggestion then pull this information from a base file such as base-hunting where we can list a great many options without editing ;combat.  

I did trim the match strings though to be less crazy looking but should still be specific enough to only match the action.

KatoakDR: Good points, thanks @Hiinky

PR #4497, opened 2020-09-18
Add toggle to stop releasing cyclics at the end of combat

Hiinky:
While most cyclics are dangerous, some are not such as necro SO mitigation, empath cylics, bard buffs, etc.  This allows a toggle to forego that release.  This is a frequently asked question and there is much grumbling over being told to recast before moving.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -30,6 +30,8 @@ class SetupProcess
     echo("  @cycle_regalia: #{@cycle_regalia}") if $debug_mode_ct
     @default_armor = settings.default_armor_type
     echo("  @default_armor: #{@default_armor}") if $debug_mode_ct
+    @release_cyclic = settings.release_combat_cyclic
+   echo("  @release_cyclic: #{@release_cyclic}") if $debug_mode_ct 
 
     @gearsets = settings.gear_sets
     @warhorn = settings.warhorn
@@ -3889,7 +3891,7 @@ class GameState
 end
 
 before_dying do
-  DRCA.release_cyclics unless get_settings.do_not_release_combat_cyclic
+  DRCA.release_cyclics unless @release_cyclic
   DRCA.shatter_regalia?
   fput('close fissure') if DRStats.warrior_mage? && DRRoom.room_objs.any? { |x| /fissure/ =~x }
   Flags.delete('using-corpse')
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3889,7 +3889,7 @@ class GameState
 end
 
 before_dying do
-  DRCA.release_cyclics
+  DRCA.release_cyclics unless get_settings.do_not_release_combat_cyclic
   DRCA.shatter_regalia?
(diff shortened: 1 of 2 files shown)


profiles/base.yaml
```diff
@@ -123,6 +123,10 @@ offensive_spells:
 # offensive_spell_cycle is a list of offensive_spells names in the order you want to cast them.  Most useful if you have multiple TM or DEBIL spells to cast
 offensive_spell_cycle:
 # list of actions to take while aiming when stealth still needs to be trained
+
+# toggle to not release cyclic spells at the end of combat.  USE WITH CAUTION as most combat cyclics will get you arrested in a hurry.
+do_not_release_combat_cyclic: false
```
KatoakDR, line 128: @Hiinky I often get confused with the double negatives with toggles/flags. Thoughts on renaming this to `release_combat_cyclic` where `true` means yes, release it after combat, and `false` means no, don't release it.

Then `DRCA.release_cyclics unless get_settings.do_not_release_combat_cyclic` would become `DRCA.release_cyclics if get_settings.release_combat_cyclic`

combat-trainer.lic
```diff
@@ -3889,7 +3889,7 @@ class GameState
 end
 
 before_dying do
-  DRCA.release_cyclics
+  DRCA.release_cyclics unless get_settings.do_not_release_combat_cyclic
```
rpherbig, line 3892: I'm not a fan of using `get_settings` in `before_dying`, especially for `combat-trainer` which people often override with sub-YAMLs. Could we initialize a global during setup?

profiles/base.yaml
rpherbig: I also like this idea

rpherbig:
No further comments ✔️

Hiinky:
Updated - should be ready.

combat-trainer.lic
```diff
@@ -3889,7 +3891,7 @@ class GameState
 end
 
 before_dying do
-  DRCA.release_cyclics
+  DRCA.release_cyclics unless @release_cyclic
```
rpherbig, line 3894: This is an instance var and used outside of a class - I wouldn't have expected this to work. Did you test to make sure this change works as expected?

jayRyan24: I think maybe the best course is to move this to the cleanup process, as a whole. This entire thing. Know what I mean? As it stands now, it's a bit wonky to do things like this without doing things like get_settings for a single setting, etc. I could see an accessor being made for it, but I am not sure how the before_dying is handled within the scripts before the instance is deleted, etc. Does that make sense?

jayRyan24: I was pointed out that this is or kill-all, crazy stuff. Dunno. Maybe one of the foundational classes needs to store cleanup variables, not really sure.

Hiinky:
After some trial and error, the original version is the most straight forward option that I know works.  I reverted to that while we discuss.  I realize its not ideal to pull settings just to kill but I will need pointers for an alternate process.

rpherbig:
Did you try using a global, rather than instance var? `$foo` rather than `@foo`, similar to how the debug flag works.

PR #4503, opened 2020-09-24
Update base-items.yaml

formayor:
added writ for "a sealed writ" also fragment, lump, shard, tear, ingot, and fist for possible metal loot.

### DIFF
--- a/data/base-items.yaml
+++ b/data/base-items.yaml
@@ -28,6 +28,13 @@ lootables:
 - package
 - permit
 - profile
+- writ
+- fragment
+- lump
+- shard
+- tear
+- ingot
+- fist
 
 box_nouns:
 - coffer


data/base-items.yaml
```diff
@@ -28,6 +28,13 @@ lootables:
 - package
 - permit
 - profile
+- writ
+- fragment
+- lump
+- shard
+- tear
+- ingot
```
Hiinky, line 36: Could we remove ingot?  I would hate to see young characters or new players getting messed up by too high encumbrance from accidentally looting multiple heavy items and not knowing to remove this.

Hiinky: Also further explanation that if someone wants to loot ingots, they add loot_additions: to cover pretty much any scenario.  Nearly every sample.yaml has examples on how to do a loot_addition or a loot_special.

formayor: Or just comment it out by default.  I didn't know what the possibilities were so I thought I would add them since the rest of the file is very informative on other loot types.

Hiinky: In this case you _should never ever ever ever edit a base.yaml_ unless you're on dev mode so commenting it won't work.  Instead of making everyone change their files to confirm to this list, it will be much more user friendly to remove the one  problematic item and let folks add them according to preference.  Most folks that want to pick up these items likely already have a loot_addition.

Personally I want to pick up as little "extra" as possible and have zero patience for fiddling with nuggets/bars/ingots because of weight and inventory space.  I don't really want to loot the rest of those either but they're rare enough that it shouldn't cause any snags and others may find them useful supposing they're not already looting them.

BinuDR: I'd suggest removing ingot, shard, and fragment.

@formayor What's this writ thing?  If it's some sort of dropped ticket type item then this is a good place to add it.
As for the other things, why are you adding them to base.yaml?
You can just add them to your own  yaml like this:
```loot_additions:
  - ingot
  - shard
  - lump
  - fragment
  - tear```

formayor: The writ is a F2P citizenship token.  The others are there in case people don't know about them.  Nuggets and bars are both #out in base-items by default.  Maybe give ingot the same treatment?

Hiinky: I don't see nuggets or bars in base but good call on the writ.  Most of the other F2P items are listed but that was clearly missing.  Thanks!

PR #4512, opened 2020-09-28
bescort - added flying broom/carpet use to faldesu and segoltha

BinuDR:
I added functionality to use flying mounts and set it up to work for both the segoltha and faldesu rivers.

You need to add flying_mount setting to your yaml.  Here's an example with one of the new flying carpets.
`flying_mount: silk carpet`

I updated the map today and it should be pushed through to everyone by tomorrow or day after.


BinuDR:
I'll be modifying this for traveling north in the segoltha soon.  I need the map changes to go through so I can make a change there too.

Instead of having it stop at the crossing river bank, I'll have it go through the river, over the bank, then to the slope and over that too so that all the athletics checks are bypassed.
19373->15886->922

BinuDR:
Okay, all of the updates are done.  This is ready to be merged from my end.  The map updates are forthcoming.

bescort.lic
```diff
@@ -227,6 +227,7 @@ class Bescort
     ]
 
     args = parse_args(arg_definitions)
+    @flying_mount = get_settings.flying_mount
```
rpherbig, line 230: 💡 Looking at the rest of the file, I think the pattern has been to call `get_settings` within the specific travel function (e.g. `use_flying_mount`). I'm not a fan of that in general, but it does mean we don't have a bunch of instance vars for travel types that will never be used (in that specific execution of `bescort`).

What do you think about storing `get_settings` in an instance var here and then each place in `bescort` can just grab the N settings it needs? That could also be pushed off to a future PR, whether by you or someone else.

bescort.lic
```diff
@@ -301,6 +302,39 @@ class Bescort
 
   private
 
+  def use_flying_mount(type, mode, speed = 'fly')
```
rpherbig, line 305: What do you think about having two separate functions, one for mount and one for dismount?

bescort.lic
```diff
@@ -301,6 +302,39 @@ class Bescort
...
+    case mode
+    when /^mount/
+      bput("get my #{@flying_mount}", /You get /, /You are already holding/)
```
rpherbig, line 308: Since you're referencing `@flying_mount` here, why is it also passed in as the function argument `type`?

bescort.lic
```diff
@@ -1152,7 +1202,7 @@ class Bescort
 
   def segoltha(mode)
     EquipmentManager.new.empty_hands
-    have_changed_gear = EquipmentManager.new.wear_equipment_set?('swimming')
+    have_changed_gear = EquipmentManager.new.wear_equipment_set?('swimming') unless @flying_mount
```
rpherbig, line 1205: This makes the assignment conditional, which can leave `have_changed_gear` as nil. While `nil` is falsey, I think it would be clearer to make this assignment always happen by including `@flying_mount` in the boolean calculation.

Something like `have_changed_gear = !@flying_mount && EquipmentManager.new.wear_equipment_set?('swimming')`.

What do you think?

bescort.lic
```diff
@@ -301,6 +302,39 @@ class Bescort
 
   private
 
+  def use_flying_mount(type, mode, speed = 'fly')
+    case mode
+    when /^mount/
+      bput("get my #{@flying_mount}", /You get /, /You are already holding/)
+      case type
+      when /broom/i
```
jandersson, line 310: it might be worth just downcasing the `type` at the start of the method.  Then you can do string matching here `/broom/i` -> `'broom'`

bescort.lic
jandersson, line 230: New setting? Add it to base.yaml as well

rpherbig: Good catch

BinuDR: Sounds good.

bescort.lic
BinuDR: I started separating it and ended up with method names like this:
mount_flying_mount()
dismount_flying_mount()

They sounded strange so I'm going to leave it as one function for now.  I'm thinking about moving this whole method into common-travel.lic later so I'll address it then.

bescort.lic
BinuDR: It was from an earlier form of the method and I forgot to update it.  Done.

bescort.lic
BinuDR: I ended up using your favorite ternary operator to get around both issues.
`have_changed_gear = flying_mount ? false : EquipmentManager.new.wear_equipment_set?('swimming')`

bescort.lic
BinuDR: Done for the mode, which makes the change from:
`/^mount/` -> `'mount'`  and `/^dismount/i` -> `'dismount'`

But I'm leaving the mount type variable as is because I want to encourage people to enter their mount with both adjective and noun and I only am trying to match whether it's a carpet or broom.

ie: `flying_mount: silk carpet`
so `/carpet/i` will match the yaml setting no matter the adjective and also even if they use any capital letters.

bescort.lic
BinuDR: I'll add it.

BinuDR:
Ok @rpherbig I think I did that correctly.

Let me know if there's anything else.

rpherbig:
It looks like there's a merge conflict with `master`, can you merge that into your branch?

BinuDR:
How does it look now @rpherbig ?

PR #4513, opened 2020-09-28
Updated profile.

Dartellum:


profiles/Dartellum-setup.yaml
```diff
@@ -1155,31 +1199,34 @@ empty_hunting_room_messages:
...
+ - xxxxx
+ - xxxxx
+ - xxxxx
+ - xxxxx
+ - xxxxx
+ - xxxxx
+ - xxxxx
+ - xxxxx
+
+ combat_teaching_skill: Parry
```
rpherbig, line 1226: ❌ There is a leading space before this setting that is breaking the YAML parser

Dartellum: Fixed.

PR #4517, opened 2020-09-29
Update base-hunting.yaml

formayor:
Added more rock troll rooms and got rid of the room with the un-STOW-able rock.

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -631,12 +631,23 @@ hunting_zones:
   - 1554
   # https://elanthipedia.play.net/Rock_Troll_(1)                           60-73
   rock_trolls:
-  - 8946
   - 8947
   - 8954
-  - 8955
   - 8948
   - 8949
+  - 8950
+  - 8951
+  - 8952
+  - 8953
+  - 8972
+  - 8970
+  - 8958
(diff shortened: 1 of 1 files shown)


data/base-hunting.yaml
```diff
@@ -631,12 +631,23 @@ hunting_zones:
   - 1554
   # https://elanthipedia.play.net/Rock_Troll_(1)                           60-73
   rock_trolls:
-  - 8946
```
jandersson, line 634: Is this the room with the unmovable rock? Can you please comment it out if so and and comment about why it was removed. This is just so that we can remember it within the codebase and nudge others from re-adding it.

rpherbig:
Minor comment: in the future I'd leave a comment saying X room number can't be used, so no one tries to add it back in

data/base-hunting.yaml
jandersson: Ah missed my chance :P but feel free to address the comment in a future PR :)

formayor: 8955 is the room with the unmoveable rock.  8946 is the first room in the zone across the barrier and group members will be dumped there if you are not careful.  i took it out for this reason.

PR #4521, opened 2020-10-01
Adding an auto famine usage setting for barbs.

jayRyan24:


validate.lic
```diff
@@ -389,6 +389,13 @@ class DRYamlValidator
             .each { |skill, item| error("Item in weapon_training could not be found in gear listings. '#{skill}: #{item}'") }
   end
 
+  def assert_that_barb_famine_healing_has_both_required_settings(settings)
+    return if settings.barb_famine_healing.empty?
+    return if settings.barb_famine_healing.size() < 2
```
rpherbig, line 394: ❓ What do you think about making this more specific by checking that both `barb_healing['health_threshold']` and ``barb_healing['inner_fire_threshold']` aren't nil?

The error message could also be more specific

rpherbig:
Looks good aside from the one comment ✔️

validate.lic
jayRyan24: Yeah, I was second guessing myself when I put this in, was thinking maybe expanding the system, but regardless, your way is more explicit.

PR #4524, opened 2020-10-04
Pattern hues script. Will take settings from yaml and will do random …

Dartellum:
…colors.

Settings are as follows:

```### Pattern Hues settings ###
### gesture is either random, custom, or a specific setting
pattern_hues:
  duration: 600
  gesture: custom
  pattern_hues_styles:
    pattern:
    - bright
    - hazy
    - gleaming
    - lucent
    hue:
    - soot black
    - emerald
    - bone white
    - snowflake white
    - yellow
    - charcoal black
    - sky blue
##############################

If no colors are set under pattern and hue, it will grab from the entire list of colors.

### CODE REVIEW

**Comment by rpherbig** [comment]
@jayRyan24

**Comment by jayRyan24** [comment]
So, this is fine, but I have some points. 

1) What happens when this runs into other spells being cast, etc. How does it manage to deal with conflicting scripts? 
2) What do you think about making this either a one shot or a loop? You can just have a single parameter `noloop` and it would just cast the random Pattern and then exit? So for different hunts it's either a one shot or a during. My biggest issue is the concern I have for it with conflicting scripts. It could be a non-issue.

Short hunt:

```
before:
- pattern-hues noloop
```

Long hunt:
```
during:
- pattern-hues
```

**Comment by jayRyan24** [comment]
> …colors.
> 
> Settings are as follows:
> 
> ### Pattern Hues settings
> ### gesture is either random, custom, or a specific setting
> pattern_hues:
> duration: 600
> gesture: custom
> pattern_hues_styles:
> pattern:
> - bright
> - hazy
> - gleaming
> - lucent
> hue:
> - soot black
> - emerald
> - bone white
> - snowflake white
> - yellow
> - charcoal black
> - sky blue
> ##############################
> 
> If no colors are set under pattern and hue, it will grab from the entire list of colors.

I do like this, can you put some code tags around the yaml settings in here, in case people come here for the example later. They all look like separate settings and not under the main PH tag.

**Comment by Dartellum** [comment]
> > …colors.
> > Settings are as follows:
> > ### Pattern Hues settings
> > ### gesture is either random, custom, or a specific setting
> > pattern_hues:
> > duration: 600
> > gesture: custom
> > pattern_hues_styles:
> > pattern:
> > 
> > * bright
> > * hazy
> > * gleaming
> > * lucent
> >   hue:
> > * soot black
> > * emerald
> > * bone white
> > * snowflake white
> > * yellow
> > * charcoal black
> > * sky blue
> >   ##############################
> > 
> > If no colors are set under pattern and hue, it will grab from the entire list of colors.
> 
> I do like this, can you put some code tags around the yaml settings in here, in case people come here for the example later. They all look like separate settings and not under the main PH tag.

Done. :)

**Comment by Dartellum** [comment]
> So, this is fine, but I have some questions.
> 
> 1. What happens when this runs into other spells being cast, etc. How does it manage to deal with conflicting scripts?
> 2. What do you think about making this either a one shot or a loop? You can just have a single parameter `noloop` and it would just cast the random Pattern and then exit? So for different hunts it's either a one shot or a during.
> 
> Short hunt:
> 
> ```
> before:
> - pattern-hues noloop
> ```
> 
> Long hunt:
> 
> ```
> during:
> - pattern-hues
> ```
> 
> My biggest issue is the concern I have for it with conflicting scripts. It could be a non-issue.

I can add two things to this: pause/unpause script functions and an arg to pass in for one run or loop. I may have time later this week to do that. The bput just keep retrying if it was in RT from combat-trainer. So, a quick pause then unpause would take care of that. 

Potential arg `now` - `pattern-hues now` - to cast now and no arg is the loop. Thoughts?

**Comment by Dartellum** [comment]
Can run as part of startup routine, as a before, or as a during. Climbing practice gave me a dump, put catch message in and testing.

File: pattern-hues.lic [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+    ]
+    ]
+
+    args = parse_args(arg_definitions)
+    @settings = get_settings
+    now = Time.now
+    duration = @settings.pattern_hues['duration']
+
+    adj = @settings.pattern_hues['pattern_hues_styles']['pattern'] ||
+      adj = ['bright', 'crystalline', 'gleaming', 'glimmering', 'glistening',\
```
rpherbig, line 24: You don't need the second assignment (`adj =`) (same with line 29)

rpherbig:
✔️ looks good to me, let me know when it's ready @Dartellum @jayRyan24

pattern-hues.lic
Dartellum: Removed. I am testing another item and will push once I feel it is working as I expect.

Dartellum:
Removed the Time.now calculations as that was resource-intensive. Used `pause @duration` for the loop. Also, I feel pause and unpause all are used enough they should go into common. However, that is another day and PR. This one is solely focused on Pattern Hues.

Dartellum:
Found a small challenge. I am testing.

Dartellum:
> Found a small challenge. I am testing.

Okay, working well now.

PR #4525, opened 2020-10-05
Changed save to support vault books.

jayRyan24:


inventory-manager.lic
```diff
@@ -63,21 +64,43 @@ class InventoryManager
...
+    if use_vault_book
+      @item_data[@active_character] = @item_data[@active_character].select { |line| !line.include?'vault' }
+    else
+      @item_data[@active_character] = @item_data[@active_character].select { |line| line.include?'vault' }
+    end
 
-    bput('inv list', 'You have:')
-    while line = get
-      break if line =~ /Roundtime/i
+    if use_vault_book
```
rpherbig, line 74: This is basically two different functions with a conditional. I'd prefer to see this split up. Thoughts?

jayRyan24: That's fine, at one point I just had the conditional change up the command and break, but then realized the formatting for the vault book was different, so continued on the same path. I'll separate them, no problem

PR #4533, opened 2020-10-09
Minor change for the removal of go2 from play_no_play

jayRyan24:
I believe this was an error of adding go2 to this list.


jayRyan24:
They wanted to add theurgy to the sanowret no play, I added in go2 because that was in my version. It needs to be pushed without it.

PR #4534, opened 2020-10-10
Add craft overrides and sew for outfitting

jayRyan24:
OK, you can override recipes using the below settings. It doesn't require them all. I also added the ability for higher tier outfitting, this works fine, I just need testers for all the things I can't see. 

```
craft_overrides:
  Outfitting:
    chapter: 5
    item: a knitted blanket
    volumes: 7
    type: knit
  Forging:
    item: a shallow metal cup
  Engineering:
    chapter: 7
    item: a wood bracelet
```


jayRyan24:
Been using this since I submitted this, works well.

PR #4539, opened 2020-10-12
Add unpause to the end of burgle.

robbintt:
Still testing this, I think it's just missing.

### DIFF
--- a/burgle.lic
+++ b/burgle.lic
@@ -564,3 +564,5 @@ Script.running.find_all { |s| !s.paused? && !s.no_pause_all && s.name != Script.
 end
 
 Burgle.new
+
+Script.running.find_all { |s| s.paused? && !s.no_pause_all && $scripts_to_unpause.include?(s.name) }.each(&:unpause)


KatoakDR:
@robbintt It looks like the unpause logic is already in the script when it exits:

```ruby
before_dying do
  #resume the previously paused scripts
  Script.running.find_all { |s| s.paused? && !s.no_pause_all && $scripts_to_unpause.include?(s.name) }.each(&:unpause)
  Flags.delete('burgle-footsteps')
end
```

Can you elaborate on what issues you're having specifically?

robbintt:
For sure, this script hangs after burgle closes in this little loop.  burgle.lic doesn't unpause this script when exiting: 

https://github.com/robbintt/dr-scripts/blob/robbintt/main/resting-chain2.lic#L40

BinuDR:
Which script isn't unpausing?  The one you linked?
If so, it's already using wait_for_script_to_complete, so you don't need that script itself to pause.
Did you try adding `no_pause_all` to the top of your script?

vtcifer:
As Katoak pointed out, there is already unpause logic handled in burgle, in a much safer fashion than just placing it after after the Class instantiation. 

> For sure, this script hangs after burgle closes in this little loop. burgle.lic doesn't unpause this script when exiting:

Hangs is not the same as not unpausing.  If your custom script hangs, but is not paused, then issue is not with burgle but with your custom script.

For the record I could not get your issue to reproduce, so a log would be helpful here.

robbintt:
Nothing hangs, thanks for bringing it up. It's an issue with unpausing.




Best,

Trent Robbins
https://www.linkedin.com/in/trentrobbins

robbintt:
This PR doesn't work anyways, the unpause still does not happen.

PR #4542, opened 2020-10-12
Added my to lines 206 and 211.

Dartellum:
Added **my** to lines 206 and 211 to use your own glue or stain.


KatoakDR:
FYI to those following along, this is continuing the effort of https://github.com/rpherbig/dr-scripts/pull/4538

PR #4544, opened 2020-10-14
Add pick.lic method to balance the lockpick container.

robbintt:
The purpose of this is to evenly wear picks so they can be repaired by a Thief. Playing with the idea of using a flag for this, see code for example settings flag.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -41,6 +41,7 @@ class LockPicker
         echo '***STOPPING DUE TO MINDLOCK***'
         break
       end
+      balance_lockpick_container
       attempt_open(box)
       break if @make_pets && @pet_count >= @pet_goal
     end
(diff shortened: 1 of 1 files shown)


Dartellum:
Could make this optional by adding a new variable to base and setting it to false. 
In base under lock pick settings:
`balance_lockpick_container: false`

Set to true in your yaml if you want to do this. 

In script:
`balance_lockpick_container If balance_lockpick_container`

pick.lic
```diff
@@ -207,6 +208,11 @@ class LockPicker
     end
   end
 
+  def balance_lockpick_container
+    return unless @settings.balance_lockpick_container
```
rpherbig, line 212: `balance_lockpick_container` needs to be added to `base.yaml`

robbintt:
close in favor of https://github.com/rpherbig/dr-scripts/pull/5171 which uses new pick.lic

PR #4547, opened 2020-10-14
Use own brazier by default if you have one assigned.

robbintt:
Previously the script preferred the society brazier.

### DIFF
--- a/enchant.lic
+++ b/enchant.lic
@@ -43,7 +43,7 @@ class Enchant
 
     @use_own_brazier = true
     DRRoom.room_objs.each do |obj|
-      if obj.include?("enchanter's brazier")
+      if obj.include?("enchanter's brazier") and not @brazier
         @brazier = "enchanter's brazier"
         @use_own_brazier = false
       end


Hiinky:
Redundant.  Feature already exists, just assign `enchanting_room:` to prefer your own.

robbintt:
It's not redundant because you should be able to use your own enchanting equipment even if you are in the society -- which should be preferred given the poor quality of the society brazier.

Hiinky:
I do find it redundant and would very strongly prefer it not be added in quite this way.  There is no reason someone with tools should be blocking someone without them simply by using that room.  I would concede that the current setting is not necessarily obvious for how to use your own tools but it exists that way to free up rooms for folks that need the society brazier.  

Perhaps defining your own brazier defaults you to enchant in your safe_room?  It would be more user friendly than the current redirect but accomplishes both goals of using better tools and keeping the society clear.

jayRyan24:
So let me understand how this works. It will go into an enchanting room with a brazier, and then it will use it's own brazier while within the room? This will effectively lead to people entering the room, finding someone in there (even though they aren't using the brazier), and then leaving to find a new place?

This seems kind of iffy to me, as in very rude. It's redundant in that if you have a brazier, you shouldn't be hogging a room that has one and making the limited room/tool situation even more congested. 

If this isn't the case, I am missing something and need clarification.

robbintt:
If this extra condition also needs mitigated then maybe it would be better to require an enchanting room to be specified if you specify your own brazier.  Then the user is at fault if they specify the enchanting society room, and if there is a special need for it they can do it.

robbintt:
I also like the idea of default safe room if not specified.

robbintt:
Superceded by `@use_own_brazier`

PR #4552, opened 2020-10-17
base.yaml - Added flying_mount setting

BinuDR:
Added `flying_mount: ` to base.yaml
Currently it's only used in bescort.


rpherbig:
Can this be rolled into #4512 instead of a separate PR?

PR #4556, opened 2020-10-18
[script] [combat-trainer] Use loaded ammo first if using a repeater crossbow

KatoakDR:
## Changes
* When training a crossbow, if it's a [repeating crossbow](https://elanthipedia.play.net/Repeating_Crossbow) then if there's already loaded ammunition in the chamber then use the quicker `push` command to ready it rather than the slower combination of `load` then `push`.
* In `AttackProcess` class, adds `@crossbow_is_repeater` variable and initializes it to `nil` which means we don't know yet.
* In the `AttackProcess.attack_aimed` method we detect if the crossbow is a repeater and if so then will use the `push` command to ready the ammo for firing more quickly than `load`.
* If there is no more preloaded ammo in the crossbow the script does not reload it fully again. It will do regular "load/aim/fire" routine as if it weren't a repeating crossbow. Maybe in a future revision we could introduce a yaml config to reload the crossbow fully during a combat routine, but for now player's can handle that before calling `combat-trainer`.

## Tests
I've tested with the following weapons:
* sleek diacan repeating arbalest with a wickedly curved lathe (repeating)
* compact ironwood riot crossbow with a yew lath (repeating)
* lever-drawn light crossbow (non-repeating)


combat-trainer.lic
```diff
@@ -370,7 +370,7 @@ class LootProcess
 
     @rituals = get_data('spells').rituals
     echo("  @rituals: #{@rituals}") if $debug_mode_ct
-    
+
```
KatoakDR, line 373: apparently some trailing whitespace got zapped

combat-trainer.lic
```diff
@@ -2606,6 +2606,9 @@ class AttackProcess
   def initialize(settings)
     echo('New AttackProcess') if $debug_mode_ct
 
+    @crossbow_is_repeater = nil # unknown, will detect when train a crossbow
+    echo("  @crossbow_is_repeater: #{@crossbow_is_repeater}") if $debug_mode_ct
```
KatoakDR, line 2610: This new class property is not read from yaml config, not a whole lot of importance in logging that's it nil on init but the script was echoing the values of all the other properties so followed suit

rpherbig: The other properties are only logged because they come from the YAML file. No need to echo nil.

rpherbig: See line 2652 for an example of not echoing.

rpherbig:
Would it be easier to add something like an `repeating_crossbow` property to the item definition?

Dartellum:
> Would it be easier to add something like an `repeating_crossbow` property to the item definition?

You mean like:
```- :name: crossbow
  :adjective:
  :is_leather: true
  :hinders_lockpicking: false
  :is_worn: false
  :swappable: false
  :tie_to:
  :wield: true
  :repeater: true```

I know, I changed from `repeating_crossbow` to `repeater`.

KatoakDR:
> Would it be easier to add something like an `repeating_crossbow` property to the item definition?

@rpherbig @Dartellum Great idea! Gives player flexibility how they want their repeater to be used. Maybe they don’t want all the ammo unloaded in a hunt so it’s ready for that unplanned pvp skirmish or ambush. 😏

I’ll submit updates to this effect. 

Question about the property name: are we going with `repeating_crossbow` or `repeater` or <insert option here>?

rpherbig:
> Question about the property name: are we going with repeating_crossbow or repeater or ?

I suggest the former to be very specific, but go with whatever you think will be clearest to users.

Dartellum:
> > Question about the property name: are we going with repeating_crossbow or repeater or ?
> 
> I suggest the former to be very specific, but go with whatever you think will be clearest to users.

Do repeating stonebows or pelletbows exist? My crossbow is one of the very first in the game, if not the first. 
```tap repeating crossbow
You tap a heavy repeating crossbow constructed from sun-bleached mammoth bones inside your thigh bag.```

KatoakDR:
> Do repeating stonebows or pelletbows exist?
> That is why I suggested repeater in-case other types exit.

That's a good question. I'm not sure if repeating crossbows exist as the nouns "stonebows" or "pelletbows".

However, they all train the "Light/Heavy Crossbow" skill so fall in the family of weapons as [crossbows](https://elanthipedia.play.net/Crossbow). If we go by the concept of the weapon type, rather than weapon noun, then the property `repeating_crossbow` could make sense.

I don't think we would ever see repeating slings or bows, though, as an example.

KatoakDR:
@Dartellum I forgot about this PR, I’ll loop back to give this some love. Thanks

KatoakDR:
Closing due to my own lack of activity on this topic. I'll re-open this when I loop back to it. But for now being closed _should_ stop pinging people every time I backmerge my branches.

PR #4562, opened 2020-10-22
Another minor with a new condition

jayRyan24:


locksmithing.lic
```diff
@@ -98,7 +98,10 @@ class Locksmithing
         echo ('YOU HAVE NO LOCKPICKS ON YOUR LOCKPICK RING/BELT.  SORT THAT OUT!')
         exit
       when /it opens|isn't locked/
-        bput("lock my #{trainer}", 'You quickly lock', 'already locked')
+        case bput("lock my #{trainer}", 'You quickly lock', 'already locked', 'Maybe you should close')
+        when 'Maybe you should close'
+          bput("close my #{trainer}", 'You close')
```
rpherbig, line 103: ❓ Do you also need to try to lock it again?

jayRyan24: Yes, otherwise we go through the loop again. I mean, can remove it, but it's just doubling through the logic again. What do you want, it's minor to me.

jayRyan24:
Canceling this since Katoak is doing something similar.

PR #4565, opened 2020-10-23
[script] [common-items] 'get_item_list' handles empty container

KatoakDR:
### Changes
* Add `bput` string match for `There is nothing` (empty container) so that an empty array is returned rather than a `bput` error about unmatched string.
* Moved `.match` and `.split` calls to their own lines; the original line was very long.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -253,7 +253,9 @@ module DRCI
   end
 
   def get_item_list(container)
-    container_contents = DRC.bput("look in #{container}",'In the .* you see (.*)\.').match(/In the .* you see (?:a|an|some) (?<items>.*)\./)[:items].split(/(?:, | and )?(?:a|an|some) /)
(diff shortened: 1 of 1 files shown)


common-items.lic
```diff
@@ -253,7 +253,9 @@ module DRCI
   end
 
   def get_item_list(container)
-    container_contents = DRC.bput("look in #{container}",'In the .* you see (.*)\.').match(/In the .* you see (?:a|an|some) (?<items>.*)\./)[:items].split(/(?:, | and )?(?:a|an|some) /)
+    container_contents = DRC.bput("look in #{container}", 'In the .* you see (.*)\.', 'There is nothing')
```
rpherbig, line 256: 💡 I think you could also remove `container_contents =` since the var is never used; the function just automatically returns the last evaluation

PR #4566, opened 2020-10-23
[script] [transfer-items] Refactor to use existing utilities

KatoakDR:
### Changes
* `source` and `destination` container names don't have to be full `<adj> <noun>`.
  * Examples: "blue backpack", "blue.back", "backpack", "back"
  * Previously only "blue backpack" (full adj. and noun) or "backpack" (full noun) would work.
* Refactor `transfer_items` method to accept source and destination container names as parameters; removed need for class properties for the same.
* Leverage `DRCI.get_item_list` existing method; able to remove duplicate code.

### Example Usage
```ruby
> ,transfer blue.back thorn.loots

--- Lich: transfer-items active.

[transfer-items]>look in blue.back

In the blue backpack you see a black blindfold, a foo, a bar, and a baz.
> 
[transfer-items]>get blindfold from my blue.back

You get a black blindfold from inside a rugged blue backpack which is in your hitman's backpack.
> 
[transfer-items]>put blindfold in my thorn.loots

You put your blindfold in your thornweave lootsack.

...

--- Lich: transfer-items has exited.
```

### DIFF
--- a/transfer-items.lic
+++ b/transfer-items.lic
@@ -2,41 +2,32 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#transfer-items
 =end
 
+custom_require.call %w[common common-items]
+
 class ItemTransfer
+  include DRC
+  include DRCI
+
   def initialize
     arg_definitions = [
       [
-        { name: 'source', regex: /^[A-z\s\-]+$/i, variable: true, description: 'Source container' },
-        { name: 'destination', regex: /^[A-z\s\-]+$/i, variable: true, description: 'Destination container' }
+        { name: 'source', regex: /^[A-z\.\s\-]+$/i, variable: true, description: 'Source container' },
+        { name: 'destination', regex: /^[A-z\.\s\-]+$/i, variable: true, description: 'Destination container' }
       ]
     ]
     args = parse_args(arg_definitions)
-    @source = args.source
-    @destination = args.destination
-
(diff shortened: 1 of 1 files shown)


transfer-items.lic
```diff
@@ -2,41 +2,32 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#transfer-items
 =end
 
+custom_require.call %w[common common-items]
+
 class ItemTransfer
+  include DRC
```
rpherbig, line 8: 💡 Where is `common` used? I think it could be removed

transfer-items.lic
```diff
@@ -2,41 +2,32 @@
...
-      .split(',')
-      .map { |full_name| full_name.split(' ') }
-      .map { |parts| parts.last.sub('.', '') }
+  def transfer_items(source, destination)
+    DRCI.get_item_list(source)
+      .map { |full_name| full_name.split(' ').last }
       .each do |item|
-        fput("get #{item} from my #{@source}")
-        fput("put #{item} in my #{@destination}")
+        fput("get #{item} from my #{source}")
```
rpherbig, line 26: 💡 Alternatively, these could be changed to `bputs` (and keep `common` above)

KatoakDR: Good point, I'll submit a revision using `bput`

PR #4573, opened 2020-10-26
Targeted Tattoo Support

jayRyan24:
Added a setting to the spells to prepare and target if setting is true.

```
- skill: Sorcery
  prep: invoke tattoo
  tattoo_tm: true
  cast_only_to_train: true
  mana: 25
  cambrinth:
  - 25
```

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -92,7 +92,7 @@ module DRCA
     end
   end
 
-  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_type = 'buff')
+  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false)
     return false unless abbrev
     
     DRC.bput('prep symb', 'You recall the exact details of the', 'But you\'ve already prepared') if symbiosis
@@ -101,13 +101,13 @@ module DRCA
     case match
     when 'Your desire to prepare this offensive spell suddenly slips away'
       pause 1
-      return prepare?(abbrev, mana, symbiosis, command, tattoo_type)
+      return prepare?(abbrev, mana, symbiosis, command, tattoo_tm)
     when 'Something in the area interferes with your spell preparations'
       DRC.bput('rel symb', 'You release the', 'But you haven\'t') if symbiosis
       return false
     end
 
-    DRC.bput("target", 'You begin to weave mana lines') if tattoo_type.include?("targ")
+    DRC.bput("target", 'You begin to weave mana lines') if tattoo_tm
 
     match
   end
@@ -458,10 +458,7 @@ module DRCA
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    tattoo_type = 'buff'
-    tattoo_type = data['tattoo_type']
-
-    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command, tattoo_type)
+    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command, data['tattoo_tm'])
     prepare_time = Time.now
 
     unless settings.cambrinth_items[0]['name']
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -458,7 +458,10 @@ module DRCA
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command, data['tattoo_tm'])
+    tm_tattoo = false
+    tm_tattoo = data['tattoo_tm'] if data['tattoo_tm']
+
+    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command, tm_tattoo)
     prepare_time = Time.now
 
     unless settings.cambrinth_items[0]['name']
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -456,7 +458,10 @@ module DRCA
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command)
+    tm_tattoo = false
```
KatoakDR, line 461: 💅 Minor inconsistency: Here the variable is named tm_tattoo but in the prepare method is spelled tattoo_tm

common-arcana.lic
```diff
@@ -92,7 +92,7 @@ module DRCA
     end
   end
 
-  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare')
+  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false)
```
KatoakDR, line 95: ❓combat-trainer doesn’t look like it’ll pass in the tattoo argument per https://github.com/rpherbig/dr-scripts/blob/596442d3cdb5782cb2b95cf73d01ca80cce2537a/combat-trainer.lic#L1693

Does this mean you plan another PR to that file or am I misreading combat trainer?

common-arcana.lic
```diff
@@ -101,12 +101,14 @@ module DRCA
...
     when 'Your desire to prepare this offensive spell suddenly slips away'
       pause 1
-      return prepare?(abbrev, mana, symbiosis, command)
+      return prepare?(abbrev, mana, symbiosis, command, tattoo_tm)
     when 'Something in the area interferes with your spell preparations'
       DRC.bput('rel symb', 'You release the', 'But you haven\'t') if symbiosis
       return false
     end
 
+    DRC.bput("target", 'You begin to weave mana lines') if tattoo_tm
```
KatoakDR, line 110: 💡Thoughts on adding match strings if there isn’t anything to target? For example, critter dies or something and you no longer can see or are engaged with something targetable.

common-arcana.lic
rpherbig: Combat-trainer would have to be in a separate PR.

Short answer: they have to be separate because common files are not updated in-memory at the same time scripts are.

Breaking changes to common files have to be merged first in a backwards compatible way. Then the change to the consuming script can be merged a few days later, once people (or dependency) have had time to update.

common-arcana.lic
rpherbig: I'm in favor of consistency

common-arcana.lic
```diff
@@ -456,7 +458,10 @@ module DRCA
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command)
+    tm_tattoo = false
+    tm_tattoo = data['tattoo_tm'] if data['tattoo_tm']
```
rpherbig, line 462: Ruby treats nil as falsey. Can you simplify lines 461 and 462 to `tm_tattoo = data['tattoo_tm']` (or even in-line it on line 464)?

common-arcana.lic
jayRyan24: Yeah, I renamed it a couple times, thank you.

common-arcana.lic
jayRyan24: I did it the way above to avoid nils. So, it will set false, if the data has it's own setting it will then set it to whatever it was, Then it will go through. Where is the nil? I am just missing something, I did originally have it 
```prepare?(data['abbrev'], data['mana'], data['symbiosis'], command, data['tattoo_tm'])```
but figured if that line was nill, it's a shitshow. I am just missing something.

File: common-arcana.lic (hunk 3 above)
Comment (jayRyan24) [reply common-arcana.lic:110]: Fixed, ty.

File: common-arcana.lic (hunk 2 above)
Comment (jayRyan24) [reply common-arcana.lic:95]: Yeah, I've made changes to all of them at the same time and it is a cluster. :P

File: common-arcana.lic (hunk 4 above)
Comment (rpherbig) [reply common-arcana.lic]: If `data['tattoo_tm']` is nil, the `if` inside of `prepare?` will evaluate to false

File: common-arcana.lic [hunk 5]
```diff
@@ -101,12 +101,14 @@ module DRCA
...
     when 'Your desire to prepare this offensive spell suddenly slips away'
       pause 1
-      return prepare?(abbrev, mana, symbiosis, command)
+      return prepare?(abbrev, mana, symbiosis, command, tattoo_tm)
     when 'Something in the area interferes with your spell preparations'
       DRC.bput('rel symb', 'You release the', 'But you haven\'t') if symbiosis
       return false
     end
 
+    DRC.bput("target", get_data('spells').prep_messages) if tattoo_tm
```
rpherbig, line 110: ❓ Does `You begin to weave mana lines` need to be added to `prep_messages`?

common-arcana.lic
jayRyan24: NM, I realized my issue when I finally woke.

common-arcana.lic
jayRyan24: No clue, I will check, figured it was a part of it.

jayRyan24: Nope, it's in there. Ty for the double check!

PR #4581, opened 2020-10-28
[script] [equipmanager] Validate gear config matches item's adjective/noun exactly

KatoakDR:
### Background
* This is a precursor to https://github.com/rpherbig/dr-scripts/pull/4484

### Changes
* Adds new temporary method `check_item_matches_regex_pr4484?(item, name_to_match)`, which determines if an item's adjective/name from the gear yaml match the item to use exactly. If it doesn't then shows a warning to the player about the discrepancy. Once #4484 is merged then this method could be removed.
* Refactor spots that compared an item's desc to the `item.short_regex` property to now call a new method `item_matches_desc?(item, desc)`. This method does the same comparison but also makes a validation call to  `check_item_matches_regex_pr4484?(item, name_to_match)`.
* The regex check and warnings can be disabled via user variable `UserVars.equipmanager_check_item_matches_regex` being `false`. By default you are opted in to this new check.

### Out of Scope
* The new  `check_item_matches_regex_pr4484?(item, name_to_match)` method only determines if the YAML config matches an item's `<adjective> <name>` exactly and alerts the player if doesn't. The new method cannot tell the player what values they should be using because that would involve tapping a potentially large permutation of words in an item's description to discern the correct two words to use. Doing so takes a considerable amount of time.
* I'm working on a complementary script `TapIt` that would determine the adjective/noun of your inventory that users could run on their own to influence their YAML config for gear.

#### Example 1
Given you have a sleek diacan brigandine balaclava with twisted black gold edging.

```yaml
# item "starts with" shorthand syntax
gear:
- :adjective: brig
  :name: bala
```

```
| ************************************************
| ***                                             
| *** IMPORTANT: Your gear YAML does not match    
| *** this item's 'adjective' and 'noun' exactly. 
| ***                                             
| *** In a future Lich update, this may cause     
| *** scripts to not work as expected or desired. 
| ***                                             
| *** To reduce chances that scripts grab or trash
| *** the wrong items, please make the following  
| *** corrections to your gear YAML settings.     
| ***                                             
| *** WHAT YOU HAVE:                              
| ***      :adjective: brig         
| ***      :name: bala                        
| ***                                             
| *** ITEM TO MATCH:                              
| ***       sleek diacan brigandine balaclava with twisted black gold edging                      
| ***                                             
| *** To disable these warnings, run vars script: 
| *** vars set equipmanager_check_item_matches_regex=false
| ************************************************
```

#### Example 2
Given you have a sleek diacan brigandine balaclava with twisted black gold edging.

```yaml
# exact adjective and name
gear:
- :adjective: brigandine
  :name: balaclava
```

(no warning would appear)

### DIFF
--- a/equipmanager.lic
+++ b/equipmanager.lic
@@ -9,6 +9,8 @@ class EquipmentManager
   include DRCI
 
   def initialize(settings = nil)
+    @debug_mode_em = UserVars.equipmanager_debug.to_s == 'true' # you opt in to debug mode
+    @check_item_matches_regex = UserVars.equipmanager_check_item_matches_regex.to_s != 'false' # you opt out of regex validation
     items(settings)
   end
 
@@ -61,7 +63,7 @@ class EquipmentManager
   end
 
   def item_by_desc(description)
-    items.find { |item| item.short_regex =~ description }
+    items.find { |item| item_matches_desc?(item, description) }
   end
 
   def notify_missing(lost_items)
@@ -83,7 +85,7 @@ class EquipmentManager
     end
 
     missing_items = worn_items
-                    .reject { |item| combat_items.find { |c_item| item.short_regex =~ c_item } }
+                    .reject { |item| combat_items.find { |c_item| item_matches_desc?(item, c_item) } }
                     .reject { |item| [right_hand, left_hand].grep(item.short_regex).any? ? (stow_weapon(item.short_name) || true) : false }
 
     echo("wear missing items #{missing_items}") if !missing_items.empty? && UserVars.equipmanager_debug
@@ -97,8 +99,8 @@ class EquipmentManager
       echo(worn_items.map(&:short_name).join(',').to_s)
     end
     combat_items
-      .reject { |description| worn_items.find { |item| item.short_regex =~ description } }
-      .map { |description| items.find { |item| item.short_regex =~ description } }
+      .reject { |description| worn_items.find { |item| item_matches_desc?(item, description) } }
+      .map { |description| item_by_desc(description) }
       .compact
       .each { |item| remove_item(item) }
   end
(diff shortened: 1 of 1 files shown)


equipmanager.lic
```diff
@@ -395,4 +397,82 @@ class EquipmentManager
       stow_helper("stow my #{weapon_name}", weapon_name, 'You put', 'You should unload', 'You easily strap', 'You secure your')
     end
   end
+
+  # Given an item and a description (ajd + noun or just noun)
+  # then returns true if the item matches, else false.
+  def item_matches_desc?(item, desc)
+    return false unless item && desc
+    check_item_matches_regex_pr4484?(item, desc)
+    return (item.short_regex =~ desc) != nil
```
KatoakDR, line 406: Line 405: we check if their yaml config matches the item's adj/noun exactly and warn otherwise in preparation for https://github.com/rpherbig/dr-scripts/pull/4484/files#diff-3cc8e7e4d7960e8338d3d8b4201e6c398747480769cfc458fb288bc1eb2d5b57R322

Line 406: the original comparison logic is still in place

equipmanager.lic
```diff
@@ -395,4 +397,82 @@ class EquipmentManager
...
+      echo "item: #{name_to_match}"
+      echo "old regex: #{old_regex}"
+      echo "old matches? #{old_matches}"
+      echo "new regex: #{new_regex}"
+      echo "new matches? #{new_matches}"
+    end
+
+    # If the old regex matches but the new regex doesn't then
+    # notify user of the discrepancy.
+    if @check_item_matches_regex && old_matches && !new_matches
```
KatoakDR, line 450: Note to self, move `@check_item_matches_regex` to line 405 so only call this method if the setting is true

jayRyan24:
Could this logic be moved to the validate script, since what you're doing is validation and what validate does is validation of settings.

jayRyan24:
Could this logic be moved to the validate script, since what you're doing is validation and what validate does is validate settings for syntax? This seems outside the scope of equipment-manager.

KatoakDR:
> Could this logic be moved to the validate script, since what you're doing is validation and what validate does is validate settings for syntax? This seems outside the scope of equipment-manager.

Yep, good suggestion!

And if we move forward with #4484 we still need a way to proactively call the validation to preemptively warn users for N weeks if their config would no longer work. Then when (if) 4484 gets merged then this proactive warning would be removed. It’s only temporary to let people know changes are coming so they can get out ahead of it. Other thoughts on this goal?

Thanks

jayRyan24:
I would say we just warn people of the change, since it's such a big change, at start up. Maybe give them a week of warning, and then merge it and continue warning, and recommend validate. Just seems weird to alter a class with validation warnings, even if just for a time. I am not sure honestly, this could be a moot issue, but possibly could be. Dunno.

rpherbig:
I'm a fan of this being a validator, and then we can tell people to run it to specifically check for this

I'm not a fan of a temporary change this extensive change to a core file, it seems risky

KatoakDR:
Thanks for the feedback. I'll close this PR in favor of moving the gear config check into `validate` script.

Thoughts on how we could have a message presented to users each time they connect? Would that have to be in `dependency` or something else?

PR #4587, opened 2020-11-03
Find-darkbox - Updated room list and added exit for no money.

Dartellum:
Also changed bin to bucket.

May add in to go get money later.

### DIFF
--- a/find-darkbox.lic
+++ b/find-darkbox.lic
@@ -5,15 +5,18 @@ if Script.running?('roomnumbers')
   before_dying { start_script('roomnumbers') }
 end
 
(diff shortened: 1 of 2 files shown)


find-darkbox.lic
```diff
@@ -42,7 +42,17 @@ loop do
   #16315 bank
   fput('play dark')
   pause 1
-  break if Flags['darkbox-no-money']
+  #break if Flags['darkbox-no-money']
+  if Flags['darkbox-no-money']
+    DRCT.walk_to(16_315)
+    case DRC.bput('withdraw 3 platinum', 'we are not lending money at this time', 'The clerk counts out')
+    when 'we are not lending money at this time'
+      exit
```
rpherbig, line 50: 💡 It would be nice to `echo` or `message` to the user that they need to get more money

Dartellum: I will add in on the next enhancement. The Darkbox completely disappeared from the map last night. The script ran through trying to find it and then returned to the last known location and tried to play. I am going to add exit if Darkbox is not in the room after it runs everywhere.

PR #4592, opened 2020-11-07
Lamprey game

Dartellum:


lamprey.lic
```diff
@@ -0,0 +1,57 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#lamprey
+=end
+
+custom_require.call(%w[events common-travel common common-items])
```
rpherbig, line 5: :bulb: I don't think `events` needs to be here

lamprey.lic
```diff
@@ -0,0 +1,57 @@
...
+    case bput('get lamprey from trough', 'You bite down', 'You just got a lamprey.')
+    when 'You just got a lamprey.'
+      message("*** You will need to wait 10 minutes to play again! ***")
+      exit
+    end
+  end
+
+  def check_prize
+    junk_list = get_settings.darkbox_junk.map { |x| /#{x}/i }
+    prize = DRC.right_hand
```
rpherbig, line 35: 💡 If you're relying on the prize to be in your right hand, should you empty hands before starting?

PR #4593, opened 2020-11-07
Add examine script.

robbintt:
Perform a series of examine verbs on an object.

### DIFF
--- a/examine.lic
+++ b/examine.lic
@@ -1,0 +1,25 @@
+class Examine
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'target', regex: /^[A-z\s\-]+$/i, variable: true, description: 'Object to examine' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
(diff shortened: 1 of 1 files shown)


examine.lic
```diff
@@ -0,0 +1,25 @@
+class Examine
```
rpherbig, line 1: 💡 All scripts should have a link here to elanthipedia & an entry in elanthipedia with more info on the script and its args/etc.

examine.lic
```diff
@@ -0,0 +1,25 @@
...
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'target', regex: /^[A-z\s\-]+$/i, variable: true, description: 'Object to examine' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
+    @target = args.target
```
rpherbig, line 9: 💡 I don't think this needs to be an instance var, it could just be an arg to `examine_object`

PR #4594, opened 2020-11-08
[script] [crowns] Handle empty hands when stow prize

KatoakDR:
### Background
* Kamisama got `Lich: error: can't convert NilClass to String` and confirmed it happened when his hands were empty
* Discord thread: https://discord.com/channels/745675889622384681/771867662551744534/774875895092477962

### Changes
* Consolidated logic for checking and stowing prizes in left or right hand
* Coil rope otherwise it's too long to fit in a container

FYI @Dartellum

### DIFF
--- a/crowns.lic
+++ b/crowns.lic
@@ -12,9 +12,9 @@ class Dice
 
   def initialize
     arg_definitions = [
-    [
-      { name: 'bet', regex: /\d+/, optional: true, description: 'Optional override for betting amount.' }
-    ]
+      [
+        { name: 'bet', regex: /\d+/, optional: true, description: 'Optional override for betting amount.' }
+      ]
     ]
 
     settings = get_settings
@@ -117,27 +117,21 @@ class Dice
   end
 
   def check_prize
-    if @trash_items.any? { |item| Regexp.new(item) =~ right_hand }
-      bput("put my #{righthand} in bucket", 'You put', 'You drop')
-    else
-      case bput("stow #{righthand}", 'You put', 'to fit in the', 'What were', 'Stow what?')
-      when 'to fit in the'
-        message("*** Item is too big to fit in your stow container! ***")
-        exit
+    [left_hand, right_hand]
+      .compact
+      .each do |in_hand|
+        if @trash_items.any? { |item| Regexp.new(item) =~ in_hand }
+          bput("put my #{in_hand} in bucket", 'You put', 'You drop')
+        else
+          case bput("stow #{in_hand}", 'You put', 'to fit in the', 'What were', 'Stow what?')
+          when 'to fit in the'
+            message("*** Item is too big to fit in your stow container! ***")
+            exit
+          end
+          pause
+        end
       end
-      pause
-    end
-    # Check left hand as well. On lag may have items in both hands.
-    if @trash_items.any? { |item| Regexp.new(item) =~ left_hand }
-      bput("put my #{lefthand} in bucket", 'You put', 'You drop')
-    else
-      case bput("stow #{lefthand}", 'You put', 'to fit in the', 'What were', 'Stow what?')
-      when 'to fit in the'
-        message("*** Item is too big to fit in your stow container! ***")
-        exit
-      end
-      pause
-    end
+    pause
   end
 end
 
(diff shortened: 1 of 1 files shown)


crowns.lic
```diff
@@ -117,27 +117,21 @@ class Dice
...
-    # Check left hand as well. On lag may have items in both hands.
-    if @trash_items.any? { |item| Regexp.new(item) =~ left_hand }
-      bput("put my #{lefthand} in bucket", 'You put', 'You drop')
-    else
-      case bput("stow #{lefthand}", 'You put', 'to fit in the', 'What were', 'Stow what?')
-      when 'to fit in the'
-        message("*** Item is too big to fit in your stow container! ***")
-        exit
+    [left_hand, right_hand]
+      .compact
```
KatoakDR, line 121: Build list of the items to stow as prizes. If either hand is empty then `.compact` will reduce the size of the array. If no items in our hands then the array is empty and this method won't do anything and just return. No error.

crowns.lic
```diff
@@ -117,27 +117,21 @@ class Dice
...
+      .each do |in_hand|
+        if @trash_items.any? { |item| Regexp.new(item) =~ in_hand }
+          bput("put my #{in_hand} in bucket", 'You put', 'You drop')
+        else
+          case bput("stow #{in_hand}", 'You put', 'to fit in the', 'What were', 'Stow what?')
```
KatoakDR, line 126: 💡 @Dartellum this may be a good spot to configuration for _which_ container someone wants their loot to be stowed in in case their default container isn't appropriate. They could specify it in `dice_loot_container`. Thoughts?

crowns.lic
```diff
@@ -117,27 +117,25 @@ class Dice
...
+    [left_hand, right_hand]
+      .compact
+      .each do |in_hand|
+        if @trash_items.any? { |item| Regexp.new(item) =~ in_hand }
+          bput("put my #{in_hand} in bucket", 'You put', 'You drop')
+        else
+          # Coil rope so we can store it
+          if /\brope\b/ =~ in_hand
+            fput("coil my #{in_hand}")
+          end
```
KatoakDR, line 129: Coil rope otherwise too long to stow and script will abort prematurely

Dartellum: I found this method was not always catching if the item was on the junk list or not. Rope is on the junk list. I went back to 
`thing = DRC.right_hand || DRC.left_hand
    waitrt?
    if thing
      case thing
      when *trash_items
        DRC.bput("put my #{thing} in bucket", 'You put', 'You drop')` and it catches everything better. 

I also consolidated the old lists into `hollow_eve_junk` for a new PR (have not yet put in). 

Could add `dice_loot_container:` at the same time.

crowns.lic
```diff
@@ -117,27 +117,25 @@ class Dice
...
+          case bput("stow #{in_hand}", 'You put', 'to fit in the', 'What were', 'Stow what?')
+          when 'to fit in the'
```
Dartellum, line 131: This should have caught the rope was too big.

KatoakDR: It did catch that the rope was too long but the script aborted unecessarily because it was a recoverable error. That's why I added the `coil rope` step above so that it could be stored. You pointed out that the junk filter wasn't working always and would have filtered the rope out to begin with, but that didn't happen for me so I added the behavior to handle it just in case.

Dartellum: No worries. I have the junk list working better but I like the catch. Who knows, the person may want the rope. :)

rpherbig:
✔️ Looks ok to me from a code-quality point of view.

@KatoakDR @Dartellum let me know when you two are done discussing/revising and this can be merged.

Dartellum:
He hit the same error with something not on his list. I find that weird because if it is not on the list it should stow. I incorporated these changes into my local copy to test. I also added `dice_loot_container` while at it.

Dartellum:
> He hit the same error with something not on his list. I find that weird because if it is not on the list it should stow. I incorporated these changes into my local copy to test. I also added `dice_loot_container` while at it.

I had him test again with my updated version with this cahnge in it and it worked for the cambrinth ring (Errored out before).

Dartellum:
I say go ahead and merge this. I will then put together the PR with the consolidated junk list, the configurable loot container, and this method in the other game scripts. Trying to keep everything consistent.

KatoakDR:
Thanks @Dartellum!

@rpherbig you're good to merge this one when you're able (to get it into people's hands sooner rather than later) though Dartellum will have a follow up PR at some point for more enhancements.

Thanks!

PR #4596, opened 2020-11-09
Hollow eve games update

Dartellum:
Made all the game scripts this year use the same logic for handling junk that @KatoakDR put in for crowns.lic. Added loot storage containers with defaults in `base.yaml.` Made one list for junk/trash items as the games use the same pool now. The name is `hollow_eve_junk.` Made all these scripts use `trash_items:` for consistency. 

For the loot containers, should I just make one, `hollow_eve_loot_container:,` or keep the individual ones I have created?

### DIFF
--- a/find-darkbox.lic
+++ b/find-darkbox.lic
@@ -13,12 +13,12 @@ arg_definitions = [
 ]
 
 args = parse_args(arg_definitions)
-@override = args.override
-
-trash_items = get_settings.hollow_eve_junk.map { |x| /#{x}/i }
-darkbox_stop_on_wounded = get_settings.darkbox_stop_on_wounded
-he_use_herbs_remedies = get_settings.he_use_herbs_remedies
-hollow_eve_loot_container = get_settings.hollow_eve_loot_container
+override = args.override
+settings = get_settings
+trash_items = settings.hollow_eve_junk.map { |x| /#{x}/i }
+darkbox_stop_on_wounded = settings.darkbox_stop_on_wounded
+he_use_herbs_remedies = settings.he_use_herbs_remedies
+hollow_eve_loot_container = settings.hollow_eve_loot_container
 
 if Script.running?('roomnumbers')
   stop_script('roomnumbers')
@@ -78,7 +78,7 @@ loop do
   end
 
   if reget(10, 'your wounds make it impossible') && he_use_herbs_remedies
-    if @override
+    if override
       DRC.wait_for_script_to_complete('heal-remedy', ['quick', 'override'])
     else
       DRC.wait_for_script_to_complete('heal-remedy', ['quick'])
--- a/find-darkbox.lic
+++ b/find-darkbox.lic
@@ -15,9 +15,10 @@ arg_definitions = [
 args = parse_args(arg_definitions)
 @override = args.override
 
-junk_list = get_settings.darkbox_junk.map { |x| /#{x}/i }
+junk_list = get_settings.hollow_eve_junk.map { |x| /#{x}/i }
 darkbox_stop_on_wounded = get_settings.darkbox_stop_on_wounded
 he_use_herbs_remedies = get_settings.he_use_herbs_remedies
+darkbox_loot_container = get_settings.darkbox_loot_container
 
 if Script.running?('roomnumbers')
   stop_script('roomnumbers')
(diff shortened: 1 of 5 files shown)


KatoakDR:
> For the loot containers, should I just make one, hollow_eve_loot_container:, or keep the individual ones I have created?
I vote for one `hollow_eve_loot_container:` variable shared by all the HE game scripts

Dartellum:
> > For the loot containers, should I just make one, hollow_eve_loot_container:, or keep the individual ones I have created?
> 
> I vote for one `hollow_eve_loot_container:` variable shared by all the HE game scripts because it's less config to maintain and, at least for now, the games pull from the same or very nearly the same loot sources, right?

I will make the change tonight. Does make it simpler. I will also fix the failed check.

Dartellum:
<<I will also fix the failed check.

I am not sure why the check failed. All I did was change the names of the trash variable in the one and then the loot variable in the latest.

rpherbig:
> <<I will also fix the failed check.
> 
> I am not sure why the check failed. All I did was change the names of the trash variable in the one and then the loot variable in the latest.

TravisCI reports:
```
TestValidYaml#test_all_settings_in_base_yaml [/home/travis/build/rpherbig/dr-scripts/test/test_valid_yaml.rb:32]:
Expected all settings to be in base.yaml.
--- expected
+++ actual
@@ -1 +1 @@
-["darkbox_junk", "pumpkin_junk"]
```

find-darkbox.lic
```diff
@@ -15,9 +15,10 @@ arg_definitions = [
 args = parse_args(arg_definitions)
 @override = args.override
 
-junk_list = get_settings.darkbox_junk.map { |x| /#{x}/i }
+trash_items = get_settings.hollow_eve_junk.map { |x| /#{x}/i }
 darkbox_stop_on_wounded = get_settings.darkbox_stop_on_wounded
```
rpherbig, line 19: 💡 I'd prefer to only call `get_settings` once and cache it

rpherbig:
✔️ I'm ok with merging this once the test is fixed

find-darkbox.lic
Dartellum: Made settings = get_settings and adjusted the rest accordingly.

Dartellum:
> > <<I will also fix the failed check.
> > I am not sure why the check failed. All I did was change the names of the trash variable in the one and then the loot variable in the latest.
> 
> TravisCI reports:
> 
> ```
> TestValidYaml#test_all_settings_in_base_yaml [/home/travis/build/rpherbig/dr-scripts/test/test_valid_yaml.rb:32]:
> Expected all settings to be in base.yaml.
> --- expected
> +++ actual
> @@ -1 +1 @@
> -["darkbox_junk", "pumpkin_junk"]
> ```

How do I fix this? Both of those were removed and replaced with `hollow_eve_junk`.

rpherbig:
> How do I fix this? Both of those were removed and replaced with `hollow_eve_junk`.

We can't have any profiles with the bad settings.

Btw, if you're renaming settings it could cause problems for people who are already using those settings. Is that a worry here?

Dartellum:
> > How do I fix this? Both of those were removed and replaced with `hollow_eve_junk`.
> 
> We can't have any profiles with the bad settings.
> 
> Btw, if you're renaming settings it could cause problems for people who are already using those settings. Is that a worry here?

I'll add the two settings back with nothing in them. Once the new scripts are merged, the settings will change, unless they have them hardcoded into their yaml. I can make a general announcement on LNET and Discord that the junk list names changed.

rpherbig:
> > > How do I fix this? Both of those were removed and replaced with `hollow_eve_junk`.
> > 
> > 
> > We can't have any profiles with the bad settings.
> > Btw, if you're renaming settings it could cause problems for people who are already using those settings. Is that a worry here?
> 
> I'll add the two settings back with nothing in them. Once the new scripts are merged, the settings will change, unless they have them hardcoded into their yaml. I can make a general announcement on LNET and Discord that the junk list names changed.

There are profiles in source control right now with these settings. I think it's very possible adding unused settings to `base.yaml` could confuse users.

Dartellum:
> There are profiles in source control right now with these settings. I think it's very possible adding unused settings to `base.yaml` could confuse users.

I could label them depreciated. Remember, all the games disappeared or the script (darkbox) changed this year.

PR #4608, opened 2020-11-16
Hollow eve games update #3

Dartellum:
Following up on a comment that stow, in `case DRC.bput("stow #{in_hand} in my #{@hollow_eve_loot_container}", 'You put', 'to fit in the', 'What were', 'Stow what?')` was changing store default if one had that item defined. Found the common-items function `DRCI.put_away_item?` works well. Yes, I could have just changed the stow to put but we already have a function so I decided to use it. 

I experimented with result = DRCI.... then case result but it did not work. The reget function (or are these methods?) grabbed what I needed to find. In the scripts I could, I made a function for beeping and exiting. I also fixed how the scripts were getting the settings. I made it so settings = get_settings was used instead of each setting reaching back to the yaml. 

Also added another catch for the disposal of the junk item for those instances one emptied their hands or another script did before this line was called.

### DIFF
--- a/crowns.lic
+++ b/crowns.lic
@@ -122,20 +122,28 @@ class Dice
       if in_hand
         case in_hand
         when *@trash_items
-          DRC.bput("put my #{in_hand} in bucket", 'You put', 'You drop')
+          DRC.bput("put my #{in_hand} in bucket", 'You put', 'You drop', 'What were', 'Stow what?')
         else
           # Coil rope so we can store it
           if /\brope\b/ =~ in_hand
             fput("coil my #{in_hand}")
           end
-          case DRC.bput("stow #{in_hand} in my #{@hollow_eve_loot_container}", 'You put', 'to fit in the', 'What were', 'Stow what?')
-          when 'to fit in the'
-            DRC.message("*** Item is too big to fit in your stow container! ***")
-            exit
+          DRCI.put_away_item?(in_hand, @hollow_eve_loot_container)
+          if reget(5, /to fit in the/)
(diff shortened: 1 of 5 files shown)


crowns.lic
```diff
@@ -122,22 +122,30 @@ class Dice
       if in_hand
         case in_hand
         when *@trash_items
-          DRC.bput("put my #{in_hand} in bucket", 'You put', 'You drop')
+          DRC.bput("put my #{in_hand} in bucket", 'You put', 'You drop', 'What were', 'Stow what?')
```
rpherbig, line 125: 💡 I think there's a 'dispose junk' method in DRCI that might help

crowns.lic
```diff
@@ -122,22 +122,30 @@ class Dice
...
           # Coil rope so we can store it
           if /\brope\b/ =~ in_hand
             fput("coil my #{in_hand}")
           end
-          case DRC.bput("stow #{in_hand} in my #{@hollow_eve_loot_container}", 'You put', 'to fit in the', 'What were', 'Stow what?')
-          when 'to fit in the'
-            DRC.message("*** Item is too big to fit in your stow container! ***")
-            exit
+          DRCI.put_away_item?(in_hand, @hollow_eve_loot_container)
+          if reget(5, /to fit in the/)
```
rpherbig, line 132: 💡 The `?` in `put_away_item?` means that it returns a boolean (true/false). You could capture that and use it instead of `reget`. However, that would mean you couldn't differentiate these two cases and instead have a more generic 'couldn't put away' error message

rpherbig (approved):

PR #4610, opened 2020-11-16
Preserve before butchering

Corillian:
A corpse can be preserved before it's butchered, I added this ability for some extra Thanatology training.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -652,6 +652,7 @@ class LootProcess
     echo "Attempting necromancer ritual #{ritual} on #{mob_noun}" if $debug_mode_ct
 
     if ritual.eql?('butcher')
+      do_necro_ritual(mob_noun, 'preserve', game_state)
       butcher_corpse(mob_noun, ritual, game_state)
 
       @last_ritual = ritual
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -362,6 +362,9 @@ class LootProcess
     @ritual_type = thanatology['ritual_type'].downcase
     echo("  @ritual_type: #{@ritual_type}") if $debug_mode_ct
 
+    @preserve_before_butcher = thanatology['preserve_before_butcher']
+    echo("  @preserve_before_butcher: #{@preserve_before_butcher}") if $debug_mode_ct
+
     @cycle_rituals = @ritual_type == 'cycle'
     echo("  @cycle_rituals: #{@cycle_rituals}") if $debug_mode_ct
 
@@ -652,7 +655,10 @@ class LootProcess
     echo "Attempting necromancer ritual #{ritual} on #{mob_noun}" if $debug_mode_ct
 
     if ritual.eql?('butcher')
-      do_necro_ritual(mob_noun, 'preserve', game_state)
+      if @preserve_before_butcher
+        do_necro_ritual(mob_noun, 'preserve', game_state)
+      end
+      
       butcher_corpse(mob_noun, ritual, game_state)
 
       @last_ritual = ritual
--- a/profiles/Melborne-setup.yaml
+++ b/profiles/Melborne-setup.yaml
@@ -206,6 +206,7 @@ necromancer_healing:
 dissect_and_butcher: true
 thanatology:
   ritual_type: cycle
+  preserve_before_butcher: true
   heal: true
   store: true
   harvest_container: haversack
--- a/profiles/SampleNecromancer-setup.yaml
+++ b/profiles/SampleNecromancer-setup.yaml
@@ -305,6 +305,8 @@ dissect_and_butcher: true
 thanatology:
   # cycles through rituals, or you can set the specific type you want to do based on availiabity or preference
   ritual_type: cycle
+  # performs the preserve ritual before butchering to maximize Thanatology training
+  preserve_before_butcher: true
   # heal during rituaal
   heal: true
   # Do or do not store body parts, you will just drop them otherwise
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -135,6 +135,7 @@ dance_actions_stealth:
 # hash of necromancer-related ritual settings. See Melborne-setup
 thanatology:
   ritual_type:
+  preserve_before_butcher: true
   heal: false
   store: false
   harvest_container: 


combat-trainer.lic
```diff
@@ -652,6 +652,7 @@ class LootProcess
     echo "Attempting necromancer ritual #{ritual} on #{mob_noun}" if $debug_mode_ct
 
     if ritual.eql?('butcher')
+      do_necro_ritual(mob_noun, 'preserve', game_state)
```
rpherbig, line 655: 💡 What do you think about instead moving the `butcher` check after line 663 and adding `butcher` to the array on line 663?

rpherbig:
The code looks fine to me ✔️ 

@jayRyan24 you implemented support for butcher in #2550 - do you have any thoughts on this change?

cc @izawwlgood

combat-trainer.lic
jayRyan24: This is more a question for @izawwlgood or other Necros. I'm so far detached from what is good/proper. :(

jayRyan24: Seems like this would be better as optional rather than required, could add a lot of RT and maybe isn't required, these are my first thoughts, just no real foundation to make any accurate statements beyond that.

jayRyan24: Just double checked this, doesn't appear to be needed! 

The butcher_corpse method handles this already!

```
  def butcher_corpse(mob_noun, ritual, game_state)
    return unless ritual.eql?('butcher')
    echo "Butchering the #{mob_noun}'s corpse!" if $debug_mode_ct
    echo ' Only butchering it once!' if $debug_mode_ct && @dissect_and_butcher && !@ritual_type.eql?('butcher')

    do_necro_ritual(mob_noun, 'preserve', game_state)
```

jayRyan24 (changes requested):

jayRyan24:
This should be closed as this is already handled within the butcher_corpse method.

Corillian:
Hmm, let me double check. I actually made this change some time ago but realized I never pushed it so it's possible it's obsolete now.

jayRyan24:
> Hmm, let me double check. I actually made this change some time ago but realized I never pushed it so it's possible it's obsolete now.

The butcher_corpse method did this upon first implementation from my original PR. #2550

izawwlgood:
My preference would be to have adding preserve first to be optional. The
ritual sequences were set up to give you choice, and preserve->butcher
isn't really a needed addition for training Thanatology.

Butcher is primarily included for maxing out your Butchery ranks, not for
training Than/FA/Skinning.

Corillian:
@jayRyan24 I'm not seeing where `Preserve` is being called prior to butchering...

@izawwlgood Is there a situation where you don't want the extra Thanatology?

jayRyan24:
> @jayRyan24 I'm not seeing where `Preserve` is being called prior to butchering...
> 
> @izawwlgood Is there a situation where you don't want the extra Thanatology?

This: 
```    if ritual.eql?('butcher')
      butcher_corpse(mob_noun, ritual, game_state)
```

Your changed adds the preserve ritual before butcher_corpse in the above location. The butcher_corpse method calls the same method you are calling before. So, you are calling it twice now, rather than once. 

The `def butcher_corpse` method, look it up:
```  def butcher_corpse(mob_noun, ritual, game_state)
    return unless ritual.eql?('butcher')
    echo "Butchering the #{mob_noun}'s corpse!" if $debug_mode_ct
    echo ' Only butchering it once!' if $debug_mode_ct && @dissect_and_butcher && !@ritual_type.eql?('butcher')

    do_necro_ritual(mob_noun, 'preserve', game_state)
```

Corillian:
Yes that's correct, `do_necro_ritual` is being called twice, which can possibly be avoided but there's so much additional logic in there I decided it was difficult to asses the full impact of _not_ calling it twice. If you're the original author than I you're definitely the one to assess whether or not I should make a slimmer call to `Preserve`. You still haven't shown me where `Preserve` is being called prior to butchering - it's not being called, not in the outer `do_necro_ritual` or anywhere else, that's the whole reason I made this change in the first place.

@izawwlgood There is only a single situation under which I can think roundtime would be an issue - which is if you're extremely high circle and constantly running `USOL` to quickly burn through corpses in an extremely high spawn room. The whole reason I added this is that my character, which is sub 100, _needed_ the extra Thanatology. I could certainly make it optional but it seems like the edge case for not wanting it would only apply to special circumstances for extremely high circle characters.

izawwlgood:
The xp from preserve is low relative to the rt (4s). I just think it's kind
of moot to include. If you want to train than you do dissect. Fa/skin, you
do harvest. Butcher is only used by people who are maxing butchery and once
the happens you switch to cycle which alternates between dissect and
harvest.

Corillian:
@izawwlgood  The `combat-trainer` does not kill creatures very quickly. In my experience every little bit of Thanatology is a big help, particularly before you get `USOL`.

Corillian:
If the consensus is to add a check or a flag that's totally fine with me, I'm just not sure where this pre-existing call to `Preserve` is @jayRyan24 has mentioned.

jayRyan24:
> Yes that's correct, `do_necro_ritual` is being called twice, which can possibly be avoided but there's so much additional logic in there I decided it was difficult to asses the full impact of _not_ calling it twice. If you're the original author than you're definitely the one to assess whether or not I should make a slimmer call to `Preserve`. You still haven't shown me where `Preserve` is being called prior to butchering - it's not being called, not in the outer `do_necro_ritual` or anywhere else, that's the whole reason I made this change in the first place.

It's not OUTSIDE of the butcher ritual, but the butcher ritual method calls preserve from within. So, when you call it from outside of the method, and then the first thing the method does is call it again, you are calling it twice...

Corillian:
This is the original `butcher_corpse`:

``` Ruby
def butcher_corpse(mob_noun, ritual, game_state)
    return unless ritual.eql?('butcher')
    echo "Butchering the #{mob_noun}'s corpse!" if $debug_mode_ct
    echo ' Only butchering it once!' if $debug_mode_ct && @dissect_and_butcher && !@ritual_type.eql?('butcher')

    loop do
      result = bput("perform #{ritual} on #{mob_noun}", @rituals['butcher'], @rituals['failures'])
      break if result.empty? || @rituals['failures'].any? { |msg| result.include?(msg) }
      bput("drop my #{right_hand}", 'You drop', 'You discard', 'Please rephrase')
      break if @dissect_and_butcher && !@ritual_type.eql?('butcher')
    end

    do_necro_ritual(mob_noun, 'dissect', game_state) if @dissect_and_butcher && !@ritual_type.eql?('butcher')
  end
```

Wherein that is `Preserve` being called?

jayRyan24:
> Wherein that is `Preserve` being called?

OK, so here's what I'm going to do. I will state I was wrong because the code is different and changed from when I first implemented it. I will say though that your passive aggressiveness is straight up a dick thing to do. 

I clearly gave you examples of the code I was referencing twice, I then also gave you an example and said this is what it is when I implemented it. A normal person would have said "Oh, yours is different than mine, here's what I have" instead of this bullshit passive aggressive dickhead comments. 

Maybe I should have as well, but I'm currently editing code for work, and also in a meeting at the same time while trying to respond to this.

This should be implemented with a conditional Thanatology check, that's my last input for what needs to be done here.

jayRyan24:
> @izawwlgood The `combat-trainer` does not kill creatures very quickly. In my experience every little bit of Thanatology is a big help, particularly before you get `USOL`.

It uses "attack" as the default, unless you turn on weak attacks. If you want to kill as fast as possible, make sure they're off, and make sure to use more mana for TMs. I think I have an example of how I trained TM and other skills using a TM USOL hunt. The trainer kills things fast or slow, depending on how you specialize the training with it.

izawwlgood:
If you're high circle, you've already maxed Butchery and aren't running it
anymore.
If you want corpses fast, you're running USOL/QE/CFB. That and ritual:
cycle is what I do to train than/skin/fa.

Corillian:
@jayRyan24 You clearly gave examples of **the wrong code**. I was looking at _the current code_, not realizing you were posting something different, and asking for clarification. There is nothing passive aggressive about that, I _promise_ you would know it if I was trying to be a dick, so spare me your bruised ego. I'll update the PR with a conditional later this evening, thanks.

jayRyan24:
> @jayRyan24 You clearly gave examples of **the wrong code**. I was looking at _the current code_, not realizing you were posting something different, and asking for clarification. 

If you didn't notice that the code was "clearly the wrong code", why didn't you say "it's different than what I have? That shows me you didn't even look at my examples as it would have been clear. 

>There is nothing passive aggressive about that, I _promise_ you would know it if I was trying to be a dick, so spare me your bruised ego. I'll update the PR with a conditional later this evening, thanks.

`Wherein that is Preserve being called?` Is a passive aggressive statement given the context of the entire conversation. There is no bruised ego here, I simply had the wrong code and was out of date, and since I showed you multiple examples of what I was references, there's nothing wrong with what I did. I simply admitted to having different code right when you showed me yours, since I actually looked at your reference. If you had taken a small amount of time to look over the code I posted to you, in 4 separate examples, you would have noticed something out of whack. Since you didn't reference my code being different from yours each time I posted, I could only assume that the code was the same as yours.

rpherbig:
Good evening, folks. Apparently this was the wrong day for me to be out sick. Let's please keep it civil and focus on the point of this PR:
1. Does `combat-trainer` already preserve before butchering?
1. If not, should it?
1. If it should, should it be conditional?

Here's what I've gathered from reading over the discussion:
1. `combat-trainer` used to preserve before butchering, but does not now.
1. @Corillian would like it to preserve before butchering
1. It looks like this was intentionally removed in #3510. This tells me that we have people with different goals, so a conditional seems like the best path forward.

I leave it to the participants (hereafter "the brain trust") to decide what that conditional should be. If the brain trust can come up with a condition that doesn't require a new setting, I'd love that (I've already seen one suggestion to make it conditional on Thanatology field experience). Otherwise, a new setting seems reasonable.

jayRyan24:
> Good evening, folks. Apparently this was the wrong day for me to be out sick. Let's please keep it civil and focus on the point of this PR:
> 
> 1. Does `combat-trainer` already preserve before butchering?
> 2. If not, should it?
> 3. If it should, should it be conditional?
> 
> Here's what I've gathered from reading over the discussion:
> 
> 1. `combat-trainer` used to preserve before butchering, but does not now.
> 2. @Corillian would like it to preserve before butchering
> 3. It looks like this was intentionally removed in #3510. This tells me that we have people with different goals, so a conditional seems like the best path forward.
> 
> I leave it to the participants (hereafter "the brain trust") to decide what that conditional should be. If the brain trust can come up with a condition that doesn't require a new setting, I'd love that (I've already seen one suggestion to make it conditional on Thanatology field experience). Otherwise, a new setting seems reasonable.

The issue is that butchering, as of now, already checked thanatology as a conditional, so this might not even be worth it. Unless we move the check to a lower number. I think it doesn't need it, if it has been removed and is training fine already, and as @izawwlgood stated, it trains very little for the required RT. 

As a side note, I don't think butchering should even have a requirement of thanatology, given that you might want to butcher without training. I have no actual dog in this hunt as I sold my Necro a long time ago, so I'll let the people that actually use it to figured it out.

jayRyan24:
Just make it a setting and add it to thanatology...
```
thanatology:
  ritual_type: butcher
  preserve_before_butcher: true
  heal: false
```

izawwlgood:
To be clear butchery is a separate track to master. So once mastered, you
no longer need to use butchery to train.

I forget the specific conversation back when we first set up
cycle/dissect/harvest/butcher/preserve, but someone pointed out that the
thanatology xp from butchery was so low it shouldn't be used for training.
The only time you want it automated is to grind out the butchery mastery.

For younger necromancers they may lose out on a squeek of thanatology xp,
but they can use butcher in a hunt designed to maximize corpses for a short
period of time until butchery is mastered.

jayRyan24:
I know I wrote all this shit, I'm just so out of the loop, and don't remember much from it. Was kind of a flash, thanks for the input.

Corillian:
> I forget the specific conversation back when we first set up cycle/dissect/harvest/butcher/preserve, but someone pointed out that the thanatology xp from butchery was so low it shouldn't be used for training.

In my experience butchering is the _best_ way to train Thanatology. I exclusively use butchering for my training and combining it with `Preserve` maximizes the utility of a corpse, particularly prior to acquiring `USOL`. I've pushed a flag for toggling `Preserve` per @jayRyan24's suggestion. Ultimately it's up to the powers that be as to whether or not you want to merge or if you'd like any additional modifications.

izawwlgood:
You may be thinking of maximizing Thanatology per corpse, which is largely
unnecessary.

Some short testing shows the following:
preserve -
2 fa 2 than, 2s

butcher -
2 than, 7s

dissect -
fa 3 than 3, 10s

harvest -
2 skin 2 than, 4s

Butcher teaches poorly for the RT, and corpses aren't in short supply. If
you want to maximize Than locking, you use cycle while running USOL.

rpherbig:
The code changes look good ✔️ 

❓ What should this new setting default to? On the one hand, `false` means users do not see a change in behavior (i.e. it's opt-in). On the other hand, `true` means if this does train better the users see an improvement immediately.

> Butcher teaches poorly for the RT, and corpses aren't in short supply. If you want to maximize Than locking, you use cycle while running USOL.

Perhaps `true` for the sample since it's set up for a low circle Necro who doesn't have USOL, and `false` in `base.yaml`?

Corillian:
> Butcher teaches poorly for the RT, and corpses aren't in short supply. If you want to maximize Than locking, you use cycle while running USOL.

Right, this is what I've been saying the _entire_ time - the scenario you're presenting is only workable if you are in a high swarm-rate room running USOL. This puts everyone below circle 80 at a significant training disadvantage. Below circle 80 it usually takes all or most of your mana to kill a single creature so corpses _are_ difficult to come by and the RT doesn't matter because it gives your mana time to recharge. I think the flag I've added straddles the need of low-circle necro's and higher circle necro's who just want to sit back with USOL and mow down everything in a room.

izawwlgood:
QE is available very early on. As I said, I think this applies to a very
narrow range of Necromancers - specifically, those who are for some reason
hunting something they cannot kill quickly, and who also have not yet
capped Butchery ranks.

Once you've maxed your Butchery ranks, no one is using Butcher as their
ritual anymore. More options are nice, but I think this only really applies
to people still pushing Butchery ranks, and squeaks out an extremely small
amount of additional Thanatology XP per corpse. :shrug:, have at it.

Hiinky:
I'm late but my two cents: I believe this was explicitly removed about a year ago for folks not wanting to waste time.

Bmoney420:
For a secondary point of view - I've just trained a Necro, from 0-100, using the existing settings. In no way did I have trouble training Thanatology. I never needed to do specific Thanatology hunts because I wasn't learning enough from it. Having now capped Butchery and turned that off - I think cycling the Rituals is more than enough to keep Thanatology moving. I even spent some time hunting Constructs (short term in both instances) and was able to push into the next zone before I came close to needing Thanatology for circling purposes, or before it fell massively behind.

MahtraDR:
@rpherbig I think we can close this one out. No activity in a year, and I think it's not needed today in any case.

PR #4625, opened 2020-11-26
Missed the at sign, @, on these games.

Dartellum:
When I switched to the new put_away_item? function I did not add the global variable symbol, @, in these games as I cut and pasted from one of the games that did not require it.


Dartellum:
@rpherbig I know it is Thanksgiving, but could you merge this?

PR #4640, opened 2020-12-03
Will invoke runestone spell.

Dartellum:
Question - change script name to invoke-rune since that is what it is doing?

This script will take the spell name of the runestones available for purchase at [Sierack's Reagents](https://elanthipedia.play.net/Sierack%27s_Reagents) in **Shard** as an argument. 

Usage: `;focus-rune` "Clear Vision" will find the proper runestone then cast the spell. If the runestone is not available, one can have the script purchase one. The script saves the current room, purchases the proper runestone, then returns to the room and continues on. You can leave runestone_purchase false (default value) and let your normal restock routine replenish your runestone(s).

Settings:
```
##########################
### Runestone settings ###
##########################
runestone_storage: backpack
runestone_harness: 30
### Setting runestone_purchase to true means you must have access to Shard.
### These runestones are only purchasable there.
### Uses restock section.
runestone_purchase: false 
```

```
#######################
### Restock section ###
#######################
restock:
  arrow:
  bolt:
  ammo:
    hometown: Hibarnhvidar
    name: smooth rocks
    room: 12172
    price: 90
    size: 30
    stackable: true
    quantity: 60
  cv_rune:
    hometown: Shard
    name: avaes runestone
    room: 13105
    price: 4510
    size: 1
    stackable: false
    quantity: 2
```

Calling in T2, before, or after sections:
`- focus-rune "Clear Vision"`

This work can lead to using runestones in combat via the buff method. I also added the runestones to base-consumables. My specific use case was for CV for locksmithing.


KatoakDR:
> Question - change script name to invoke-rune since that is what it is doing?
That's my vote :)

data/base-consumables.yaml
```diff
@@ -80,6 +80,105 @@ Shard:
     size: 1
     stackable: false
     quantity: 1
+  ### Runestones ###
+  ec_rune:
```
KatoakDR, line 84: Thoughts on naming the runes with their full spell name, or including the full spell name as a property or as a comment? I don't have the spell acronyms memorized so having the spell names listed somewhere to see at a glance here would be helpful.

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+    echo @rune if @debug
+
+    ### Because Protection from Evil has a lowercase letter for the middle word
+    ### this work around is required.
+    if spell == "protection from evil"
+      temp = "Protection from Evil"
+    else
+      temp = spell.split(/ |\_|\-/).map(&:capitalize).join(" ")
+    end
+    echo temp if @debug
```
KatoakDR, line 50: Consider putting this into its own method, something like:
```ruby
def format_spell_name(spell)
  if spell == "protection from evil"
    return "Protection from Evil"
  else
    return spell.split(/ |\_|\-/).map(&:capitalize).join(" ")
  end
end
```
Then you can call this earlier in the script when you grab the spell argument:
```ruby
spell = format_spell_name(args.spell)
```

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+    until DRSpells.active_spells.include?(temp)
+      pause 1 while mana < 40
```
KatoakDR, line 52: Instead of hardcoding `40` mana limit, could we reuse `buff_spell_mana_threshold` setting from the user's yaml?

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+      if DRCI.exists?("#{@rune} runestone in #{@runestone_storage}")
+        runestone_cast
+      elsif runestone_purchase
+        buy_runestone
+      else
+        DRC.message("*** Out of #{@rune} runestones! ***") unless runestone_purchase
```
KatoakDR, line 58: I think the `unless` bit can be removed here since per the if/else chain you can't get into this block _unless_ runestone_purchase is already false, correct?
```suggestion
        DRC.message("*** Out of #{@rune} runestones! ***")
```

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+        beep
+        exit
+      end
+    end
+  end
+
+  def spell_name(rune)
```
KatoakDR, line 65: Suggest a more explicit name of what this function does, such as `get_rune_name_for_spell(spell)`

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+    runestone_map = { "eagle's cry" => 'electrum', 'protection from evil' => 'topaz',
+        'refresh' => 'quartz', "glythtide's gift" => 'axinite',
+        'shadows' => 'azurite', 'arc Light' => 'calavarite',
+        'geyser' => 'celestite', 'bless' => 'elbaite',
+        'fire shards' => 'rhodonite', 'athleticism' => 'selenite',
+        'calm' => 'xibaryl', 'zephyr' => 'imnera',
+        'compost' => 'asketine', "clear vision" => 'avaes' }
+    runestone_map[rune] ? runestone_map[rune] : rune
```
KatoakDR, line 73: If the argument is not found in the `runestone_map`, what is the use case for returning the argument as-is rather than printing a warning message and exiting?

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def spell_name(rune)
+    runestone_map = { "eagle's cry" => 'electrum', 'protection from evil' => 'topaz',
+        'refresh' => 'quartz', "glythtide's gift" => 'axinite',
+        'shadows' => 'azurite', 'arc Light' => 'calavarite',
+        'geyser' => 'celestite', 'bless' => 'elbaite',
+        'fire shards' => 'rhodonite', 'athleticism' => 'selenite',
+        'calm' => 'xibaryl', 'zephyr' => 'imnera',
+        'compost' => 'asketine', "clear vision" => 'avaes' }
```
KatoakDR, line 72: For readability, I suggest formatting each key => value entry on its own line, standardize enclosing each value in double-quotes, and sorting alphabetically:
```suggestion
    runestone_map = {
      "arc Light" => "calavarite",
      "athleticism" => "selenite",
      "bless" => "elbaite",
      "calm" => "xibaryl",
      "clear vision" => "avaes",
      "compost" => "asketine",
      "eagle's cry" => "electrum",
      "fire shards" => "rhodonite",
      "geyser" => "celestite",
      "glythtide's gift" => "axinite",
      "protection from evil" => "topaz",
      "refresh" => "quartz",
      "shadows" => "azurite",
      "zephyr" => "imnera"
    }
```

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+    DRCI.get_item("#{@rune} runestone", @runestone_storage)
+    if reget(3, "You get a useless #{@rune} runestone")
+      DRCI.dispose_trash("#{@rune} runestone")
+      return
+    end
+    case bput('invoke my rune', 'Closing your eyes', 'Well, that was fun', 'backfire', 'The spell pattern resists')
+    when 'Well, that was fun'
+      waitrt?
+      DRCI.dispose_trash("#{@rune} runestone")
+      runestone_cast
```
KatoakDR, line 86: Line 86 will recall this method, if you are now out of runestones then when execution gets back to line 82 then the command could hang if you don't have a runestone held, correct? If so, then I suggest adding a matchstring to line 82 to that effect then adding a `when` block to handle that the player has run out of runestones trying to cast.

focus-rune.lic
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def buy_runestone
+    current_room = Room.current.id
+    DRC.wait_for_script_to_complete('restock')
+    DRCT.walk_to(current_room)
+  end
+
+  before_dying do
+
```
KatoakDR, line 105: 🧹 deleting extra newline
```suggestion
```

KatoakDR (commented):
I like the idea of this script, thanks! I've left a few comments/suggestions.

data/base-consumables.yaml
Dartellum: Done. I could now use the restock logic in the purchase def and key in only on the one we want instead of running all the restock items.

focus-rune.lic
Dartellum: Added in.

focus-rune.lic
Dartellum: I like the `buff_spell_mana_threshold` option.

focus-rune.lic
Dartellum: True. When I rewrote this section I did not remove that bit.

focus-rune.lic
Dartellum: I ordered them by the order in the catalog, in case you were wondering why it was the way it was.

focus-rune.lic
Dartellum: Fixed.

```
def runestone_cast
    get_runestone
    case bput('invoke my rune', 'Closing your eyes', 'Well, that was fun', 'backfire', 'The spell pattern resists', 'Invoke what?')
    when 'Well, that was fun'
      waitrt?
      DRCI.dispose_trash("#{@rune} runestone")
      runestone_cast
    when 'backfire','The spell pattern resists'
      return
    when 'Invoke waht?'
      out_of_runestone
    end
```

Made a new def.

```
def out_of_runestone
    DRC.message("*** Out of #{@rune} runestones! ***")
    beep
    exit
  end
```

focus-rune.lic
Dartellum: If you provide an invalid argument this happens:
```
>;invoke-rune Clear Vision
[invoke-rune: ***INVALID ARGUMENTS DON'T MATCH ANY PATTERN***]
Provided Arguments: 'Clear Vision'
   ;invoke-rune <spell> [debug]
   spell        Specify spell to cast from runestone. Spells are: 'Eagle's Cry'
                  'Protection from Evil' Refresh 'Glythtide's Gift' Shadows 'Arc Light'
                  Geyser Bless 'Fire Shards' Althleticism Calm Zephyr Compost
                  'Clear Vision.' Use double quotes. 
   debug         
--- Lich: invoke-rune has exited.
```

focus-rune.lic
Dartellum: Changed.

Dartellum:
For the purchasing section, I really do not want to bring all of the restock logic in... perhaps I could make restock take an arg. I will ponder this while walking my dog.

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#focus-rune
```
rpherbig, line 2: ❌ Should match script name

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
...
+    runestone_purchase = settings.runestone_purchase
+
+    echo spell if @debug
+    echo @harness if debug
+    @rune = get_rune_name_for_spell_name(spell)
+    echo @rune if @debug
+    spell = format_spell_name(args.spell)
+
+    until DRSpells.active_spells.include?(spell)
+      pause 1 while mana < buff_spell_mana_threshold
```
rpherbig, line 46: Why isn't this check done before each cast?

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
...
+        runestone_cast
+      elsif runestone_purchase
+        buy_runestone
+      else
+        out_of_runestone
+      end
+    end
+  end
+
+  def format_spell_name(spell)
```
rpherbig, line 57: Why not just downcase all spell names instead?

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
...
+
+    echo spell if @debug
+    echo @harness if debug
+    @rune = get_rune_name_for_spell_name(spell)
+    echo @rune if @debug
+    spell = format_spell_name(args.spell)
+
+    until DRSpells.active_spells.include?(spell)
+      pause 1 while mana < buff_spell_mana_threshold
+      if DRCI.exists?("#{@rune} runestone in #{@runestone_storage}")
```
rpherbig, line 47: Would `DRCI.inside?("#{rune} runestone", @runestone_storage)` be clearer?

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
...
+      "eagle's cry" => "electrum",
+      "fire shards" => "rhodonite",
+      "geyser" => "celestite",
+      "glythtide's gift" => "axinite",
+      "protection from evil" => "topaz",
+      "refresh" => "quartz",
+      "shadows" => "azurite",
+      "zephyr" => "imnera"
+    }
+    runestone_map[rune] ? runestone_map[rune] : rune
```
rpherbig, line 84: If the spell name isn't in the map, isn't that an error?

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
...
+    ### Because Protection from Evil has a lowercase letter for the middle word
+    ### this work around is required.
+    if spell == "protection from evil"
+      return "Protection from Evil"
+    else
+      return spell.split(/ |\_|\-/).map(&:capitalize).join(" ")
+    end
+  end
+
+  def get_rune_name_for_spell_name(rune)
```
rpherbig, line 67: Should this arg be `spell_name` or something similar?

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
...
+      "protection from evil" => "topaz",
+      "refresh" => "quartz",
+      "shadows" => "azurite",
+      "zephyr" => "imnera"
+    }
+    runestone_map[rune] ? runestone_map[rune] : rune
+  end
+
+  def runestone_cast
+    get_runestone
```
rpherbig, line 88: Does this mean if it takes multiple invokes for the spell to be cast, it'll get/stow/get/stow/etc. (i.e. get and stow each attempt)?

invoke-rune.lic
```diff
@@ -0,0 +1,134 @@
...
+                                               'The spell pattern resists',\
+                                               'Invoke what?')
+    when 'Well, that was fun'
+      waitrt?
+      DRCI.dispose_trash("#{@rune} runestone")
+      runestone_cast
+    when 'backfire','The spell pattern resists'
+      return
+    when 'Invoke what?'
+      out_of_runestone
```
rpherbig, line 101: Why doesn't this buy a runestone if you passed in that arg?

invoke-rune.lic
Dartellum: I changed the script name and forgot this. :(

invoke-rune.lic
Dartellum: This checks against the game's name for it. I could play around with downcase this.

invoke-rune.lic
Dartellum: I will test. :)

invoke-rune.lic
Dartellum: It is but it spits out the usage screen if that happens.

invoke-rune.lic
Dartellum: This returns the rune name. That is why I used rune. Yes, it feeds in spell name.

invoke-rune.lic
Dartellum: It does. I will move so it does not.

invoke-rune.lic
Dartellum: testing
```
    when 'Invoke what?'
      buy_runestone
      runestone_cast
    end
```

invoke-rune.lic
rpherbig: Using `rune` for the name of the argument confuses me since the input is the spell name, not rune name.

invoke-rune.lic
rpherbig: If we're confident that rune is in the map, why do the ternary?

invoke-rune.lic
Dartellum: Okay.

invoke-rune.lic
Dartellum: Moved into spell casting section.

invoke-rune.lic
Dartellum: Works great.

invoke-rune.lic
Dartellum: Changed logic here to match only the spell name. I had an instance where it cast the wrong spell because it was not bound (/\b .. \b/).

invoke-rune.lic
Dartellum: I believe I fixed this. I tested and seems to work as expected.

invoke-rune.lic
Dartellum: The new logic is working as expected.

invoke-rune.lic
Dartellum: Changed to spell_name.

invoke-rune.lic
Dartellum: Downcase was not working for me. I kept getting undefined variable downcase.

Dartellum:
Many small updates to the script. I have tested with a useless runestone, one that goes useless while trying to invoke, and other ways.

PR #4642, opened 2020-12-04
[script] [common] Even more match strings for 'collect' command

KatoakDR:
Encountered these on a new character with sub 30 ranks of outdoorsmanship trying to collect rocks while hunting ship rats in Crossing


Dartellum:
I had the lifeless one on two of my characters (one has almost 1000 outdoormanship ranks). I believe you get that message if you cannot forage in that room.

PR #4651, opened 2020-12-11
Add support for custom t2 avoids

vtcifer:


KatoakDR (approved):
Nice!

t2.lic
```diff
@@ -23,6 +20,15 @@ class T2
     UserVars.repair_timer_snap ||= Time.now
     UserVars.t2_timers ||= {}
 
+    t2_avoids = @settings.t2_avoids
+    t2_avoids.each do |avoid|
+      if avoid['state']
```
rpherbig, line 25: ❓ Does this mean if you have `state: off` this will still trigger?

vtcifer: I think I get where you're going with this, and agree the syntax for the settings is a little funky, but with how avoids are structured I'm not sure of a good way around it.  

If you have state: off it will trigger "avoid join" for example: ex:

```yaml
t2_avoids:
  - type: hold
    state: on
  - type: drag
    state: on
  - type: join
    state: off
```
Produces:

> --- Lich: t2 active.
> [t2]>awaken
> But you are not sleeping!
> > 
> [t2]>avoid !hold
> You're now avoiding attempts to hold.
> > 
> [t2]>avoid !drag
> You're now avoiding attempts to drag.
> > 
> [t2]>avoid join
> You're now allowing attempts to join.

rpherbig: Ah, I'd forgotten about the conversion for YAML. Could you update the comment to be clearer?

vtcifer: Can you clarify? Do you mean in t2.lic or in base.yaml?

rpherbig: In base.yaml, I suggest using true instead of on. That should follow the established pattern enough that a comment isn't necessary.

Dartellum:
This documented any?

PR #4652, opened 2020-12-12
[script][restock] Added and extra equal sign on line 37 as a single equal sign did not work.

Dartellum:
My F2P guy was trying to go to Shard as it was evening. Checked with `== Shard` now working as expected.

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -160,3 +160,4 @@ module DRCM
       end
     end
   end
+end
--- a/common-money.lic
+++ b/common-money.lic
@@ -160,16 +160,3 @@ module DRCM
       end
     end
   end
-
-  def deposit_coins_from_specific_bank?(keep_copper, settings, town)
-    DRCT.walk_to(get_data('town')[town]['deposit']['id'])
-    DRC.release_invisibility
-    DRC.bput('wealth', 'Wealth:')
-    case DRC.bput('deposit all', 'you drop all your', 'You hand the clerk some coins', "You don't have any", 'There is no teller here', 'reached the maximum balance I can permit', 'You find your jar with little effort', 'Searching methodically through the shelves')
-    when 'There is no teller here'
-      return
-    end
-    minimize_coins(keep_copper).each { |amount| withdraw_exact_amount?(amount, settings) } if settings.hometown == town
-    DRC.bput('check balance', 'your current balance is', 'You find your jar with little effort', 'Searching methodically through the shelves')
-  end
-end
--- a/common-money.lic
+++ b/common-money.lic
@@ -128,7 +128,7 @@ module DRCM
   def deposit_coins(keep_copper, settings, hometown = nil)
     return if settings.skip_bank
     hometown = settings.hometown if hometown == nil
-    
+
     DRCT.walk_to(get_data('town')[hometown]['deposit']['id'])
     DRC.release_invisibility
     DRC.bput('wealth', 'Wealth:')
@@ -159,5 +159,17 @@ module DRCM
         break false
       end
     end
+  end
+
+  def deposit_coins_from_specific_bank?(keep_copper, settings, town)
+    DRCT.walk_to(get_data('town')[town]['deposit']['id'])
+    DRC.release_invisibility
+    DRC.bput('wealth', 'Wealth:')
+    case DRC.bput('deposit all', 'you drop all your', 'You hand the clerk some coins', "You don't have any", 'There is no teller here', 'reached the maximum balance I can permit', 'You find your jar with little effort', 'Searching methodically through the shelves')
+    when 'There is no teller here'
+      return
+    end
+    minimize_coins(keep_copper).each { |amount| withdraw_exact_amount?(amount, settings) } if settings.hometown == town
+    DRC.bput('check balance', 'your current balance is', 'You find your jar with little effort', 'Searching methodically through the shelves')
   end
 end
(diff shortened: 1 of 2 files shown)


restock.lic
```diff
@@ -34,6 +34,7 @@ class Restock
       num_needed = item['quantity'] - remaining
       buy_num = (num_needed / item['size'].to_f).ceil
       coin_needed += buy_num * item['price']
+      coin_needed += 2502 if UserVars.sun['night'] = true && town == "Shard"
```
rpherbig, line 37: Could you add a comment explaining why this is necessary?

Dartellum: When you go to Shard at night as a non citizen you need to bribe the night attendant to enter the shop. The bribe is 2501 normally. If you do not bribe the attendant you cannot enter the store. If you use the funds you fetched from the bank just for the runestones, or whatever restock items you are trying to get, you are then short 2501 for the items you tried to obtain. The runestone shop keeper has his dog attack you if you do not have enough money.

rpherbig: Is this extra money per item to buy (i.e. inside the loop), or just a one-time cost (i.e. outside the loop)?

Dartellum: It is a one time cost as long as you stay inside the shop and purchase everything at once. I need to watch restock closer, or check the logs, to see if all the items are purchased at once.

Dartellum:
Changed to draft to continue testing. I am going to add in a deposit after purchasing from remote town.

Dartellum:
Added a new method to common-money to deposit any leftover money withdrawn from the town back excluding the amount to keep on hand for your hometown area. This became necessary as I was left with lots of copper on me.

restock.lic
```diff
@@ -47,13 +48,16 @@ class Restock
...
+    deposit_coins_from_specific_bank?(@keep_copper, @settings, town)
   end
 
   def setup
     @settings = get_settings
     @restock = @settings.restock
     @hometown = @settings.hometown
     @runestone_storage = @settings.runestone_storage
+    amount, denom = @settings.sell_loot_money_on_hand.split(' ')
+    @keep_copper = convert_to_copper(amount, denom)
```
rpherbig, line 60: I think this can be simplified by just storing `sell_loot_money_on_hand` and making only one call to `withdraw_exact_amount` in `deposit_coins_from_specific_bank`

common-money.lic
```diff
@@ -160,4 +160,16 @@ module DRCM
       end
     end
   end
+
+  def deposit_coins_from_specific_bank?(keep_copper, settings, town)
```
rpherbig, line 164: How does this function differ from `deposit_coins`?

restock.lic
```diff
@@ -47,13 +48,16 @@ class Restock
         reget(2, 'runestone') ? DRCI.put_away_item?(item['name'], @runestone_storage) : stow_hands
       end
     end
+    deposit_coins_from_specific_bank?(@keep_copper, @settings, town)
```
rpherbig, line 51: What do you think about guarding this call on `town == "Shard"`?

restock.lic
```diff
@@ -37,6 +37,7 @@ class Restock
       item['buy_num'] = buy_num
       items_to_restock.push(item)
     end
+    coin_needed += 2502 if UserVars.sun['night'] == true && town == "Shard"
```
rpherbig, line 40: Could you add a comment to the script explaining why this is necessary?

restock.lic
Dartellum: You mean switch to /\bShard\b/?

restock.lic
Dartellum: Added.

common-money.lic
Dartellum: The only difference is no @hometown function is in it, just like get_money_from_specific_bank? 

I put in the minimize_coins line because I found it deposited all my money in my hometown. Or it did on early runs. I need to test again if I need it or not. Def `deposit_coins` assumes hometown and even sets it explicitly. I could see about combining the two different defs, `get_money_from_bank` with `get_money_from_specific_bank?` and `deposit_coins` with `get_money_from_specific_bank?`, this week or so. 

Now that I am really looking at it, let me try both in testing and see what happens. :) Going to change this back to draft.

restock.lic
Dartellum: I may not need `deposit_coins_from_specific_bank?`. I was following what was put in to make it work with restock but see the original had an option already but defaults to `settings.hometown` if nil. Testing with the old. I will look into this.

Dartellum:
Removed `deposit_coins_from_specific_bank?` as I can get the original to work. I am going to update the other two scripts that use `get_money_from_specific_bank?` - trade and herb-stock, as not needed if you pass town as `"#{town}"` in the call.

restock.lic
```diff
@@ -37,7 +37,9 @@ class Restock
       item['buy_num'] = buy_num
       items_to_restock.push(item)
     end
-    get_money_from_specific_bank?("#{coin_needed} copper #{town_currency(town)}", town) if coin_needed > 0
+    ### The following line is for the bribe to the night attendant to let you in.
+    coin_needed += 2502 if UserVars.sun['night'] = true && town == /\bShard\b/
```
KatoakDR, line 41: I think there's a typo in `UserVars.sun['night'] = true` as that's assigning the value `true` to the variable rather than checking if it's nighttime. I think you can just remove the `= true` and ruby will evaluate to true if the expression is truthy. To be super explicit, then need a second equal sign for comparison.
```suggestion
    coin_needed += 2502 if UserVars.sun['night'] && town == /\bShard\b/
```

Dartellum: Changed to ` coin_needed += 2502 if UserVars.sun['night'] && town == /\bShard\b/`

Dartellum:
Okay, this has worked for all four currently running characters. The F2P do not try to go to leave their bounds.

PR #4666, opened 2020-12-21
[script] [common-arcana] Match strings to retry activating barb ability

KatoakDR:
### Changes
* Add match string for `While swimming?` when you fail to meditate in water
* Add match string for `You must be unengaged` when you fail to meditate while engaged
* Return `true` if ability activated, `false` otherwise so calling scripts can respond accordingly
* Renamed method to include question mark to indicate it's a predicate, which matches format for `activate_khri?` too

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -81,14 +81,21 @@ module DRCA
 
   def activate_barb_buff(name)
     ability_data = get_data('spells').barb_abilities[name]
-    case DRC.bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand', 'You must be sitting')
+    case DRC.bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand', 'You must be sitting', 'You must be unengaged', 'While swimming?')
+    when 'You must be unengaged'
+      DRC.retreat
+      activate_barb_buff(name)
     when 'You must be sitting'
       DRC.retreat
-      DRC.bput('sit', 'You sit', 'You are already', 'You rise')
+      DRC.bput('sit', 'You sit', 'You are already', 'You rise', 'While swimming?')
       activate_barb_buff(name)
     when 'You should stand'
       DRC.fix_standing
       activate_barb_buff(name)
+    when ability_data['activated_message']
+      true
+    else
+      false
     end
   end
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -72,26 +72,36 @@ module DRCA
     skills
       .reject { |name| DRSpells.active_spells[name] }
       .each do |name|
-        activate_barb_buff(name)
+        activate_barb_buff?(name)
         waitrt?
         pause settings.meditation_pause_timer.to_i if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -81,14 +81,21 @@ module DRCA
 
   def activate_barb_buff(name)
```
rpherbig, line 82: It would be great to rename this to `activate_barb_buff?` now that it's a predicate. But I realize that makes it a little more complicated of a change. What are your thoughts?

common-arcana.lic
```diff
@@ -81,14 +81,21 @@ module DRCA
...
     ability_data = get_data('spells').barb_abilities[name]
-    case DRC.bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand', 'You must be sitting')
+    case DRC.bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand', 'You must be sitting', 'You must be unengaged', 'While swimming?')
+    when 'You must be unengaged'
+      DRC.retreat
+      activate_barb_buff(name)
     when 'You must be sitting'
       DRC.retreat
-      DRC.bput('sit', 'You sit', 'You are already', 'You rise')
+      DRC.bput('sit', 'You sit', 'You are already', 'You rise', 'While swimming?')
```
rpherbig, line 90: Won't this lead to a loop of sitting and getting that error message repeatedly?

common-arcana.lic
KatoakDR: Sounds reasonable to me, and should just be a "find and replace" change. Thanks

common-arcana.lic
KatoakDR: I'll do another test and grab the logs to confirm if its entering recursive loop here or just spammy in the combat-trainer event loop trying to activate a buff in an area it'll never be able to.

Maybe worth exploring some kind of backoff/sleep on barb abilities that aren't able to be activated so that they aren't instantly repeated (and fail).

KatoakDR:
Switched back to draft while I address code review comments

common-arcana.lic
rpherbig: > should just be a "find and replace" change

Kind of. Since this PR is only for changes to the `common*` files, you'd probably want two functions (new name and old name) for a brief period of time (with one just calling the other). In the next PR they can be collapsed.

KatoakDR: Done, I've added new method `activate_barb_ability?` to include ? since it's a predicate now like `activate_khri?`. The previous method delegates to this new one and will be removed in a future PR.

common-arcana.lic
KatoakDR: Good catch, if you try to sit in too deep of water then now return `false` because won't be able to activate the ability

common-arcana.lic
```diff
@@ -72,26 +72,36 @@ module DRCA
...
-        activate_barb_buff(name)
+        activate_barb_buff?(name)
         waitrt?
         pause settings.meditation_pause_timer.to_i if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
         DRC.fix_standing
       end
   end
 
+  # Deprecated, will be removed in a future update
+  # Use 'activate_babr_buff?' instead.
```
rpherbig, line 83: 💡 `activate_barb_buff?`

PR #4667, opened 2020-12-21
[script] [combat-trainer] Don't assume barb ability activated successfully

KatoakDR:
### Background
* If combat trainer fails to activate a barb buff and that buff wasn't already active then the flag `"ap-#{name}-expired"` won't ever reset because the expiry message won't ever be seen. This causes the ability to never be retried for the rest of that hunt.

### Changes
* ~~Depends on https://github.com/rpherbig/dr-scripts/pull/4666~~
* Don't assume the barb ability was activated just because it was attempted to be activated (you may have lacked inner fire at the time or in a location that prevents the ability from being activated, such as trying to meditate in deep water)
* If the ability is activated _then_ resets the flag
* Only tries to activate ability if you have sufficient inner fire based on new numeric setting, `barb_buffs_inner_fire_threshold`

### Config Example
_Added to base.yaml_
```yaml
# The minimum inner fire needed to activate a barb_buff in your buff_nonspells.
# You won't try to activate a barb buff until you have this much inner fire.
# You need some inner fire in reserve to pay the periodic cost of abilities.
barb_buffs_inner_fire_threshold: 15
```


rpherbig:
Looks good (once the other PR is merged, of course)

combat-trainer.lic
```diff
@@ -1915,11 +1918,16 @@ class AbilityProcess
     @barb_buffs
       .select { |name| Flags["ap-#{name}-expired"] }
       .each do |name|
-        Flags.reset("ap-#{name}-expired")
-        activate_barb_buff(name)
-        echo "Activated barb_buff #{name}!" if $debug_mode_ct
-        waitrt?
-        pause 6 if get_data('spells').barb_abilities[name]['type'].eql? 'meditation'
+        if mana < @barb_buffs_inner_fire_threshold
+          echo "Not activating barb_buff #{name} until have at least #{@barb_buffs_inner_fire_threshold} inner fire" if $debug_mode_ct
```
rpherbig, line 1922: 💡 Maybe include in the echo what the current inner fire level is?

combat-trainer.lic
```diff
@@ -1915,11 +1918,16 @@ class AbilityProcess
...
+        elsif activate_barb_buff(name)
```
rpherbig, line 1923: ❓ Should this now reference `activate_barb_buff?`? If so, can `activate_barb_buff` be removed from DRCA?

KatoakDR: Yes, now that https://github.com/rpherbig/dr-scripts/pull/4666 got merged I need to update this to use the predicate named method name and remove the old name from DRCA. Thanks

combat-trainer.lic
KatoakDR: Yeah, that'd be a nice addition

KatoakDR:
@rpherbig This is ready for your review again. Also note, we could wait until https://github.com/rpherbig/dr-scripts/pull/4679 is merged and then this PR could remove the Flag management for tracking barb buffs because we’d know if something was up or not via `DRSpells['ability name']` like checking any buff that’s up.

rpherbig:
> we could wait until #4679 is merged and then...

That would be my preference. I think that PR can be merged soon.

KatoakDR:
Converted to draft -- need to merge `master` and take advantage that barb abilities are now available via `DRSpells.active_spells`

KatoakDR:
Closing due to my own lack of activity on this topic. I'll re-open this when I loop back to it. But for now being closed _should_ stop pinging people every time I backmerge my branches.

KatoakDR:
Replaced by https://github.com/rpherbig/dr-scripts/pull/4752

PR #4668, opened 2020-12-24
Switched from stow to Equipmentmanager method to honor containers for items.

Dartellum:
Updated to so you put your items away in the proper location based on sheath or container settings under gear.


rpherbig:
Looks good to me

jandersson (approved):

PR #4669, opened 2020-12-24
[script] [multi] Port 'multi' script for Lich built-in repo to GitHub

KatoakDR:
### Background
* This script is downloadable from Lich's built-in repository but isn't part of our GitHub cannon.
* People use the `multi` script and it's a good candidate to begin managing in version control.

### Changes
* Code is inside of a class now, aptly called `Multi`
* If one of the things you want executed is a script, will wait for that script to complete before moving on to next item in your commands list

### DIFF
--- a/multi.lic
+++ b/multi.lic
@@ -1,0 +1,64 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#multi
+=end
+
+custom_require.call(%w[common])
+
+class Multi
+  include DRC
+
+  def initialize(args = nil)
+    if args.nil? || args.empty?
+      show_usage
+    elsif args[0] == "help"
+      show_help
+    else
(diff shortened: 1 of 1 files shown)


multi.lic
```diff
@@ -0,0 +1,64 @@
...
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#multi
+=end
+
+custom_require.call(%w[common])
+
+class Multi
+  include DRC
+
+  def initialize(args = nil)
+    if args.nil? || args.empty?
```
rpherbig, line 11: ❓ Would you be willing to use our arg system?

KatoakDR: I'll look into it more. Since they were complex arguments I didn't tackle that bringing the script over, but worth a shot. Thanks

PR #4671, opened 2020-12-27
Adding in a setting to never blind pick, it's annoying many people.

jayRyan24:


profiles/base.yaml
```diff
@@ -817,6 +817,7 @@ consumable_lockboxes:
 
 # pick.lic settings for live boxes
 # https://elanthipedia.play.net/Lich_script_repository#pick
+never_blind_pick: false
 stop_pick_on_mindlock: true
 pick_blind: false
```
KatoakDR, line 822: Could we leverage the existing `pick_blind: false` setting instead of add a new setting?

pick.lic
```diff
@@ -344,7 +345,11 @@ class LockPicker
     when *@pick_quick
       pick_speed(box, 'quick')
     when *@pick_blind
-      pick_speed(box, 'blind')
+      if @never_blind_pick
+        pick_speed(box, 'quick')
```
KatoakDR, line 349: This new setting can conflict with the existing `pick_blind: true` because the speed is overwritten in the `pick_speed(..)` method itself. I vote for modifying the `pick_speed(..)` method to not do blind if `pick_blind: false`, so moving your logic a bit.

Original logic:
```ruby
  def pick_speed(box, speed)
    if @settings.pick_blind && DRSkill.getxp('Locksmithing') < 34
      @original_speed = speed != 'blind' ? speed : nil
      speed = 'blind'
    end
    # .. snip ..
  end
```

Proposed change: something like this
```ruby
  def pick_speed(box, speed)
    # If we don't want to pick blind, throttle back.
    if @settings.pick_blind == false && speed == 'blind'
      speed = 'quick'
    end
    # If we want to pick blind, speed it up.
    if @settings.pick_blind && DRSkill.getxp('Locksmithing') < 34
      @original_speed = speed != 'blind' ? speed : nil
      speed = 'blind'
    end
    waitrt?
    # .. snip ..
  end
```

profiles/base.yaml
jayRyan24: No, this should be "always_blind_pick".

pick.lic
jayRyan24: Hm, I didn't see this. I'll need to look into this, since blind picking is occurring with blind pick set to false. Thanks!

jayRyan24: So, pick_blind is forcing the system to _always_ pick blind (and should frankly be renamed to `always_pick_blind`. So, my change is saying "even if my skill is able to blind pick this, never do it". Am I missing something here?

KatoakDR: > So, pick_blind is forcing the system to always pick blind (and should frankly be renamed to always_pick_blind)

Yep

> So, my change is saying "even if my skill is able to blind pick this, never do it". Am I missing something here?

To avoid having to mutually exclusive flags for the script, my suggestion is to have logic one way or the other based on the value of a single flag. 

For example, what should happen in this scenario?
pick_blind: true
never_blind_pick: true

Rather than have that ambiguous scenario, modify the script such that `pick_blind: true` always tries to do blind and `pick_blind: false` never attempts blind (will use quick, normal, careful).

jayRyan24: This doesn't work. Sometimes people don't want it to always blind pick, but do want it to when capable. You are missing a logic gate.

jayRyan24: If you can show me how to implement this with 1 setting for the same results, I will do so, but I don't see it.

KatoakDR: > Sometimes people don't want it to always blind pick, but do want it to when capable

That's fair

> If you can show me how to implement this with 1 setting for the same results, I will do so, but I don't see it.

What's rolling around in my head now to support the 3 logic gates is continue to use the one existing setting, `pick_blind` but modify the script to support a third new value, but also be backwards compatible with the old true or false values.

For example, legacy use of the variable will be:
```yaml
pick_blind: true    # always pick blind, live dangerously :)
pick_blind: false   # pick blind when it's "safe" to do so
```
New use of the variable will be:
```yaml
pick_blind: always    # same as legacy value 'true'
pick_blind: capable  # same as legacy value 'false'
pick_blind: never      # new value: never pick blind. ever. fastest possible will be 'quick'
```
During script init, would read the yaml value and if it's true or false then map those to the new values, else use the value as-is. If the configured value isn't one of the allowed values, print warning and fallback to a safe default of "capable" or "never".

Thoughts?

jayRyan24: Your asking for a logical rewrite with a check for deprecated code to be hard coded into the system. Also, 'capable' as a setting seems super ambiguous. I will rename pick_blind to `always_pick_blind` and make that have a deprecation check, but would rather not do all this rigmarole for something so simple.

jayRyan24: If that's not good enough, I will simply close this and ignore it.

KatoakDR (approved):
Giving my approval to clear the air that I'm onboard with your revised approach since I was advocating for something different earlier. Thanks!

profiles/base.yaml
```diff
@@ -817,8 +817,12 @@ consumable_lockboxes:
 
 # pick.lic settings for live boxes
 # https://elanthipedia.play.net/Lich_script_repository#pick
+# Set to false if you want to pick all boxes ignoring exp mind states.
 stop_pick_on_mindlock: true
-pick_blind: false
```
rpherbig, line 821: Removing `pick_blind` from `base.yaml` means a bunch of YAMLs checked into `/profiles` suddenly become invalid. That's why the TravisCI build is failing.

rpherbig:
✔️ Code changes look good, ready to merge once the build passes.

profiles/base.yaml
KatoakDR: Could the test script be updated in this PR to handle the renamed setting?

`dr-scripts/test/test_valid_yaml.rb:32`

jayRyan24:
@rpherbig Should be good to go, thank you.

PR #4674, opened 2020-12-28
Adding a pray-chadatru script for t2 use

jayRyan24:
For use with t2. I will move the core as a call to crossing-training after this is merged.


pray-chadatru.lic
```diff
@@ -0,0 +1,37 @@
+custom_require.call(%w[common drinfomon common-travel])
```
rpherbig, line 1: 💡 Elanthipedia documentation

PR #4684, opened 2021-01-03
Adding empathylink to repo.

Dartellum:
Adding empathylink to repo. 

Added `noheal_empathylink:` for the list of people not to heal.

### DIFF
--- a/empathylink.lic
+++ b/empathylink.lic
@@ -9,10 +9,11 @@ class Empathylink
     if UserVars.empathylink.nil?
       UserVars.empathylink=Hash.new
     end
-    populate_noheal_list
     settings = get_settings
     @buddies = settings.hunting_buddies
     @use_hodierna = settings.has_hodierna
+    noheal = settings.noheal_empathylink
+    populate_noheal_list(noheal)
     in_room = DRRoom.pcs
     while true
       in_room.each do | target |
@@ -25,11 +26,12 @@ class Empathylink
     end
   end
 
-  def populate_noheal_list
+  def populate_noheal_list(noheal)
     ### This setting is to never heal the people listed here even if on
     ### hunting_buddies list.
-    #UserVars.empathylink['xxxx'] = 2496928411
-    #UserVars.empathylink['xxxx'] = 2496928411
+    noheal.each do | no_heal |
+      UserVars.empathylink[no_heal] = 2496928411
+    end
   end
 
   def should_heal?(target)
--- a/empathylink.lic
+++ b/empathylink.lic
@@ -1,0 +1,118 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#empathylink
+=end
+
+custom_require.call(%w[common drinfomon equipmanager])
+
+class Empathylink
+  def initialize
+    if UserVars.empathylink.nil?
+      UserVars.empathylink=Hash.new
+    end
+    populate_noheal_list
+    settings = get_settings
+    @buddies = settings.hunting_buddies
+    @use_hodierna = DRC.get_settings.hashodierna
+    in_room = DRRoom.pcs
+    while true
+      in_room.each do | target |
+        if shouldheal(target)
+          touch(target)
+        end
+      end
+      pause 0.5
+      in_room = DRRoom.pcs
+    end
+  end
+
+  def populate_noheal_list
+    #UserVars.empathylink['xxxx'] = 2496928411
+    #UserVars.empathylink['xxxx'] = 2496928411
+  end
+
+  def shouldheal(target)
+    now = Time.now.to_i
+    healTime = UserVars.empathylink[target]
+    if healTime.nil? or now > healTime and Script.running?('combat-trainer')
+      pause 5
+      if DRRoom.pcs.include?(target) and  @buddies.include?(target)
+        #echo "Should heal #{target}"
+        return true
+      end
+    end
+    #echo "Shouldn't heal #{target}"
+    return false
+  end
+
+  def touch(target)
+    pause_script('combat-trainer') if Script.running?('combat-trainer')
+    waitrt?
+    case DRC.bput("touch #{target}",
+        "You touch",
+        "You lay your hand on",
+        "Touch what",
+        "I could not find",
+        "You rest your hand on")
+    when 'You touch','You lay your hand on', 'You rest your hand on'
+      if @use_hodierna == true
+        hodierna_heal_patient(target)
+      else
+        heal_patient(target)
+      end
+    when 'Touch what', 'I could not find'
+      echo "Failed to establish link with #{target}"
+      UserVars.empathylink[target] = Time.now.to_i + 1800 # Put them on hold for 30m
+    end
+  end
+
+  def hodierna_heal_patient(target)
+    echo "In hodierna healing patient"
+    case DRC.bput("link #{target} persistent",
+      'gradually strengthening the fragile diagnostic link',
+      'I could not find',
+      'You have not established an empathic link',
+      'You already have a persistent empathic link')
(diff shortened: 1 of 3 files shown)


empathylink.lic
```diff
@@ -0,0 +1,118 @@
...
+          touch(target)
+        end
+      end
+      pause 0.5
+      in_room = DRRoom.pcs
+    end
+  end
+
+  def populate_noheal_list
+    #UserVars.empathylink['xxxx'] = 2496928411
```
rpherbig, line 29: After reading through the script, I think the value of this hash is the next time that person can be checked. I read 'noheal' as 'people to never heal'. Do you think this could be confusing to users of this script?

empathylink.lic
```diff
@@ -0,0 +1,118 @@
...
+
+class Empathylink
+  def initialize
+    if UserVars.empathylink.nil?
+      UserVars.empathylink=Hash.new
+    end
+    populate_noheal_list
+    settings = get_settings
+    @buddies = settings.hunting_buddies
+    @use_hodierna = DRC.get_settings.hashodierna
```
rpherbig, line 15: If this is a new setting, it needs to be in `base.yaml`. I would also suggest renaming it to `has_hodierna`.

Would it be possible to do a room check or something instead?

rpherbig, line 15: This should reuse `settings`, not make another call to `get_setttings`

empathylink.lic
```diff
@@ -0,0 +1,118 @@
...
+      in_room = DRRoom.pcs
+    end
+  end
+
+  def populate_noheal_list
+    #UserVars.empathylink['xxxx'] = 2496928411
+    #UserVars.empathylink['xxxx'] = 2496928411
+  end
+
+  def shouldheal(target)
```
rpherbig, line 33: What do you think about renaming this to `should_heal?` (note the `_` and `?`)?

empathylink.lic
```diff
@@ -0,0 +1,118 @@
...
+        #echo "Should heal #{target}"
+        return true
+      end
+    end
+    #echo "Shouldn't heal #{target}"
+    return false
+  end
+
+  def touch(target)
+    pause_script('combat-trainer') if Script.running?('combat-trainer')
```
rpherbig, line 48: I highly recommend scoping the `unpause_script` to the same function as the `pause_script`. If I'm reading this right, `combat-trainer` could be paused here and never unpaused if we enter the case on line 62.

rpherbig:
How does this script differ from `healer` or `smarttransfer`?

empathylink.lic
```diff
@@ -0,0 +1,118 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#empathylink
```
rpherbig, line 2: How does this script differ from `healer` or `smarttransfer`?

Dartellum: Smarttransfer and healer are not setup to work in combat. This script creates links with people in your hunting-buddies list to link with, with Hodierna, to keep the buddies wound free.

empathylink.lic
Dartellum: All the time Ceindrech has used this the script never stayed paused. I will look into it though.

empathylink.lic
Dartellum: I can do that.

empathylink.lic
Dartellum: Right.

Dartellum: Right for adding to base.yaml. The @buddies pulls all your buddies in and then it only performs the link on those in the room.

empathylink.lic
Dartellum: I commented out the names that were in there and left them as an example. So, if you have a name here to never heal and the name was on your buddies list, no healing would occur.

Dartellum: Added comment for clarification in script.

empathylink.lic
Dartellum: Changed.

Dartellum: Added to base.

:question: Would it be possible to do a room check or something instead? It does a room check and compares room PCs to buddies list to touch.

empathylink.lic
Dartellum: Put unpause in same def as pause. I also made def p_script and u_script to call for the pauses - in case I still need to add them to other defs.

Dartellum:
Let me have Ceindrech run this for a bit.

profiles/base.yaml
```diff
@@ -103,6 +103,8 @@ dance_actions:
 - weave
 # how many times to retry a failed analyze when using Analyze in training_abilities
 combat_analyze_retry_count: 3
+# Empath setting for using script empathylink script for combat.
```
rpherbig, line 106: 💡 Does this need to be clearer on how true vs false alter the script's behavior, or would that be intuitive for empath players?

rpherbig:
> Let me have Ceindrech run this for a bit.

✔️ Code looks fine, let me know when you're ready for this to be merged.

profiles/base.yaml
Dartellum: I can make clearer. :)

empathylink.lic
```diff
@@ -0,0 +1,130 @@
...
+=end
+
+custom_require.call(%w[common drinfomon equipmanager])
+
+class Empathylink
+  def initialize
+    if UserVars.empathylink.nil?
+      UserVars.empathylink=Hash.new
+    end
+    populate_noheal_list
```
jandersson, line 12: What is this? I couldn't find it in the repository. It's going to raise an exception.

empathylink.lic
```diff
@@ -0,0 +1,130 @@
...
+    settings = get_settings
+    @buddies = settings.hunting_buddies
+    @use_hodierna = settings.has_hodierna
+    in_room = DRRoom.pcs
+    while true
```
jandersson, line 17: This is going to loop forever as far as i can tell. Do you want this? **documentation nudge**

empathylink.lic
Dartellum: populate_noheal_list is a function just down from there, line 28.

empathylink.lic
jandersson: I like to use the pull request description as a first draft of documentation - you can write it there if inclined. Highly recommended if you respect reviewers time.

Dartellum: Yes, we want it to loop because you start it and let it run. Then, when hunting, it will check if buddies are in the room, make a link, and start the healing.

jandersson: Do you want it to check if leaves the room or anything? seems risky if any random script triggers that causes you to move to the bank or somesuch.

Dartellum: Although, I am going to change the no heal list from populating here to inside your yaml.

Dartellum:
This PR is ready.

profiles/base-empty.yaml
```diff
@@ -96,4 +96,5 @@ empty_values:
   hollow_eve_loot_container: []
   wands: []
   wand_watcher_no_use_scripts: []
-  wand_watcher_no_use_rooms: []
\ No newline at end of file
+  wand_watcher_no_use_rooms: []
+  noheal_empathylink: []
```
rpherbig, line 100: This should _also_ be in `base.yaml`

Dartellum: I think I did put in base but forgot to add to her. Let me check. :smile:

Dartellum: Okay, added and committed to the files.

rpherbig:
There's a merge conflict that needs resolved before I can merge the PR

Dartellum:
> There's a merge conflict that needs resolved before I can merge the PR

Fixed. Second time was the charm. :smile:

PR #4679, opened 2020-12-31
[script] [spellmonitor] consolidate logic and fix for barbarians

KatoakDR:
### Background
* While playing a barbarian without the [Powermonger](https://elanthipedia.play.net/Powermonger) mastery, noticed that `DRSpells.active_spells` was empty despite having multiple abilities and buffs on.
* Prompted [this](https://discord.com/channels/745675889622384681/745675890242879671/794012967652950017) Discord question.
* Which led to [this](https://discord.com/channels/745675889622384681/745696628714897508/794040128308903977) Discord journey.
* Upon review, the regular expressions for the barbarian and non-barbarian code paths weren't really that different and the XML stream received as a Barbarian, Warrior Mage, Paladin, Moon Mage, and Ranger -- with and without cyclics, and with and without buffs from wands or self-cast buffs -- were consistent.

### Changes
* Consolidate the logic for parsing spells -- there are no longer guild specific code paths
* Use named groups in regular expressions to simplify extracting matches (e.g. spell name and duration)
* Handle oddity where DragonRealms uses unicode codepoint 32 as "whitespace" before/within/after spell names and could not be trimmed via Ruby's [String.strip](https://ruby-doc.org/core-2.5.5/String.html#method-i-strip) method
* Document the XML data that's being parsed and scenarios that each regular expression is matching for

PR #4693, opened 2021-01-06
[script] [spellmonitor] Fix: add parens to support OR expression

KatoakDR:
### Background
* The recent update to spellmonitor per https://github.com/rpherbig/dr-scripts/pull/4679 introduced a bug for Clerics using the Osrel Meraud spell -- the regex was not performing an OR between `indefinite|om` because the parens in that expression were actually escaped to match literally, not as a "either this or that" expression.
* This caused combat-trainer to repeatedly try to cast buffs because it didn't realize the spells were active.

### Changes
* Add () around `indefinite|om` regex to be an OR expression

## Tests

_[Koujia](https://discord.com/channels/745675889622384681/745676326190579843/796257768154726432) in the Lich discord tested the change._

```
--- Lich: exec1 active.

[exec1: {"Osrel Meraud"=>75, "Benediction"=>1000, "Shield of Light"=>1000, "Minor Physical Protection"=>1000, "Persistence of Mana"=>1000}]
```

### DIFF
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -127,7 +127,7 @@ spell_action = proc do |server_string|
       # Spell fading away
       spell = Regexp.last_match[:spell]
       duration = 0
-    when /(?<spell>[^<>]+?)\s+\((?:indefinite|om)\)/i
+    when /(?<spell>[^<>]+?)\s+\((?<duration>indefinite|om)\)/i
       # Cyclic spell or Osrel Meraud cyclic spell
       spell = Regexp.last_match[:spell]
       duration = UNKNOWN_DURATION
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -127,7 +127,7 @@ spell_action = proc do |server_string|
       # Spell fading away
       spell = Regexp.last_match[:spell]
       duration = 0
-    when /(?<spell>[^<>]+?)\s+\(indefinite|OM\)/i
+    when /(?<spell>[^<>]+?)\s+\((?:indefinite|om)\)/i
       # Cyclic spell or Osrel Meraud cyclic spell
       spell = Regexp.last_match[:spell]
       duration = UNKNOWN_DURATION


spellmonitor.lic
```diff
@@ -127,7 +127,7 @@ spell_action = proc do |server_string|
       # Spell fading away
       spell = Regexp.last_match[:spell]
       duration = 0
-    when /(?<spell>[^<>]+?)\s+\(indefinite|OM\)/i
+    when /(?<spell>[^<>]+?)\s+\((?:indefinite|om)\)/i
```
KatoakDR, line 130: I mistakenly read `\(` as the start of a regular `(` but, in fact, that was an escaped literal parenthesis, not the start of regex group. This fix puts `indefinite|om` inside a regex group so the `|` is treated as logical OR.

spellmonitor.lic
```diff
@@ -127,7 +127,7 @@ spell_action = proc do |server_string|
       # Spell fading away
       spell = Regexp.last_match[:spell]
       duration = 0
-    when /(?<spell>[^<>]+?)\s+\((?:indefinite|om)\)/i
+    when /(?<spell>[^<>]+?)\s+\((?<duration>indefinite|om)\)/i
```
KatoakDR: I mistakenly read `\(` as the start of a regular `(` but, in fact, that was an escaped literal parenthesis, not the start of regex group. This fix puts `indefinite|om` inside a regex group so the `|` is treated as logical OR.

spellmonitor.lic
```diff
@@ -127,7 +127,7 @@ spell_action = proc do |server_string|
       # Spell fading away
       spell = Regexp.last_match[:spell]
       duration = 0
-    when /(?<spell>[^<>]+?)\s+\(indefinite|OM\)/i
+    when /(?<spell>[^<>]+?)\s+\((?<duration>indefinite|om)\)/i
```
KatoakDR, line 130: I mistakenly read `\(` as the start of a regular `(` but, in fact, that was an escaped literal parenthesis, not the start of regex group. This fix puts `indefinite|om` inside a regex group so the `|` is treated as logical OR.

PR #4696, opened 2021-01-07
Revert "Stage 1 of runestones working with buff and combat-trainer."

Dartellum:
Reverts rpherbig/dr-scripts#4695 

@rpherbig  Could you revert https://github.com/rpherbig/dr-scripts/pull/4695? Two reasons: First, without the updated combat-trainer it is not preparing properly, and two, seems some are getting into the else match with crossing-training and I am not sure how (line 117-121 common-arcana.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -109,31 +109,22 @@ module DRCA
     end
   end
 
-  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false, runestone_name = '', runestone_tm = false)
+  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false)
     return false unless abbrev
 
     DRC.bput('prep symb', 'You recall the exact details of the', 'But you\'ve already prepared') if symbiosis
 
-    if runestone_name.nil?
-      match = DRC.bput("#{command} #{abbrev} #{mana}", get_data('spells').prep_messages)
-    else
-      match = DRC.bput("#{command} my #{runestone_name}", get_data('spells').invoke_messages)
-    end
+    match = DRC.bput("#{command} #{abbrev} #{mana}", get_data('spells').prep_messages)
     case match
     when 'Your desire to prepare this offensive spell suddenly slips away'
       pause 1
-      return prepare?(abbrev, mana, symbiosis, command, tattoo_tm, runestone_name, runestone_tm)
+      return prepare?(abbrev, mana, symbiosis, command, tattoo_tm)
     when 'Something in the area interferes with your spell preparations'
       DRC.bput('rel symb', 'You release the', 'But you haven\'t') if symbiosis
       return false
-    when 'Well, that was fun'
-      DRCI.dispose_trash(runestone_name)
-      return false
-    when 'You\'ll have to hold it'
-      return false
     end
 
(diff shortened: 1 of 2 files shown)


common-arcana.lic
```diff
@@ -109,31 +109,22 @@ module DRCA
     end
   end
 
-  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false, runestone_name = '', runestone_tm = false)
+  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false)
     return false unless abbrev
 
     DRC.bput('prep symb', 'You recall the exact details of the', 'But you\'ve already prepared') if symbiosis
 
-    if runestone_name.nil?
```
rpherbig, line 117: I think your problem is this check.

```
def foo(a = '')
	puts a.nil?
end
 
foo(nil)
foo()
```

> true
> false

common-arcana.lic
```diff
@@ -147,7 +138,7 @@ module DRCA
     command = spell['prep'] if spell['prep']
     command = spell['prep_type'] if spell['prep_type']
 
-    return unless prepare?(spell['abbrev'], spell['mana'], spell['symbiosis'], command, tattoo_tm = false, spell['runestone_name'], spell['runestone_tm'])
```
rpherbig, line 150: I also just noticed that you have `tattoo_tm = false`. It's probably safer to just use `false`.

common-arcana.lic
```diff
@@ -503,14 +475,7 @@ module DRCA
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    if data['runestone_name']
-      if !prepare_to_cast_runestone?(data, settings)
-        return
-      end
-    end
-
-    return unless prepare?(data['abbrev'], data['mana'], data['symbiosis'], command, tattoo_tm = false, data['runestone_name'], data['runestone_tm'])
```
rpherbig, line 512: I also just noticed that you have `tattoo_tm = false`. The original used `data['tattoo_tm']`. Why the change?

common-arcana.lic
```diff
@@ -676,7 +641,7 @@ module DRCA
     command = data['prep'] if data['prep']
     command = data['prep_type'] if data['prep_type']
 
-    prepare?(data['abbrev'], data['mana'], data['symbiosis'], command, tattoo_tm = false, data['runestone_name'], data['runestone_tm'])
```
rpherbig, line 679: I also just noticed that you have `tattoo_tm = false`. It's probably safer to just use `false`.

Dartellum: I think I copied it from line 112 because I accidentally deleted it here on one of the versions.

Dartellum: Also made this `data['tattoo']`.

common-arcana.lic
Dartellum: I think I copied it from line 112 because I accidentally deleted it here on one of the versions.

common-arcana.lic
Dartellum: Also made it `spell['tattoo_tm'].

common-arcana.lic
Dartellum: Make `runestone_name = nil` ?

rpherbig:
I think that will fix the problem you described

PR #4700, opened 2021-01-08
Add new Silverfish Tiers to P1

Hiinky:


data/base-hunting.yaml
```diff
@@ -407,6 +407,31 @@ hunting_zones:
   - 1446
   - 1447
   - 1448
+  silverfish:                                                             
+  # https://elanthipedia.play.net/Silverfish                              10-32
+  # Do Not Use 6079.  It is a terrible quadruplicate.
```
jandersson, line 412: What is a terrible quadrublicate?

Hiinky: theres four rooms that are dups in mapdb and cannot be mapped better than what I did.  should never hunt there.  it's also noted on a new map image i have pending

jandersson: Just say why? Also this is a data file, it feels wierd to leave notes like this in my opinion

Hiinky: Updated

PR #4727, opened 2021-01-29
validate - Added missing training_abilities

BinuDR:
App Bundle and Tessera weren't listed in the section that checks against the training_abilities in your yaml.  I've added them so that it doesn't give an error when you validate yamls containing them.


KatoakDR:
Good catch, @BinuDR, thanks

PR #4473, opened 2020-08-30
[script] [common-summoning] Update logic to hold moon weapons

KatoakDR:
### Changes
* Changes ambiguous use of `get moon` with `DRCMM.hold_moon_weapon?`
* Changes ambiguous use of `glance moon` with checking what's held in your hands

Bulk of the logic is to ensure the regex patterns match to actual moon weapons and not get confused by items whose nouns are or start with the word "moon", like "a trio of moons" or "moonsilk shirt". I'm personally invested in these changes because my moon mage wears both those items, haha.

PR #4732, opened 2021-02-12
Update combat-trainer.lic

Akhorishaan:
Corrected a variable check to make moonblades work again.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -193,7 +193,7 @@ class SetupProcess
 
     game_state.reset_action_count
 
-    if DRStats.moon_mage? && !game_state.summoned_weapons.empty? && moon_used_to_summon_weapon.nil?
+    if DRStats.moon_mage? && !game_state.summoned_weapons.empty? && !moon_used_to_summon_weapon.nil?
       echo('skipping summoned weapons because no moonblade available') if $debug_mode_ct
       weapon_training = weapon_training.reject { |skill, _| game_state.summoned_info(skill) }
     end


rpherbig:
This conditional hasn't changed since May 2019. I'm curious how we only found/fixed this now.

@jandersson @Sarvatt @jayRyan24 any thoughts on this change?

cc @KatoakDR @BinuDR

KatoakDR:
Hi @Akhorishaan, thanks for contributing to the Lich script repository. Could you share the pertinent pieces of your logs where combat-trainer is skipping using your summoned weapon? And, would you please share a copy of your character's yaml file, it can aid with us testing the change and seeing if we can reproduce the problem.

Thanks!

Akhorishaan:
Of course, of course.  This probably wasn't fixed for so long because it's a niche application and nobody uses moonblade but me and maybe two other people, so there's no pressure to fix the problem.

KatoakDR:
Thanks, we'll take a look and follow up

KatoakDR:
> noticed that it was skipping summoned weapons every time it looked for a new weapon to train

Yes, I'm seeing this, too. Investigating...

Akhorishaan:
Oh yeah, I see that now, it's only nil when it isn't in your hand, but if it is in your hand then it is populated with the moon.  Which explains why if you grab the moonblade manually with the original version of the script, it'll go through a cycle of using each type before forgetting it exists again.  May just have to add a separate variable that doesn't change based on what is in your hands, as you mentioned.

KatoakDR:
> May just have to add a separate variable that doesn't change based on what is in your hands, as you mentioned. Like just something that gets set once whenever a new moonblade is summoned.

Yep! Trying that out right now :)

KatoakDR:
I got it working by tracking its own variable for the last moon used. Polishing up the changes and will submit a PR. Thanks

rpherbig:
The logic in `moon_used_to_summon_weapon` changed only a few months ago in PR #4473 - maybe that's related

KatoakDR:
> The logic in `moon_used_to_summon_weapon` changed only a few months ago in PR #4473 - maybe that's related

Good catch, and apologies that I introduced the issue.

At the time I submitted that earlier PR, we weren't sure how many name variants a moon* weapon could have -- but seems it's either a `moonblade` or a `moonstaff` so we may be able to revert that method to not rely on looking at someting **in your hands** but rather if the game is able to glance at a `moon(blade|staff)`. The original "glance moon" was insufficient because wearing jewelry or other items whose noun is "moon" would cause issues.

I went to bed last night before submitting my PR that tracks the moon that the moonblade spell was cast on. Might be useful to have both changes?

KatoakDR:
OK, before we modify `combat-trainer`, I've submitted enhancements to `common-summoning` and `common-moonmage` to better handle the detection and holding of moon weapons. Once those are merged let's continue to test combat-trainer as-is to see if that resolves our issue. If not, we can move forward with adding an internal variable to `game_state` to keep track of the moon used when moonblade spell is cast. Thanks!

* https://github.com/rpherbig/dr-scripts/pull/4744 - fixes bug where you had to be holding the moon weapon to detect the moon used
* https://github.com/rpherbig/dr-scripts/pull/4745 - simplifies logic for taking hold of a moon weapon

KatoakDR:
@rpherbig This PR  can be closed as it was superceded by the following merged PRs:
* #4744 - fixes bug where you had to be holding the moon weapon to detect the moon used
* #4745 - simplifies logic for taking hold of a moon weapon
* #4746 - adds clarifying comments to combat-trainer when using moonblade spell

Thanks!

PR #4736, opened 2021-02-13
Stage 2b of runestones working with buff and combat-trainer.

Dartellum:
Merge a few days after Stage-1b-of-combat-trainer-and-buff-runestones.

Runestone casting within combat-trainer.


Dartellum:
This PR is ready to go. I only heard of two small challenges for the first stage, but they were due to multiple characters logged in and logging in an updated common-arcana.

PR #4737, opened 2021-02-13
Stage 1b of runestones working with buff and combat-trainer.

Dartellum:
Merge this one first.

Common-arcana and base-spells updates to make runestone casting work from buff and combat-trainer.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -109,22 +109,31 @@ module DRCA
     end
   end
 
-  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false)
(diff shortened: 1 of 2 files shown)


rpherbig:
@Dartellum see #4814 and #4817 - `common-arcana` needs to `custom_require` common-items, but that introduces a circular dependency (common-arcana -> common-items -> common-travel -> common->arcana).

We need a way to address this, perhaps pulling the runestone code into a different common file?

wstampley:
Some thoughts on the above - 

Common Items uses Common Travel for the 'go here and buy / order / trash' something functions; I wonder if all the shopping related stuff could be moved somewhere else to make common items a 'near-bottom' (`common` only) dependency that only focuses on personal item manipulation.

Common Travel uses Common Arcana for the 'find a room with good mana' function, which seems like an okay fit, but could maybe just be duplicated code to break that link.

PR #4738, opened 2021-02-14
Arrows script updates

Dartellum:
Fixed some messaging due to output change. Other fixes added: check glue left and buy more if under 5 uses, fixed a message match string, and added stow_item(part) to store the extra arrowheads in hand if you have more than 5 in the stack.


Dartellum:
Made draft because I added a continue function. I need to test it.

PR #4741, opened 2021-02-15
create_remedies.lic

Dartellum:
This script uses the new alchemy forage and process method then will create the remedy. All one needs to do is enter the product - ;create_remedies "limb ungent" "pewter ingot" tools. The script will do the rest. The pewter ingot is an optional catalyst to use and tools will retrieve and store the tools with the Clerk.

This script will not purchase the herbs from the Society. The assumption in using this script is you have adequate Outdoorsmanship skill to forage the herbs and the remedy is not for `workorders` or skill increase only, aka `;craft`.

### DIFF
--- a/create_remedies.lic
+++ b/create_remedies.lic
@@ -1,0 +1,59 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#create_remedies
+=end
+
+custom_require.call(%w[common common-crafting])
+
+class Create_remedies
+  include DRC
+  include DRCC
+
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'target', regex: /\w+/, description: 'Recipe.' },
+        { name: 'catalyst', regex: /\w+/i, optional: true, description: 'Catalyst to use. Default is coal.' },
+        { name: 'tools', regex: /tools/i, optional: true },
+        { name: 'debug', regex: /debug/i, optional: true }
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+    settings = get_settings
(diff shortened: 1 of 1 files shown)


create_remedies.lic
```diff
@@ -0,0 +1,59 @@
...
+    get_store_tools("store") if args.tools
+  end
+
+  def get_prep_herb(herb)
+    wait_for_script_to_complete('alchemy', [herb, 'forage', 25])
+    wait_for_script_to_complete('alchemy', [herb, 'prepare'])
+  end
+
+  def get_store_tools(option)
+    echo("Option for tools is: {option}.") if debug
```
rpherbig, line 54: I don't think `debug` is in scope inside this function

create_remedies.lic
```diff
@@ -0,0 +1,59 @@
...
+    # Store tools?
+    get_store_tools("store") if args.tools
+  end
+
+  def get_prep_herb(herb)
+    wait_for_script_to_complete('alchemy', [herb, 'forage', 25])
+    wait_for_script_to_complete('alchemy', [herb, 'prepare'])
+  end
+
+  def get_store_tools(option)
```
rpherbig, line 53: Why not have two helpers?

create_remedies.lic
```diff
@@ -0,0 +1,59 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#create_remedies
+=end
+
+custom_require.call(%w[common common-crafting])
+
+class Create_remedies
```
rpherbig, line 7: Ruby convention would be `CreateRemedies`

PR #4748, opened 2021-02-17
Fix to the github comms deprecation.

### LINKED ISSUE #4478: Dependency github communication deprecation
By early march of 2021 github will stop supporting our method of communicating with the dr-scripts repository for retrieval of scripts.

https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

We use this [here](https://github.com/rpherbig/dr-scripts/blob/master/dependency.lic#L451). It can be corrected by removing that and adding it as a header to the request.

```
request = Net::HTTP::Get.new(uri.request_uri)
request['Authorization'] = "token #{@git_token}"
```

Should do the trick, but should of course be tested and verified.

jayRyan24:
Tested the fix with a clean install of Lich.

1. Made change to dependency.lic within `make_request` method.
2. Commented out the update for dependency.lic from within itself...
```
  def download_script(filename, force = false)
    return if filename.nil? || filename.empty?
    echo("downloading:#{filename}") if @debug
    info = get_file_status(filename)
    return unless info
    return if @developer
    return if get_versions[filename] == info['sha'] && !force
    if filename == 'dependency.lic'
      echo "Skipping dependency.lic update..."
      return
    end
```
3. The above step requirement actually confirms it works, but moved on for further testing.
4.  Made changes to the top of a few scripts like combat-trainer. 
5. Re-ran dependency.lic, files updated and were overwritten successfully. 
6. The args and uri.query seem to be unnecessary, removed.

Let me know if you all believe any more testing is required.

Resolves #4478

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -448,13 +448,12 @@ class ScriptManager
 
   def make_request(raw_uri)
     uri = URI.parse(raw_uri)
-    args = { access_token: @git_token }
-    uri.query = URI.encode_www_form(args)
     http = Net::HTTP.new(uri.host, uri.port)
     http.use_ssl = true
     http.verify_mode = OpenSSL::SSL::VERIFY_NONE
 
     request = Net::HTTP::Get.new(uri.request_uri)
+    request['Authorization'] = "token #{@git_token}"
 
     response = http.request(request)
     JSON.parse(response.body)


KatoakDR (approved):
I followed @jayRyan24 test procedures and worked like a champ for me, too. Thanks!

PR #4758, opened 2021-02-18
Bolts creation script new PR.

Dartellum:
This PR replaces 4739 as that PR I accidentally put `shape` into it and could not get it out. All of the updates throughout PR 4739 are in this PR.


Dartellum:
Okay, this one was tested by Kiriawen and is good to go.

PR #4752, opened 2021-02-17
[script] [combat-trainer] Improve barb buff activations

KatoakDR:
### Background
* Supercedes https://github.com/rpherbig/dr-scripts/pull/4667
* If combat trainer fails to activate a barb buff and that buff wasn't already active then the flag `"ap-#{name}-expired"` won't ever reset because the expiry message won't ever be seen. This causes the ability to never be retried for the rest of that hunt.
* Per https://github.com/rpherbig/dr-scripts/pull/4679, detection of active barb buffs can be simplified to checking `DRSpells.active_spells["foo"]`
* Just like mages need a minimum amount of mana/concentration to cast spells or maintain cyclics, so too do barbarians need sufficient inner fire.

### Changes
* Remove code that managed flags to track barb abilities (e.g. `setup_barb_buff_flags`); using `DRSpells.active_spells` now.
* Don't assume the barb ability was activated just because it was attempted to be activated (you may have lacked inner fire at the time or in a location that prevents the ability from being activated, such as trying to meditate in deep water).
* Only try to activate ability if you have sufficient inner fire based on new numeric setting, `barb_buffs_inner_fire_threshold`.
* If activated a meditation, use the pre-existing setting `meditation_pause_timer` to control how many seconds to wait for it to take effect.
* Reduced `meditation_pause_timer` in `base.yaml` to 6 seconds from 25 seconds.
  * As a 30th circle barbarian, I haven't noticed taking more than 6 seconds for Tenacity meditation to take effect.
  * The `SampleBarbarian-setup.yaml` uses 6 seconds, too
  * And `combat-trainer` was only waiting 6 seconds if it activated a meditation, too.
  * Other than the setting being introduced to `combat-trainer` (rather than the pause amount being hard coded), this setting was only used elsewhere in `common-arcana`'s `start_barb_abilities` method, and that method isn't used anywhere in the Lich codebase. Therefore, I don't foresee any risk lowering the default value.

### Config Example
_Added to base.yaml_
```yaml
# The minimum inner fire needed to activate a barb_buff in your buff_nonspells.
# You won't try to activate a barb buff until you have this much inner fire.
# You need some inner fire in reserve to pay the periodic cost of abilities.
barb_buffs_inner_fire_threshold: 15
```
_Updated in base.yaml_
```yaml
# Used to set a pause timer for barbarian meditations.
# This number should go down with a chakrel, circles, and stats...
# When buff script is activating meditations,
# this is how many seconds to wait for it to activate
# before moving on to the next action.
[...]

PR #4759, opened 2021-02-20
Revert "[script] [combat-trainer] Improve barb buff activations"

rpherbig:
Reverts rpherbig/dr-scripts#4752


KatoakDR:
This change was reverted because barbarians without knowledge of the [Flame](https://elanthipedia.play.net/Flame) meditation do not see in their active spells window what abilities they've activated. I honestly thought barbarians always knew, I'd never not ever picked the Flame ability after joining the guild before so didn't realize how intertwined it was with the spell window.

This now explains why combat-trainer was relying on tracking the expiry messages rather than using DRSpells.active_spells.

Not ideal, but this isn't the first nor will it be the last quirk of DR that we code around.

Issue #3141, opened 2018-09-20
Outdoorsmanship keeps removing symbiosis harness


drgabellia:
--- Lich: outdoorsmanship active.

[outdoorsmanship]>release spell

You aren't preparing a spell.

[outdoorsmanship]>release mana

You aren't harnessing any mana.

[outdoorsmanship]>release symb

Are you sure you'd lke to remove the Harness symbiosis from your memory?  Repeat this command in 30 seconds to confirm.

--- Lich: outdoorsmanship has exited.

--- Lich: go2 active.

****Did some climbing here (athletics) in xalas*****

Then.....

[Wilderness, Deep Forest]
The path disappears here amid small bushes and tall trees.  High overhead, a battered hip boot swings casually in the wind apparently caught at the ankle by a snare.  More tripped, but empty snares dangle from various tree limbs, silent testimony to their makers lack of skill.  You also see a mischievous snow white kitten with taffelberry-stained paws.
Obvious paths: southwest.

Room Number: 6154

[go2: travel time: 0:00:02]

--- Lich: go2 has exited.

--- Lich: outdoorsmanship active.

[outdoorsmanship]>release spell

You aren't preparing a spell.

[outdoorsmanship]>release mana

Your body below the waist is dripping with water.  Pools of it are forming all around you.
You feel fully rested.
You aren't harnessing any mana.

[outdoorsmanship]>release symb

You intentionally wipe the Harness symbiosis from your memory.

I really don't want to wipe the Harness symbiosis from my memory...can this be changed so it doesn't happen?

PR #4763, opened 2021-02-23
Fixed magic_cleanup for outdoorsmanship to properly release symbiosis

ubiquitous42:
While using t2 with outdoorsmanship, I noticed that it wasn't releasing symbiosis, and caused me to backfire a few spells.  Easy fix.

### DIFF
--- a/outdoorsmanship.lic
+++ b/outdoorsmanship.lic
@@ -60,6 +60,7 @@ class Outdoorsmanship
 
     bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
     bput('release mana', 'You release all', "You aren't harnessing any mana")
+    bput('release symb', "But you haven't", 'You release', 'Repeat this command')
   end
 end
 


Sarvatt:
I feel like there was a reason I didn't add this, it was releasing learned symbs I believe

Sarvatt:
Yeah there's no way to guarantee you don't call release symb twice before the timeout with how often this script is called and it'll release your symbs completely, much worse than a backfire since there's nothing handling learning a new one

Sarvatt:
maybe track it via a uservar timer across multiple scripts to guard against that

Sarvatt:
This commit just reintroduced this bug https://github.com/rpherbig/dr-scripts/issues/3141#issue-361990390

rpherbig:
I'll revert and add a comment until we have a better approach.

ubiquitous42:
Thanks for the context, @Sarvatt !  Sorry for the conflict, folks.

PR #4762, opened 2021-02-22
[script] [combat-trainer] Improve barb buff activations (round 3)

KatoakDR:
### Dependencies
* ~Depends on #4772~
* ~Depends on #4773~

### Background
* Supercedes #4752
* If combat trainer fails to activate a barb buff and that buff wasn't already active then the flag `"ap-#{name}-expired"` won't ever reset because the expiry message won't ever be seen again. This causes the ability to never be retried for the rest of that hunt.
* Just like mages need a minimum amount of mana/concentration to cast spells or maintain cyclics, so too do barbarians need sufficient inner fire.

### Changes
* Add comments why barb buffs are tracked via Flags and not DRSpells.active_spells (spoiler: it's because barbarians must know Power meditation or Powermonger mastery to see their active abilities...)
* Don't assume the barb ability was activated just because it was attempted to be activated (you may have lacked inner fire at the time or in a location that prevents the ability from being activated, such as trying to meditate in deep water).
* Only try to activate ability if you have sufficient inner fire based on new numeric setting, `barb_buffs_inner_fire_threshold`.
* If activated a meditation, use the pre-existing setting `meditation_pause_timer` to control how many seconds to wait for it to take effect.
* Set `meditation_pause_timer` in `base.yaml` to 20 seconds from 25 seconds.
  * As a 30th circle barbarian, I haven't noticed taking more than 6-10 seconds for Tenacity meditation to take effect.
  * Kali said her young barbarian takes 18 seconds to activate a meditation.
* Add `DRCA` prefix to khri activation methods

### Config Example
_Added to base.yaml_
```yaml
# The minimum inner fire needed to activate a barb_buff in your buff_nonspells.
# You won't try to activate a barb buff until you have this much inner fire.
# You need some inner fire in reserve to pay the periodic cost of abilities.
barb_buffs_inner_fire_threshold: 15
```
_Updated in base.yaml_
```yaml
# Used to set a pause timer for barbarian meditations.
# This number should go down with a chakrel, circles, and stats...
# When buff script is activating meditations,
# this is how many seconds to wait for it to activate
# before moving on to the next action.
meditation_pause_timer: 20
```

## Tests

The following abbreviated log shows that combat-trainer monitors your inner fire level to attempt activations and will re-activate an ability if it expires.

```
[combat-trainer]>form monkey

An interesting fluidity fills your muscles as you assume the movements of an agile monkey!
Roundtime: 2 sec.
> 
[combat-trainer: Activated barb_buff Monkey]

> 
[...]

PR #4772, opened 2021-03-01
[script] [common-arcana] activate barb abilities

KatoakDR:
### Background
* Inspired by testing #4762
* The logic for pausing after activating a meditation was responsibility of code that called `activate_barb_buff?`, which appeared in multiple places and in multiple scripts: combat-trainer and common-arcana

### Changes
* Consolidate the logic for pausing after activating a meditation to the `activate_barb_buff?` method so calling code doesn't need to repeat the logic.
* Update `activate_barb_buff?` to accept an optional second argument, `settings`, to reference if need to pause after activating a meditation.
* Add comments about why the pause is necessary.
* In `start_barb_abilities` method, renamed `skills` variable to `abilities` to disambiguate from experience skills you train.
* Fix bug where the case statement for `ability_data['activated_message']` is not matched because it's value was structured like a regular expression in `data/base-spells.yaml` but was treated literally in the code.

### Pause then wait out any remaining RT
```
> ,e echo DRCA.activate_barb_buff?('Tenacity', { 'meditation_pause_timer' => 5 })

--- Lich: exec1 active.

[exec1]>med tenacity

You must be sitting to properly reinforce your physical resistance.
> 
[exec1]>sit

You sit down.

s> 
[exec1]>med tenacity

You begin to meditate upon the chakrel amulet, your inner fire swelling as you center your mind, body, and spirit.
Roundtime: 9 sec.
s> 

(script pauses for 5 seconds, the value of meditation_pause_timer)

Peering into the flame you visualize swords, teeth and worse all rending your flesh.
s> 
You draw the flames close and bring forth a physical hardening to your body.
s> 
The flame infuses your flesh, reinforcing it against physical damage!
You feel a jolt as your vision snaps shut.
s> 

(script waits for any remaining RT, of which there is 4 seconds)

(then script terminates)

[exec1: true]

--- Lich: exec1 has exited.
```

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -72,35 +72,42 @@ module DRCA
     end
   end
 
-  def start_barb_abilities(skills, settings)
-    skills
+  def start_barb_abilities(abilities, settings)
+    # Note, you must know Power meditation or Powermonger mastery
+    # for your active abilities to be detected by DRSpells.
+    abilities
       .reject { |name| DRSpells.active_spells[name] }
-      .each do |name|
-        activate_barb_buff?(name)
-        waitrt?
-        pause settings.meditation_pause_timer.to_i if get_data('spells').barb_abilities[name]['type'].eql?('meditation')
-        DRC.fix_standing
-      end
+      .each { |name| activate_barb_buff?(name, settings) }
+    DRC.fix_standing
   end
 
-  def activate_barb_buff?(name)
+  def activate_barb_buff?(name, settings = nil)
     ability_data = get_data('spells').barb_abilities[name]
     case DRC.bput(ability_data['start_command'], ability_data['activated_message'], 'But you are already', 'Your inner fire lacks', 'find yourself lacking the inner fire', 'You should stand', 'You must be sitting', 'You must be unengaged', 'While swimming?')
     when 'You must be unengaged'
       DRC.retreat
-      activate_barb_buff?(name)
+      activate_barb_buff?(name, settings)
     when 'You must be sitting'
       DRC.retreat
       case DRC.bput('sit', 'You sit', 'You are already', 'You rise', 'While swimming?')
       when 'While swimming?'
         false # can't sit here, water is too deep
       else
-        activate_barb_buff?(name)
+        activate_barb_buff?(name, settings)
       end
     when 'You should stand'
       DRC.fix_standing
-      activate_barb_buff?(name)
-    when ability_data['activated_message']
+      activate_barb_buff?(name, settings)
+    when /#{ability_data['activated_message']}/
+      # Pause at least for the preferred amount of time
+      # to let the meditation take effect else it may fail.
+      # Note, we use bracket notation to reference settings here
+      # because this method may be invoked with either JSON or an OpenStruct.
+      if ability_data['type'].eql?('meditation') && settings['meditation_pause_timer']
+        pause settings['meditation_pause_timer'].to_i
+      end
+      # Wait for any remaining RT before proceeding
+      waitrt?
       true
     else
       false


common-arcana.lic
```diff
@@ -72,35 +72,42 @@ module DRCA
...
-    skills
+  def start_barb_abilities(abilities, settings)
+    # Note, you must know Power meditation or Powermonger mastery
+    # for your active abilities to be detected by DRSpells.
+    abilities
       .reject { |name| DRSpells.active_spells[name] }
-      .each do |name|
-        activate_barb_buff?(name)
-        waitrt?
-        pause settings.meditation_pause_timer.to_i if get_data('spells').barb_abilities[name]['type'].eql?('meditation')
```
KatoakDR, line 81: This logic is moved to `activate_barb_buff?` method below

common-arcana.lic
```diff
@@ -72,35 +72,42 @@ module DRCA
...
-        DRC.fix_standing
```
KatoakDR, line 82: Moved to end of the method rather than after each activation of an ability

common-arcana.lic
```diff
@@ -72,35 +72,42 @@ module DRCA
...
-      end
+      .each { |name| activate_barb_buff?(name, settings) }
+    DRC.fix_standing
   end
 
-  def activate_barb_buff?(name)
+  def activate_barb_buff?(name, settings = nil)
```
KatoakDR, line 84: Add new, optional argument for backwards compatibility with combat-trainer. Once this PR is merged and then after #4762 then we can decide if this second argument should remain optional or not. As of this PR, the settings is only used if activating a meditation so the script can know how long to pause.

common-arcana.lic
```diff
@@ -72,35 +72,42 @@ module DRCA
...
       else
-        activate_barb_buff?(name)
+        activate_barb_buff?(name, settings)
       end
     when 'You should stand'
       DRC.fix_standing
-      activate_barb_buff?(name)
-    when ability_data['activated_message']
+      activate_barb_buff?(name, settings)
+    when /#{ability_data['activated_message']}/
```
KatoakDR, line 101: Express the match statement as regular expression because in `base-spells.yaml` the activated message is `^You .* to meditate`. The bug was that the case statement was trying to match on that literally. Now it'll match as a regex.

common-arcana.lic
```diff
@@ -72,35 +72,42 @@ module DRCA
...
+      # Pause at least for the preferred amount of time
+      # to let the meditation take effect else it may fail.
+      # Note, we use bracket notation to reference settings here
+      # because this method may be invoked with either JSON or an OpenStruct.
+      if ability_data['type'].eql?('meditation') && settings['meditation_pause_timer']
+        pause settings['meditation_pause_timer'].to_i
+      end
+      # Wait for any remaining RT before proceeding
+      waitrt?
```
KatoakDR, line 110: Consolidated the logic for pausing after activating a meditation to this one method. Calling code, of which there's at least two other references, no longer have to remember to do this step.

common-arcana.lic
rpherbig: Is this desirable? What if you have meditation, something else, meditation. Would you want to stand up in between the two meditations?

common-arcana.lic
rpherbig: Nice catch!

common-arcana.lic
KatoakDR: Resolved per https://github.com/rpherbig/dr-scripts/pull/4772/commits/714857a3b33c36cb86d9209d30d39a7a3f831dd3

common-arcana.lic
```diff
@@ -72,39 +72,48 @@ module DRCA
...
+      activated = true
     else
-      false
+      activated = false
     end
+    DRC.fix_standing
+    activated
```
KatoakDR, line 116: After attempting to activate an ability, always try to stand back up in case you had to sit down

PR #4776, opened 2021-03-03
Small change to make sure we're all awake when we start training!

jayRyan24:


t2.lic
```diff
@@ -20,6 +20,8 @@ class T2
     UserVars.repair_timer_snap ||= Time.now
     UserVars.t2_timers ||= {}
 
+    fput('awaken')
```
KatoakDR, line 23: Oh yeah! I see your `awaken` and raise you a double `sleep`!

```yaml
training_list:
- skill:
  - Perception
  start: 99
  scripts:
  - multi 2,sleep
```

KatoakDR (approved):

t2.lic
KatoakDR: @jayRyan24 This seems to already be done at [line 15](https://github.com/rpherbig/dr-scripts/pull/4776/files#diff-196f691e98c0e3dacb1962cd4b22ff3016aa966350832fd650709251f13cf916R15). This may not be necessary?

jayRyan24: Lord jesus, why the fuck are people complaining about this then... I have a older T2 with this in there I added myself, why are they complaining that T2 is asleep... thanks.

PR #4778, opened 2021-03-03
Base-recipes clean-up and additional recipes added. 

Dartellum:
The goal is to put all current discipline recipes in the file. 

**March 3, 2021** - All Forging recipes are in and confirmed. Eningeering carve and shape recipes confirmed. I am working on adding tinkering recipes and then adding to workorders. Before adding tinkering, I will verify Tailoring and Alchemy recipes. For sanity, all recipes start at the first chapter recipes start in the book and are in page order as of this time. However, it does not appear these change order or has not for me. I added a comment after each recipe indicating page number. This step is the sanity check. :smile: Example showing Chapter comment and then page number on recipe:
```yaml
# Shaping Chapter 10 recipes
- name: a rough wood table #Page 1
  noun: table
  volume: 15
  type: shaping
  work_order: false
  chapter: 10
  part:
  - long pole
  - long pole
  - long pole
  - long pole
 ```

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -1,14 +1,19 @@
 ---
 crafting_recipes:
-- name: a metal ring vest
-  noun: vest
-  volume: 18
+##### Page numbers are as of 2/20/2021 15:48 ET #####
+##################################
+##### Forging section begins #####
+##################################
+# Armorsmithing Chapter 1 recipes
+- name: a metal ring aventail #Page 1
+  noun: aventail
+  volume: 5
   type: armorsmithing
-  work_order: false
+  work_order: true
   chapter: 1
   part:
-  - large padding
-- name: a metal ring mask
+  - small padding
+- name: a metal ring mask #Page 2
   noun: mask
   volume: 3
   type: armorsmithing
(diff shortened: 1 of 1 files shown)


rpherbig:
I don't really have a way to verify a change like this. Maybe have someone spot check by crafting a few random newly added items?

Dartellum:
> I don't really have a way to verify a change like this. Maybe have someone spot check by crafting a few random newly added items?

I have many more to confirm, add, or delete duplicates.

rpherbig:
Three recipes (see #5176) are still duplicated in base-recipes: `some elbow-length gloves` and `a double-wrapped belt` and `a deeply-hooded cloak`. I don't see any reason that these wouldn't still be a problem.

@Dartellum can you confirm that all three of these can be crafted with both cloth and leather?

Dartellum:
> Three recipes (see #5176) are still duplicated in base-recipes: `some elbow-length gloves` and `a double-wrapped belt` and `a deeply-hooded cloak`. I don't see any reason that these wouldn't still be a problem.
> 
> @Dartellum can you confirm that all three of these can be crafted with both cloth and leather?

Yes, all three are both cloth and leather. The cloth recipe is in chapter 2 and the leather recipe is in chapter 7.

rpherbig:
> > Three recipes (see #5176) are still duplicated in base-recipes: `some elbow-length gloves` and `a double-wrapped belt` and `a deeply-hooded cloak`. I don't see any reason that these wouldn't still be a problem.
> > @Dartellum can you confirm that all three of these can be crafted with both cloth and leather?
> 
> Yes, all three are both cloth and leather. The cloth recipe is in chapter 2 and the leather recipe is in chapter 7.

Can you confirm that the crafting scripts support crafting both versions of each of these recipes?

PR #4779, opened 2021-03-05
Added a delay for users to specificy between attempts. 

jayRyan24:
Requested by people.


rpherbig:
Please add `mining_attempt_timer` to `base.yaml`

jayRyan24:
Crap, I had, just forgot to add it with the other file!

PR #4780, opened 2021-03-05
Added an exit of current mine location if returning from safe-room

jayRyan24:
while in danger comes up to the room now having a pc.

Basically, people go heal, snapshot room is taken, they return and begin to mine again, and someone else is there. We should exit mine and move on. This needs testing to make sure mining-buddy recovers correctly.

### DIFF
--- a/mine.lic
+++ b/mine.lic
@@ -56,7 +56,9 @@ class Mine
       store_mining_tool
       wait_for_script_to_complete('safe-room')
       walk_to(snapshot)
-      exit unless DRRoom.npcs.empty?
+      if @mining_skip_populated && !DRRoom.npcs.empty?
+        DRC.message("Returned from healing and another player is in the room. Returning to mining-buddy to move on!")
+      end
       get_mining_tool
     end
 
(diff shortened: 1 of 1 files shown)


rpherbig:
Please add `mining_skip_populated` to `base.yaml`

> This needs testing to make sure mining-buddy recovers correctly.

No further comments, let me know when you feel good about your testing.

jayRyan24:
> Please add `mining_skip_populated` to `base.yaml`
> 
> > This needs testing to make sure mining-buddy recovers correctly.
> 
> No further comments, let me know when you feel good about your testing.

Should be there, this was a previous setting I am just taking advantage of. This is pretty rare situation, we tried to spark it to test, but was difficult without killing him outright. He is currently testing. 

The condition though can be simply tested within the CLI.
```
;e echo true if get_settings.mining_skip_populated && !DRRoom.pcs.empty?
--- Lich: exec1 active.
[exec1: true]
--- Lich: exec1 has exited.
```
With this, you can see it will definitely exit (Had a PC in the room with me). It shouldn't have taken out the mining implement either, so it will just exit and continue on, letting mining-buddy pick up where it handed off the reins to mine.lic.

rpherbig:
Fair enough, I'm good to merge this now or we can wait on testing. What do you think?

jayRyan24:
I'm fine with the merge, if it breaks, which I doubt, it will be so one off, probably won't happen for a while.

PR #4788, opened 2021-03-07
Restock deposit fix - made it so it does not deposit if no purchases were made.

Dartellum:
Behavior as of the last commit added in depositing the money withdrawn from the town you are in. For example, if I have restock setup for smooth rock from Hib, restock would count my arrows and determine if I had the proper number. If not, it will go purchase. However, if I have the proper number, the place deposit is now, restock runs you to Hib to deposit money even if you do not have money from that bank. 

If you have enough items or do not have restock set, no movement should happen. A lot of extra running around is now removed from the script. I was going under the assumption anyone using restock had it as a separate script. I did not see it was called by hunting-buddy. Running to the bank is not a bad thing unless you are a necro or do not have prehunt_buff room set. Hopefully, this explanation is clear.

The deposit is necessary to prevent burden. In testing, I had money left over causing a burden as I went into combat.

### DIFF
--- a/restock.lic
+++ b/restock.lic
@@ -21,6 +21,7 @@ class Restock
   end
 
   def restock_items(item_list, town)
+    purchased = false
     items_to_restock = []
     coin_needed = 0
 
@@ -37,10 +38,13 @@ class Restock
       item['buy_num'] = buy_num
       items_to_restock.push(item)
     end
-    ### The following line is for the bribe to the night attendant to let you in.
-    coin_needed += 2502 if UserVars.sun['night'] && town == /\bShard\b/
-    get_money_from_bank("#{coin_needed} copper", "#{town_currency(town)}", "#{town}") if coin_needed > 0
+    if coin_needed > 0
+      ### The following line is for the bribe to the night attendant to let you in.
+      coin_needed += 2502 if UserVars.sun['night'] && town =~ /\bShard\b/
+      get_money_from_bank("#{coin_needed} copper", "#{town_currency(town)}", "#{town}")
+    end
     items_to_restock.each do |item|
+      purchased = true
       item['buy_num'].times do
         buy_item(item['room'], item['name'])
         if reget(3, 'Seeing that you are too encumbered')
@@ -49,7 +53,7 @@ class Restock
         reget(2, 'runestone') ? DRCI.put_away_item?(item['name'], @runestone_storage) : stow_hands
       end
     end
-    deposit_coins(@keep_copper, @settings, "#{town}")
+    deposit_coins(@keep_copper, @settings, "#{town}") if purchased
   end
 
   def setup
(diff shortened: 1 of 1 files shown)


Dartellum:
Felonie in LNET tested and it behaved as expected. Trying to get one more tester.

KatoakDR:
I’m not in a place to test this, but my question is if it’ll put unused coins back in the bank? I think yes per your description in the PR but the change is conditionally depositing so makes me question it Perhaps I’m not understanding why this change is necessary? 

It may be clearer once I’m able to test it

Dartellum:
> I’m not in a place to test this, but my question is if it’ll put unused coins back in the bank? I think yes per your description in the PR but the change is conditionally depositing so makes me question it Perhaps I’m not understanding why this change is necessary?
> 
> It may be clearer once I’m able to test it

I failed to see hunting-buddy called restock. The way it was implemented in the push today, it will always go to the bank and deposit money even if nothing needed restocking. Some people have no prehunt_buff room so they were trying to buff in the bank. This PR fixes that behavior and will only deposit if money was withdrawn to purchase items.

Dartellum:
Taurture on LNET tested as well and said the update works as expected before the PR push this morning.

Dartellum:
@KatoakDR did you happen to run the current `restock` and then add this update and see the difference?

jayRyan24:
@rpherbig Can we get this merged, it's messing up a lot of people.

restock.lic
```diff
@@ -37,10 +38,13 @@ class Restock
...
-    ### The following line is for the bribe to the night attendant to let you in.
-    coin_needed += 2502 if UserVars.sun['night'] && town == /\bShard\b/
-    get_money_from_bank("#{coin_needed} copper", "#{town_currency(town)}", "#{town}") if coin_needed > 0
+    if coin_needed > 0
+      ### The following line is for the bribe to the night attendant to let you in.
+      coin_needed += 2502 if UserVars.sun['night'] && town =~ /\bShard\b/
+      get_money_from_bank("#{coin_needed} copper", "#{town_currency(town)}", "#{town}")
+    end
     items_to_restock.each do |item|
+      purchased = true
```
jayRyan24, line 47: This could be moved outside the loop. Just above:
```ruby
purchase = true unless items_to_restock.empty?
items_to_restock.each do |item|
  etc...
```

jayRyan24 (changes requested):
It works as is, but could be a simple chnge to make it cleaner.

restock.lic
Dartellum: Testing real quick.

Dartellum:
> It works as is, but could be a simple chnge to make it cleaner.

Done.

jayRyan24:
@rpherbig Need this merged, it's breaking a lot of people.

PR #4808, opened 2021-03-17
Was bugging out, running out of critters and dancing

jayRyan24:
this wasn't sheathing the offhand weapon so it would fail.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -71,6 +71,7 @@ class SetupProcess
     was_retreating = game_state.retreating?
     game_state.update_room_npcs
     if game_state.dancing?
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -71,6 +71,7 @@ class SetupProcess
     was_retreating = game_state.retreating?
     game_state.update_room_npcs
     if game_state.dancing?
+      game_state.sheath_whirlwind_offhand
```
KatoakDR, line 74: Will the whirlwind offhand be wielded again after dancing? As in, is this fixing one issue but introducing a new problem when resuming whirlwind?

PR #4813, opened 2021-03-18
Update tarantula.lic - add skip alternate setting

BinuDR:
I've added a setting that allows you to skip the alternate skill being chosen so that it's not necessary to exclude so many skills if you just don't want the alternate to happen.
With this set to true, the script will only pick skills you list under `tarantula:`
`tarantula_skip_alternate: true`


Dartellum:
Please add setting to base.yaml.

BinuDR:
I was waiting for some testers to respond that it's working before adding the base setting.  I'll be making another PR for it.

PR #4814, opened 2021-03-18
Adding missed dependency for DRCI

wstampley:
Not having this explicit dependency in DRCA seems to create an ordering/race condition for some scripts - buff, magic trainer, etc. - when used right at login. Looks like the DRCI dependencies are mostly around runestone handling.

Curious about peoples' thoughts here, or if there's a reason to avoid the common-arcana to common-items dependency.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-arcana
 =end
 
-custom_require.call(%w[common events spellmonitor drinfomon])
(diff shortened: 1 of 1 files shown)


wstampley:
@rpherbig - Looks like this might be causing a problem for people on the mainline. Might need to re-evaluate how to fix this, cause I think it creates a circular dependency (common-arcana -> common-items -> common-travel -> common->arcana)...

rpherbig:
Thanks for catching this!

PR #4816, opened 2021-03-18
Update common-arcana.lic

formayor:
remove common-items, it is breaking people's lich


jayRyan24:
We should just revert and not have it as if it were a new addition.

PR #4817, opened 2021-03-18
Revert "Adding missed dependency for DRCI"

jayRyan24:
Reverts rpherbig/dr-scripts#4814

Broke everyone apparently. Can we get this reverted please! Thanks!
@rpherbig


rpherbig:
Thanks for catching this!

PR #4818, opened 2021-03-19
Update base-hunting.yaml

formayor:
nothing spawns in 19193 (cave_trolls)


rpherbig:
Could you instead comment out that room with a note nothing spawns there? That way it doesn't get accidentally re-introduced later.

PR #4820, opened 2021-03-19
base.yaml - Added new tarantula setting

BinuDR:
Added `tarantula_skip_alternate: false` in here.  This is related to https://github.com/rpherbig/dr-scripts/pull/4813


rpherbig:
FYI: this change could have been part of #4813 (i.e. they do not need to be separate PRs)

Issue #4833, opened 2021-03-22
appraisal.lic missing match strings related to getting/stowing weapons from a weapon harness


Nyctus:
**Summary:**
When appraisal.lic pulls a weapon [diamondwood staff] from a weapon harness [embossed weapon harness] the messaging begins with "You deftly remove." This string does not appear to have a match that the script is listening for and so the script times out waiting for a matching string.

**Note:**
I encountered this issue on my Moon Mage.

**Repro:**
1. obtain an embossed weapon harness
2. wear the harness
3. obtain a diamondwood staff
4. put the diamondwood staff in the harness
5. ensure diamondwood staff and harness are listed in the relevant equipment sections of setup YAML
6. run ;appraisal in-game

**Results:**
When the diamondwood staff is drawn from the harness and held, the script will time out waiting for a match.

**Expected:**
The script is able to detect when the diamondwood staff is successfully removed from the weapon harness.

**Example game messaging:**

```
[appraisal]>get diamondwood.staff from my harness
You deftly remove the diamondwood staff from your harness.
>
[appraisal: *** No match was found after 15 seconds, dumping info]
[appraisal: messages seen length: 3]
[appraisal: message: Aura Sight  (12 roisaen)]
[appraisal: message: Braun's Conjecture  (56 roisaen)]
[appraisal: message: You deftly remove the diamondwood staff from your harness.]
[appraisal: checked against [/You get/i, /You pick/i, /You pluck/i, /You are already holding/i, /You fade in for a moment as you get/i, /I could not/i, /What were you/i, /Get what/i, /You need a free hand/i]]
[appraisal: for command get diamondwood.staff from my harness]
[appraisal]>get my diamondwood.staff
You are already holding that.
```

BinuDR:
Are you a warrior mage by chance?

Nyctus:
Moon Mage, sorry for leaving that out. I'll go mention it in the bug. Also, where can I get you the version info if needed?

BinuDR:
It's fine, I'll fix the script.  I know some spells alter the wield/sheath messaging.  Can you also show me message for sheathing the staff?

Nyctus:
Sure thing!
```>put staff in harness
You secure your diamondwood staff to your harness.```

BinuDR:
Thanks

Nyctus:
[...]

PR #4834, opened 2021-03-22
common-items - Add match for get_item_unsafe

BinuDR:
Added a match for `You deftly remove` to the get_item_unsafe method.

This addresses issue https://github.com/rpherbig/dr-scripts/issues/4833


KatoakDR (approved):
Thanks

PR #4835, opened 2021-03-23
Clumsy Arcana Check Matches

wstampley:
Adding matches for when you aren't able to use cambrinth while worn (due to arcana penalties, etc.)

These changes alone will make scripts that use DRCA to charge cambrinth no longer hang when the current arcana skill isn't good enough to use a cambrinth without removing it. This doesn't address harnessing instead, or any other handling we might want to do with this edge case; it just lets scripts keep going as normal instead of waiting for the 'no match was seen' timeout.

I tested this with a misconfigured YAML (marking an item's capacity lower than it is), in several scripts (combat-trainer, buff, etc.)

@Hiinky, @KatoakDR or anyone else - thoughts?

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-arcana
 =end
 
-custom_require.call(%w[common common-items events spellmonitor drinfomon])
+custom_require.call(%w[common events spellmonitor drinfomon])
 
 $MANA_MAP = {
   'weak' => %w[dim glowing bright],
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -8,6 +8,7 @@ charge_messages:
 - lack the mental stamina to charge
 - resists your efforts to charge it
 - You fail to channel
+- you find it too clumsy to charge
 
 invoke_messages:
 - Images of streaking stars falling from the heavens flash
@@ -25,6 +26,7 @@ invoke_messages:
 - Well, that was fun
 - Invoke what
 - You'll have to hold it
+- you find it too clumsy to invoke
 
 prep_messages:
 - You lick the tip of your finger and trace a sigil in the air


KatoakDR:
@wstampley Thanks for the PR. Could you clarify what I need to change in my YAML to test this with a 32 mana armband, I'd like to test the change too to reproduce the problem you all ran into. Thanks

wstampley:
To recreate the same behavior I got by having a sorcery backlash, you can take a camb that is too big for you to use while worn, and set its capacity low enough in your YAML that it thinks it can be used while worn.

I should have been able to use my pendant, but my backlash dumped me below the skill threshold so everything was just failing to match.

KatoakDR:
> @KatoakDR To recreate the same behavior I got by having a sorcery backlash, you can take a camb that is too big for you to use while worn, and set its capacity low enough in your YAML that it thinks it can be used while worn.
> 
> I should have been able to use my pendant, but my backlash dumped me below the skill threshold so everything was just failing to match.

Gotcha, thanks. I've submitted https://github.com/wstampley/dr-scripts/pull/1 that will fallback to holding your cambrinth in these scenarios if you'd like to test it out. The PR includes my test cases/output.

wstampley:
> > @KatoakDR To recreate the same behavior I got by having a sorcery backlash, you can take a camb that is too big for you to use while worn, and set its capacity low enough in your YAML that it thinks it can be used while worn.
> > I should have been able to use my pendant, but my backlash dumped me below the skill threshold so everything was just failing to match.
> 
> Gotcha, thanks. I've submitted [wstampley#1](https://github.com/wstampley/dr-scripts/pull/1) that will fallback to holding your cambrinth in these scenarios if you'd like to test it out. The PR includes my test cases/output.

@KatoakDR - I pulled your change into my branch. Thanks for the team-up.

Anything else you think we need to test / verify before moving forward with this?

KatoakDR:
> Anything else you think we need to test / verify before moving forward with this?

🤔 Can you think of any way my proposed fallback to remove the cambrinth, charge, wear it, could ever get into an infinite recursive loop? Like, after removing the cambrinth... is there any chance the error message `you find it too clumsy` ever occurs while you're **holding** cambrinth? If not, and I don't think it would happen, then I don't think there's an issue.

rpherbig:

Let me know when edge case conversation is resolved

KatoakDR:

Thanks :)

> Let me know when edge case conversation is resolved

I've submitted https://github.com/wstampley/dr-scripts/pull/2 to avoid the theoretical infinite recursion scenario. I don't know that it could occur, but, better safe than sorry.

I also thought maybe this recovery logic to use cambrinth if you've been cursed or just have it configured wrong in your yaml is too much, but then thought as a player my preference is for the scripts to try and "self heal" and "make a best effort". It could be that this retry logic helps ensure a player maintains their buffs and doesn't die. So, from that perspective, I think the retry logic is worth it.

Thanks

common-arcana.lic
```diff
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-arcana
 =end
 
-custom_require.call(%w[common events spellmonitor drinfomon])
+custom_require.call(%w[common common-items events spellmonitor drinfomon])
```
wstampley, line 6: @KatoakDR - Does this create the infinite loop with arcana -> items -> travel -> arcana?

KatoakDR: I've not had any startup issues, but yes, it does create a circular dependency. Hrmmm... I can try to test without it, I'll need to stop all my auto starts b/c I imagine common-items is being started by something else for me already and not being an issue at this point.

KatoakDR: Well... good news and bad news.

Good news, I was able to reproduce the circular dependency issue when `common-items` is included in the require list. I removed all autostart scripts from my character's config then reconnected then ran `common-arcana` and Lich broke. 🔥 

Bad news, when I remove `common-items` from the require list and repeat my test steps then when the code gets to the point where it needs to use DRCI then it fails because that file hasn't been loaded into memory yet.
```
> ,e echo DRCA.charge?('cambrinth manacle', 1)

--- Lich: exec1 active.

[exec1]>charge my cambrinth manacle 1

Try though you may, you find it too clumsy to charge the cambrinth manacle while wearing it.
> 
| *** Your arcana skill is too low to charge your cambrinth while worn

--- Exception: uninitialized constant DRCA::DRCI
Did you mean?  DRC
               DRCA

common-arcana:327:in `charge?'

--- Lich: exec1 has exited.
```
Line 327 mentioned above is `unless DRCI.in_hands?(cambrinth)`

KatoakDR: However, there are other parts in common-arcana that use `DRCI` before I suggested added `common-items` to the require list, so I suspect folks must be running some other script(s) that are loading that file before they get to this point.

And since the Lich community isn't complaining about those other DRCI method calls, then this _might_ be an acceptable risk. As in, what are the odds that you _never_ use Lich in a way that loads `common-items` before you get the `you find it too clumsy` error message charging worn cambrinth?

wstampley: I think we can submit this then, and I'll put some time into fixing the circular dependency next. Should be easy, since most of the fixes can be as simple as moving around some things in various commons.

wstampley:
@rpherbig - I think we're good with this now. It's at least no worse dependency wise than it already was. I'll work on the DRCA circular dependency as a follow-up.

PR #4840, opened 2021-03-26
 Updated to support percWindow which does not begin at the start of the stream

dr-leustyin:


KatoakDR:
Code looks good to me and it addresses my feedback from the prior PR. Thanks 👍

PR #4849, opened 2021-03-28
[script] [hunting-buddy] Enter watchdog loop immediately

KatoakDR:
### Background
* If `combat-trainer` script fails to start for any reason, line 260 that pauses until the scripts starts will never stop.
* Sitting in combat taking no actions and no watchdog to bail if you get hurt leads to you dying. womp womp

### Changes
* Don't pause and wait for `combat-trainer` to start, immediately begin watchdog loop to monitor your vitals.

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -257,7 +257,6 @@ class HuntingBuddy
     message("***STATUS*** beginning hunt '#{args}' for '#{duration}' minutes")
     verify_script('combat-trainer')
     start_script('combat-trainer', args)
-    pause until $COMBAT_TRAINER.running
     counter = 0
     loop do
       clear


KatoakDR:
@Hiinky @jayRyan24 @wstampley in case you're aware of any reason the pause _should_ exist. I think it's a liability (because one of my characters died from it, heh). Thanks

rpherbig:
IIRC there was a race condition in which you ran out to a hunting zone, combat-trainer started, but the stop_on_skills trigger in hunting-buddy happened before combat-trainer was ready to receive the kill command. This was added 4-5 years ago so we can give it a try without the pause

KatoakDR:
> IIRC there was a race condition in which you ran out to a hunting zone, combat-trainer started, but the stop_on_skills trigger in hunting-buddy happened before combat-trainer was ready to receive the kill command. This was added 4-5 years ago so we can give it a try without the pause

Thanks for the background -- that would be a good reason 😅 

Hopefully no more issues without the pause 🤞 🍀

PR #4861, opened 2021-04-05
Update base-hunting.yaml

formayor:
moved thugs #6111 potential no mob spawn room to the bottom of the list.


rpherbig:
Should the room instead be commented out (with a note on why)? I.e. if there's no spawn, why send anyone there?

PR #4862, opened 2021-04-06
athletics.lic - added outdoorsmanship method

BinuDR:
Added a method to call outdoorsmanship between climbing runs.
It consolidates the code and also fixes the issue where your outdoorsmanship is already over 29/34 which stops the outdoorsmanship script, causing you to start the climbing loop again immediately.  This will now pause the remainder of the time so that you don't make a climbing loop and learn nothing.

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -66,6 +66,15 @@ class Athletics
     stow_athletics_items
   end
 
+  def outdoorsmanship_waiting(n)
+    start = Time.now
+    walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
+    stow_athletics_items
+    wait_for_script_to_complete('outdoorsmanship', [n])
+    pause (n*15) - (Time.now - start)
+    get_athletics_items
+  end
+
   def get_athletics_items
     return if @settings.held_athletics_items.empty?
 
@@ -337,10 +346,7 @@ class Athletics
         walk_to(11_126)
         walk_to(9515)
         break if done_training?
-        walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-        stow_athletics_items
-        wait_for_script_to_complete('outdoorsmanship', ['4'])
-        get_athletics_items
+        outdoorsmanship_waiting(4)
       end
     else
       climb_wyvern
@@ -354,10 +360,7 @@ class Athletics
       walk_to(9607)
       walk_to(2900)
       break if done_training?
-      walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-      stow_athletics_items
-      wait_for_script_to_complete('outdoorsmanship', ['4'])
-      get_athletics_items
+      outdoorsmanship_waiting(4)
     end
   end
 
@@ -373,10 +376,7 @@ class Athletics
       walk_to(13_117)
       walk_to(6443)
       break if done_training?
-      walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-      stow_athletics_items
-      wait_for_script_to_complete('outdoorsmanship', ['3'])
-      get_athletics_items
+      outdoorsmanship_waiting(3)
     end
   end
 
@@ -386,10 +386,7 @@ class Athletics
       walk_to(12_838)
       walk_to(6154)
       break if done_training?
-      walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-      stow_athletics_items
-      wait_for_script_to_complete('outdoorsmanship', ['4'])
-      get_athletics_items
+      outdoorsmanship_waiting(4)
     end
   end
 end


athletics.lic
```diff
@@ -373,10 +376,7 @@ class Athletics
       walk_to(13_117)
       walk_to(6443)
       break if done_training?
-      walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
-      stow_athletics_items
-      wait_for_script_to_complete('outdoorsmanship', ['3'])
-      get_athletics_items
+      outdoorsmanship_waiting(3)
```
KatoakDR, line 379: Noticed the other sections that call this function were passing in 4 but this one passes in 3. Is that on purpose? If not, suggesting we change it to 4 for consistency (or, remove the numerical argument and have the method just do 4 rounds or some configurable value that's passed in to the script or read from yaml). Thoughts?
```suggestion
      outdoorsmanship_waiting(4)
```

BinuDR: It's a very long climb run so more than 15 seconds have passed between the time the first climb occurs and the time you get back to the spot where it runs the outdoorsmanship section so it only collects 3 times.

athletics.lic
```diff
@@ -66,6 +66,15 @@ class Athletics
     stow_athletics_items
   end
 
+  def outdoorsmanship_waiting(n)
+    start = Time.now
```
KatoakDR, line 70: Since the purpose of this script is to train athletics, some folks may not understand why the script pauses and begins training outdoorsmanship between loops. I'm suggesting a couple echos to explain that a wait is needed and if the script pauses at the end for any amount of time so they don't think something's wrong.
```suggestion
    echo "Need to wait a bit before resume athletics training"
    start = Time.now
```

Example output:
```
[athletics: Need to wait a bit before resume athletics training]

--- Lich: outdoorsmanship active.

[outdoorsmanship]>collect rock

With skill and nimble fingers you avoid the centipede that runs along in front of your fingers.
Roundtime: 15 sec.
>
You take a step back and run up to the pile of rocks.  Bringing your foot smashing down onto the pile, you send bits and pieces of debris scattering in every direction.  Looking around for the now defunct mountain of glory, you fail to find any signs of its presence.

...
 
--- Lich: outdoorsmanship has exited.

[athletics: Almost ready, need to wait 11 seconds before restart athletics training]
... 
```

KatoakDR, line 70: Since this script is about athletics training, some folks (👋) might not realize why outdoorsmanship gets called. Suggesting we put an echo to clarify.

```suggestion
    echo "Need to wait a bit before can revisit athletics training obstacles"
    start = Time.now
```

athletics.lic
```diff
@@ -66,6 +66,15 @@ class Athletics
...
+    walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
+    stow_athletics_items
+    wait_for_script_to_complete('outdoorsmanship', [n])
+    pause (n*15) - (Time.now - start)
```
KatoakDR, line 74: Since this script is about athletics training, some folks (👋) might not realize there's potentially a pause after the script runs when nothing is hapenning. Suggesting we put an echo to clarify.
```suggestion
    time_to_wait = (n*15) - (Time.now - start)
    if time_to_wait > 0
      echo "Almost ready, need to wait #{time_to_wait.to_i} seconds before resume athletics training"
      pause time_to_wait
    end
```

athletics.lic
BinuDR: I'm trying to remove bloat, not add more.  99% of the people who play DR know that there's a timer for learning between climbs.

KatoakDR: > I'm trying to remove bloat, not add more

Hi Binu, when you use those words to describe my feedback it hurts my feelings and demotivates me to contribute. 💔 I'm not trying to create problems. Like you, I'm trying to support and improve people's experience with Lich.

I respect your perspective on the echos. That's fine. Many of you have been playing this game and using Lich far longer than I have. My intentions behind the suggestions was only to help coach/teach those who use or maintain the scripts to know a little bit more about why it was doing what it does. 👨‍🏫

Thanks

BinuDR: I'm not trying to be rude at all here.  Let me explain it like this.

Those messages would be useful the 1st time someone sees them and then unnecessary scroll anytime after.  It's one of those things that's better to let people ask on lnet or the discord server and learn that way, rather than to have it messaging every time.

The outdoorsmanship training has been in the script for a long time.  Maybe I'll add a comment in the script itself as to what it's doing.

I appreciate your contributions to lich and I'm not trying to discourage you from continuing.  I was busy and so I left a short comment in response to your suggestions so I could resolve them and let the script get merged.  It wasn't meant to be rude.

KatoakDR: > I'm not trying to be rude at all here. Let me explain it like this.

Thank you for the follow up and clarification. 🤗 

> Those messages would be useful the 1st time someone sees them and then unnecessary scroll anytime after.

Excellent point. I feel super silly suggesting it now. 🐑 

> Maybe I'll add a comment in the script itself as to what it's doing.

😄 

> I appreciate your contributions to lich and I'm not trying to discourage you from continuing. I was busy and so I left a short comment in response to your suggestions so I could resolve them and let the script get merged. It wasn't meant to be rude.

❤️ Thanks, Binu

PR #4872, opened 2021-04-08
athletics.lic - pause fix

BinuDR:
Pause time must be positive.   Added in a check to make sure it doesn't pause if it's negative.

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -71,7 +71,8 @@ class Athletics
     walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
     stow_athletics_items
     wait_for_script_to_complete('outdoorsmanship', [n])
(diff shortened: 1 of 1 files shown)


athletics.lic
```diff
@@ -71,7 +71,8 @@ class Athletics
     walk_to(@outdoorsmanship_rooms.sample) unless @outdoorsmanship_rooms.empty?
     stow_athletics_items
     wait_for_script_to_complete('outdoorsmanship', [n])
-    pause (n*15) - (Time.now - start)
+    remaining_pause = ((n*15) - (Time.now - start))
+    pause remaining_pause if remaining_pause > 0
```
wstampley, line 75: Weird that pause doesn't already do nothing for negative pauses...

BinuDR: Yeah.  I am tempted to edit lich.rbw too.

Issue #4867, opened 2021-04-07
Bescort ways - A rare message can stall ;bescort when attempting to leave the first room upon entering the Ways


Nyctus:
Repro:

1. Use ;bescort to enter the Ways toward any valid destination
2. Observe the script as it exits the first room after being pulled through the unstable moongate

Results:
Very occasionally a failed skill check results in the message "You end up walking in circles" and bescort will stall.

Expected:
When "You end up walking in circles" appears upon leaving the first room of the Ways, bescort will re-attempt the first 'pow'/move.

Example when entering via Taniendar and attempting to move toward the Microcosm:
`[bescort]>northeast
You move across the unreal landscape, navigating by the streams of Lunar energy.
You end up walking in circles.`

Related lines 1497 to 1500 of bescort:
`direction_to_pattern = {
      'center' => /the microcosm is to the (?<direction>.*)\.$/,
      'exit' => /You believe the end of the conduit lies (?<direction>.*)\.$|^You are already at the end of the conduit\.$/
    }`

KatoakDR:
Add link to discord conversation, https://discord.com/channels/745675889622384681/745676101392793651/828436928629047296

PR #4874, opened 2021-04-08
[script] [bescort] When astro walking, retry if find yourself walking in circles

KatoakDR:
### Background
* Reported by [Sinistrad](https://discord.com/channels/745675889622384681/745676101392793651/828436928629047296) and [Akhorishaan](https://discord.com/channels/745675889622384681/745676101392793651/829001622821535755) in the Lich discord
* Fixes https://github.com/rpherbig/dr-scripts/issues/4867

### Changes
* When try to move in a direction, if "You end up walking in circles" then retry power walking

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -1394,7 +1394,12 @@ class Bescort
     if Flags['pattern-shift']
       Flags.reset('pattern-shift')
     else
+      Flags.reset('walking-circles')
       move Regexp.last_match[:direction]
+      # If you only walk in circles then you failed to move, try again.
+      if Flags['walking-circles']
+        power_walk(pattern)
+      end
     end
   end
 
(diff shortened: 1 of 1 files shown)


bescort.lic
```diff
@@ -1394,7 +1394,12 @@ class Bescort
     if Flags['pattern-shift']
       Flags.reset('pattern-shift')
     else
+      Flags.reset('walking-circles')
       move Regexp.last_match[:direction]
+      # If you only walk in circles then you failed to move, try again.
+      if Flags['walking-circles']
+        power_walk(pattern)
```
rpherbig, line 1401: Do you need to clear the flag before power walking, in case it happens again?

rpherbig:
Seems reasonable to me ✔️

bescort.lic
```diff
@@ -1394,7 +1394,12 @@ class Bescort
     if Flags['pattern-shift']
       Flags.reset('pattern-shift')
     else
+      Flags.reset('walking-circles')
       move Regexp.last_match[:direction]
```
KatoakDR, line 1398: Feedback from [VTCifer](https://discord.com/channels/745675889622384681/745676101392793651/829707485908434995) in Lich discord:

> I think you're still going to run into an issue with missing match in move unless you define a new move method specific to dr-scripts instead of core lich.

KatoakDR:
I'm of the opinion this will either resolve the issue or have no net effect. I vote to move forward.

rpherbig:
> I'm of the opinion this will either resolve the issue or have no net effect. I vote to move forward.

I'm willing to give it a try. @ me if something goes wrong and this needs reverted.

Nyctus:
I'll patch tonight and try to repro again to confirm the fix. I repro'd again last night but I didn't have the test code as it got patched over when I had to reboot. Argh.

It seems having a couple minor injuries greatly increases the odds of this happening (or I could just be seeing patterns where there are none).

KatoakDR:
Thanks, @Nyctus

PR #4878, opened 2021-04-10
[script] [t2] Removed unused variables

KatoakDR:
### Background
* The assignments to `@repair_timer` and `UserVars.repair_timer_snap` were in the original commit of this script years ago, but were never used. The assignments may have been part of some copy/paste.

### Changes
* Remove assignments to `@repair_timer` and `UserVars.repair_timer_snap` as those variables are not used in this script

### DIFF
--- a/t2.lic
+++ b/t2.lic
@@ -16,8 +16,6 @@ class T2
 
     @counter = 0
     @settings = get_settings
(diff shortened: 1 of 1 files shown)


rpherbig:
It looks like those were copied over from `training-manager`

PR #4891, opened 2021-04-13
Chop-wood Match Improvements

wstampley:
Matches for when you're chopping wood at the same time as someone else, and try to loot their findings.

This happens pretty rarely for me, so it might be a bit before I get to test this case again.


rpherbig:
Sorry, GitHub didn't notify me this went from draft -> ready to merge

PR #4893, opened 2021-04-14
[script] [dependency] delete unused variable @item_url

KatoakDR:
### Background
* While investigating what [Melindrha](https://discord.com/channels/745675889622384681/745675890242879671/831709254699581440) reported in Lich discord about `dependency` exiting due to failure of api.github.com response, found this (unrelated) unused variable.

### Changes
* Delete `@item_url` because it's not used and the URL gives a 404 not found error


rcuhljr:
The api failure may be related to this https://github.com/rpherbig/dr-scripts/issues/4478 anyone who hasn't updated to the new version of dependency will get failures during the brownout period, and lose all access to dependency in May. (someone may need to replace the dependency in the ;repository so our first time setup instructions don't break forever.) https://github.com/rpherbig/dr-scripts/wiki/First-Time-Setup

PR #4890, opened 2021-04-13
[script] [combat-trainer] Robustify stowing non-lodged ammo after shooting

KatoakDR:
### Dependencies
* ~~Depends on https://github.com/rpherbig/dr-scripts/pull/4894~~

### Background
* Fixes https://github.com/rpherbig/dr-scripts/issues/4877 - misidentifies what ammo was shot and doesn't pick it up
* Fixes https://github.com/rpherbig/dr-scripts/issues/4871 - doesn’t stow ammo if “Face what?” message appears first

### Changes
* To handle when game messages arrive out of order or are flat out wrong (e.g. "Face what?" appearing before the "you fire..." message), switch to using `Flags` to detect that you fired your ranged weapon. If yes, after RT then stow the ammo.
* Since the purpose of identifying the kind of ammo shot is for stowing it, and you can't stow lodged ammo, I removed the complicated regex and instead added a simpler flag that specific looks for ammo that doesn't lodge (falls to your feet, lands nearby, passes through incorporeal undead, etc).

## Tests

I tested with the following ammo on incorporeal undead and non-undead:
- crossbow bolt
- Sunderstone shard
- lemon-striped arrow
- blunt-tipped arrows with chevron stripes

Golameth tested with the following ammo on umbra moth's and arthelun cabalists:
- matte indurium sphere

<details>
<summary>Click to toggle tests</summary>

### should stow non-lodging ammo after powershot maneuver (matte indurium sphere)
```
[combat-trainer]>maneuver powershot
You square up your feet and arch your back while searching for an engaged enemy to target.
>
Knowledge from your sanowret crystal about Lunar Magic rings clear in your mind:
The youngest of the magical disciplines, Lunar magic's scholarly study begins in the Imperial era.  Broadly, it is the study of the powers derived from the celestial bodies, which include magical ('enlightened') geometry, psychic phenomenon, and light manipulation.
>
With a loud twang, you let fly your sphere!
 
Your matte indurium sphere lands a hard hit (5/23) to a fuligin umbral moth's left front leg!
The matte indurium sphere falls to the ground!
A fuligin umbral moth's wings spasm one last time, drifting away from its body like a cloud of acrid smoke.  The creature goes still, now nothing more than an ungainly mound of disintegrating shadow.
With expert skill you end the attack and maneuver into a better position.
>
[combat-trainer]>stow my matte sphere
The fuligin moth's massive proboscis licks the air lasciviously as it stands motionless.
>
You pick up the sphere lying at your feet.
[...]

PR #4894, opened 2021-04-14
[script] [events] Store raw match data so can use named groups

KatoakDR:
### Background
* Suggested by @rcuhljr (https://github.com/rpherbig/dr-scripts/pull/4890#discussion_r612112575) to facilitate using named capturing groups in regular expressions used with `Flags` class.
* I'm planning to use named capturing groups with a revised regex for identifying ammo shot in `combat-trainer`

### Changes
* Store the raw regex match data in the flag rather than an array of the captured groups

### regex without names can reference groups by number (regression test)
GIVEN:
```
Your birthday is more than 4 months away.
```
THEN:
```ruby
    Flags.add('birthday', /Your birthday is more than (\d+) (\w+) away/)
    fput('info')
    echo Flags['birthday'][1] # 4
    echo Flags['birthday'][2] # months
```

### regex with names can reference groups by name (new functionality)
GIVEN:
```
Your birthday is more than 4 months away.
```
THEN:
```ruby
    Flags.add('birthday', /Your birthday is more than (?<qty>\d+) (?<unit>\w+) away/)
    fput('info')
    echo Flags['birthday'][:qty] # 4
    echo Flags['birthday'][:unit] # months
```

### DIFF
--- a/events.lic
+++ b/events.lic
@@ -47,7 +47,7 @@ events_action = proc do |server_string|
   Flags.matchers.each do |key, regexes|
     regexes.each do |regex|
       if matches = server_string.match(regex)
-        Flags.flags[key] = matches.to_a
+        Flags.flags[key] = matches
         break
       end
     end


wstampley:
Are they still indexable with the same index key as matchers, or is this going to force any flag based regexs to update to the new syntax?

rcuhljr:
The change will allow both versions to work, it should be a non breaking change in all situations unless some complete madman was type checking the flag return values for arrays as some kind of boolean check. In which case they should feel bad and deserve to have it break. (alternatively someone might loop over all the flag matches but that's a one line fix)

KatoakDR:
> Are they still indexable with the same index key as matchers, or is this going to force any flag based regexs to update to the new syntax?

Reiterating **rcuhljr** said, "The change will allow both versions to work, it should be a non breaking change".

The only gotcha I'm aware of is that if a regex mixes named groups with unnamed groups then you should **only** extract matches out of the result by the named groups. This is because when a named group exists then there's no guarantee to the index/order of the unnamed groups. I noticed this in testing and read about it [here](https://www.regular-expressions.info/named.html) -- "Mixing named and numbered capturing groups is not recommended because flavors are inconsistent in how the groups are numbered. If a group doesn't need to have a name, make it non-capturing..."

For example:

**GOOD: unnamed groups and access results by index**
```
    Flags.add('birthday', /Your birthday is more than (\d+) (\w+) away/)
    fput('info')
    echo Flags['birthday'][1] # 4
    echo Flags['birthday'][2] # months
```

**GOOD: named groups and access results by name**
```
    Flags.add('birthday', /Your birthday is more than (?<qty>\d+) (?<unit>\w+) away/)
    fput('info')
    echo Flags['birthday'][:qty] # 4
    echo Flags['birthday'][:unit] # months
```

**GOOD: mix of named and unnamed groups and access only by name**
```
    Flags.add('birthday', /Your birthday is more than (\d+) (?<unit>\w+) away/)
    fput('info')
    echo Flags['birthday'][:unit] # months
```

**BAD: mix of named and unnamed groups and access only by index**
```
    Flags.add('birthday', /Your birthday is more than (\d+) (?<unit>\w+) away/)
    fput('info')
    echo Flags['birthday'][1] # this MAY or MAY NOT be what you think it should be, no guarantee
```

Hiinky:
This is currently breaking random scripts, most notably killing moonies trying to walk the ways with ;bescort.  Should we revert it while the issues get sorted out?

rcuhljr:
`     harness_mana([20]) unless Flags['harness-check'].include?('effortlessly')` is the line to fix bescort, someone meant to write [0] but used includes incorrectly.

PR #4895, opened 2021-04-14
Correct Harness Flag

Hiinky:
Old typo exposed by the change to Flags.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -1384,7 +1384,7 @@ class Bescort
   def power_walk(pattern)
     fput 'exit' if dead? || Flags['ap-danger']
     if Flags['harness-check']
-      harness_mana([0]) unless Flags['harness-check'].include?('effortlessly')
+      harness_mana([1]) unless Flags['harness-check'].include?('effortlessly')
       Flags.reset('harness-check')
     end
     result = bput('pow', pattern)
--- a/bescort.lic
+++ b/bescort.lic
@@ -1384,7 +1384,7 @@ class Bescort
   def power_walk(pattern)
     fput 'exit' if dead? || Flags['ap-danger']
     if Flags['harness-check']
-      harness_mana([20]) unless Flags['harness-check'].include?('effortlessly')
+      harness_mana([0]) unless Flags['harness-check'].include?('effortlessly')
       Flags.reset('harness-check')
     end
     result = bput('pow', pattern)


Hiinky:
@rcuhljr  Is this what you meant?  Or did I miss something.

bescort.lic
```diff
@@ -1384,7 +1384,7 @@ class Bescort
   def power_walk(pattern)
     fput 'exit' if dead? || Flags['ap-danger']
     if Flags['harness-check']
-      harness_mana([20]) unless Flags['harness-check'].include?('effortlessly')
+      harness_mana([1]) unless Flags['harness-check'].include?('effortlessly')
```
KatoakDR, line 1387: > harness_mana([20]) unless Flags['harness-check'].include?('effortlessly') is the line to fix bescort, someone meant to write [1] but used includes incorrectly.

@Hiinky Robert was referring to accessing the capturing group `1` instead of using `.include?('effortlessly')`, not the mana level being harnessed. For example, something like `harness_mana([20]) unless Flags['harness-check'][1] == 'effortlessly'` instead of assuming the response was an array.

In [#4897](https://github.com/rpherbig/dr-scripts/pull/4897/files#diff-3f09e588727190f4177ac2420ce5c2f2e824b59d06dd03cf135caa68e963d9d2R1387) I'm updating the line to be `harness_mana([20]) unless Flags['harness-check'].to_a.include?('effortlessly')` so that it converts the MatchData's results to an array again, what this script assumes the result to be.

I think you can close this PR, thank you for diving in to help fix the issues I introduced

rcuhljr: I'd avoid the to_a approach, it's kind of pointless because it's never going to give a different result then seeing if the first capture group is effortlessly.

KatoakDR: Noted, thanks

KatoakDR: @rcuhljr It should be like this? https://github.com/rpherbig/dr-scripts/pull/4897/files#diff-3f09e588727190f4177ac2420ce5c2f2e824b59d06dd03cf135caa68e963d9d2R1387

```ruby
harness_mana([20]) unless Flags['harness-check'][1] == 'effortlessly'
```

Hiinky:
I'm totally fine closing this and letting it go with your PR so it stays packaged nice n tidy.

KatoakDR:
Thanks @Hiinky and @rcuhljr for the help!

rpherbig:
Closing due to #4897

bescort.lic
rcuhljr: Yes, so that's the ideal correction. The old one works but it's just kind of nonsensical when you have a single 
 or no match groups.  Even with many matching groups it's still a super weird case that'd you want to use includes on a raw match array.

```
irb(main):001:0> reg = /(\w+)\s+(\w+)\s+(\w+)/
=> /(\w+)\s+(\w+)\s+(\w+)/
irb(main):002:0> data = "red blue green these are the color balls".match(reg)
=> #<MatchData "red blue green" 1:"red" 2:"blue" 3:"green">
irb(main):003:0> data.to_a
=> ["red blue green", "red", "blue", "green"]
```
So here someone could argue well what if I want to know if the red ball appeared? However this isn't really the case, either you knew you wanted to know about the red ball before the regex ran, in which case it should just be in the regex. Or we didn't know the color we're interested in until afterwards at which point it'smore likely that we should be storing out the three ball colors `balls = data[1..3]` and then you can start asking questions like `balls.include?('red') and it makes more sense than including the overall match string 'red blue green' in the array.

tl;dr; I can't think of a realistic situation where calling includes? on the match array still containing the overall capture makes sense, and the pain here is worthwhile for the added benefits having the raw match data provides. In addition to having named group references the matchdata contains the ability to retrieve the entire matching string.

```
irb(main):009:0> data.string
=> "red blue green these are the color balls"
```

PR #4897, opened 2021-04-14
Support <MatchData> response from Flags class introduced in #4894

KatoakDR:
### Background
* https://github.com/rpherbig/dr-scripts/pull/4894 updated Flags class to return `MatchData` regex result rather than array of the string matches to support named groups
* Some scripts treated the response as an array and called methods on the result rather than accessing groups by index, and those scripts are now [breaking](https://discord.com/channels/745675889622384681/745675890242879671/831956248261296158)
* Supercedes #4895

```
--- Lich: error: undefined method `last' for #<MatchData:0x124107c0>
    combat-trainer:937:in `tend_lodged'
    combat-trainer:898:in `execute'
```

```
--- Lich: error: undefined method first' for #<MatchData:0x10e94ea8>
    common-arcana:439:in perc_aura'
    combat-trainer:1575:in `check_starlight'
```

### Changes
* Update scripts that treated Flags result as an array to call `.to_a` on the result first
* Use named capture groups where appropriate
* Replace Flags['name'].first/last with Flags['name'][0]

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -223,7 +223,7 @@ class Bescort
       [
         { name: 'currach', regex: /currach/i, description: "Row a currach between Halasa Temple and Aesry Surlaenis'a." },
         { name: 'mode', options: %w[halasa aesry], description: "halasa or aesry?" }
-      ]  
+      ]
     ]
 
     args = parse_args(arg_definitions)
@@ -346,10 +346,10 @@ class Bescort
   def currach(mode)
     if (mode.include?('aesry') && [5555, 5557, 5558].include?(Room.current.id))
       echo "You're already there silly!"
-      exit    
+      exit
     elsif (mode.include?('halasa') && [15863].include?(Room.current.id))
       echo "You're already there silly!"
-      exit    
+      exit
     elsif mode == 'aesry'
       row_direction = 'pier'
     elsif mode == 'halasa'
@@ -357,7 +357,7 @@ class Bescort
     else
       echo("Unrecognized argument: #{mode}")
       exit
-    end  
+    end
 
     if [5555, 5557, 5558, 15863].include?(Room.current.id)
       move "go currach"
@@ -373,7 +373,7 @@ class Bescort
       else
         bput("row #{row_direction}", 'You pull strongly', 'What were you referring to',  "You can't do that right now")
       end
-    end 
+    end
   end
 
   def jolas(mode)
@@ -385,7 +385,7 @@ class Bescort
 
     if (mode.include?('merkresh') && Room.current.id == 6542) || (mode.include?('harajaal') && Room.current.id == 15253)
       echo "You're already there silly!"
-      exit    
+      exit
     elsif (mode.include?('merkresh') && Room.current.id == 15253) || (mode.include?('harajaal') && Room.current.id == 6542)
       hide? unless DRRoom.room_objs.find { |x| x =~ /The Jolas/ }
       pause 1 until DRRoom.room_objs.find { |x| x =~ /The Jolas/ }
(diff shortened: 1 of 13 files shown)


rcuhljr:
Looks good, a number of those don't actually need to change, but won't hurt anything. (anything with a range [i..j] returns an array on match data like expected. Ideally we'd go in and stop using .includes? on match arrays because that's basically never correct, but as a quick fix it should do.

wstampley:
I didn't find any other cases, and these fixes look like the safe approach to me (I don't know enough about some of the names you gave capture groups, but you're usually pretty good about that stuff).

bescort.lic
```diff
@@ -1384,7 +1384,7 @@ class Bescort
   def power_walk(pattern)
     fput 'exit' if dead? || Flags['ap-danger']
     if Flags['harness-check']
-      harness_mana([20]) unless Flags['harness-check'].include?('effortlessly')
+      harness_mana([20]) unless Flags['harness-check'].to_a.include?('effortlessly')
```
KatoakDR, line 1387:
```suggestion
      harness_mana([20]) unless Flags['harness-check'][1] == 'effortlessly'
```

combat-trainer.lic
```diff
@@ -211,7 +211,7 @@ class SetupProcess
   end
 
   def last_stance
-    Flags['last-stance'][0] =~ /(\d+)%.* (\d+)%.* (\d+)%.* (\d+)/
+    Flags['last-stance'].to_a[0] =~ /(\d+)%.* (\d+)%.* (\d+)%.* (\d+)/
```
KatoakDR, line 214:
```suggestion
    Flags['last-stance'][0] =~ /(\d+)%.* (\d+)%.* (\d+)%.* (\d+)/
```

common-arcana.lic
```diff
@@ -436,7 +436,7 @@ module DRCA
     Flags.add('aura-growing?', 'Local conditions permit optimal growth of your aura', 'Local conditions are hindering the growth of your aura')
     aura = {}
     DRC.bput('perceive aura', 'Roundtime')
-    aura['level'] = Flags['aura-level'] ? starlight_messages.index(Flags['aura-level'].first) : 0
+    aura['level'] = Flags['aura-level'] ? starlight_messages.index(Flags['aura-level'].to_a.first) : 0
```
KatoakDR, line 439:
```suggestion
    aura['level'] = Flags['aura-level'] ? starlight_messages.index(Flags['aura-level'][0]) : 0
```

forge.lic
```diff
@@ -405,10 +405,10 @@ class Forge
   def assemble_part
     while Flags['forge-assembly']
       part = nil
-      part = if Flags['forge-assembly'].first == 'ready to be woven into and around the material'
+      part = if Flags['forge-assembly'].to_a.first == 'ready to be woven into and around the material'
```
KatoakDR, line 408:
```suggestion
      part = if Flags['forge-assembly'][0] == 'ready to be woven into and around the material'
```

mine.lic
```diff
@@ -30,7 +30,7 @@ class Mine
   end
 
   def resource_level
-    case Flags['resource-level'].first
+    case Flags['resource-level'].to_a.first
```
KatoakDR, line 33:
```suggestion
    case Flags['resource-level'][0]
```

mm.lic
```diff
@@ -74,7 +74,7 @@ class MoonMagery
     cast_spell(data, @settings)
     pause 0.25
     if Flags['locate-fail']
-      case Flags['locate-fail'].first
+      case Flags['locate-fail'].to_a.first
```
KatoakDR, line 77:
```suggestion
      case Flags['locate-fail'][0]
```

rpherbig:
@KatoakDR looks good to me, is this ready to be merged?

KatoakDR:
> @KatoakDR looks good to me, is this ready to be merged?

Yes, thank you. Big 👏 to @Hiinky @rcuhljr @wstampley for helping triage while I was away

PR #4901, opened 2021-04-18
athletics - added max argument

BinuDR:
I've added an optional argument named max.

This can be used in conjunction with t2 or when calling it manually so that even if you start at 0/34, it will keep cycling until you've reached or passed 29/34 field experience.
For example, currently I use this to make sure athletics trains enough:

```training_list:
- skill: Athletics
  start: 10
  scripts:
  - athletics xalas
  - athletics xalas
```

This causes the second run to start immediately after the first one finishes if athletics is still below 29/34 and in doing so it does not wait for the timer between climbs.
With this change you can run this:

```training_list:
- skill: Athletics
  start: 10
  scripts:
  - athletics xalas max
```

It will keep training until your athletics is >= 29/34 and respect the athletics timer.

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -30,12 +30,14 @@ class Athletics
         { name: 'undergondola', regex: /^un.*/i, optional: true, description: 'Climb branch etc undergondola' },
         { name: 'xalas', regex: /^xala.*/i, optional: true, description: 'Climb xalas in zoluren in instead of undergondola' },
         { name: 'stationary', regex: /^stat.*/i, optional: true, description: 'Stand still and use a climbing rope' },
-        { name: 'cliffs', regex: /^cliff.*/i, optional: true, description: 'Climb undergondola without attempting branch' }
+        { name: 'cliffs', regex: /^cliff.*/i, optional: true, description: 'Climb undergondola without attempting branch' },
+        { name: 'max', regex: /^max/i, optional: true, description: 'Keep training until 29/34+' }
       ]
     ]
 
     args = parse_args(arg_definitions)
 
+    @end_exp = 29 if args.max
     start_script('skill-recorder') unless Script.running?('skill-recorder')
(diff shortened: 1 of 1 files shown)


athletics.lic
```diff
@@ -30,12 +30,14 @@ class Athletics
...
         { name: 'stationary', regex: /^stat.*/i, optional: true, description: 'Stand still and use a climbing rope' },
-        { name: 'cliffs', regex: /^cliff.*/i, optional: true, description: 'Climb undergondola without attempting branch' }
+        { name: 'cliffs', regex: /^cliff.*/i, optional: true, description: 'Climb undergondola without attempting branch' },
+        { name: 'max', regex: /^max/i, optional: true, description: 'Keep training until 29/34+' }
       ]
     ]
 
     args = parse_args(arg_definitions)
 
+    @end_exp = 29 if args.max
```
KatoakDR, line 40: Thoughts on reusing `crossing_training_max_threshold` setting? This would give players more granular control at which learning rate they want to stop at.

Another example is how `outdoorsmanship` does it, you can pass in the number of mindstates you'd like to increment by, so maybe you want a short run to go up 5 or 10 mindstates and move on to something else. Especially helpful if you're starting at 0 mind states so you don't monopolize your entire time trying to reach 29+, you can go do something else for bit.

BinuDR: The script currently aims for 15 mindstates higher than where you start, with 29 as the max in case you start at > 14/34.

I'd rather not tie it to a crossing-training yaml setting, and I'm not a fan of how the outdoorsmanship script functions with the mindstates agument.

With these changes, if people use crossing-training it will continue to function as it already does.
It will also function as it already does If they call it manually or have it called with t2 without using the max argument.
This setting is specifically for preventing issues with having to call it multiple times in t2.

I would rather not add more yaml settings since it already functions fine without the argument if you don't want to train too many mindstates.

PR #4917, opened 2021-04-24
Adding optional way to configure combat buffs with a waggle

wstampley:
Many people configure all their spells in their char-setup.yaml, and have one or more hunt files. While this works nicely, it can be a pain to use different buffs in each of those hunts, due to the inability to use YAML anchors/references across files. This lets you define a waggle for each hunt, and refer to it by name in the hunt specific file.

If you make no changes to your YAML, you will continue to use `buff_spells:` for all hunts as normal.

@ubiquitous42 - I think this was something you had explicitly asked for as well, if you want to test.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -978,7 +978,11 @@ class SpellProcess
 
     @settings = settings
 
-    @buff_spells = settings.buff_spells
+    if settings.combat_buff_waggle && settings.waggle_sets[settings.combat_buff_waggle]
+      @buff_spells = settings.waggle_sets[settings.combat_buff_waggle]
+    else
+      @buff_spells = settings.buff_spells
+    end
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -978,7 +978,11 @@ class SpellProcess
 
     @settings = settings
 
-    @buff_spells = settings.buff_spells
+    if settings.combat_buff_waggle && settings.waggle_sets[settings.combat_buff_waggle]
```
rpherbig, line 981: `combat_buff_waggle` needs added to `base.yaml` (and maybe `base-empty.yaml`)

wstampley: Added to `base.yaml` and expanded the existing comment around `buff_spells`. It didn't look like there were any other string style properties in `base-empty.yaml`, so I didn't put this one there either.

KatoakDR (approved):

PR #4923, opened 2021-04-27
[script] [common-arcana] Treat 'stowed' cambrinth as worn if you're wearing it

KatoakDR:
I'm tired, will fill this out later :)

### Background
* TODO

### Changes
* TODO

## Tests

TODO

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -265,7 +265,10 @@ module DRCA
 
   def find_cambrinth(cambrinth, stored_cambrinth, cambrinth_cap)
     if stored_cambrinth
-      DRC.bput("get my #{cambrinth}", 'You get', 'You are already holding that')
+      case DRC.bput("get my #{cambrinth}", 'You get', 'You are already holding that', 'But that is already in your inventory')
+      when "But that is already in your inventory"
+        find_cambrinth(cambrinth, false, cambrinth_cap)
+      end
     elsif DRSkill.getrank('Arcana').to_i < cambrinth_cap * 2 + 100
       DRC.bput("remove my #{cambrinth}", 'You remove', 'You slide', 'You sling', 'You take', 'You detach', 'You aren\'t wearing that.', 'You grab your', 'You clutch', 'You need a free hand')
     end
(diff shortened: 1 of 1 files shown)


wstampley:
If it's flagged to be stored, and you are wearing it instead, should it remove it and use it as declared?

KatoakDR:
> If it's flagged to be stored, and you are wearing it instead, should it remove it and use it as declared?

That's what the change to `find_cambrinth` does -- if it fails to "get" the item because you're wearing it, then it will retry by removing it.

Are you asking about the `stow_cambrinth` method? Rather than just stop at adding the missing match string to avoid hanging, go ahead and remove then stow the cambrinth?

wstampley:
> That's what the change to `find_cambrinth` does -- if it fails to "get" the item because you're wearing it, then it will retry by removing it.

Hm. It looked to me like it would only remove it if your skill was too low. I guess that's just semantics if you made the other change to stow. The 'desired' state when the whole process is done is that if your cambrinth was marked to be stowed, then it's in your bag instead of worn at the end, is all I'm saying. Otherwise, you hit this case every time you try to cast anything.

KatoakDR:
> It looked to me like it would only remove it if your skill was too low

Ahhh, I see what you’re talking about now. Yeah, it should just remove the cambrinth because according to config the player said it wasn’t a worn item so if it’s being worn then that’s not what the player intended (assuming the config was correct). 

> The 'desired' state when the whole process is done is that if your cambrinth was marked to be stowed, then it's in your bag instead of worn at the end

Yes, that’d be good. I was too focused on recovery logic and didn’t think about just course correcting the scenario for future use. Thanks

KatoakDR:
@wstampley With the new common-items methods, I've updated the recovery logic for finding a cambrinth item that's out of place at runtime contrary to what your config says where the item should be. Would appreciate your 👀 on this again when you get a chance. Thanks!

wstampley:
> @wstampley With the new common-items methods, I've updated the recovery logic for finding a cambrinth item that's out of place at runtime contrary to what your config says where the item should be. Would appreciate your 👀 on this again when you get a chance. Thanks!

All the changes look good. If I have time this weekend, I'll try to test this out some.

KatoakDR:
Status: testing

KatoakDR:
@rpherbig @wstampley This is ready for your all's review. I've also added automated tests to verify the logic (Huge kudos to wstampley for figuring out the mocking 🎉 )

PR #4930, opened 2021-04-30
[scripts] [common-items] [01] Add "DRCI" prefix to constants

KatoakDR:
This is the 1st of many PRs I'll be submitting for common-items.

To make it easier to review the changes, I'm phasing them in.

This one focuses on psuedo namespacing the constants.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -3,15 +3,15 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-items
 =end
 
-$TRASH_STORAGE = %w[basket bin gloop barrel bucket urn log arms stump tree statue chamberpot]
+$DRCI_TRASH_STORAGE = %w[basket bin gloop barrel bucket urn log arms stump tree statue chamberpot]
 
-$DROP_TRASH_SUCCESS_PATTERNS = [
+$DRCI_DROP_TRASH_SUCCESS_PATTERNS = [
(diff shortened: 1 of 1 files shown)


wstampley:
With constants like this - is the intent that they are just re-used inside DRCI, or that they are supposed to be available to other scripts/files to use as well?

I ask, because if they are considered internals for DRCI, I think we could put them inside the module itself, which would keep them from being global.

wstampley:
It also might be nice to mark all the dependent PRs as drafts or something, because they're hard to review until this change goes in.

KatoakDR:
> With constants like this - is the intent that they are just re-used inside DRCI, or that they are supposed to be available to other scripts/files to use as well?
> 
> I ask, because if they are considered internals for DRCI, I think we could put them inside the module itself, which would keep them from being global.

Just used within DRCA class. They were originally defined as $ constants so I ran with it. To not blow up the train of PRs already queued up, I’ll add a new PR to move the constants to class variables. Thanks

KatoakDR:
> It also might be nice to mark all the dependent PRs as drafts or something, because they're hard to review until this change goes in.

To help with reviewing the downstream PRs, I’ve included in the description a link to the specific commit that PR is introducing. That’ll allow you to review and comment on the specific change without all the noise. Thanks

PR #4943, opened 2021-05-02
[script] [combat-trainer] Optimize load weapon attempts

KatoakDR:
### Dependencies
* ~~Depends on https://github.com/rpherbig/dr-scripts/pull/4945~~

### Background
* Occasionally, the game may not respond to the `load` command to load your ranged weapon, requiring you to send the command again
* `combat-trainer` handles that nuance by always doing two load attempts (`fput` followed by `bput`), even when not necessary
* In the happy path, the current implementation always incurs an unnecessary second network roundtrip

Note in the below examples that two load attempts are made, even though one was sufficient.

**Example 1**
```
[combat-trainer]>load
>
You reach into your quiver to load the battle stonebow with your last stone shard.
Roundtime 7 sec.
>
[combat-trainer]>load
Your battle stonebow is already loaded with a senci stone shard.
```

**Example 2**
```
[combat-trainer]>load
Your battle stonebow is already loaded with a senci stone shard.
>
[combat-trainer]>load
Your battle stonebow is already loaded with a senci stone shard.
>
[combat-trainer]>aim
You shift your target to a dire bear.
```

### Changes
* Consolidate the success and failure match strings for dual and single loading sections to reduce duplicated code
* Use a flag to know if the first load attempt succeeded, and if not then try the second load attempt
* To handle that the first load attempt (due to the game bug) is a noop, we use `wget` with a 1 second timeout on the first attempt then `bput` on the second attempt (like original logic)
  - `wget` is used instead of `fput` because the `case/when` statements look for specific string matches and we don't want to move forward due to any game response that happens to be received first
  - The 1 second timeout with `wget` ensures that first attempt is quick like `fput` if the game bug occurs so that we can quickly move into the second load attempt

**Result 1**
_Note that two load attempts were necessary (due to game bug), but with help of https://github.com/rpherbig/dr-scripts/pull/4945 no errors about "no match was found" are echoed from the first load (desired behavior)._
```
[combat-trainer]>loot

> 
[combat-trainer]>load

[combat-trainer]>load

You reach into your quiver to load the assassin's bow with your last lemon-striped arrow.
Roundtime 2 sec.
```

**Result 2**
_Note that only one load attempt was needed, saving time and scroll by not needing to perform a reundant load command._
```
[combat-trainer]>remove my assassin's.bow

You sling a bone-inlaid assassin's bow off from over your shoulder.
> 
[combat-trainer]>load

You reach into your quiver to load the assassin's bow with your last lemon-striped arrow.
Roundtime 2 sec.
```


KatoakDR:
Status: testing

rpherbig:
What's left before this is ready to be merged?

KatoakDR:
> What's left before this is ready to be merged?

I need to get back to you on that... I didn't take good enough notes on _where_ I left off during testing 😬

KatoakDR:
Closing in favor of https://github.com/rpherbig/dr-scripts/pull/4890

PR #4945, opened 2021-05-03
[script] [common] Add options arg to bput/wput

KatoakDR:
### Background
* I want to perform commands with variable wait times and take an action based on if a response was received within the timeout.
* In these cases, I may also want to suppress the "*** No match was found" error messages because it's acceptable if a response was not received within the timeout.
* In these cases, echoing the error would serve only to confuse the players and create scroll spam.

### Changes
* Refactor `wput` method so that it's second argument is an options hash map instead of a timeout number.
  - Initially supported option values are:
    - `timeout` _number of seconds to wait for an expected response before returning. Default is 15._
    - `debug` _boolean flag to echo debug info, such as what command, options, and match strings are being evaluated. Default is false._
    - `suppress_no_match` _boolean flag that when true will not echo the "*** No match was found" error messages. Default is false._
* Refactor `bput` method to check if the first element in the matches array is an options hash map (way to go dynamic languages!).
  - This keeps `bput` method's behavior and signature backwards compatible while also supporting accepting a new options argument.
 * In test class, remove unused `@result = nil` from setup method

## Usage Examples
<details>
<summary>Click to toggle examples</summary>

### command only
```
> ,e DRC.bput('enc')

--- Lich: exec1 active.

[exec1]>enc

  Encumbrance : None (0/11)
> 
[exec1: *** No match was found after 15 seconds, dumping info]

[exec1: messages seen length: 1]

[exec1: message:   Encumbrance : None]

[exec1: checked against []]

[exec1: for command enc]

--- Lich: exec1 has exited.
```

### command and match strings
```
> ,e DRC.bput('enc', 'Encumbrance')

--- Lich: exec1 active.

[exec1]>enc

  Encumbrance : None (0/11)
> 
--- Lich: exec1 has exited.
```

### command and no matches, suppress errors
```
> ,e DRC.bput('enc', { 'suppress_no_match' => true, 'debug' => true }, 'Obvious paths')

--- Lich: exec1 active.

[exec1: bput.message=enc]

[exec1: bput.options={"suppress_no_match"=>true, "debug"=>true, "timeout"=>15}]

[exec1: bput.matches=[["Obvious paths"]]]

[exec1]>enc

  Encumbrance : None (0/11)
> 

... (wait 15 seconds) ...

--- Lich: exec1 has exited.
```

### command with configurable timeout, show errors
_For demonstration, using an impossibly short timeout_
```
> ,e echo DRC.bput('load', { 'timeout' => 0.1, 'debug' => true }, 'You load', 'You carefully load', 'You reach into', 'is already loaded')

--- Lich: exec1 active.

[exec1: bput.message=load]

[exec1: bput.options={"timeout"=>0.1, "debug"=>true}]

[exec1: bput.matches=[["You load", "You carefully load", "You reach into", "is already loaded"]]]

[exec1]>load

[exec1: *** No match was found after 0.1 seconds, dumping info]

[exec1: messages seen length: 0]

[exec1: checked against [/You load/i, /You carefully load/i, /You reach into/i, /is already loaded/i]]

[exec1: for command load]

[exec1: ]

--- Lich: exec1 has exited.
```

### command with configurable timeout, suppress errors
_For demonstration, using an impossibly short timeout_
```
> ,e echo DRC.bput('load', { 'timeout' => 0.1, 'suppress_no_match' => true, 'debug' => true }, 'You load', 'You carefully load', 'You reach into', 'is already loaded')

--- Lich: exec1 active.

[exec1: bput.message=load]

[exec1: bput.options={"timeout"=>0.1, "suppress_no_match"=>true, "debug"=>true}]

[exec1: bput.matches=[["You load", "You carefully load", "You reach into", "is already loaded"]]]

[exec1]>load

[exec1: ]

--- Lich: exec1 has exited.
```
</details>

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -96,6 +96,7 @@ module DRC
   # Will wait for matching text up to 15 seconds then timeout.
   def bput(message, *matches)
     options = (matches.shift if matches.first.is_a?(Hash)) || {}
+    options['timeout'] ||= 15
     wput(message, options, matches)
   end
 
@@ -103,7 +104,7 @@ module DRC
   # before performing command and do smart retries.
   # Will wait for matching text up til the given timeout in seconds.
   def wput(message, options, *matches)
-    timeout = (options['timeout'] ||= 15)
+    timeout = options['timeout']
     suppress = options['suppress_no_match']
 
     if options['debug']
--- a/common.lic
+++ b/common.lic
@@ -95,13 +95,23 @@ module DRC
   # before performing command and do smart retries.
   # Will wait for matching text up to 15 seconds then timeout.
   def bput(message, *matches)
-    wput(message, 15, matches)
+    options = (matches.shift if matches.first.is_a?(Hash)) || {}
+    wput(message, options, matches)
   end
 
   # Like `fput` but better because will wait for RT
   # before performing command and do smart retries.
   # Will wait for matching text up til the given timeout in seconds.
-  def wput(message, timeout, *matches)
+  def wput(message, options, *matches)
+    timeout = (options['timeout'] ||= 15)
+    suppress = options['suppress_no_match']
+
+    if options['debug']
+      echo "bput.message=#{message}"
+      echo "bput.options=#{options}"
+      echo "bput.matches=#{matches}"
+    end
+
     waitrt?
     log = []
     matches.flatten!
@@ -154,11 +164,15 @@ module DRC
         end
       end
     end
-    echo "*** No match was found after #{timeout} seconds, dumping info"
-    echo "messages seen length: #{log.length}"
-    log.reverse.each { |logged_response| echo "message: #{logged_response}" }
-    echo "checked against #{matches}"
-    echo "for command #{message}"
+
+    unless suppress
+      echo "*** No match was found after #{timeout} seconds, dumping info"
+      echo "messages seen length: #{log.length}"
+      log.reverse.each { |logged_response| echo "message: #{logged_response}" }
+      echo "checked against #{matches}"
+      echo "for command #{message}"
+    end
+
     ''
   end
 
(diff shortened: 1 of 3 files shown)


KatoakDR:
@rpherbig @wstampley @binu this is ready for your review, thanks

common.lic
```diff
@@ -95,13 +95,23 @@ module DRC
...
+    options = (matches.shift if matches.first.is_a?(Hash)) || {}
+    wput(message, options, matches)
   end
 
   # Like `fput` but better because will wait for RT
   # before performing command and do smart retries.
   # Will wait for matching text up til the given timeout in seconds.
-  def wput(message, timeout, *matches)
+  def wput(message, options, *matches)
+    timeout = (options['timeout'] ||= 15)
```
rpherbig, line 106: Should 15 be a property of `wput` or `bput`?

rpherbig:
Code makes sense to me ✔️

common.lic
KatoakDR: Yeah, that makes more sense that it's part of `bput` so that `wput` is simpler, just running with whatever is given to it. I've made that change per https://github.com/rpherbig/dr-scripts/pull/4945/commits/bfa5a0d2fd436b6d5cca8c136a4b312b99dcac32

With this change, technicaly, we don't need `wput` method anymore (which I originally introduced so that I had a way to specify custom timeouts for scripts like `feed-cloak`).

Do you think we should keep `wput` or consolidate back down to just `bput`?

rpherbig: I think just keeping bput is fine, but I don't have a strong preference

KatoakDR: I'll combine them again. The only use of `wput` I found was in `feed-cloak` which is the script for which I introduced `wput` for

KatoakDR:
@rpherbig I combined `wput` back into `bput` now that `bput` can handle the option for configurable timeout. This is ready for your final review. I've been using this version of `common` script all week with no issues. Thanks

PR #4951, opened 2021-05-08
[test] configure simplecov gem to collect code coverage

KatoakDR:
### Background
* @wstampley and I are working on improving test coverage of the Lich scripts
* Code coverage is a helpful metric to know which methods and logic branches have been tested

### Changes
* Configure [simplecov](https://github.com/simplecov-ruby/simplecov) to collect code coverage from the `minitest` tests
* Due to how simeplecov instruments the ruby runtime to track code coverage, our scripts need to be loaded at test time via `require_relative` command, which is very limited in which file extensions it'll load.
  * As workaround, the `Rakefile` test task creates a copy of the `*.lic` scripts as `test/tmp/*.rb` and then the `test_harness.rb` file loads the scripts from that location when they're tested.
  * These temp files are deleted once the test task completes.
* Update `.ruby-version` to **2.5.5**, the same version that the https://bit.ly/WinLich all-in-one installer uses

### Code Coverage
image


Gemfile.lock
```diff
@@ -1,30 +1,72 @@
 GEM
```
KatoakDR, line 1: I deleted my `Gemfile.lock` file locally then ran `bundle`. This is the result.

Rakefile
```diff
@@ -1,10 +1,38 @@
 #!/usr/bin/env rake
 
 require 'rake/testtask'
+require 'fileutils'
+
+TEST_TMP_DIR = 'test/tmp'
```
KatoakDR, line 6: The path `test/tmp` was chosen because it was one of the already ignored paths in `.gitignore`

rpherbig:
Looks fine to me ✔️ 

@rcuhljr want to take a look at this?

KatoakDR:
I've backmerged `head` into this PR and fixed a couple lines that shook out of that. All tests pass now (via `rake`). Ready for your all's final review and feedback. Thanks!

PR #4957, opened 2021-05-16
[script] [find] Find NPCs in more places, add unit tests (100% line coverage)

KatoakDR:
### Background
* I'd like to use `find` to search for any NPC in any town where we've set up rooms to search in.
* Currently, the script handles very specific NPCs (e.g. Ozursus in Asemath Academy) or defaults to Crossing city.

### Changes
* Move room numbers to `base-town.yaml` file so that we can more easily maintain and grow the list, as well be much easier to test by injecting values.
* While preserving backwards compatibilty for named NPCs like 'ozursus', 'theurgist', 'monk', and 'shaman', you can now specify a town to search for an arbitrary NPC via optional `town` argument.
  - `;find crossing guard`
  - `;find shard urchin`
* Removed 'includes' and added module prefixes (DRC, DRCT, etc) where appropriate to clarify code usage.
 * Add `test_find.rb` as this script did not have unit tests yet
   - Refactored some methods and conditions to improve testability for code coverage
 * Add mocks for `DRRoom` and `Map` to `test_harness.rb`
 * Updated test harness to reset internal mocked constants (DRStats, DRRoom, etc) between tests

## Tests
See `test/test_find.rb`
```
100.0% lines covered
97.44% branches covered
52 relevant lines. 52 lines covered and 0 lines missed.
39 total branches, 38 branches covered and 1 branches missed.
```


KatoakDR:
Feedback from @wstampley:
> Should the room numbers in base-find.yaml just be part of base-town.yaml?
> For example:
```
Crossing:
  regions:
  - streets: [rooms]
  - academy: [rooms]
 ```

**Comment by KatoakDR** [comment]
@wstampley I've moved the room data into `base-towns.yaml` per your suggestion.

I've also finished with the tests and achieved 100% line coverage 🎉

File: find.lic [hunk 1]
```diff
@@ -2,76 +2,109 @@
...
+    when /shaman/i
+      town = 'Dirge' if town.nil? || town.empty?
+      region = 'rooms'
+    when /theurgist/i
+      town = 'Shard' if town.nil? || town.empty?
+      region = 'rooms'
+    when /monk/i
+      town = 'Boar Clan' if town.nil? || town.empty?
+      region = 'pilgrimage_trail' if town == 'Boar Clan'
+    end
```
KatoakDR, line 51: For backwards compatibility with just specifying an NPC's name without the town/area to search in, this logic does that. It'll set the town automatically unless you specified one in the script args.

find.lic
```diff
@@ -2,76 +2,109 @@
...
-      walk_to(room_id)
-      found?
+  # If no town provided then 'Crossing' is used as default.
+  # Anything that starts with 'boar' is expanded to 'Boar Clan'.
+  # Each word in the town is capitalized to match values in base-town.yaml.
+  def format_town(town)
+    town = '' if town.nil?
+    case town
+    when /^boar/i
+      town = 'Boar Clan'
```
KatoakDR, line 67: Use convenience to specify shorter, or one-word variants of towns/areas and the script will expand the values to correct lookup value in `base-towns.yaml`.

For now, **Boar Clan** is the only city with wandering NPC rooms that needs this help.

Others could be "River" for "Riverhaven" and "Theren" for "Therenoborough", as future examples.

test/test_afk.rb
```diff
@@ -6,27 +6,6 @@
...
-  def id
-    1
-  end
-end
-
-class XMLData
-  def self.room_title
-    'Middle of Nowhere'
-  end
-end
```
KatoakDR, line 28: Lines 9-28, moved to `test_harness.rb` for easier mocking

find.lic
```diff
@@ -2,76 +2,109 @@
...
-    pilgrimage_trail = [4179, 4180, 4181, 4182, 4183, 4184, 4185, 4186, 4187, 4188, 4189, 4190, 4191, 4192, 4193, 4194]
+  def initialize
     arg_definitions = [
       [
-        { name: 'npc', regex: /\w+/, description: 'Unique name/noun of the NPC to find' },
-        { name: 'follow', regex: /follow/i, optional: true, description: 'Follow the NPC once it is found' }
+        { name: 'town', options: ['crossing', 'dirge', 'shard', 'boar'], optional: true, description: 'Town/area to search' },
+        { name: 'npc', regex: /^(?!follow)([\w\s\-']+)$/i, optional: true, description: 'Unique name/noun of the NPC to find' },
```
rpherbig, line 13: Why is `npc` optional? It seems to me that would be the one required arg...

rpherbig:
Two comments, the rest of the code changes look good to me ✔️

find.lic
KatoakDR: Yes, I should have explained that change. When `npc` was required then the first argument to the script was always assigned to that variable. This made trying to specify the optional town difficult. For example `;find shard guard` would assign "npc=shard, flex=[guard]".

I think it has something to do with how `parse_args` may be giving priority to assign all the required arguments first then assign out the optional ones.

I found that if I made them all optional then the arguments get assigned in the order they're given. So `;find shard guard` becomes "town=shard, npc=guard".

As a trade-off, the script itself now checks if `npc` is blank and shows the help menu manually.

I need to distill the above into comments in the code to explain all that.

find.lic
KatoakDR: > Could you update `join-thieves.lic`?

What would you like updated? Are you referring to https://github.com/rpherbig/dr-scripts/blob/master/join-thieves.lic#L12

Technically, `;find beggar` will still work because if no town is specified and if the NPC isn't one of the "special ones" the code already had logic for, then the town will default to Crossing. However, for clarity, we could update the join-thieves script to explicitly pass in "crossing" as the town.

KatoakDR: `join-thieves` updated per https://github.com/rpherbig/dr-scripts/pull/4957/commits/78ee29318146249e0cbd0861754f9102bff73eb3

find.lic
KatoakDR: Code comments updated, https://github.com/rpherbig/dr-scripts/pull/4957/commits/632a7dde164fafe53dca72bf505dd2de238976ce

find.lic
```diff
@@ -2,76 +2,117 @@
...
+    when /^boar/i
+      town = 'Boar Clan'
     end
+    town.split(' ').map(&:capitalize).join(' ')
+  end
 
-    echo "***STATUS*** Failed to find an NPC named '#{@npc}', there might be a room missing from the search path"
-    exit
+  # Determine the rooms to search for the requested NPC in the suggested town.
+  # Certain well-known NPCs, like Ozursus in Asemath Academy, ignore the requested town.
```
KatoakDR, line 81: Remove obsolete comment
```suggestion
```

PR #4960, opened 2021-05-19
[script] [common-items] Add retry logic to stow/put away item

KatoakDR:
### Background
* You may get the message "Something appears different about the {item}, perhaps try doing that again." when an outdated item is updated upon use, for example [leather lotion](https://elanthipedia.play.net/Item:Leather_lotion)
  - You can retry the command (get/stow/put/etc) in this scenario.
* You may get the message "Oddly, when you attempt to stash it away safely" when you're cursed and unable to let go of any items from your hands.

### Changes
* Add new set of match strings for when a stow/put operation should be retried.
* Add new match strings for when you're cursed and consider the action failed.

## Tests
See `test/test_common_items.rb`

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -39,6 +39,7 @@ module DRCI
   @@drop_trash_retry_patterns = [
     # You may get the next message if the item would be damaged upon dropping.
     /If you still wish to drop it/,
+    /would damage it/,
     # You may get the next messages when an outdated item is updated upon use.
     # "Something appears different about the <item>, perhaps try doing that again."
     # Example: https://elanthipedia.play.net/Item:Leather_lotion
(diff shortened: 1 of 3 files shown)


common-items.lic
```diff
@@ -34,7 +53,7 @@ module DRCI
     /You pluck/,
     /You deftly remove/,
     /You are already holding/,
-    /You fade in for a moment as you get/
+    /You fade in for a moment as you/
```
KatoakDR, line 56: There's two variants:
1. "as you get..."
2. "as you pick up..."

common-items.lic
```diff
@@ -69,7 +88,7 @@ module DRCI
 
   @@wear_item_failure_patterns = [
     /You can't wear/,
-    /You .* unload/,
+    /You (need to|should) unload/,
```
KatoakDR, line 91: To more clearly self-document which phrases are being handled

PR #4966, opened 2021-05-22
[script] [heal-remedy] [herb-stock] Add support for new @remedy_container setting

KatoakDR:
### Background
* [Astarin](https://discord.com/channels/745675889622384681/745675890242879671/845317619888554044) and [Naimo](https://discord.com/channels/745675889622384681/745675890242879671/845767728210903080) asked for `heal-remedy` script to use "my" qualifier and allow use of a container to mitigate where the script tries to use random items on the ground or perhaps share a noun but aren't the remedy you wanted to use.

### Changes
* Add new yaml setting, `remedy_container:`, a container where players store their ready-to-use healing remedies.
  - This container is in contrast to `herb_container:` and `alchemy_herb_storage:` settings that are used by `alchemy.lic` and `remedy.lic` to store raw ingredients and crafted products.
  - I did not want to reuse either of those existing settings out of precaution to breaking those scripts inadvertently.
  - This also means that to use the new `remedy_container:` setting is opt-in, if a player does not define it then the logic of `remedy-heal.lic` and `herb-stock` to generically "get/stow" behaves the same.
  - Use `DRCI.get_item?` and `DRCI.put_away_item?` methods that handle if a container argument is or isn't provided, which provides the backwards compatibility for people that don't opt in to this behavior by not defining `remedy_container:` setting.
  - Rename `key` variable to be more explicit of its value (e.g. body part, remedy) and avoid shadowing (same variable name was being used for different purposes as lexical scope changed).
* Since it's [not always clear](http://www.tinyheroes.com/forums/DragonRealms/Lore/Bugs%20-%20Lore/thread/1765003) which verb to use to consume an herb, if unable to drink a remedy, try to eat it instead (logic existed in one method but needed to exist consistently in more scenarios per [this Discord conversation](https://discord.com/channels/745675889622384681/745675890242879671/846092459264049182))

## Tests
[Recruiting testers](https://discord.com/channels/745675889622384681/745675890242879671/845803023271985182) from Lich discord since I don't use any of these scripts myself 😁  

Per [Naimo](https://discord.com/channels/745675889622384681/745675890242879671/845821208914165771):
> The test phase, it seems to me, is to throw a bunch of forage and found herbs on the ground and running ;heal-remedy.  I'll also run the ;remedy craft script to see if there is any interference.

Per [Astarin](https://discord.com/channels/745675889622384681/745675890242879671/845823525558419486):
> Seems to work so far.  I'll keep an eye on it, but nothing crazy happened when I went to heal an eye wound that required getting sufil sap out and putting it away.


heal-remedy.lic
```diff
@@ -91,9 +91,12 @@ def remedy_apply_wounds(body_part)
     if DRCI.get_item?(remedy, @remedy_container)
       case remedy
       when /salve/, /ungent/, /poultices/, /ointment/
-        DRC.bput("rub #{remedy}", 'You')
+        DRC.bput("rub my #{remedy}", 'You')
       when /potion/, /tonic/, /elixir/, /draught/
-        DRC.bput("drink #{remedy}", 'You')
+        case DRC.bput("drink my #{remedy}", "You drink", "Drink what?", "can't drink")
```
wstampley, line 96: Feel like this would make a good little helper method:
```
consume_remedy?(remedy_name, try_drink_first)
  if try_drink_first
     drink
        when fail, eat
  else
    eat
      when fail, drink
  end
end
```

KatoakDR: Agree

rpherbig:

rpherbig:
What's left before this is ready to be merged?

KatoakDR:
> What's left before this is ready to be merged?

Thanks for the ping. I need to move some duplicate code into a helper function per https://github.com/rpherbig/dr-scripts/pull/4966#discussion_r638369996

Will get that done here shortly

heal-remedy.lic
KatoakDR: Addressed per https://github.com/rpherbig/dr-scripts/pull/4966/commits/a76fe0555590fd1f8f702eded9f64ce5f9fdb2ee

KatoakDR:
@rpherbig I _think_ this is ready to go. Naimo and Astarin said they didn't have issues with the script. I've put out multiple calls for folks to test/review in case they might be affected by the changes and haven't heard any negative feedback.

Thanks

PR #4974, opened 2021-05-29
Update base-hunting.yaml to correct dire_bears

Donntae:
Removed room 1578 from dire_bears because it consistently spawns only silver-backed bears.


rpherbig:
Thank you for the PR!

Donntae:
My pleasure!

PR #4986, opened 2021-06-03
Empath safe magic

wstampley:
Had my empath get her first shock today due to absolution falling off during combat, and combat trainer not having any checks for harmful spells.

- Centralize logic for if an empath can take harmful actions into game state
- Add "aggressive" guard on cast to abort non-harmful spells in the right situations
- Add filters to offense and training spell selection based on if offense is allowed and spells are harmless
- Update base-spells.yaml to include harmless tags on most of the spells an empath would use that don't cause shock

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1123,6 +1123,7 @@ class SpellProcess
     Flags.add('ct-starlight-depleted', 'enough starlight') if DRStats.trader?
     Flags.add('ct-regalia-expired', 'Your .* glimmers? weakly') if DRStats.trader?
     Flags.add('ct-regalia-succeeded', 'You cup your palms skyward to bask in') if DRStats.trader?
+    Flags.add('ct-shock-warning', 'You are distinctly aware that completion of this spell pattern may bring you shock')
 
     @offensive_spells
       .select { |spell| spell['expire'] }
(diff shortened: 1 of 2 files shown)


rpherbig:
Changes look good to me (once the comments are resolved) ✔️

combat-trainer.lic
```diff
@@ -1655,6 +1664,7 @@ class SpellProcess
                    .select { |spell| spell['min_threshold'] ? game_state.npcs.length >= spell['min_threshold'] : true }
                    .select { |spell| spell['max_threshold'] ? game_state.npcs.length <= spell['max_threshold'] : true }
                    .select { |spell| spell['expire'] ? Flags["ct-#{spell['abbrev']}"] : true }
+                   .select { |spell| !game_state.is_offense_allowed? ? spell['harmless'] : true }
```
KatoakDR, line 1667: Thanks for accepting my suggestion on line 1453, I believe the same can be applied here, too?
```suggestion
                   .select { |spell| game_state.is_offense_allowed? || spell['harmless'] }
```

wstampley:
An example of the shock guard:
```
[combat-trainer]>target FS 10
With tense movements you prepare your body for the Fire Shards spell.  Casting rationality aside, you prepare to yoke unseen mana streams to fuel your sorcery.
You are distinctly aware that completion of this spell pattern may bring you shock, and you reflexively become slightly queasy.
You begin to weave mana lines into a target pattern around an Adan'f shadow mage.
>
[combat-trainer: Dropping spell: Got shock warning.]
[combat-trainer]>release spell
The shadow mage points its staff at an Adan'f shadow mage, and the sockets of the skull glow with black light.
>
You let your concentration lapse and feel the spell's energies dissipate.
>
[combat-trainer]>release mana
```

combat-trainer.lic
```diff
@@ -3614,6 +3625,20 @@ class GameState
     !@clean_up_step.nil?
   end
 
+  def is_offense_allowed?
+    return true unless @is_empath
+    return true if @construct_mode
+    return true if @is_permashocked
```
KatoakDR, line 3631: Any value in reusing `is_permashocked?` here?
```suggestion
    return true if is_permashocked?
    return true if @construct_mode
```

wstampley:
Testers were requested on the 2nd: https://discord.com/channels/745675889622384681/745676101392793651/849865463910891520

No one has mentioned any issues, and neither my empath nor my other characters have had any problems running this version.

PR #4993, opened 2021-06-06
[script] [hunting-buddy] Add option for max searches for a room

KatoakDR:
_(Note, per feedback the option to hunt with people not in your buddies list has been removed. The description has been updated to that effect.)_

### Background
* When `hunting-buddy` cannot find a suitable hunting room then it will go back to your safe room and wait 20-40 seconds then return to the area and look for a room again.
* This loop runs forever until a suitable room opens up, which could take several minutes or never happen.
* It would be nice to have an upper limit to the number of times the script will scan an area for a suitable room before giving up. Secondary and tertiary hunting zones can be configured for other places to check per the `zone:` property on `hunting_info:` as shown in [Dartellum's example](https://github.com/rpherbig/dr-scripts/pull/4993#issuecomment-855422270).

### Changes
* Update `common-travel`'s `find_empty_room` method to
  - accept a new `max_search_attempts` parameter (number, default to infinity) and after N passes of the hunting zone and no suitable room has been found then break out of the loop
  - pass a counter as new argument to the predicate so that it can take into consideration how many times the character has tried to search for a suitable room (i.e. is this the first pass through the area? the second? the third?)
* Update `hunting-buddy`'s `find_hunting_room?` method to
  - include comments in the predicate lambda for its true/false decisions
  - `return` the result of calling `find_empty_room(..)` to fix bug that when no suitable room is found (after exhausting max searches or if no idle room) that you'd begin hunting in the first room of the hunting area no matter what. With this fix you'll return to your safe room and hunting-buddy will exit rather than hunting in a random room.

### Configuration
_base.yaml_
```yaml
# When unable to find a room to hunt in,
# hunting-buddy will run back to a safe room and wait 20-40 seconds
# before returning to the hunting zone to look for a room again.
# How many times should hunting-buddy revist the area to look for
# a suitable hunting room before ultimately giving up?
# Note, this is not the number of physical rooms to be checked.
# This is the number of times to wait then revisit the area to find a room.
hunting_room_max_searches: 30
```

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -208,10 +208,16 @@ class HuntingBuddy
     else
       @exit = nil
       find_empty_room(rooms, waiting_room,
-                      lambda do
+                      lambda do |search_attempt|
+                        # Skip room if has more people than your max limit
                         return false if DRRoom.pcs.size > @hunting_buddies_max
+                        # Skip if one of your nemesis is in the room
                         return false if (DRRoom.pcs & UserVars.hunting_nemesis).any?
+                        # Continue if one of your friends is in the room
                         return true if (DRRoom.pcs & UserVars.friends).any?
+                        # Continue if this is not the first search attempt and character will settle for "good enough" room
+                        return true if search_attempt > 1 && !@settings.hunting_room_strict_empty_or_buddies
+                        # Skip if anyone not in your group is in the room
                         return false if (DRRoom.pcs - DRRoom.group_members).any?
                         # No visible friends in the room and no visible people
                         UserVars.friends.each { |friend| Flags.add("room-check-#{friend}", friend) }
@@ -232,7 +238,9 @@ class HuntingBuddy
                         true
                       end,
                       @settings.hunting_room_min_mana,
-                      @settings.hunting_room_strict_mana)
+                      @settings.hunting_room_strict_mana,
+                      @settings.hunting_room_max_searches
+                    )
     end
 
     true
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -207,7 +207,7 @@ class HuntingBuddy
       @exit = [escort_info['area'], 'exit']
     else
       @exit = nil
-      find_empty_room(rooms, waiting_room,
+      return find_empty_room(rooms, waiting_room,
                       lambda do |search_attempt|
                         # Skip room if has more people than your max limit
                         return false if DRRoom.pcs.size > @hunting_buddies_max
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -208,39 +208,37 @@ class HuntingBuddy
     else
       @exit = nil
       return find_empty_room(rooms, waiting_room,
-                      lambda do |search_attempt|
-                        # Skip room if has more people than your max limit
-                        return false if DRRoom.pcs.size > @hunting_buddies_max
-                        # Skip if one of your nemesis is in the room
-                        return false if (DRRoom.pcs & UserVars.hunting_nemesis).any?
-                        # Continue if one of your friends is in the room
-                        return true if (DRRoom.pcs & UserVars.friends).any?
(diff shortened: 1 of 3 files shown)


Hiinky:
I'm not quite sure how I feel about this edit.  While yes it could help lichies find rooms, it may also bring an incredible amount of negative backlash from the larger DR community when lichies suddenly roll into any room ever.  I am VERY concerned this is going to generate lich-hate when we've always gone out of our way to ease the friction where we can.

hunting-buddy.lic
```diff
@@ -207,11 +207,17 @@ class HuntingBuddy
...
+      return find_empty_room(rooms, waiting_room,
+                      lambda do |search_attempt|
+                        # Skip room if has more people than your max limit
                         return false if DRRoom.pcs.size > @hunting_buddies_max
+                        # Skip if one of your nemesis is in the room
                         return false if (DRRoom.pcs & UserVars.hunting_nemesis).any?
+                        # Continue if one of your friends is in the room
                         return true if (DRRoom.pcs & UserVars.friends).any?
+                        # Continue if this is not the first search attempt and character will settle for "good enough" room
+                        return true if search_attempt > 1 && !@settings.hunting_room_strict_empty_or_buddies
```
wstampley, line 219: It would be helpful for me and the way I handle getting my attention if something non-standard is going on, if this echo'd something when it makes this choice (like the mana one does) so that I could then know I should say something to the random people i'm now hunting with.

bastionwt:
> I'm not quite sure how I feel about this edit.  While yes it could help lichies find rooms, it may also bring an incredible amount of negative backlash from the larger DR community when lichies suddenly roll into any room ever.  I am VERY concerned this is going to generate lich-hate when we've always gone out of our way to ease the friction where we can.

Absolutely this. There are now plenty of hunting zones at almost every rank range. This is a very bad direction to head in that will increase animosity towards players running Lich, not reduce it. 

If you are in a spot where you need to relax criteria, evaluate the creatures you are hunting and choose another in the range.

Dartellum:
Also, those with multiple areas set just for this reason will not work:
```yaml
hunting_info:
- :zone:
  - xalas_1 #750-1000
  - xalas_1 #750-1000
  - juvenile_wyverns #700-1100
```

This example will try xalas twice then head to wyverns. We already have an overflow, or no room found, mechanism in place. I know at lower levels the areas are limited, especially for F2P. On my F2P guys, I have the same area set multiple times.

KatoakDR:
Thanks everyone for sharing your feedback and concerns and alternatives to my proposed changes. 

I’ll close this pull request.

hunting-buddy.lic
```diff
@@ -207,32 +207,38 @@ class HuntingBuddy
       @exit = [escort_info['area'], 'exit']
     else
       @exit = nil
-      find_empty_room(rooms, waiting_room,
```
KatoakDR, line 210: Enable "Hide whitespace changes" when reviewing this diff otherwise it looks like more changed than what actually did.

The lambda didn't need to be indented so far to the right and that's the whitespace changes.

hunting-buddy.lic
KatoakDR: Per feedback, I've removed this aspect from the pull request to relax criteria to hunt with people not explicitly listed in your buddies list.

KatoakDR:
@Hiinky @Dartellum @bastionwt @wstampley Per your all's feedback, I've removed the "easy" option to hunt with characters not in your buddies list. This pull request is now focused on adding the max search limit to give players control of how long they'll wait for a particular hunting zone to open up before moving on to check other zones (if configured like https://github.com/rpherbig/dr-scripts/pull/4993#issuecomment-855422270)

Would appreciate your all's thoughts on this new direction.

Thanks

rpherbig:

Let me know how you want to proceed

KatoakDR:
> 
> Let me know how you want to proceed

I'm good with this going out. I have no further changes to make unless anyone has more feedback. Thanks

PR #4999, opened 2021-06-08
Support Worn Trashcans

wstampley:
Adding common-support so that scripts can use worn trashcans if they want to.

Opening this PR as a draft so people with trashcans can test it. I don't own one, personally, but it has been mentioned several times and Skinz requested it again recently for lockpicking speedup (not dismantling).

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -253,7 +253,7 @@ module DRCI
     return if item.nil?
 
     if worn_trashcan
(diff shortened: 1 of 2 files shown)


KatoakDR (approved):
✅ Code looks good to me. However, I also don't have a portable worn trashcan to test with

rpherbig:
Code looks good to me ✔️

wstampley:
This version isn't causing my characters any issues (without a trashcan), and Skinz's testing with one has gone well. Of course, scripts that want to take advantage of this behavior will have to be updated to opt in (by passing settings).

PR #5000, opened 2021-06-08
Vault room name is wrong for Shard

wstampley:
```
>go door
[Shard, Carousel Chamber]
A sturdy vault, set in the back wall...
```


KatoakDR (approved):
Thanks

PR #5003, opened 2021-06-10
[script] [common-healing] Fix parsing for 'tail' bleeders

KatoakDR:
### Background
* Reported by Moiyra on Lnet that `tendme` was not tending their tail bleeders.
* Upon investigation, the `tail` body part was not part of any of the regex patterns (oops!)

### Changes
* Add `tail` to the bleeder regex body part patterns

## Tests
See `test/test_check_health.rb`, and also Moiyra confirmed the fix.

### DIFF
--- a/common-healing.lic
+++ b/common-healing.lic
@@ -194,7 +194,7 @@ module DRCH
   # are the severity and the values are the list of bleeding wounds.
   def parse_bleeders(health_lines)
     bleeders = Hash.new { |h, k| h[k] = [] }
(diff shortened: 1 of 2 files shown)


wstampley:
This is probably an issue for the perc health methods too...

KatoakDR:
> This is probably an issue for the perc health methods too...

Perceive health delegates to the same parse bleeders logic that check health does, so we win there! :)

And the regex patterns for perceive health don’t specify body part explicitly like the parse bleeder regex did so we might win there, too

PR #5006, opened 2021-06-10
[script] [steal] buff after removing gear; use DRCI methods

### LINKED ISSUE #2439: Stealing: Should use waggle_set.
Refactor stealing to use a waggle_set.

KatoakDR:
### Background
* I noticed that my invisibility was released when the script was removing my gear, which then hampered my stealth and I got caught stealing. 😢 

### Changes
* Swap order of buffing and removing gear so that gear is taken off first, then you buff.
* General script updates to use new DRCI methods and add DRC* prefixes to methods.
* Refactored buff logic to use `DRCA.do_buffs(..)`
* Fixes #2439 - adds support for `steal` waggle set with fallback to the original `stealing_buffs` setting.

### Configuration
```yaml
waggle_sets:
  # Due to some rooms preventing movement while invisible,
  # the script may fail to travel to some shops. As a workaround,
  # don't use an invisibility buff and instead set `hide_to_steal: true`
  steal:
  - Plunder
  - Darken

hide_to_steal: true

# The stealing buffs setting is supported for backwards compatibility.
stealing_buffs:
  # Thieves would specify their khri
  khri:
  - Plunder
  - Darken
  # Magic users would specify their spells
  spells:
  - Shadows
```

### DIFF
--- a/steal.lic
+++ b/steal.lic
@@ -55,28 +55,30 @@ class Steal
       return
     end
 
+    # Do buffs after swapping gear otherwise equipmanager
+    # may cause you to drop interferring buffs like Khri Silence.
     @equipment_manager = EquipmentManager.new
     @equipment_manager.empty_hands
     @equipment_manager.wear_equipment_set?('stealing')
-    if left_hand || right_hand
+    if DRC.left_hand || DRC.right_hand
       DRC.message('*** ITEMS ARE STILL IN HANDS, EXITING ***')
       @equipment_manager.wear_equipment_set?('standard')
       return
     end
 
-    # Do buffs after swapping gear otherwise equipmanager
-    # may cause you to drop interferring buffs like Khri Silence
-    stealing_buffs = settings.stealing_buffs
-    do_buff(stealing_buffs, settings)
+    # Backwards compatibility with `stealing_buffs` setting.
+    stealing_buffs = (DRStats.thief? ? settings.stealing_buffs['khri'] : settings.stealing_buffs['spells'])
+    settings.waggle_sets['steal'] ||= stealing_buffs
+    DRCA.do_buffs(settings, 'steal')
 
     start_script('performance', ['noclean']) unless settings.hide_to_steal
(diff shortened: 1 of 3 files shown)


rpherbig:

steal.lic
```diff
@@ -112,13 +111,13 @@ class Steal
 
   def do_buff(buffs, settings)
```
wstampley, line 112: Should this just be DRCA.do_buff?

steal.lic
```diff
@@ -408,20 +372,20 @@ class Steal
       update_target(target, 'exceptionally') if count == 1
       bin_item(target['item'])
     when /but decide you have too many outstanding fines as is/
-      beep
+      DRC.beep
       echo('Too many fines, cannot steal anything')
```
wstampley, line 376: Anything worth beeping should probably use DRC.message for more visibility (since I know for me at least, beep doesn't work).

steal.lic
KatoakDR: I think so; I hadn’t thought about it since I was focused on other things in this PR. I can test that

steal.lic
KatoakDR: Agree, beep doesn’t work for me either

KatoakDR: I replaced the beeps with DRC.message

steal.lic
KatoakDR: I switched the buffing to DRCA.do_buffs

KatoakDR:
I've pushed some changes to respond to feedback and more testing. This is ready for final review, thanks

rpherbig:
Speaking of `stealing_buffs`, I'd be all for deprecating it in favor of the waggle set you just added support for.

KatoakDR:
> Speaking of `stealing_buffs`, I'd be all for deprecating it in favor of the waggle set you just added support for.

Cool -- I kept support because I didn't want to "pull the rug" out from under people if they were relying on that setting. I'll put another update to begin warning people if they are using the old setting and not using a waggle and then after some time we can remove the backwards compatibility support altogether.

PR #5023, opened 2021-06-18
[Update] [weave.lic] Changed weave to use hometown setting. 

Dartellum:
Added all information to base-crafting for loom-rooms and made definition to find empty room with loom.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -78,6 +78,16 @@ module DRCC
     end
   end
 
+  def find_loom_room(hometown, override = nil)
+    if override
+      DRCT.walk_to(override)
+    else
+      loomrooms = get_data('crafting')['tailoring'][hometown]['loom-rooms']
+      idle_room = get_data('crafting')['tailoring'][hometown]['idle-room']
+      DRCT.find_sorted_empty_room(loomrooms, idle_room)
+    end
+  end
+
   def find_shaping_room(hometown, override = nil)
     if override
       DRCT.walk_to(override)
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -82,9 +82,9 @@ module DRCC
     if override
       DRCT.walk_to(override)
     else
-      loomrooms = get_data('crafting')['tailoring'][hometown]['loom-rooms']
+      loom_rooms = get_data('crafting')['tailoring'][hometown]['loom-rooms']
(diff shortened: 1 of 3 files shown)


common-crafting.lic
```diff
@@ -78,6 +78,16 @@ module DRCC
     end
   end
 
+  def find_loom_room(hometown, override = nil)
+    if override
+      DRCT.walk_to(override)
+    else
+      loomrooms = get_data('crafting')['tailoring'][hometown]['loom-rooms']
+      idle_room = get_data('crafting')['tailoring'][hometown]['idle-room']
+      DRCT.find_sorted_empty_room(loomrooms, idle_room)
```
KatoakDR, line 87: Disclaimer, I'm not familiar with this script. If there isn't an idle room configured, should it fall back to your safe room?

rpherbig: @BinuDR @Dartellum - you two have contributed to this script, what do you think?

common-crafting.lic
```diff
@@ -78,6 +78,16 @@ module DRCC
     end
   end
 
+  def find_loom_room(hometown, override = nil)
+    if override
+      DRCT.walk_to(override)
+    else
+      loomrooms = get_data('crafting')['tailoring'][hometown]['loom-rooms']
```
rpherbig, line 85: `loom_rooms`

data/base-crafting.yaml
```diff
@@ -467,16 +467,20 @@ tailoring:
     - 16661
     - 16663
     sewing-rooms:
-    - 19_037
-    - 19_036
-    - 16_670
+    - 19037
```
rpherbig, line 470: The underscore here is a Rubocop style thing.

Which reminds me, we should probably run Rubocop on the repo sometime soon. I just don't have it/Ruby installed on any machines at the moment.

common-crafting.lic
Dartellum: I just cut and pasted `def find_shaping_room` and changed `shapeingrooms` to `loomrooms`. Changed!

common-crafting.lic
Dartellum: We can add an idle room then. I will add the room right outside the weave rooms.

Dartellum: Updated.

PR #5025, opened 2021-06-19
Improving common-arcana's handling of khri 

wstampley:
Several issues were discovered during testing with the new pick (which checks buffs more regularly). This will make it easier to use khri in multiple ways through the multiple entry points (combat-trainer, buff, etc.)

- format tolerant
- common usability
- improved detection for already running to prevent unnecessary retreat / kneeling

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2058,19 +2058,18 @@ class AbilityProcess
       fput action
       waitrt?
     end
+
     @khri
       .map { |name| "Khri #{name}" }
       .each do |name|
         next if game_state.danger && DRCA.kneel_for_khri?(@kneel_khri, name)
-        timer = game_state.cooldown_timers[name]
-        next unless !timer || (Time.now - timer).to_i > 30
         if DRCA.activate_khri?(@kneel_khri, name)
-          game_state.cooldown_timers[name] = Time.now
           echo "Activated khri #{name}" if $debug_mode_ct
         else
           echo "Did not activate khri #{name}" if $debug_mode_ct
         end
       end
+
     @barb_buffs
       .select { |name| Flags["ap-#{name}-expired"] }
       .reject { |name| DRSpells.active_spells[name] }
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2063,7 +2063,11 @@ class AbilityProcess
       .map { |name| "Khri #{name}" }
       .each do |name|
         next if game_state.danger && DRCA.kneel_for_khri?(@kneel_khri, name)
+        timer = game_state.cooldown_timers[name]
+        next unless !timer || (Time.now - timer).to_i > 30
+
         if DRCA.activate_khri?(@kneel_khri, name)
+          game_state.cooldown_timers[name] = Time.now
           echo "Activated khri #{name}" if $debug_mode_ct
         else
           echo "Did not activate khri #{name}" if $debug_mode_ct
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -70,10 +70,10 @@ module DRCA
     abilities = abilities.drop(1) if should_delay
 
     # Check each khri in the list against Active Spells, Drop any that are active
-    needed_abilities = abilities.select { |ability| DRSpells.active_spells["Khri #{ability}"].nil? }
+    needed_abilities = abilities.select { |ability_to_check| DRSpells.active_spells["Khri #{ability_to_check}"].nil? }
     return false if needed_abilities.empty?
 
-    kneel = needed_abilities.any? { |ability| kneel_for_khri?(settings_kneel, ability) }
+    kneel = needed_abilities.any? { |ability_to_check| kneel_for_khri?(settings_kneel, ability_to_check) }
     DRC.retreat if kneel
     DRC.bput('kneel', 'You kneel', 'You are already', 'You rise') if kneel && !kneeling?
 
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -81,7 +81,7 @@ module DRCA
     waitrt?
     DRC.fix_standing
 
-    ['Your body is willing', 'You have not recovered'].include?(result)
+    ['Your mind and body are willing', 'Your body is willing', 'You have not recovered'].exclude?(result)
   end
 
   def kneel_for_khri?(kneel, ability)
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -71,7 +71,7 @@ module DRCA
 
     # Check each khri in the list against Active Spells, Drop any that are active
     needed_abilities = abilities.select { |ability_to_check| DRSpells.active_spells["Khri #{ability_to_check}"].nil? }
-    return false if needed_abilities.empty?
+    return true if needed_abilities.empty?
 
     kneel = needed_abilities.any? { |ability_to_check| kneel_for_khri?(settings_kneel, ability_to_check) }
     DRC.retreat if kneel
@@ -81,7 +81,7 @@ module DRCA
     waitrt?
     DRC.fix_standing
 
-    ['Your mind and body are willing', 'Your body is willing', 'You have not recovered'].exclude?(result)
+    return !['Your mind and body are willing', 'Your body is willing', 'You have not recovered'].include?(result)
   end
 
   def kneel_for_khri?(kneel, ability)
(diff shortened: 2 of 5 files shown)


KatoakDR:
Code looks good to me. I’ll try it out tonight-tomorrow. 

Would you mind adding tests where multiple khri try to be activated together? For example, mock Khri Hasten is active/inactive and do`DRCA.activate_khri?(false, 'Khri Hasten Focus')`

wstampley:
> Code looks good to me. I’ll try it out tonight-tomorrow.
> 
> Would you mind adding tests where multiple khri try to be activated together? For example, mock Khri Hasten is active/inactive and do`DRCA.activate_khri?(false, 'Khri Hasten Focus')`

Yeah. I'm still adding test cases, that's why it is still in draft.

KatoakDR:
Thanks for the optimizations! So far so good 👍 

```yaml
waggle_sets:
  default:
  - Focus Hasten
  - Avoidance Harrier
  - Dampen
```


```
> khri harrier
> 
With mental effort, you center your mind on improving your physical prowess.
Roundtime: 3 sec.

>;buff

--- Lich: buff active.

[buff]>Khri Focus Hasten
> 
Focusing intently, you mentally weave together the two mantras until their essences flow through you, increasing your abilities.

With deep breaths, you recall your training and focus your mind to hone in on improving your physical dexterity.
Centering your mind, you allow your practiced discipline to spread throughout your body, making thought and motion one.
Roundtime: 5 sec.
> 
```
_Note in the next activation it skips Harrier because it's already active_ 👍 
```
[buff]>Khri Avoidance

Taking a deep breath, you focus on making your mind and body one, your mental discipline trained on quickly noticing, analyzing, and dodging approaching threats.
Roundtime: 3 sec.
> 
[buff]>Khri Dampen

Focusing your mind, you attempt to attune your senses to even your most minor movements.  You quickly feel more comfortable with your ability to control your motions to prevent detection.
Roundtime: 3 sec.
> 
--- Lich: buff has exited.
```

rpherbig:
Code looks good to me ✔️

wstampley:
Looks like fixing the return value of the activate_khri function (true when it actually activates one like it seems the intent was) triggers some buff timer/logic that shouldn't really be in combat trainer anymore.

Should probably wait a little bit while I do some more testing here.

KatoakDR:
> Looks like fixing the return value of the activate_khri function (true when it actually activates one like it seems the intent was) triggers some buff timer/logic that shouldn't really be in combat trainer anymore.
> 
> Should probably wait a little bit while I do some more testing here.

Should the last line of `activate_khri?` return the opposite of what it currently does?
```ruby
['Your body is willing', 'You have not recovered'].include?(result)
```
This looks like to me that it returns `true` if you're NOT able to activate the khri but we should be returning `true` if the result does NOT include those strings, right?

wstampley:
> > Looks like fixing the return value of the activate_khri function (true when it actually activates one like it seems the intent was) triggers some buff timer/logic that shouldn't really be in combat trainer anymore.
> > Should probably wait a little bit while I do some more testing here.
> 
> Should the last line of `activate_khri?` return the opposite of what it currently does?
> 
> ```ruby
> ['Your body is willing', 'You have not recovered'].include?(result)
> ```
> 
> This looks like to me that it returns `true` if you're NOT able to activate the khri but we should be returning `true` if the result does NOT include those strings, right?

Yep. But when I fix that (and the tests for it), it starts triggering a block of combat-trainer, that looks like it sets a timer on all khri starts for some reason.

combat-trainer.lic
```diff
@@ -2063,7 +2063,11 @@ class AbilityProcess
       .map { |name| "Khri #{name}" }
       .each do |name|
         next if game_state.danger && DRCA.kneel_for_khri?(@kneel_khri, name)
+        timer = game_state.cooldown_timers[name]
+        next unless !timer || (Time.now - timer).to_i > 30
+
```
KatoakDR: What happened that the timers were needed? I would think that `spellmonitor` keeping track of your abilities and `DRCA.activate_khri?` using `DRSpells.active_spells` would mean it'd know if the abilities were up or not.

KatoakDR: Or, is this more of a "don't check to rebuff more of than every 30 seconds"?

wstampley: I think this is added primarily to keep it from re-checking too often for things that succeeded (especially for people who run 'delay <whatever>' in their khri for the slow start + better effect versions).

It might be interesting to try and figure out a 'pending spells' mechanic somewhere in the system so that activate khri w/ delay could say "consider this running for 30 seconds until it really is running", but for now, this timer fixes it for combat, and most people aren't spam running buff so it doesn't matter there.

common-arcana.lic
```diff
@@ -46,23 +46,42 @@ module DRCA
...
-    return false if DRSpells.active_spells[ability]
+    abilities = ability.split.map(&:capitalize)
 
-    kneel = kneel_for_khri?(settings_kneel, ability)
+    # Standardize for with/without 'Khri' on the front
+    abilities = abilities.drop(1) if abilities.first.casecmp('khri') == 0
+
+    # Handling for 'Delay'
+    should_delay = abilities.first.casecmp('delay') == 0
+    abilities = abilities.drop(1) if should_delay
```
KatoakDR, line 70: This may make as a good utility method. I could see this being re-used in `combat-trainer` at line 2062 to split out the khris specified in the `buff_nonspells.khri` setting, particularly if people specify "khri delay xxx" or "khri xxx yyy zzz".

common-arcana.lic
```diff
@@ -46,23 +46,42 @@ module DRCA
...
     DRC.retreat if kneel
     DRC.bput('kneel', 'You kneel', 'You are already', 'You rise') if kneel && !kneeling?
-    result = DRC.bput(ability, get_data('spells').khri_preps)
+
+    result = DRC.bput("Khri #{should_delay ? 'Delay ' : ''}#{needed_abilities.join(' ')}", get_data('spells').khri_preps)
     waitrt?
     DRC.fix_standing
 
-    ['Your body is willing', 'You have not recovered'].include?(result)
+    return !['Your mind and body are willing', 'Your body is willing', 'You have not recovered'].include?(result)
```
KatoakDR, line 84: 6 of one, half dozen of another. What are your thoughts on changing `include?` to `none?` ?
```suggestion
    return ['Your mind and body are willing', 'Your body is willing', 'You have not recovered'].none?(result)
```

common-arcana.lic
KatoakDR, line 70: Lines 63-70 may make as a good utility method. I could see this being re-used in `combat-trainer` at line 2062 to split out the khris specified in the `buff_nonspells.khri` setting, particularly if people specify "khri delay xxx" or "khri xxx yyy zzz".

common-arcana.lic
wstampley: I don't see that on this version of ruby:
https://ruby-doc.org/core-2.5.5/Array.html

Am I just missing it?

common-arcana.lic
wstampley: the lines from buff_nonspells.khri end up passing through this code anyway.

The .map in combat_trainer that pins 'khri' on the front of everything could probably be removed as it shouldn't be needed for the timer key, and isn't needed as an input to this method anymore.

common-arcana.lic
KatoakDR: It’s inherited from the Enumerable parent class: https://ruby-doc.org/core-2.5.5/Enumerable.html#method-i-none-3F

I missed it originally too; those docs don’t make it clear what methods are available due to inheritance, you have to go to the parent classes to find them :(

wstampley: Well, that's decidedly annoying. It shows some methods, but not all of them - made me assume all that worked were documented. I'll poke around at it.

wstampley:
@rpherbig - I think this is ready.

KatoakDR (approved):
Been working well for me for the past few days with the following:

```yaml
waggle_sets:
  combat:
  - Strike Steady
  - Terrify
  - Focus Hasten
  - Avoidance Harrier
  - Dampen
  pick:
  - Focus Hasten Safe
  steal:
  - Plunder Darken
 ```

rpherbig:
@wstampley the code looks good - can you fix the merge conflict?

wstampley:
> @wstampley the code looks good - can you fix the merge conflict?

This is fixed.

PR #5034, opened 2021-06-23
[script] [common-travel] delete flags when no longer needed

KatoakDR:
### Background
* Inspired by conversation with @wstampley 
* Typical scripts run then exit, and any flags in them can be deleted in a `before_dying` block.
* Modules that are loaded in memory and not "ran and exited" don't have this cleanup step, so flags need to be deleted manually.
* Deleting unused flags can improve performance by reducing the amount of processing over received messages

### Changes
* Move declaration of the travel flags in `walk_to` method to the place where they're actually needed, by-passing several `return` statements that exit the method and wouldn't have cleaned up the flags.
* At end of method, delete the travel flags tracking if a shop is closed or you're in combat.

## Tests
* I tested while engaged with ship rats -- able to retreat then move to destination.
* I tested in Arthe Dale trying to enter room 19237 in the evening when it's closed up.

### DIFF
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -86,9 +86,6 @@ module DRCT
   end
 
   def walk_to(target_room, restart_on_fail = true)
-    Flags.add('travel-closed-shop', 'The door is locked up tightly for the night', 'You smash your nose', '^A servant (blocks|stops)')
-    Flags.add('travel-engaged', 'You are engaged')
-
     target_room = tag_to_id(target_room) if target_room.is_a?(String) && target_room.count("a-zA-Z") > 0
 
     return false if target_room.nil?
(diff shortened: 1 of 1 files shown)


common-travel.lic
```diff
@@ -118,13 +115,25 @@ module DRCT
...
       # Shop closed, stop script and open door then restart
       if Flags['travel-closed-shop']
         Flags.reset('travel-closed-shop')
         kill_script(script_handle)
         if /You open/ !~ DRC.bput('open door', 'It is locked', 'You .+', 'What were')
-          return false
+          # You cannot get to where you want to go,
+          # and no amount of retries will get you through the locked door.
+          restart_on_fail = false
+          break
```
KatoakDR, line 136: To still hit the code at the end of this method that deletes the flags, I changed the `return` to just don't restart on fail and then break out of the loop. Line 171 will now _not_ execute (retry movement) and the method will still return false because you're not at your destination.

rpherbig:
> Deleting unused flags can improve performance

These common files are kept around in memory long-term and there are only two flags here. Is this a noticeable performance change?

KatoakDR:
> > Deleting unused flags can improve performance
> 
> These common files are kept around in memory long-term and there are only two flags here. Is this a noticeable performance change?

Noticeable? Doubtful. 

Originally we thought the leftover engaged flag was keeping Sekmeht in a loop trying to retreat from spiders under Theren’s keep,  but turns out that you just can’t retreat there and you’re stuck, with or without this change. 

This PR is not a “must have”. If any concerns, I can close it. 

@wstampley any feedback from your perspective?

Thanks

wstampley:
> This PR is not a “must have”. If any concerns, I can close it.
> 
> @wstampley any feedback from your perspective?

I like the cleanup, personally. Every stranded flag is another regex on every line of text you get from the game.

KatoakDR:
@rpherbig this is ready for your final review, thanks!

PR #5040, opened 2021-06-26
[data] [base-spells] Remove duplicate `harmless: true` entry

KatoakDR:
### Background
* The spell `Compel` has two entries for `harmless: true`, only one is needed

### Changes
* Remove the duplicate `harmless: true` from the `Compel` spell definition


wstampley (approved):

wstampley:
Oops?

KatoakDR:
> Oops?

No worries, it was _harmless_ 😉

PR #5042, opened 2021-06-27
[common-crafting] [Update] Added .* to line 232 to catch extra long brazier names.

Dartellum:
Per https://discord.com/channels/745675889622384681/745762938991804556/858509717092433920, needed to add the extra .* to catch names like `a potbellied silver brazier`. Alternative pointed it out and will test. 

I also changed to the regex method instead of single or double-quotes.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -229,8 +229,8 @@ module DRCC
         order_enchant(stock_room, stock_needed, stock_number, bag, belt)
       end
     when /I could not find what you were referring to./
-      case bput('tap my fount on my brazier', 'You tap .* atop a brazier.', 'I could not find what you were referring to.')
-      when /You tap (.*) atop a brazier./
+      case bput('tap my fount on my brazier', /You tap .* atop a .* brazier./, /I could not find what you were referring to./)
+      when /You tap (.*) atop a (.*) brazier./
         /(\d+)/ =~ bput('analyze my fount on my brazier', 'This appears to be a crafting tool and it has approximately \d+ uses remaining')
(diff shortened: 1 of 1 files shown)


Dartellum:
Alternative tested and said it was working: https://discord.com/channels/745675889622384681/745762938991804556/858519173563285524.

wstampley (approved):

common-crafting.lic
```diff
@@ -229,8 +229,8 @@ module DRCC
         order_enchant(stock_room, stock_needed, stock_number, bag, belt)
       end
     when /I could not find what you were referring to./
-      case bput('tap my fount on my brazier', 'You tap .* atop a brazier.', 'I could not find what you were referring to.')
-      when /You tap (.*) atop a brazier./
+      case bput('tap my fount on my brazier', /You tap .* atop a .* brazier./, /I could not find what you were referring to./)
```
wstampley, line 232: Do you actually want `You tap .* atop a .*brazier` so that it matches both:
- You tap a burin atop a fancy brazier
- You tap a burin atop a brazier
I don't enchant, but if this was working for some people before, with the second string, it looks like this would break them.

rpherbig: This concerns me too.

`atop a .* brazier` won't match `atop a brazier` because of the space

rpherbig:
Code changes look good aside from the one comment ✔️

common-crafting.lic
Dartellum: The regex tester worked with a space or no space. :man_shrugging: I removed the space.

PR #5045, opened 2021-06-27
Script and example profile settings to run yaml autostarts.

robbintt:
Add a script to autostart scripts according to a YAML configuration.

When added to dependency autostart, this script creates an alternative path for autostart scripts to be configurable as code, per character, in `profiles/my.yaml`.

FUTURE: integrate this more tightly into dependency.

### DIFF
--- a/autostart_yaml_scripts.lic
+++ b/autostart_yaml_scripts.lic
@@ -1,0 +1,23 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#autostart_yaml_scripts
+=end
+
+custom_require.call(%w[common])
+
+no_kill_all
+no_pause_all
+
+class AutostartYamlScripts
+  def initialize
+    settings = get_settings
+    @autostart_scripts = settings.autostart_scripts
+    @autostart_scripts_delay = settings.autostart_scripts_delay
+
+    pause @autostart_scripts_delay
+    for s in @autostart_scripts
+      start_script(s)
+    end
+  end
+end
+
+AutostartYamlScripts.new
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -5,6 +5,15 @@ hometown: Crossing
 # Cap on time crossing-training runs
 training_manager_town_duration:
 
+# delays yaml autostart scripts on startup
+autostart_scripts_delay: 5
+# autostart scripts can also be specified in the YAML
+autostart_scripts:
+  - spellmonitor
+  - drinfomon
+  - sanowret-crystal
+  - skill-recorder
+
 # T2 Settings
 training_list:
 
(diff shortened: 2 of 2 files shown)


vtcifer:
The idea of migrating autostarts into yamls, and away from the db is probably in general a good idea, however I'm not sure this PR is ready for merging:
- The new autostart script still needs to be added to an autostart (dependency's autostart or it will fail because it makes use of dr-scripts calls) but isn't managed in any way with this PR.
- It doesn't support global vs character autostarts which is a loss of functionality comparatively 
- Further complicates the _already_ confusing autostart situation for people ( https://xkcd.com/927 )

Dartellum:
I use a separate script that autostarts all the stuff I want and is called by the global autostart after dependency starts. The script also keeps the stuff running if it exits for some reason. 

Example:
`start_script('jail-buddy') if not Script.running?('jail-buddy')`

wstampley:
My two cents - 

I would love to have a more configurable autostart mechanism - that can handle dependencies or readiness checks (some scripts don't run well until other things are up and ready), Dart's optional restarting if things drop for some reason, etc.

I think I would happily give up "do this for all my characters" in order to get it, since it's easy to copy paste a block of text into a YAML.

KatoakDR:
I agree with what everyone has said above.

I also like the idea behind this pull request -- moving the autostart config to yaml. We field questions pretty often from people having issues with their autostarts, and it doesn't help that Lich has it's own "autostart" function and our fork has a different mechanism that's also called "autostart" (e.g. `;autostart` vs. `;e autostart`)

Since `dependency` is already doing the effort to run autostarts configured in the database, and it loads up the yamls, I _think_ we could add the hook to dependency's [SetupFiles.run_queue](https://github.com/rpherbig/dr-scripts/blob/master/dependency.lic#L770) method. By that time the core scripts should be loaded and the yaml's will be loaded. That avoids people needing to add anything to their `;e autostart` -- just add things to your yaml and they'll run. Another advantage is that the yaml could support arguments being passed to scripts ;)

rpherbig:
I'm worried about merging this in, only to have yet another way to do it in the future. I'm inclined to wait on this PR until it integrates with `dependency`. Thoughts?

vtcifer:
Just reiterating my thoughts, but I agree with your worries.  IMO if we're going to introduce a new way of doing autostarts, which people already get confused by, it should replace the old way fully, before it gets merged.

rcuhljr:
Yes to manage autostarts via yaml instead of `;e autostart` :100:. Agree on killing off the old one though as part of this.

MahtraDR:
Hi @robbintt Do you plan to come back to this? Or shall we close it off? Thanks!

Kiriawen:
@robbintt @MahtraDR

Bump

MahtraDR:
Closing this one out as abandoned.

PR #5046, opened 2021-06-28
Esp script for activating a gweth.

robbintt:
Whenever putting on or reactivating a gweth, too many commands are required. This script does the work.

### DIFF
--- a/esp.lic
+++ b/esp.lic
@@ -1,0 +1,13 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#esp
+=end
+custom_require.call(%w[common])
+
+bput("esp listen general", "With a moment of focus")
+bput("esp listen trade", "With a moment of focus")
+bput("esp listen personal", "With a moment of focus")
+bput("esp listen guild", "With a moment of focus")
+bput("esp listen race", "With a moment of focus")
+bput("esp listen local", "With a moment of focus")
+
+bput("esp send general", "You prepare to project")


esp.lic
```diff
@@ -0,0 +1,13 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#esp
+=end
+custom_require.call(%w[common])
+
+bput("esp listen general", "With a moment of focus")
```
wstampley, line 6: Should the channels be in base, and overwritable, or does everyone tune most of these channels?

KatoakDR: Agree, all the channels should be listed in `base.yaml` with option for users to override the list in their yaml.

Something like:
```yaml
esp_listen_channels:
  - general
  - trade
  - personal
  ...

esp_default_channel: general
```

Any channels NOT listed in `esp_listen_channels` would explicitly be turned off when the script runs.

The last line in this script, `bput("esp send general", "You prepare to project")`, would use the setting `esp_default_channel`

KatoakDR (changes requested):

KatoakDR:
> Whenever putting on or reactivating a gweth, too many commands are required. This script does the work.

They do take an annoying amount of time to set back up, thanks for proposing this script. I left some comments towards refactoring to read from the character's yaml which channels they want on/off and which to set as the default.

Thanks

rpherbig:
@robbintt code changes look good to me ✔️ 

let me know what the next step is here

KatoakDR:
> @robbintt code changes look good to me ✔️
> 
> let me know what the next step is here

As a new script, I think it’s ok to merge as is. I’d prefer it to already have [some tweaks](https://github.com/rpherbig/dr-scripts/pull/5046#discussion_r659455560) to support activating channels from your yaml but that can definitely be added in a follow up PR.

PR #5047, opened 2021-06-28
Update dependency.lic

rcuhljr:
Github has a new token format so I'd like to switch to it for this and eventually delete the old token. We should figure out the right way to do this in the future and stop being hacky.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -154,7 +154,7 @@ class ScriptManager
 
   def initialize(debug)
     @debug = debug
-    @git_token = '4491b4c6' + '20864db1575' + '3542a5d8729cf362acd1c'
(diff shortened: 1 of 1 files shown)


rpherbig:
The change seems reasonable. I'm not really sure of a better way to do this, though.

KatoakDR:
As long as we're using a read only token only for `dependency` to download files, this may be fine (it's what I assume has been used for years prior).

The GitHub API does allow for unauthenticated calls, https://docs.github.com/en/rest/guides/getting-started-with-the-rest-api#authentication, and that reduces the rate limit per IP address per hour to 60 down from 5,000 if authenticated.

I have not tested without using a token if the APIs we need still work, though.

KatoakDR (approved):
🔑 works for me

PR #5048, opened 2021-06-28
dependency.lic updates:

### LINKED ISSUE #4958: Deprecated ostruct method in use in dependency.lic
As of 2.4.0 (ruby/ruby@7c89ca5) an internal OpenStruct method (`modifiable`) made private, but that commit also aliased the old method the new public method. 
Those aliases were removed in 2.7.0 (ruby/ruby@dbcc224)

This breaks this line of code:
https://github.com/rpherbig/dr-scripts/blob/56d3afd732fa8390057f14e4054a370b8adf30bc/dependency.lic#L140

Most people should be on a higher version than 2.3.0 (2.5.5 is current version with the all-in-one), but we should update to be within spec / future compatibility. 

This fixes in my test, and works on both 2.5.5, and 2.7.3 
```diff
+140      if !respond_to?(:modifiable?, true) # for backwards compatibility <2.3
-140      modifiable[new_ostruct_member(mname)] = args[0]
+141        modifiable[new_ostruct_member(mname)] = args[0]
+142      else                                # new >=2.3 method
+143        modifiable?[new_ostruct_member(mname)] = args[0]
+144      end
```

I have no pre-2.3 version to test, and no environment to commit for the PR at this time.

vtcifer:
- handle deprecated ostruct method (closes #4958)
 - Introduce min ruby version
   - prohibits new install on versions below min ruby version
   - warn users (weekly) if version lower than min
 - bump dependency minor version due to above
 - change developer toggles to prevent accidently turning it off and losing work
 - create toggle for ignoring just dependency update: far easier to test changes during install
 - added additional info on number of steps during install process
 - also removed go2 from repsoitory updates to prevent overwrites of our own version
 - removed hardcoded lich script start charcters in favor of variable to handle different FEs
 - additional output when running help_me to direct to lnet window for link

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -1244,12 +1244,14 @@ end
 
 def set_dev_mode
   echo "Now turning dev mode on"
-  echo "Developer mode now set to: #{$manager.developer = true}"
+  $manager.developer = true
+  echo "Developer mode now set to: #{$manager.developer}"
 end
 
 def unset_dev_mode
   echo "Now turning dev mode off"
-  echo "Developer mode now set to: #{$manager.developer = false}"
+  $manager.developer = false
+  echo "Developer mode now set to: #{$manager.developer}"
 end
 
 def check_dev_mode
@@ -1262,12 +1264,14 @@ end
 
 def ignore_dependency
   echo "Now ignoring dependency updates"
-  echo "ignore_dependency now set to: #{$manager.ignore_dependency = true}"
+  $manager.ignore_dependency = true
+  echo "ignore_dependency now set to: #{$manager.ignore_dependency}"
 end
 
 def unignore_dependency
   echo "Now processing dependency updates"
-  echo "ignore_dependency now set to: #{$manager.ignore_dependency = false}"
+  $manager.ignore_dependency = false
+  echo "ignore_dependency now set to: #{$manager.ignore_dependency}"
 end
 
 def check_dependency
--- a/dependency.lic
+++ b/dependency.lic
@@ -1392,7 +1392,7 @@ def help_me(user = nil)
   result = $manager.submit_help_pastebin
   if user
     LNet.send_message({ 'type' => 'private', 'to' => user }, result)
-    echo("Attempted to  PM YAML link on lnet to #{user}.  Check lnet window for success or failure.")
+    echo("Attempted to PM YAML link on lnet to #{user}.  Check lnet window for success or failure.")
   else
     LNet.send_message({ 'type' => 'channel' }, result)
     echo('YAML link sent to lnet.  Check lnet window.')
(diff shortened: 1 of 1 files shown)


wstampley (approved):

dependency.lic
```diff
@@ -1280,8 +1392,10 @@ def help_me(user = nil)
   result = $manager.submit_help_pastebin
   if user
     LNet.send_message({ 'type' => 'private', 'to' => user }, result)
+    echo("Attempted to  PM YAML link on lnet to #{user}.  Check lnet window for success or failure.")
```
KatoakDR, line 1395: 💅 I tried, I really tried to ignore the extra whitespace 😅 
```suggestion
    echo("Attempted to PM YAML link on lnet to #{user}.  Check lnet window for success or failure.")
```

KatoakDR (approved):
Looking forward to using the `set_ignore_dependency` option, I've needed that a couple times this week 😁

dependency.lic
```diff
@@ -190,8 +196,32 @@ class ScriptManager
     @autostarts = UserVars.autostart_scripts + Settings['autostart']
   end
 
-  def toggle_developer
-    @developer = !@developer
+  def set_developer
+    @developer = true if !@developer
+    @developer
```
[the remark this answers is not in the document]
vtcifer: TBQH not entirely sure.   I thought I had copied earlier logic and just updated, and not being a ruby expert by any means left the same, but now I'm not 100% sure.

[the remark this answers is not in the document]
vtcifer: I could probably just make them accessors.

`attr_accessor :developer, :ignore_dependency`

rpherbig:
@rcuhljr want to take a look at this?

dependency.lic
```diff
@@ -1220,28 +1242,65 @@ def stop_autostart(script)
   end
 end
 
+def set_dev_mode
+  echo "Now turning dev mode on"
+  echo "Developer mode now set to: #{$manager.developer = true}"
```
rpherbig, line 1247: I'm not a fan of doing the assignment in the string interpolation, but it won't hold up merging.

dependency.lic
```diff
@@ -1220,28 +1242,65 @@ def stop_autostart(script)
...
+end
+
+def unset_dev_mode
+  echo "Now turning dev mode off"
+  echo "Developer mode now set to: #{$manager.developer = false}"
```
rpherbig, line 1252: I'm not a fan of doing the assignment in the string interpolation, but it won't hold up merging.

rpherbig:

Waiting on @rcuhljr and I'll merge this

dependency.lic
vtcifer: I don't mind changing this, as I agree it's awkward. Was really just a result of couple of naive find->replace when changing those from custom methods to accessors.

PR #5052, opened 2021-06-29
[script] [wandwatcher] add cli options

vtcifer:
- add cli option to override start up delay from yaml
 - add option to clear all timing variables for wands

### DIFF
--- a/wand-watcher.lic
+++ b/wand-watcher.lic
@@ -12,12 +12,32 @@ class WandWatcher
   include DRCI
 
   def initialize
+
+    arg_definitions = [
+      [
+        {
+          name: 'delay', 
+          regex: /^\d+$/, 
+          optional: true, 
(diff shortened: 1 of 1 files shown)


KatoakDR (approved):
Looks good to me ☑️

wand-watcher.lic
```diff
@@ -12,12 +12,32 @@ class WandWatcher
...
+    arg_definitions = [
+      [
+        {
+          name: 'delay', 
+          regex: /^\d+$/, 
+          optional: true, 
+          description: 'Override wand_watcher_startup_delay from yaml' 
+        },
+        {
+          name: 'reset',
```
KatoakDR, line 25: What’s a use case that someone would want/need to do this? Just curious if this is something I may need to add to my routine or not. Thanks

vtcifer: If timers get out of sync for some reason.

PR #5053, opened 2021-06-30
[script] [common-arcana] Add option for barbs to sit to meditate

KatoakDR:
### Background
* Unless you have the Yogi mastery, Barbarians must sit to meditate
* Attempting to meditate while not sitting will waste your inner fire (sad trombone)

### Changes
* Add new setting `sit_to_meditate: boolean` for players to indicate if they want to sit to meditate or not (modeled after the `kneel_khri` setting for Thieves). As new behavior, this defaults to `false` so players must opt-in by setting it to `true`.
* Update `common-arcana`'s `activate_barb_buff?` to accept a new third, optional, parameter to denote if should sit or not.

### Configuration
```yaml
# Barbarian only - kneel to activate barbarian meditation
# With the Yogi mastery, you can meditate while standing
sit_to_meditate: true
```

### Tests
* See `test/test_common_arcana.rb`

### DIFF
--- a/test/test_common_arcana.rb
+++ b/test/test_common_arcana.rb
@@ -1,6 +1,5 @@
 require_relative 'test_helper'
 require 'yaml'
-require 'ostruct'
 
 load 'test/test_harness.rb'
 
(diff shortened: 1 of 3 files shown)


test/test_common_arcana.rb
```diff
@@ -10,19 +9,28 @@ class TestDRCA < Minitest::Test
...
   def setup
     reset_data
+    load_base_settings
   end
 
   def teardown
     @test.join if @test
   end
 
+  def load_base_settings(settings = {})
```
wstampley, line 19: Should this have gone in the base harness?

KatoakDR: Good point, I’ll move it. There’s a couple other tests where this method or similar exist in them too, I’ll evaluate refactoring them. Thanks

PR #5064, opened 2021-07-05
[script] [esp] configure channels via yaml, add recovery logic

KatoakDR:
### Background
* Following up on https://github.com/rpherbig/dr-scripts/pull/5046#pullrequestreview-693522835

### Changes
* Convert script to a class named `ESP`
* Read from YAML which channels to subscribe you to
* Add error recovery and handling logic if you're not wearing a gweth or it breaks
* Handle that `esp listen <channel>` toggles on/off you listening to that channel so the command may need to be run twice to get your preference configured

### Configuration
```yaml
# ESP Settings
# List the channels you want to subscribe to when `esp` script runs.
# You'll unsubscribe from any channels not listed.
esp_channels:
  - General
  - Trade
  - Guild
  - Race
  - Local
  - Private
  - Personal

# Which channel to set you up to send to by default.
esp_default_channel: General
```

### DIFF
--- a/esp.lic
+++ b/esp.lic
@@ -13,10 +13,11 @@ class ESP
     channels = settings.esp_channels.map { |channel| channel.downcase.capitalize }
     default_channel = settings.esp_default_channel
 
-    activate_gweth
-    listen_to_these_channels_and_not_others(channels)
-    set_default_send_channel(default_channel)
-    show_channel_subscriptions
+    if activate_gweth?
+      listen_to_these_channels_and_not_others(channels)
+      set_default_send_channel(default_channel)
+      show_channel_subscriptions
+    end
   end
 
   def show_channel_subscriptions
@@ -49,10 +50,11 @@ class ESP
     end
   end
 
-  def activate_gweth(retry_count = 0)
+  def activate_gweth?(retry_count = 0)
+    return false
     case DRC.bput("touch my gweth", { 'timeout' => 3, 'suppress_no_match' => true }, "A chorus of foreign thoughts joins your own", "Since you already have telepathy, nothing more happens", "Touch what?", "The chain snaps in two places and falls to the ground")
     when /A chorus of foreign thoughts joins your own/, /Since you already have telepathy, nothing more happens/
-      true
+      return true
     when /Touch what?/
       DRC.message("You don't have a gweth!")
       exit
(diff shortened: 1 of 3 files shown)


KatoakDR:
@robbintt Thanks for your earlier submision of the `esp` script. Here's my proposal to refactor it to read from YAML settings and such. Would like your review when you get a chance. Thanks

esp.lic
```diff
@@ -1,13 +1,75 @@
...
+
+  def listen_to_these_channels_and_not_others(channels)
+    @all_channels.each do |channel|
+      listen = channels.include?(channel)
+      update_channel_subscription(channel, listen)
+    end
+  end
+
+  def update_channel_subscription(channel, listen)
+    # We don't have an easy way to know whether you are or are not currently
```
rpherbig, line 38: This feels like a good place for a UserVar

esp.lic
```diff
@@ -1,13 +1,75 @@
...
+    when /Touch what?/
+      DRC.message("You don't have a gweth!")
+      exit
+    when /The chain snaps in two places and falls to the ground/
+      DRC.message("Uh oh, your gweth broke!")
+      exit
+    else
+      # Not wearing a gweth, try to wear one and retry
+      if retry_count < 1
+        DRCI.get_item_if_not_held?('gwethdesuan')
```
rpherbig, line 65: An opportunity for an early exit if this returns false?

esp.lic
```diff
@@ -1,13 +1,75 @@
...
+        DRCI.wear_item?('gwethdesuan')
```
rpherbig, line 66: An opportunity for an early exit if this returns false?

esp.lic
```diff
@@ -1,13 +1,75 @@
...
+      listening = Regexp.last_match[1] == 'open'
+      if listen != listening
+        update_channel_subscription(channel, listen)
+      end
+    else
+      DRC.message("Unrecognized channel: #{channel}")
+    end
+  end
+
+  def activate_gweth(retry_count = 0)
```
rpherbig, line 52: `activate_gweth?` maybe?

esp.lic
```diff
@@ -1,13 +1,84 @@
...
+      if listen != listening
+        update_channel_subscription(channel, listen)
+      end
+    else
+      DRC.message("Unrecognized channel: #{channel}")
+    end
+  end
+
+  def activate_gweth?(retry_count = 0)
+    return false
```
rpherbig, line 54: Is this leftover test code?

esp.lic
```diff
@@ -1,13 +1,84 @@
...
+      return true
+    when /Touch what?/
+      DRC.message("You don't have a gweth!")
+      exit
+    when /The chain snaps in two places and falls to the ground/
+      DRC.message("Uh oh, your gweth broke!")
+      exit
+    else
+      # Not wearing a gweth, try to wear one and retry
+      if retry_count < 1
```
rpherbig, line 66: If there is really only going to be a single retry, I'd prefer this to be a boolean

esp.lic
KatoakDR: > Doesn't this negate the purpose of the function? Is this leftover dev/test code?

Doh! Residue from earlier testing (sigh)

esp.lic
KatoakDR: changed to boolean

PR #5069, opened 2021-07-08
[script] [combat-trainer] charged maneuvers don't have ramp time anymore

KatoakDR:
### Background
* http://forums.play.net/forums/DragonRealms/Combat%20-%20Weapons%20and%20Armor/Game%20Master%20and%20Official%20Announcements/thread/1941605?get_newest=true

### Changes
* Remove the 7 second pause before performing the combat maneuver since it's not needed anymore


rpherbig:
🎉

PR #5044, opened 2021-06-27
[script] [combat-trainer] New option: prioritize maneuver doublestrike

KatoakDR:
### Background
* `combat-trainer` already prioritizes the doublestrike maneuver for Barbarians that are whirlwinding since they're actively holding two melee weapons.
* For characters that are not whirlwinding (regardless their guild), there lacks an option to prioritize the doublestrike maneuver.
* I have a desire to doublestrike as much as possible for...
    - the fun of it
    - kill critters fast to fuel my inner fire
    - I don't have whirlwind and would like to use this maneuver to gain EXP in multiple melee weapons with one attack similar to how aiming_trainables lets you gain EXP in two skills when you're training a ranged weapon

### Changes
* Add setting `prioritize_maneuver_doublestrike: boolean`
* Update method `charged_maneuver` to also choose doublestrike when this setting is true and you're training a melee weapon skill
* Add method `determine_doublestrike_skill` that, similar to `determine_aiming_skill`, picks an appropriate skill from your aiming_trainables list as what to equip offhand
* Update the `attack_melee` method that when the doublestrike maneuver should be performed to equip a suitable melee weapon in  your offhand, perform the maneuver, then put away the offhand weapon

### Configuration
```yaml
# If true then whenever you're able to perform a doublestrike maneuver, do so :)
# If false then will perform a specific manuever based on the equipped mainhand weapon.
prioritize_maneuver_doublestrike: true

training_abilities:
  # How often you want to attempt a charged combat maneuver.
  # https://elanthipedia.play.net/Combat_maneuvers#Charged_Maneuvers
  Charged Maneuver: 60 

# When you can do doublestrike maneuver then a non-thrown, non-ranged, non-brawling skill
# will be chosen from this list to temporarily equip in your offhand for the strike.
aiming_trainables:
  - Small Edged
  - Small Blunt
  - Light Thrown
  - Heavy Thrown
  - Staves
  - Polearms
  - Brawling
```

### Example
<details>
<summary>click to view log snippet</summary>

```
[combat-trainer]>wield my throwing.club

You draw out your throwing club from the armor pack, gripping it firmly in your right hand.

... attack with club for a bit ...

... time to do maneuver doublestrike! ...

[combat-trainer]>wield my witchclaw.pilum
You draw out your witchclaw pilum from the armor pack, gripping it firmly in your left hand.

[combat-trainer]>maneuver doublestrike
You step to the side and adjust your stance.

... charge up ...

You turn to face a zombie boar.
[...]

PR #5070, opened 2021-07-09
[script] [combat-trainer] Advanced Combat Maneuvers

KatoakDR:
## Dependencies
* Depends on https://github.com/rpherbig/dr-scripts/pull/5076

### Background
* "Charged Combat Maneuvers" were changed to "Advanced Combat Maneuvers" ([forum post](https://elanthipedia.play.net/Post:Charged_Combat_Maneuvers_-_07/07/2021_-_00:32), [wiki page](https://elanthipedia.play.net/Combat_maneuvers#Advanced_Combat_Maneuvers)).
* Without the need for the ramp up time https://github.com/rpherbig/dr-scripts/pull/5069, they are more viable to be part of regular combat routines, especially for Barbarians who have high likelihoods of reducing the 90 second cooldown to 60 seconds.
* Continuation of https://github.com/rpherbig/dr-scripts/pull/5044, which introduced prioritizing doublestrikes maneuvers.

### Changes
* Add new setting `doublestrike_trainables` to define which mainhand weapon skills you want to use when attempting doublestrikes. This lets you, for example, opt out of trying (and failing) doublestrikes with staves if you use a quarterstaff (two-handed) instead of a short staff (one-handed). Same with polearms like halberds vs. light spears.
* For the true combat maneuver enthusiast, including suggested YAML configuration below for how to perform maneuvers frequently.
* When prioritizing doublestrikes, if that specific maneuver is on cooldown then fallback to trying the specific maneuver for the mainhand weapon, and if that maneuver is on cooldown then do regular attack.
* When tracking cooldowns for each combat maneuver, detect when Barbarians reduce their cooldown so that the maneuver will be attempted sooner rather than later.
* Support dual loading with powershot maneuver. Powershot will only fire one arrow from a dual loaded bow, so in these scenarios to not waste the other arrow, the script will perform the second shot immediately.
* Format `@skill_map` to be (subjectively) more readable and maintainable.

### Configuration
<details>
<summary>toggle sample configuration</summary>

```yaml
training_abilities:
  Charged Maneuver:
    # By default, maneuvers are performed as long as the Expertise skill needs training.
    # To make combat maneuvers a regular attack in your routine, set the :check: property to a
    # non-existant skill that'll never reach the value of combat_training_abilities_target.
    :check: Always
    # Also, since every maneuver type has its own cooldown (e.g. you can 'cleave' then 'crash'
    # back to back but then have to wait for their respective cooldowns before you can do them again),
    # set the training ability cooldown to a lowish value so that the script is regularly performing
    # maneuvers for your equipped weapons as soon as they are off cooldown.
    # This example checks every 10 seconds, and if a maneuver is ready then it's performed.
    # This increases the chances of performing a maneuver, especially when you swap to a new weapon.
    :cooldown: 10

# If "Charged Maneuver" is set in your training_abilities, this setting
# will attempt doublestrikes by temporarily wielding an "aiming trainable"
# in your offhand when your holding a "doublestrike trainable" in your main hand.
# If this is false then doublestrikes are only attempted if you are a Barbarian using whirlwind.
# I recommend enabling this, especially if you already use aiming trainables with ranged weapons,
# because it increases the training of multiple weapons at the same time.
prioritize_maneuver_doublestrike: true

# When prioritizing doublestrike maneuvers, this list indicates which
# melee, one-handed weapon skills in your mainhand signal that
# it's ok to attempt doublestrikes. If a doublestrike is attempted then
# a weapon from your aiming trainables will be equipped offhand.
#
# This setting is necessary if you use twohanded staves or polearms so
# that you can opt them out of doublestrike attempts because they will
# fail because the weapon itself requires two hands.
# If you're able to use the weapon for an aiming trainable then it's
# suitable to be a doublestrike trainable. This separate setting gives you control,
# however, to use different weapons for aiming and doublestrike purposes.
doublestrike_trainables:
  - Small Edged
  - Small Blunt
  - Large Edged
  - Large Blunt
  # - Staves
  # - Polearms

# In addition to being the skills trained while aiming ranged weapons,
# when prioritizing doublestrike maneuvers, one of the
# non-thrown, one-handed melee weapons in this list will be
# equipped temporarily in your offhand to do the strike.
aiming_trainables:
  - Small Edged
  - Small Blunt
  - Large Edged
  - Large Blunt
  - Light Thrown
  - Heavy Thrown
  # - Staves
  # - Polearms
  - Brawling
```
</details>

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -269,17 +269,17 @@ class SetupProcess
     if !last_summoned
       @equipment_manager.stow_weapon(game_state.last_weapon_name)
       game_state.sheath_whirlwind_offhand
-      # Reset state that was specific to the prior weapon.
-      # This prevents, for example, attempting to do a powershot maneuver
-      # as soon as you equip another ranged weapon before you've had
-      # a chance to load it because the game_state thinks you're still holding a loaded weapon.
-      game_state.loaded = false
-      game_state.selected_maneuver = nil
     elsif !next_summoned && !DRStats.moon_mage?
       break_summoned_weapon(game_state.last_weapon_name)
     elsif !next_summoned && DRStats.moon_mage?
       DRCMM.wear_moon_weapon?
     end
+
+    # Reset state that was specific to the prior weapon.
+    # This prevents, for example, attempting to do a powershot maneuver
+    # as soon as you equip another ranged weapon before you've had
+    # a chance to load it because the game_state thinks you're still holding a loaded weapon.
+    game_state.loaded = false
 
     # Prepare the next weapon
     if next_summoned
@@ -2875,6 +2875,7 @@ class AttackProcess
     Flags.add('ct-ranged-ready', 'You think you have your best shot possible now.')
     Flags.add('ct-face-what', 'Face what')
     Flags.add('ct-maneuver-cooldown-reduced', "With expert skill you end the attack and maneuver into a better position")
+    Flags.add('ct-attack-out-of-range', "You aren't close enough to attack")
   end
 
   def execute(game_state)
@@ -2925,6 +2926,8 @@ class AttackProcess
 
   def attack_melee(charged_maneuver, game_state)
     waitrt?
+
+    maneuver_success = false
     if charged_maneuver
       # If you're prioritizing doublestrike but only wielding one melee weapon
       # then temporarily wield a second weapon offhand to perform the maneuver.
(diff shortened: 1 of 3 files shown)


rpherbig:
@jayRyan24 can you take a look at this?

jayRyan24:
Looks solid to me.

KatoakDR:
I ran into a snag with repeating crossbows and when combat-trainer would try to perform `maneuver powershot`. Rather than kludge a fix, I'm going to submit a second PR to improve repeating crossbow support then come back to this PR. I figure it'd be better to review/approve/merge those changes separately since they aren't really related.

rpherbig:
Thank you for the separate PR

KatoakDR:
This is ready for final review, thanks

PR #5074, opened 2021-07-10
[script] [combat-trainer] Invoke current weapon in case have multiple bonded items

KatoakDR:
### Background
* Reported by [Tekhelet](https://discord.com/channels/745675889622384681/745675890242879671/863456621823524914) in the Lich discord
* Tekhelet has two bonded weapons and since combat-trainer only performs `invoke` without specifying the bonded item then the game randomly decides which to return to their hands, which may be the wrong weapon.
* Having the wrong item returned causes the combat routine to go awry.

### Changes
* Update `thrown_retrieve_verb` method to specify the weapon name when using the invoke verb

### Tests
* Awaiting Tekhelet's results as I don't have any bonded items to test with myself

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4177,7 +4177,7 @@ class GameState
 
   def thrown_retrieve_verb
     if bound_weapon?
-      'invoke'
+      "invoke #{weapon_name}"
     else
       "get my #{weapon_name}"
     end
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4177,7 +4177,7 @@ class GameState
 
   def thrown_retrieve_verb
     if bound_weapon?
-      "invoke #{weapon_name}"
+      "invoke bond #{weapon_name}"
     else
       "get my #{weapon_name}"
     end


rpherbig:
Looks good to me ✔️

KatoakDR:
The plot thickens.... in the official discord, [GM Zadraes](https://discord.com/channels/619301383451181075/813953565705371728/863651373958299668) gave this clarification regarding `invoke bond`:
> You can't specify a weapon with INVOKE BOND...or you can, but it doesn't do anything.  When a bonded item leaves your possession, a tracker gets added to you pointing to the weapon.  When you INVOKE BOND, the first tracker is ID'd (likely the last item to leave your possession) and the item belonging to that tracker is initialized.

Need to further investigate if the real solution to this issue is `invoke bond <weapon>` or if the logic on when to use invoke in our script needs to change or if there's actually a bug in the game and we need GM support.

KatoakDR:
[Tekhelet](https://discord.com/channels/619301383451181075/813953565705371728/863654826772922368) confirmed that `invoke bond <weapon>` actually doesn't work the way we thought it was and Zadraes' comment holds.
> Looks like you CAN'T invoke a particular weapon. If I drop them both, it's invoking the first one dropped first, then the second.

KatoakDR:
Tekhelet is going to do some more investigating into his game logs and what's going on when combat-trainer invokes the seemingly wrong weapon into his hand. If we can figure out the root cause then we can revisit what needs to change. For now, will close this PR.

PR #5078, opened 2021-07-12
[script] [outdoorsmanship] support training perception using collect

vtcifer:
- Updates outdoorsmanship to allow for training perception

### DIFF
--- a/outdoorsmanship.lic
+++ b/outdoorsmanship.lic
@@ -12,6 +12,7 @@ class Outdoorsmanship
   def initialize
     arg_definitions = [
       [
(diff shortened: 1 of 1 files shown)


KatoakDR (approved):
Looks good to me

vtcifer:
As a note, this is WILDLY inefficient, but wanted to add the option for non-combat perception training.  Right now crossing-training uses it, but if your outdoorsmanship locks before perception (which it probably will due to exp allocation) this causes problems.  

For most who also train in combat this isn't a huge deal.

Next step is updating crossing-training to use this, and after that I've got an idea for a better way to train perception based on hometown, with this as a fallback.

PR #5079, opened 2021-07-12
[DRAFT] Add spellbook utility and add unleash feature to prepare? method. These will probably be split up.

robbintt:
Clean up some rough edges and split up this PR.

1. Determine how to add a yaml for unleash:

```
buffs:
  prehunt_buffs:
    Sentinel's Resolve:
      abbrev: sr
      recast: 2
      cambrinth:
        - 20
      unleash_item: flame-hued scroll  # need safety 'glance scroll' match for Sentinel's Resolve
      unleash_container: spidersilk pouch
      # or
      unleash_item: diamondwood spellbook
      unleash_container: spellbook case
      unleash_mana: 15
```

2. Update unleash logic so that it is more context free. Today the standard is that you can cast_spell() and common-arcana.lic will do everything for you, and that means item handling. So we will need to do item handling and spellbook handling inside the unleash context.  For spellbook users, this will add ;go2 to their waggle_sets and a `spellbook_safe_room` yaml variable. Spellbooks cannot get wet, so it's better to not reuse any other rooms that could be outdoors.

3. Add unleash tattoo logic (see other tattoo logic and just update the data map.
4. probably just pass `data` into prepare?
  - A very long laundry list of data properties are now being passed and it makes more sense just to give prepare? the payload.

5. spellbook: allow specifying more parameters as arguments for one-offs or script-based usage (but still prefer waggle sets).

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -137,8 +137,32 @@ module DRCA
     activated
   end
 
-  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false, runestone_name = nil, runestone_tm = false)
+  def prepare?(settings, abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false, runestone_name = nil, runestone_tm = false, unleash_item = false)
     return false unless abbrev
+
+    # minimum settings to handle unleash items
+    # 1. yaml entry of unleash spells, their items, and containers: unleash: list of spells with parameters: item, container
+    # 2. safety function for unleashing the right scroll by "glance #{unleash.spell.unleash_item} in #{unleash.unleash_container}"
+    # 3. option to integrate spellbook with unleash at that point with like "unleash.spell.is_spellbook = true"
+    # 4. special spellbook logic from this
+    if unleash_item
+      # TODO: This makes me think that any ritual spell should have a manifest with: container, item...
+      # TODO: maybe there should be a helper to get a scroll with a particular read, as well, or at least safety check the read
+      if not DRCI.in_hands?(unleash_item)
+        # TODO: this is not normal, if we can get the logic inside this module, then these items can be used in buff scripts
+        # TODO: however, one excellent case for this is the spellbook, which requires travel, special logic, and more yaml settings
+        # TODO: Spellbook management is not a good fit for this module, but maybe it should be called in this clause
+        echo("*** WARNING: Unleash item management must be handled outside of the cast_spell function.")
+        return false
+      end
+      cast_action = "cast my #{unleash_item}" # nb: ruby doesn't interpolate inside the data payload
+      # TODO: Unleash can be optimized based on the cost of the spell, requires an overhaul to discern logic based on skill+spell
+      # TODO: add a prep action for tattoo from settings
+      unleash_data = {"abbrev" => "unleash", "mana" => 15, "cast" => cast_action}
+      cast_spell(unleash_data, settings)
+      DRC.bput('prepare symbiosis', 'You recall the exact details of the', 'But you\'ve already prepared') if symbiosis
+      return true
+    end
 
     DRC.bput('prepare symbiosis', 'You recall the exact details of the', 'But you\'ve already prepared') if symbiosis
 
@@ -150,7 +174,7 @@ module DRCA
     case match
     when 'Your desire to prepare this offensive spell suddenly slips away'
       pause 1
-      return prepare?(abbrev, mana, symbiosis, command, tattoo_tm, runestone_name, runestone_tm)
+      return prepare?(settings, abbrev, mana, symbiosis, command, tattoo_tm, runestone_name, runestone_tm, unleash_item)
     when 'Something in the area interferes with your spell preparations', 'You shouldn\'t disrupt the area right now', 'You have no idea how to cast that spell', 'You have yet to receive any training in the magical arts'
       DRC.bput('release symbiosis', 'You release the', 'But you haven\'t') if symbiosis
       return false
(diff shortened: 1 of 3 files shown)


common-arcana.lic
```diff
@@ -137,9 +137,33 @@ module DRCA
...
+        # TODO: however, one excellent case for this is the spellbook, which requires travel, special logic, and more yaml settings
+        # TODO: Spellbook management is not a good fit for this module, but maybe it should be called in this clause
+        echo("*** WARNING: Unleash item management must be handled outside of the cast_spell function.")
+        return false
+      end
+      cast_action = "cast my #{unleash_item}" # nb: ruby doesn't interpolate inside the data payload
+      # TODO: Unleash can be optimized based on the cost of the spell, requires an overhaul to discern logic based on skill+spell
+      # TODO: add a prep action for tattoo from settings
+      unleash_data = {"abbrev" => "unleash", "mana" => 15, "cast" => cast_action}
+      cast_spell(unleash_data, settings)
```
rpherbig, line 162: I'm not a fan of a code path that has `prepare?` call `cast_spell`

KatoakDR: Agree. I’m also not really following what this change is trying to do and why

KatoakDR: From reading https://elanthipedia.play.net/Unleash, it seems like you need to cast two spells: unleash and then the scroll spell. 

I recommend adding a new method that understands that’s what needs to happen and it’d make two calls to cast_spell — one for unleash and one for the scroll spell

Disclaimer, I don’t have unleash. This is my suggestion for how to reuse existing code rather than be casting spells in the prepare method.

spellbook.lic
```diff
@@ -0,0 +1,196 @@
...
+    weather = bput("weather", indoors_str, indoors_portal_str)
+    if weather == indoors_str or weather == indoors_portal_str
+      return true
+    end
+    return false
+  end
+
+  # this is mainly to help this script fail out
+  def check_holding_spellbook
+    hands = bput("inv held", @spellbook_name, "Both of your hands are empty.", "right hand", "left hand")
```
KatoakDR, line 94: You may be able to replace this method with `DRCI.in_hands?(@spellbook_name)`

spellbook.lic
```diff
@@ -0,0 +1,196 @@
...
+    return page_number
+  end
+
+  # TODO NB: It is not necessary to read the spellbook to unleash it.
+  def turn_to_spell_page(page_number)
+    bput("turn my #{@spellbook_name} to page #{page_number}", "You turn to page", "is already open to page")
+  end
+
+  def unleash_spellbook(data)
+    cast_spell(data, @settings)
```
KatoakDR, line 192: Cast unleash _then_ cast the scroll spell?

common-arcana.lic
```diff
@@ -137,9 +137,33 @@ module DRCA
     activated
   end
 
-  def prepare?(abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false, runestone_name = nil, runestone_tm = false)
+  def prepare?(settings, abbrev, mana, symbiosis = false, command = 'prepare', tattoo_tm = false, runestone_name = nil, runestone_tm = false, unleash_item = false)
```
KatoakDR, line 140: Why do you need the settings argument?

common-arcana.lic
robbintt: Can you elaborate? This is a very appropriate usage for cast spell.  The function contexts are separate so there is no scope concern.

spellbook.lic
robbintt: Unfortunately the `cast spellbook` action of unleash doubles as the prep action for the next spell. Because of the way common-arcana.lic is written, it would require some disruptive state tracking and/or rewriting to make a cast action of 1 spell the same action as the prep action for the next spell. Each has a lot of contexts.

common-arcana.lic
robbintt: It's needed for `cast_spell` and doesn't really have any negative effects. I am also planning to just pass data instead of all these parameters, that is more in-keeping with the rest of the library.

bastionwt:
How unleash works, for those who aren't MMs:

- prep unleash 100
- waitfor prep time
- cast SCROLL
- The spell that was in your scroll becomes your new already-being-prepped-wait-for-the-timer spell
- waitfor prep time
- cast SPELLNAME

common-arcana.lic
robbintt: Looks like prepare is called in many contexts.  I will probably prefer exposing settings globally in common-arcana.lic or whatever the accepted pattern is, rather than passing a new function parameter, although it may be better to revise all existing uses of `prepare?` to be of the form `prepare?(settings, data)` anyways, just like the calling function, so that less special context exists across other scripts.

common-arcana.lic
wstampley: Just my two cents - I almost feel like cast_spell (or somewhere up the chain) needs to be refactored to decide what kind of thing we're doing, and split out from there.

It currently looks like this one method would handle unleashing a tattoo or runestone, neither of which makes any sense, and it's just getting more and more bloat.

As to the fact that you can 'unleash and unleash scroll to unleash a <other> scroll', this feels like classic "you got so caught up in figuring out if you could, you never stopped to consider if you should" requirements.

robbintt: I may take on a cast spell refactor this week.

MahtraDR:
Hi @robbintt Do you plan to come back to this? Or shall we close it off? Thanks!

KatoakDR:
Closing due to inactivity. Feel free to reopen this pull request and address code review feedback if you'd like to pursue this further. Thanks!

PR #5083, opened 2021-07-13
[script] [echo] Code style only changes to match spellmonitor's convention

KatoakDR:
### Changes
* parens around arguments
* comments about infinite loop
* `before_dying` is a `do/end` block
* remove hook when script starts

### DIFF
--- a/echo.lic
+++ b/echo.lic
@@ -4,15 +4,19 @@
 
 echo_action = proc do |server_string|
   open('echo_log_room.txt', 'a') do |f|
-    f.puts server_string
+    f.puts(server_string)
   end
   server_string
 end
+
+DownstreamHook.remove('echo_action')
 DownstreamHook.add('echo_action', echo_action)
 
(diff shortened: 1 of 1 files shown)


echo.lic
```diff
@@ -4,15 +4,19 @@
...
 
-before_dying { DownstreamHook.remove('echo_action') }
+before_dying do
+  DownstreamHook.remove('echo_action')
+end
 
-loop do
-  pause 10
-  clear
+until script.gets.nil?
```
wstampley, line 19: Seems to me like there's a performance change with this code change. It looks like every script in lich is allocated a thread, and this is just to keep this "script" alive, even though it has no work to do (cause the real work is handled by the main thread in the DownstreamHook stuff)?

If that understanding is right, we switched from a thread that sleeps for 10 seconds, clears the scripts message queue, and then sleeps, to a thread that just spins pulling records off as fast as it can.

Can anyone verify if my understanding is right, and why we'd prefer one over the other?

PR #5091, opened 2021-07-13
[scripts] [common dependency validate] Enhance DRC.message functionality

vtcifer:
Adds support to allow to redirect output to customizable windows and highlight message based on FE presets
of DRC.message method.

This change introduces the following:
- Allows custom windows to be created, if the FE supports dynamic windows similar to SF and Genie.
- Allows modification of settings of the default windows (ex: allows to redirect output from one window to another)
- Allows DRC.message to direct it's output in (moderately) user customizable way.  (Existing scripts will need to be updated to take advantage of this).
  - If a script specifies a severity on DRC.message (and ensuring yaml settings for that severity exist) users can select which window and what kind of highlighting are applied to these messages.  
  - All DR preset highlighting are supported, but the FE must support this 
    - SF: as the Simutronics official one supports all
    - Genie: does not due to it's janky setup
    - Frostbite: Unknown
    - Profanity: Unknown
  - Scripts that have not been updated to take advantage of this will continue to operate as normal.

Notes:
- I'm not 100% on the naming of these settings/ methods / yaml setup.  If there are any suggestions for more clear naming/help text I'm more than willing to consider changes.  This is the best I could come up with but even then I'm not particularly happy with what I came up with.
- I'm not sure current changes to base-empty.yaml (adding default values instead of actual empty hash, and using merge in dependency) are the best way to accomplish having some defaults that users will not accidently remove completly.  Also open to suggestions here as well.
- As this is making some significant changes, having additional FE users test is strongly desired prior to merging.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -716,7 +716,9 @@ module DRC
 
   def message(text, severity = 'none')
     # legacy message format - not customizable
-    if severity == 'none'
+    # frostbite doesn't suppot extra user defined or system generated windows as of at least 1.12.0b
+    # also allow user selectable legacy mode  (current default)
+    if severity == 'none' || $frontend == 'frostbite' || $LEGACY_DRC_MESSAGE
       clean_text = clean_xml(text)
       string = ''
       $fake_stormfront ? string.concat("\034GSL\r\n ") : string.concat("<pushBold\/>")
@@ -732,7 +734,7 @@ module DRC
     else
       # variables have been cleared for some reason, try reloading
       if UserVars.message_extensions['severities'] == nil || UserVars.message_extensions['severities'].empty?()
-        reload_message_formats
+        reload_message_extensions
         if UserVars.message_extensions['severities'] == nil || UserVars.message_extensions['severities'].empty?()
           # if those are still failing fall back to legacy
           message(text,'none') 
@@ -770,6 +772,8 @@ module DRC
         if $frontend == 'genie'   #fix genie monsterbold
           str1 = str1 + "<pushBold/>" + clean_message
           str2 = "<popBold/>" + str2
+        elsif $frontend == 'profanity'  # profanity hack
+          str1 = str1 + "<pushBold/>" + clean_message + "<popBold/>"
         else
           str1 = str1 + "<preset id='bold'>" + clean_message + "</preset>"
         end
--- a/common.lic
+++ b/common.lic
@@ -714,17 +714,78 @@ module DRC
     DRC.bput("stance set #{stance}", /Setting your/)
   end
 
-  def message(text)
-    string = ''
-    $fake_stormfront ? string.concat("\034GSL\r\n ") : string.concat("<pushBold\/>")
-    string.concat("\r") if $frontend == 'genie'   #fix genie monsterbold
-    if text.index('\n')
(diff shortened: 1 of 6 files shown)


rpherbig:
How will this be tested?

vtcifer:
> How will this be tested?

Posted yesterday in discord looking for testers, hoping to get traction.

vtcifer:
This should be good to merge, assuming we're OK with the slight deviation in formatting / look and feel for base-empty.yaml

I've tested in Genie, SF,  Frostbite and ProfanityFE, support for this change boils down to the following:
- SF: Full support
  - All highlight options work
  - Full window creation/updates/redirects work
- Genie: Partial support
  - Not all highlighting options work, only roomName, thought, and bold/monster bold work.
  - Full window creation/updates/redirects work
- Profanity: Partial Support
  - All highlight options work
  - No dynamic window creation or redirection, windows have to be pre-defined in your ProfanityFE settings.xml
- Frostbite: no support, legacy mode only

SF with dynamically created log window showing utilization of presets highlights:
image

Genie with dynamically created log window showing supported preset highlights:
image

ProfanityFE with 'log' window set, and custom highlights defined in settings.xml:
image

Thought process on default highlighting:

info -> no highlighting (similar to echo)
warning -> bold (same as highlight message behavior for today)
severe / fatal -> I judged based off of "general" game reactions.  If someone is talking in the same room as me I'll pay a little more attention, if someone is whispering to me I'll pay a lot more attention.

vtcifer:
Thinking more on this, perhaps would be better to set those default settings in a base-xxxx.yaml in data, and pull them out of base-empty. 

Specifically:
```yaml
  base_message_extension: 
    windows:
      log:                        # required - window id 
        :title: Log               # optional - default: capitalization case of window id
        :location: center         # optional - default: center, [left,center,right]
        :styleIfClosed: ''        # optional - default: '', [select from presets - roomName, bold, monsterbold, speech, whisper, thought, watching, link, selectedLink, command, normal]
        :windowIfClosed: main     # optional - default: '', window id of where to send text if this window is closed
        :timestamp: true          # optional - default: true, use timestampes on this window
    severities:
      info:
        :window: log
        :highlight: none
      warning:
        :window: log
        :highlight: bold
      severe:
        :window: log
        :highlight: speech
      fatal:
        :window: log
        :highlight: whisper
```

Thoughts?

rpherbig:
The changes look good, but I don't have confidence in a visual inspection of this PR.

When you feel it's been tested, let me know and I'll merge it.

vtcifer:
Trying to troubleshoot a setting reload issue, that I'm really not sure what it's root cause is.

vtcifer:
I think it should be all fixed, and ready, but I'm going to let it soak for a bit before setting back to ready for review.

Dartellum:
With profanity, you can make dynamic windows work but you need to use GTK. Checkout `vars` and `narost` for GTK examples.

vtcifer:
It works fine in profanity without GTK, it's not a GUI window, but like combat / experience / perc etc.  Just needs to be defined in the layout.

MahtraDR:
Hi @vtcifer  Do you plan to come back to this? Or shall we close it off? Thanks!

BinuDR:
I'll test this in Frostbite and post results.

PR #5092, opened 2021-07-14
[script] [hunting-buddy] Option to stop when encumbered

KatoakDR:
### Background
* Some of my characters are weak and easily become encumbered in battle looting boxes, coins, gems, and bundles.
* Once they reach "Burdened" or higher then they really start taking a beating.
* I'd like to exit combat before I need to exit due to wounds.

### Changes
* New per-hunt setting, `stop_on_encumbrance`, to indicate at which encumbrance level you want `hunting-buddy` to end combat prematurely. The range is 0 (None) to 11 (It's amazing you aren't squashed!). Default is nil to maintain current behavior; players opt-in.
* Move the `need_boxes?` method lower in the file, closer to related methods.
* Because we only know your encumbrance by running the `encumbrance` command, put the check in the one-minute heartbeat section so as not to spam the output with encumbrance checks.

### Configuration
```yaml
hunting_info:
  - :zone:
      - rats
    # If specified then this is the encumbrance level at which you want hunting-buddy to abandon combat.
    # Being too encumbered hinders your defenses and increases your chances of receiving wounds.
    # Set a numeric value from 0 (None) to 11 (It's amazing you aren't squashed!)
    # https://elanthipedia.play.net/Encumbrance
    stop_on_encumbrance: 2
    ...
```


rpherbig:
Is there a case in which you'd want different encumbrance values for different hunts?

vtcifer:
Yeah, I feel like the simplest way of handling this settings wise is just a numeric threshold in the hunting_info: and set the default if not used to like 99, then you're always below the threshold.

KatoakDR:
> Is there a case in which you'd want different encumbrance thresholds for different hunts?

> Yeah, I feel like the simplest way of handling this settings wise is just a numeric threshold in the hunting_info: and set the default if not used to like 99, then you're always below the threshold.

ok

hunting-buddy.lic
```diff
@@ -252,11 +265,39 @@ class HuntingBuddy
     stop_on_low_skills && stop_on_low_skills.flatten.any? { |skill| DRSkill.getxp(skill) < @stop_on_low_threshold }
   end
 
+  # Is your current encumbrance at or higher than the threshold?
+  # Threshold is a number from 0 (None) to 11 (It's amazing you aren't squashed!)
+  def encumbered?
+    return false unless @hunting_encumbrance_threshold
+    fput('encumbrance') # drinfomon script will parse the output then update DRStats
```
wstampley, line 272: Should there be a function on DRStats to refresh your encumbrance, that this just calls?

hunting-buddy.lic
```diff
@@ -81,6 +77,7 @@ class HuntingBuddy
       stop_on_low_skills = info['stop_on_low']
       stop_on_boxes = info['boxes']
       stop_on_no_moons = info['moons']
+      stop_on_encumbrance = info['encumbrance']
```
wstampley, line 80: I see that you followed the pattern here for boxes/moons/etc, but it feels like those property names aren't super clear what they do to the average person.

Should we try to update the hunt config to something like:
```
  stop_on:
    low_skills:
    - Theurgy
    encumbrance: true
    max_boxes: false
 ```
 etc.?

File: hunting-buddy.lic (hunk 1 above)
Comment (KatoakDR) [reply hunting-buddy.lic:272]: Yep, I'll move that logic to drinfomon and silence the output so that DRStats can always return the current encumbrance value without this hack

File: hunting-buddy.lic (hunk 2 above)
Comment (KatoakDR) [reply hunting-buddy.lic:80]: Yeah, I had half a mind to add support for `stop_on_boxes` and `stop_on_moons` so it'd be clearer but then decided against it because of existing player base uses the ambiguous names already. Not an excuse for the new encumbrance setting to be that way, but as you noticed, I just went with the convention that had been set

File: hunting-buddy.lic [hunk 3]
```diff
@@ -52,11 +52,6 @@ class HuntingBuddy
     @hunting_info = format_hunting_info(hunting_info)
   end
 
-  def need_boxes?
-    return false unless @settings.box_hunt_minimum
-    DRCI.count_all_boxes(@settings) <= @settings.box_hunt_minimum
-  end
```
KatoakDR, line 58: Not deleted, just moved down to `over_box_limit?` so the box methods are co-located

hunting-buddy.lic
KatoakDR: I agree with this feedback and will follow up on how to do it. For the initial version of this change, I'd like to roll it out with the `encumbrance` command being called once a minute (current design). Amid all the other combat scroll, I don't really notice it myself. Thanks

KatoakDR:
> From @rpherbig: Is there a case in which you'd want different encumbrance thresholds for different hunts? 
>
> From @vtcifer: Yeah, I feel like the simplest way of handling this settings wise is just a numeric threshold in the hunting_info: and set the default if not used to like 99, then you're always below the threshold.

Done. I've removed the global setting and instead the per-hunt setting is where you specify the threshold. By default, the setting is considered nil and so ignored (original behavior), so players have to opt-in by adding `stop_on_encumbrance: <number>` like they would for `stop_on` etc

Thanks

hunting-buddy.lic
```diff
@@ -253,11 +265,39 @@ class HuntingBuddy
...
   end
 
+  # Is your current encumbrance at or higher than the threshold?
+  # Threshold is a number from 0 (None) to 11 (It's amazing you aren't squashed!)
+  def encumbered?(threshold)
+    return false unless threshold
+    fput('encumbrance') # drinfomon script will parse the output then update DRStats
+    pause 1 # give drinfomon a chance to update DRStats so we use the value from now, not last time
+    encumbrance_index = [
+      "None",
```
rpherbig, line 275: What do you think about using this instead?

https://github.com/rpherbig/dr-scripts/blob/19d06c8de3ece9bd2185960706a39c07df822dbd/common.lic#L52

rpherbig:
Code changes look good (aside from the one comment) ✔️

hunting-buddy.lic
KatoakDR: Ha! I didn't know that was there, thanks

KatoakDR:
Doing more testing, I've gone down the rabbit hole of hunting-buddy and wrapping my head around how it iterates and passes arguments between multiple hunting files.

KatoakDR:
I've gone a different direction with my hunting-buddy updates, will close this in favor of forthcoming PRs, thanks

PR #5084, opened 2021-07-13
[script] [common-items] Use canonical model DRCI::Item to get name/regex

KatoakDR:
### Background
* Continuation of https://github.com/rpherbig/dr-scripts/pull/5041
* Some methods in `common-items` were on their own trying to determine the name or noun of an item, or create a regex pattern for it. The success rate depended significantly on how complete one's yaml configuration was when specifying the items.

### Changes
* Use the new `DRCI::Item.from_text(string)` method to convert a string argument into an `Item` class instance.
* Reuse the logic in `Item` class for `name`, `short_name`, and `short_regex` for consistency

PR #5094, opened 2021-07-14
[script] [common-items] Fix when item name includes 'my' prefix

KatoakDR:
### Background
* Reported by [DMAN](https://discord.com/channels/745675889622384681/745675890242879671/864884343151132714) in the Lich discord
* Prior to https://github.com/rpherbig/dr-scripts/pull/5084, the `remove_item_unsafe?(item)` used the entire contents of `item` argument. That pull request incorrectly split the string and used the first/last words only. So when the phrase "my cambrinth armband" was given then the updated command operated on "my armband" which now ambiguously refers to "leather armbands" and "cambrinth armbands", etc.

### Changes
* Update `Item.from_text(text)` method to preserve the `my` prefix as part of the item object's adjective.

### Tests
```
> ,e echo i = DRC::Item.from_text('armband') ; echo "adj=#{i.adjective}, noun=#{i.name}"
[exec1: adj=, noun=armband]

> ,e echo i = DRC::Item.from_text('my armband') ; echo "adj=#{i.adjective}, noun=#{i.name}"
[exec1: adj=my, noun=armband]

> ,e echo i = DRC::Item.from_text('my cambrinth armband') ; echo "adj=#{i.adjective}, noun=#{i.name}"
[exec1: adj=my cambrinth, noun=armband]

> ,e echo i = DRC::Item.from_text('my cambrinth something armband') ; echo "adj=#{i.adjective}, noun=#{i.name}"
[exec1: adj=my cambrinth, noun=armband]
```

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -369,13 +369,21 @@ module DRC
         .sub('.', ' ') # convert 'foo.bar' => 'foo bar' so more easily split into adjective and noun
         .squeeze(' ') # condense repeated runs of whitespace with a single space
         .strip # remove leading/trailing whitespace
+      has_my_prefix = text.start_with?('my ')
+      text = text.delete_prefix('my ')
       parts = text.split
       if text.nil? || text.empty?
         nil
       elsif parts.size > 1
-        DRC::Item.new(adjective: parts.first, name: parts.last)
+        DRC::Item.new(
+          adjective: "#{has_my_prefix ? 'my ' : ''}#{parts.first}",
+          name: parts.last
+        )
       else
-        DRC::Item.new(name: text)
+        DRC::Item.new(
+          adjective: has_my_prefix ? 'my' : nil,
+          name: text
+        )
       end
     end
   end


common.lic
```diff
@@ -369,13 +369,21 @@ module DRC
...
         .strip # remove leading/trailing whitespace
+      has_my_prefix = text.start_with?('my ')
+      text = text.delete_prefix('my ')
       parts = text.split
       if text.nil? || text.empty?
         nil
       elsif parts.size > 1
-        DRC::Item.new(adjective: parts.first, name: parts.last)
+        DRC::Item.new(
+          adjective: "#{has_my_prefix ? 'my ' : ''}#{parts.first}",
```
wstampley, line 379: Baking the 'my' into an item adjective feels like it will have repercussions elsewhere. The item should be the item, and the method using the item should probably decide to use 'my' or not, IMO.

rpherbig:
Yeah, I'm not sure I like this approach. There's already some special `my` handling in a few of the places that call this method (e.g. `have_item_by_look`) and it feels like we're layering on workarounds.

vtcifer:
I think break-fix wise, we should revert #5084, and discuss here future changes.

KatoakDR:
> I think break-fix wise, we should revert #5084, and discuss here future changes.

Agree, I've posted a request to revert it, too, at https://github.com/rpherbig/dr-scripts/pull/5084#issuecomment-880285830

KatoakDR:
> I agree with @wstampley - I'm not sure I like this approach. There's already some special `my` handling in a few of the places that call this method (e.g. `have_item_by_look`) and it feels like we're layering on workarounds.

Closing this PR, this was a quick attempt to try and fix the issues that https://github.com/rpherbig/dr-scripts/pull/5084 introduced but it's better we revert that PR and I'll go back to the drawing board. Thanks

common.lic
KatoakDR: Gotcha, thanks for the feedback

rpherbig:
#5084 has been reverted

Issue #4138, opened 2020-01-30
steal.lic - the item selection process can lead to very bad inefficiencies in travel


vtcifer:
At my current ranks, with a hometown of shard, and setting stealing towns as below, steal will start in crossing, go to shard, go back to crossing, go back to shard, go to leth on the first time it's run.  

Either after item selection in find_all_targets, the targets should be sorted based on order in stealing town, or find_all_targets should be run multiple times each for town.  I'm not sure which would be better.

**Character/settings:**
Guild: Thief
Thievery: 485
```yaml
stealing_towns:
 - Shard
 - Leth Deriel
 - Crossing
```

vtcifer:
I tried getting targets to sort based on town, but I have no idea why `targets.sort_by!{ |id | }` works, but `targets.sort_by!{ |town | }` doesn't.

Tarhish:
The `targets.select { |id|` thing is more like 'take the data that's in the first level of 'targets' and call it variable id.  So targets.sort_by! { |town| } is actually the same as targets.sort_by! { |id| } except it names the variable differently.

That said, given stealing options, I think that     `@stealing_options.sort_by { |data| data['town'] }` should work?

vtcifer:
After digging and staring a lot, it's because targets isn't what I thought it was, it's a triple-index array (that I'm still not sure how it's being used/why it's setup the way it is), but later on I was able to figure out that per_room_targets.values is the final list, and is an array as opposed to a triple indexed array of hashes.  Much easier to sort.  

Or even better use a for each and select to allow the yaml to pick the order.  Working on this now.

vtcifer:
Some quick testing and it's 

```ruby
@stealing_towns.each do |current_town|
  targets.select { |target| target['town'] == current_town }
  .each { |target| steal(target)}
end
```
instead of 

```ruby
targets.each { |target| steal(target) }
```

I'll run with this for a while but I think that's it. 😊

PR #5095, opened 2021-07-14
[script] [steal.lic] minor qol updates

vtcifer:
For anyone who might still be using it (might still be useful for thieves repairing rep), I've had these in my working directory for a while.  

Minor updates:
 - allow hometown override for bin (force_bin_town)
 - retry getting bin settings to mitigate intermittent failure
 - change shop selection criteria to stick within a town before moving to next
 - close stealing bag before redressing to prevent issues caused by stealing an item with same name as your equipment

### DIFF
--- a/steal.lic
+++ b/steal.lic
@@ -18,23 +18,19 @@ class Steal
     @stealing_bag = settings.stealing_bag
     stealing_hometown = settings.hometown 
     stealing_hometown = settings.force_bin_town if settings.force_bin_town
-    hometown_data = get_data('town')[stealing_hometown]
-    @bin_id = hometown_data['thief_bin'] && hometown_data['thief_bin']['id']
     @stealing_towns = settings.stealing_towns
     @bin_stolen = settings.bin_stolen
-    if @bin_stolen && !@bin_id
-      DRC.message("*** Binning check failed first attempt for #{stealing_hometown}. Trying again:")
-      pause 2
-      hometown_data = get_data('town')[stealing_hometown]
-      @bin_id = hometown_data['thief_bin'] && hometown_data['thief_bin']['id']
-      if !@bin_id
+    if DRStats.thief?
+      @bin_id = get_hometown_target_id(@hometown,'thief_bin')
+      if !(@bin_id)
+        message("Binning not supported in #{@hometown}.  You should turn it off or change hometown.")
         @bin_stolen = false
-        DRC.message("*** Binning check failed second attempt.  Binning is not yet supported in #{settings.hometown}")
-        pause 10 
-      else
-        DRC.message("*** Binning check succeeded second attempt.  Binning supported in #{settings.hometown}")
-        pause 10 
+        pause 5
       end
+    else
+      message("You are not a thief.  You can't use thief bins.  You should turn it off.")
+      @bin_stolen = false
+      pause 5
     end
     @hide_to_steal = settings.hide_to_steal
     @stealing_low_acceptable_count = settings.stealing_low_acceptable_count
@@ -97,12 +93,12 @@ class Steal
 
     # close stealing container before getting re-dressed 
     # in case you stole something with same name as your equipment
-    fput ("close my #{@stealing_bag}") if @bin_stolen
+    fput("close my #{@stealing_bag}") if @bin_stolen
 
     @equipment_manager.wear_equipment_set?('standard')
 
     # reopen stealing container after getting re-dressed 
-    fput ("open my #{@stealing_bag}") if @bin_stolen
+    fput("open my #{@stealing_bag}") if @bin_stolen
 
     # As is done when `burgle` script finishes, release invisibility
     # because it can interfere with other actions like using banks.
--- a/steal.lic
+++ b/steal.lic
@@ -16,13 +16,25 @@ class Steal
 
     @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
     @stealing_bag = settings.stealing_bag
-    hometown_data = get_data('town')[settings.hometown]
+    stealing_hometown = settings.hometown 
+    stealing_hometown = settings.force_bin_town if settings.force_bin_town
+    hometown_data = get_data('town')[stealing_hometown]
     @bin_id = hometown_data['thief_bin'] && hometown_data['thief_bin']['id']
     @stealing_towns = settings.stealing_towns
     @bin_stolen = settings.bin_stolen
     if @bin_stolen && !@bin_id
-      DRC.message("*** Binning is not yet supported in #{settings.hometown}")
-      @bin_stolen = false
+      DRC.message("*** Binning check failed first attempt for #{stealing_hometown}. Trying again:")
+      pause 2
+      hometown_data = get_data('town')[stealing_hometown]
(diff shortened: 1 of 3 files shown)


steal.lic
```diff
@@ -16,13 +16,25 @@ class Steal
 
     @stealing_options = get_data('stealing').stealing_options.reject { |target| settings.dont_steal_list.include?(target['id'].to_i) }
     @stealing_bag = settings.stealing_bag
-    hometown_data = get_data('town')[settings.hometown]
+    stealing_hometown = settings.hometown 
+    stealing_hometown = settings.force_bin_town if settings.force_bin_town
```
rpherbig, line 20: `force_bin_town` needs to be in `base.yaml`

steal.lic
```diff
@@ -16,13 +16,25 @@ class Steal
...
+    hometown_data = get_data('town')[stealing_hometown]
     @bin_id = hometown_data['thief_bin'] && hometown_data['thief_bin']['id']
     @stealing_towns = settings.stealing_towns
     @bin_stolen = settings.bin_stolen
     if @bin_stolen && !@bin_id
-      DRC.message("*** Binning is not yet supported in #{settings.hometown}")
-      @bin_stolen = false
+      DRC.message("*** Binning check failed first attempt for #{stealing_hometown}. Trying again:")
+      pause 2
+      hometown_data = get_data('town')[stealing_hometown]
```
rpherbig, line 28: Does this need to be reassigned?

vtcifer: Yea, the _usual_ cause of the failure that requires the 2nd attempt is a failure at loading town data.  I've never been able to nail down what specifically causes the problem with the load, and why it sonly seems to impact steal / burgle but a 2nd attempt usually fixes it.

vtcifer:
Also: closes #4138.

steal.lic
```diff
@@ -16,13 +16,25 @@ class Steal
...
+      @bin_id = hometown_data['thief_bin'] && hometown_data['thief_bin']['id']
+      if !@bin_id
+        @bin_stolen = false
+        DRC.message("*** Binning check failed second attempt.  Binning is not yet supported in #{settings.hometown}")
+        pause 10
```
KatoakDR, line 33: ❓ What are the `pause 10` for on lines 33 and 36?

steal.lic
```diff
@@ -74,9 +86,24 @@ class Steal
...
+    # steal within one area before moving to next
+    @stealing_towns.each do |current_town|
+      targets.select { |target| target['town'] == current_town}
+      .each { |target| steal(target) }
+    end
+
+
+    # close stealing container before getting re-dressed 
+    # in case you stole something with same name as your equipment
+    fput ("close my #{@stealing_bag}") if @bin_stolen
```
KatoakDR, line 100: 💡 Possible candidate for `DRCI.close_container?(@stealing_bag)`?  Otherwise, my next ~suggestion~ preference is remove the space between `fput` and the open paren 😁

steal.lic
```diff
@@ -74,9 +86,24 @@ class Steal
...
+
     @equipment_manager.wear_equipment_set?('standard')
 
+    # reopen stealing container after getting re-dressed 
+    fput ("open my #{@stealing_bag}") if @bin_stolen
```
KatoakDR, line 105: 💡 Possible candidate for `DRCI.open_container?(@stealing_bag)`?  Otherwise, my next ~suggestion~ preference is remove the space between `fput` and the open paren 😁

steal.lic
```diff
@@ -74,9 +86,24 @@ class Steal
...
+    # make semi-intelligent choices about which order to steal things in:
+    # steal within one area before moving to next
+    @stealing_towns.each do |current_town|
+      targets.select { |target| target['town'] == current_town}
+      .each { |target| steal(target) }
+    end
+
+
+    # close stealing container before getting re-dressed 
+    # in case you stole something with same name as your equipment
```
KatoakDR, line 99: Good catch!

steal.lic
```diff
@@ -74,9 +86,24 @@ class Steal
...
     start_script('performance', ['noclean']) unless settings.hide_to_steal
     npcs.each { |npc| steal_from_npc(npc, settings.npc_stealing_attempt_count) }
-    targets.each { |target| steal(target) }
+    
+    # make semi-intelligent choices about which order to steal things in:
+    # steal within one area before moving to next
+    @stealing_towns.each do |current_town|
+      targets.select { |target| target['town'] == current_town}
+      .each { |target| steal(target) }
+    end
```
KatoakDR, line 95: Nice optimization

steal.lic
rpherbig: This is unintuitive, but I get that weird things happen. Could you add a comment explaining it so someone doesn't try to refactor/optimize this out in the future?

steal.lic
vtcifer: Not needed imo.  That's useful when you want to robust handling of close / failure to close (/ open / failure to open) but here, there's not a good code path if there is a problem AFAIK.

steal.lic
vtcifer: 1\ Allows timing for whatever issue prevented the data load from occurring to resolve itself.
2\ in case of actual non-bin support for an area, moving on too quickly means the user doesn't see the bad setting.  

Once DRC.message updates are in the second one can possibly be removed, as we can direct that to a window other than main so the user doesn't miss it.

I probably should actually convert this to the DRCT method written at the same time and used in burgle.lic that encapsulates all of this.

KatoakDR: Gotcha, if the pauses stay then would you mind adding a comment explaining this. Similar to https://github.com/rpherbig/dr-scripts/pull/5095#discussion_r670432665

thanks!

steal.lic
vtcifer: The switch to using the DCT.get_hometown_target_id method should take care of this as it's documented in there.

steal.lic
vtcifer: The switch to using the DCT.get_hometown_target_id method should take care of this as it's documented in there.

rpherbig:
These changes look good to me. What's left before this is ready to be merged?

vtcifer:
> These changes look good to me. What's left before this is ready to be merged?

Not quite ready.  The most recent changes I made just in code and have actually tested yet.  I want to get a couple of runs in at least to validate first.  Should have time this week to.

vtcifer:
✅ Should be good to merge now.

steal.lic
```diff
@@ -16,13 +16,21 @@ class Steal
...
-      DRC.message("*** Binning is not yet supported in #{settings.hometown}")
+    if DRStats.thief?
+      @bin_id = DRCT.get_hometown_target_id(stealing_hometown,'thief_bin')
+      if !(@bin_id)
+        DRC.message("Binning not supported in #{stealing_hometown}.  You should turn it off or change hometown/force_bin_town.")
+        @bin_stolen = false
+        pause 5
+      end
+    else
+      DRC.message("You are not a thief.  You can't use thief bins.  You should turn it off.")
```
KatoakDR, line 31: Looks like in this update that the check for if `@bin_stolen` is set was removed from original line 23. I think this message will appear every time for non-thieves. It doesn't hurt, but may give false positives.

PR #5100, opened 2021-07-19
Arenawatch + Combat-Training Cleanup

wstampley:
Had several runs where combat would get into an RT before arenawatch could do its thing. I'm sure this needs to be more configurable, but the chat in discord prompted me to throw it here for feedback at least.

### DIFF
--- a/arenawatch.lic
+++ b/arenawatch.lic
@@ -1,20 +1,20 @@
-custom_require.call(%w[common common-items common-summoning common-travel drinfomon equipmanager events spellmonitor])
-
-$arena_watch_timer = nil
+custom_require.call(%w[common common-summoning common-travel drinfomon equipmanager events spellmonitor])
 
 def mainloop
   ignore_npcs = %w[Usdiwi Servant, Traumatologist Usdiwi]
   npcs = DRRoom.npcs - ignore_npcs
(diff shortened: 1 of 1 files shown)


arenawatch.lic
```diff
@@ -1,4 +1,6 @@
-custom_require.call(%w[common common-arcana common-healing common-items common-summoning common-travel drinfomon equipmanager events spellmonitor])
+custom_require.call(%w[common common-items common-summoning common-travel drinfomon equipmanager events spellmonitor])
+
+$arena_watch_timer = nil
```
KatoakDR, line 3: Why is the variable needed as a global rather than local to the mainLoop method?

wstampley: Cause in my head I said, "You can't put it in a loop, it will get recreated each time." and the method name is "main loop", therefore it is a loop. (it could totally live in the mainLoop method before the loop starts)

KatoakDR (approved):
Looks good to me, and the tie in with `combat-trainer` gives non-Thieves an alternate to `stabbity`

rpherbig:
Code looks fine to me ✔️

wstampley:
I think this is good to merge.

PR #5105, opened 2021-07-24
[script] [tendme] Skin wounds aren't tendable

KatoakDR:
### Background
* Sorcerous backlash and other causes may give you a skin bleeder.
* Although skin bleeders are considered external, they are not tendable.

```
[tendme]>tend my skin

TEND allows for the tending of wounds on yourself or others.
Syntax:
TEND {MY|<character>} {area}
{area} may be one of:
       head          chest
       abdomen       back
       neck          tail
       right arm     left arm
       right hand    left hand
       right leg     left leg
       right eye     left eye
Examples:
TEND MY HEAD
TEND GREGORY RIGHT ARM
The demeanor of your target may affect your ability to tend them.
```

### Changes
* Exclude skin body part from tendable bleeders so that `tendme` doesn't get hung up trying to tend an untendable and then you bleed out and die.

### DIFF
--- a/tendme.lic
+++ b/tendme.lic
@@ -26,7 +26,7 @@ class TendMe
       health_data['parasites'].values.flatten +
       health_data['bleeders'].values.flatten
     )
-    .reject { |wound| wound.internal? || wound.bleeding_rate =~ /tended|clotted/ || wound.body_part == 'skin' } # untendable
+    .reject { |wound| !wound.tendable? || wound.bleeding_rate =~ /tended|clotted/ } # untendable
     .select do |wound|
       if wound.bleeding? && !DRCH.skilled_to_tend_wound?(wound.bleeding_rate)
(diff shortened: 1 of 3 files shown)


tendme.lic
```diff
@@ -26,7 +26,7 @@ class TendMe
       health_data['parasites'].values.flatten +
       health_data['bleeders'].values.flatten
     )
-    .reject { |wound| wound.internal? || wound.bleeding_rate =~ /tended|clotted/ } # untendable
+    .reject { |wound| wound.internal? || wound.bleeding_rate =~ /tended|clotted/ || wound.body_part == 'skin' } # untendable
```
wstampley, line 29: Should 'tendable?' be a wound class property so this doesn't have to go everywhere?

KatoakDR: Yeah, I debated that too. I was just being lazy. I'll add a tendable property on the wound class.

rpherbig:

KatoakDR:
This is ready, thanks

PR #5112, opened 2021-08-05
[script] [ combat-trainer] Added Debilitation and Targeted Magic to cyclic training list.

Dartellum:
Nothing Earthshattering here with this change, just the ability to swap between two additional cyclic spells during combat-trainer. Note, it will not stop a running one if the other cyclic skills do not need training regardless of timer. For example, I have EE and Rimefang defined in `combat_spell_training:`. If Debilitation needs training, combat-trainer will start EE and then run the 5 minutes and 25 seconds before, if TM training is needed, it will switch to Rimefang. If TM did not need training, EE would keep running. I see nothing wrong with it running as, if the hunt is long enough, Debil will drain to below the threshold for it to require training again, and then the switch will occur. 

The three messages were added to make sure the skill was reading and when the timer was set to change if needed.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1488,7 +1488,7 @@ class SpellProcess
     return if @training_spells_max_threshold && game_state.npcs.length > @training_spells_max_threshold
     return if DRStats.mana < @training_spell_mana_threshold
 
-    needs_training = %w[Warding Utility Augmentation Sorcery]
+    needs_training = %w[Warding Utility Augmentation Sorcery Debilitation Targeted\ Magic]
                      .select { |skill| @training_spells[skill] }
                      .select { |skill| game_state.is_offense_allowed? || @training_spells[skill]['harmless'] }
                      .select { |skill| DRSkill.getxp(skill) < @magic_exp_training_max_threshold }
@@ -1497,13 +1497,16 @@ class SpellProcess
                      .select { |skill| @training_spells[skill]['day'] ? UserVars.sun['day'] : true }
     needs_training = game_state.sort_by_rate_then_rank(needs_training).first
     return unless needs_training
+    DRC.message "needs_training: #{needs_training}"
 
     data = @training_spells[needs_training]
     @training_cast_timer = Time.now + @training_spells_wait
     if data['cyclic']
       @training_cyclic_timer = Time.now + 325
+      DRC.message "@training_cyclic_timer of Time.now + 325: #{@training_cyclic_timer}"
     elsif @training_cyclic_timer < @training_cast_timer
       @training_cyclic_timer = @training_cast_timer
+      DRC.message "@training_cyclic_timer less than @training_cast_timer: #{@training_cyclic_timer}"
     end
 
     game_state.hide_on_cast = true if data['use_stealth']
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1488,7 +1488,6 @@ class SpellProcess
     return if @training_spells_max_threshold && game_state.npcs.length > @training_spells_max_threshold
     return if DRStats.mana < @training_spell_mana_threshold
 
-    #needs_training = %w[Warding Utility Augmentation Sorcery Debilitation Targeted\ Magic]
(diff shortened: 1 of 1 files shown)


KatoakDR:
Timely, I just got Fire Rain and was wondering how I could switch between Electrostatic Eddy (debil) and Fire Rain (tm) cyclic spells

combat-trainer.lic
```diff
@@ -1497,13 +1497,16 @@ class SpellProcess
                      .select { |skill| @training_spells[skill]['day'] ? UserVars.sun['day'] : true }
     needs_training = game_state.sort_by_rate_then_rank(needs_training).first
     return unless needs_training
+    DRC.message "needs_training: #{needs_training}" if $debug_mode_ct
```
KatoakDR, line 1500: Do these need to be `DRC.message`? Most debug statements in this file use `echo`

combat-trainer.lic
```diff
@@ -1488,7 +1488,7 @@ class SpellProcess
     return if @training_spells_max_threshold && game_state.npcs.length > @training_spells_max_threshold
     return if DRStats.mana < @training_spell_mana_threshold
 
-    needs_training = %w[Warding Utility Augmentation Sorcery]
+    needs_training = %w[Warding Utility Augmentation Sorcery Debilitation Targeted\ Magic]
```
KatoakDR, line 1491: I suggest we refactor this list to standard looking array of strings to avoid ambiguity about the escaped space `\ `
```suggestion
    needs_training = ['Warding', 'Utility', 'Augmentation', 'Sorcery', 'Debilitation', 'Targeted Magic']
```

Dartellum:
> Timely, I just got Fire Rain and was wondering how I could switch between Electrostatic Eddy (debil) and Fire Rain (tm) cyclic spells

You could do the way I did before deciding to add to this. Hint, that was the long `offensive_spell_cycle` list I used to use.

combat-trainer.lic
Dartellum: I was working within the bounds of %w and could not get it to work without the escaped space. Wonder why it was %w now...

combat-trainer.lic
Dartellum: Switching to DRC.message where I can. I like you can see the output in the console as it is visible (red).

combat-trainer.lic
KatoakDR: Someone trying to be cheeky with Ruby syntax, heh

Dartellum: Either way, when using %w, you need to escape `\` spaces. That took a bit to find. I will test the other way.

rpherbig: `%w[]` is the convention when it's easy to do (Rubocop will auto-fix this). Since we need the escaped space, I'm fine with going to `['...']` syntax

rpherbig:
Changes look fine to me ✔️ 

Let me know when it's tested and ready to be merged

combat-trainer.lic
Dartellum: Changed to `['..']`. Works as expected.

Dartellum:
Working as expected. I may revisit if we want the cyclic to actually end when `@training_cyclic_timer` runs out or leave it run as it currently does. Does not hurt to leave it run and it will end it to cast a new one when that skill needs trained. :man_shrugging:

KatoakDR:
> Working as expected. I may revisit if we want the cyclic to actually end when `@training_cyclic_timer` runs out or leave it run as it currently does. Does not hurt to leave it run and it will end it to cast a new one when that skill needs trained. 🤷‍♂️

I think it's fine to keep the cyclic running and it only be cancelled when a new cyclic is attempted to be cast. But, to give people options and control, I can see the merit of a toggle for how that should behave (let the last cyclic just stay on or proactively release it after the timer). Though I don't know why someone would want to cancel their cyclic, unless they suffer from bad attunement and run out of mana?

wstampley:
> Though I don't know why someone would want to cancel their cyclic, unless they suffer from bad attunement and run out of mana?

For me it would be because damaging cyclics that are generating XP I don't need are effectively taking XP from weapons / defenses / etc. depending on spawn, but I also think it's fine as is.

Issue #4340, opened 2020-06-10
Two artificing base-recipes errors


KillashandreaDR:
artificing: -simple gwethdesuan...
it indicates "bone totem" which works up until finishing the enchantment, after which it becomes a "totem" (no adjective).  The workorder script them tries to pick up a "bone totem" and bundle it with the logbook, but instead it pulls a "bone totem" from your backpack.  In order to get it to work, you need to either drop the adjective from the enchant and workorder scripts only when picking up the item from your feet and when bundling with the logbook...or just remove "bone" from the base recipe.

artificing: -aura device...
it orders sigil #4 when it needs to order sigil #2.

Dartellum:
I will work through this as I update all the recipes. I am getting closer to done. I will also work through each of these recipes to make sure the noun setting is correct. I see many nouns as bone totem but most are `workorders: false` so no one should encounter them doing workorders. :smile:

Dartellum:
> artificing: -simple gwethdesuan...
> it indicates "bone totem" which works up until finishing the enchantment, after which it becomes a "totem" (no adjective). The workorder script them tries to pick up a "bone totem" and bundle it with the logbook, but instead it pulls a "bone totem" from your backpack. In order to get it to work, you need to either drop the adjective from the enchant and workorder scripts only when picking up the item from your feet and when bundling with the logbook...or just remove "bone" from the base recipe.
> 
> artificing: -aura device...
> it orders sigil #4 when it needs to order sigil #2.

@KillashandreaDR 
Did you mean `simple gwethsmasher` in chapter 2 or `a gwethdesuan` in chapter 5?

KillashandreaDR:
Sorry been so long I don’t remember.  But since I wrote “simple” and based off of epedia and the components list, I am 99% sure I meant the smasher.

Dartellum:
Okay fixed that one but it is set to false for workorders because one of the sigils, metamorphosis, is not store-bought. However, it will work for overrides now once the `base-recipes` PR is merged.

Dartellum:
Fix for picking up the correct item. During testing, I had both ways, get my #{@item} and get #{@item}, retrieve the wrong item occasionally. This fix, https://github.com/rpherbig/dr-scripts/pull/5114, makes sure the item is picked up from the at feet location. Or, it did during my testing. I would like you to test it and report back. :smile:

PR #5114, opened 2021-08-08
[script] [enchant] Updated cleanup section to make sure the item on the ground it picked up and not retrieved from container.

Dartellum:
Sometimes `get my #{@item}` would retrieve the item from the container instead of picking the item up from the feet location. This change tries both get ways to retrieve the item. This satisfies the second part of Issue https://github.com/rpherbig/dr-scripts/issues/4340. Putting in draft until some more can test, specifically KillashandreaDR. Yes, it works for me.


Dartellum:
This PR works as expected.

PR #5117, opened 2021-08-11
Revert "Temporary Map Fix"

Hiinky:
Reverts rpherbig/dr-scripts#5116

Tillmen responded!  Map DB has been fixed!


Hiinky:
Thanks for the super quick response last night.  saved SO much complaining and brokenness.

PR #5122, opened 2021-08-12
Offensive Spell Debugging

wstampley:
Adding some logging to offensive spell function in CT, since it's a somewhat regular debugging topic in lnet/discord.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1696,10 +1696,12 @@ class SpellProcess
   end
 
   def check_offensive(game_state)
+    echo "Checking offensive spells:" if $debug_mode_ct
     return if game_state.casting
     return if game_state.npcs.empty?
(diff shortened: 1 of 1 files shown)


Dartellum:
> > Should use DRC.message over echo. That way at the cli you can see the red line pop up.
> 
> Ooo..gives me another idea for the DRC.message enhancements down the line, debug window, so you can dump all that stuff to that window and not _need_ to clutter main window. Although you lose some context surrounding it....

Have it output to both places.

wstampley:
> Should use DRC.message over echo. That way at the cli you can see the red line pop up.

I would be okay with that, in general. Probably best to keep it consistent within a script.

I think we probably ought to have a version of DRC.debug_message that could be used consistently, and then could be configured based on user preference / using VTC's customizations, etc.

PR #4764, opened 2021-02-23
Restock improvements for ritual foci

ubiquitous42:
I wasn't satisfied with the fact that restock can only replace ritual foci when they completely run out, due to the order in which I run things.  I made it so that it would work for Moon Mages, but I have NOT tested it for the other limited-use ritual foci.

(Also yes, I know I screwed up this PR by accidentally bundling my outdoorsmanship changes into it.  I'm not sure how I can best separate that change without starting a new repo, since I accidentally committed it to my own master.)

PR #5126, opened 2021-08-13
[script] [restock] better consumable ritual focus support

vtcifer:
- supports better checks for ritual support items


wstampley:
Looks like this will supersede https://github.com/rpherbig/dr-scripts/pull/4764 and allow us to close that lingering PR.

rpherbig:
> Looks like this will supersede #4764 and allow us to close that lingering PR.

Good catch!

PR #5001, opened 2021-06-08
Rewrite of picking and the base-picking data it needs

wstampley:
What started as a refactor basically turned into rewriting pick.lic.

Major Changes:
  Resolves Issue: #1669 (Ignores difficulty on Reaper boxes)
  Resolves Issue: #4347 (Roundtime match race condition)
  Resolves Issue: #4339 (Improve 'early out' options for pick - xp, no boxes)
  Resolves Issue: #4226 (Use correct speed for each difficulty level)
  Resolves Issue: #3135 (Support Thief Glance)
  Feature: Relies on common-items methods to reduce code duplication
  Feature: Supports multiple box containers natively (Usually Autoloot + Stow)
  Feature: Optimizes startup/shutdown to minimize gear removal with nothing to do
  Feature: A run of bad 'something shifts' appropriately slows down / gives up on a box
  Feature: Configurable 'Speed' Levels by difficulty numbers for disarm and pick
  Feature: Supports worn trashcans
  Feature: Supports Trap Blacklist
  Feature: Supports Tendme before running Safe-room for people who want it
  Feature: Supports 'all' argument for overwriting stop_pick_on_mindlock setting from CLI
  Feature: Supports 'assume' argument (uses the 'speed' values) to skip identification and assume the difficulty
  
Depends on https://github.com/rpherbig/dr-scripts/pull/4999

New Configuration Example:
```
worn_trashcan:  # can be set to the adj. + noun for a wearable trash container
worn_trashcan_verb: # can be set to the verb that empties your trash container

pick:
  debug: true # turns on debug statements so that if something goes wrong, your log will be more useful
  use_glance: false # turns off glance if you're a thief, not required otherwise
  trash_empty_boxes: false # if true, tries to trash instead of dismantling boxes after looting
  # optional property to skip the identify step and just assume that a box is 'blind/quick/careful/etc.'
  # Warning: Since this skips identify, it also skips the blacklist and 'too hard' checks for traps
  assumed_difficulty: quick/careful/etc.
  # optionally set to a number 0 -> 16 to control speed for when locks are that hard
  # anything less than the quick threshold will be done blind
  pick_quick_threshold: 
  pick_normal_threshold:
  pick_careful_threshold:
  # optionally set to a number 0 -> 16 to control speed for when traps are that hard
  # anything less than the quick threshold will be done blind
  disarm_quick_threshold:
  disarm_normal_threshold:
  disarm_careful_threshold:
  # set a threshold (0-16) for when a trap is too hard to mess with for you
[...]

PR #5135, opened 2021-08-16
pick.lic - Added args to ignore exp mindstate limits and skip perception checks

mikejcook:
Two new arguments, "force" and "perception"

Force: ignore experience mindstate limits that would normally stop picking
Perception: Skip identify checks (for when perception skill is insufficient)

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -107,7 +107,7 @@ class LockPicker
   end
 
   def stop_picking?
-    @stop_pick_on_mindlock && DRSkill.getxp('Locksmithing') >= 30
+    @stop_pick_on_mindlock && !@force_repair && DRSkill.getxp('Locksmithing') >= 30
   end
 
   def setup
@@ -116,12 +116,16 @@ class LockPicker
         { name: 'pets', regex: /pets/, optional: true, description: 'Disarm boxes and place them in the pet box container.' },
         { name: 'count', regex: /\d+/, optional: true, description: 'How many pet boxes to make.' },
         { name: 'refill', regex: /refill/, optional: true, description: 'Refills your lockpick ring.' },
-        { name: 'source', regex: /(source=(\w+))/, optional: true, description: 'Container with boxes to pick.'}
+        { name: 'source', regex: /(source=(\w+))/, optional: true, description: 'Container with boxes to pick.'},
+        { name: 'force', regex: /force/i, optional: true, description: 'Force picking, ignoring any experience mindstate limits' },
+        { name: 'perception', regex: /perception/i, optional: true, description: 'Skip perception checks' }
       ]
     ]
 
     args = parse_args(arg_definitions, true)
-
+	
+    @skip_identify = args.perception
+    @force_repair = args.force
     @make_pets = args.pets
     @pet_goal = args.count.to_i
     @pet_count = 0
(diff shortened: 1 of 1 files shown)


pick.lic
```diff
@@ -116,12 +116,16 @@ class LockPicker
...
+        { name: 'force', regex: /force/i, optional: true, description: 'Force picking, ignoring any experience mindstate limits' },
+        { name: 'perception', regex: /perception/i, optional: true, description: 'Skip perception checks' }
       ]
     ]
 
     args = parse_args(arg_definitions, true)
-
+	
+    @skip_identify = args.perception
+    @force_repair = args.force
```
rpherbig, line 128: `_repair` doesn't feel like the right name. What does this have to do with repairing?

mikejcook: Corrected with new commit

pick.lic
```diff
@@ -68,7 +68,7 @@ class LockPicker
     @harvest_traps = false
 
     boxes.each do |box|
-      if DRSkill.getxp('Locksmithing') >= 30
+      if DRSkill.getxp('Locksmithing') >= 30 && !force_picking
```
rpherbig, line 71: Should this be `@force_picking`?

pick.lic
```diff
@@ -116,12 +116,16 @@ class LockPicker
         { name: 'pets', regex: /pets/, optional: true, description: 'Disarm boxes and place them in the pet box container.' },
         { name: 'count', regex: /\d+/, optional: true, description: 'How many pet boxes to make.' },
         { name: 'refill', regex: /refill/, optional: true, description: 'Refills your lockpick ring.' },
-        { name: 'source', regex: /(source=(\w+))/, optional: true, description: 'Container with boxes to pick.'}
+        { name: 'source', regex: /(source=(\w+))/, optional: true, description: 'Container with boxes to pick.'},
+        { name: 'force', regex: /force/i, optional: true, description: 'Force picking, ignoring any experience mindstate limits' },
```
rpherbig, line 120: How does this arg differ from setting `stop_pick_on_mindlock` to false?

mikejcook: This can be done on the fly, rather than having to change your YAML. I generally prefer to stop on mindlock, but if I'm e.g. opening boxes for someone else, that isn't the case.

Dartellum: It is not different. If I set `stop_pick_on_mindlock: false` I keep picking when I hit 34/34.

Dartellum:
**FYI** A very large rewrite of pick is underway, see PR https://github.com/rpherbig/dr-scripts/pull/5001.

mikejcook:
> 
> 
> **FYI** A very large rewrite of pick is underway, see PR #5001.

I've gone ahead and closed the pull request, since there's a re-write already, but you might consider adding one or both args to the new re-write. I use them in my own local version of pick. Yes, you can have a different YAML, but it's nice being able to just throw "force" to keep picking past locked. And then, some locks I am not able to identify but can disarm/pick, and the script just hangs forever on trying, hence why I added that argument.

rpherbig:
I can see the use case for wanting a script arg in addition to a YAML setting. The only thing I'd suggest is using the script arg to modify the assignment to `stop_on_mindlock`, instead of adding a new conditional

PR #5137, opened 2021-08-17
[data] [crafting] remove crossing anvil room outside society

vtcifer:
Removed 760 from crossing anvils due to it being outside not in no-stealing room, and thus prone to getting stolen from


vtcifer:
Inadvertently merged in other pending changes - closing will resubmit.

PR #5139, opened 2021-08-18
[script][base-lumber] Added more spots to lumberjack. 

Dartellum:
Also broke the Adan'f woods up to avoid the heavy hitters. Credit to Damiza for doing this, I am just opening the PR.


rpherbig:
@Dartellum could you or Damiza advertise this change in Discord/lnet so people are aware they may need to change their lumber config in case they wanted some of the old behavior?

Kaesken:
Okies, I let them know!  <3

PR #5141, opened 2021-08-19
[script][base-mining]Damiza went around and updated the mining information. Yay!

Dartellum:


data/base-mining.yaml
```diff
@@ -104,11 +172,27 @@ mining_buddy_rooms:
...
   - 7020
   - 7021
+  - 10008
+  - 10009
+  - 10011
+  - 7023
+  - 7024
+# Divided this section because you must kneel
+# You cannot retreat in this location.
+  breech_tunnels2:
```
wstampley, line 182: This is likely to expose some issues in mining-buddy / mine, where it will try to enter, and have to kneel, and then be unable to mine. Probably fine to add them here, but might be good to figure out exactly what happens in those scripts when/if someone tries to use this.

wstampley (approved):

data/base-mining.yaml
Dartellum: These were already in, just not broken out like this. You will need a trigger (another script) to kill the spiders since you cannot retreat. I would rather include all the mining areas and let the player ask or figure out a solution. :smile:

Issue #5143, opened 2021-08-24
combat-trainer.lic - issue with empath self-healing


mikejcook:
There's an issue where combat-trainer won't cast Empath healing spells when facing non-constructs/undead with the error message "Dropping spell: Offensive magic is not allowed right now..."

I tracked it down to the following issue:

In the method check_health_empath, the empath spells are created on the fly (objects are named "data"), but the harmless flag isn't set to true.

So, when we then hit our cast_spell method, we end up with this check returning false on both sides:
if game_state.is_offense_allowed? || @prepared_spell['harmless']

So, the spell falls through to the error message and it's dropped.

I was able to hack around this with the following change:
Original: data = { 'abbrev' => 'vh', 'mana' => @empath_spells['VH'].first, 'cambrinth' => @empath_spells['VH'][1..-1]}
New: data = { 'abbrev' => 'vh', 'mana' => @empath_spells['VH'].first, 'cambrinth' => @empath_spells['VH'][1..-1], 'harmless' => true }

I'm not sure if this is necessarily the "best" way to fix it, though, so I created the issue rather than a commit.

rpherbig:
@Dartellum @vtcifer @wstampley @LuckyThiefDR @Hiinky @KatoakDR could one of you take a look at this?

vtcifer:
I have no empath and am not really familiar with the specifics of their hunting, but a quick manual look at the code suggests that's probably the best immediate fix.  Empath healing only considers 3 spells, vh, foc, heal, all of which are harmless.  Anything else added to that shouldn't be used as far as I can tell.

A longer term fix might be merging with data from base-spells?

wstampley:
I'll put in the quick fix for now. Merging with base-spells, and rewriting it to use the same healing waggle stuff as healme is probably all stuff on the future todo list.

wstampley:
https://github.com/rpherbig/dr-scripts/pull/5144

rpherbig:
Thanks for the report @mikejcook !

PR #5144, opened 2021-08-24
Quick Fix: Empath Healing + Shock Guards

wstampley:
Marking the spell data for "on the fly" generated heal spells as 'harmless' so that the shock guards don't drop the spells


wstampley:
https://github.com/rpherbig/dr-scripts/issues/5143

PR #5151, opened 2021-09-02
titlecheck.lic - added arguments to optionally list ALL titles and/or SORT them

mikejcook:
I found that parsing the standard title output was challenging and wanted to see titles in a nice list. titlecheck provided that functionality, but only for titles new since the last run. I added some simple arguments that bypass checking the existing list and optionally sort the titles alphabetically.

### DIFF
--- a/titlecheck.lic
+++ b/titlecheck.lic
@@ -8,25 +8,36 @@ class TitleCheck
   def initialize
     arg_definitions = [
       [
-        { name: 'title_set', options: %w[blunt ranged brawling genedged specedged thrown pole shield slings weapons weaponmaster performer primarymagic magic money ownership survival1 survival2 survival3 lore criminal generic racial premium order religion novice practitioner dilettante aficionado adept expert professional authority genius savant master grandmaster guru legend cleric trader warmage necromancer empath ranger paladin bard barbarian thief moonmage], description: 'What set of titles are you checking?' }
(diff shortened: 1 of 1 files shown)


Dartellum:
I will grab these changes and give them a go.

Dartellum:
Okay, works. However, alphabetical sorting removes the progression of titles. Yes, the blob output is hard to read and I like the list. Perhaps make the list display as in the original order and offer the sort.

mikejcook:
I'm not sure what you mean. Sorting only happens when you use the sort argument. Without that argument, it'll display in the original, progression order.

Without any additional arguments, it's functionally identical to the original titlecheck

With all, it'll list all titles, regardless of what you've viewed previously

With sort, it'll list the titles sorted alphabetically, whether viewing only new titles or all

Dartellum:
> I'm not sure what you mean. Sorting only happens when you use the sort argument. Without that argument, it'll display in the original, progression order.
> 
> Without any additional arguments, it's functionally identical to the original titlecheck
> 
> With all, it'll list all titles, regardless of what you've viewed previously
> 
> With sort, it'll list the titles sorted alphabetically, whether viewing only new titles or all

What I am saying is I like the list format you did when you pass sort in but I would like it to list it that way without sorting as well (optional, of course). The blob from the original is overwhelming sometimes. :smile:

mikejcook:
Well, the list format was part of titlecheck originally. I just added the all and sort arguments.

So if you do ;titlecheck <cat> or ;titlecheck<cat> all, you should see what you are asking for? It'll put it in a list format, but without the sort. Just don't pass in the sort argument.

Dartellum (approved):

PR #5152, opened 2021-09-02
Add Sorcery Option to Crafting/Outdoors Magic

Hiinky:
Defaults to false for safety.  Allows people control to add more magic training at their own risk.  @Dartellum @wstampley @vtcifer @BinuDR   Do you guys have any particular feels about this add?

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -731,6 +731,9 @@ use_own_ingot_type:
 deed_own_ingot: false
 # spells that get woven in to various scripts, do not use cambrinth
 crafting_training_spells:
+# Toggle Sorcery during crafting or outdoorsmanship. Sorcery will still be blocked by forge.lic or false.
+#ADVANCED OPTION MAY RESULT IN ARREST, USE WITH CAUTION!!
+cast_sorcery_training: false
 # mindstate craft.lic will skip making an item at
 craft_max_mindstate: 31
 # amount of yarn to keep on hand for craft.lic, don't exceed 300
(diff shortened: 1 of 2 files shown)


BinuDR:
I would add a message in validate if it finds sorcery included that warns the user to use at their own risk.
Maybe even to cast the sorcery spell at a lower mana level than they usually train at so as to reduce the risk of backlash.  We might also need to add in handling for if someone blows their hand off and other more severe backlashes.

Even with all of that, I think it's fine to give people the option to take the risk if they want to.  I would have used it for Binu if it had been available.

Dartellum:
I agree with Binu. Choice is good if we can program it in. :smile:

profiles/base.yaml
```diff
@@ -731,6 +731,9 @@ use_own_ingot_type:
 deed_own_ingot: false
 # spells that get woven in to various scripts, do not use cambrinth
 crafting_training_spells:
+# Toggle Sorcery during crafting or outdoorsmanship. Sorcery will always be blocked by forge.lic
+#ADVANCED OPTION MAY RESULT IN ARREST, USE WITH CAUTION!!
+cast_sorcery_training: false
```
rpherbig, line 736: If this setting is intended to be used only in very specific circumstances (i.e. crafting or outdoorsmanship), what do you think about giving it a more specific name?

KatoakDR:
@MahtraDR To confirm, did #5307 supercede this PR?

MahtraDR:
> @MahtraDR To confirm, did #5307 supercede this PR?

Yes.

Issue #5154, opened 2021-09-04
[combat-trainer] dissect hangup after casting blend/hiding


echo1432:
[combat-trainer]>dissect troll
That's going to be hard to accomplish while concealed.
>
You feel fully rested.
>
The troll decomposes.  Not that it looks much different that way.
>
[combat-trainer: *** No match was found after 15 seconds, dumping info]
[combat-trainer: messages seen length: 10]
[combat-trainer: message: Blend  (3 roisaen)]
[combat-trainer: message: Oath of the Firstborn  (17 roisaen)]
[combat-trainer: message: Instinct  (12 roisaen)]
[combat-trainer: message: Essence of Yew  (22 roisaen)]
[combat-trainer: message: Senses of the Tiger  (20 roisaen)]
[combat-trainer: message: Claws of the Cougar  (4 roisaen)]
[combat-trainer: message: See the Wind  (16 roisaen)]
[combat-trainer: message: The troll decomposes.  Not that it looks much different that way.]
[combat-trainer: message: You feel fully rested.]
[combat-trainer: message: That's going to be hard to accomplish while concealed.]
[combat-trainer: checked against [/You'll gain no insights from this attempt/i, /You succeed in dissecting the corpse/i, /What exactly are you trying to dissect/i, /You'll learn nothing further/i, /While likely a fascinating study/i, /You cannot dissect/i, /would probably object/i, /should be left alone./i]]
[combat-trainer: for command dissect troll]

rpherbig:
@Hiinky

PR #5155, opened 2021-09-05
Update new Dissect match strings

Hiinky:
"That's going to be hard"  -  can't dissect while hidden
"That'd be a waste of time." - undead no longer valid


Hiinky:
Addresses #5154

PR #5157, opened 2021-09-06
[data] [base-hunting] Separating Leth dryads

LuckyThiefDR:
The twisted dryads (36) and dryad priestesses (34) spawn in different areas and have different rank ranges. It could be useful for people dealing with the low-level hunting ladder to start at 34 and move up to 36 later. This change leaves the 'dryads' zone in place, and moves 4 of those rooms into a separate 'dryad_priestess' zone.

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -968,6 +968,10 @@ hunting_zones:
   - 9773
   - 9772
   - 9767
+  - 9768
+  - 9769
+  - 9770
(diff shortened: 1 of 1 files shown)


LuckyThiefDR:
The second commit copies the 4 priestess rooms back into the 'dryads' zone they were taken from. This way, the original zone is unchanged for the people who are using it but they have the option of specifying the priestess rooms using the new zone.

PR #5162, opened 2021-09-08
TT Update: Wilderness Thicket Shuffle

Hiinky:
Quite a few areas were split into standard vs premie and rooms split to boost spawn.


LuckyThiefDR:
The two rock guardian areas are both named 'rock_guardians' - I think the new one was supposed to be named 'rock_guardians_premie' or something.

rpherbig:
Good catch, thank you @LuckyThiefDR !

PR #5164, opened 2021-09-08
[script] [combat-trainer] berserk training_abilities

LuckyThiefDR:
This adds berserk Landslide and berserk Avalanche as training abilities for warding and utility.  I find this useful on a smaller barbarian who does not gain much field exp from the research options.  It's preferable to have this feature rather than rely on buff_nonspells: which does not respect inner fire thresholds or field exp targets or roundtime conflicts.  Avalanche and Landslide are accessible early trainers which may become staples when the upcoming barbarian mastery split is live (though maybe we'll all be using Tenacity for warding early on).

If this makes sense, seems useful, and isn't obviously broken in some way, I'll look for people willing to test.


rpherbig:
Makes sense to me, code changes look good ✔️

combat-trainer.lic
```diff
@@ -2284,6 +2284,8 @@ class TrainerProcess
       'Barb Research Debilitation' => 'Debilitation',
       'Barb Research Utility' => 'Utility',
       'Barb Research Warding' => 'Warding',
+      'Landslide' => 'Warding',
+      'Avalanche' => 'Utility',
```
KatoakDR, line 2288: ❓  Thoughts on prefixing these terms to clarify what these abilities are? Similar to how `App` and `Barb Research` and `Khri` prefix a few other abilities?
```suggestion
      'Berserk Landslide' => 'Warding',
      'Berserk Avalanche' => 'Utility',
```

combat-trainer.lic
```diff
@@ -2487,6 +2492,14 @@ class TrainerProcess
...
     when /^Barb Research Utility$/i
       bput('meditate research prediction', 'You clear ')
+    when /^Landslide$/i
+      if DRStats.mana >= @barb_buffs_inner_fire_threshold
+        bput('berserk landslide', 'A supernatural', 'But you are already', 'Your inner fire lacks')
+      end
+    when /^Avalanche$/i
+      if DRStats.mana >= @barb_buffs_inner_fire_threshold
+        bput('berserk avalanche', 'The momentus rage', 'But you are already', 'Your inner fire lacks')
+      end
```
KatoakDR, line 2502: 💡  You may be able to reuse common-arcana helper methods here

```suggestion
    when /^Landslide$/i
      if DRStats.mana >= @barb_buffs_inner_fire_threshold
        DRCA. activate_barb_buff?('Landslide')
      end
    when /^Avalanche$/i
      if DRStats.mana >= @barb_buffs_inner_fire_threshold
        DRCA. activate_barb_buff?('Avalanche')
      end
```

KatoakDR (commented):
Interesting idea, thanks! I've made a couple suggestions to consider

combat-trainer.lic
LuckyThiefDR: Yeah, I think that's a good call.

combat-trainer.lic
LuckyThiefDR: Yeah, why not? I think the names here would need to be updated to reflect adding "Berserk ..." in the previous committed suggestion. I'll make that happen and test the updated version

combat-trainer.lic
```diff
@@ -2487,6 +2492,14 @@ class TrainerProcess
       bput('meditate research earthquake', 'You clear ')
     when /^Barb Research Utility$/i
       bput('meditate research prediction', 'You clear ')
+    when /^Berserk Landslide$/i
+      if DRStats.mana >= @barb_buffs_inner_fire_threshold
+        DRCA. activate_barb_buff?('Landslide')
```
rpherbig, line 2497: Does this work with a space in `DRCA. act.......`?

LuckyThiefDR: I wondered about that space, but it's working fine. Is that space not supposed to be there? I can take it out.

rpherbig:

LuckyThiefDR:
I've had all my characters running this as we've been collaborating - it works as planned for my barb and doesn't cause any problems for other guilds.

mikejcook:
It seems like it would be worth adding Cyclone, Wildfire, and Tornado (Augmentation) as well as Famine (Utility) and Flashflood (Warding). That would cover all of the early Barb berserks for training purposes.

PR #5165, opened 2021-09-14
[script] [combat-trainer] Barb maneuver toggle

LuckyThiefDR:
This adds a setting to allow barbs to keep using the 'Charged Maneuver' training ability beyond the field exp threshold set for training abilities. It's nice for young barbs to benefit from the increased IF regen applied to maneuver killing blows, and doublestrike is a decent weapon training tool pre-whirlwind.

Also, there is a small fix that adds missing match strings for attack commands related to using ambush stun on flying creatures and undead (issue reported by user).

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2413,6 +2413,9 @@ class TrainerProcess
 
     @barb_buffs_inner_fire_threshold = settings.barb_buffs_inner_fire_threshold
     echo("  @barb_buffs_inner_fire_threshold: #{@barb_buffs_inner_fire_threshold}") if $debug_mode_ct
+
+    @always_use_maneuvers = settings.always_use_maneuvers
+    echo("  @always_use_maneuvers: #{@always_use_maneuvers}") if $debug_mode_ct
   end
 
   def execute(game_state)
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2413,6 +2413,9 @@ class TrainerProcess
 
     @barb_buffs_inner_fire_threshold = settings.barb_buffs_inner_fire_threshold
     echo("  @barb_buffs_inner_fire_threshold: #{@barb_buffs_inner_fire_threshold}") if $debug_mode_ct
+
+    @always_use_maneuvers = settings.always_use_maneuvers
```
rpherbig, line 2417: This setting needs to be in `base.yaml`

rpherbig:
I'm not super familiar with the use case. Would it be useful to make the setting an array so you can, say, always use doublestrike but let powershot obey field exp?

mikejcook:
So, my understanding is based entirely on what epedia says, which is:

A chance to regenerate some Inner Fire when using a CCM Combat Charged Maneuver if an Expertise skill check is passed.

Based on my reading of that, we have a chance to benefit from any maneuver that would train Expertise, so it's not necessarily the double hit, but rather the execution of the maneuver that earns inner fire. So, assuming that's true, there's no benefit in limiting which maneuvers are executed.

rpherbig:
Code changes look fine aside from the one comment ✔️

mikejcook:
@rpherbig I know the author mentioned they were going to be away for a while on Discord. I've been using these changes live and they really help my innerfire. is there any way someone else can make the change to base.yaml? I looked and I'm not sure how to add a file to the PR (or maybe lack permission).

rpherbig:
I wasn't aware the author was going to be away for a while, thank you for letting me know. I happened to have a few minutes to spare and made the change to `base.yaml`.

PR #5167, opened 2021-09-17
Pick tweaks and bugfixes

wstampley:
Based on initial feedback to the pick rewrite, here are some more tweaks

- Fixing a bug I noticed with making sure certain containers are open beforehand
- Re-adding trap harvesting (based on request)
- Moving how armor is put back on (based on request in Discord from @Hiinky)
- Fixing a bug with skeleton keys
- Fixing a bug with non-pouch gem pouches
- Fixing a bug with loot_specials
- Fixing a bug with thief glance when there is a matching box on the ground in your room

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -717,7 +717,7 @@ class Pick
 
     special = @settings.loot_specials.find { |x| /\b#{x['name']}\b/i =~ item_long }
     if special
-      DRCI.put_away_item?(item_long, special['bag'])
+      DRCI.put_away_item?(item, special['bag'])
       return
     end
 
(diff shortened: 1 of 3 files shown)


pick.lic
```diff
@@ -90,8 +90,8 @@ class Pick
     @all_trap_messages = picking_data['traps']
     @disarmed_trap_messages = picking_data['disarmed_traps']
 
-    @harvest_traps = @settings.harvest_traps
-    @component_container = @settings.component_container
+    @harvest_traps = @settings.pick['harvest_traps'] || @settings.harvest_traps
```
rpherbig, line 93: Should these two be added to base.yaml? https://github.com/rpherbig/dr-scripts/blob/master/profiles/base.yaml#L962

wstampley: Added.

PR #5169, opened 2021-09-19
Move check_slivers to drcs for use elsewhere.

robbintt:
This can be used downstream for sliver management in the future.

### DIFF
--- a/common-summoning.lic
+++ b/common-summoning.lic
@@ -90,6 +90,46 @@ module DRCS
     return nil
   end
 
+  def check_slivers(game_state)
+    return unless DRStats.moon_mage?
+    return if game_state.casting
+    return if @tk_ammo
+    return unless @tk_spell
+    return if DRSpells.slivers
+    return if UserVars.moons['visible'].empty?
+    # With low skill you may fail to create the slivers
+    # when breaking your moonblade. If so, keep retrying.
+    retry_count = 0
+    max_retries = 3
+    created_slivers = false
+    # We don't need to cast the spell a fancy way
+    # so rather than asking users to define a new waggle
+    # then simply cast the spell at minimum prep.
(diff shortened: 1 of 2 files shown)


common-summoning.lic
```diff
@@ -90,6 +90,46 @@ module DRCS
     return nil
   end
 
+  def check_slivers(game_state)
```
wstampley, line 93: Should this be in DRCS or DRCMM?

Dartellum: It is summoning but we do have a `DRCMM`, `common-moonmage.lic` file. I believe it should go into there. Also, I assume you tested this and made sure when it calls this the `game_state` is known.

robbintt: I colocated it with moonblade, if I move it to drcmm, I will probably move more. Changing this to draft for now.

rpherbig:
A few thoughts:
1. This will have to be multiple PRs. The first adds function(s) to `common-*` files, the follow-up PRs then use those functions. The reason is that we have to give changes to common files a few to be picked up by players or it can cause problems.
2. If the code is MM-specific, I recommend putting it in common-moonmage. There is some logical overlap between MM and S, and arguably most of S could be "WM".

KatoakDR:
@robbintt Can you share what kind of slivers management you're thinking of? 

I may be a bit short sighted on the utility of this function. To me it feels very combat-trainer specific to quickly create some slivers for use as TKT ammo.

As-is, I'm inclined to say the method is fine where it is right now until we have definitive plans to reuse it outside of combat-trainer.

Thanks!

KatoakDR:
Closing due to inactivity. Feel free to reopen this pull request and address code review feedback if you'd like to pursue this further. Thanks!

PR #5170, opened 2021-09-19
Update Logic for using Crafted

Hiinky:
Previous logic excluded crafted remedies being called from ;combat  `wait_for_script_to_complete('heal-remedy', ['quick'])` because it doesn't/can't send the override arg without a fair bit of tinkering.  This returns to the previous functional conditions while including the override option for on-the-spot healing.

### DIFF
--- a/heal-remedy.lic
+++ b/heal-remedy.lic
@@ -37,6 +37,9 @@ if @settings.herbs && !@override_mode
 elsif @override_mode
   $remedies = get_data('remedies')
   DRC.message("first remedies condition") if $debug_mode_hr
+else
+  $remedies = get_data('remedies')
+  DRC.message("first remedies condition") if $debug_mode_hr
 end
 
 @wounds_applied = FALSE
--- a/heal-remedy.lic
+++ b/heal-remedy.lic
@@ -34,9 +34,6 @@ DRC.message("@override_mode: #{@override_mode}") if $debug_mode_hr
 if @settings.herbs && !@override_mode
   $remedies = get_data('herbs')
   DRC.message("herbs") if $debug_mode_hr
-elsif @override_mode
-  $remedies = get_data('remedies')
-  DRC.message("first remedies condition") if $debug_mode_hr
 else
   $remedies = get_data('remedies')
   DRC.message("first remedies condition") if $debug_mode_hr


Dartellum (approved):

heal-remedy.lic
```diff
@@ -37,6 +37,9 @@ if @settings.herbs && !@override_mode
 elsif @override_mode
   $remedies = get_data('remedies')
   DRC.message("first remedies condition") if $debug_mode_hr
+else
+  $remedies = get_data('remedies')
```
rpherbig, line 41: I'm unclear on what the difference is between this case and line 37.

If there is no difference, should line 37 change to `else`?

Hiinky: if @settings.herbs && !@override_mode   
-- uses storebought herbs if herbs: are in yaml with no override arg.

elsif @override_mode  
-- is only an arg to manually start the script to use expensive/limited crafted remedies on demand for better results for things like dark box or when badly hurt while still using cheap storebought during ;combat.  

else
-- required for folks that want to use their crafted remedies during ;combat because ;combat doesn't have handling to send the override arg which effectively excludes crafted.

Dartellum: We could update combat-trainer to check for the override and put the call in for it. 🤷‍♂️

Hiinky: Seems like a couple added lines here is a lot less trouble than adding more to the beast that is ;combat

rpherbig: What I'm saying is, the code seems identical in the two cases, so why not remove the `elif` and just make it an `else` (so it covers both cases)?

Hiinky: That would block folks from using crafted remedies on the fly?  Unless I missed something.  You would have to comment out your herbs: in the yaml to make it use crafted.

rpherbig: It won't block anything. All we're doing is combining two cases into the more-accepting branch (i.e. `else`).

This is all based on the fact the code is identical within the two branches. If it should be different between the two branches, then we'll need both conditionals.

Hiinky:
Once I had a moment to stop and walk my tired brain through the options, I see what you mean.  I'm totally fine just make it an else.

PR #5171, opened 2021-09-19
Add option to balance lockpick ring.

robbintt:
Default to off since this is a waste of time unless you're a thief or frugal.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -237,8 +237,6 @@ class Pick
     bput("turn my #{@settings.lockpick_container} to best", "You fiddle with", "You think about it")
   end
 
-  end
-
   def stop_picking?
     @stop_pick_on_mindlock && DRSkill.getxp('Locksmithing') >= 30
   end
--- a/pick.lic
+++ b/pick.lic
@@ -104,6 +104,7 @@ class Pick
       echo "- Sources: #{@sources}"
       echo "- use_lockpick_ring: #{@use_lockpick_ring}"
       echo "- lockpick_container: #{@lockpick_container}"
+      echo "- balance_lockpick_container: #{@balance_lockpick_container}"
       echo "- harvest_traps: #{@harvest_traps}"
       echo "- component_container: #{@component_container}"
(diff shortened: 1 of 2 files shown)


pick.lic
```diff
@@ -231,6 +232,11 @@ class Pick
     @equipment_manager.wear_items(@removed_items)
   end
 
+  def balance_lockpick_container
+    return unless @use_lockpick_ring and @balance_lockpick_container
```
wstampley, line 236: I think you'll need something at the top to pull @settings.balance_lockpick_container into the class variable @balance_lockpick_container, and probably to add it to the debug output of all the settings.

pick.lic
```diff
@@ -231,6 +232,11 @@ class Pick
...
+    bput("turn my #{@settings.lockpick_container} to best", "You fiddle with", "You think about it")
```
wstampley, line 237: This version of pick doesn't do the Ruby include of any commons to make it easier to unit test, so you will need to prefix this bput with DRC.

profiles/base.yaml
```diff
@@ -999,6 +999,7 @@ skeleton_key: skeleton key
 # Set to false to use lockpicks instead.
 use_skeleton_key: false
 use_lockpick_ring: true
+balance_lockpick_container: false
```
wstampley, line 1002: Other new properties that are only used the new pick were nested under "pick:", i.e.
```
pick:
  balance_lockpick_container: false
```
Obviously not a 'must change', just something to consider.

pick.lic
```diff
@@ -39,6 +39,7 @@ class Pick
 
     @use_lockpick_ring = @settings.use_lockpick_ring
     @lockpick_container = @settings.lockpick_container
+    @balance_lockpick_container = @settings.pick['balance_lockpick_container']
```
rpherbig, line 42: Should this be output inside the block on line 101?

rpherbig:
Code changes look good aside from the two comments ✔️

PR #5172, opened 2021-09-19
Refactor tarantula combat detection

robbintt:
Easier for players to expand combat script detection:
* allow list to be moved to settings later
* move decision to function

### DIFF
--- a/tarantula.lic
+++ b/tarantula.lic
@@ -40,17 +40,20 @@ class Tarantula
     UserVars.tarantula_last_skillset = @last
   end
 
-  def choose_skill
-    _respond( bold("*** Choosing skill ***") ) if @debug
-    skills = []
-    in_combat = false
+  def in_combat?
     @combat_scripts.each do |script|
       if Script.running?(script)
         in_combat = true
         break
       end
     end
-    if in_combat
+  end
+
+  def choose_skill
+    _respond( bold("*** Choosing skill ***") ) if @debug
+    skills = []
+    in_combat = false
+    if in_combat?
       @tarantula.each.reject{ |k,_| k == @last }.each{ |k,v| skills << v["combat"] if v["combat"] }
     else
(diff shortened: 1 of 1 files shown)


tarantula.lic
```diff
@@ -39,10 +40,19 @@ class Tarantula
     UserVars.tarantula_last_skillset = @last
   end
 
+  def in_combat?
+    @combat_scripts.each do |script|
```
rpherbig, line 44: I think you could simplify this to `@combat_scripts.any? { |script| Script.running?(script) }`?

rpherbig:
I'm fine to merge this, or you can do the refactor. Thoughts?

tarantula.lic
robbintt: Do you know about this `running?` method in lich.rbw? Looks like it might take a list but has some other conditions as well. Could be much simpler: 

https://github.com/dragon-realms/dr-lich/blob/master/lich.rbw#L9790

robbintt: Interesting, yeah it works as expected afaict:
``` 
  1 scripts = ["banana", "potato", "dependency"]
  2 running?(scripts) ? echo("#{scripts} -- something in here is running") : echo("nothing in the list is running"
```

wstampley: Interesting. Based on a quick read of the code in that function, I would have expected it to require that ALL of the scripts in the list are running.

There are some interesting edge cases with the way that works today:
```
R>;la
--- Lich: dependency, roomnumbers, drinfomon, afk, spellmonitor, smartlisten, attention, sanowret-crystal, skill-recorder, moonwatch, t2, events, jail-buddy, hunting-buddy, textsubs, combat-trainer, lnet, log

>;e echo running?(['notcombat-trainer'])
[exec1: true]

R>;e echo running?(['foo', 'fi', 'bar'])
[exec1: true]
```

robbintt: good eye, the regex match is definitely not what we want. i will probably go back to the recommendation from @rpherbig

rpherbig (approved):

PR #5174, opened 2021-09-23
Random cleanup

wstampley:
- Improving vis of warnings in combat related to shock + spells (switch from echo -> message)
- Fixing bug with return value of DRCA.cast_spells, adding wrapper cast_spells? to follow the ruby convention
- Fixing similar bug in common theurgy, returning the right value for holy water checking
- Adding a recipe I forge all the time to the recipe list
- Fixing room name for Theren vault
- Fixing race condition in hunting-buddy, where your stop-on becomes true while go2'ing your hunting group

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -587,6 +587,11 @@ module DRCA
     pause 0.5 while UserVars.moons.empty?
   end
 
+  def cast_spell?(data, settings, force_cambrinth = false, cast_lifecycle_lambda = nil)
+    result = cast_spell(data, settings, force_cambrinth, cast_lifecycle_lambda)
+    result ? true : false
+  end
+
   def cast_spell(data, settings, force_cambrinth = false, cast_lifecycle_lambda = nil)
     return unless data
     return unless settings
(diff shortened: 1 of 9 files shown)


common-arcana.lic
```diff
@@ -587,6 +587,11 @@ module DRCA
     pause 0.5 while UserVars.moons.empty?
   end
 
+  def cast_spell?(data, settings, force_cambrinth = false, cast_lifecycle_lambda = nil)
+    result = cast_spell(data, settings, force_cambrinth, cast_lifecycle_lambda)
+    result ? true : false
```
rpherbig, line 592: Why not return `cast_spell` directly?

rpherbig:
Code changes look good ✔️

common-arcana.lic
wstampley: Cause ? methods are supposed to return a boolean, but not all paths through cast_spell actually do, so this forces the weirder paths (like rituals?) into a boolean until a better refactor gets done.

rpherbig: Makes sense to me, thanks

PR #5177, opened 2021-09-26
Add mindstate_goal argument to performance.

robbintt:
Allow mindstate_goal in performance arguments for a smaller training goal.

### DIFF
--- a/performance.lic
+++ b/performance.lic
@@ -22,7 +22,7 @@ class Performance
     @settings = get_settings
     @song_list = get_data('perform').perform_options
     skip_clean = args.noclean
-    @mindstate_goal = 30 # default
+    @mindstate_goal = 31
     if args.mindstate_goal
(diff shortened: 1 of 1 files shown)


performance.lic
```diff
@@ -12,7 +12,8 @@ class Performance
     start_time = Time.new.to_i
     arg_definitions = [
       [
-        { name: 'noclean', regex: /noclean/i, optional: true, description: 'Skip cleaning your instrument' }
+        { name: 'noclean', regex: /noclean/i, optional: true, description: 'Skip cleaning your instrument' },
+        { name: 'mindstate_goal', regex: /(mindstate_goal=(\d+))/, optional: true, description: 'Train until mindstate >= mindstate_goal' }
```
rpherbig, line 16: Why introduce text into the arg? Rather than simply making it an integer?

PR #5181, opened 2021-10-04
Improving handling for harvest + no component container

wstampley:
The data in base-picking for trap parts wasn't specific enough for people who are harvesting but not keeping trap components, which would leave them holding a trap part in their left hand, and cause problems picking the box.

This PR:
- Fixes the base-picking data based on the table here: https://elanthipedia.play.net/Locksmithing_skill (which could still be wrong? we'll see)
- Adds an error handling path in `pick identify` that clears your off hand and continues
- Adds more debugging lines for future use

### DIFF
--- a/data/base-picking.yaml
+++ b/data/base-picking.yaml
@@ -176,21 +176,19 @@ picking:
   - steel striker
   - black cube
   - chitinous leg
-  - spring
+  - coiled spring
   - brown clay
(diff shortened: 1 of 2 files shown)


wstampley:
I've been running the updated version just fine, and a couple others are running it from Discord. Should be good to go.

PR #5185, opened 2021-10-09
rezz.lic update - Find bodies with roombrief on

BinuDR:
The script doesn't detect dead people if you have RoomBrief on.

[rezz]>look people
You take a moment to look for everybody in the area and see the body of Tohaka (prone).

I've updated the regex to deal with this.


rpherbig:
Regular expressions 🤷 ✔️

PR #5188, opened 2021-10-11
[script] [t2] QOL Updates: controlled shutdown, 'naming' of sections

vtcifer:
- Adds the ability to gracefully shutdown/cancel shutdown of  T2 at next section, using:
   `;e $T2.shutdown` / `;e $T2.noshutdown`
  (It is recommended to setup an alias)
  - After shutdown, will run list of scripts defined by `t2_after_shutdown:`, which defaults to
      ```yaml
      t2_after_shutdown:
        - gosafe
        - sell-loot
        - gosafe
      ```
- Adds the ability to name sections, so shorter output is generated when starting up sections:
    ```yaml
    training_list:
      - skill:                  # Combat - main hunt
        - Targeted Magic
        - Debilitation
        - Small Edged
        - Large Edged
        - Shield Usage
        - Chain Armor
        # [SNIP - shortened to not take up space]
        name: Combat - Main Hunt
        start: 9
        scripts:
          - hunting-buddy setup
          - crossing-repair
    ```
    If no name is specified, will default to old style list from `skill:`

### DIFF
--- a/t2.lic
+++ b/t2.lic
@@ -12,10 +12,12 @@ class T2
   include DRCT
 
   def initialize
+    @shutdown = false
     fput('awaken')
 
     @counter = 0
     @settings = get_settings
+    @after_shutdown = @settings.t2_after_shutdown
     UserVars.t2_timers ||= {}
 
     t2_avoids = @settings.t2_avoids
@@ -26,8 +28,11 @@ class T2
         fput("avoid #{avoid['type']}")
       end
     end
+  end
 
+  def run
     loop do
+      break if @shutdown
       trainables = @settings['training_list']
       EquipmentManager.new.empty_hands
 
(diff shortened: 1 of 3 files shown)


t2.lic
```diff
@@ -70,6 +80,25 @@ class T2
       wait_for_script_to_complete(script_name, action_parts)
     end
   end
+
+  def shutdown 
+    if @shutdown == false
```
rpherbig, line 85: It's minor, but I have two preferences around this style:
1. not explicitly check `== false` (or `== true`)
2. whenever possible, check the `true` case first

So something like:
```
if @shutdown
    # output it's already set to shutdown
else
    # set @shutdown and output
end
````

File: t2.lic [hunk 2]
```diff
@@ -41,12 +46,17 @@ class T2
...
+        else
+          echo "***STATUS*** Starting (#{trainable['start']}) #{trainable['name']}"
+        end
         execute_actions(trainable['scripts'])
-        update_cooldown trainable['skill'] if has_cooldown? trainable['skill']
+        update_cooldown(trainable['skill']) if has_cooldown?(trainable['skill'])
         break
       end
     end
+    execute_actions(@after_shutdown) unless @after_shutdown.nil?
```
rpherbig, line 59: If I'm reading this PR right, `@after_shutdown` should never be `nil`, only empty.

Should this check be removed, changed to `.empty?`, or left as-is? Does `execute_actions` handle an empty array correctly?

rpherbig:
This is ready to merge after the conditional is confirmed to be correct as-is or changed. ✔️

t2.lic
vtcifer: Q on the the preferred style preferences:

```ruby
    if !@after_shutdown.empty?
      echo '***STATUS*** Controlled shutdown, executing shutdown actions.'
      execute_actions(@after_shutdown) 
    end
```
vs
```ruby
    unless @after_shutdown.empty?
      echo '***STATUS*** Controlled shutdown, executing shutdown actions.'
      execute_actions(@after_shutdown) 
    end
```

I tend to avoid `unless`, unless it's at the end, as it feels awkward and confusing when used before the code block itself to me, but that could because I'm not used to it as a Boolean check, but am open to what is considered more standard ruby practices.

PR #5207, opened 2021-10-27
Fix common-arcana telescope storage

wstampley:
common-arcana uses 'stow' instead of correctly putting things back where it should, which then makes my other scripts very angry.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1814,7 +1814,7 @@ class SpellProcess
       @equipment_manager.stow_weapon(game_state.weapon_name)
     end
 
-    data = update_astral_data(data)
+    data = update_astral_data(data, @settings)
     if data # update_astral_data returns nil on failure
       check_invoke(game_state)
       ritual(data, @settings)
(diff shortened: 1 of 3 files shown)


Dartellum:
I left a comment, or review, on why using DRCI instead of DRCMM. Checkout DRCMM and you will see the store telescope method.

asechrest:
Hi, all.

We've had additional reports of this issue in Discord. It looks like this PR is stalled for a couple reasons: 1) needs to be split into two PRs since part of the changes effect a common-*, and 2) should probably utilize DRCMM which has a ready-made function for putting away a telescope.

I don't want to step on any toes, but if this is stalled, we've got some fixes ready. Ya'll just let me know.

rpherbig:
> I don't want to step on any toes, but if this is stalled, we've got some fixes ready. Ya'll just let me know.

Go ahead and open a new PR, just note the overlap in it.

MahtraDR:
Closing. resolved in https://github.com/rpherbig/dr-scripts/pull/5544

PR #5220, opened 2021-11-08
[script][common-healing][update] Added worn trash parameters to dispo…

Dartellum:
…se call. If no worn trash parameters are in yaml, the default (old) action occurs. The option for worn trash is already present in `common-items.`

### DIFF
--- a/common-healing.lic
+++ b/common-healing.lic
@@ -330,8 +330,8 @@ module DRCH
     waitrt?
     case result
     when *tend_dislodge
-      DRCI.dispose_trash(DRC.left_hand) if DRC.left_hand != snap.first
-      DRCI.dispose_trash(DRC.right_hand) if DRC.right_hand != snap.last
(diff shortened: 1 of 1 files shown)


common-healing.lic
```diff
@@ -330,8 +330,8 @@ module DRCH
     waitrt?
     case result
     when *tend_dislodge
-      DRCI.dispose_trash(DRC.left_hand) if DRC.left_hand != snap.first
-      DRCI.dispose_trash(DRC.right_hand) if DRC.right_hand != snap.last
+      DRCI.dispose_trash(DRC.left_hand, get_settings.worn_trashcan, get_settings.worn_trashcan_verb) if DRC.left_hand != snap.first
```
rpherbig, line 333: I'm not a fan of calling `get_settings` four times, or even calling it at all inside this method (it should be provided by the calling script).

That said, it won't hold up merging, it would just be nice to have it passed in (which I realize is a bigger change).

PR #5222, opened 2021-11-09
[script][smash-shells][update]Changed walk_to location in script so y…

Dartellum:
…ou stay in the room with the bucket when checking if wounded. Fragments need to go into the bucket or else. Otters are now performing social justice on people littering their room with fragment.

### DIFF
--- a/smash-shells.lic
+++ b/smash-shells.lic
@@ -11,11 +11,11 @@ trash_items = settings.hollow_eve_junk.map { |x| /\b#{x}\b/i }
 hollow_eve_loot_container = settings.hollow_eve_loot_container
 
 DRC.message("*** Heading to grab a shell ***")
-DRCT.walk_to(16_236)
+
 loop do
   break if checkbleeding
   break if checkhealth < 50
-
+  DRCT.walk_to(16_236)
   until checkright || checkleft
     DRC.fix_standing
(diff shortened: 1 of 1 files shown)


Dartellum:
Giving to the Discord chat so some can try. As always, working for me. 

https://discord.com/channels/745675889622384681/771867662551744534/907287447816253510

Dartellum:
Working as expected, per some testers.

smash-shells.lic
```diff
@@ -66,12 +66,11 @@ loop do
...
 
   waitrt?
-  DRCT.walk_to(16_236)
   next unless Flags['shell-fragments']
   pause 4
   DRC.wait_for_script_to_complete('tendme')
-  pause 20
-  DRC.wait_for_script_to_complete('tendme')
+  #pause 20
+  #DRC.wait_for_script_to_complete('tendme')
```
rpherbig, line 73: Why was this commented out?

Dartellum: Tendme was fixed to get all lodged and bleeders in one go. I left in just in case it was needed but I have run it since the beginning of HE and have not had a problem, other than getting killed by an otter horde because I dropped the fragments in the room.

rpherbig: Cool, thanks

PR #5226, opened 2021-11-14
[script] [common-items] Look in watery portal if fail to get item from the container the first time

KatoakDR:
### Background
* For some reason the GameMasters [changed](http://forums.play.net/forums/DragonRealms/Discussions%20with%20DragonRealms%20Staff%20and%20Players/Game%20Master%20and%20Official%20Announcements/view/1899
) how the watery portal (eddy) from Hollow Eve 2020 works and it now requires you to periodically `LOOK` in the container before you can get items from it.

### Changes
* In `common-items`, when fail to get an item from the container and that container includes the word `portal` then retry once after looking in the portal.

### Tests
* See `test/test_common_items.rb`

```
> get bloodwood branch from my watery portal
What were you referring to?

> look in my watery portal
[Assuming you mean a swirling eddy of incandescent light bound by a gold-striated coralite frame.]
In the swirling eddy you see ...

> get bloodwood branch from my watery portal
You get a curving bloodwood branch from inside a swirling eddy of incandescent light bound by a gold-striated coralite frame.
```

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -557,6 +557,23 @@ module DRCI
     when *@@get_item_success_patterns
       return true
     else
+      if container =~ /\bportal\b/i
+        return get_item_from_eddy_portal?(item, container)
+      else
+        return false
+      end
+    end
+  end
+
+  # Workaround to game changes where you must periodically look in
+  # the portal for the contents to be available.
+  # http://forums.play.net/forums/DragonRealms/Discussions%20with%20DragonRealms%20Staff%20and%20Players/Game%20Master%20and%20Official%20Announcements/view/1899
+  def get_item_from_eddy_portal?(item, container)
+    DRCI.look_in_container(container)
+    case DRC.bput("get #{item} #{container}", @@get_item_success_patterns, @@get_item_failure_patterns)
+    when *@@get_item_success_patterns
+      return true
+    else
       return false
     end
   end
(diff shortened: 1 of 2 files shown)


Dartellum:
We need to tread carefully with this enhancement as the GMs did what they did because of the amount of inventory loaded in the system was consuming way too many resources. 

```
Lyneya — 10/28/2021
We had over 150K items in that one room that they're stored in.  It was causing a "heartbeat" pattern in the system performance every time that segment saved out.  For comparison, the next largest segment in terms of item count has 9k items in it.

Lyneya — 10/28/2021
Right now, there's so few items in the seg that it doesn't matter if everyone online has theirs loaded all of the time (600ppl at 200 items is 120K items and we were ok with the 152K, but it was showing itself on the performance charts).  More people will have them after Hollow Eve.  If I want to expand their size, depending on how much I add or what methods I use it might become an issue down the road.  So we'll take a while to see how player behavior handles the update before I decide whether you can have more space in them, and how much of it we can spare.
```

vtcifer:
> We need to tread carefully with this enhancement as the GMs did what they did because of the amount of inventory loaded in the system was consuming way too many resources.

I think this change is in the spirit of what GMs are looking for (forcing a load by looking in it only when trying to get something out of it).  The thing we need to not do is basically just check the portal every X time unit, even if you're not using it.

PR #5227, opened 2021-11-14
[script] [t2] Allow reload of settings on the fly

vtcifer:
Adds the ability to reload t2 settings (really `training_list:` and `t2_after_shutdown:` settings).

`;e $T2.reload_settings` - Will reload, no output.

`;e $T2.reload_settings(true)` - Will reload, and output the `training_list:` settings in yaml format.

It does some basic checks to ensure not reloading malformed settings, regardless of above.

(It is recommended to setup an alias).

### DIFF
--- a/t2.lic
+++ b/t2.lic
@@ -17,11 +17,9 @@ class T2
 
     @counter = 0
     @settings = get_settings
-    @after_shutdown = @settings.t2_after_shutdown
     UserVars.t2_timers ||= {}
 
-    t2_avoids = @settings.t2_avoids
-    t2_avoids.each do |avoid|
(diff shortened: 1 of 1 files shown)


t2.lic
```diff
@@ -17,11 +17,9 @@ class T2
 
     @counter = 0
     @settings = get_settings
```
rpherbig, line 19: Why not call `reload_settings` here so we check for validity at first startup (and not just on reload)?

vtcifer: Probably not a bad idea, I've accidently run it before on characters that weren't setup for T2 yet and it took me a bit to figure out why it kept stowing things over and over.

PR #5228, opened 2021-11-14
[script] [common-money] Add return value to deposit_coins of balance and currency

vtcifer:
Changes to be used in PR #5229

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -137,7 +137,27 @@ module DRCM
       return
     end
     minimize_coins(keep_copper).each { |amount| withdraw_exact_amount?(amount, settings) } if settings.hometown == hometown
-    DRC.bput('check balance', 'your current balance is', 'You find your jar with little effort', 'Searching methodically through the shelves')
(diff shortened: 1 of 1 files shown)


common-money.lic
```diff
@@ -137,7 +137,27 @@ module DRCM
       return
     end
     minimize_coins(keep_copper).each { |amount| withdraw_exact_amount?(amount, settings) } if settings.hometown == hometown
-    DRC.bput('check balance', 'your current balance is', 'You find your jar with little effort', 'Searching methodically through the shelves')
+    case DRC.bput('check balance',
```
rpherbig, line 140: IIRC, the `check balance` here was to refresh something in `drinfomon` that already tracked money. How does that differ from this?

vtcifer: The benefit here is immediate return, without needing to cross reference which bank / hometown etc.

vtcifer: Also I think CharSettings are only accessible to the script that set them (so technically that data is only accessible to drinfomon).

PR #5230, opened 2021-11-15
[script] [textsubs] Add option to textsubs to group at plat level

vtcifer:
Allows users to group money at plat level, isntead of just based on thousands.  Ex:

Default (currenty):
`textsubs_use_plat_grouping: false`
 > 1219 platinum, 56 gold, 64 silver, 59 bronze, and 56 copper Kronars (12,253,046 copper Kronars).
  No Lirums.

New:
`textsubs_use_plat_grouping: true`
 > 1219 platinum, 56 gold, 64 silver, 59 bronze, and 56 copper Kronars (1,225,3046 copper Kronars).
  No Lirums.

Also groups textsubs settings in base.yaml together.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -23,10 +23,6 @@ t2_after_shutdown:
   - sell-loot
   - gosafe
 
-# TextSubs
-private_textsubs:
-  # old text: new text
-
 # Combat settings
 # https://elanthipedia.play.net/Lich_script_repository#combat-trainer
 aim_fillers:
@@ -1807,9 +1803,15 @@ sorter:
 
 stabbity:
 
+# TextSubs
+private_textsubs:
+  # old text: new text
 # For use with shitlist and textsubs
 # Will allow textsubs to put an (SL) at the end of the names of people on your list
 integrate_shit_list_with_textsubs: false
+#turn on heal/remedy text subs by ingredient name: yelith (limb-IW) tonic
+herb_textsubs: false
+textsubs_use_plat_grouping: false
 
 # Settings for smoke script
 smoke:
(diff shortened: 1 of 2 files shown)


rpherbig:
I don't think I understand what this setting is doing.

The two outputs look identical to me except for the comma placement.

Why do they both say copper? I assumed "group money at plat level" means we'd see something like `(1,225.3046 platinum Kronars)`.

In the second example, why are there four numbers are the last comma? What does `5,3046` mean numerically?

vtcifer:
> The two outputs look identical to me except for the comma placement.

That's exactly what it does, it changes  where the first (right-to-left) set of comma 'grouping' happens. the idea here is that, generally, RL comma grouping follows the hundreds.  But in DR specifically, most people think in terms of Plat (once you get really high numbers), so group first at plat level, and from there normally.

I didn't want to change it to a decimal dot, because as you also noted, it should then change the copper to plat, but not everywhere has that extra bit, making it more compelx to deal with.

`textsubs_use_plat_grouping: false`
```
> shop staff
Short: a slender alabaster staff
  Tap: a slender alabaster staff with a silver cobra coiled along it
 Cost: 1,437,500 Kronars
 Look: It is labeled "Holy".
 Read: There is nothing there to read.
You can buy this item if you like.
```

`textsubs_use_plat_grouping: true`
```
> shop staff
Short: a slender alabaster staff
  Tap: a slender alabaster staff with a silver cobra coiled along it
 Cost: 143,7500 Kronars
 Look: It is labeled "Holy".
 Read: There is nothing there to read.
You can buy this item if you like.
```

> In the second example, why are there four numbers are the last comma? What does 5,3046 mean numerically?

It's just to facilitate 'mental' grouping.

rpherbig:
I don't think I really see the use case, but this is optional and seems to do what you want it to, so it's all good.

PR #5231, opened 2021-11-15
[script] [attunement] Add support for sigil walking (scholarship training)

vtcifer:
- Adds support to do sigil walking to train scholarship
- Cleans up target mindstate code

Priority for list of sigil walking rooms:
1. `sigil_walk_rooms:`
2. hometown:sigil_walk_rooms (from base-town.yaml)
3. hometown:attunement_rooms (from base-town.yaml)

### DIFF
--- a/attunement.lic
+++ b/attunement.lic
@@ -11,10 +11,24 @@ class Attunement
 
   LUNAR_PERCEIVE_COMMANDS = ['mana', 'moons', 'planets']
 
+  PRIMARY_SIGILS_PATTERN = /\b(?:abolition|congruence|induction|permutation|rarefaction) sigil\b/
+  SECONDARY_SIGILS_PATTERN = /\b(?:antipode|ascension|clarification|decay|evolution|integration|metamorphosis|nurture|paradox|unity) sigil\b/
+
   def initialize
     arg_definitions = [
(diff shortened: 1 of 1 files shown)


rpherbig:
The code looks fine, but I wonder about a bigger issue for this script. It's named `attunement`, but sometimes trains Empath and sometimes Scholarship.

There are so many `if`s scattered around it really feels like three different scripts that share the concept of "walk to these rooms and issue this command". I wonder if we should make the walking logic more common and break this up into separate scripts? Just rename it to be more generic? ...

vtcifer:
> The code looks fine, but I wonder about a bigger issue for this script. It's named `attunement`, but sometimes trains Empath and sometimes Scholarship.
> 
> There are so many `if`s scattered around it really feels like three different scripts that share the concept of "walk to these rooms and issue this command". I wonder if we should make the walking logic more common and break this up into separate scripts? Just rename it to be more generic? ...

Indeed there was some discussion on that in the discord, and I think that's ultimate the right way way forward, but that even with this adding to the current technical debt, it's still better to have it available, and on a longer scale refactor to be more generic. (Which tbh, some of the discussion was over my head wrt ruby knowledge).

Issue #5224, opened 2021-11-13
transfer-items


simmish:
Please add transferring of items from a worn/held container into another container not on oneself.  For instance, transferring from a backpack to a donation shelf or trashcan.  would be amazing!

asechrest:
I'm pretty green but I'll give this a whirl. No timeframe. On Discord there was a suggestion to make this a separate script from transfer-items, which is primarily designed to safely transfer items between things on your own person. Transferring off your person would be inherently unsafe, thus the suggestion to make it separate. I'm thinking "offload-items" or something similarly named.

simmish:
Hey no timeframe needed!  I tried to figure it out but my eyes cross with ruby and dr stuff.  Probably doesn't help I'm a novice at it.

Makes sense for a separate script for sure.  Just for the backlash people would post if they weren't paying attention or made a typo

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

asechrest:
@rpherbig  This can be closed via #5232

PR #5232, opened 2021-11-15
Add offload items script

asechrest:
Responding to #5224 which requests a way to offload items from your person to something like a donation shelf which is not on your person. The original request asked for additional functionality to `;transfer-item`. However, discussion in Discord suggested a separate script from `;transfer-item` since the requested functionality is an inherently unsafe script to take many items from your person and offload to a container not on your person. The existing `;transfer-item` is functionally designed to safely transfer items from one owned container to another contain on your person or that you own.

Created new script, borrowing heavily from `;transfer item`, to offload all items from a source container on your person to another container not on your person. Optionally accepts a noun argument that will offload all found items of 'noun' from the source container on your person to the destination container not on your person.

Warns the script user that this is unsafe and requires 1 affirmative `;unpause` (if transferring limited items of <noun>) or 2 affirmative `;unpause` (if transferring every item from source container) to complete the script.

Testers welcome before merging. I'm not yet proficient at scripting/Ruby.

### DIFF
--- a/offload-items.lic
+++ b/offload-items.lic
@@ -76,7 +76,7 @@ class OffloadItems
     DRC.message("\n")
     
     unless noun
-      DRC.message("You are about to move EVERY SINGLE ITEM from from #{source} to #{destination}.")
+      DRC.message("You are about to move EVERY SINGLE ITEM from #{source} to #{destination}.")
       DRC.message("Unpause the script to continue. Otherwise, ;kill the script to stop.")
       pause_script
       DRC.message("Are you SURE you want to offload EVERY SINGLE ITEM from YOUR #{source} to #{destination}?")
--- a/offload-items.lic
+++ b/offload-items.lic
@@ -18,12 +18,14 @@ class OffloadItems
       ]
     ]
     args = parse_args(arg_definitions)
+
+    # Warn the script user that this is designed to offload items off your person
     warn_before_offload(args.source, args.destination, args.noun)
+    # Call the main method. If no preposition was specified, default to "in"
     offload_items(args.source, args.destination, args.preposition || "in", args.noun)
   end
 
   def offload_items(source, destination, preposition, noun)
-    echo "Preposition is #{preposition}"
     # If container is very full then LOOK may not list all of them.
     # If you're moving a specific item, then sort those to the top
     # to increase chances we find and move all of them in one go.
@@ -37,7 +39,6 @@ class OffloadItems
         # We need to look again to see what we can now see.
         # Keep doing this until fill the destination or exhaust source.
         if item =~ /lot of other stuff/
-          echo "About to REcall offload items"
           offload_items(source, destination, preposition, noun)
           return
         end
@@ -47,8 +48,8 @@ class OffloadItems
             trash_item(item)
           else
             unless put_away_item_unsafe?(item, destination, preposition)
+              # Failed. Message user and return item to source container.
               DRC.message("Unable to offload #{item} to #{destination}. It may not accept items, or you may need to change the preposition (e.g. 'on' vs. 'in')")
-              # Failed. Return item to source container.
               DRCI.put_away_item?(item, source)
               exit
             end
--- a/offload-items.lic
+++ b/offload-items.lic
@@ -1,0 +1,158 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#offload-items
+=end
+
+custom_require.call %w[common common-items]
+
+class OffloadItems
+  include DRC
+  include DRCI
+
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'source', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Source container' },
+        { name: 'destination', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Destination container including preposition. E.g. "on shelf", "in bucket"' },
+        { name: 'noun', regex: /^[A-z\.\s\-]+$/i, optional: true, variable: true, description: 'If specified, only items with this noun will be transferred.'}
+      ]
+    ]
+    args = parse_args(arg_definitions)
+    offload_items(args.source, args.destination, args.noun)
+  end
+
+  @@offload_items_success_patterns = [
+    /^You drop/i,
+    /^You set/i,
+    /^You put/i
+  ]
+
+  # Messages that when trying to put away an item you're warned.
+  # To continue you must retry the command.
+  @@offload_items_retry_patterns = [
+    # You may get the next messages when an outdated item is updated upon use.
+    # "Something appears different about the <item>, perhaps try doing that again."
+    # Example: https://elanthipedia.play.net/Item:Leather_lotion
+    /Something appears different about/,
+    /perhaps try doing that again/
+  ]
+
+  @@offload_items_failure_patterns = [
+    /^You can't put that there/i,
+    /^What were you referring to/i,
+    /^Please rephrase that command/,
+    /^I could not find what you were referring to/,
+    /^You can't put that in|on|under|behind there/i,
+    /^Stow what/,
+    /^Please rephrase that command/,
+    /^What were you referring to/,
+    /^I could not find what you were referring to/,
+    /even after stuffing it/,
+    /is too .* to (fit|hold)/,
+    /close the fan/,
+    /no matter how you arrange it/,
+    /There isn't any more room in/,
+    /There's no room/,
+    /to fit in the/,
+    /too heavy to go in there/,
+    /You (need to|should) unload/,
+    /You can't do that/,
+    /You just can't get/,
+    /You can't put items/,
+    /You can only take items out/,
+    /Perhaps you should be holding that first/,
+    /Containers can't be placed in/,
+    /The .* is not designed to carry anything/,
+    # You may get the next message if you've been cursed and unable to let go of items.
+    # Find a Cleric to uncurse you.
+    /Oddly, when you attempt to stash it away safely/,
+    /completely full/,
+    /That doesn't belong in there!/,
+    /exerts a steady force preventing/
+  ]
+
+  def offload_items(source, destination, noun)
+    # Pause script and warn that the script is designed to offload items
+    # from your person to somewhere else. Item loss may occur.
+    DRC.message("WARNING: This script is designed to offload items from YOU to somewhere ELSE.")
+    DRC.message("WARNING: It also has the ability to move every item from the source container.")
+    DRC.message("WARNING: Item loss may occur. Use with caution.")
+    DRC.message("WARNING: If you want to transfer items between containers you own, use ;transfer-items instead.")
+    DRC.message("\n")
+    
+    unless noun
+      DRC.message("You are about to move EVERY SINGLE ITEM from from #{source} to #{destination}.")
+      DRC.message("Unpause the script to continue. Otherwise, ;kill the script to stop.")
+      pause_script
+      DRC.message("Are you SURE you want to offload EVERY SINGLE ITEM from YOUR #{source} to #{destination}?")
+      pause_script
+    else
+      DRC.message("You are about to move all #{noun}s from #{source} to #{destination}.")
+      DRC.message("Unpause the script to continue. Otherwise, ;kill the script to stop.")
+      pause_script
+    end
+    
+    # If container is very full then LOOK may not list all of them.
+    # If you're moving a specific item, then sort those to the top
+    # to increase chances we find and move all of them in one go.
+    bput("sort #{noun} in my #{source}", "are now at the top", "What were you referring to", "Please rephrase that command", "You may only sort items in your inventory") if noun
+    DRCI.get_item_list(source, 'look')
+      .map { |full_name| full_name =~ /lot of other stuff/ ? full_name : full_name.split(' ').last }
+      .select { |item| noun ? /\b#{noun}\b/ =~ item : true }
+      .each do |item|
+        # This indicates there is more items than LOOK can show
+        # and we've reached the end of what we last saw.
+        # We need to look again to see what we can now see.
+        # Keep doing this until fill the destination or exhaust source.
+        if item =~ /lot of other stuff/
+          offload_items(source, destination, noun)
+          return
+        end
+        # Attempt to get the item from the source container.
+        if DRCI.get_item(item, source)
+          if destination == 'trash'
+            trash_item(item)
+          else
+            unless move_item?(item, source, destination)
+              DRC.message("Unable to put #{item} in #{destination}. It may not accept items, or you may need to change the preposition (e.g. 'on shelf' vs. 'in shelf')")
+              # Failed. Return item to source container.
+              DRCI.put_away_item?(item, source)
+              exit
+            end
+          end
+        else
+          DRC.message("Unable to get #{item} from #{source}.")
+          DRC.message("Your hands are full!") if (DRC.left_hand && DRC.right_hand)
+          exit
+        end
+      end
+  end
+ 
+  def move_item?(item, source, destination)
+    # If the preposition isn't specified for the destination
+    # then default to "in <container>"
+    if !destination.include?(" ")
+      command = "put #{item} in #{destination}"
+    else
+      command = "put #{item} #{destination}"
+    end
+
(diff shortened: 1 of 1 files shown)


offload-items.lic
```diff
@@ -0,0 +1,159 @@
...
+          end
+        else
+          DRC.message("Unable to get #{item} from #{source}.")
+          DRC.message("Your hands are full!") if (DRC.left_hand && DRC.right_hand)
+          exit
+        end
+      end
+  end
+ 
+  def move_item?(item, source, destination)
```
KatoakDR, line 131: ❓ Could the `move_item?` method be replaced by `DRCI.put_away_item_unsafe?(item, destination)`?

The `transfer-items` script uses `DRCI.put_away_item?` because it ensures the **my** prefix to the container to ensure you're transfering items on your person. In this case, you want to be able to put items in containers not on your person so no need to automatically add the **my** prefix, which is what the "unsafe" method variant allows.

offload-items.lic
```diff
@@ -0,0 +1,159 @@
...
+      DRC.message("You are about to move EVERY SINGLE ITEM from from #{source} to #{destination}.")
+      DRC.message("Unpause the script to continue. Otherwise, ;kill the script to stop.")
+      pause_script
+      DRC.message("Are you SURE you want to offload EVERY SINGLE ITEM from YOUR #{source} to #{destination}?")
+      pause_script
+    else
+      DRC.message("You are about to move all #{noun}s from #{source} to #{destination}.")
+      DRC.message("Unpause the script to continue. Otherwise, ;kill the script to stop.")
+      pause_script
+    end
```
KatoakDR, line 93: 💡 Suggest moving lines 75-93 into their own method, maybe something like `warn_before_offload`, and call it right before line 20. This way the warning logic isn't intermingled with the item transfer logic.

offload-items.lic
asechrest: I considered this and actually initially designed the script around this helper method but the success patterns in DRCI don't properly consider the success messages we need when offloading items from our person to outside our person. It also doesn't properly handle putting items on|under|behind because it seems to be designed around putting items away on your person.

Also doesn't appear to properly handle my idea of including a preposition before the destination container such that you can specify "in shelf" or "on shelf" or "behind shelf" or "under shelf".

I considered modifying DRCI with helper methods to handle what I need, but figured that might be above my pay grade, so I handled it fully within this script.

Still, welcome suggestions on how I might utilize helper methods in DRC/DRCI.

offload-items.lic
asechrest: Thanks you. Can you give a hint with pseudo code or otherwise how I can encapsulate this messaging as you suggest? (I'm quite new to coding and very new to Ruby.)

KatoakDR: Sure, something like this:

_Lines 75-93 get moved into a new method defined in this class:_
```ruby
def warn_before_offload(source, destination, noun)
    # Pause script and warn that the script is designed to offload items
    # from your person to somewhere else. Item loss may occur.
    DRC.message("WARNING: This script is designed to offload items from YOU to somewhere ELSE.")
    DRC.message("WARNING: It also has the ability to move every item from the source container.")
    DRC.message("WARNING: Item loss may occur. Use with caution.")
    DRC.message("WARNING: If you want to transfer items between containers you own, use ;transfer-items instead.")
    DRC.message("\n")

    unless noun
      DRC.message("You are about to move EVERY SINGLE ITEM from from #{source} to #{destination}.")
      DRC.message("Unpause the script to continue. Otherwise, ;kill the script to stop.")
      pause_script
      DRC.message("Are you SURE you want to offload EVERY SINGLE ITEM from YOUR #{source} to #{destination}?")
      pause_script
    else
      DRC.message("You are about to move all #{noun}s from #{source} to #{destination}.")
      DRC.message("Unpause the script to continue. Otherwise, ;kill the script to stop.")
      pause_script
    end
end
```

_Call the new warn method when the class is initialized:_
```ruby
  def initialize
    # ... existing logic ...
    warn_before_offload(args.source, args.destination, args.noun)
    offload_items(args.source, args.destination, args.noun)
  end
```

asechrest: Ah, duh, makes sense now that you outline it. Will probably be tomorrow but I'll edit and commit. Let me know what you think about my comments on `DRCI.put_away_item_unsafe?`. Would have liked to use that helper method but it didn't seem to do what I needed.

offload-items.lic
KatoakDR: Gotcha. What you've proposed is fine to me, and we can iterate on if/how/when to propose changes into the DRC/DRCI helper methods to accommodate the new match strings in a future PR.

> Also doesn't appear to properly handle my idea of including a preposition before the destination container such that you can specify "in shelf" or "on shelf" or "behind shelf" or "under shelf".

I see what you mean now. The `DRCI.put_item_away?(..)` method hardcodes the `in` preposition. It needs to be refactored to be like `DRCI.get_item_unsafe(..)` that handles the in/on/under/etc.

> I considered modifying DRCI with helper methods to handle what I need, but figured that might be above my pay grade, so I handled it fully within this script.

Not at all! We have code reviews and tests to help folks new the codebase :)

You're thinking about all the right things 👍

asechrest: Let me clarify a little why I feel like `DRCI.put_away_item_unsafe?` does or doesn't do what I need and necessitates me writing my own method:

1. Hard-codes putting the item IN destination container (I think?). I want to support calling the script with "on shelf" and "in cabinet" and "behind desk", etc.
2. Related to above, doesn't allow us to understand if we've passed in a preposition
3. Doesn't include the proper @@success_patterns for offloading an item OFF your person, like `/^You set/i`

vtcifer: The idea here is having a generic method that is used to limit the amount of methods you have to update down the line.  In that vein, my thoughts are adding an (optional) third  argument to the method 'preposition', which defaults to 'in':
- Should be OK as a third optional argument, as it would only make sense to include if you've also included a container.
- This should be non-breaking to anything already using it 
- Allows flexibility down the line where needed (like this)

```ruby
  def put_away_item_unsafe?(item, container = nil, preposition = 'in')
    command = "put #{item} #{preposition} #{container}" if container
    command = "stow #{item}" unless container
    result = DRC.bput(command, @@container_is_closed_patterns, @@put_away_item_success_patterns, @@put_away_item_failure_patterns, @@put_away_item_retry_patterns)
    case result
    when *@@container_is_closed_patterns
      return false unless open_container?(container)
      return put_away_item_unsafe?(item, container)
    when *@@put_away_item_retry_patterns
      return put_away_item_unsafe?(item, container)
    when *@@put_away_item_success_patterns
      return true
    when *@@put_away_item_failure_patterns
      return false
    else
      return false
    end
  end
```

Some additional testing error scenarios might need to be taken into account (wrong preposition, if you can't manipulate the container to open it).

asechrest: I like the idea of updating the DRCI method. The success patterns would need to be updated as well because DRCI currently doesn't account for successfully putting things elsewhere - e.g. "You set the cloak on the shelves". Do you have any concern intermingling success messaging for keeping items on your person (safe) with success messaging for offloading items off your person (unsafe)?

I'm just trying to be real thoughtful about item loss, is all.

vtcifer: I don't think there should be any issues.  The patterns are just there to trap game output (and not end up with a 15 sec delay on command).  You've already executed the command itself, just trapping output.  

Keep in mind updates to common-* should be non-breaking to other scripts, and need to be rolled out in advance of regular scripts that depend on them, so you might need to break this into 2 PRs.

asechrest: Ok, you and Katoak have convinced me to do some work on DRCI. So my first PR was to combat-trainer of all things, and now my second PR will be to common-items. Wheee.

KatoakDR: > The idea here is having a generic method that is used to limit the amount of methods you have to update down the line if things from the game change. In that vein, my thoughts are adding an (optional) third argument to the method 'preposition', which defaults to 'in'

+1

asechrest: Handled via #5237

asechrest:
Additional cleanup and consolidation of this script. Note now that this relies on the update to `;common-items` via #5237 and thus will need to wait to merge for that one to be reviewed, merged, and settled.

offload-items.lic
asechrest: Broke the warning messaging out per @KatoakDR .

rpherbig:
> Additional cleanup and consolidation of this script. Note now that this relies on the update to `;common-items` via #5237 and thus will need to wait to merge for that one to be reviewed, merged, and settled.

The real risk to pushing common* changes is that a script will break if common* hasn't been reloaded. Because this is a _new_ script, we should be able to push it immediately. Nothing should break for _existing_ scripts; the worst case is that someone runs this new script and it fails. They reload common* and now it works.

@KatoakDR how's the code change look to you?

asechrest:
> > Additional cleanup and consolidation of this script. Note now that this relies on the update to `;common-items` via #5237 and thus will need to wait to merge for that one to be reviewed, merged, and settled.
> 
> The real risk to pushing common* changes is that a script will break if common* hasn't been reloaded. Because this is a _new_ script, we should be able to push it immediately. Nothing should break for _existing_ scripts; the worst case is that someone runs this new script and it fails. They reload common* and now it works.
> 
> @KatoakDR how's the code change look to you?

That makes sense.

asechrest:
I will say one of my main dislikes of this script I built is that it can take as many as four arguments (two optional), and they're in an order that is not especially intuitive. For example:

`;offload-items <source> <destination> <preposition> <noun>`

If we want to offload all cloaks from our backpack onto a shelf, the input is:

`;offload-items backpack shelf on cloak`

The order that the arguments must be input is not real intuitive. I'd much prefer:

`;offload-items cloak backpack on shelf`

But I could not figure out how to code it given that, for example, in that order, the first and third argument are a set of known possible inputs but are actually optional, and other inputs are required inputs but could be any possible input . I didn't know how to code it to handle this. If that is even possible.

The script functions as-is according to my testing, I just wish I could have figured out coding it with a more intuitive argument order.

KatoakDR:
> I will say one of my main dislikes of this script I built is that it can take as many as four arguments (two optional), and they're in an order that is not especially intuitive.

@asechrest I know that the `transfer-items` scripts takes two required arguments (source, destination) then one optional (noun), but I don't think this new script has to be bound by that format if it's unintuitive.

What are your thoughts on making it take 3 required arguments (source, preposition, destination) and one optional (noun)?

Usage would be like:
```
;offload-items backpack in vault

;offload-items backpack on table arrow
```

That also has the benefit of ensuring the player is explicit about where the item(s) are going (in/on/under/behind), especially since the items are moving _off_ the person. You have the warning message, too, but I'm looking for a positive if you did make the preposition the 2nd argument and required.

Other than that, script looks good to me. I can also test it out later tonight. Thanks

asechrest:
> What are your thoughts on making it take 3 required arguments (source, preposition, destination) and one optional (noun)?
> 
> Usage would be like:
> 
> ```
> ;offload-items backpack in vault
> 
> ;offload-items backpack on table arrow
> ```

I like this order of arguments a lot better than how I currently have it coded. And, by making the second argument (preposition) required, I believe I can properly write the code. Great idea. 

I would love to know, though, if something like this is possible:

`;offload-items <required arg> <optional arg of known format> <required arg>`

So that it would accept either of the following:

```
;offload-items backpack on shelf # Offloads all items from backpack, using preposition ON, to shelf
;offload-items backpack shelf # Offloads all items from backpack, recognizes that the user has passed no preposition, and recognizes SHELF as the "third" arg, rather than thinking it's the preposition.
```

I think my overarching coding question is - can we accept a 2nd arg IF it exists, and still allow a "third" arg without the script thinking it's the 2nd arg?

In words this would be: accept a first argument whatever word it is and assign as source. Accept a second argument IFF it is of the format in|on|under|behind and assign that to preposition. Accept a final argument and assign that to destination without the script thinking it's the SECOND arg if the user has not specified the optional preposition.

KatoakDR:
> can we accept a 2nd arg IF it exists, and still allow a "third" arg without the script thinking it's the 2nd arg?

Come to think of it, yes, I think that might be possible by defining more argument definitions in the `arg_definitions` array (note that it's an array of _arrays_). Each nested array is a valid combination of arguments for just this purpose.

I think you could do something like this, and note that I've removed the `optional` property from the arguments so that Lich does an exact match on number of arguments provided for how it will interpret which variable they represent:
```ruby
    arg_definitions = [
      # if exactly four arguments: <source> <preposition> <destination> <noun>
      [
        { name: 'source', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Source container' },
        { name: 'preposition', options: %w[in on under behind], variable: true, description: 'Preposition to use with the destination (e.g. "ON shelf"' },
        { name: 'destination', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Destination container' }
        { name: 'noun', regex: /^[A-z\.\s\-]+$/i, variable: true, description: 'If specified, only items with this noun will be transferred.'}        
      ],
      # if exactly two arguments: <source> <destination>
      [
        { name: 'source', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Source container' },
        { name: 'destination', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Destination container' }
      ],
      # if exactly three arguments: <source> <preposition> <destination>
      [
        { name: 'source', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Source container' },
        { name: 'preposition', options: %w[in on under behind], variable: true, description: 'Preposition to use with the destination (e.g. "ON shelf"' },
        { name: 'destination', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Destination container' }
      ]
    ]
```

offload-items.lic
```diff
@@ -0,0 +1,94 @@
...
+    # Pause script and warn that it is designed to offload items
+    # from your person to somewhere else. Item loss may occur.
+    DRC.message("WARNING: This script is designed to offload items from YOU to somewhere ELSE.")
+    DRC.message("WARNING: It also has the ability to move every item from the source container.")
+    DRC.message("WARNING: Item loss may occur. Use with caution.")
+    DRC.message("WARNING: If you want to transfer items between containers you own, use ;transfer-items instead.")
+    DRC.message("\n")
+    
+    unless noun
+      DRC.message("You are about to move EVERY SINGLE ITEM from from #{source} to #{destination}.")
```
KatoakDR, line 79: 💅  duplicate 'from'
```suggestion
      DRC.message("You are about to move EVERY SINGLE ITEM from #{source} to #{destination}.")
```

asechrest:
> > can we accept a 2nd arg IF it exists, and still allow a "third" arg without the script thinking it's the 2nd arg?
> 
> Come to think of it, yes, I think that might be possible by defining more argument definitions in the `arg_definitions` array (note that it's an array of _arrays_). Each nested array is a valid combination of arguments for just this purpose.
>

Mind blown. Thanks so much for this. I'm working backward and learning from existing scripts. Hadn't seen anything like this and just couldn't wrap my mind around how we could handle the situation I was asking.

asechrest:
Implemented @KatoakDR 's suggestion for 3 required script args (reordered) and an optional noun. Reordered method args to match. Additionally, a little more detailed messaging to the script user in the warn method.

KatoakDR:
> Mind blown. Thanks so much for this. I'm working backward and learning from existing scripts. Hadn't seen anything like this and just couldn't wrap my mind around how we could handle the situation I was asking.

You're very welcome! And you're learning Lich scripts the same way I, and many others, did. Just copy/paste/experiment. I don't think the multiple sets of arg definitions is common, took me a few minutes to find a script where I'd seen it before. Glad it helps!

rcuhljr:
I swear I documented parse_args somewhere when I wrote it but I'm not finding it in my searching, my bad. Might be handy for someone to add some documentation for it in a developer guides section.

KatoakDR (approved):
I did some tests in a vault to offload items from a container on my person **IN** the vault, and again from my container to **ON** a hook in the vault. Worked nicely for me, thanks 🚀

asechrest:
> I did some tests in a vault to offload items from a container on my person **IN** the vault, and again from my container to **ON** a hook in the vault. Worked nicely for me, thanks 🚀

Gotta' be the longest comment string for a simple script that's ever existed. But great fun, and I learned a lot. Thanks to you and the others who helped. I'll wait for the next opportunity to contribute to present itself.

rpherbig:
Thank you for all your hard work on this @asechrest !

PR #5237, opened 2021-11-18
Update common items

asechrest:
Update method `put_away_item_unsafe?()` to avoid hard-coding the "in" preposition and allows to pass in a preposition of another type (e.g. "ON shelf"). Still defaults the preposition to "in" if no preposition is specified to avoid breaking scripts using this method as it exists. Also updates the success messaging to recognize when we've successfully put an item from our person to off our person.

This update will help support PR #5232 (in progress) which is a script to handle offloading items from our person to outside our person.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -804,7 +804,7 @@ module DRCI
   # Puts away an item, optionally into a specific container, optionally
   # using a container preposition (e.g. "ON shelf")
   # If no container specified then uses the default stow location.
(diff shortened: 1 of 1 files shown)


KatoakDR (approved):
Looks good to me, thanks

asechrest:
Forgot to mention I also updated the failure patterns to recognize failure for `/^You can't put that in|on|under|etc. there/`.

PR #5239, opened 2021-11-20
Hunting-Buddy.lic allow the threshold for stop_on to be specified

Valherun:
replaced huntig-buddy.lic's default of 32 when checking stop_on skills to now use stop_on_threshold to allow stopping on lower mindstates. 

Added to base.yaml with a default of 32 as specified in the original hunting-buddy.

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -20,6 +20,7 @@ class HuntingBuddy
     @escort_zones = data.escort_zones
     @hunting_zones = data.hunting_zones
     @stop_on_low_threshold = @settings.stop_on_low_threshold
+    @stop_on_threshold = @settings.stop_on_threshold || 32
     @magic_exp_threshold = @settings.magic_exp_training_max_threshold
     @magic_skills = ['Arcane Magic', 'Holy Magic', 'Life Magic', 'Elemental Magic',\
                      'Lunar Magic', 'Attunement', 'Arcana', 'Targeted Magic', 'Inner Fire',\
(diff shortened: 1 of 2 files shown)


hunting-buddy.lic
```diff
@@ -20,6 +20,7 @@ class HuntingBuddy
     @escort_zones = data.escort_zones
     @hunting_zones = data.hunting_zones
     @stop_on_low_threshold = @settings.stop_on_low_threshold
+    @stop_on_threshold = @settings.stop_on_threshold || 32
```
rpherbig, line 23: If it's in `base.yaml`, it doesn't have to be specified here.

profiles/base.yaml
```diff
@@ -402,6 +402,8 @@ hunting_room_strict_mana: false
 hunting_room_max_searches: 30
 hunting_file_list:
 stop_on_low_threshold: 2
+# sets a default stop_on threshold for hunting-buddy.lic 
+stop_on_threshold: 32
```
rpherbig, line 406: What do you think about naming this `stop_on_high_threshold` to make it clearer how this differs from the setting above it?

hunting-buddy.lic
Valherun: I can remove it, figured it would be good as a failsafe incase someone on dev mode pulls the updated hunting-buddy.

profiles/base.yaml
Valherun: sounds good, will make the changes.

Dartellum (approved):

Issue #5242, opened 2021-11-20
Tactics manuevers fail if not standing, missing 'You should stand up first'


Chewmatca:
"You should stand up first" is not included in combat-trainer.lic
can search for instances of: 'You must be standing'
Lines 2464 and Line 2742 look to need to have this added, or your dude just stands there gettin whacked while prone.

Others also advised it could be same in issue for combat manuevers:

Zin — Today at 11:11 AM
re: the standing thing, the same issue exists for combat manuevers
If you fall before the command goes through and get a failure for not standing

Chewmatca:
Full log, expect the script to stand when it did not, due to not catching "You must be standing" after being knocked to the ground.

An ice archon raises its massive icy foot and brings it downward!
The massive impact shakes the ground!
You fall over as the world shakes beneath you!
[You're mighty, very badly balanced with opponent overwhelming you.]


The ice archon moves into a position to parry.
P> 
An ice archon backs away from combat, sending a shower of ice shards from its body as it does so.
An ice archon moves ponderously northwest, dragging its feet and leaving huge snowy scratches behind it.

P> 
[combat-trainer]>analyze

True Bard Natarian just arrived.

P> 
True Bard Natarian goes southeast.

P> 
You should stand up first.
P> 
P> 
Jobber Sall just arrived.

P> 

The ice archon begins to advance on Sall.
P> 
Jobber Sall strolls southeast.

P> 
Your silver glow flares.
The air around you solidifies into a blinding yellow luminescence, collapsing under the force of the attack in a silent cascade of glassy fragments that vanish before reaching the ground.

Your silver glow flares.

You feel fully attuned to the mana streams again.
P> 
Your silver glow flares.

Your silver glow flares.

Your silver glow flares.


| ***STATUS*** stopping due to bleeding

An ice archon backs away from combat, sending a shower of ice shards from its body as it does so.
An ice archon moves ponderously northeast, dragging its feet and leaving huge snowy scratches behind it.

!SP> 
Your silver glow flares.

!SP> 
!P> 
Your silver glow flares.

!P> 
Your silver glow flares.

!P> 
[combat-trainer: *** No match was found after 15 seconds, dumping info]

[combat-trainer: messages seen length: 69]

[combat-trainer: message: [You're very badly hurt, mighty, very badly balanced with opponent overwhelming you.]]

[...]

PR #5245, opened 2021-11-21
[script][combat-trainer][update] Fix multiple failures when prone

asechrest:
Addressing issue #5242. Add a proper bput failure message for if you're knocked prone when trying Analyze. This failure message may need to be handled elsewhere. For now, handling the specific error case.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4573,6 +4573,11 @@ class GameState
       fput('engage')
     when 'You fail to find any'
       perform_analyze?(combo_type, expertise_requirement)
+    when 'You should stand up first'
+      # We're somehow prone. Attempt to stand up and retry the maneuver.
+      DRC.fix_standing
+      waitrt?
+      perform_analyze?(combo_type, expertise_requirement)
     when 'Analyze what?'
       case bput('face next', 'There is nothing else', 'You turn', 'What are you trying')
       when 'You turn'
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4550,7 +4550,7 @@ class GameState
     return false if combo_type != 'flame' && !Flags["ct-#{combo_type}-ready"] && DRStats.barbarian?
     return false if DRSkill.getrank('Expertise') < expertise_requirement
 
-    result = bput("analyze #{combo_type}", 'cannot repeat', 'Analyze what\?', 'by landing an? .*', 'You need to hold', 'You must be closer', 'You fail to find any','What are you trying to attack\?')
+    result = bput("analyze #{combo_type}", 'cannot repeat', 'Analyze what\?', 'by landing an? .*', 'You need to hold', 'You must be closer', 'You fail to find any', 'What are you trying to attack\?', 'You should stand up first')
     waitrt?
 
     case result
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2461,7 +2461,7 @@ class TrainerProcess
       waitrt?
       fix_standing
     when /^Tactics$/i
-      bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty?
+      bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack', 'You should stand up first') unless game_state.npcs.empty?
     when /^Analyze$/i
       analyze(game_state, 0)
     when /^Hunt$/i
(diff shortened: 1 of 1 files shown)


asechrest:
Note: all this fix does is avoid the 15 second match failure and breaks us fully out of the analyze code block. The script then moves on to the next thing, which includes a call to `fix_standing` elsewhere.

I considered handling the error within `def analyze()` using a call to `DRC.fix_standing()` and the calling analyze() again, but we'd want to iterate the failure count, I think, to avoid an infinite loop. For now just opted for the simple fix.

asechrest:
I added a commit to handle prone in the Barbarian analyze section. **I am unable to test this as I don't have a Barb.** Looks to me like it should work fine, but someone may want to test.

Dartellum:
> Note: all this fix does is avoid the 15 second match failure and breaks us fully out of the analyze code block. The script then moves on to the next thing, which includes a call to `fix_standing` elsewhere.
> 
> I considered handling the error within `def analyze()` using a call to `DRC.fix_standing()` and the calling `analyze()` again, but we'd want to iterate the failure count, I think, to avoid an infinite loop. For now just opted for the simple fix.

Fail count should not change since no attempt to analyze was given. Do we count that it could not because prone as a failure to analyze? Need to test if it does enter a loop. I suppose it depends on the mob and if they can knock you down consistently, like rock trolls casting tremor.

asechrest:
Most recent commit handles being prone when trying a Maneuver.

```
>lie
You lie down.
>
A musk hog hisses viciously and lunges at you!
* A musk hog claws, swiping with its hooves at you.  You block solidly with a dented pot lid dappled with scorch marks.  
[You're very badly balanced with opponent in superior position.]
>
[combat-trainer: ***Ready to use charged maneuver: crash***]
[combat-trainer]>maneuver crash
You must be standing to perform that maneuver.
>
[combat-trainer]>stand
You stand back up.
>
[combat-trainer]>maneuver crash
Bantree just arrived.
>
Taking a full step back, you plant your feet and ready a mighty attack.
Dashing in, you attempt to crash your sledgehammer through a musk hog!

A musk hog deftly sidesteps your pine-handled sledgehammer.

[You're badly balanced with opponent in strong position.]
Roundtime: 4 sec.
```

vtcifer:
Have we thought about adding the "stand up fix" to bput instead, as another default "try again" match?

Dartellum:
> Have we thought about adding the "stand up fix" to bput instead, as another default "try again" match?

I put that in my suggestions of code.

asechrest:
@Dartellum I don't see your code suggestion?

@vtcifer Do you mean revising the bput method itself?

asechrest:
I believe this now recovers from a prone position for both Tactics, Analyze, Charged Maneuvers, and Barbarian Analyze. Still unable to test the barb stuff, but the other stuff is tested on my character. Welcome further suggestions.

Dartellum:
> @Dartellum I don't see your code suggestion?
> 
> 
> 
> @vtcifer Do you mean revising the bput method itself?

Look under the Commits tab.

asechrest:
> > @Dartellum I don't see your code suggestion?
> > @vtcifer Do you mean revising the bput method itself?
> 
> Look under the Commits tab.

Hmmm. Am I missing it?

image

Dartellum:
On the iPhone app it was under conversations. On the Web page, it is under Conversations. I started a review for each one. I also found the match should go here as well:

```
when /^Tactics$/i
      bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack', 'You should stand up first') unless game_state.npcs.empty?
```

image
image
image

vtcifer:
> On the iPhone app it was under conversations. On the Web page, it is under Conversations. I started a review for each one. I also found the match should go here as well:

I don't think your review is published and viewable by anyone else yet (at least it's not showing up for me).

vtcifer:
> @vtcifer Do you mean revising the bput method itself?

Indeed, I mean updating the DRC.bput method to natively handle the failure condition via trap, stand, retry, similar to how it handles RT/typeahead related issues.

```ruby
# [...]
      case response
      when /(?:\.\.\.wait |Wait |\.\.\. wait )([0-9]+)/
        unless ignore_rt
          pause(Regexp.last_match(1).to_i - 0.9)
          waitrt?
          put message
          timer = Time.now
        end
        next
      when /Sorry, you may only type ahead/
        pause 1
        put message
        timer = Time.now
        next
      when /^You should stand up first/
        fix_standing
        put message
        next
# [...]
```

asechrest:
> > @vtcifer Do you mean revising the bput method itself?
> 
> Indeed, I mean updating the DRC.bput method to natively handle the failure condition via trap, stand, retry, similar to how it handles RT/typeahead related issues.
> 
> ```ruby
> 
> ```

You know, my first though was, "is that really the purview of `bput`"? But I see it already handles a coupe of other cases that are similar. I like the idea of it being centralized in DRC.

What sort of informal rule do we have about whether it should be handled in DRC.bput vs. when it should be handled in the script? I'm thinking something like: "if it's a failure case where we would nearly always want to fix the issue and retry the exact same bput, it can be handled in DRC.bput"?

asechrest:
Handling this within DRC.bput.

PR #5247, opened 2021-11-23
Update common.lic

asechrest:
Addressing original issue #5242 , in which parts of combat-trainer fail when your character is unexpectedly prone.

We were originally addressing this directly within combat-trainer via #5245 which is now closed without merging.

After additional discussion we've decided this is best handled within DRC.bput itself, which already handles some limited conditions wherein we have a common failure condition and want to simply fix that condition and re-input the same exact bput without any complex handling.

Thus, this PR edits bput to handle standing you up when you are prone. Separately, I'll do another PR to combat-trainer to remove the redundant failure output for standing since it'll now be handled in DRC.bput.

@vtcifer @Dartellum @KatoakDR

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -155,7 +155,7 @@ module DRC
         put message
         timer = Time.now
         next
-      when /^You must be standing/, /^You should stand up first/, /^You can't do that while (sitting|kneeling|lying)/
+      when /^You must be standing/, /^You should stand up first/, /You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /You should be sitting up/
         fix_standing
         waitrt?
         put message
--- a/common.lic
+++ b/common.lic
@@ -159,7 +159,7 @@ module DRC
         put message
         timer = Time.now
         next
-      when /^You must be standing/, /^You should stand up first/, /You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /You should be sitting up/, /You really should be standing to play/
+      when /^You must be standing/, /^You should stand up first/, /You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /^You should be sitting up/, /^You really should be standing to play/
         fix_standing
         waitrt?
         put message
--- a/common.lic
+++ b/common.lic
@@ -155,7 +155,7 @@ module DRC
         put message
         timer = Time.now
         next
-      when /^You must be standing/, /^You should stand up first/
+      when /^You must be standing/, /^You should stand up first/, /^You can't do that while (sitting|kneeling|lying)/
         fix_standing
         waitrt?
         put message
--- a/common.lic
+++ b/common.lic
@@ -159,7 +159,7 @@ module DRC
         put message
         timer = Time.now
         next
-      when /^You must be standing/, /^You should stand up first/, /You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /^You should be sitting up/, /^You really should be standing to play/
+      when /^You must be standing/, /^You should stand up first/, /^You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /^You should be sitting up/, /^You really should be standing to play/
         fix_standing
         waitrt?
         put message
--- a/common.lic
+++ b/common.lic
@@ -155,6 +155,12 @@ module DRC
         put message
         timer = Time.now
         next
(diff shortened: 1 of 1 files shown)


common.lic
```diff
@@ -155,6 +155,12 @@ module DRC
         put message
         timer = Time.now
         next
+      when /^You must be standing/, /^You should stand up first/
```
KatoakDR, line 158: These are other positions you could find yourself in that would prevent movement that need you to stand back up
```suggestion
      when /^You must be standing/, /^You should stand up first/, /^You can't do that while (sitting|kneeling|lying)/
```

KatoakDR (approved):

asechrest:
Some additional testing indicates the handling of standing within DRC.bput preempts duplicate handling of the same regex within other scripts, for example combat-trainer. Because of this, I'm inclined to do nothing to those other scripts initially, and let this PR stand alone to handle the prone failures that exist at least within combat-trainer, and perhaps elsewhere.

rpherbig:
I like the simplicity of this change. I like that it'll help out a number of scripts, not just combat-trainer (search the repo for `fix_standing`).

@rcuhljr do you see any problems with this approach?

common.lic
```diff
@@ -155,6 +155,12 @@ module DRC
         put message
         timer = Time.now
         next
+      when /^You must be standing/, /^You should stand up first/, /^You can't do that while (sitting|kneeling|lying)/
```
rpherbig, line 158: If I'm reading this code right 
https://github.com/rpherbig/dr-scripts/blob/7953d9c04f406c6e8da141c97c4e898de9b7924a/coordinator.lic#L459
then there is one more case: `'You should be sitting up'`

rpherbig: If I'm reading this code right 
https://github.com/rpherbig/dr-scripts/blob/45ae6be4d505615899eadfd30b54ddfb339b8fe1/stabbity.lic#L140
then there is one more case: `You'll need to stand up first`

rpherbig: If I'm reading this code right 
https://github.com/rpherbig/dr-scripts/blob/acc89c9a453fcb5afe06e4fa36e31e8b7b8bd4dd/boggle_blast.lic#L97
then there is one more case: `You really should be standing to play`

asechrest (commented):
Good call on these. Will get them added.

common.lic
Dartellum: `Boggle_blast.lic` is a dead script at this point. Should we remove all the old HE games, the ones from The Spider?

vtcifer: Admittedly git will keep the history, but we may want to move them into an archive folder, just in case.

asechrest:
Add the suggested match phrases. Re-tested locally with analyze and charge maneuver.

asechrest:
Sorry, accidentally closed this. Meant to just resolve the comments. 

I've added the suggested match phrases and re-tested. I think this may be ready if all agree.

rpherbig:

@KatoakDR @Chewmatca @Dartellum @vtcifer any other thoughts/questions before I merge this?

common.lic
```diff
@@ -155,6 +159,12 @@ module DRC
         put message
         timer = Time.now
         next
+      when /^You must be standing/, /^You should stand up first/, /You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /You should be sitting up/, /You really should be standing to play/
```
KatoakDR, line 162: Please add the `^` prefix character the the last two new regex strings, thanks
```suggestion
      when /^You must be standing/, /^You should stand up first/, /You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /^You should be sitting up/, /^You really should be standing to play/
```

KatoakDR (approved):

common.lic
```diff
@@ -155,6 +159,12 @@ module DRC
         put message
         timer = Time.now
         next
+      when /^You must be standing/, /^You should stand up first/, /You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /^You should be sitting up/, /^You really should be standing to play/
```
Dartellum, line 162: Miss `/You'll need to stand up first/` => `/^You'll need to stand up first/`.

asechrest: Blech, I need to pay better attention.

asechrest:
Thanks, all, for the comments, support, and assistance!

PR #5249, opened 2021-11-25
[script] [appraisal] Skip items that aren't found

KatoakDR:
### Background
* When using `appraisal_training: [gear]` then the script iterates all items in your `gear_set:`
* If you have items listed that are not currently on your person then the script tries (and fails) to appraise something that you didn't get.

### Changes
* If don't get item then skip it and move on to the next item.

### DIFF
--- a/appraisal.lic
+++ b/appraisal.lic
@@ -96,11 +96,12 @@ class Appraisal
   end
 
   def appraise_item?(item)
-    @equipment_manager.get_item?(item)
-    bput("appraise #{item.short_name} quick", 'Roundtime')
-    waitrt?
-    pause 1
-    @equipment_manager.empty_hands
+    if @equipment_manager.get_item?(item)
+      bput("appraise #{item.short_name} quick", 'Roundtime')
+      waitrt?
+      pause 1
+      @equipment_manager.empty_hands
+    end
     DRSkill.getxp('Appraisal') < 30
   end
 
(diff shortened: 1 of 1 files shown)


appraisal.lic
```diff
@@ -96,11 +96,12 @@ class Appraisal
...
-    @equipment_manager.get_item?(item)
-    bput("appraise #{item.short_name} quick", 'Roundtime')
-    waitrt?
-    pause 1
-    @equipment_manager.empty_hands
+    if @equipment_manager.get_item?(item)
+      bput("appraise #{item.short_name} quick", 'Roundtime')
+      waitrt?
+      pause 1
+      @equipment_manager.empty_hands
```
KatoakDR, line 103: These 4 lines are now indented within the IF statement, no other change to the logic

rpherbig: Thank you for pointing that out

appraisal.lic
```diff
@@ -96,11 +96,12 @@ class Appraisal
...
+    end
```
rpherbig, line 104: I wouldn't mind seeing an else that reports the problem to the user, but it's not a big deal either.

KatoakDR: Agree, thankfully the `equipment_manager.get_item?` method already logs when it doesn't get the item: https://github.com/rpherbig/dr-scripts/blob/master/equipmanager.lic#L230

The appraise script just needed to be smart enough to skip trying to appraise the non-existant item.

Thanks!

rpherbig: Good point!

PR #5250, opened 2021-11-25
[scripts] [common-items] open eddy THEN look in portal THEN get item

KatoakDR:
### Background
* If you've closed your eddy for some reason, the recovery logic we recently put in to LOOK in the container will still fail because it doesn't also check that the container is open.
* When the eddy is closed, trying to get an item from your portal doesn't give the typical "But that's closed" kind of message, instead you get a "What were you referring to?", thus we need special handling in the `get_item_from_eddy_portal?` method.

### Changes
* Ensure the **eddy** is open then look in the **portal**

### Tests
* See `test/test_common_items.rb`
```
--- Lich: exec1 active.

[exec1]>get demonbone.mace from my watery portal

What were you referring to?

[exec1]>open my eddy

You open your swirling eddy.

[exec1]>look in my watery portal

In the swirling eddy you see ....
> 
[exec1]>get demonbone.mace from watery portal

You get a jagged demonbone mace with a wrought iron haft from inside a swirling eddy of incandescent light bound by a gold-striated coralite frame.
```


fetch.sh
```diff
@@ -1,46 +0,0 @@
-#!/bin/sh
```
KatoakDR: This is now part of my global .gitignore

PR #5254, opened 2021-11-26
[script] [sell-loot] add hometown argument

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5256

### Background
* Feature requested by [Tirost](https://discord.com/channels/745675889622384681/912127186419482664/913661837462089788)

> I'm requesting being able to use `;sell-loot crossing` or `;sell-loot hib` and have the script take me to that hometown's gem appraiser, tanner and bank. My hope is to be able to use `sell-loot` in locations other than the hometown I use for training to sell off invasion spoils, money of a denomination I'd rather not pay exchange fees on or that's been acquired in an RP event, or return money after shopping or trading.

### Changes
* Add `hometown` as a new script argument to override that value from character's YAML

### Example Usage
_Original two options, and these still exist_
```
;sell-loot
;sell-loot 10 silver
```
_New options allow you to specify the hometown to use_
```
;sell-loot riverhaven
;sell-loot shard 5 gold
```

PR #5256, opened 2021-11-26
[script] [common] method to lookup town canonical name by colloquialism

KatoakDR:
### Background
* Supports https://github.com/rpherbig/dr-scripts/pull/5254
* Answers [Gildaren's question](https://discord.com/channels/745675889622384681/912127186419482664/913670178309558302), _"How do you handle if someone calls `sell-loot` with a hometown abbreviation that doesn't exactly match the hometown in `base-town.yaml`? So for example, `sell-loot lang`?"_
* [Worksaround](https://discord.com/channels/745675889622384681/912127186419482664/913689331791712327) that parsed script arguments are lowercase but the `data/base-town.yaml` town keys are case-sensitive.

### Changes
* Add a new map of canonical town names (i.e. as listed in `data/base-town.yaml`) and a regular expression of the official or colloquial name that matches.
* Add new method `DRC.get_town_name(text)`

### returns nil if does not match any regex
```
> ,e echo DRC.get_town_name('foo')
[exec1: ]
```

### returns canonical name if matches regex
```
> ,e echo DRC.get_town_name('haven')
[exec1: Riverhaven]

> ,e echo DRC.get_town_name('Riverhaven')
[exec1: Riverhaven]
```


rpherbig:
Will this break `burgle.lic` if merged? https://github.com/rpherbig/dr-scripts/blob/5e1a7615464e1f6cf9d7b23eb298788531554f59/burgle.lic#L24

common.lic
```diff
@@ -382,6 +390,19 @@ module DRC
...
+  # Utility to help identify the canonical town name based on arbitrary text.
+  # For example, "Theren" for "Therenborough" and "Haven" for "Riverhaven".
+  # It also handles missing apostrophes and the occasional space between names
+  # like "merkresh" or "Mer'Kresh" or "ainghazal" or "Ain Ghazal".
+  # Returns nil if unable to find a match.
+  def get_town_name(text)
+    $HOMETOWN_REGEX
+      .select { |town, regex| regex =~ text }
+      .keys
+      .first
```
rpherbig, line 403: I'm wondering about this `.first` call. Is it a problem if `text` matches more than 1 town?

Should we notify the user we're choosing the first one but they should be more explicit?

KatoakDR: Good question. The regexes _should not_ (in my opinion) allow for ambiguous results. There should be exactly 0 or 1 town name found.

But, the day that a regex were to be added/updated that was too loose could cause multiples to be found (as you point out) and we should protect the user from unexpected results.

At the moment I'm leaning towards the script should error that prints the results and asks the user to use a more specific town value and to contact the devs/raise an issue about the ambiguity.

common.lic
```diff
@@ -382,6 +390,22 @@ module DRC
...
+  # Returns nil if unable to find a match.
+  def get_town_name(text)
+    towns = $HOMETOWN_REGEX.select { |town, regex| regex =~ text }.keys
+    if towns.length > 1
+      DRC.message("Found multiple towns that match '#{text}': #{towns}")
+      DRC.message("Using first town that matched: #{towns.first}")
+      DRC.message("To avoid ambiguity, please use the town's full name: https://elanthipedia.play.net/Category:Cities")
+    end
+    towns.first
+  end
```
KatoakDR, line 407: In response to https://github.com/rpherbig/dr-scripts/pull/5256#discussion_r758321544

When more than one town is found given the search text then a warning is given with advice to use the town's full name, then returns the first match.

_The following test had two towns that would match 'cross'_
```
,e echo DRC.get_town_name('cross')

--- Lich: exec1 active.

| Found multiple towns that match 'cross': ["Crossing", "Crossroads"]
| Using first town that matched: Crossing
| To avoid ambiguity, please use the town's full name: https://elanthipedia.play.net/Category:Cities

[exec1: Crossing]

--- Lich: exec1 has exited.
```

PR #5259, opened 2021-11-29
Create smart tipping script

asechrest:
From the notes/comments on the script:

    > # Background: I've always been annoyed by the need to specify tips in coppers.
    > # DR's required syntax leaves you counting zeroes if you want to tip someone.
    > # So a tip of 10 platinum kronars requires you to tip 100000 kronars. How many
    > # zeroes is that? I lost count already. This simple script allows you to tip 
    > # someone, in a similar syntax, but use the denomination to do so (e.g. "platinum").
    > # <;tip gildaren 10 plat kron> now works. If you don't specify a currency,
    > # this script also chooses the currency of the greatest coin you have on hand.

In other words, we all think of DR money in denominations. Primarily plats, but perhaps gold and silver. So if we want to tip someone, we're likely thinking in plats or other denominations. Empath healed you and you want to tip 2 plats? It'd be nice to `tip <person> 2 plat kronars`. This script allows us to do that, in a very similar syntax to the current DR tipping mechanism, such that we don't have to remember how to call the script. Just call it similar to how we'd tip anyway. Instead of <tip Gildaren 10000 kronars>, we can now input <;tip Gildaren 1 plat kron>, and it'll function fine.

It also allows omitting of the currency, in which case the script will choose your current max currency on hand from which to tip.

This is my second from-scratch script. I wrote it primarily for my own edification, but thought I'd offer it as a PR. If it's not worthy of being considered in the main script list for users, that's fine, and I won't be upset.

As always, I welcome comments and, especially, suggestions on better or more efficient coding.

Thanks.

### DIFF
--- a/tip.lic
+++ b/tip.lic
@@ -1,0 +1,99 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#tip
+=end
+
+# Background: I've always been annoyed by the need to specify tips in coppers.
+# DR's required syntax leaves you counting zeroes if you want to tip someone.
+# So a tip of 10 platinum kronars requires you to tip 100000 kronars. How many
+# zeroes is that? I lost count already. This simple script allows you to tip 
+# someone, in a similar syntax, but use the denomination to do so (e.g. "platinum").
+# <;tip gildaren 10 plat kron> now works. If you don't specify a currency,
+# this script also chooses the currency of the greatest coin you have on hand.
+
+custom_require.call %w[common common-money]
+
+class Tip
+
+  # include DRC for the message function
+  # include DRCM for use of converting amounts to copper
+  
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'recipient', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'The recipent of your tip. Abbreviations ok.' },
+        { name: 'amount', regex: /^\d+$/i, variable: true, description: 'Numeric amount to tip, no commas (e.g. the number "100" in 100 kronars)' },
+        { name: 'denomination', regex: /(\bp(l|la|lat|lati|latin|latinu|latinum)?\b)|(\bg(o|ol|old)?\b)|(\bs(i|il|ilv|ilve|ilver)?\b)|(\bb(r|ro|ron|ronz|ronze)?\b)|(\bc(o|op|opp|oppe|opper)?\b)/i, variable: true, description: 'The denomination. Abbrevations ok. (e.g. plat, gold, silv)' },
+        { name: 'currency', regex: /(\bk(r|ro|ron|rona|ronar|ronars)?\b)|(l(i|ir|iru|irum|irums)?\b)|(\bd(o|ok|oko|okor|okora|okoras)?\b)/i, optional: true, variable: true, description: 'Optional currency. Abbreviations ok: kron, lirums, dokor' }         
+      ]
+    ]
+    args = parse_args(arg_definitions)  
+    gratuity(args.recipient, args.amount, args.denomination, args.currency)
+
+  end
+
+  def gratuity(recipient, amount, denomination, currency)
+    
+    total = convert_total(amount, denomination)
+    currency = handle_currency(total, currency)
+    
+    fput("tip #{recipient} #{total} #{currency}")
+    
+  end
+
+  def convert_total(amount, denomination)
+    # Take the amount, and the denomination, and determine the total
+    # in coppers. For example, 1 platinum would be 10000 coppers.
+
+    full_denom = nil
+
+    case denomination
+    when /\bp(l|la|lat|lati|latin|latinu|latinum)?\b/i
+      full_denom = "platinum"
+      return DRCM.convert_to_copper(amount, full_denom)
+    when /\bg(o|ol|old)?\b/i
+      full_denom = "gold"
+      return DRCM.convert_to_copper(amount, full_denom)
+    when /\bs(i|il|ilv|ilve|ilver)?\b/i
+      full_denom = "silver"
+      return DRCM.convert_to_copper(amount, full_denom)
+    when /\bb(r|ro|ron|ronz|ronze)?\b/i
+      full_denom = "bronze"
+      return DRCM.convert_to_copper(amount, full_denom)
+    when /\bc(o|op|opp|oppe|opper)?\b/i
+      full_denom = "copper"
+      return DRCM.convert_to_copper(amount, full_denom)
+    end
+  end
+
+  def handle_currency(total, currency)
+    # If user hasn't specified the currency, choose based
+    # on max currency on hand, else parse specified currency
+    # in case user has abbreviated it.
+    unless currency
+      # Get current wealth by currency
+      purse = {
+        "kronars": DRCM.check_wealth("Kronars"),
+        "lirums":  DRCM.check_wealth("Lirums"),
+        "dokoras": DRCM.check_wealth("dokoras")
+      }
+      # Finds the currency of our current max on hand
+      currency = purse.max_by{|k,v| v}[0]
+    else
+      # Parse abbreviations of the currency in the script call when given
+      case currency
+      when /\bk(r|ro|ron|rona|ronar|ronars)?\b/i
+        currency = "kronars"
+      when /\bl(i|ir|iru|irum|irums)?\b/i
+        currency = "lirums"
+      when /\bd(o|ok|oko|okor|okora|okoras)?\b/i
+        currency = "dokoras"
+      end
+    end
+    
+    return currency
+
+  end
+
+end
+
+Tip.new
--- a/tip.lic
+++ b/tip.lic
@@ -2,96 +2,82 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#tip
 =end
 
-# Background: I've always been annoyed by the need to specify tips in coppers.
-# DR's required syntax leaves you counting zeroes if you want to tip someone.
-# So a tip of 10 platinum kronars requires you to tip 100000 kronars. How many
-# zeroes is that? I lost count already. This simple script allows you to tip 
-# someone, in a similar syntax, but use the denomination to do so (e.g. "platinum").
-# <;tip gildaren 10 plat kron> now works. If you don't specify a currency,
-# this script also chooses the currency of the greatest coin you have on hand.
-
 custom_require.call %w[common common-money]
 
 class Tip
 
-  # include DRC for the message function
-  # include DRCM for use of converting amounts to copper
+  # include DRC for the bput function
+  # include DRCM for various money handling functions
   
   def initialize
     arg_definitions = [
+      # Two required arguments - recipient and amount; One optional argument, currency
+      # Mirrors DR's mechanics of "tip <person> 10000 kron", and also allows "tip <person> 10000" and assumes coppers
       [
         { name: 'recipient', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'The recipent of your tip. Abbreviations ok.' },
         { name: 'amount', regex: /^\d+$/i, variable: true, description: 'Numeric amount to tip, no commas (e.g. the number "100" in 100 kronars)' },
-        { name: 'denomination', regex: /(\bp(l|la|lat|lati|latin|latinu|latinum)?\b)|(\bg(o|ol|old)?\b)|(\bs(i|il|ilv|ilve|ilver)?\b)|(\bb(r|ro|ron|ronz|ronze)?\b)|(\bc(o|op|opp|oppe|opper)?\b)/i, variable: true, description: 'The denomination. Abbrevations ok. (e.g. plat, gold, silv)' },
-        { name: 'currency', regex: /(\bk(r|ro|ron|rona|ronar|ronars)?\b)|(l(i|ir|iru|irum|irums)?\b)|(\bd(o|ok|oko|okor|okora|okoras)?\b)/i, optional: true, variable: true, description: 'Optional currency. Abbreviations ok: kron, lirums, dokor' }         
+        { name: 'currency', regex: Regexp.union($CURRENCY_REGEX_MAP.values), optional: true, variable: true, description: 'Optional currency. Abbreviations ok: kron, lirums, dokor' }         
+      ],
+      # Three required arguments - recipient, amount, denomination - plus a fourth optional argument - currency
+      # Allows mechanic of "tip person 10 plat" and chooses your current max on hand
+      # or "tip person 10 plat kron"
+      [
+        { name: 'recipient', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'The recipent of your tip. Abbreviations ok.' },
+        { name: 'amount', regex: /^\d+$/i, variable: true, description: 'Numeric amount to tip, no commas (e.g. the number "100" in 100 kronars)' },
+        { name: 'denomination', regex: Regexp.union($DENOMINATION_REGEX_MAP.values), variable: true, description: 'The denomination. Abbrevations ok. (e.g. plat, gold, silv)' },
+        { name: 'currency', regex: Regexp.union($CURRENCY_REGEX_MAP.values), optional: true, variable: true, description: 'Optional currency. Abbreviations ok: kron, lirums, dokor' }
       ]
     ]
+    
     args = parse_args(arg_definitions)  
     gratuity(args.recipient, args.amount, args.denomination, args.currency)
 
   end
 
-  def gratuity(recipient, amount, denomination, currency)
+  @tip_success_patterns = [
+    /^You offer/i
+  ]
+
+  @tip_failure_patterns = [
+    /^I don't know who/i,
+    /you really should keep every bronze/i,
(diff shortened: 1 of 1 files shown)


KatoakDR:
I hate counting the zeroes, too! This will be a great convenience, thanks

tip.lic
```diff
@@ -0,0 +1,99 @@
...
+
+  # include DRC for the message function
+  # include DRCM for use of converting amounts to copper
+  
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'recipient', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'The recipent of your tip. Abbreviations ok.' },
+        { name: 'amount', regex: /^\d+$/i, variable: true, description: 'Numeric amount to tip, no commas (e.g. the number "100" in 100 kronars)' },
+        { name: 'denomination', regex: /(\bp(l|la|lat|lati|latin|latinu|latinum)?\b)|(\bg(o|ol|old)?\b)|(\bs(i|il|ilv|ilve|ilver)?\b)|(\bb(r|ro|ron|ronz|ronze)?\b)|(\bc(o|op|opp|oppe|opper)?\b)/i, variable: true, description: 'The denomination. Abbrevations ok. (e.g. plat, gold, silv)' },
```
rpherbig, line 25: I swear we've done something like this before... I just can't remember where.

@KatoakDR @Dartellum @jayRyan24 anyone else remember this?

tip.lic
```diff
@@ -0,0 +1,99 @@
...
+        "kronars": DRCM.check_wealth("Kronars"),
+        "lirums":  DRCM.check_wealth("Lirums"),
+        "dokoras": DRCM.check_wealth("dokoras")
+      }
+      # Finds the currency of our current max on hand
+      currency = purse.max_by{|k,v| v}[0]
+    else
+      # Parse abbreviations of the currency in the script call when given
+      case currency
+      when /\bk(r|ro|ron|rona|ronar|ronars)?\b/i
```
rpherbig, line 84: If we're repeating these RegEx's here, maybe they should be put somewhere more common?

tip.lic
```diff
@@ -0,0 +1,99 @@
...
+    # If user hasn't specified the currency, choose based
+    # on max currency on hand, else parse specified currency
+    # in case user has abbreviated it.
+    unless currency
+      # Get current wealth by currency
+      purse = {
+        "kronars": DRCM.check_wealth("Kronars"),
+        "lirums":  DRCM.check_wealth("Lirums"),
+        "dokoras": DRCM.check_wealth("dokoras")
+      }
```
KatoakDR, line 78: 💡 Not necessary for this pull request, but to avoid multiple calls to `wealth <currency>` game command, perhaps the `DRCM.wealth` function could be updated such that if no currency argument is provided then do `wealth` and parse the output to infer each currency.

```
> wealth

Wealth:
  3 gold Kronars (3000 copper Kronars).
  9 bronze, and 2 copper Lirums (92 copper Lirums).
  No Dokoras.
```

Would return the `purse` variable you've defined.
```ruby
{
  "kronars" => 3000,
  "lirums" => 92,
  "dokoras" => 0
}

File: tip.lic [hunk 4]
```diff
@@ -0,0 +1,99 @@
...
+      # Finds the currency of our current max on hand
+      currency = purse.max_by{|k,v| v}[0]
+    else
+      # Parse abbreviations of the currency in the script call when given
+      case currency
+      when /\bk(r|ro|ron|rona|ronar|ronars)?\b/i
+        currency = "kronars"
```
KatoakDR, line 85: 💡 Another candidate for a helper method in `common-money` to `get_canonical_currency(currency)` string

tip.lic
```diff
@@ -0,0 +1,99 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#tip
+=end
+
+# Background: I've always been annoyed by the need to specify tips in coppers.
```
rpherbig, line 5: This is the kind of info that would normally reside on elanthipedia in the docs for this script. Not opposed to keeping it here if you think it needs to stay, just pointing it out.

tip.lic
```diff
@@ -0,0 +1,99 @@
...
+    gratuity(args.recipient, args.amount, args.denomination, args.currency)
+
+  end
+
+  def gratuity(recipient, amount, denomination, currency)
+    
+    total = convert_total(amount, denomination)
+    currency = handle_currency(total, currency)
+    
+    fput("tip #{recipient} #{total} #{currency}")
```
rpherbig, line 39: Can this be a bput?

Something like this https://github.com/rpherbig/dr-scripts/blob/acc89c9a453fcb5afe06e4fa36e31e8b7b8bd4dd/bankbot.lic#L108

tip.lic
```diff
@@ -0,0 +1,99 @@
...
+
```
rpherbig, line 40: Tipping requires the other party to accept/reject/time out. Would it be desirable for this script to wait until the "tip process" has completed?

tip.lic
asechrest: It could be a bput, yes. And it originally was. Then I started handling the failure cases and realized that for many of them, the game already handles it with appropriate output, like if it can't find a recipient, or you don't have enough money to tip the desired amount.

But, I may have missed the other advantages of bput -- like that it'll wait for roundtime and recover from some other failure types and retry.

tip.lic
asechrest: At the moment I sort of have this designed to be an active script, meaning you're there using it live at the keyboard. Essentially just a quality of life script so we're not having to multiply by 1000 and carry the 2 just to figure out how much to tip.

BUT, I could flesh this out to be a more robust tipping script if ya'll thought it would be useful. Before writing this script I took a brief look at, I think it was bankbot, to see how they do it. In that script they do handle waiting for the tip process to complete, I believe by setting flags (which I haven't learned yet, but it would be a good chance to). But I believe that's designed to be fully automated.

Thoughts?

tip.lic
asechrest: No problem. I can put it on the -pedia.

tip.lic
```diff
@@ -0,0 +1,99 @@
...
+  def convert_total(amount, denomination)
+    # Take the amount, and the denomination, and determine the total
+    # in coppers. For example, 1 platinum would be 10000 coppers.
+
+    full_denom = nil
+
+    case denomination
+    when /\bp(l|la|lat|lati|latin|latinu|latinum)?\b/i
+      full_denom = "platinum"
```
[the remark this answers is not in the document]
asechrest: This is a great idea.

BUT, how do I limit the user's arg input in the initial script call to just currencies or abbreviations thereof, without having to write out the regex patterns as I've done above?

tip.lic
asechrest: Another good idea. If I make this changed it'll be fully backward compatible, right?

tip.lic
rpherbig: In general, bput is designed to be a replacement for fput

tip.lic
rpherbig: I'd stick with your particular use case for now

rpherbig: Maybe just note that it doesn't wait in the documentation or output a reminder to the user

tip.lic
asechrest: Gotcha. Will update.

tip.lic
Dartellum: Safe-room has a tipping method I believe. A setting in base exists.

asechrest: Just checked it out. If I'm reading it right, it's a tip method for a PC empath(s) of known name(s), embedded within the safe-room script. It also uses the give verb instead of the tip verb.

tip.lic
KatoakDR: After reviewing the `DRCM.check_wealth(currency)` and `DRCM.wealth(hometown)` methods, it may be more prudent to introduce a new function `DRCM.get_wealth` that takes no arguments and returns the hash of your wealth broken down by currency.

Then the `DRCM.check_wealth(currency)` method can be refactored to call `DRCM.get_wealth[currency]`, or be left as-is if you really didn't want to touch existing code. That function "ain't broke", my goal is to have a way that the tip script didn't need to spam `wealth` command 3 times :)

tip.lic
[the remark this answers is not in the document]
KatoakDR: Well, might have to define a global variable for the regexes in `common-money.lic` then reference them from `tip.lic`.

The `burgle.lic` script references `$HOMETOWN_LIST` which is defined in `common.lic`.
https://github.com/rpherbig/dr-scripts/blob/master/burgle.lic#L24

You might try something like:
```ruby
# common-money.lic
$DENOMINATION_REGEX_MAP = {
  'platinum' => /\bp(l|la|lat|lati|latin|latinu|latinum)?\b/i,
  'gold' => ...,
  'silver' => ...,
  'bronze' => ...,
  'copper' => ...
}

$CURRENCY_REGEX_MAP = {
  'kronars' => /\bk(r|ro|ron|rona|ronar|ronars)?\b/i,
  'lirums' => ...,
  'dokoras' => ...
}

# tip.lic
    arg_definitions = [
      [
        { name: 'recipient', regex: /^[A-z\.\s\-']+$/i, ... },
        { name: 'amount', regex: /^\d+$/i, ... },
        { name: 'denomination', regex: Regexp.union($DENOMINATION_REGEX_MAP.values), ... },
        { name: 'currency', regex: Regexp.union($CURRENCY_REGEX_MAP.values), ... }         
      ]
    ]

```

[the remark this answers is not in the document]
asechrest: > ```Regexp.union($DENOMINATION_REGEX_MAP.values)```

Now THAT is fancy.

asechrest:
Good comments so far. Got a few things in the works on this one. Going to end up submitting a PR to `common-money` and then working on this one again to seamlessly accept a few additional formats, so that this script can be used for tipping of any reasonable format. One ;tip to rule them all, you might say.

```ruby
;tip Gildaren 10000 kron #Functions exactly like the existing DR tip mechanic, assumes coppers
;tip Gildaren 10000 #No denomination or currency, assumes coppers and uses greatest currency on hand
;tip Gildaren 10 plat #No currency, tips 10 plat using greatest currency on hand
;tip Gildaren 10 plat kronars #Tips 10 plat kronars
```

asechrest:
I've significantly revised this PR. It now allows two sets of args for greater flexibility in tipping format. Uses bput for the tip input. Utilizes new and revised DRCM helper methods submitted via PR #5271 . Thus, will need to wait for that PR to merge.

This script now allows most reasonable formats of tipping, including formats that include the denomination or not, those that include the currency or not, etc.

The idea for this script, at least for me, is that it will fully replace DR's in-built tipping mechanic, because it allows the traditional DR tipping format (requires amount in coppers, requires currency specified), but also allows a number of other input formats, including tipping using the denomination.

Below are some examples of the script in use.

Tipping with no currency OR denomination specified. Assumes coppers and uses the greatest currency on hand.
```
>;tip juvat 10
--- Lich: tip active.
[tip]>wealth

Wealth:
  3 silver Kronars (300 copper Kronars).
  No Lirums.
  100 copper Dokoras (100 copper Dokoras).
>
[tip]>tip juvat 10 kronars
You offer Juvat a tip of 10 Kronars.  Type CANCEL TIP to withdraw the offer.
```

Tipping with denomination specified but no currency:
```
>;tip juv 10 bronze
--- Lich: tip active.
[tip]>wealth

Wealth:
  3 silver Kronars (300 copper Kronars).
  No Lirums.
  100 copper Dokoras (100 copper Dokoras).
>
[tip]>tip juv 100 kronars
You offer Juvat a tip of 100 Kronars.  Type CANCEL TIP to withdraw the offer.
```

Tipping with no denomination, but currency, specified. Currency abbreviated.
```
>;tip juv 100 kron
--- Lich: tip active.
[tip]>tip juv 100 kronars
You offer Juvat a tip of 100 Kronars.  Type CANCEL TIP to withdraw the offer.
```

Tipping using both currency and denomination. Heavy abbreviation of both.
```
>;tip juv 10 br kr
--- Lich: tip active.
[tip]>tip juv 100 kronars
You offer Juvat a tip of 100 Kronars.  Type CANCEL TIP to withdraw the offer.
```

tip.lic
```diff
@@ -0,0 +1,85 @@
...
+    /^You already have a tip offer outstanding/i,
+    /^But you don't have that much/i,
+    /^Hmmm, tipping yourself/i    
+  ]
+
+  def gratuity(recipient, amount, denomination, currency)    
+    # Grab the total amount in copper given a numerical amount
+    # and a denomination (e.g. 100 bronze). Utilize the
+    # helper method in common-money.
+    total = DRCM.convert_to_copper(amount, denomination)
```
KatoakDR, line 53: 💡 To now avoid the [warning](https://github.com/rpherbig/dr-scripts/pull/5271/files#diff-7dae83dd2163ccf532277ddc13167af83188df20c3bdccd306234d70ec92eb34R53) from common-money's `convert_to_copper` about unknown denomination, `Tip` can provide a reasonable default of `coppers`.

```suggestion
    denomination ||= 'copper'
    total = DRCM.convert_to_copper(amount, denomination)
```

tip.lic
asechrest: Added to common-money

tip.lic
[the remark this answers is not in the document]
asechrest: Added to common-money

tip.lic
asechrest: Now deployed as a bput.

tip.lic
asechrest: Thanks for this. Neat reading on double-pipe equals, and truthiness, falseyness.

asechrest:
All comments resolved. Re-tested the various ;tip formats.

(To remind, relies on the common-money PR.)

```
>;tip gild 10
--- Lich: tip active.
[tip]>wealth

Wealth:
  3 silver Kronars (300 copper Kronars).
  No Lirums.
  2 gold, and 39 copper Dokoras (2,039 copper Dokoras).
>
[tip]>tip gild 10 dokoras
Hmmm, tipping yourself?  That's an interesting concept, but it won't work.
>
--- Lich: tip has exited.
>;tip gild 10 bronze
--- Lich: tip active.
[tip]>wealth

Wealth:
  3 silver Kronars (300 copper Kronars).
  No Lirums.
  2 gold, and 39 copper Dokoras (2,039 copper Dokoras).
>
[tip]>tip gild 100 dokoras
Hmmm, tipping yourself?  That's an interesting concept, but it won't work.
>
--- Lich: tip has exited.
>;tip gild 100 kron
--- Lich: tip active.
[tip]>tip gild 100 kronars
Hmmm, tipping yourself?  That's an interesting concept, but it won't work.
>
--- Lich: tip has exited.
>;tip gild 10 br kr
--- Lich: tip active.
[tip]>tip gild 100 kronars
Hmmm, tipping yourself?  That's an interesting concept, but it won't work.
>
--- Lich: tip has exited.
```

KatoakDR (approved):

tip.lic
```diff
@@ -0,0 +1,87 @@
...
+    /^But you don't have that much/i,
+    /^Hmmm, tipping yourself/i    
+  ]
+
+  def gratuity(recipient, amount, denomination, currency)    
+    # Grab the total amount in copper given a numerical amount
+    # and a denomination (e.g. 100 bronze). Utilize the
+    # helper method in common-money. If no denomination given,
+    # set to copper.
+    denomination ||= 'copper'
```
rpherbig, line 54: The idiomatic way is to do this in the function argument list (and avoid re-assigning to a function input), but it's pretty minor

tip.lic
```diff
@@ -0,0 +1,87 @@
...
+    currency = handle_currency(currency)
+    
+    DRC.bput("tip #{recipient} #{total} #{currency}", @tip_success_patterns, @tip_failure_patterns)    
+  end
+
+  def handle_currency(currency)
+    # If user hasn't specified the currency, choose based
+    # on max currency on hand, else parse specified currency
+    # in case user has abbreviated it.
+    unless currency
```
rpherbig, line 70: I prefer `if` over `unless` in this case where both branches are defined (e.g. switch unless to if and swap the branches so the logic is the same), but it's pretty minor.

rpherbig:
Code changes look good ✔️ 

The common PR has been merged.

Remind me to merge this in a few days.

tip.lic
asechrest: Gotcha. Just convention? Or easier to follow the logic (which I agree with)?

rpherbig: It's easier for people unfamiliar with Ruby's `unless` to follow; and IMO a best practice, barring performance concerns where you need the branches in a specific order likely due to skewed % of taking each branch.

tip.lic
KatoakDR: I agree, and that's what I did originally; however, with this method it accepts 4 non-optional arguments and it's the 3rd argument that I'm trying to provide a default for if `nil` is passed in.

For example, `gratuity('katoak', 50, nil, nil)` is what will get called so `nil` will get assigned to the arguments. Specifying the default in the function argument list would not have an effect here since the 3rd and 4th arguments are being provided.

asechrest:
I'm thinking this one is good to merge now? Thanks.

PR #5263, opened 2021-12-02
Update stowing of enchantables to be consistent

Hiinky:
stow_hands uses stow and stow_items honors crafting bag.  Will snag if crafting bag is not default container.  Other tiers are correct, just updating this last one to match.


rpherbig:
This whole section looks ripe for extracting a helper function, but it's not necessary for this PR

PR #5264, opened 2021-12-02
Adding astrology to the list of scripts that ;play shouldn't bother

MahtraDR:
;play tried to bother my ;astro run today, and I realized that shouldn't be happening.


MahtraDR:
> researcher is another that ;play tries to, and really shouldn't mess with.

PR #5269, opened 2021-12-03
[data] [base-spells] Default `recast` and `prep` properties

KatoakDR:
### Background
* `combat-trainer` will recast buffs defined in `buff_spells:` of your yaml if the spell data has either `recast`, `recast_every`, or `expire` properties.
* Not all spells in base data had one of those properties and so they weren't being cast in combat (e.g. [Gam Inran](https://github.com/rpherbig/dr-scripts/pull/5267) which was recently fixed, and [Aether Wolves](https://discord.com/channels/745675889622384681/745678330933805157/916347740252954635) as reported by Mahtra).
* `dependency` sets if doesn't exist a `prep: target` for `offensive_spells:` defined in your yaml if they are for the Targeted Magic skill, or if a Sorcery that refers to a Targeted Magic skill. However, if those spells are referenced in other yaml properties, like `combat_spell_training:` then that logic isn't being applied and leads to TM spells being prepared but never targeted and consequentially never cast. Rather than continue to play whack-a-mole in dependency, opting to ensure every spell has a default `prep` property.

### Changes
* For spells without a `recast` property:
  - Add `recast: 2` for ritual spells (to account for time needed to invoke the ritual focus)
  - Add `recast: 1` for all other spells
* For spells without a `prep` property:
  - Add `prep: prepare` for non-TM spells
  - Add `prep: target` for TM spells
* For consistency, and based on [this conversation](https://discord.com/channels/745675889622384681/745696628714897508/916461610598547558), cyclics use `recast: -1` to emphasize that their duration is as long as the spell is maintained.
* For consistency, renamed `prep_type` to `prep` for the handful of spells that used that alias

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -484,7 +484,7 @@ spell_data:
     cyclic: true
     mana: 7
     mana_type: elemental
-    recast: 1
+    recast: -1
   Absolution:
     skill: Utility
     harmless: true
@@ -525,7 +525,7 @@ spell_data:
     prep: prepare
     cyclic: true
     mana: 5
-    recast: 1
+    recast: -1
     mana_type: life
     harmless: true
   Aesrela Everild:
@@ -543,7 +543,7 @@ spell_data:
     cyclic: true
     mana: 4
     mana_type: elemental
-    recast: 1
+    recast: -1
   Aether Spheres:
     skill: cantrip
     abbrev: C AE S
@@ -562,7 +562,7 @@ spell_data:
     cyclic: true
     mana: 2
     mana_type: elemental
-    recast: 1
+    recast: -1
   Aethereal Image:
     skill: cantrip
     abbrev: C AE I
@@ -637,7 +637,7 @@ spell_data:
     cyclic: true
     mana: 5
     mana_type: elemental
-    recast: 1
+    recast: -1
   Anther's Call:
     skill: Debilitation
     abbrev: ANC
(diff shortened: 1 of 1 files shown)


data/base-spells.yaml
```diff
@@ -632,111 +661,134 @@ spell_data:
...
   Artificer's Eye:
     skill: Augmentation
     abbrev: ART
     mana: 5
     recast: 1
     mana_type: lunar
+    prep: prepare
   Aspects of the All-God:
-    skill: Augmentation  # Also utility
+    skill: Augmentation # Also Utility
```
rpherbig, line 681: I keep forgetting about these dual-skill spells. We ought to make this a list at some point...

rpherbig:
The changes look good to me ✔️ 

How do you suggest this be tested?

KatoakDR:
> The changes look good to me ✔️
> 
> How do you suggest this be tested?

Thanks -- I'm in the process myself of testing on a warmie (who has cyclics), a paladin, a necromancer, and a ranger.

I'll try to recruit some folks who have clerics, bards, empaths, and traders.

And on that note, I'll put this back in draft until collect more test confidence.

KatoakDR:
A couple folks and I have been running with these changes without adverse effects. I think we're good to move forward if you're comfortable with it.

data/base-spells.yaml
KatoakDR: Someday, but not this day 😉

PR #5271, opened 2021-12-04
[script][common-money] Broaden common money

asechrest:
This PR makes a few updates to common-money. The impetus for this was separate PR #5259 which is a smart tipping script. These updates centralize and revise some of the money handling and may be useful for future use.

1. Adds global regex maps for use in parsing denominations and currencies and abbreviations thereof. I'm using this in my tip script to limit arg inputs.
2. Revises the `convert_to_copper` method to handle conversion if the denomination has been abbreviated. Now also handles conversion if the denomination is given explicitly as copper.
3. Adds method `get_canonical_currency`, which takes an abbreviated currency and returns the full canonical currency name.
4. Adds a `get_total_wealth method`, which returns a hash of the character's current total wealth in copper by currency.

Special thanks to @KatoakDR for significant Ruby help.

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -137,26 +137,18 @@ module DRCM
     # each line after we recognize we've hit the Wealth block.
     wealth_lines.each do |line|
       case line
-      when  /^Wealth:/i
+      when /^Wealth:/i
         # This is the start of our Wealth lines.
         # We don't need to parse this line. Break out of loop.
         break
-      when /\(\d+ copper Kronars\)|^No Kronars/i
-        kronars_line = line
-      when /\(\d+ copper Lirums\)|^No Lirums/i
-        lirums_line = line
-      when /\(\d+ copper Dokoras\)|^No Dokoras/i
-        dokoras_line = line
+      when /\(\d+ copper Kronars\)/i
+        kronars = line.scan(\d+).first.to_i
+      when /\(\d+ copper Lirums\)/i
+        lirums = line.scan(\d+).first.to_i
+      when /\(\d+ copper Dokoras\)/i
+        dokoras = line.scan(\d+).first.to_i
       end
     end
-
-    # Scan each wealth line for the value in coppers,
-    # capture that numeric value, then integerize.
-    # If the line indicates no wealth of that currency, accept
-    # the instantiated value of 0 for that currency.
-    kronars = kronars_line.scan(/\((\d+) copper Kronars\)/i).first.first.to_i unless kronars_line =~ /^No/    
-    lirums = lirums_line.scan(/\((\d+) copper lirums\)/i).first.first.to_i unless lirums_line =~ /^No/        
-    dokoras = dokoras_line.scan(/\((\d+) copper dokoras\)/i).first.first.to_i unless dokoras_line =~ /^No/
     
     # Set up a hash of currency and corresponding value
     # in coppers. Return the hash for future use.
--- a/common-money.lic
+++ b/common-money.lic
@@ -138,7 +138,7 @@ module DRCM
         # We don't need to parse this line. Break out of loop.
         break
       when /\(\d+ copper Kronars\)/i
-        kronars = line.scan(/\((\d+) copper Kronars\)/i).first.first.to_i
+        kronars = line.scan(/\((\d+) copper kronars\)/i).first.first.to_i
       when /\(\d+ copper Lirums\)/i
         lirums = line.scan(/\((\d+) copper lirums\)/i).first.first.to_i
       when /\(\d+ copper Dokoras\)/i
--- a/common-money.lic
+++ b/common-money.lic
@@ -138,11 +138,11 @@ module DRCM
         # We don't need to parse this line. Break out of loop.
         break
       when /\(\d+ copper Kronars\)/i
-        kronars = line.scan(/\d+/).first.to_i
+        kronars = line.scan(/\((\d+) copper Kronars\)/i).first.first.to_i
       when /\(\d+ copper Lirums\)/i
-        lirums = line.scan(/\d+/).first.to_i
+        lirums = line.scan(/\((\d+) copper lirums\)/i).first.first.to_i
       when /\(\d+ copper Dokoras\)/i
-        dokoras = line.scan(/\d+/).first.to_i
+        dokoras = line.scan(/\((\d+) copper dokoras\)/i).first.first.to_i
       end
     end
     
--- a/common-money.lic
+++ b/common-money.lic
@@ -42,7 +42,8 @@ module DRCM
   def convert_to_copper(amount, denomination)
     # Convert to copper given denomination (abbreviation permitted)
     # If no denomination specified, return the integer amount (assumed to be coppers)
-    if denomination
+    denomination = denomination.strip # trim whitespace and also convert nil to empty string
+    if !denomination.empty?
       return (amount.to_i * 10_000) if 'platinum'.start_with?(denomination.downcase)
       return (amount.to_i *   1000) if 'gold'.start_with?(denomination.downcase)
       return (amount.to_i *    100) if 'silver'.start_with?(denomination.downcase)
--- a/common-money.lic
+++ b/common-money.lic
@@ -49,10 +49,9 @@ module DRCM
       return (amount.to_i *    100) if 'silver'.start_with?(denomination.downcase)
       return (amount.to_i *     10) if 'bronze'.start_with?(denomination.downcase)
       return (amount.to_i *      1) if 'copper'.start_with?(denomination.downcase)
-    else
-      DRC.message("Unknown denomination, assuming coppers: #{denomination}")
-      amount.to_i
     end
+    DRC.message("Unknown denomination, assuming coppers: #{denomination}")
+    amount.to_i
   end
 
   # Returns full canonical currency if given an abbreviation
--- a/common-money.lic
+++ b/common-money.lic
@@ -8,6 +8,25 @@ custom_require.call(%w[common common-travel drinfomon])
 module DRCM
   module_function
 
+  # Map of regex abbreviations for coin denominations
+  # Supports abbreviations of input like DR
+  $DENOMINATION_REGEX_MAP = {
+    'platinum' =>   /\bp(l|la|lat|lati|latin|latinu|latinum)?\b/i,
+    'gold' =>       /\bg(o|ol|old)?\b/i,
+    'silver' =>     /\bs(i|il|ilv|ilve|ilver)?\b/i,
+    'bronze' =>     /\bb(r|ro|ron|ronz|ronze)?\b/i,
+    'copper' =>     /\bc(o|op|opp|oppe|opper)?\b/i
+  }
+
+  # Map of regex abbreviations for currency
+  # Supports abbreviations of input like DR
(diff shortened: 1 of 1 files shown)


common-money.lic
```diff
@@ -64,11 +99,74 @@ module DRCM
...
+      end
+    end
+
+    # Scan each wealth line for the value in coppers,
+    # capture that numeric value, then integerize.
+    # If the line indicates no wealth of that currency, accept
+    # the instantiated value of 0 for that currency.
+    kronars = kronars_line.scan(/\((\d+) copper Kronars\)/i).first.first.to_i unless kronars_line =~ /^No/    
+    lirums = lirums_line.scan(/\((\d+) copper lirums\)/i).first.first.to_i unless lirums_line =~ /^No/        
+    dokoras = dokoras_line.scan(/\((\d+) copper dokoras\)/i).first.first.to_i unless dokoras_line =~ /^No/
```
KatoakDR, line 158: 💡 Since the parsing of amount out of the wealth lines is contextual to the denomination, we can consolidate the logic to occur in the loop that identifies the wealth lines, thus removing some duplicated regex patterns.

```suggestion
    wealth_lines.each do |line|
      case line
      when /^Wealth:/i
        # This is the start of our Wealth lines.
        # We don't need to parse this line. Break out of loop.
        break
      when /\(\d+ copper Kronars\)/i
        kronars = line.scan(\d+).first.to_i
      when /\(\d+ copper Lirums\)/i
        lirums = line.scan(\d+).first.to_i
      when /\(\d+ copper Dokoras\)/i
        dokoras = line.scan(\d+).first.to_i
      end
    end
```

Additionally, with the above you can then get rid of the `<denomination>_line` variables as they aren't used anymore.

KatoakDR (commented):
Well done pulling this reusable logic out of `tip` and into the `common-money` library. I had a few comments where some repeated logic could be consolidated, but overall good job 👍

common-money.lic
```diff
@@ -64,11 +101,65 @@ module DRCM
...
 
+  def get_total_wealth
+    # This method captures your current total on-hand wealth
+    # and returns a hash representing the numerical value in
+    # coppers of each currency.    
+
+    # Set up variables to capture each line of wealth output by currency
+    kronars_line = nil
+    lirums_line = nil
+    dokoras_line = nil
```
KatoakDR, line 116: Thanks for the quick updates, these lines can now be deleted since no longer being used
```suggestion
```

asechrest: > Thanks for the quick updates, these lines can now be deleted since no longer being used

Thanks for the suggestions. I'm working on deleting those now. I'm also getting an error, so diagnosing that.

common-money.lic
```diff
@@ -64,11 +101,65 @@ module DRCM
   def check_wealth(currency)
     DRC.bput("wealth #{currency}", /\(\d+ copper #{currency}\)/i, /No #{currency}/i).scan(/\d+/).first.to_i
   end
-
   def wealth(hometown)
```
KatoakDR, line 104: Oops, I suggested an entire line be deleted when I meant to suggest just remove the whitespace. This suggestion adds the blank line back (no whitespace) between the `check_wealth` and `wealth` methods.
```suggestion
  end

  def wealth(hometown)
```

asechrest: > Applying suggestions on deleted lines is currently not supported.

I'll fix and commit.

asechrest: Added back blank line. Also fixed the regex in .scan.

KatoakDR (approved):
🥳

asechrest:
Testing ;tip again, and it doesn't appear to choose the max currency correctly. Can't figure that's the fault of these changes, so I'll diagnose within ;tip.

```
>;tip gild 100
--- Lich: tip active.
| Unknown denomination, assuming coppers: 
[tip]>wealth

Wealth:
  3 silver Kronars (300 copper Kronars).
  No Lirums.
  100 copper Dokoras (100 copper Dokoras).
>
[tip]>tip gild 100 **dokoras**
```

asechrest:
I take that back, it is these changes. I see the issue, and will fix after I get back from the grocery store. These changes grab the first set of digits in a currency line. We want to grab the set of digits after the first paren.

KatoakDR:
> These changes grab the first set of digits in a currency line. We want to grab the set of digits after the first paren.

Sorry about that!

asechrest:
> > These changes grab the first set of digits in a currency line. We want to grab the set of digits after the first paren.
> 
> Sorry about that!

No apology needed! These are good suggestions that help me understand some more advanced considerations.

common-money.lic
```diff
@@ -69,6 +106,56 @@ module DRCM
...
+        # This is the start of our Wealth lines.
+        # We don't need to parse this line. Break out of loop.
+        break
+      when /\(\d+ copper Kronars\)/i
+        kronars = line.scan(/\d+/).first.to_i
+      when /\(\d+ copper Lirums\)/i
+        lirums = line.scan(/\d+/).first.to_i
+      when /\(\d+ copper Dokoras\)/i
+        dokoras = line.scan(/\d+/).first.to_i
+      end
```
KatoakDR, line 146: Apologies for the bug my suggestion introduced. I forgot that at this point the context of `line` is the entire line, not just the regex pattern that was matched. In the following, the use of `Regexp.last_match` ensures we're looking at only the part of the line that indeed matched the pattern. I also named the capturing group that matched the coins so that you can grab them right out of the matched result, no further scanning or parsing necessary.

My character is out hunting at moment so I haven't tested it, but I'm pretty confident with the below snippet.

```ruby
      when /\((?<amount>\d+) copper Kronars\)/i
        kronars = Regexp.last_match[:amount].to_i
      when /\((?<amount>\d+) copper Lirums\)/i
        lirums = Regexp.last_match[:amount].to_i
      when /\((?<amount>\d+) copper Dokoras\)/i
        dokoras = Regexp.last_match[:amount].to_i
      end
```

This next bit isn't necessary, just sharing as FYI. You could get fancier with naming capturing groups to get the currency, too, and then be more dynamic with populating the total wealth map result.

```ruby
      when /\((?<amount>\d+) copper (?<currency>Kronars|Lirums|Dokoras)\)/i
        total_wealth[Regexp.last_match[:currency].downcase] = Regexp.last_match[:amount].to_i
      end

```

KatoakDR: You don't have to use what I proposed above, just trying to atone for my mistake :)

common-money.lic
```diff
@@ -21,11 +40,29 @@ module DRCM
...
   def convert_to_copper(amount, denomination)
-    return (amount.to_i * 10_000) if 'platinum' =~ /^#{denomination}/
-    return (amount.to_i * 1000) if 'gold' =~ /^#{denomination}/
-    return (amount.to_i * 100) if 'silver' =~ /^#{denomination}/
-    return (amount.to_i * 10) if 'bronze' =~ /^#{denomination}/
-    amount.to_i
+    # Convert to copper given denomination (abbreviation permitted)
+    # If no denomination specified, return the integer amount (assumed to be coppers)
+    denomination = denomination.strip # trim whitespace and also convert nil to empty string
+    if !denomination.empty?
```
rpherbig, line 46: This is a change in behavior in the case where denomination is not-empty and also fails to match any of the checks. E.g. someone passes in `glod` or `sliver`. Do you think that case should also return `amount.to_i`?

rpherbig:
Change look good once comments are resolved ✔️

common-money.lic
asechrest: This is a good question. 

If I'm not mistaken, the original accepted a denomination arg. If it matched the full DR denominations, it did conversions. If it didn't match (coper, sliver, etc.), or was nil, it integerized the amount and returned it.

I believe my first revision unintentionally changed that behavior. If denomination was not nil, it checked for matches, including substrings, to the DR denominations. If denomination WAS nil, it integerized the amount arg, and returned it. But if the denomination was existed but was misspelled, for example, I'm betting my revision would do nothing at all.

Now this revision per @KatoakDR  is a change in behavior from MY revision, but I think behaves similarly to the original, with the added advantage that it now warns the user that the passed-in denomination arg is not known, and still completes the function by integerizing the amount arg and returning it.

I kind of like Katoak's revision best. But now that we've had some discussion on it, I could easily mirror the original's behavior if that's best.

common-money.lic
asechrest: Haven't quite wrapped my head around these fancier versions. But I did commit a fix that is within my current ability.

common-money.lic
KatoakDR: > This is a change in behavior in the case where denomination is not-empty and also fails to match any of the checks

Good catch. To restore that original behavior then I think we just need to move the default logic out of the ELSE so it always runs if the script gets to that point. In the below proposal, if the denomination is not blank and does not match the platinum/gold/silver/bronze/copper checks then it hits the default case and will message the warning and assume coppers.

```ruby
  def convert_to_copper(amount, denomination)
    denomination = denomination.strip # trim whitespace and also convert nil to empty string
    if !denomination.empty?
      return (amount.to_i * 10_000) if 'platinum'.start_with?(denomination.downcase)
      return (amount.to_i *   1000) if 'gold'.start_with?(denomination.downcase)
      return (amount.to_i *    100) if 'silver'.start_with?(denomination.downcase)
      return (amount.to_i *     10) if 'bronze'.start_with?(denomination.downcase)
      return (amount.to_i *      1) if 'copper'.start_with?(denomination.downcase)
    end
    DRC.message("Unknown denomination, assuming coppers: #{denomination}")
    amount.to_i
end
```

common-money.lic
```diff
@@ -69,6 +106,56 @@ module DRCM
...
+      when /^Wealth:/i
+        # This is the start of our Wealth lines.
+        # We don't need to parse this line. Break out of loop.
+        break
+      when /\(\d+ copper Kronars\)/i
+        kronars = line.scan(/\((\d+) copper Kronars\)/i).first.first.to_i
+      when /\(\d+ copper Lirums\)/i
+        lirums = line.scan(/\((\d+) copper lirums\)/i).first.first.to_i
+      when /\(\d+ copper Dokoras\)/i
+        dokoras = line.scan(/\((\d+) copper dokoras\)/i).first.first.to_i
```
KatoakDR, line 145: 💅 Purely for consistency within the regex patterns, suggest they all follow either capital case or lower case. Note the current code has Kronars, lirums, dokoras; a mix of upper and lower case.

common-money.lic
asechrest: Returned to original behavior.  Test:

```
>;tip gil 10 floof
--- Lich: tip active.
[tip: Denomination is floof]
[tip: Currency is ]
| Unknown denomination, assuming coppers: floof
[tip]>wealth

Wealth:
  3 silver Kronars (300 copper Kronars).
  No Lirums.
  2 gold, and 39 copper Dokoras (2,039 copper Dokoras).
>
[tip]>tip gil 10 dokoras
Hmmm, tipping yourself?  That's an interesting concept, but it won't work.```

common-money.lic
asechrest: Done!

asechrest:
I believe all comments are now resolved.

KatoakDR (approved):

PR #5275, opened 2021-12-05
[script] [sell-loot] Support new setting pick.component_container

KatoakDR:
### Background
* With the [rewrite](https://github.com/rpherbig/dr-scripts/pull/5167) of `pick` script, the new setting for storing harvested trap components is the nested `pick.component_container` property.
* The original `component_container` property is still supported for backwards compatibility, but if you've moved to the new setting then `sell-loot` was not selling your traps.

### Changes
* Check for the old and new settings just like the `pick` script does

### DIFF
--- a/sell-loot.lic
+++ b/sell-loot.lic
@@ -58,7 +58,7 @@ class SellLoot
 
     sell_bundle if @settings.sell_loot_bundle
 
-    sell_traps(@settings.component_container) if @settings.sell_loot_traps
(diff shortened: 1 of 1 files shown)


rpherbig:
Is the goal to deprecate the root-level `component_container` setting? If so, a validator would be prudent (to warn the user).

KatoakDR:
> Is the goal to deprecate the root-level `component_container` setting? If so, a validator would be prudent (to warn the user).

Yes, will add a validation to convert to the new setting, thanks

PR #5277, opened 2021-12-07
[script] [combat-trainer] Add `strict_weapon_stance` setting

KatoakDR:
### Background
* In the Lich Discord, a group of players have been discussing concerns about how `stances:` and `priority_defense` were not being respected by ranged weapons.
* After review of the code, we realized that even when a weapon stance is provided, combat-trainer will still dynamically choose between the first two preferred defenses and prioritize the one that has lower learning rate/rank (presumably to optimize learning).
* Some players shared that their characters get hurt bad if combat-trainer switches them into a Shield first stance.
* To provide people the option to opt-out of this dynamic stance logic, this PR introduces a new setting `strict_weapon_stance` (boolean) that defaults to `false` to preserve current behavior.

### Changes
* Add `strict_weapon_stance` to `base.yaml`, and update comments on the stance settings.
* Update the `check_stance` method in `combat-trainer` to use the exact weapon stance as configured in the yaml when `strict_weapon_stance: true`, otherwise do its current behavior.

### Example Config

_original behavior, no change necessary_
```yaml
stances:
  Bow:
  - Evasion
  - Shield Usage
  - Parry Ability

# If false then combat-trainer does its current behavior,
# it will dynamically choose between the first two listed defenses
# of the weapon (in this case between Evasion and Shield) and
# stance to the one that has lowest learning rate/rank of the two.
strict_weapon_stance: false
```

_new option_
```yaml
stances:
  Bow:
  - Evasion
  - Shield Usage
  - Parry Ability

# If true then combat-trainer will stance the character precisely
# as the defenses are listed in the weapon stance in the yaml.
# In this case, when training bows then the character will
# prioritize Evasion, allocate remaining to Shield, and any extra to Parry.
# The script would never dynamically decide to stance Shield before Evasion
# when training the bow weapon skill.
strict_weapon_stance: true
```

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -245,11 +245,22 @@ class SetupProcess
     points = override || previous.values.inject(&:+)
 
     priority = if game_state.current_weapon_stance
-                 game_state.sort_by_rate_then_rank(game_state.current_weapon_stance[0..1]) + [game_state.current_weapon_stance.last]
+                 if game_state.strict_weapon_stance
+                   # Player wants their weapon stance strictly adhered to, no change.
+                   game_state.current_weapon_stance
+                 else
+                   # Player has a preference for the first two stances for this weapon
+                   # and is open to having them dynamically prioritized to optimize learning.
+                   game_state.sort_by_rate_then_rank(game_state.current_weapon_stance[0..1]) + [game_state.current_weapon_stance.last]
+                 end
                elsif @priority_defense
+                 # Player does not have a weapon specific stance but does want
+                 # a preferred defense to always be 100%.
                  rest = ['Evasion', 'Parry Ability', 'Shield Usage'] - [@priority_defense]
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -3727,6 +3738,9 @@ class GameState
     @target_increment = settings.combat_trainer_target_increment
     echo("  @target_increment: #{@target_increment}") if $debug_mode_ct
 
+    @strict_weapon_stance = settings.strict_weapon_stance
```
rpherbig, line 3741: It's minor, but I don't think this needs to be in `game_state` if it's only going to be used once by a single process. I'd rather have it pulled out in `SetupProcess::initialize` (see line 10-ish).

KatoakDR: Ah, gotcha. I was following the pattern of `current_weapon_stance` to define the new variable. I can switch it to be a class variable in `SetupProcess` instead.

PR #5278, opened 2021-12-07
[script] [combat-trainer] Mitigate accidentally dropping items after script unpaused

KatoakDR:
### Background
* [Naimo](https://discord.com/channels/745675889622384681/745678330933805157/917646189032263700) and others were sharing in Lich Discord that `combat-trainer` would occassionally [drop items](https://pastebin.com/amKRSMwt) if the script was paused/unpaused (usually be `wand-watcher`, but could have been `almanac` or other script) in the middle of the stow loot routine.
* The issue is a race condition between `combat-trainer` becoming unpaused and the Lich thread catching up with what's currently in your hand isn't still the wand or almanac the pausing script just had in your hands.

### Changes
* In an effort to ensure the Lich thread synchronously parses the game output to know what's in your hands before `combat-trainer` does the logic to determine if it should trash what's in your hand, add a `glance` action.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -595,21 +595,24 @@ class LootProcess
       game_state.wield_whirlwind_offhand
     end
     return unless tried_loot
-    # Mitigating a race condition when wand-watcher or another script pauses
-    # combat-trainer while it's looting, then resumes in the middle of this function.
-    # Occassionally, the thread hasn't updated the left_hand/right_hand variables
-    # to know what's currently in your hand and will trash what's in your hand!
     pause 1
-    bput('glance', 'You glance .*')
-    if left_hand != pair.first && !@equipment_manager.is_listed_item?(left_hand)
-      echo("out of room, failed to store #{left_hand}")
-      game_state.unlootable(GameObj.left_hand.noun.downcase)
-      dispose_trash(left_hand)
+    if DRC.left_hand != pair.first || DRC.right_hand != pair.last
+      # Mitigating a race condition when wand-watcher or another script pauses
+      # combat-trainer while it's looting, then resumes in the middle of this function.
+      # Occassionally, the thread hasn't updated the left_hand/right_hand variables
+      # to know what's currently in your hand and will trash what's in your hand!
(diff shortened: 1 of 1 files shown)


rpherbig:
Oof, that's a bad user experience. And a rough interaction of scripts.

I don't like glancing that often (every time we are looting, right?), but if we need to, then ok.

Is there a way to detect `combat-trainer` was paused and only glance then? What about only glancing if the items don't match (i.e. check if we should trash, if so then glance and check again)?

KatoakDR:
> Oof, that's a bad user experience. And a rough interaction of scripts.
> 
> I don't like glancing that often (every time we are looting, right?), but if we need to, then ok.
> 
> Is there a way to detect `combat-trainer` was paused and only glance then? What about only glancing if the items don't match (i.e. check if we should trash, if so then glance and check again)?

I like the "only glance if the items don't match" to cut down on how often the glancing occurs, particularly since the glancing is only to try and ensure if the items mismatched _really are mismatched._ If there is no mismatch then shouldn't be a problem. Good suggestion!

combat-trainer.lic
```diff
@@ -596,15 +596,23 @@ class LootProcess
...
+    end
+    if DRC.left_hand != pair.first && !@equipment_manager.is_listed_item?(DRC.left_hand)
+      DRC.message("Out of room, failed to store: #{DRC.left_hand}")
+      game_state.unlootable(DRC.left_hand_noun.downcase)
+      DRCI.dispose_trash(DRC.left_hand)
+    end
+    if DRC.right_hand != pair.last && !@equipment_manager.is_listed_item?(DRC.right_hand)
+      DRC.message("Out of room, failed to store: #{DRC.right_hand}")
+      game_state.unlootable(DRC.right_hand_noun.downcase)
+      DRCI.dispose_trash(DRC.right_hand)
```
KatoakDR, line 615: * Added explicit use of `DRC` and `DRCI` prefixes
* Switch use of `GameObj` with `DRC` to maintain the abstraction from core Lich
* Use `DRC.message` instead of `echo` to display the error

rpherbig:
Looks good to me ✔️

PR #5281, opened 2021-12-08
[script][common-arcana] Honor settings.telescope_storage

asechrest:
Issue first reported by @MahtraDR [in Discord](https://discordapp.com/channels/745675889622384681/912127186419482664/917606108410032158). Special thanks to him for helping test these changes.

Basically, sometimes ;buff and ;combat-trainer will not honor your telescope storage case. Each seems to _eventually_ rely on `DRCA.find_visible_planets`, which in its un-revised state uses a simple stow instead of using telescope storage. There is an existing, but stalled, PR #5207 which attempts to update both the common-* and other files in a single PR.

This PR overlaps that one. But also, instead of storing your telescope via `bput`, utilizes the `DRCMM.store_telescope` method. All we need to do is properly pass a settings args to access.

This will support another PR, eventually, to `combat-trainer`, which passes the @settings arg through to also now utilize `DRCA.store_telescope`.

@MahtraDR has tested this change on his MM, but we welcome others.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-arcana
 =end
 
-custom_require.call(%w[common common-items events spellmonitor drinfomon])
+custom_require.call(%w[common common-items common-moonmage events spellmonitor drinfomon])
 
 $MANA_MAP = {
   'weak' => %w[dim glowing bright],
@@ -518,16 +518,17 @@ module DRCA
     end
   end
 
-  def update_astral_data(data)
+  def update_astral_data(data, settings)
     if data['moon']
       data = set_moon_data(data)
     elsif data['stats']
-      data = set_planet_data(data)
+      data = set_planet_data(data, settings)
     end
     data
   end
 
-  def find_visible_planets(planets)
+  def find_visible_planets(planets, settings)
+    return unless settings.have_telescope
     return if DRC.bput('get my telescope', 'You get', 'What were you', 'You are already') == 'What were you'
 
     Flags.add('planet-not-visible', 'turns up fruitless')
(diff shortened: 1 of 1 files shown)


MahtraDR:
Given the original failures could easily be reproduced by buffing IOTS out of combat (for buff.lic) and in combat (for combat-trainer.lic) the testing methodology consisted of pre-buffing IOTS and checking the telescope was then placed in the telescope container as expected. With IOTS set as a buff, I then went into combat and released IOTS, which caused combat trainer to recast it. Again, the telescope ended up in the correct container as per the setting. Based on these two conditions being met, I think this is a very good change! Thanks @asechrest !

asechrest:
Having trouble with Github desktop tonight for the companion PR to comab-settings. Will work on that PR tomorrow.

common-arcana.lic
```diff
@@ -518,16 +518,17 @@ module DRCA
     end
   end
 
-  def update_astral_data(data)
+  def update_astral_data(data, settings)
```
KatoakDR, line 521: 💡  As a new argument, this is a breaking  change to other scripts that may use these methods and aren't currently passing in the settings. I suggest providing a default `settings = OpenStruct.new` in each of the updated method signatures.
```suggestion
  def update_astral_data(data, settings = OpenStruct.new)
```

common-arcana.lic
```diff
@@ -518,16 +518,17 @@ module DRCA
...
-      data = set_planet_data(data)
+      data = set_planet_data(data, settings)
     end
     data
   end
 
-  def find_visible_planets(planets)
+  def find_visible_planets(planets, settings)
+    return unless settings.have_telescope
     return if DRC.bput('get my telescope', 'You get', 'What were you', 'You are already') == 'What were you'
```
KatoakDR, line 532: 💡 Go ahead and update the "getting" logic to use the equivalent DRCMM method, too
```suggestion
    DRCMM.get_telescope(settings.telescope_storage)
    return unless DRCI.in_hands?('telescope')
```

asechrest:
Converted to draft while we discuss, as this may be a breaking change.

MahtraDR:
Closing. resolved in https://github.com/rpherbig/dr-scripts/pull/5544

PR #5283, opened 2021-12-09
Fixing BG prep: line since it doesn't use target

MahtraDR:
Same deal as IZ was. This spell uses "prep" not "target"


MahtraDR:
This spell cannot be targeted.
target bg
This spell cannot be targeted.
prep bg
Since you're not feeding enough power into the spell pattern to make it coherent, you quickly work your way to the minimum required.

You trace a geometric sigil in the air, shaping the pattern of the Blufmor Garaen spell..

PR #5287, opened 2021-12-10
Fixing gryphon room numbers as per https://github.com/rpherbig/dr-scr…

MahtraDR:
…ipts/issues/4099


MahtraDR:
Note I did not test. I am just going through trying to clean up old issues that are easy/quick.

Issue #5285, opened 2021-12-10
;circlecheck is currently not calculating correctly.


Hiinky:
@asechrest Can you see what's going on?  Seems related to https://github.com/rpherbig/dr-scripts/pull/5284
The non-updated file is still working as intended.

```
>ask soth about level
"Well, SOMEONE," the guildleader says, "You have some work to do before you can advance to circle 26 in our order.

Regarding your lore knowledge, you should work on 3rd Lore (Appraisal).

Once you've accomplished that, come see me again."
```

```
>cc
--- Lich: circlecheck active.
Your next level will award: 126 TDPs
***You're ready to circle!***
You have enough 2nd Magic (Arcana) for circle 27 and need 3 (112) ranks for circle 28
```

asechrest:
Yep, will check it out. Thanks for advising, sorry about that.

asechrest:
@Hiinky 

Any chance you have a copy/paste of your `exp 0` and your guild? If the old file works, and my revised one doesn't, it's probably something I did, by if you have an `exp 0` it might help me diagnose. Thanks.

asechrest:
@Hiinky 

I will need more info from you to figure this out. While I did so some string manipulation with my edit, I didn't change the fundamental calculations for circles. Your paste of game output is not my revised file. Could you do some other cut and pastes for me?

Here is my character talking to the guildleader, running the new ;cc, and running the old ;cc. It seems to be accurate:

```
>ask tom about level
"Well, Gildaren," the Guildleader says, "You have some work to do before you can advance to circle 60.

Regarding your weapons knowledge, you should work on 1st Weapon (Small Blunt).

Once you've accomplished that, come see me again."
>;cc
--- Lich: could not find script 'cc' in directory C:/Ruby4Lich/lich/scripts
>cc
--- Lich: circlecheck active.
--- Lich: spellmonitor active.
| ________________________________________________________________________________________________________________
CURRENT CIRCLE: 59
Your next level will award: 160 TDPs
You have enough  Small Blunt     1st Weapon    for circle   59  and need     3.48   (210)  ranks for circle    60

You have enough  Light Armor     1st Armor     for circle   78  and need     0.71   (236)  ranks for circle    79
You have enough  Small Edged     2nd Weapon    for circle   80  and need     2.29   (203)  ranks for circle    81
You have enough  Perception      3rd Survival  for circle   93  and need     2.17   (390)  ranks for circle    94
[...]

Issue #2854, opened 2018-06-27
;circlecheck broken for Thieves


ghost:
>;circle
--- Lich: circlecheck active.
Your next level will award: 250 TDPs
You have enough 3rd Lore (Engineering) for circle 145 and need 15 (350) ranks for circle 150
You have enough 1st Weapon (Small Edged) for circle 149 and need 1 (620) ranks for circle 150
You have enough 4th Survival (Backstab) for circle 150 and need 6 (735) ranks for circle 151
You have enough 1st Survival (Perception) for circle 151 and need 7 (800) ranks for circle 152
You have enough 7th Survival (Thievery) for circle 151 and need 10 (596) ranks for circle 152
You have enough 2nd Survival (Stealth) for circle 151 and need 12 (760) ranks for circle 152
You have enough 3rd Survival (Skinning) for circle 151 and need 14 (750) ranks for circle 152
You have enough 5th Survival (Evasion) for circle 155 and need 4 (718) ranks for circle 156
You have enough 6th Survival (Locksmithing) for circle 155 and need 13 (688) ranks for circle 156
You have enough Inner Magic for circle 156 and need 3 (530) ranks for circle 157
You have enough Thievery for circle 156 and need 4 (590) ranks for circle 157
You have enough 1st Magic (Augmentation) for circle 165 and need 4 (530) ranks for circle 166
You have enough 2nd Weapon (Brawling) for circle 165 and need 4 (620) ranks for circle 166
You have enough 1st Armor (Light Armor) for circle 167 and need 2 (524) ranks for circle 168
You have enough 1st Lore (Tactics) for circle 169 and need 5 (660) ranks for circle 170
You have enough 2nd Magic (Debilitation) for circle 172 and need 8 (524) ranks for circle 173
You have enough 8th Survival (Athletics) for circle 172 and need 8 (554) ranks for circle 173
You have enough Stealth for circle 177 and need 2 (750) ranks for circle 178
You have enough Parry Ability for circle 183 and need 6 (642) ranks for circle 184
You have enough 2nd Lore (Appraisal) for circle 184 and need 8 (650) ranks for circle 185
--- Lich: circlecheck has exited.

info:
Gender: Male   Age: 23   Circle: 149

From olwydd:
What you need for circle: 150
You have enough 1st Weapon (Small Edged) for Circle 149 and need 1 (620) ranks for circle 150

Sarvatt:
[...]

PR #5284, opened 2021-12-10
[script][circlecheck] Prettify output

asechrest:
This revision prettifies the output of `;circlecheck`.

I took this on mainly as practice with formatting in Ruby. But I did add some additional output that I thought were nice quality of life updates. I use circle check via "cc" alias quite frequently. One thing I like to see at a glance is 1) my current circle and 2) my current progress to next rank in given skills.

So this PR does a few things:

1. Outputs your current circle
2. Defines, with yellow highlights, the beginning and end of the circle check
3. Organizes the sections in columns for easy at-a-glance interpretation
4. Changes the order of presentation of the SKILL (e.g. Small Blunt) with the category (e.g. 1st Weapon)
4. Lets you know, down to the remaining percentage, how much you have left to next level

I was able to test a number of output scenarios with the EXCEPTION of someone who is max 200 circle. If someone would offer to help me test the formatting of the output, I'd appreciate it. I think it will need some tweaking.

In a reply here I'll give some examples of old vs. new formatting. Thanks.

PR #5300, opened 2021-12-17
[script][circlecheck] Pretty output option

asechrest:
I previously submitted PR #5284 which provided a nicer formatting for `;circlecheck` output. Some folks liked it. Others weren't sure about it. Learned a good lesson to offer new things as an option, rather than forcing it on users.

This PR adds the nicer column formatting as an option, set in your yaml, but defaults to the traditional output via default setting in base.yaml. I also made some changes after seeing some formatting hiccups with other users. Specifically, the output line-wrapped for some users because of the length of the rows. Also, for users who have characters who are high level and have some skills in the sub 200 circle requirements and others beyond, the split formatting wasn't to my liking. Both of these should now be fixed. I've removed superfluous words in the columns that makes them more compact and the output is now substantially smaller width than even the traditional `;circlecheck` output. This should prevent line wrap for most users. And for users who have exceeded the circle 200 requirements, display now uses a simple plus sign to signify the amount over circle 200 reqs, and maintains the column formatting.

The new yaml flag is `circlecheck_prettyprint`. It defaults to false in base.yaml. When false, traditional output is maintained.

NOTE: There is a known issue where the Genie FE double-spaces the output. This is NOT a result of this PR. The un-revised ;circlecheck does the same thing. We're working on understanding why, but most Genie users are utilizing the built in plugin anyway.

Traditional format:

image

When the yaml `circlecheck_prettyprint` is true, new column-formatted output is displayed:

image

Displaying requirements for a target circle well above the current circle:

image

Example output from a character that has some skills sub 200 circle req, and some over it. Note this is on Genie, so it's doublespaced:

image


asechrest:
Thanks to @MahtraDR for helping test. Also tagging @KatoakDR for review.

circlecheck.lic
```diff
@@ -498,6 +498,9 @@ def main
 
   $debug_mode_cc = args.debug
 
+  # Get settings so we can check for yaml prettyprint true or false
+  @settings = get_settings
```
rpherbig, line 502: If `@settings` is only going to be used on line 516, then I don't think it should be a local var. I instead suggest moving line 516 up here and passing the `prettyprint` value into `display_requirements`.

circlecheck.lic
```diff
@@ -506,37 +509,153 @@ end
 def display_requirements(requirements, level_target, brief)
   can_circle = false
   first_circle_seen = nil
+  column_headings_displayed = false
   spacer_displayed = false
+
+  # Grab the circlecheck_prettyprint yaml setting
+  prettyprint = @settings.circlecheck_prettyprint #Set manually for testing
```
rpherbig, line 516: I'd prefer `pretty_print`

rpherbig:

Let me know if you want more 👀 on this or are ready for it to be merged.

circlecheck.lic
asechrest: Fair, I'll change this.

circlecheck.lic
asechrest: You got it.

asechrest:
> 
> Let me know if you want more 👀 on this or are ready for it to be merged.

I think we're good to merge. Thanks for reviewing.

asechrest:
@rpherbig this is ready to merge after the commit I just made. That commit addresses issue #2854 which said ;circlecheck was broken for thieves.

After quite a bit of review, we've determined via preponderance of evidence that Elanthipedia gives the wrong circle requirements for 3rd Lore for Thieves. And because the original author of ;circlecheck used Elanthipedia's numbers, we believe it is also wrong. I've checked against 3 different online circle calculators as well as checking against Genie's exp plugin.

So, my final commit updates the Thief 3rd Lore requirements, and I believe we can close #2854. Let's also close #5285 due to no further response from OP and unable to recreate the issue.

This should be ready to merge.

asechrest:
Resolves #2854 
Resolves #5285

PR #5304, opened 2021-12-18
Deprecate id: <roomnumber> syntax for theurgy_prayer_mat_room, and make it work like other _room settings

MahtraDR:
…rematurely

Prior to this small change, calling theurgy with a prayer mat defined caused it to exit prematurely:
with 

```
theurgy_use_prayer_mat: true
```
and with a theurgy_prayer_mat_room: defined prior to this fix, this happens

```
 --- Lich: error: no implicit conversion of String into Integer                                                                                                                                                                               
         theurgy:394:in `[]'                                                                                                                                                                                                                  
         theurgy:394:in `walk_to_altar_or_prayer_mat'                                                                                                                                                                                         
 --- Lich: theurgy has exited.   
```

It now no longer happens, and the script runs through to the end.

Thanks @asechrest for the help!

### DIFF
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -392,7 +392,11 @@ class TheurgyActions
     return if @mat_unrolled
 
     if @prayer_mat
-      walk_to @prayer_mat_room if @prayer_mat_room
+      if @settings.theurgy_prayer_mat_room.is_a?(Hash)
+        walk_to @prayer_mat_room['id']
+      else @settings.theurgy_prayer_mat_room.is_a?(Integer)
+        walk_to @prayer_mat_room
+      end
       stow_hands
       DRCI.get_item_safe(@prayer_mat, @theurgy_supply_container)
       bput("unroll #{@prayer_mat}",
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -392,10 +392,12 @@ class TheurgyActions
     return if @mat_unrolled
 
     if @prayer_mat
-      if @settings.theurgy_prayer_mat_room.is_a?(Hash)
+      if @settings.theurgy_prayer_mat_room.is_a?(Hash) && @settings.theurgy_prayer_mat_room.has_key?('id')
         walk_to @prayer_mat_room['id']
-      else @settings.theurgy_prayer_mat_room.is_a?(Integer)
+      elsif @settings.theurgy_prayer_mat_room.is_a?(Integer)
         walk_to @prayer_mat_room
+      else 
+        echo "theurgy_prayer_mat_room does not contain valid settings. A valid setting looks like \"theurgy_prayer_mat_room: 1900\""
       end
       stow_hands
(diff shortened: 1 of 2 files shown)


MahtraDR:
This may be a case of user ignorance on my part... the documentation  https://elanthipedia.play.net/Lich_script_repository#theurgy actually states 

```
theurgy_prayer_mat_room:
 id: 1234
```

In a way, I'm hesitant to change this, and at the same time, no other _room setting uses this format either. I'll change it to accept either convention.

MahtraDR:
Don't merge this in yet, but we may have to rename it. I see now that for example, tithing has never worked. That data set isn't in base-theurgy at all (it is in base-towns now)  and the script is still checking Mech Lore ranks for carving beads... So, I'm having a bit of an overhaul going, and testing. It's a bit of a learning opportunity, so it might not be done right away.

MahtraDR:
I'm not touching this yet until I learn how to make it work with either a hash or an array.

I don't know how many people are using this, and I don't want to break it for them, but I do also think we should use the new syntax all the other _room settings use. It seems possible to do this, I just don't know how yet. I'll figure it out...

rpherbig:
I think making it support both forms is best, and then also add a validator that the `id` form is deprecated.

MahtraDR:
Tested on my cleric.
with the old syntax
```
theurgy_prayer_mat_room:
 id: 1900
```
I get 
```
 [validate: WARNING:< theurgy_prayer_mat_room is set as a key value pair id: 1900 this is deprecated. It should be set as a room number "theurgy_prayer_mat_room: 1900"   >]                                                                  
```

with the new syntax
```
theurgy_prayer_mat_room: 1900
```
no warnings ensue. 

No need to touch base.yaml now, since with this change, the setting is named in a way that is in line with all the other _room settings...

MahtraDR:
Adding a note here so I don't forget to edit https://elanthipedia.play.net/Lich_script_repository#theurgy later to account for this change.

theurgy.lic
```diff
@@ -391,7 +392,11 @@ class TheurgyActions
     return if @mat_unrolled
 
     if @prayer_mat
-      walk_to @prayer_mat_room['id'] if @prayer_mat_room
+      if @settings.theurgy_prayer_mat_room.is_a?(Hash)
+        walk_to @prayer_mat_room['id']
+      else @settings.theurgy_prayer_mat_room.is_a?(Integer)
```
rpherbig, line 397: This should be `elsif`. As written, the conditional part is ignored.

```ruby
a = 1

if a == 2 then
	puts '2'
else a == 3
	puts '3'
end
```

Results in `3`.

-----

While you're editing this, can you add an `else` that outputs an error?

MahtraDR: Thank you! Have a look at the current iteration, it's tested, and I made the conditionals tighter, assuming they do what i think they do.

PR #5305, opened 2021-12-18
[script] [theurgy] fixing tithing and bead-carving in theurgy.lic resolves https://githu…

MahtraDR:
…b.com/rpherbig/dr-scripts/issues/3288

carve-bead.lic actually works, so we call that instead.
Likewise, tithe.lic actually works... so not reinventing the wheel (poorly) here.

Neither of these two functions could have worked before. Carve-bead was checking on Mech Lore, so it hasn't been touched in awhile... and tithing was looking for data that didn't exist in common-theurgy.

### DIFF
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -17,6 +17,7 @@ class TheurgyActions
     @hometown = @settings.hometown
     all_theurgy_data = get_data('theurgy')
     @data = all_theurgy_data[@hometown]
+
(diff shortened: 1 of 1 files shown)


MahtraDR:
To note, both of these have been tested on my end, solid 24 hours total test time on my cleric and nothing breaks that was there, and it does now tithe, and carve if conditions are met.

I just got Truffenyi, and will test that in a bit, but that doesn't look as obviously broken as these two did.

PR #5315, opened 2021-12-22
[script] [combat-trainer] paving way to support Damaris weapons

KatoakDR:
### Background
* For some upcoming changes to `combat-trainer` to support [Damaris weapons](https://elanthipedia.play.net/Category:Damaris_weapons), need methods to help identify valid weapon skills, in this specific case whirlwinding.

### Changes
* Added to `GameState` class a `determine_whirlwind_weapon_skill` method mirroring the existing `determine_doublestrike_skill` and `determine_aiming_skill` methods. In the "Damaris Weapon" PR that's forthcoming, each of these "determine" methods will be updated to filter out skills that require use of the Damaris weapon that's currently in your hand.

### Deferred to another pull request
* Also fixed `determine_doublestrike_skill` to actually pick a skill from your `doublestrike_trainables:` setting (which often is the same as your `aimable_trainables:` but now the method ensures it's picking from your list of weapons you explicitly allowlisted for doublestrikes. The other method, `doublestrikable_weapon_skill?`, already was correctly looking at `doublestrike_trainables:`


combat-trainer.lic
```diff
@@ -4516,13 +4516,13 @@ class GameState
...
   def determine_doublestrike_skill
-    options = @aiming_trainables
+    options = @doublestrike_trainables
     options -= [weapon_skill] # can't be your main hand skill, it's in your main hand
     options -= $twohanded_skills # can't be a twohanded skill, you need both hands
     options -= $aim_skills # can't use aimable weapons in your offhand
     options -= $martial_skills # shouldn't be brawling, you need a weapon in your offhand
-    echo "determine_doublestrike_skill::options: #{options}" if $debug_mode_ct
     options -= weapon_training.select { |skill, weapon| weapon == weapon_training[weapon_skill] }.keys # can't be same weapon that's equipped in your mainhand
+    echo "determine_doublestrike_skill::options: #{options}" if $debug_mode_ct
```
KatoakDR, line 4525: Swapped lines to move comment at the end so what's logged is the actual filtered list of options. Fixing my earlier mistake.

combat-trainer.lic
```diff
@@ -4516,13 +4516,13 @@ class GameState
 
   # Determine the weapon skill to equip offhand to perform a doublestrike maneuver.
   def determine_doublestrike_skill
-    options = @aiming_trainables
+    options = @doublestrike_trainables
```
rpherbig, line 4519: Is it possible some players are relying on this behavior? I.e. they've configured/customized `aiming_trainables` for doublestrike purposes and haven't configured `doublestrike_trainables`?

rpherbig:
Code changes look good aside from the one comment ✔️

combat-trainer.lic
KatoakDR: Good question, I've [asked](https://discord.com/channels/745675889622384681/745675890242879671/923284803736993803) in Discord for feedback on this change. Will follow up.

KatoakDR: On the positive side, any weapon listed in `doublestrike_trainables:` is allowable in both your main hand and offhand, even if you normally can't perform an _offhand_ attack with the weapons. Just tested on my Warrior Mage with a LB and LE. I could doublestrike with them but can't "attack left" with either.

The default config in base.yaml is:
```yaml
prioritize_maneuver_doublestrike: false

doublestrike_trainables:
  - Small Edged
  - Small Blunt
  - Large Edged
  - Large Blunt
 ```

If someone opts-in to doueblstrikes and only overrode `prioritize_maneuver_doublestrike: true`, then as long as their `weapon_training:` list includes 2 or more of those skills then they would still successfully doublestrike.

KatoakDR:
I'm going to move the Damaris weapon support to its own PR, separate from the doublestrike changes. I shouldn't have intermingled them to begin with.

rpherbig:
> I'm going to move the Damaris weapon support to its own PR, separate from the doublestrike changes. I shouldn't have intermingled them to begin with.

Thank you, this made reviewing much easier!

Issue #3152, opened 2018-09-23
DR: Favor altar not getting clean so training manager stalls


Brashton:
I was trying to get a favor and the altar didn't get cleaned twice, both times training-manager stalled and I had to restart it.  Here is the log of what happened

Room Number: 5834
<streamWindow id='main' title='Story' subtitle=" - [Eyes of the Thirteen, Eye of Hav'roth - 5834]" location='center' target='drop'/>
<streamWindow id='room' title='Room' subtitle=" - [Eyes of the Thirteen, Eye of Hav'roth - 5834]" location='center' target='drop' ifClosed='' resident='true'/>
[go2: travel time: 0:00:01]
--- Lich: go2 has exited.
[favor]>clean silver altar with water in chalice
You begin to wash a low silver altar with your holy water.
Roundtime: 10 sec.
Ethereal Shield  (2 roisaen)
You diligently work, getting between the cracks and behind the tight corners of the silver altar.
Roundtime: 10 sec.
Ethereal Shield  (2 roisaen)
Noticing that the silver altar does not look any cleaner, you furrow your brow in confusion.  What a failure.
[favor]>get my cobra primer
You get a parchment cobra primer from inside your duffel bag.
[favor]>put my cobra primer on silver altar
You reverently place a parchment cobra primer on the silver altar.
[favor]>pray
You begin to pray, kneeling before the altar.  A sense of peace and tranquility fills you as you murmur softly.
Ethereal Shield  (2 roisaen)
Ethereal Shield  (2 roisaen)
You feel a sense of shame as your offering is thrown violently from the dusty surface of the altar.

Sarvatt:
https://github.com/rpherbig/dr-scripts/blob/master/favor.lic#L72 needs to be a separate method that gets checked recursively until it succeeds instead of just trying twice

Tarhish:
If it uses a bit of holy water on failure then you could run out mid loop, but I don't know if that's true or not.  I can't seem to really trigger this.

PR #5316, opened 2021-12-22
[script] [favor] making altar cleaning a loop till the altar is completely clean. reso…

MahtraDR:
Made altar-cleaning a loop till it's really really clean, and caught the case where you run out of water cleaning too.


MahtraDR:
Also resolves https://github.com/rpherbig/dr-scripts/issues/3152 which I just noticed now, so didn't tag in my commit messages.

PR #5318, opened 2021-12-22
[script] [combat-trainer] Prioritize aborting if combat-trainer is shutting down

KatoakDR:
### Background
* When `hunting-buddy` or `combat-trainer` itself determines that it's time to bail then `combat-trainer` goes through a structured shutdown.
* The `SpellProcess`, however, was checking its "abort" conditions _after_ possibly preparing another spell to buff or to train, which would delay the shutdown process.
* My character got boo-boos 🩹 from `combat-trainer` not bailing fast enough.

### Changes
* Refactor `SpellProcess` such that the check for aborting happens _first_ instead of _last_ in its `execute` method.
* Remove the condition to abort (see new line 1261) if the character is currently casting a spell (see old line 1290). Instead, the shutdown process will release the spell and move on to get you out of the situation.

In other words,
```diff
- if game_state.finish_spell_casting? && !game_state.casting
+ if game_state.finish_spell_casting?
```

Personally, I've been running with SpellProcess refactored this way for 5+ months.


Dartellum:
Will this fix the condition that, for some reason, you arrive at the hunting location and you meet the criteria to exit combat as it is starting up?

KatoakDR:
> Will this fix the condition that, for some reason, you arrive at the hunting location and you meet the criteria to exit combat as it is starting up?

I'm not sure, what criteria/conditions/logs were people experiencing in the scenario you're referring?

What this change will do, is not try to cast a buff or spells to train magic if combat-trainer has received the signal to stop hunting.

Dartellum:
Some conditions exist where if `combat-trainer` is spinning up and `hunting-buddy` determines the criteria are met, `combat-trainer` continues even though `hunting-buddy` exists. I have logs.

rpherbig:
This change makes sense to me and the code looks good ✔️ 

I'm ready to merge this aside from the resolution of the one comment.

KatoakDR:
> Some conditions exist where if `combat-trainer` is spinning up and `hunting-buddy` determines the criteria are met, `combat-trainer` continues even though `hunting-buddy` exists. I have logs.

@Dartellum I'm not sure if what you're describing is the same or different than the scenario this PR is addressing. It might still have positive effects because character's won't try to train magic while combat-trainer is trying to stop.

Do you see any issues with moving forward with these changes, or were you just asking questions about other scenarios this might be solving?

Thanks

Hiinky:
I had a super edge case years ago related to this spell cleanup process that caused some mega frustration.  Without logs of exactly what Dart's looking at, I would guess his is probably some version of what I experienced and this may solve his issue also.

KatoakDR:
Thanks for the feedback @Hiinky 

@rpherbig I think we're good to merge this PR, thanks

Issue #4888, opened 2021-04-12
Prepare to move this repo to the dragon-realms organization


rpherbig:
I'm thinking of this process:

1. Update `dependency` to handle a fallback URL (see https://github.com/rpherbig/dr-scripts/blob/6c4f93716cbd1047f51fef1b542b91f1f01b92e0/dependency.lic#L161-L162) - the primary would be this URL, the fallback would be https://github.com/dragon-realms/dr-scripts/
2. Wait a while for the change to propagate (two weeks?)
3. Move the repo
4. Update the 'new' repo to no longer check the `rpherbig` URL

Based on the GitHub docs (https://docs.github.com/en/github/administering-a-repository/transferring-a-repository)

KatoakDR:
Woo hoo! Moving to an organization will unlock a lot more capabilities for maintaining this open source project.

One of which I'm excited about is ability to grant folks [triage access](https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-permission-levels-for-an-organization) so that they can manage issues and pull requests without write access.

rpherbig:
From @BinuDR :
> I would suggest that when we move the repo over there that we add in a samples folder into profiles and move all of the sample yamls into that so that the profiles folder can be only for base, base-empty, and all personal yamls

I like the idea

MahtraDR:
> From @BinuDR :
> 
> > I would suggest that when we move the repo over there that we add in a samples folder into profiles and move all of the sample yamls into that so that the profiles folder can be only for base, base-empty, and all personal yamls
> 
> I like it - note: it doesn't have to be part of the first PR, I just needed a place to record the idea

Any objection to the Samples folder now? I like this idea, and it's easy.

BinuDR:
> > From @BinuDR :
> > > I would suggest that when we move the repo over there that we add in a samples folder into profiles and move all of the sample yamls into that so that the profiles folder can be only for base, base-empty, and all personal yamls
> > 
> > 
> > I like it - note: it doesn't have to be part of the first PR, I just needed a place to record the idea
> 
> Any objection to the Samples folder now? I like this idea, and it's easy.

No objections from me.

I had another idea too.  Maybe we should have subfolders for each guild so it's easier for everyone to find a yaml for the guild they're looking for.

rpherbig:
Works for me

MahtraDR:
[...]

PR #5325, opened 2021-12-23
Clean profile dirs sort samples into guild folders

MahtraDR:
When/if this is merged I'll go through the tutorials and fix any broken links that arise too.

Following up on discussion here: https://github.com/rpherbig/dr-scripts/issues/4888


rpherbig:
Judging by the use of `restock_shop`, Cidemon is a Trader. Do you want to move that to `/Samples/`?

MahtraDR:
I didn't think his yaml added very much to the existing body, but it's back in now in any case. :)

rpherbig:
Hmm, something that just occurred to me: if someone is using their repo as their Lich directory _AND_ they had a character's YAML which was checked in, then won't this break that character (because their YAML will have moved)?

MahtraDR:
If theirs is one of the yamls we've moved, then yes, it will break their setup, and they'll have to keep a local copy of their yaml (which is part of the goal of this effort) in that scenario they'd have a local copy, and a copy in our repo.

rpherbig:
To avoid surprising anyone, could you ask around in Slack/Discord/Lnet/etc. to see if this impacts anyone?

Also, I think we need to add the profiles directory (i.e. that which Lich actually uses) to the `.gitignore` then - that way no one is accidentally checking in YAMLs from there.

MahtraDR:
Thinking about this more. Wouldn't this only impact people checking their yamls into the repo today? If so, the plan would be to contact them directly, and add to gitignore for any newcomers. How's that sound?

Hiinky:
Just my two cents:  I think instead of potentially creating a lot of issues needlessly, we just remove any profiles older than X months.  There are quite a few names in this last that have sold a few times and would impacted without have any idea ahead of time through no fault of their own.

MahtraDR:
Someone who bought their character wouldn't be impacted by this. No profiles are being deleted here, they're just being moved.

The worse that could happen is someone who uses exclusively the profile in this repo, finds it's missing one day, because it's been moved to the Samples folder. They copy it back, and chug along their merry way.

With a caveat that I don't think anyone actually does this. The up to date yamls are all edited, at a minimum to remove safe room IDs, anyway... So we don't have live, in-use yamls on this repo I don't think.

MahtraDR:
@rpherbig Zero bites on lnet for any feedback.

Discussions on Discord suggest this wouldn't impact anyone. dependency won't change any existing profile files, and nobody could see a problem with this.

PR #5331, opened 2021-12-24
[script] [T2] Making t2 use the common-arcana method for cyclic release

MahtraDR:
Changing from a blanket fput for releasing cyclics, to using the common-arcana method, as we prepare to change common-arcana to allow for not releasing certain cyclics.

### DIFF
--- a/t2.lic
+++ b/t2.lic
@@ -10,7 +10,7 @@ custom_require.call(%w[common drinfomon common-travel equipmanager])
 class T2
   include DRC
   include DRCT
-  include DRCA
+  # include DRCA for the ciclyc release handling at script exit
 
   def initialize
     @shutdown = false
@@ -123,7 +123,7 @@ before_dying do
     stop_script(script_name) if Script.running?(script_name)
   end
   # Don't bother releasing cyclic for NMU guilds
-  release_cyclics unless %w[Commoner Barbarian Thief].include?(DRStats.guild)
+  DRCA.release_cyclics unless %w[Commoner Barbarian Thief].include?(DRStats.guild)
 end
 
 $T2 = T2.new
--- a/t2.lic
+++ b/t2.lic
@@ -123,7 +123,7 @@ before_dying do
     stop_script(script_name) if Script.running?(script_name)
   end
   # Don't bother releasing cyclic for NMU guilds
-  DRCA.release_cyclics unless %w[Commoner Barbarian Thief].include?(DRStats.guild)
+  release_cyclics unless %w[Commoner Barbarian Thief].include?(DRStats.guild)
 end
 
 $T2 = T2.new
--- a/t2.lic
+++ b/t2.lic
@@ -10,6 +10,7 @@ custom_require.call(%w[common drinfomon common-travel equipmanager])
 class T2
   include DRC
   include DRCT
+  include DRCA
 
   def initialize
     @shutdown = false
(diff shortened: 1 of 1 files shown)


t2.lic
```diff
@@ -10,6 +10,7 @@ custom_require.call(%w[common drinfomon common-travel equipmanager])
 class T2
   include DRC
   include DRCT
+  include DRCA
```
KatoakDR, line 13:
```suggestion
```
Technically, the `include DRCA` isn't necessary because you call `release_cyclics` already qualified with `DRCA`.

In Ruby, the `include <module>` copies all the methods from the `<module>` to this class as if they had been defined in this class. So it's unnecessary bloat in the long term.

KatoakDR (commented):
Code looks good aside from my one comment

t2.lic
MahtraDR: would it be equally valid to change the release_cyclics line to read
```
release_cyclics unless %w[Commoner Barbarian Thief].include?(DRStats.guild)
```
 and keep the include?

asechrest:
Note: I took the advice of @KatoakDR (I think?) and others and I am trying to get into the habit of always using the explicit calls. For someone new to the code it makes it hugely easier to understand where the helper methods are coming from. When I'm dissecting the flow of a script I end up seeing a call to a method, doing a control-F within that script, failing to find the method definition, then searching the repository or the scripts folder for it to find out where it exists. If we always use the explicit call, we always know where it exists and can easily look it up.

In my [;tip script](https://github.com/rpherbig/dr-scripts/blob/master/tip.lic) I commented out the include lines and used them essentially just as reference, and used explicit calls to the helper methods in my code.

MahtraDR:
@KatoakDR @asechrest Thank you both for your constructive feedback. I appreciate it. I've changed it to call DRCA explicitly, and left the include commented with a note.

t2.lic
rpherbig: The `include`s scattered around the codebase was from when we decided terseness was better. So we used `include` everywhere and didn't explicitly reference modules.

Over time we decided that being explicit with where helpers came from helped new people acclimate to the codebase. No one has really taken the time to refactor the entire codebase, though.

rpherbig: Ha, and @asechrest just said the same thing below. That'll teach me to read the entire thread before responding...

t2.lic
```diff
@@ -10,6 +10,7 @@ custom_require.call(%w[common drinfomon common-travel equipmanager])
 class T2
   include DRC
   include DRCT
+  # include DRCA for the ciclyc release handling at script exit
```
rpherbig, line 13: "cyclic"

t2.lic
```diff
@@ -122,9 +123,9 @@ before_dying do
     stop_script(script_name) if Script.running?(script_name)
   end
   # Don't bother releasing cyclic for NMU guilds
-  fput('release cyclic') unless %w[Commoner Barbarian Thief].include?(DRStats.guild)
+  DRCA.release_cyclics unless %w[Commoner Barbarian Thief].include?(DRStats.guild)
```
rpherbig, line 126: Maybe not part of this PR, but _if_ we feel this check is necessary, I think it should be moved to a guard clause within `release_cyclics` itself.

rpherbig, line 126: If we're going to use functionality in `DRCA`, then it needs to be added to the `custom_require` call on line 8.

t2.lic
rpherbig: And another typo on line 5: "scrips" -> "scripts"

t2.lic
MahtraDR: We're planning on touching the function in common-arcana anyway in the near future, to make not releasing some cyclics optional. It would make sense to add the check there at that time, if you're ok with that idea.

MahtraDR:
@rpherbig Thank you so much for your detailed review. I have learned from this, and it's prompted me to modify a decades old vimrc to include spell checking. :)

t2.lic
rpherbig: Sounds great

PR #5332, opened 2021-12-24
[script] [common-arcana] [data] [base.yaml] Make cyclic spell release by scripts optional

MahtraDR:
Needs more testing, but getting started here.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -407,7 +407,7 @@ module DRCA
       .select { |_name, properties| properties['cyclic'] }
       .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
       .reject { |_name, properties| properties['skip_release'] }
-      .reject { |_name, properties| properties['harmless'] && settings.magic_cyclic_skip_release }
+      .reject { |_name, properties| properties['harmless'] }
       .map { |_name, properties| properties['abbrev'] }
       .each { |abbrev| fput("release #{abbrev}") }
   end
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -23,7 +23,7 @@ module DRCA
     failure = ['as if it hungers for more', 'Your infusion fails completely', 'You don\'t have enough harnessed mana to infuse that much', 'You have no harnessed']
 
     loop do
-      pause 5 while mana <= amount
+      pause 5 while mana <= 40
       harness_mana([amount]) if harness
       break if success.include?(DRC.bput("infuse om #{amount}", success, failure))
       pause 0.5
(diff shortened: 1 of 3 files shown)


common-arcana.lic
```diff
@@ -406,6 +406,8 @@ module DRCA
       .spell_data
       .select { |_name, properties| properties['cyclic'] }
       .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
+      .reject { |_name, properties| properties['skip_release'] }
+      .reject { |_name, properties| properties['harmless'] }
```
KatoakDR, line 410: 💡 The list of spells being filtered from come `get_data('spells')` on line 405, which does not have any per-character customizations in them. For example, if `skip_release` is a property people configure in their yaml then that property won't be set here. On the other hand, `harmless` is defined in `base-spells.yaml` so it will be useful as a filter here.

MahtraDR:
Katoak and I are discussing this. We can't just skip release for harmless spells, because we've defined harmless as spells that don't cause shock (even for non empaths...) That leaves us with either creating a new property in base-spells, "justice" maybe, ot use a local per spell toggle in buff_spells, or something of the kind. I think the latter is simpler, but the former would be a neater solution globally.

MahtraDR:
After some additional discussion with @KatoakDR and @Hiinky the current plan is to make a global toggle to drop only TM/Debil cyclics (and we need special cases for awaken forest for rangers, and guardian spirit for empaths) which would make for relatively small code changes to enable this. We'd keep the default to the current behavior of dropping all cyclics.

KatoakDR:
I haven't thought through this next question, just asking it: would it make things easier to manage if we had a new setting where players list the spell names they don't want released?

The solution is more general and the implementation doesn't need to understand nuances of the spells in base data (is it TM or Debil or harmless or justice or ...)

MahtraDR:
> I haven't thought through this next question, just asking it: would it make things easier to manage if we had a new setting where players list the spell names they don't want released?
> 
> ```yaml
> no_release_cyclic_spells:
>   - Death Monkey
>   - Keep It Up Charm
> ```
> 
> The solution is more general and the implementation doesn't need to understand nuances of the spells in base data (is it TM or Debil or harmless or justice or ...), which gives players more control over which cyclics they keep up.

This is my favorite idea. It defaults to empty, which is the current status quo anyway. It gives full flexibility as to what spells to drop, or not. And it's just one small change to one method in common-arcana, and documentation in base.yaml

Hiinky:
I think I like this option best.  Some folks won't care at all for the updates and can just ignore it but its easy enough to explain for folks that do want to opt-in.

PR #5334, opened 2021-12-25
[script] [common-money] Handle decimals when convert to copper

KatoakDR:
### Background
* `convert_to_copper` casts the amount variable to integer, then does multiplication, and returns the result.
* For expressions like '1.5 plat' this loses precision because 1.5 => 1 and then becomes 10,000 instead of 15,000.
* I came across this working on a script that parses output from `SHOP <surface>` to know the price of items. Sometimes things are listed as "X.Y {denomination}" instead of always in coppers.

### Changes
* Convert amount to float first, do multiplication, then convert to integer to truncate any remaining decimals (if any).
* Trimmed whitespace on file save.

### Tests
```
> ,e echo DRCM.convert_to_copper(1.5, 'platinum')

[exec1: 15000]
```
```
> ,e echo DRCM.convert_to_copper(1.6254, 'platinum')

[exec1: 16254]
```
```
> ,e echo DRCM.convert_to_copper(1.6254, 'gold')

[exec1: 1625]
```


asechrest:
Good call.

PR #5336, opened 2021-12-26
[script][circlecheck] - Bugfix, visual updates

asechrest:
Fixed a bug where the `pretty_print` version of `;circlecheck` reported the wrong needed ranks for next circle when the current rank was exactly 0% to next rank and at the same time made the calculation code a bit more efficient. Also some visual updates and a bit more info on TDPs - listing your current TDPs, the amount of TDPs you'll earn next level, and the resulting total.

image

Note: Still requires yaml `circlecheck_prettyprint: true` to display new format.

Here is a log of the error:

image

### DIFF
--- a/circlecheck.lic
+++ b/circlecheck.lic
@@ -542,8 +542,8 @@ def display_requirements(requirements, level_target, brief, pretty_print)
     # This outputs the beginning section of circle check, which
     # gives your current circle and your TDPs at next level.
     DRC.message("___________________________________________________________")
-    _respond "            Current Circle:  <pushBold/>#{DRStats.circle}<popBold/>" + 
-      "    |    Next Level TDPs:  #{DRStats.tdps} <pushBold/>+ #{tdps}<popBold/> = #{DRStats.tdps + tdps}"
+    _respond " Current Circle:  <pushBold/>#{DRStats.circle}<popBold/>" + 
+      " | Next Level TDPs:  #{DRStats.tdps} <pushBold/>+ #{tdps}<popBold/> = #{DRStats.tdps + tdps}"
     respond("")
     
     # Set column headings for later display for prettyprint
--- a/circlecheck.lic
+++ b/circlecheck.lic
@@ -611,7 +611,11 @@ def display_requirements(requirements, level_target, brief, pretty_print)
 
       # Formulate the needed ranks to target circle including 
       # fractional amount. Also prep total target ranks for proper display.
-      subtraction == 1.0 ? needed_ranks = missing_ranks : needed_ranks = missing_ranks - 1
+      if subtraction == 1.0
+        needed_ranks = missing_ranks
+      else
+        needed_ranks = missing_ranks - 1
+      end
       needed_ranks = needed_ranks.to_s + percent_remaining.to_s
       missing_ranks_plus_ranks = "(" + (missing_ranks + ranks).to_s + ")"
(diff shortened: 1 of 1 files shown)


asechrest:
@MahtraDR , @KatoakDR for a quick test if ya'll get a moment. Other than the bugfix, mostly just cosmetic and back-end better code now that I've learned a bit more, so not a big deal.

circlecheck.lic
```diff
@@ -542,7 +542,8 @@ def display_requirements(requirements, level_target, brief, pretty_print)
     # This outputs the beginning section of circle check, which
     # gives your current circle and your TDPs at next level.
     DRC.message("___________________________________________________________")
-    respond("CURRENT CIRCLE:" + "#{DRStats.circle}" + " | Next level TDPS: " + "#{tdps}")
+    _respond "            Current Circle:  <pushBold/>#{DRStats.circle}<popBold/>" +
```
KatoakDR, line 545: ❓ What led to using `_respond` instead of `respond`? Line 545 and 547 use different method calls.

circlecheck.lic
```diff
@@ -604,17 +605,14 @@ def display_requirements(requirements, level_target, brief, pretty_print)
...
-        percent_remaining = 00
-      end
+      # we can easily append decimal to current rank.
+      subtraction = (100 - DRSkill.getpercent(sub_skill).to_f)/100.00
+      percent_remaining = format('%.2f', subtraction).sub(/^./, '')
 
       # Formulate the needed ranks to target circle including 
       # fractional amount. Also prep total target ranks for proper display.
-      needed_ranks = (missing_ranks -1).to_s + "." + percent_remaining.to_s
+      subtraction == 1.0 ? needed_ranks = missing_ranks : needed_ranks = missing_ranks - 1
```
KatoakDR, line 614: 💡 It took me a minute to understand this syntax. Thoughts on refactoring to one of these alternatives such that there's exactly one assignment per statement and the "right hand side" is only deciding on a return value?
```ruby
      needed_ranks = (subtraction == 1.0) ? missing_ranks : missing_ranks - 1
```
```ruby
      needed_ranks = missing_ranks
      needed_ranks -= 1 if subtraction == 1.0
```

circlecheck.lic
asechrest: The honest answer is that I wanted to use `<pushBold/>` on a portion of the line's text and it didn't seem to work with `respond` vs. `_respond`. Happy to learn anything additional, though.

circlecheck.lic
asechrest: I was exploring the use of Ruby's single-line if/else syntax. It's hard for me to read.

Ok to just use a standard if/else?

```
if subtraction == 1.0
    needed_ranks = missing_ranks
else
    needed_ranks = missing_ranks - 1
end
```

circlecheck.lic
asechrest: Example: pushbold used to highlight current circle and highlight the amount of next level's TDPs to visually separate it from the current TDPs.

image

circlecheck.lic
KatoakDR: Yep, readability > syntax magic

circlecheck.lic
KatoakDR: Oh neat!

PR #5337, opened 2021-12-26
appraisal.lic use drci.get/pull for pouches

Valherun:
Appraisal.lic was failing to find full pouches contained in Eddy's.

Updated to use common-items ``get_item?`` and ``put_away_item?``

### DIFF
--- a/appraisal.lic
+++ b/appraisal.lic
@@ -77,21 +77,17 @@ class Appraisal
     end
     pause 1
     $ORDINALS.each do |ordinal|
-      case bput("get #{ordinal} #{gem_pouch_adjective} #{gem_pouch_noun} from my #{full_pouch_container}", '^You get ', '^What were you referring')
-      when /^You get /
-        case bput("appraise my #{gem_pouch_adjective} #{gem_pouch_noun} quick", 'Roundtime', /You.ll need to open the .* to examine its contents./)
-        when /You.ll need to open the .* to examine its contents./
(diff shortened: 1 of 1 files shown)


appraisal.lic
```diff
@@ -77,21 +77,17 @@ class Appraisal
...
+      when /You.ll need to open the .* to examine its contents./
+        bput("open my #{gem_pouch_adjective} #{gem_pouch_noun}", /You open your .* ./, /That is already open/)
+        bput("appraise my #{gem_pouch_adjective} #{gem_pouch_noun} quick", 'Roundtime')
+        bput("close my #{gem_pouch_adjective} #{gem_pouch_noun}", /You close your/, /That is already closed/)
       end
-      break if DRSkill.getxp('Appraisal') >= 30
+      waitrt?
+      DRCI.put_away_item?( "#{gem_pouch_adjective} #{gem_pouch_noun}", full_pouch_container )
+      pause 1
+    break if DRSkill.getxp('Appraisal') >= 30
```
KatoakDR, line 90: Thanks for the update. 

💅  Minor stylistic request that lines 80-90 be indented two more spaces so that they fall nested wtihin the each loop defined at line 79

Valherun: good call, missed that when I updated things

Dartellum: I do not believe lines 81 through 89 require two spaces indented. You are not in a specific unless loop as `break unless ...` is its own loop code line.

PR #5319, opened 2021-12-22
[script] [combat-trainer] Add Damaris Weapons support

KatoakDR:
### Dependencies
* #5318
* #5317
* #5315

### Background
* Adds support for [Damaris Weapons](https://elanthipedia.play.net/Category:Damaris_weapons), a new multi-tool weapon introduced at Corn Maze 2021.

### Changes
* Support new yaml setting: `damaris_weapon_sets` which is a hash of your weapon set name to a list of Damaris weapons that have been combined into that set. The weapon names must match adj + noun of weapons in your `weapon_training:` config.
* Refactored code that wields weapons to go through `game_state.wield_xyz` methods instead of directly to `@equipment_manager` so that Damaris weapon logic, or any wielding logic for that matter, can be processed. Code that was temporarily putting a weapon away then retrieving it again was not changed to go through the new methods. The new methods are specifically designed when needing to change to a new weapon to train.
* Supports Damaris weapons for main hand, offhand, whirlwind, aiming trainables, and doublestrike trainables.

### Known Limitations
* Damaris weapons that run out of "turnable" time may cause undesirable effects as `combat-trainer` won't be able to switch to a new weapon noun that it wants to train. The script won't try to apply infuser stones on the player's behalf. It's the player's responsibility to ensure their Damaris weapon has sufficient infusions.
* Although multiple Damaris weapon sets are supported (see example config below), `combat-trainer` does not handle if the same weapon (e.g. a Damaris mace) exists in multiple sets. It is assumed that a Damaris weapon exists in exactly one set.

_example config using one set to rule them all_
```yaml
damaris_weapon_sets:
  one_set_to_rule_them_all:
    - Damaris slings
    - Damaris shortbow
    - Damaris stonebow
    - Damaris sabre
    - Damaris broadsword
    - Damaris greatsword
    - ...

weapon_training:
  Small Edged: Damaris sabre
  Large Edged: Damaris broadsword
  Twohanded Edged: Damaris greatsword
  Slings: Damaris sling
  Bow: Damaris shortbow
  Crossbow: Damaris stonebow
```

_example config separating melee from ranged waepons_
```yaml
damaris_weapon_sets:
  melee:
    - Damaris sabre
    - Damaris broadsword
    - Damaris greatsword
    - ...
  ranged:
    - Damaris slings
    - Damaris shortbow
    - Damaris stonebow

weapon_training:
  Small Edged: Damaris sabre
  Large Edged: Damaris broadsword
  Twohanded Edged: Damaris greatsword
[...]

PR #5340, opened 2021-12-27
[script] [equipmanager] Explicitly return true/false when try to wield weapon

KatoakDR:
### Background
* Part of the road to [Damaris Weapons support](https://github.com/rpherbig/dr-scripts/pull/5319)

### Changes
* Introduce new method names that use `?` predicate convention
  - `wield_weapon?(..)`
  - `wield_weapon_offhand?(..)`
* Avoid unnecessarily trying to swap a weapon (to another skill or another hand) that never made it to your hands.
* Provide a consistent return value so calling code can make appropriate decisions.
* Use `DRC.message` instead of `echo` to log warnings/errors to player.


rpherbig:
What do you think about adding a new function with `?` in the name to follow the predicate convention? (and adding a comment to the original functions that they're deprecated)

KatoakDR:
> What do you think about adding a new function "overload" with `?` in the name to follow the predicate convention? (and adding a comment to the original functions that they're deprecated)

Sure, will do

KatoakDR:
> What do you think about adding a new function "overload" with `?` in the name to follow the predicate convention? (and adding a comment to the original functions that they're deprecated)

Resolved in https://github.com/rpherbig/dr-scripts/pull/5340/commits/7b6f8319d110846bae2766b7098b5353e4935d76

PR #5344, opened 2021-12-28
[script] [buff] Enable forcibly recasting buffs even if active

KatoakDR:
### Background
* Sometimes I want to recast a spell, even if it's active, and I want to use `buff` script because it handles the prep, cambrinth, cast, etc.
* `buff` uses `common-arcana` to cast spells, and `common-arcana` by design optimizes to not recast spells that are already active.
* As a workaround, I release the spell, wait for it to fade and for Lich to realize it's gone, then I re-cast the buff.
* I don't want to spend time waiting for the spell to release just so I can use `buff` to recast it as I fine-tune mana and cambrinth numbers.

### Changes
* 🆕 Add optional `force` argument that when set will recast spells even if they're currently active.
* 🆕 Add warning if the waggle set to cast isn't found. Previously, the script just wouldn't cast anything but didn't say why.
* 💅 Remove unnecessary modules from `custom_require` and `include`.

### Implementation Note
* The trick to set `spell_data['recast'] = 99` is borrowed from `buffother`.

### Usage
```
# Casts spells from your 'default' waggle set
# but won't recast any spell that's currently active
>,buff
```
```
# Casts spells from your 'something ' waggle set
# but won't recast any spell that's currently active
>,buff something 
```
```
# Casts spells from your 'force' waggle set
# but won't recast any spell that's currently active.
# Although the word 'force' is used, it matches the waggle set argument.
# To use the new `force` argument then both waggle set and force must be specified.
>,buff force
```
```
# Casts spells from your 'something' waggle set
# and will recast spells even if they're currently active
>,buff something force
```


bastionwt:
Why not just create a second waggle with higher-than-normal recast values? CTRL-C, CTRL-V, change the recast values from the normal 1's or 2's to 10's or 20's. 

(This is the common tactic currently employed)

KatoakDR:
> Why not just create a second waggle with higher-than-normal recast values? CTRL-C, CTRL-V, change the recast values from the normal 1's or 2's to 10's or 20's.
> 
> (This is the common tactic currently employed)

Thanks for the feedback. Although that may work I'd rather not duplicate spell definitions in my yaml across my 15 characters. Also, I define a waggle per spell so I can cast them individually, as well as define waggles to cast groups of spells. 

By having the script automatically adjust the recast value for me means I don't have to change anything about my yaml files and only additionally type "force" word to buff commands I already use. 

Let me know if you have other questions about this approach. 🙂

MahtraDR:
Tested this on mine. It's so good. Thank you! Zero problems. The code looks good too.

asechrest:
I haven't tested this yet but it's something I've wanted. I've got manual a custom spell release script I've been using to release spells and then recast. This is a good solution.

Hiinky:
I haven't tested either but I do like the idea.  I've used the "separate waggle" approach for a long time because I typically have different everything in mind when I force buffs up (primarily for pvp/events) but I can see where this could be useful for uphunting with mandatory buffs needing to be up before you arrive.

KatoakDR:
@rpherbig this is ready for your review/merge, thanks!

PR #5350, opened 2021-12-29
Add Premie Eidolon Steeds + Death Spirits

Hiinky:


Hiinky:
Added death spirits since  this is still open and its all related to the new premie thresholds being better.

PR #5355, opened 2021-12-30
[script] [bescort] Added checks for skate wearing around the ice road, so it will only g…

### LINKED ISSUE #4948: bescort ice-skate error handling
There are some unhandled edge cases in bescort for the ice road -

in this case, there was a YAML issue parsing the footwear property, so it didn't take my boots off, got an error message trying to put the skates on, and then never noticed the failure messages / falling over to switch to slow mode.

At least catching that you can't put on the skates seems like it would be an easy fix.

--- Lich: bescort active.
[bescort]>tap my skates
You tap a pair of white ice skates with fine brass buckles inside your traveler's pack.
>
[dependency: *** ERROR PARSING YAML FILE ***]
[dependency: (./scripts/profiles/Corgar-setup.yaml): did not find expected key while parsing a block mapping at line 825 column 3]
>
[dependency: *** ERROR PARSING YAML FILE ***]
[dependency: (./scripts/profiles/Corgar-setup.yaml): did not find expected key while parsing a block mapping at line 825 column 3]
[bescort]>get skates
You get a pair of white ice skates with fine brass buckles from inside your traveler's pack.
>
[bescort]>wear skates
>
You can't wear any more items like that.
>
[bescort]>se
You run southeast.
[Himineldar Shel, Mountain Road]
As night descends upon this mountainous region, falling temperatures increase the wind's bitter fury.  Blinding clouds of snow race across the slopes, disappearing into the darkness of the night.  The awesome power of nature unleashed against the mountain's solid stone is spellbinding.  Shadowy peaks, barely visible in the dark, are a reminder of the endless procession of mountains.
Obvious paths: east, northwest.
Roundtime: 2 seconds.
Room Number: 4367
>
[bescort]>e
...wait 1 seconds.
>
[bescort]>e
You had better slow down! The ice is far too treacherous for you to move as quickly as you are.
You go east.
[Himineldar Shel, Mountain Road]
Creeping across the mountains, the deepening shadows slowly engulf the trail in darkness.  Fractured rocks, the shattered bones of the mountain, tell the tale of a recent slide and lie scattered about the path.  Howling winds, strengthened by nightfall and plunging temperatures, rush down from the highest peaks with a banshee-like shrieking.  Along the horizon rise the shadowy images of majestic snowcapped mountains, standing in silent judgment in the darkness.
Obvious paths: east, west.
Roundtime: 2 seconds.
Room Number: 4419
>
[bescort]>e
>
Your excessive speed causes you to lose your footing.
 
Falling down you slide east.
[Himineldar Shel, Mountain Road]
Plunged into darkness and assaulted by brutal winds, the narrow trail slowly wends it way through the pass.  Sheer rock walls, rising to either side of the path, do nothing but increase the fury of nature.  The raging winds frequently dislodge large boulders from the cliffs above.  In the darkness, these missiles, with little warning, shatter violently on impact with the bottom of the pass and shower the trail and snow drifts with debris.
Obvious paths: northeast, west.
Some ice has hurt your right leg in the fall.
Room Number: 4420
>
[bescort]>ne
You can't do that while lying down.
>
[bescort]>stand
Your wounds are making standing more difficult.
You stand back up.

MahtraDR:
…o fast if skates are actually on... resolves #4948 , depends on #5354

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -980,12 +980,14 @@ class Bescort
           end
         else
           DRCI.wear_item?(footwear)
+          DRC.message("Couldn't put away #{footwear}, going slow")
         end
       elsif footwear && !DRCI.remove_item?(footwear)
         if DRCI.get_item?("skates") && DRCI.wear_item?("skates")
           move_fast = true
         else
           DRCI.put_away_item?("skates")
+          DRC.message("Couldn't wear skates, going slow")
         end
       end
     end
--- a/bescort.lic
+++ b/bescort.lic
@@ -971,22 +971,21 @@ class Bescort
     move_fast = false
 
     if DRCI.exists?("skates")
-      if !footwear
-        DRCI.get_item?("skates") && DRCI.wear_item?("skates")
+      if !footwear && DRCI.get_item?("skates") && DRCI.wear_item?("skates")
             move_fast = true
       elsif footwear
         if DRCI.remove_item?(footwear)
           if DRCI.put_away_item?(footwear)
             if DRCI.get_item?("skates") && DRCI.wear_item?("skates")
               move_fast = true
-            else
-              DRCI.put_away_item?("skates")
             end
           else
             DRCI.wear_item?(footwear)
           end
         else
           DRCI.put_away_item?("skates")
+          DRCI.get_item?(footwear) if footwear
+          DRCI.wear_item?(footwear) if footwear
         end
       end
     end
--- a/bescort.lic
+++ b/bescort.lic
@@ -984,6 +984,8 @@ class Bescort
       elsif !footwear
         DRCI.get_item?("skates") && DRCI.wear_item?("skates")
             move_fast = true
+      else
+        DRCI.put_away_item?("skates")
       end
     end
 
--- a/bescort.lic
+++ b/bescort.lic
@@ -969,45 +969,49 @@ class Bescort
     EquipmentManager.new.empty_hands
     footwear = get_settings.footwear
     move_fast = false
+    footwear_removed = false
+    skates_worn = false
 
     if DRCI.exists?("skates")
-      if !footwear && DRCI.get_item?("skates") && DRCI.wear_item?("skates")
-        move_fast = true
-      elsif footwear && DRCI.remove_item?(footwear)
-        if DRCI.put_away_item?(footwear)
-          if DRCI.get_item?("skates") && DRCI.wear_item?("skates")
-            move_fast = true
+
+    # If I don't have footwear, or I removed them, then they're off
+    footwear_removed = !footwear || DRCI.remove_item?(footwear)
+    DRCI.put_away_item?(footwear) # may fail, but they'll just be in our hands until the end
+  
+    # Nothing on out feet, try to get and wear the skates
+      if footwear_removed
+        if DRCI.get_item?("skates")
+          if DRCI.wear_item?("skates")
+            skates_worn = true
+          else
+            # Couldn't wear them for some reason
+            # You might be wearing something and `footwear:` config is incorrect
+            DRCI.put_away_item?("skates")
           end
-        else
-          DRCI.wear_item?(footwear)
-          DRC.message("Couldn't put away #{footwear}, going slow")
-        end
-      elsif footwear && !DRCI.remove_item?(footwear)
-        if DRCI.get_item?("skates") && DRCI.wear_item?("skates")
-          move_fast = true
-        else
-          DRCI.put_away_item?("skates")
-          DRC.message("Couldn't wear skates, going slow")
         end
       end
-    end
 
-    path.each do |movement|
-      move(movement)
-      waitrt?
-      unless move_fast
-        DRC.collect('rock') 
+      move_fast = skates_worn
+
+      path.each do |movement|
+        move(movement)
         waitrt?
+        unless move_fast
+          DRC.collect('rock') 
+          waitrt?
+        end
+      end
+
+      if skates_worn
+        DRCI.remove_item?("skates")
+        DRCI.put_away_item?("skates")
+      end
+
+      if footwear && footwear_removed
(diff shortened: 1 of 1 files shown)


MahtraDR:
Tested in both directions, including with invalid settings in my yaml. Works as expected now. If for whatever reasons skates can't be won, it'll go the slow way. Thanks to @KatoakDR for the lesson in common methods.

KatoakDR (commented):
I haven't tested this but Mahtra and I discussed the algorithm, and so the code looks good to me 😆 

It tries to remove/stow/get/wear to swap out your footwear and skates, and if it fails to do any of the operations along the way it tries to recover and re-wear your original footwear and walk slowly.

MahtraDR:
This is not ready, switching to draft for now. I've tested the case of invalid yamls, and broken yamls... but I just tested not having any footwear... and it doesn't use the skates. I am going to bed just now, but will fix tomorrow, and test again.

MahtraDR:
Could use another review. It works as expected now. Tested valid and invalid yamls, including with and without footwear. @KatoakDR

bescort.lic
```diff
@@ -966,28 +966,47 @@ class Bescort
...
+            if DRCI.get_item?("skates") && DRCI.wear_item?("skates")
+              move_fast = true
+            else
+              DRCI.put_away_item?("skates")
+            end
+          else
+            DRCI.wear_item?(footwear)
+          end
+        else
+          DRCI.put_away_item?("skates")
```
KatoakDR, line 989: I don't think this `else` block to put skates away is necessary because you wouldn't have pulled out the skates at that point.

bescort.lic
```diff
@@ -966,28 +966,47 @@ class Bescort
...
+      if !footwear
+        DRCI.get_item?("skates") && DRCI.wear_item?("skates")
+            move_fast = true
+      elsif footwear
+        if DRCI.remove_item?(footwear)
+          if DRCI.put_away_item?(footwear)
+            if DRCI.get_item?("skates") && DRCI.wear_item?("skates")
+              move_fast = true
+            else
+              DRCI.put_away_item?("skates")
```
KatoakDR, line 983: Also need to get out and wear your footwear as part of the failure recovery

MahtraDR:
@KatoakDR thank you! Change submitted. I'll test it more today, but it's looking good for now

MahtraDR:
Tested the following whilst having skates

wearing footwear, as defined in the yaml > works goes fast
wearing footwear, not defined in the yaml > works goes slow
not wearing footwear > goes fast
not wearing footwear, but it's defined in the yaml > works, goes fast, puts footwear back on
for whatever reason can't wear skates > works, goes slow

bescort.lic
```diff
@@ -966,28 +966,46 @@ class Bescort
...
-      path.each do |movement|
-        move(movement)
+    EquipmentManager.new.empty_hands
+    footwear = get_settings.footwear
+    move_fast = false
+
+    if DRCI.exists?("skates")
+      if !footwear && DRCI.get_item?("skates") && DRCI.wear_item?("skates")
+        move_fast = true
+      elsif footwear && DRCI.remove_item?(footwear)
```
rpherbig, line 976: If I'm reading this right, then we remove the footwear here. If any of the next steps (put away footwear, get skates, wear skates) fail, we never wear the footwear again (since `move_fast` is never set to true). Is that right?

bescort.lic
```diff
@@ -966,28 +966,46 @@ class Bescort
...
-        move(movement)
-        waitrt?
-        collect('rock')
+    end
+
+    path.each do |movement|
+      move(movement)
+      waitrt?
+      unless move_fast
+        collect('rock')
```
rpherbig, line 997: Could you change this to `DRC.collect('rock')`?

bescort.lic
MahtraDR: Actually, I thought this would be caught by the else, or do I need a local else here?

rpherbig: If `footwear` isn't defined _OR_ removing it fails, then control moves to the next `elsif` (line 982). However, once those two succeed and control moves into _THIS_ `elsif` (line 976), control will never go into the other `elsif` or `else` blocks. They're all mutually exclusive.

To do what you're describing, we could add the conditionals on lines 977 and 978 to the `elsif` on line 976 (and make a similar refactor to the other `elseif` with a nested `if` on line 982). If we're relying on the `else`, then the cleanup there needs to be very flexible/general to handle the different states we could be in.

MahtraDR: I ended up testing, and I see what you mean now. I think the current version catches oddities well.

rpherbig: I still think this isn't catching all of the cases. For example, on line 978 if getting skates succeeds but wearing them fails, then we never set `move_fast` to true (which means the cleanup on line 1002 never happens). But we still have skates in our hands.

rpherbig: > I don't think I have a preference as long as both handle each state.

I thought about this more. Stylistically, I have a strong preference for the merged `if`s with a single general-purpose recovery/cleanup `else`. I'm a bit concerned about the approach though, because my gut says there could be edge cases (what if the user has two boots with the same name?).

MahtraDR: Let me think on it more. 

The goal here was to not die on the road, which this achieves now. 

I wouldn't worry about two boots named the same, because that problem exists today. If we keep that specific problem, but make the overall situation better, I'd call that progress. Does that sound reasonable?

rpherbig:
Good point. Works for me.

MahtraDR:
@rpherbig Sorry about the hassle. Completely reworked that section with @asechrest and @KatoakDR 's help and advice. Tested all kinds of scenarios now, including containers closed, or not existing, etc... broke my poor character's leg a few times, but this version looks good now. :)

rpherbig (approved):

PR #5356, opened 2021-12-30
[data] [base-hunting] Removing 8355 because the room description causes a loot action on th…

MahtraDR:
…e word stones

As reported on Discord.

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -3168,7 +3168,7 @@ hunting_zones:
   - 8325
   # https://elanthipedia.play.net/Suw_bizar                                70-90
   suw_bizar:
-  - 8355
+# - 8355 # because room has "stones" in the description
   - 8364
   - 8362
   - 8360


Hiinky:
Just add at a note to use a loot sub?  There's a fair number of rooms with this issue game-wide and the yaml is probably the most flexible fix.
```
loot_subtractions:
- stones
```

rpherbig:
I'm inclined to go with the note about `loot_subtractions`. It does concern me a bit that it could cause problem with someone using stones (people still use those for slings, right?).

Any thoughts @KatoakDR @asechrest @Valherun ?

MahtraDR:
Note I'm happy to go with stones. The only reason I went with commenting that room is that it's not at all a busy area. No action is needed for someone who in all likelihood will be the only person in the zone, and in passing at that, since the rank ranger here is relatively narrow. Had it been 2, or even 4 rooms only, I wouldn't remove one. With 8 available rooms... I'm happy with a blunt fix (or any fix)

rpherbig:
Also a good point. Let's go with the quick fix and merge this so the immediate problem is dealt with. If we decide to bring it back in we can add a comment - and I think it's a good fit for a validator, too.

PR #5359, opened 2021-12-31
Add toggle to release cyclic spells on low mana

Hiinky:
Low mana can seriously hinder training or survivability.  One of the biggest offenders for bad mana are mana-heavy cyclics.  This feature allows to the user to toggle off the drain.

-  Updated to combat-trainer training section.  
  -  Chose this location because it triggers often and is the section most impacted by low mana.
-  Add default values to base.
  -  `release_cyclic_on_low_mana` default toggle is false to mimic current behavior.
  -  `release_cyclic_threshold` default value set to 30 to allow some margin of flexibility of account for the initial mana cost of a buff or training spell.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1583,12 +1583,8 @@ class SpellProcess
     return if game_state.casting
     return unless @training_spells
     return if @training_spells_max_threshold && game_state.npcs.length > @training_spells_max_threshold
-    if DRStats.mana < @release_cyclic_threshold && @release_cyclic_on_low_mana == true
-      release_cyclics
-      return
-    elsif DRStats.mana < @training_spell_mana_threshold 
-      return      
-    end
+    release_cyclics if @release_cyclic_on_low_mana && DRStats.mana < @release_cyclic_threshold
+    return if DRStats.mana < @training_spell_mana_threshold 
 
     needs_training = ['Warding', 'Utility', 'Augmentation', 'Sorcery', 'Debilitation', 'Targeted Magic']
                      .select { |skill| @training_spells[skill] }
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1159,6 +1159,12 @@ class SpellProcess
 
     @training_spell_mana_threshold = settings.training_spell_mana_threshold
     echo("  @training_spell_mana_threshold: #{@training_spell_mana_threshold}") if $debug_mode_ct
+    
+    @release_cyclic_on_low_mana = settings.release_cyclic_on_low_mana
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -1577,7 +1583,12 @@ class SpellProcess
     return if game_state.casting
     return unless @training_spells
     return if @training_spells_max_threshold && game_state.npcs.length > @training_spells_max_threshold
-    return if DRStats.mana < @training_spell_mana_threshold
+    if DRStats.mana < @release_cyclic_threshold && @release_cyclic_on_low_mana == true
+      release_cyclics
+      return
```
KatoakDR, line 1588: Should this only apply if you actively have a cyclic up?

Are you seeing combat-trainer spam `release cyclic` over and over in the event loop while waiting for your mana to increase above the threshold?

KatoakDR (commented):
I'm looking forward to this on my baby warmie; he runs out of mana quickly when electrostatic eddy is up

combat-trainer.lic
Hiinky: release_cyclic method checks if its up or not.  It does not spam. It was tested.  The only time you see it try it more than once is in the seconds after its released *and* you have nothing else to do so it immediately returns to training spells section before everything actually recognizes the release. (Like I was doing sitting in my safe room with no mobs and only two spells to cast.)   In active combat it should release and just move to the next action in queue.

combat-trainer.lic
```diff
@@ -1577,7 +1583,12 @@ class SpellProcess
     return if game_state.casting
     return unless @training_spells
     return if @training_spells_max_threshold && game_state.npcs.length > @training_spells_max_threshold
-    return if DRStats.mana < @training_spell_mana_threshold
+    if DRStats.mana < @release_cyclic_threshold && @release_cyclic_on_low_mana == true
```
rpherbig, line 1586:
```suggestion
    if @release_cyclic_on_low_mana && DRStats.mana < @release_cyclic_threshold
```

combat-trainer.lic
```diff
@@ -1577,7 +1583,12 @@ class SpellProcess
...
+      release_cyclics
+      return
+    elsif DRStats.mana < @training_spell_mana_threshold
```
rpherbig, line 1589: Why change this to `elsif` instead of keeping the `return if...` guard clause? If I'm understanding this correctly the two conditionals aren't really related to each other (aside from both being "check for types of low mana"), and this seems to artificially couple them.

Hiinky: True.  I think I ended up there while contemplating different tests.  This does work in both cases but I agree they're not necessarily mutually exclusive.

Hiinky:
Fixed!

PR #5364, opened 2022-01-02
[script][pick.lic] Add option to pick while standing

asechrest:
[Requested by](https://discordapp.com/channels/745675889622384681/912127186419482664/926937134374477834) B-Money420 in Discord. Adds an option to pick while standing.

Example use:

```
;pick stand
```

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -36,13 +36,9 @@ class Pick
       exit
     end
 
-    if args.stand
-      @stand = true
-    else
-      @stand = false
-    end
-
     check_deprecated_settings
+
+    @stand = args.stand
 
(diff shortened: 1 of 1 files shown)


pick.lic
```diff
@@ -35,6 +36,12 @@ class Pick
       exit
     end
 
+    if args.stand
```
KatoakDR, line 39: This can be a one liner: 
```
@stand = args.stand
```
If the arg is not nil and not false then will be treated "truthy".

asechrest: Ah, you just taught me this, too. Shall fix.

PR #5369, opened 2022-01-03
[script][theurgy] fix pouring wine on prayer rug to actually bless the wine...

MahtraDR:
Crude fix for now to make the pouring of wine actually work, since it needs to be blessed. The whole file needs attention still...

### DIFF
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -744,11 +744,24 @@ class TheurgyActions
 
   def wine
     walk_to_altar_or_prayer_mat
-    DRCI.get_item_safe('wine', @theurgy_supply_container)
-    if @prayer_mat
-      fix_standing
-      bput("kneel #{@prayer_mat}", 'You humbly kneel')
+    if DRCI.get_item?("wine", @theurgy_supply_container)
+      if @prayer_mat
+        fix_standing
+        bput("kneel #{@prayer_mat}", 'You humbly kneel')
+      else
+        return
+      end
+    else
+      return
     end
+
+    complete_or_interrupt_research
+    cast_spell({ 'abbrev' => 'bless',
+                 'mana' => 1,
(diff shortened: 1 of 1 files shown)


theurgy.lic
```diff
@@ -744,11 +744,24 @@ class TheurgyActions
 
   def wine
     walk_to_altar_or_prayer_mat
-    DRCI.get_item_safe('wine', @theurgy_supply_container)
-    if @prayer_mat
-      fix_standing
-      bput("kneel #{@prayer_mat}", 'You humbly kneel')
+    if DRCI.get_item?("wine", @theurgy_supply_container)
```
rpherbig, line 747: I'm a big fan of minimizing indentation. A good way to do that is using guard clauses. For example:

```suggestion
    return unless DRCI.get_item?("wine", @theurgy_supply_container)
```

theurgy.lic
```diff
@@ -744,11 +744,24 @@ class TheurgyActions
...
+      if @prayer_mat
```
rpherbig, line 748: The old code would try to use a prayer mat if it was defined, but continue if not.

The new code returns if no prayer mat is defined. Is that intended?

theurgy.lic
```diff
@@ -744,11 +744,24 @@ class TheurgyActions
...
-      fix_standing
-      bput("kneel #{@prayer_mat}", 'You humbly kneel')
+    return unless DRCI.get_item?("wine", @theurgy_supply_container)
+      if @prayer_mat
+        fix_standing
+        bput("kneel #{@prayer_mat}", 'You humbly kneel')
+      else
+        return
+      end
+    else
```
rpherbig, line 754: I think this `else` should have been removed from the `return unless` was refactored

PR #5372, opened 2022-01-04
[script] [common-items] get gem nouns in container

KatoakDR:
### Background
* Helpful method to identify gems in a container 😄 

### Changes
* Expose method that returns nouns from container rummaged for gems via `rummage /G` flag
* Trim whitespace on file save

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -470,6 +470,10 @@ module DRCI
     DRC.rummage('B', container)
   end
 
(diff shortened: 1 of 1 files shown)


KatoakDR:
Actually, upon closer inspection, `DRC.get_gems(container)` already exists as a wrapper to `DRC.rummage('G', container)`.

Closing

PR #5383, opened 2022-01-07
Add optional prayer_mat_storage

Hiinky:
A prayer mat is significantly larger and more valuable than most theurgy items.  This optional field defaults to the same theurgy container or allows the user to define a specific container for added security/QOL.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2423,8 +2423,8 @@ class TrainerProcess
     @prayer_mat = settings.prayer_mat
     echo("  @prayer_mat: #{@prayer_mat}") if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -2423,6 +2423,9 @@ class TrainerProcess
     @prayer_mat = settings.prayer_mat
     echo("  @prayer_mat: #{@prayer_mat}") if $debug_mode_ct
 
+    @prayer_mat_storage = settings.prayer_mat_storage || settings.theurgy_supply_container
```
rpherbig, line 2426: `prayer_mat_storage` needs to be in `base.yaml`

Hiinky: Fixed!  And updated everywhere to be prayer_mat_container to match the theurgy_supply_container.

PR #5384, opened 2022-01-07
Update to allow Forage Toggle

Hiinky:
With new foraging items released recently, it much more effective to train with `forage` instead of collect but it is largely dependent on a trash bin nearby.  It is also quite handy to be able to change your room/item/method on the fly based on where you might be so it can be set by yaml or by args.  All args work independently or together in any combination so long as they're sent in correct order.
Defaults mirror existing behavior
```yaml
# Define as `forage` to override outdoorsmanship gathering method.  Leave blank to collect.
outdoors_method: 
# If you choose forage method, you'll want a room with a trash bin!  Leave blank to not set a default room.
outdoors_room: 
# If you choose forage method, you'll want to select a custom item for best learning rate.
forage_item: rock
```

### DIFF
--- a/outdoorsmanship.lic
+++ b/outdoorsmanship.lic
@@ -52,7 +52,7 @@ class Outdoorsmanship
     start_exp = DRSkill.getxp(@skill_name)
     @end_exp = [start_exp + @targetxp, 34].min
 
-    DRCT.walk_to(@outdoors_room) if @outdoors_room
+    DRCT.walk_to(@outdoors_room)
     train_outdoorsmanship
   end
 
--- a/outdoorsmanship.lic
+++ b/outdoorsmanship.lic
@@ -27,8 +27,6 @@ class Outdoorsmanship
     @training_spells = @settings.crafting_training_spells
     @targetxp = 4
     @targetxp = args.mindstates.to_i if args.mindstates
-    @worn_trashcan = @settings.worn_trashcan
-    @worn_trashcan_verb = @settings.worn_trashcan_verb
     @outdoors_room = @settings.outdoors_room
     if args.room
       room = args.room.split('=')[1]
@@ -64,7 +62,7 @@ class Outdoorsmanship
       if DRSkill.getrank('Outdoorsmanship') < 20 || @train_method == 'forage' 
         crafting_magic_routine(@settings)
         forage?(@forage_item)
-        DRCI.dispose_trash(@forage_item, @worn_trashcan, @worn_trashcan_verb)
+        DRCI.dispose_trash(@forage_item)
       else
         item = @forage_item
         crafting_magic_routine(@settings)
--- a/outdoorsmanship.lic
+++ b/outdoorsmanship.lic
@@ -14,9 +14,10 @@ class Outdoorsmanship
       [
         { name: 'perception', regex: /perception/, optional: true, description: 'Optional: Check Perception skill in place of Outdoorsmanship' },
         { name: 'mindstates', regex: /^\d+/, optional: true, description: 'Optional: Number of mindstates or collection attempts to perform before exiting. Defaults to 3' },
-        { name: 'room', regex: /(room=(\d+))/, optional: true, description: 'Optional: Specific room to forage in - primarily used to locate a trash bin. Syntax:  room=1234'  },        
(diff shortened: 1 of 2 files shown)


Hiinky:
Pretty sure this is ready to go but there are so many combinations of input, I am having folks test it for a day or two to make sure I haven't missed anything major.

outdoorsmanship.lic
```diff
@@ -24,30 +26,42 @@ class Outdoorsmanship
...
       @skill_name = 'Perception'
     else
       @skill_name = 'Outdoorsmanship'
     end
+
     start_exp = DRSkill.getxp(@skill_name)
-    @end_exp = [start_exp + @targetxp, 29].min
+    @end_exp = [start_exp + @targetxp, 34].min
 
+    DRCT.walk_to(@outdoors_room) if @outdoors_room
```
rpherbig, line 54: We should really just put this `if` check in `walk_to` at some point... I feel like we do this guard in a lot of places

asechrest: It might work as-is? If I'm understanding this line correctly, It walks to `outdoors_room` if one is set in the yaml. If we left the `if` off of this, it would call `walk_to(nil)`.  And `DRCT.walk_to` returns false if the target room is nil. So it should do nothing and just continue the script in your current room, I think, giving the same result.

Hiinky: Accurate.  Leaving it blank works fine.  Tested!

Hiinky:
Should be all set on this one.  I've been nagging folks to try and break it for several days and so far we're all systems go.

outdoorsmanship.lic
```diff
@@ -12,9 +12,12 @@ class Outdoorsmanship
...
-        { name: 'mindstates', regex: /\d+/, optional: true, description: 'Number of mindstates or collection attempts to perform before exiting. Defaults to 3' },
-        { name: 'collect_item', regex: /\w+/i, optional: true, description: 'Item to collect. Will use forage_item if not set' }
+        { name: 'perception', regex: /perception/, optional: true, description: 'Optional: Check Perception skill in place of Outdoorsmanship' },
+        { name: 'mindstates', regex: /^\d+/, optional: true, description: 'Optional: Number of mindstates or collection attempts to perform before exiting. Defaults to 3' },
+        { name: 'room', regex: /(room=(\d+))/, optional: true, description: 'Optional: Specific room to forage in - primarily used to locate a trash bin. Syntax:  room=1234'  }, 
+        { name: 'forage', regex: /forage/, optional: true, description: 'Optional: Default method is collect, specify forage to override. Forage is best used with a trash bin.' },
+        { name: 'collect_item', regex: /\w+/i, optional: true, description: 'Optional: Item to collect. Will use forage_item if not set. Wrap "multiple words" in double quotes.' }
+
```
KatoakDR, line 20: 🧹 deleting extra blank line
```suggestion
```

KatoakDR (commented):
Code looks good to me, just the one comment about an extra blank line.

I tested a tiny bit:
```
;outdoors
;outdoors room=123 stick
;outdoors room=123 stick 2
;outdoors room=123 forage 2 stick
```
I did not try the new yaml settings though

outdoorsmanship.lic
```diff
@@ -24,30 +27,44 @@ class Outdoorsmanship
...
 
+    DRCT.walk_to(@outdoors_room)
     train_outdoorsmanship
   end
 
   def train_outdoorsmanship
     attempt = 0 # Failsafe counter to prevent an infinite loop if item isn't collectable in the room
     while (DRSkill.getxp(@skill_name) < @end_exp) && (attempt < @targetxp)
-      if DRSkill.getrank('Outdoorsmanship') < 20
+      if DRSkill.getrank('Outdoorsmanship') < 20 || @train_method == 'forage'
```
KatoakDR, line 64: 💡 Thoughts on moving `DRSkill.getrank('Outdoorsmanship') < 20` up to where `@train_method` is being defined so that the skill check is next to the same skill check that decides the `@forage_item` should be rock.

Then here the IF statement would be
```ruby
if @train_method == 'forage'
    ... # forage
else
   ... # collect
end
```

PR #5388, opened 2022-01-08
Add Grey List of traps to always pick careful

Hiinky:
Some traps have a very high mortality rate, no matter the difficulty identified.  This grey list allows the user to define specific traps to always disarm careful.
```yaml
pick:
  debug: false
  trash_empty_boxes: true
  pick_quick_threshold: 2
  pick_normal_threshold: 4
  pick_careful_threshold: 6
  disarm_quick_threshold: 1
  disarm_normal_threshold: 3
  disarm_careful_threshold: 5
  disarm_too_hard_threshold: 8
  trap_blacklist:
  trap_greylist:  [teleport, shrapnel, shocker, concussion, lightning, boomer]  
  picking_box_sources: [lootpouch,backpack]
```

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -364,8 +364,8 @@ class Pick
 
         while holding_box?(current_box) && current_box['trapped'] && !current_box['trap_difficulty'].nil?
           if current_box['trap'] && @trap_greylist.include?(current_box['trap'])
-            current_box['trap_difficulty'] = 17
-            #17 is one above the player input range to block potential yaml setting interference
+            current_box['trap_difficulty'] = 18
+            #18 is one above the player range to block potential yaml setting interference
           end        
           disarm_trap(current_box)
         end
--- a/pick.lic
+++ b/pick.lic
@@ -364,8 +364,8 @@ class Pick
 
         while holding_box?(current_box) && current_box['trapped'] && !current_box['trap_difficulty'].nil?
           if current_box['trap'] && @trap_greylist.include?(current_box['trap'])
-            current_box['trap_difficulty'] = 18
-            #18 is one above the player range to block potential yaml setting interference
+            # force disarm careful on specific trap_greylist of traps types.
+            current_box['trap_difficulty'] = @disarm_careful_threshold
           end        
           disarm_trap(current_box)
         end
--- a/pick.lic
+++ b/pick.lic
@@ -90,6 +90,8 @@ class Pick
 
     @trap_blacklist = @settings.pick['trap_blacklist'] || []
     @blacklist_container = @settings.pick['blacklist_container']
+    
+    @trap_greylist = @settings.pick['trap_greylist'] || []
 
     @all_trap_messages = picking_data['traps']
     @disarmed_trap_messages = picking_data['disarmed_traps']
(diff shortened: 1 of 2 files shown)


Hiinky:
Much appreciation to @wstampley for the lovely pick re-write that made this so simple to add.

pick.lic
```diff
@@ -360,6 +363,10 @@ class Pick
         end
 
         while holding_box?(current_box) && current_box['trapped'] && !current_box['trap_difficulty'].nil?
+          if current_box['trap'] && @trap_greylist.include?(current_box['trap'])
+            current_box['trap_difficulty'] = 18
+            #18 is one above the player range to block potential yaml setting interference
+          end
```
KatoakDR, line 369: As I was trying to understand why there was a loop within a loop disarming a box, I remembered that boxes can have multiple traps to disarm. Got it, so we're iterating over each possible trap. So I now understand why your greylist check is within this loop so that even if the 1st trap wasn't in the greylist, the 2nd or 3rd trap _might_ be.

So that got me thinking about lines 353-363 that blacklist boxes if the first trap's type or difficulty meet criteria. I'm wondering if we should also move lines 353-363 to be within the same loop as where you're putting the greylist handling, too.

Thoughts?

image

KatoakDR: To avoid the hard-coded magic number `18`, I'm proposing that we either use the `@disarm_careful_threshold` variable which the method `disarm_trap` uses to know to disarm careful, or use a sufficiently large symbolic number that has the connotation of just "yep, this is big and important" like `999` or `Float::Infinity`

```suggestion
          if current_box['trap'] && @trap_greylist.include?(current_box['trap'])
            # The trap may or may not be too difficulty for your skill
            # but the consequence of failure is higher than usual,
            # let's take it safe just in case.
            current_box['trap_difficulty'] = @disarm_careful_threshold
          end        
```

image

Hiinky: > So that got me thinking about lines 353-363 that blacklist boxes if the first trap's type or difficulty meet criteria. I'm wondering if we should also move lines 353-363 to be within the same loop as where you're putting the greylist handling, too, since if the 1st trap doesn't meet blacklist criteria, the 2nd or 3rd trap _might_.

I believe this section is only identifying the speed for this one action and another method (disarm_trap) is actually doing the disarming and making a new determination based on the results of that method.   To me, its easier to see/read in separate sections the way it was done originally. I only want to set it to careful once that box has clearly been "approved" by nesting this check within the "yes, lets disarm this box" and it's most obvious by leaving the too hard sections as they are.

>To avoid the hard-coded magic number 18, I'm proposing that we either use the @disarm_careful_threshold variable which the method disarm_trap uses to know to disarm careful, or use a sufficiently large symbolic number that has the connotation of just "yep, this is big and important" like 999 or Float::Infinity

Setting it to @disarm_careful_threshold probably makes the most sense.  Future proofing. :)

KatoakDR: Gotcha, thanks

rpherbig:
Changes look good to me ✔️ 

Let me know when it's ready to be merged

Hiinky:
I think we're all set!  I've had a few people running in addition to my girls and no issues.

PR #5390, opened 2022-01-08
Add Commune Selection + bug fixes

Hiinky:
-- Add communes field to allow user to choose which communes to use.  Its very easy to bottom out devotion with badly timed communes.
-- Properly return water holder to theurgy container after checking so stow_hands can't accidentally stow it in the wrong place.
-- Add handling for flint in an eddy (must be in your hand to light incense) versus in a normal container (does not need to be in-hand)
-- Add handling for performance/play interfering with altar cleaning.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -1282,6 +1282,7 @@ theurgy_supply_container: backpack
 #    min: 3
 #    target: 4
 theurgy_supply_levels:
+communes: [Eluned Truffenyi Kertigen Tamsine]
 
 tithe: false
 tithe_almsbox:
--- a/theurgy.lic
+++ b/theurgy.lic
@@ -17,11 +17,9 @@ class TheurgyActions
     @hometown = @settings.hometown
     all_theurgy_data = get_data('theurgy')
     @data = all_theurgy_data[@hometown]
-
     @immortal_to_aspect = all_theurgy_data['immortal_to_aspect']
     @safe_room = @settings.safe_room
     @immortal_aspect = @settings.immortal_aspect
-
     @bag = @settings.crafting_container
     @bag_items = @settings.crafting_items_in_container
     @belt = @settings.engineering_belt
@@ -29,12 +27,11 @@ class TheurgyActions
     @prayer_mat_container = @settings.prayer_mat_container || @settings.theurgy_supply_container
     @water_holder = @settings.water_holder
     @flint_lighter = @settings.flint_lighter
-
-    (@prayer_mat = @settings.prayer_mat) if @settings.theurgy_use_prayer_mat
-    @prayer_mat_room = @settings.theurgy_prayer_mat_room
-
+    @prayer_mat = @settings.prayer_mat if @settings.theurgy_use_prayer_mat
+    @prayer_mat_room = @settings.theurgy_prayer_mat_room    
     @rituals = get_rituals
     @communes = get_communes
+    @chosen_communes = @settings.communes
 
     @item_info = collect_item_info(@rituals + @communes)
 
@@ -45,7 +42,6 @@ class TheurgyActions
       c[:items].any? { |i| !i[:on_hand] && !i[:shop] }
     end
   end
-
 
   def get_rituals
     rituals = all_rituals
@@ -73,7 +69,6 @@ class TheurgyActions
         method: :refectory
       }
     end
-
     if DRSkill.getrank('Engineering') > 140 && @immortal_aspect && @known_spells.include?('Bless')
       rituals << {
         method: :carve_bead
@@ -161,13 +156,13 @@ class TheurgyActions
 
   def get_communes
     communes = []
-    if DRStats.circle > 3 && DRSkill.getrank('Outdoorsmanship') > 20 && @data['dirt_foraging']
+    if DRStats.circle > 3 && DRSkill.getrank('Outdoorsmanship') > 20 && @data['dirt_foraging'] && @chosen_communes.include?('Eluned')
       communes << {
         name: :eluned,
         method: :commune_eluned
       }
     end
-    if DRStats.circle > 60
+    if DRStats.circle > 60 && @chosen_communes.include?('Truffenyi')
       truffenyi_commune = {
         name: :truffenyi,
         method: :commune_truffenyi,
@@ -181,13 +176,13 @@ class TheurgyActions
       end
       communes << truffenyi_commune
     end
-    if DRStats.circle > 2
+    if DRStats.circle > 2  && @chosen_communes.include?('Tamsine')
       communes << {
         name: :tamsine,
         method: :commune_tamsine
       }
     end
-    if DRStats.circle > 8
+    if DRStats.circle > 8  && @chosen_communes.include?('Kertigen')
       communes << {
         name: :kertigen,
         method: :commune_kertigen,
@@ -256,6 +251,7 @@ class TheurgyActions
   def holy_water?
     DRCI.get_item_safe(@water_holder, @theurgy_supply_container)
     DRCI.inside?('holy water', @water_holder)
+    DRCI.put_away_item?(@water_holder, @theurgy_supply_container)
   end
 
   def buy_supplies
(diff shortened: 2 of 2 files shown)


profiles/base.yaml
```diff
@@ -1282,6 +1282,7 @@ theurgy_supply_container: backpack
 #    min: 3
 #    target: 4
 theurgy_supply_levels:
+communes: [Eluned Truffenyi Kertigen Tamsine]
```
rpherbig, line 1285:
```suggestion
communes:
- Eluned
- Truffenyi
- Kertigen
- Tamsine
```

theurgy.lic
```diff
@@ -729,12 +725,24 @@ class TheurgyActions
...
+      dropped = true
+    end
 
     waitrt? while bput('light my incense with flint',
                        'nothing happens',
                        'bursts into flames',
                        'much too dark in here to do that') == 'nothing happens'
     waitrt?
+
+    if dropped == true
```
rpherbig, line 741:
```suggestion
    if dropped
```

theurgy.lic
```diff
@@ -729,12 +725,24 @@ class TheurgyActions
     else
       DRCI.get_item_safe('incense', @theurgy_supply_container)
     end
+    
+    unless exists?('flint')
+      bput('drop my incense','You')
```
rpherbig, line 730: Can we lower to feet instead of dropping? Or is there something else safer than dropping?

theurgy.lic
```diff
@@ -729,12 +725,24 @@ class TheurgyActions
     else
       DRCI.get_item_safe('incense', @theurgy_supply_container)
     end
+    
+    unless exists?('flint')
```
rpherbig, line 729: This will attempt to tap 'flint'. When would this return `false` and then line 731 would succeed?

theurgy.lic
```diff
@@ -776,7 +782,11 @@ class TheurgyActions
     return false unless holy_water?
     safe_walk_to @data['altar']['id']
     DRCI.get_item_safe(@water_holder, @theurgy_supply_container)
-    fput 'clean altar with holy water'
+    case bput('clean altar with holy water','Roundtime','You are a bit too busy performing to do that')
+    when 'You are a bit too busy performing to do that'
```
rpherbig, line 786: Doesn't `bput` already handle this?

See https://github.com/rpherbig/dr-scripts/blob/master/common.lic#L164

KatoakDR:
@Hiinky Did you mean to commit `things` and `things2` in this PR about communes?

theurgy.lic
Hiinky: We can but not really any major reason to?  Incense only costs a couple copper and you use up/buy it constantly.  One fail due to missing incense (I've never had it disappear in that 1 sec its on the ground) would be fixed on the very next running because you check supplies again.  Lowering to ground won't break anything but it could potentially cause a slight delay in leaving a room due to things atfeet.

theurgy.lic
Hiinky: If flint is in your eddy/portal its not on your person so you can't just tap thing, you have to tap thing in your eddy.  (Think portable wormhole to a vault.  You can reach through the eddy to a vault somewhere else but only if you specify). If you can just tap (because its in a normal storage) you don't have to pick it up.  If you can't because its in your eddy, you need to grab it with get_item_safe which specifies portal.

theurgy.lic
Hiinky: Yup!  I forgot entirely.  I just get mega annoyed when somehow I'm still playing a song when I get to the altar and pause forever because fput/waitfor is broken by obnoxious cowbell.  Fixed!

Hiinky:
Ready for follow-up review!

theurgy.lic
```diff
@@ -256,6 +251,7 @@ class TheurgyActions
   def holy_water?
     DRCI.get_item_safe(@water_holder, @theurgy_supply_container)
     DRCI.inside?('holy water', @water_holder)
+    DRCI.put_away_item?(@water_holder, @theurgy_supply_container)
```
rpherbig, line 254: I think this changes the meaning of `holy_water?`. Instead of returning whether or not `holy water` was found inside `@water_holder`, doesn't it now return whether or not `@water_holder` could be returned to `@theurgy_supply_container`?

theurgy.lic
rpherbig: Could you add a comment that this is to work around eddies?

theurgy.lic
rpherbig: I guess it doesn't matter much for incense, I was just thinking that in general we try to do that for safety/to make sure we don't lose items. Are there any other 'incense' nouns you could end up with in your hand that are actually valuable?

theurgy.lic
```diff
@@ -776,7 +782,7 @@ class TheurgyActions
     return false unless holy_water?
     safe_walk_to @data['altar']['id']
     DRCI.get_item_safe(@water_holder, @theurgy_supply_container)
-    fput 'clean altar with holy water'
+    bput('clean altar with holy water','Roundtime','You are a bit too busy performing to do that')
```
rpherbig, line 785: Do you even need the 'You are a bit too busy performing to do that' match? I thought bput automatically handled that as a match, too, even if unspecified?

theurgy.lic
Hiinky: It is.  By not returning it to the right container at the end of this action, it leaves it in-hand which later causes snags when a badly timed stow_hands uses `stow` which may or may not put it in the right place.  All related actions get the water holder from theurgy container (filling the container or using it) so this is the only place that randomly moves it.

theurgy.lic
Hiinky: Nope!  I just got in a hurry removing the other lines.

theurgy.lic
Hiinky: Adding!

theurgy.lic
Hiinky: True. I can't think of anything off hand but better to future proof!

theurgy.lic
rpherbig: Sorry, let me clarify. I understand _why_ you added the cleanup here.

What I'm saying is that the return value of `holy_water?` has a dramatically different _meaning_ now than it did before. For example, if you have a water holder in your supply container, but it has no water in it, I think this function now returns `true` when it used to return `false`. I'm worried that this has ripple effects.

theurgy.lic
```diff
@@ -729,12 +725,26 @@ class TheurgyActions
...
     waitrt? while bput('light my incense with flint',
                        'nothing happens',
                        'bursts into flames',
                        'much too dark in here to do that') == 'nothing happens'
     waitrt?
+
+    # based on tap test, returns flint to eddy if required
+    if dropped
+      DRCI.put_away_item?('flint', @theurgy_supply_container)
+      fput "get incense"
```
rpherbig, line 745: Can this be a `bput`?

theurgy.lic
```diff
@@ -256,6 +251,7 @@ class TheurgyActions
   def holy_water?
     DRCI.get_item_safe(@water_holder, @theurgy_supply_container)
```
rpherbig, line 252: Should line 266 also use `get_item_safe`?

https://github.com/rpherbig/dr-scripts/pull/5390/files#diff-f4e18e3d16b30f3605ba753ee5a27d679c60b459748d8708cb17ac111cc9b877R266

theurgy.lic
Hiinky: Done!

theurgy.lic
Hiinky: Can't hurt to make it all uniform.  Done!

theurgy.lic
Hiinky: Ahhhh  I caught up!  I see what you mean.

Hiinky:
```
  def holy_water?
    DRCI.get_item_safe(@water_holder, @theurgy_supply_container)
    result = DRCI.inside?('holy water', @water_holder)
    DRCI.put_away_item?(@water_holder, @theurgy_supply_container)
    result
  end
  ```
  This should work without adding more logic?

rpherbig: Looks good

Hiinky:
I think we're all set!

PR #5394, opened 2022-01-09
[script][log]adding this to revert to default behaviour

MahtraDR:
I should have added this with a different name. Someone objects to log datetimestamps and we can't do args on startup scripts, so reverting this. I'll submit datetimestamped log.lic as a different named file.

### DIFF
--- a/log.lic
+++ b/log.lic
@@ -5,11 +5,9 @@
 
 	 author: Tillmen (tillmen@lichproject.org)
 	   game: any
-	version: 0.3
+	version: 0.2
 
 	changelog:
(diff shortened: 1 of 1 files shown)


KatoakDR:
I recommend keeping the one script but relying on a yaml setting to opt-in to date stamping. That would avoid duplicate scripts and still allow for auto start of "log".

MahtraDR:
> I recommend keeping the one script but relying on a yaml setting to opt-in to date stamping. That would avoid duplicate scripts and still allow for auto start of "log". 

Great suggestion. Thank you.

PR #5399, opened 2022-01-10
[script][log] Adding timestamp toggle, and format options

MahtraDR:
I should not have changed the default behaviour of a script like that... even if it's a script from another repo...

### DIFF
--- a/log.lic
+++ b/log.lic
@@ -5,9 +5,13 @@
 
 	 author: Tillmen (tillmen@lichproject.org)
 	   game: any
-	version: 0.2
+	version: 0.4
 
 	changelog:
+                0.4 (2022-01-10:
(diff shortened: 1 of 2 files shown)


log.lic
```diff
@@ -25,6 +29,10 @@ end
 
 hide_me
 
+@settings = get_settings
```
rpherbig, line 32: Settings doesn't need to be an instance var

log.lic
```diff
@@ -5,9 +5,13 @@
 
 	 author: Tillmen (tillmen@lichproject.org)
 	   game: any
-	version: 0.2
+	version: 0.4
 
 	changelog:
+                0.4 (2022-01-10:
```
rpherbig, line 11: The indentation is strange here - tabs vs spaces?

PR #5408, opened 2022-01-12
[script] [common] Centralize logic to check if moons are visible

KatoakDR:
### Background
* The logic to check if moons are visible is repeated across a few scripts; I'd like to centralize it
  - common-arcana
  - gate
  - hunting-buddy
  - mm
  - t2 (forthcoming)

### Changes
* Add method `moons_visible?` that returns true if your moonwatch data sees moons, else false
* Add method `visible_moons` that returns array of moon names (katamba, yavash, xibar) that are visible


common.lic
```diff
@@ -224,6 +224,20 @@ module DRC
...
+  # Returns true if at least one moon (e.g. katamba, yavash, xibar)
+  # is above the horizon and won't set for at least another ~4 minutes.
+  def moons_visible?
+    !visible_moons.empty?
+  end
+
+  # Returns list of moon names (e.g. katamba, yavash, xibar)
+  # that are above the horizon and won't set for at least another ~4 minutes.
+  def visible_moons
+    DRCA.check_moonwatch
```
rpherbig, line 236: Does this introduce a circular dependency?

rpherbig: I noticed there's no `custom_require.call(%w[common-arcana])` in this file, but DRCA requires DRC

rpherbig: Should these be in DRCA or DRCMM?

KatoakDR: 🤦‍♂️ Well that's embarrassing. Good catch on the circular dependency, and yes I agree these should move to common-moonmage (as moon mages are the primary use case for these methods)

KatoakDR: I've moved the new methods to `common-arcana` alongside the `check_moonwatch` function they depend on

PR #5410, opened 2022-01-12
[script] [t2] option to skip skills that require moons

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5408

### Background
* I've been playing my Moon Mage lately and the moons have not been visible for long stretches of time while I'm in the game.
* T2 doesn't currently have a way to skip training certain skills/sections if moons aren't visible so I get in loops where I can't get past my combat-block (which bails when no moons) to do other tedious things like train scholarship or first aid, etc
* I'd rather not re-order my T2 section, I like hunting first then diving in to other things
* What I'm looking for is for T2 to skip whole sections if moons aren't visible to avoid these hangups

### Changes
* 🆕 Support new property when defining a training section: `moons: boolean` much like how `hunting-buddy` does
* 🧹 Reorder custom_require imports to sort `common-*` scripts first
* 🧹 Trim whitespace on file save
* 🧹 Add parens to method calls

### DIFF
--- a/t2.lic
+++ b/t2.lic
@@ -5,11 +5,9 @@
         and thresholds in which the scripts should be started
 =end
 
-custom_require.call(%w[common drinfomon equipmanager common-arcana])
+custom_require.call(%w[common common-arcana drinfomon equipmanager])
 
 class T2
-  # include DRC for various common calls below
-  # include DRCA for the cyclic release handling at script exit
 
   def initialize
     @shutdown = false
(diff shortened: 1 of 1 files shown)


t2.lic
```diff
@@ -41,11 +41,15 @@ class T2
         end
 
         # Skip if the skill is still on cooldown between trainings
-        next unless has_cooldown?(trainable['skill']) && cooldown_expired?(trainable['skill'])
+        if has_cooldown?(trainable['skill'])
```
rpherbig, line 44: Style-wise, I'd prefer to see the conditionals combined

```suggestion
        if has_cooldown?(trainable['skill']) && !cooldown_expired?(trainable['skill'])
            next
        end
```

rpherbig: Or one-line them like this:

```suggestion
        next if has_cooldown?(trainable['skill']) && !cooldown_expired?(trainable['skill'])
```

KatoakDR: Resolved in https://github.com/rpherbig/dr-scripts/pull/5410/commits/bcaeb120cc6206bceda7642768380696de2a971a

KatoakDR:
Space command, we are ready for launch

PR #5411, opened 2022-01-13
[script][common-arcana][validate][data][base.yaml] Adding option to prevent cyclics from being released by DRCA.release_cyclics.

MahtraDR:
Sorry for opening a new PR. I borked my branch and git and I had differences we couldn't reconcile.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -406,6 +406,7 @@ module DRCA
       .spell_data
       .select { |_name, properties| properties['cyclic'] }
       .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
+      .reject { |name| get_settings.cyclic_no_release.include?(name) }
       .map { |_name, properties| properties['abbrev'] }
       .each { |abbrev| fput("release #{abbrev}") }
   end
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -401,12 +401,12 @@ module DRCA
     return charged
   end
 
-  def release_cyclics
+  def release_cyclics(cyclic_no_release = [])
     get_data('spells')
       .spell_data
       .select { |_name, properties| properties['cyclic'] }
       .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
-      .reject { |name| get_settings.cyclic_no_release.include?(name) }
+      .reject { |name| cyclic_no_release.include?(name) }
       .map { |_name, properties| properties['abbrev'] }
       .each { |abbrev| fput("release #{abbrev}") }
   end
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -2123,3 +2123,6 @@ log_timestamp: false
 # default is 2022-01-10 01:36:46 UTC format
 log_timestamp_format: "%F %T %Z"
 
+# List of cyclic spells that won't get released by scripts when
+# generally releasing cyclics
+cyclic_no_release:
--- a/validate.lic
+++ b/validate.lic
@@ -60,11 +60,9 @@ class DRYamlValidator
   def assert_that_cyclic_no_release_spell_is_invalid_unless_listed_in_base_spells(settings)
     return if !settings.cyclic_no_release
 
-    settings.cyclic_no_release.each do |spell_name|
-      if !get_data('spells').spell_data.include?(spell_name)
-        warn ("#{spell_name} in cyclic_no_release is not a valid spell name.")
-      end
-    end
+    settings.cyclic_no_release
+      .reject { |spell_name| get_data('spells').spell_data.include?(spell_name) }
+      .each { |spell_name| error("#{spell_name} in cyclic_no_release is not a valid spell name.") }
   end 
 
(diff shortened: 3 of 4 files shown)


common-arcana.lic
```diff
@@ -406,6 +406,7 @@ module DRCA
       .spell_data
       .select { |_name, properties| properties['cyclic'] }
       .select { |name, _properties| DRSpells.active_spells.keys.include?(name) }
+      .reject { |name| get_settings.cyclic_no_release.include?(name) }
```
rpherbig, line 409: Calling `get_settings` in a `common-*` file is undesirable and should be avoided if possible because it means we'll only ever pick up the setting as defined in the main YAML file - it can't be overriden in sub-YAML files. We should try to pass either the specific setting or settings itself into this function.

profiles/base.yaml
```diff
@@ -2123,3 +2123,6 @@ log_timestamp: false
 # default is 2022-01-10 01:36:46 UTC format
 log_timestamp_format: "%F %T %Z"
 
+# List of cyclic spells that won't get released by scripts when
+# generally releasing cyclics
+cyclic_no_release:
```
rpherbig, line 2128: If this is intended to be an array, it needs an entry in `base-empty.yaml`

validate.lic
```diff
@@ -57,6 +57,16 @@ class DRYamlValidator
     warn('You have no weapons configured in weapon_training: this will likely cause problems.')
   end
 
+  def assert_that_cyclic_no_release_spell_is_invalid_unless_listed_in_base_spells(settings)
+    return if !settings.cyclic_no_release
+
+    settings.cyclic_no_release.each do |spell_name|
+      if !get_data('spells').spell_data.include?(spell_name)
+        warn ("#{spell_name} in cyclic_no_release is not a valid spell name.")
```
rpherbig, line 65: Warn or error?

validate.lic
```diff
@@ -57,6 +57,16 @@ class DRYamlValidator
     warn('You have no weapons configured in weapon_training: this will likely cause problems.')
   end
 
+  def assert_that_cyclic_no_release_spell_is_invalid_unless_listed_in_base_spells(settings)
+    return if !settings.cyclic_no_release
+
+    settings.cyclic_no_release.each do |spell_name|
```
rpherbig, line 63: You could do this with function chaining, such as (please forgive any syntax mistakes):

```suggestion
    settings.cyclic_no_release
    .reject { |spell_name| get_data('spells').spell_data.include?(spell_name) }
    .each { |spell_name| warn ("#{spell_name} in cyclic_no_release is not a valid spell name.") }
```

validate.lic
```diff
@@ -57,6 +57,14 @@ class DRYamlValidator
     warn('You have no weapons configured in weapon_training: this will likely cause problems.')
   end
 
+  def assert_that_cyclic_no_release_spell_is_invalid_unless_listed_in_base_spells(settings)
+    return if !settings.cyclic_no_release
+
+    settings.cyclic_no_release
+      .reject { |spell_name| get_data('spells').spell_data.include?(spell_name) }
```
KatoakDR, line 64: 💡 Avoid reloading the file for each spell being checked. The `reject` method is like a loop, it iterates all elements listed in `cyclic_no_release` list. The result of `get_data('spells')` won't be different each iteration.
```suggestion
    spell_data = get_data('spells').spell_data
    settings.cyclic_no_release
      .reject { |spell_name| spell_data.include?(spell_name) }
```

common-arcana.lic
KatoakDR: Additionally, the `release_cyclics` method needs an argument to force releasing them regardless the user's setting. For example, when `combat-trainer` decides it's time to cast a new cyclic spell then it must release the current one.

validate.lic
MahtraDR: Thank you. I did not know that. Commited.

validate.lic
MahtraDR: I've changed it to error. 

```
 [validate: ERROR:< garbled in cyclic_no_release is not a valid spell name.  >]                                                                                                                                                               
```

for 

```
cyclic_no_release:
  - Regenerate
  - garbled
```

In my empath's config.

common-arcana.lic
MahtraDR: Hah, yes, I hadn't considered that scenario, and beyond that, the new functionality @Hiinky just added to release cyclics on low-mana would also need to be updated. Note, however, that since there can only ever be one cyclic cast at a time, even for a bard... it would be user error to configure a spell to not be released by scripts, and expect a different cyclic to be cast. Segue functionality for bards wouldn't be affected.

MahtraDR: @KatoakDR and I chatted about this. See the current version. Once this is in, we can go fix the various scripts that might want to use this. The main focus is really T2/Hunting-Buddy, which I'll look into in a couple of days after this is merged in.

KatoakDR (commented):
Code looks good to me, thanks 👍

validate.lic
rpherbig: Nice catch!

PR #5412, opened 2022-01-13
[script][hunting-buddy] Clearing all avoids when exiting due to low health

MahtraDR:
When HB exits due to low health, in many cases a character will have their avoids set still, as they're set by t2. With this change, when a character logs back in, they can at least be dragged out immediately, instead of having to perform the manual action of clearing avoids first...

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -278,6 +278,7 @@ class HuntingBuddy
       clear
       if health < @settings.health_threshold
         message('***STATUS*** exiting due to low health')
+        fput('avoid all')
         fput('exit')
       end
       if @settings.stop_hunting_if_bleeding && bleeding?


KatoakDR:
@MahtraDR Rather than clearing all `avoid`s, should this focus specifically on `touching` and `dragging`? The ones that would impede dragging to safety and healing?

As is, if someone was avoiding coins or joining or crime, etc then those are being wiped out and people now need to be aware that they'd have to reset them.

MahtraDR:
@KatoakDR The most commonly used script for this scenario of exiting on low-health is afk.lic and that already  fputs and avoid all like we changed this to do. I understand your concern, and at the same time, I am not sure it applies here. Note that t2 already has a default avoids setting, which is additionally configurable, so in a scenario when this logs you out, assuming you're using t2, you'll get your avoids back on next run of that script regardless. Anyway... let's see how this goes, as its been merged in already.

PR #5413, opened 2022-01-15
[script][T2] T2 use cyclic no release

MahtraDR:
Adding support for new cyclic_no_release function into T2.

### DIFF
--- a/t2.lic
+++ b/t2.lic
@@ -125,7 +125,7 @@ before_dying do
   scripts.each do |script_name|
     stop_script(script_name) if Script.running?(script_name)
   end
-  DRCA.release_cyclics(@settings.cyclic_no_release)
+  DRCA.release_cyclics(get_settings.cyclic_no_release)
 end
 
 $T2 = T2.new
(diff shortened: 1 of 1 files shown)


t2.lic
```diff
@@ -125,7 +125,7 @@ before_dying do
   scripts.each do |script_name|
     stop_script(script_name) if Script.running?(script_name)
   end
-  DRCA.release_cyclics
+  DRCA.release_cyclics(@settings.cyclic_no_release)
```
KatoakDR, line 128: Although an elegant change, unfortunately instance variables (`@settings`) are not available in the `before_dying` function.

Unlike the advice about not calling `get_settings` in **common** utils, this is one scenario where you have to call `get_settings` (example from [performance](https://github.com/rpherbig/dr-scripts/blob/85d003a6ff266c2f275f1092ce2f63c8ce68b436/performance.lic#L84) script)
```suggestion
  DRCA.release_cyclics(get_settings.cyclic_no_release)
```

rpherbig:
Remind me in two days to merge this

MahtraDR:
@rpherbig reminder. :)

PR #5422, opened 2022-01-16
[script] [common-arcana] Remove duplicate code to set moon data; use new util methods

KatoakDR:
### Background
* https://github.com/rpherbig/dr-scripts/pull/5408 introduced new methods to encapsulate logic parsing `UserVars.moons` to know if any are visible and which ones are

### Changes
* 🧹 Clean up where code can be replaced by calls to the new util methods

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -560,21 +560,20 @@ module DRCA
       data['cast'] = "cast #{cast_on}"
       return data
     end
-    echo 'Could not find any planets to cast on'
+    echo "No planet available to cast #{data['abbrev']}"
     nil
   end
 
   def set_moon_data(data)
     return data unless data['moon']
 
-    check_moonwatch
-    # Look for a moon that will be up for the next few minutes
-    moon = UserVars.moons.find { |moon_name, moon_data| UserVars.moons['visible'].include?(moon_name) && moon_data['timer'] >= 4 }.first
-    unless moon
+    moon = visible_moons.first
+    if moon
+      data['cast'] = "cast #{moon}"
+    else
       echo "No moon available to cast #{data['abbrev']}"
-      return nil # Return nil to indicate that setting moon casting data failed
+      data = nil
     end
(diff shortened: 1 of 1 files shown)


common-arcana.lic
```diff
@@ -560,21 +560,20 @@ module DRCA
...
+    echo "No planet available to cast #{data['abbrev']}"
     nil
   end
 
   def set_moon_data(data)
     return data unless data['moon']
 
-    check_moonwatch
-    # Look for a moon that will be up for the next few minutes
-    moon = UserVars.moons.find { |moon_name, moon_data| UserVars.moons['visible'].include?(moon_name) && moon_data['timer'] >= 4 }.first
```
KatoakDR, line 572: The new `visible_moons` method handles parsing  `UserVars.moons` and calling `check_moonwatch`

common-arcana.lic
```diff
@@ -804,26 +803,13 @@ module DRCA
...
-        if weather =~ /inside/
-          echo "*** You're inside and there are no available moons. You're going to have a hard time casting #{data['abbrev']}"
-        end
-
-        unless moon = UserVars.moons['visible'].first
-          return
-        end
-        data['cast'] = "cast #{moon}"
-      end
-    end
```
KatoakDR, line 825: `set_moon_data` does these checks and updates the cast command

There is no need for the weather check, it's immaterial to any decisions being made

common-arcana.lic
```diff
@@ -560,21 +560,20 @@ module DRCA
       data['cast'] = "cast #{cast_on}"
       return data
     end
-    echo 'Could not find any planets to cast on'
+    echo "No planet available to cast #{data['abbrev']}"
     nil
   end
 
   def set_moon_data(data)
```
rpherbig, line 567: This probably isn't something we can change here, but the Ruby convention would be to name the function `set_moon_data!`

KatoakDR: Good to know, thanks

PR #5426, opened 2022-01-16
Add: force_crafting_hometown > Forging

Hiinky:
This is very similar to a draft @vtcifer had started ages ago then tons of things changed and some of the commons have been updated for other PRs but are applied very effectively here.

Purpose: Allow user to set `force_crafting_town:` similar to force_healer_town to utilize less-busy locations.  Most users would really only need this for forging but for the sake on uniformity and future-proofing, I am adding support to all of the craft scripts.

### DIFF
--- a/alchemy.lic
+++ b/alchemy.lic
@@ -29,7 +29,7 @@ class Alchemy
     exit if args.craft && (args.forage || args.prepare)
 
     settings = get_settings
-    @hometown = settings.hometown
+    @hometown = @settings.force_crafting_town || @settings.hometown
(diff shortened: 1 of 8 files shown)


Hiinky:
Still adding more crafting scripts and considering logic to allow setting the crafting hometown down to the discipline level (as VTC had it originally) or an array of hometowns to check in case of forging where a society room in mandatory.

rpherbig:
Code looks good to me ✔️

MahtraDR:
Closing this, as I have an up to date version in https://github.com/rpherbig/dr-scripts/pull/6871

PR #5434, opened 2022-01-17
[script] [pawn-items] 🆕 pawn items in a container

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5433

### Related
* https://github.com/rpherbig/dr-scripts/pull/5448
* https://github.com/rpherbig/dr-scripts/pull/5450

### Background
* My characters collect in Crossing pawnable items that have to be sold in other towns
* To cut down on travel, I move them all on to one character and make the trip periodically
* I want to automate selling all the items from my container to the pawn shop similar to how `sell-loot` sells gems/metals or `offload-items` moves items from a container to off your person

### Changes
* 🆕 Script modeled after `offload-items` that takes you to pawn shop and sells the items in the container. Uses same safety mechanisms to prompt user to confirm and they have to explicitly unpause the script

### DIFF
--- a/pawn-items.lic
+++ b/pawn-items.lic
@@ -1,0 +1,84 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#pawn-items
+=end
+
+custom_require.call %w[common common-items common-travel equipmanager]
+
+class PawnItems
+
+  def initialize
+    arg_definitions = [
+      [
+        { name: 'source', regex: /^[A-z\.\s\-']+$/i, variable: true, description: 'Source container' },
+        { name: 'noun', regex: /^[A-z\.\s\-]+$/i, optional: true, variable: true, description: 'If specified, only items with this noun will be sold.' }
+      ]
+    ]
+    args = parse_args(arg_definitions)
+
+    # Put away items in hands to mitigate accidentally selling them
+    EquipmentManager.new.return_held_gear
+    # Warn the script user that this is designed to sell items off your person
+    warn_before_sell(args.source, args.noun)
+    # Turn items into coins
+    sell_items(args.source, args.noun)
+  end
+
+  def sell_items(source, noun)
+    # If container is very full then LOOK may not list all of them.
+    # If you're moving a specific item, then sort those to the top
+    # to increase chances we find and sell all of them in one go.
+    DRC.bput("sort #{noun} in my #{source}", "are now at the top", "What were you referring to", "Please rephrase that command", "You may only sort items in your inventory") if noun
+    DRCI.get_item_list(source, 'look')
+      .map { |full_name| full_name =~ /lot of other stuff/ ? full_name : full_name.split(' ').last }
+      .select { |item| noun ? /\b#{noun}\b/ =~ item : true }
+      .each do |item|
+        # This indicates there is more items than LOOK can show
+        # and we've reached the end of what we last saw.
+        # We need to look again to see what we can now see.
+        # Keep doing this until exhaust source.
+        if item =~ /lot of other stuff/
+          sell_items(source, noun)
+          return
+        end
+        # Attempt to get the item from the source container.
+        if DRCI.get_item?(item, source)
+          unless DRCT.sell_item(Room.current.id, item)
+            # Failed. Message user and return item to source container.
+            DRC.message("Unable to sell #{item}. The merchant may not accept these items.")
+            DRCI.put_away_item?(item, source)
+            exit
+          end
+        else
+          DRC.message("Unable to get #{item} from #{source}.")
+          DRC.message("Your hands are full!") if (DRC.left_hand && DRC.right_hand)
+          exit
+        end
+      end
+  end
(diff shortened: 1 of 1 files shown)


pawn-items.lic
```diff
@@ -0,0 +1,84 @@
...
+    DRC.bput("sort #{noun} in my #{source}", "are now at the top", "What were you referring to", "Please rephrase that command", "You may only sort items in your inventory") if noun
+    DRCI.get_item_list(source, 'look')
+      .map { |full_name| full_name =~ /lot of other stuff/ ? full_name : full_name.split(' ').last }
+      .select { |item| noun ? /\b#{noun}\b/ =~ item : true }
+      .each do |item|
+        # This indicates there is more items than LOOK can show
+        # and we've reached the end of what we last saw.
+        # We need to look again to see what we can now see.
+        # Keep doing this until exhaust source.
+        if item =~ /lot of other stuff/
```
rpherbig, line 39: Is there a chance we never exhaust the source? E.g. you use a noun and there are none in the source, but the source has a ton of non-noun items in it?

pawn-items.lic
```diff
@@ -0,0 +1,84 @@
...
+    # Pause script and warn that it is designed to sell items
+    # from your person to somewhere else. Item loss may occur.
+    DRC.message("WARNING: This script is designed to sell items from YOU to an NPC merchant.")
+    DRC.message("WARNING: It also has the ability to sell every item from the source container.")
+    DRC.message("WARNING: Item loss may occur. Use with caution.")
+    DRC.message("WARNING: If you want to sell loot like gems and bundles, use ;sell-loot instead.")
+    DRC.message("WARNING: If you want to transfer items between containers, use ;transfer-items or ;offload-items instead.")
+    DRC.message("\n")
+
+    unless noun
```
rpherbig, line 69: I prefer to use `if` over `unless` when possible. E.g. in this case you could use `if` and swap the cases.

pawn-items.lic
```diff
@@ -0,0 +1,84 @@
...
+        # and we've reached the end of what we last saw.
+        # We need to look again to see what we can now see.
+        # Keep doing this until exhaust source.
+        if item =~ /lot of other stuff/
+          sell_items(source, noun)
+          return
+        end
+        # Attempt to get the item from the source container.
+        if DRCI.get_item?(item, source)
+          unless DRCT.sell_item(Room.current.id, item)
```
rpherbig, line 45: Why was this helper introduced to `DRCT` when you only call it with the current room ID?

KatoakDR: I followed the example of [DRCT.buy_item(room_id, item)](https://github.com/rpherbig/dr-scripts/blob/c8e20343d9545637daaebec245646b3e7269f069/common-travel.lic#L35 method when I added `sell_item`. Both methods will walk you to the merchant to do the buy/sell. 

But for why did I do `Room.current.id` .... maybe I thought there'd be more than pawn shops you could sell items to? This probably should pull from the base town data to go to the _pawn shop_ considering the name and intent of this script, and support a town override argument (if you're in Crossing and want to pawn in Shard or Fang Cove)

pawn-items.lic
KatoakDR: Noted, thanks

KatoakDR: FYI @MahtraDR @asechrest

pawn-items.lic
KatoakDR: Great question. I took a few minutes to walk through this some more to double check and I don't think this will go in to infinite recursion.

My conclusion is that because "lot of other stuff" will be the last item in the list, each prior item will follow the code path to line 44 to offload the contents.

If the item cannot be removed from the container then the script exits (avoids infinite recursion of rerummaging but never reducing item count).

If the item can be removed from the container then it's either sold (and so the container is becoming less full over time) or the item is returned to the container and the script exits (avoids infinite recursion again).

Is there a scenario you're thinking of that don't fall in to the above?

KatoakDR: > you use a noun and there are none in the source, but the source has a ton of non-noun items in it?

If you provided the `noun` argument, then lines 30 and 33 should ensure that if any items do have that noun then they'll be at the top of the container. Line 34 can only occur for items are the "noun" you're looking for.

pawn-items.lic
KatoakDR: Resolved in https://github.com/rpherbig/dr-scripts/pull/5434/commits/ed46457064b60f222f6de0079fa9ebceacbf7f2c

pawn-items.lic
KatoakDR: Resolved in https://github.com/rpherbig/dr-scripts/pull/5434/commits/ed46457064b60f222f6de0079fa9ebceacbf7f2c

pawn-items.lic
asechrest: Fair. Robert noted this when I created offload items. I just never came back and tweaked it.

PR #5437, opened 2022-01-17
[script] [almanac] Add `almanac_no_use_rooms` setting, other refactorings

KatoakDR:
### Background
* Similar run-in-the-background scripts like `wand-watcher` and `buff-watcher` support a list of exclusion rooms where not to try and run the script, either out of preference or because the room won't allow it, as well as follow a particular _template_ for how the class is structured.
* This pull request brings `almanac` more in line with them for consistency.

### Changes
* Remove `include` in favor of using `DRC` and `DRCI` prefixed methods
* Create individual methods for checking specific criteria on when it's ok to use the almanac
* Use `DRC.smart_pause_all` to only pause _pausable_ scripts, and to remember which scripts were paused by this script so that at the end only those are the ones that are unpauased and not _every_ paused script suddenly is unpaused. You might have paused an unrelated script and didn't want it unpaused yet.
* Add a slight delay after pausing scripts to handle a race condition where `combat-trainer` or `almanac` may get confused when parsing game output and mistake what's actually in your hands.
* In addition to not using an almanac when hidden, don't while invisible either.
* Group almanac yaml settings together (they were split by sanowret settings)

### Configuration
_existing settings_
```yaml
# The noun of your almanac (e.g. almanac, tome, treatise, monograph, etc)
almanac_noun: almanac

# List of skills to turn almanac to train.
# Leave blank unless you have a "pick a skill" almanac.
almanac_skills:
  - Scholarship
  - Appraisal
  - ...

# List of scripts that when running then almanac script won't run.
almanac_no_use_scripts:
  - athletics
  - burgle
  - steal
  - ...
```
_new settings_
```yaml
# List of room titles or room ids or regular expressions
# that if matches the current room then almanac script won't run.
almanac_no_use_rooms:
  - !ruby/regexp /\b(Carousel|Vault|Item Registration Office|Registrar's Office)\b/
```

### DIFF
--- a/almanac.lic
+++ b/almanac.lic
@@ -5,28 +5,26 @@
 custom_require.call(%w[common common-items])
 
 class Almanac
-  include DRC
-  include DRCI
 
   def initialize
     settings = get_settings
     UserVars.almanac_last_use ||= Time.now - 600
     @no_use_scripts = settings.almanac_no_use_scripts
+    @no_use_rooms = settings.almanac_no_use_rooms
     @almanac_skills = settings.almanac_skills
     @almanac = settings.almanac_noun
 
     passive_loop
   end
 
-  def use_almanac
-    return if Time.now - UserVars.almanac_last_use < 600
-    return if @no_use_scripts.any? { |name| Script.running?(name) }
-    return if XMLData.room_title.include? 'Carousel Chamber'
-    return if XMLData.room_title.include? 'Carousel Booth'
-    return if XMLData.room_title.include? 'Family Vault'
-    return if hidden?
-    return if checkleft
+  def passive_loop
+    loop do
+      use_almanac if should_use_almanac?
+      pause 20
+    end
+  end
 
+  def use_almanac
     unless @almanac_skills.empty?
       training_skill = @almanac_skills
                        .select { |skill| DRSkill.getxp(skill) < 18 }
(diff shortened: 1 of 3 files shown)


almanac.lic
```diff
@@ -5,64 +5,88 @@
...
   end
 
-  def use_almanac
-    return if Time.now - UserVars.almanac_last_use < 600
-    return if @no_use_scripts.any? { |name| Script.running?(name) }
-    return if XMLData.room_title.include? 'Carousel Chamber'
-    return if XMLData.room_title.include? 'Carousel Booth'
-    return if XMLData.room_title.include? 'Family Vault'
-    return if hidden?
-    return if checkleft
```
KatoakDR, line 28: Checks moved to their own methods, called from `should_use_almanac?`

profiles/base.yaml
```diff
@@ -1800,32 +1800,14 @@ spell_cycle:
...
 restock_shop:
 
-#custom almanac noun
+# The noun of your almanac (e.g. almanac, tome, treatise, monograph, etc)
 almanac_noun: almanac
-# For almanacs that let you pick which skill to use
-almanac_skills:
 
-sanowret_adjective: sanowret
-sanowret_no_use_scripts: # Don't try to use crystal while these scripts are active.
```
KatoakDR, line 1809: Despite what GitHub says, these lines were not "deleted", they simply were moved below the new almanac setting

almanac.lic
```diff
@@ -2,67 +2,91 @@
...
+  def should_use_almanac?
+    !(hidden? || invisible? || almanac_on_cooldown? || hands_full? || running_no_use_scripts? || inside_no_use_room?)
   end
+
+  def running_no_use_scripts?
+    @no_use_scripts.any? { |name| Script.running?(name) }
+  end
+
+  def inside_no_use_room?
+    @no_use_rooms.any? { |room| room === DRRoom.title.to_s()[2..-3] || room == Room.current.id }
```
rpherbig, line 79: > DRRoom.title.to_s()[2..-3]

This feels very "magical". I realize it's done in a bunch of other scripts, but it looks funny and I can't for the life of me remember why.

> s = "this is a very long string which I am making longer"
> puts s[2..-3]
> is is a very long string which I am making long

This seems to mangle the title a lot...

profiles/base.yaml
```diff
@@ -1850,6 +1836,34 @@ almanac_no_use_scripts:
 - corn-maze
 - burgle
 
+# List of room titles or room ids or regular expressions
+# that if matches the current room then almanac script won't run.
+almanac_no_use_rooms:
+  - !ruby/regexp /\b(Carousel|Vault|Item Registration Office|Registrar's Office)\b/
```
rpherbig, line 1842: Why not an array with 4 items in it?

almanac.lic
KatoakDR: Agree, those nuances should be abstracted.  It's because the room title is enclosed in double brackets [[Some Place]]. I don't know why they weren't stripped out in lich.rbw to begin with

profiles/base.yaml
KatoakDR: I think it was because as string literals then they must match the room name exactly. This matches rooms that contain those words since it's a regex match.

rpherbig: Is that a concern we have?

almanac.lic
rpherbig: I'd love to see that change, but it's not necessary for this PR

profiles/base.yaml
KatoakDR: It's handy, yes, otherwise we have to list out each vault room for every town you might be in because the town is in the room's title.

For example, `[Fang Cove, Carousel Chamber]` or `[Crossing, Carousel Booth]`

KatoakDR: And if you're asking "why do the string literals have to match the room title exactly?", I had that same question when this paradigm was being introduced 1-2 years ago and the developer at the time was insistent that it match the room name exactly or use regex. So, here we are

PR #5438, opened 2022-01-17
Make gem_pouch_adjective Optional

Hiinky:
Not everyone uses the exact same pouch for appraisal training and are severely limited by enforcing use of said variable.


KatoakDR:
> Not everyone uses the exact same pouch for appraisal training and are severely limited by enforcing use of said variable.

Should we instead recommend that these people set `gem_pouch_adjective:` to blank in their yaml file? Then no script change needed?

Hiinky:
I believe that would cause pouch swapping to break.

KatoakDR:
Got it, thanks

PR #5452, opened 2022-01-21
[script][dependency] Removing DRC references in dependency

MahtraDR:
When I switched out from Etreu's name to point to discord, I added DRC.message, which is entirely not appropriate for dependency, since it needs to be able to run entirely before any scripts exist...


rpherbig:
Nice catch

PR #5453, opened 2022-01-22
[meta] Removing execute bit on these files

MahtraDR:
There is no reason to have the execute bit set on these. Minor clean up only. Anyone wanting to execute them will know how to make that happen regarless.


rpherbig:
Huh, I never knew execute was stored in git

PR #5458, opened 2022-01-23
[script] [dependency] Rewrite SetupFiles to be thread-safe

### LINKED ISSUE #5451: Possible race condition/collision causing semi-frequent get_settings or get_data failures
We've had a number of reports, or personal experiences, with failures to `get_settings` or `get_data` within scripts. Examples:

- penguin23 on Discord reports that occasionally, my optional pretty formatting of the `;circlecheck` script output appears to fail to pull the yaml setting `circlecheck_prettyprint`. (Issue #5419 )
- @MahtraDR reports that his updates to allow releasing only certain cyclics sometimes fails to pull the yaml list of cyclics not to release.
- @KatoakDR reports that his autostarted `;buff-watcher` occasionally exits at starting because it finds no waggle sets.
- In script code [there exists some workarounds](https://github.com/rpherbig/dr-scripts/blob/8e38c9623a51893c389fc2724fd537d47c5ceb42/common-travel.lic#L330) for when this collision or race condition rears its head.

I wrote a script to see if I could cause the condition to happen on a regular basis and was successful. Granted, this was a stress test.

image

The frequency of the failure seems to be related to a lot going on, and a possible collision _between_ `get_data` and `get_settings`.

I was able to cause the most frequent failures during busy times (combat), with a number of scripts already running, and by making adjacent `get_data` and `get_settings` calls in my stress test script.

I don't have the background at the moment to really fix this issue, but I know @KatoakDR and @MahtraDR are taking a look at it. The hunch at the moment is, again, the possible collision between between the two methods and how they do what they do.

I'll let those two expound on the hunch if desired.

KatoakDR:
### Background
* Fixes #5451

### Changes
* Use `MonitorMixin` and `self.synchronize`  to make loading files and updating cache thread-safe
* Refactored code into smaller functions focused on their specific task
* Refactor `transform_settings` to reduce duplicated logic updating various settings


rpherbig:
I'll try and take a look at this tomorrow.

cc @rcuhljr

MahtraDR:
@asechrest @KatoakDR and myself have been testing this, as we've all seen variations of the problem. @asechrest wrote a nifty test script to showcase this:

```

custom_require.call(%w[common])

class Test
  include DRC

  def initialize
    arg_definitions = [
      [
        { name: 'loops', regex: /^\d+$/i, variable: true, description: 'Number of times to loop get_settings tries.' }
      ]
    ]

    # time_settings_pulls

    test_single_line_get_settings(args.loops)
  end

  def test_single_line_get_settings(loops)
    test_setting = false
    crafting_data = nil
    failure_counter = 0
    time_elapsed = 0
    loop_times = loops.to_i

    start_total_time = Time.now
    DRC.message("Starting test: #{loop_times} times")
    loop_times.times do
      start = Time.now
      test_setting = get_settings.test_setting
      finish = Time.now

      time_elapsed += (finish - start)

      if !test_setting
        failure_counter += 1
      end
      test_setting = false
      crafting_data = nil
    end
    finish_total_time = Time.now

    DRC.message("Test complete using dependency version: #{$DEPENDENCY_VERSION}")
    DRC.message("Tested a get_data assignment immediately followed by get_settings.test_setting assignment: #{loop_times} times")
    DRC.message("Total failures to pull setting: #{failure_counter}")
    DRC.message("Total test duration: #{finish_total_time - start_total_time}")
    DRC.message("Average time for each settings pull: #{(time_elapsed/loop_times).round(3)}")
  end
end

Test.new
```
which shows this difference over 1000 iterations:
```
 | Test complete using dependency version: 1.4                                                                                                                                                                                                
 | Tested a get_data assignment immediately followed by get_settings.test_setting assignment: 1000 times                                                                                                                                      
 | Total failures to pull setting: 0                                                                                                                                                                                                          
 | Total test duration: 9.312122351                                                                                                                                                                                                           
 | Average time for each settings pull: 0.009                                                                                                                                                                                                 
 --- Lich: test1 has exited.                                                                                                                                                                                                                  
 | Test complete using dependency version: 1.4                                                                                                                                                                                                
 | Tested a get_data assignment immediately followed by get_settings.test_setting assignment: 1000 times                                                                                                                                      
 | Total failures to pull setting: 0                                                                                                                                                                                                          
 | Total test duration: 11.863988086                                                                                                                                                                                                          
 | Average time for each settings pull: 0.012 
```
versus
```
 | Test complete using dependency version: 1.3                                                                                                                                                                                                
 | Tested a get_data assignment immediately followed by get_settings.test_setting assignment: 1000 times                                                                                                                                      
 | Total failures to pull setting: 94                                                                                                                                                                                                         
 | Total test duration: 423.649382503                                                                                                                                                                                                         
 | Average time for each settings pull: 0.424                                                                                                                                                                                                 
 --- Lich: test1 has exited.                                                                                                                                                                                                                  
  | Test complete using dependency version: 1.3                                                                                                                                                                                                
 | Tested a get_data assignment immediately followed by get_settings.test_setting assignment: 1000 times                                                                                                                                      
 | Total failures to pull setting: 116                                                                                                                                                                                                        
 | Total test duration: 445.086521819                                                                                                                                                                                                         
 | Average time for each settings pull: 0.445
```
To replicate this test, use test.lic, and copy it to test1.lic and run both scripts together, so as to simulate what happens when multiple scripts call get_settings at the same time.

asechrest:
I've been running the current iteration of this and it appears to function well. I have yet to register a `get_settings` failure on it, and it appears to be a number of times faster to complete as well.

MahtraDR:
Likewise to @asechrest above, I have been running the new version on my 6, plus some bonus freebies to test the last ~48 hours. Looking good, with zero failures.

KatoakDR:
Recruiting testers in discord

https://discord.com/channels/745675889622384681/745675890242879671/935428648012513321

KatoakDR:
@asechrest @MahtraDR To catch a scenario that Mahtra ran in to with scripts inadvertently affecting each other by modifying settings in-memory, I updated the test runner as below to simulate the combat-trainer khri buff snafu.

Specifically, the "test setting" line I used is `test_setting = get_settings.buff_nonspells.delete('khri')`

```ruby
custom_require.call(%w[common])

class TestDependency

  def initialize
    arg_definitions = [
      [
        { name: 'loops', regex: /^\d+$/i, variable: true, description: 'Number of times to loop get_settings tries.' }
      ]
    ]

    test_single_line_get_settings(args.loops)
  end

  def test_single_line_get_settings(loops)
    test_setting = false
    crafting_data = nil
    failure_counter = 0
    time_elapsed = 0
    loop_times = loops.to_i

    start_total_time = Time.now
    DRC.message("Starting test: #{loop_times} times")
    loop_times.times do
      start = Time.now
      # test_setting = MySetupFiles.get_settings.test_setting
      test_setting = get_settings.buff_nonspells.delete('khri')
      finish = Time.now

      time_elapsed += (finish - start)

      if !test_setting
        failure_counter += 1
      end
      test_setting = false
      crafting_data = nil
    end
    finish_total_time = Time.now

    DRC.message("Test complete using dependency version: #{$DEPENDENCY_VERSION}")
    DRC.message("Tested a get_data assignment immediately followed by get_settings.test_setting assignment: #{loop_times} times")
    DRC.message("Total failures to pull setting: #{failure_counter}")
    DRC.message("Total test duration: #{finish_total_time - start_total_time}")
    DRC.message("Average time for each settings pull: #{(time_elapsed/loop_times).round(3)}")
  end
end

TestDependency.new
```

KatoakDR:
Have made [final call](https://discord.com/channels/745675889622384681/745683361841807390/937125519135236097) for testers in Lich discord. A few folks have said they would test a few days ago, not sure if they have. No one has reported any issues either way.

@MahtraDR @asechrest and myself have been running our characters without issue. I feel very confident about it.

@rpherbig I think you can begin your code review; I don't have any outstanding work for this specific PR, thanks

MahtraDR:
My testing has been running this across 5 accounts for the last week. 

We've also had a few others testing, and whilst we don't yet have test results from a non-redeemed necro... that's the only class that hasn't had good extensive day-to-day running as a test, in case we have class-specific problems (We found an issue with buff_nonspells early on, which has been fixed now.)

MahtraDR:
As of a few minutes ago we have a new tester with non-redeemed necro. That's every class, and variation going. Will update if they reports any issues.

KatoakDR:
Hi @rpherbig, have you had a chance to review this? Thanks

rpherbig:
> Hi @rpherbig, have you had a chance to review this? Thanks

Sorry, life got busy (discovered mice in my crawlspace/walls). I'll take a look now.

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+      @path = path
+      @name = name
+      @data = data
+      @mtime = mtime
+    end
+
+    def data
+      # Create deep clone of data to mitigate scripts changing settings in-memory
+      # because that leads to hard to track down bugs and unpredictable script behaviors.
+      Marshal.load(Marshal.dump(@data))
```
rcuhljr, line 792: Nice for for safety, although we should also make sure we're correcting any submissions that are modifying settings.

dependency.lic
```diff
@@ -8,8 +8,9 @@ require 'base64'
 require 'yaml'
 require 'ostruct'
 require 'digest/sha1'
+require 'monitor'
```
rcuhljr, line 11: Not sure why I just rolled this from scratch instead of finding an existing solution.

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+    end
+
+    def inspect
+      "#<SetupFiles::FileInfo @name=#{@name}, @path=#{@path}, @mtime=#{@mtime}>"
+    end
+  end
+
+  def initialize(debug = false)
+    super() # Must call super() to initialize the MonitorMixin
+    @files_cache = {} # filename => FileInfo
```
rpherbig, line 814: What does this comment mean?

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+
+  def transform_data(original_data)
+    echo "#{self.class}::#{__callee__}" if @debug
+
+    # Enable dot access to top-level properties (e.g. data.cast_messages)
+    data = OpenStruct.new(original_data)
+
+    # ... reserved for future needs ...
+
+    return data
```
rcuhljr, line 1229: The return here is redundant and I believe non idiomatic for ruby.  I don't think the rescue block is anymore vulnerable than any other method layout for someone to come in and drop code after that would alter the return value.

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+      settings.astrology_training.unshift('events') if settings.predict_event
     end
-    unless s.astrology_training.include?('ways')
-      s.astrology_training.unshift('ways') if s.astral_plane_training['train_in_ap']
+    unless settings.astrology_training.include?('ways')
+      settings.astrology_training.unshift('ways') if settings.astral_plane_training['train_in_ap']
     end
 
-    s
+    return settings
```
rcuhljr, line 1212: ditto the other return comment

rcuhljr (approved):
Are there any performance implications from checking all files for changes in a fairly tight loop timer? Yes we're getting faster responses to load settings calls, but in practice those are far less than frequent than every tenth of a second so we're increasing our load by several orders of magnitude, just want to make sure it won't be an issue.

dependency.lic
rcuhljr: it's a map of filenames to FileInfo objects.

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+  # Includes base data plus any specified in the filenames argument.
+  def get_data_glob_patterns(filenames = [])
+    get_glob_patterns(SCRIPTS_DATA_PATH, filenames)
   end
 
-  attr_reader :settings
+  # Generates an array of file glob patterns, one per filename in the given basepath.
+  # `base*.yaml` is automatically and always included in the glob pattern.
+  def get_glob_patterns(basepath = '.', filenames = [])
+    filenames = ["base*.yaml", "include*.yaml", filenames].flatten.compact.uniq
```
rpherbig, line 906: What is `include*.yaml`?

asechrest: This was news to me, but we have the ability to load include files: https://discordapp.com/channels/745675889622384681/745675890242879671/935919107944095755

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
-    include_list.each do |include_file|
-      filepath = "./scripts/profiles/include-#{include_file}.yaml"
-      if File.exists?(filepath)
-        settings.merge!(safe_load_yaml(filepath))
+    # Waggle sets are maps of arbitrary names to map of spell names to map of spell properties.
+    # For example:
+    #   { 'burgle': { 'Refractive Field': { abbrev: 'RF' ... }, 'Death Monkey': { ... } } }
+    # Iterate each waggle's list of spells to enrich them.
+    # Enrich each's spell definition by merging base spell data with spell settings.
+    settings.waggle_sets.transform_values! do |spells_map|
```
rpherbig, line 1073: Methinks we have too many ways to define spells... not part of this PR, and super difficult to change without an upconverter for settings. Which is an interesting idea now that I mention it...

rcuhljr: I looked into settings converters back when I worked with making an in game script that would help you generate your settings file by watching you interact with your gear. The biggest issue is just that it likely wipes out all comments and doesn't respect ordering of the given yaml file, so they'll look kind of machine generated after processing.

rpherbig: Oof, yeah, good point.

rcuhljr: Most transforms were largely meant to be temporary, It might be time to start cutting some old formats off, or if we want to support certain quality of life short hands let the consuming scripts do that work instead of the settings file.

dependency.lic
rpherbig: Oh wow, I totally forgot about that. Could you add a comment explaining it since it's 1) not in the repo; 2) not used by a bunch of people; and 3) I know I'm going to forget again. Thanks

dependency.lic
rpherbig: Wait, can we only upconvert a specific setting (e.g. `offensive_spells`)? That might make things easier. Or make the conversion a separate script and ask people to run it and check the result.

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+      'forage_override_room',
+      'lockpick_room_id',
+      'outdoor_room',
+      'outfitting_room',
+      'prehunt_buffing_room',
+      'safe_room',
+      'safe_room_id',
+      'theurgy_prayer_mat_room'
+    ].each do |setting_name|
+      if settings[setting_name].is_a?(Hash)
```
rpherbig, line 1190: I might do this as a `.reject` then `.each`

rpherbig: Same thing for the other places the type matter, e.g. line 1108

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+        hometown_value = settings[setting_name][settings.hometown]
+        settings[setting_name] = hometown_value if hometown_value
+      end
+    end
 
     # Merge legacy appraisal settings into target appraisal settings
-    unless s.appraisal_training.include?('pouches')
-      s.appraisal_training.unshift('pouches') if s.train_appraisal_with_pouches
+    unless settings.appraisal_training.include?('pouches')
```
rpherbig, line 1197: I'd merge the conditionals

```suggestion
    if !settings.appraisal_training.include?('pouches') && settings.train_appraisal_with_pouches
        settings.appraisal_training.unshift('pouches')
    end
```

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
+      settings.appraisal_training.unshift('pouches') if settings.train_appraisal_with_pouches
```
rpherbig, line 1198: Why are we using `unshift` over `append`?

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
     end
-    unless s.appraisal_training.include?('gear')
-      s.appraisal_training.unshift('gear') if s.train_appraisal_with_gear
+    unless settings.appraisal_training.include?('gear')
```
rpherbig, line 1200: Similar to what @rcuhljr said, a good follow-up PR might be to start excising some of this legacy transformation logic. Not part of this PR, of course.

dependency.lic
KatoakDR: Can you clarify what "make sure we're correcting any submissions that are modifying settings"? Do you mean other humans who submit code changes to this file or something else? Thanks

dependency.lic
KatoakDR: > it's a map of filenames to FileInfo objects.

This, yep!

```suggestion
    @files_cache = {} # map of filenames => FileInfo objects
```

dependency.lic
KatoakDR: Will do

dependency.lic
```diff
@@ -772,257 +773,467 @@ class SlackbotManager
...
-      s.combat_spell_training[skill] = if spells[data['name']]
-                                         spells[data['name']].merge(data)
-                                       elsif spells[data['abbrev']]
-                                         spells[data['abbrev']].merge(data)
-                                       else
-                                         data
-                                       end
+    # Spell settings that are maps of spell names to map of spell properties.
+    # For example:
+    #   { 'Augmentation': { abbrev: 'MAF', mana: 5 ... } }
```
KatoakDR, line 1085:
```suggestion
    #   { 'Manifest Force': { abbrev: 'MAF', mana: 5 ... } }
```

dependency.lic
KatoakDR: I did some digging on the interwebs, seems that `append` as an alias for `unshift` was introduced in [Ruby 2.5](https://rubyreferences.github.io/rubychanges/2.5.html) so it may have been that the developer at the time either (a) wasn't yet on 2.5, or (b) was accustomed to using `unshift`.

I'll update them to `append` as that alias seems more intuitive now

dependency.lic
rcuhljr: Just an idle thought that we should be keeping an eye open for script submissions that are modifying the settings themselves.

dependency.lic
rcuhljr: So the latter, certainly, the former is basically where I stopped since you'd end up having to do manual text parsing. You'd basically need to read the existing yaml, convert, export that field to text, paste that text over the old block. But that means you need to be able to read and understand a yaml file, although we could do some crude approximations looking for the key and then snipping everything indented under it.

KatoakDR:
> Are there any performance implications from checking all files for changes in a fairly tight loop timer? Yes we're getting faster responses to load settings calls, but in practice those are far less than frequent than every tenth of a second so we're increasing our load by several orders of magnitude, just want to make sure it won't be an issue.

We haven't noticed any, but your question got us thinking about the need for `run_queue` background method checking for file changes constantly. I don't think we need to passively check behind the scenes since a check will be made when `get_data` or `get_settings` are explicitly called by a script on-demand. What I'm going to test out is doing a one-time load of the data/settings to get cached, and then if a player makes changes to their files then they'll only be reloaded into the cache the next time a script actually needs it. Assuming we don't notice any latency/lag with scripts picking up file changes that way, then we can eliminate that background process:

_Specifically talking about the following code where the loop calls `$setupfiles.run_queue`_
```ruby
class SetupFiles
  ... snip ...

  # Designed to be called by dependency's run loop.
  # Purpose is to check for file changes and reload them into cache.
  def run_queue
    reload_profiles(character_suffixes_to_filenames(['setup']))
    reload_data
  end

  ... snip ...

end

... snip ...

loop do
  if $turn_on_moon_watch && $moon_watch.nil?
    $moon_watch = MoonWatch.new(debug)
    $turn_on_moon_watch = nil
  end

  $manager.run_queue
  $setupfiles.run_queue # <<< we could remove this
  $bankbot.run_queue
  $reportbot.run_queue
  $slackbot.run_queue
  $moon_watch.run_queue if $moon_watch
  clear
  pause 0.1
end
```

rcuhljr:
That was what I meant, the old run_queue almost never actually ran much unless something else triggered it to actually do work. When you replaced it you moved it to a busy queue that's always checking, but it may be fine now to remove the queue entirely and just load it once during initialization and check files before opening them.

dependency.lic
KatoakDR: Added comments about base and include yamls

dependency.lic
KatoakDR: > Methinks we have too many ways to define spells...

haha, I came to the same conclusion after working on this!

dependency.lic
KatoakDR: If I understand correctly, then I came up with the following possible refactoring:
```ruby
    [
      ...,
      'safe_room_id',
      'theurgy_prayer_mat_room'
    ]
    .select do |setting_name|
      settings[setting_name].is_a?(Hash) &&
      settings[setting_name][settings.hometown]
    end
    .each do |setting_name|
      settings[setting_name] = settings[setting_name][settings.hometown]
    end
```

However, the more I think about it and take into consideration @rcuhljr comments about performance, this change doesn't sit right with me for two reasons:
1. there's now two loops, the `select` and the `each`, and
2. there's now two hometown lookups into the hash

Is there any performance concern for this script about that change? Or do you have an alternative refactoring you'd like other than what I shared above?

Thanks

dependency.lic
KatoakDR: Good catch, I hadn't realized the inner statement had yet another conditional 🤦

rpherbig:
> KatoakDR requested a review from rpherbig 2 days ago

@KatoakDR sorry for the delay, GitHub never notified me

dependency.lic
```diff
@@ -369,6 +370,8 @@ class ScriptManager
       respond "use '#{$lich_char}e check_dependency' to check the value."
       return
     end
+    # (Katoak) hack to avoid Lich overwritting my local script changes.
+    return if filename =~ /dependency/
```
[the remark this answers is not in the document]
KatoakDR: @rpherbig NO! Ugh, sorry

KatoakDR:
> sorry for the delay, GitHub never notified me

@rpherbig  No worries :)

_(and I removed my snafu line per https://github.com/rpherbig/dr-scripts/pull/5458/commits/1a7856e4c1866b5f932e9c5bc18b967593496212, thanks)_

PR #5464, opened 2022-01-24
[date][base-town]Adding repairer to SCC

MahtraDR:
As requested on Discord, adding repairer to SCC.

### DIFF
--- a/data/base-town.yaml
+++ b/data/base-town.yaml
@@ -630,6 +630,9 @@ Steelclaw Clan:
     id: 19399
   trader_outpost:
     id: 2874
+  metal_repair:
+    id: 9639
+    name: Granzer
 Stone Clan:
   currency: kronars
   shipment_clerk:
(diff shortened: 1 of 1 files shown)


data/base-town.yaml
```diff
@@ -630,6 +630,10 @@ Steelclaw Clan:
     id: 19399
   trader_outpost:
     id: 2874
+  metal_repair:
+    id: 9639
+    name: Granzer
+  deposit: *shard_deposit
```
douglascayers, line 636: To confirm, this doesn't need to be?
```yaml
deposit:
  << : *shard_deposit
```

KatoakDR, line 636: To confirm, this doesn't need to be?
```yaml
deposit:
  << : *shard_deposit
```

KatoakDR: Mahtra clarified this syntax also works in Discord, they were having issues posting reply to this comment.

PR #5483, opened 2022-02-01
[script][almanac]Adding startup delay to almanac

MahtraDR:
Without a startup delay, if you have almanac in autostarts you lose the mindstate when the game logs you in and clears mindstates from rested exp. Copied the syntax from tarantula.lic

### DIFF
--- a/almanac.lic
+++ b/almanac.lic
@@ -14,6 +14,9 @@ class Almanac
     @almanac_skills = settings.almanac_skills
     @almanac = settings.almanac_noun
 
+    @startup_delay = settings.almanac_startup_delay || 15
+
+    pause @startup_delay
     passive_loop
   end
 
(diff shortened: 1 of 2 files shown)


almanac.lic
```diff
@@ -14,6 +14,9 @@ class Almanac
     @almanac_skills = settings.almanac_skills
     @almanac = settings.almanac_noun
 
+    @startup_delay = settings.almanac_startup_delay || 15
```
rpherbig, line 17: The || isn't necessary since this is in base.yaml

rpherbig, line 17: This does not need to be an instance var (`@`)

MahtraDR: Fixed. Thanks again for your reviews!

MahtraDR: Fixed. Thanks again for your reviews!

PR #5488, opened 2022-02-02
[script][tarantula] Adding check if exists at startup, and adding common prefixes

MahtraDR:
I recently got a character that doesn't have a tarantula, and I wasn't going to invert my autostarts for one character. This script now checks if a tarantula exists, like almanac.lic checks for an almanac, and exits if none is found.

While there, I added some DR* prefixes as well.

### DIFF
--- a/tarantula.lic
+++ b/tarantula.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#tarantula
 =end
 
-custom_require.call(%w(common drinfomon))
+custom_require.call(%w(common common-items drinfomon))
 
 no_kill_all
 
--- a/tarantula.lic
+++ b/tarantula.lic
@@ -7,13 +7,16 @@ custom_require.call(%w(common drinfomon))
 no_kill_all
 
 class Tarantula
-  include DRC
 
   def initialize
     settings = get_settings
     @tarantula = settings.tarantula
-    @startup_delay = settings.tarantula_startup_delay || 15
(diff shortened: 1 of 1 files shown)


tarantula.lic
```diff
@@ -7,21 +7,24 @@ custom_require.call(%w(common drinfomon))
...
-  include DRC
 
   def initialize
     settings = get_settings
     @tarantula = settings.tarantula
-    @startup_delay = settings.tarantula_startup_delay || 15
-    @tarantula_noun = settings.tarantula_noun.nil? ? 'biomechanical tarantula' : settings.tarantula_noun
+    @startup_delay = settings.tarantula_startup_delay
+    @tarantula_noun = settings.tarantula_noun
+    unless DRCI.exists?(@tarantula_noun)
```
rpherbig, line 16: `common-items` needs to be added to `custom_require` near the top

MahtraDR: Do you use an IDE that quickly points that out to you, or do you just the The Eye for that kind of thing? I'm using vim for these, and seriously considering going RubyMine, or VSCode if they'll help me spot this kind of thing easier.

asechrest: > Do you use an IDE that quickly points that out to you, or do you just the The Eye for that kind of thing? I'm using vim for these, and seriously considering going RubyMine, or VSCode if they'll help me spot this kind of thing easier.

I use VSCode and it doesn't help me with this, but maybe I don't have it set up right.

I think @rpherbig is just damn good at this. Hah.

rpherbig: I don't use an IDE, just GitHub's diff interface

PR #5492, opened 2022-02-03
[script] [common-items] don't tap if nothing to tap

KatoakDR:
### Changes
* Updated to be consistent with the other methods that handle a `nil` argument (for example, when called `DRCI.tap(DRC.right_hand)` and there's nothing in your hand to avoid a "tap my " command)

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -398,6 +398,7 @@ module DRCI
   # Taps an item and returns the match string.
   # If no container specified then generically taps whatever's in your immediate inventory.
   def tap(item, container = nil)
+    return nil unless item
     from = container
     from = "from #{container}" if container && !(container =~ /^(in|on|under|behind|from) /i)
     DRC.bput("tap my #{item} #{from}", *@@tap_success_patterns, *@@tap_failure_patterns)


rpherbig:
I went back and forth on if the return value should be empty string, nil, a failure pattern, or something else. As far as I could tell, this function was only used here in common-items and nil seems fine, but let's keep an eye out for weird errors related to tapping.

>  As far as I could tell, this function was only used here in common-items

That said, in my search I did see a _lot_ of `DRC.bput("tap my ...", ...)` calls and I wonder if they should be replaced with `exists?`, `tap`, etc. helper function calls. That would be a good cleanup PR if you are interested.

KatoakDR:
> I went back and forth on if the return value should be empty string, nil, a failure pattern, or something else. As far as I could tell, this function was only used here in common-items and nil seems fine, but let's keep an eye out for weird errors related to tapping.

I had the same thoughts, and now I wished I had clarified my thinking process in the description. I went with `nil` in that specific method because it wasn't designed as a predicate function but rather to return the matched string. So if there's nothing to tap I went with `nil` over a boolean value.

> That said, in my search I did see a _lot_ of `DRC.bput("tap my ...", ...)` calls and I wonder if they should be replaced with `exists?`, `tap`, etc. helper function calls. That would be a good cleanup PR if you are interested.

Agree, I'll keep an eye out for those refactorings, thanks

PR #5494, opened 2022-02-03
[script][common-items] Adding pull to remove item success patterns

MahtraDR:
I've only added 'You slip" as a match pattern, but I also sorted the pattern list, which is why it looks like so many lines changed.

Added "You slip" to match this
```
 >wear my gloves
You slip some lumium lamellar gloves crafted from tempered bands onto your hands.
```

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -117,23 +117,23 @@ module DRCI
   ]
 
   @@remove_item_success_patterns = [
-    /^You remove/,
-    /^You detach/,
-    /^You sling/,
-    /^You slide/,
-    /^You take/,
-    /^You loosen/,
-    /^You untie/,
-    /^You pull/,
-    /^You tug/,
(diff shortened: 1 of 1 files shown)


KatoakDR:
I don't see the "^You slip" pattern in the code that the PR changed. I do see that they've been sorted. Is they match string perhaps missing from this PR?

rpherbig:
> I don't see the "^You slip" pattern in the code that the PR changed. I do see that they've been sorted. Is the match string perhaps missing from this PR?

Good catch!

MahtraDR:
This is what happens when you commit code at 2am (or that's my excuse anyway...) I actually had both 'You pull' and 'You slip' to add from my local copy, I managed to miss one and label this for the other, and then I misplaced a comma. Bad choices are made at 2am...

PR #5495, opened 2022-02-03
[script][performance] Reworked the whole non-worn instrument routine, and added an exit if…

MahtraDR:
… no instrument can be found.

And added mindstate_goal command line argument, so this supersedes/closes https://github.com/rpherbig/dr-scripts/pull/5177 as well.

### DIFF
--- a/performance.lic
+++ b/performance.lic
@@ -75,9 +75,8 @@ class Performance
           DRC.play_song?(@settings, song_list, is_instrument_worn, skip_clean)
         end
       when /Perhaps you should find somewhere drier/
-          wear_hand_armor_and_put_away_instrument
-          exit
-        end
+        wear_hand_armor_and_put_away_instrument
+        exit
       end
     end
   end
@@ -115,6 +114,7 @@ class Performance
     DRC.stop_playing
     Flags['ct-song'] = true
   end
+end
 
 before_dying do
   fput('stop play')
--- a/performance.lic
+++ b/performance.lic
@@ -75,7 +75,6 @@ class Performance
           DRC.play_song?(@settings, song_list, is_instrument_worn, skip_clean)
         end
       when /Perhaps you should find somewhere drier/
-        if done_training?
           wear_hand_armor_and_put_away_instrument
           exit
         end
--- a/performance.lic
+++ b/performance.lic
@@ -21,7 +21,7 @@ class Performance
 
     @settings = get_settings
     song_list = get_data('perform').perform_options
-    @mindstate_goal = args.mindstate_goal.to_i
+    @mindstate_goal = args.mindstate_goal.to_i || @settings.performance_mindstate_goal
     skip_clean = args.noclean
     check_song_only = args.checksong
 
@@ -83,12 +83,17 @@ class Performance
   end
 
   def remove_hand_armor_and_get_instrument
-    if DRCI.remove_item?(@hand_armor)
+    if DRCI.remove_item?(@hand_armor) && DRCI.stow_item?(@hand_armor)
       @hand_armor_removed = true
-      DRCI.stow_item?(@hand_armor)
+    else
+      DRC.message("Could not remove or stow #{hand_armor}, exiting")
+      exit
     end
     if DRCI.get_item?(@instrument)
       @instrument_out = true
+    else
+      DRC.message("Could not get #{@instrument}, exiting")
+      exit
     end
   end
 
@@ -103,11 +108,7 @@ class Performance
   end
 
   def done_training?
-    if @mindstate_goal
       DRSkill.getxp('Performance') >= @mindstate_goal
-    else
-      DRSkill.getxp('Performance') >= @settings.performance_mindstate_goal
-    end
   end
 
   def stop_play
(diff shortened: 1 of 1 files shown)


performance.lic
```diff
@@ -2,43 +2,54 @@
...
+	{ name: 'mindstate_goal', regex: /(\d+)/, optional: true, description: 'Train until mindstate >= mindstate_goal' }
       ]
     ]
 
     args = parse_args(arg_definitions)
 
     @settings = get_settings
-    @song_list = get_data('perform').perform_options
+    song_list = get_data('perform').perform_options
+    @mindstate_goal = args.mindstate_goal.to_i
```
rpherbig, line 24:
```suggestion
    @mindstate_goal = args.mindstate_goal.to_i || @settings.performance_mindstate_goal
```

rpherbig: And then line 107 can be more or less unchanged

performance.lic
```diff
@@ -52,37 +63,64 @@ class Performance
...
       end
     end
   end
 
+  def remove_hand_armor_and_get_instrument
+    if DRCI.remove_item?(@hand_armor)
+      @hand_armor_removed = true
+      DRCI.stow_item?(@hand_armor)
+    end
+    if DRCI.get_item?(@instrument)
```
rpherbig, line 91: Won't this try and get an instrument even if it shouldn't be removed (e.g. zills)? I don't see any guards in here

performance.lic
```diff
@@ -52,37 +63,64 @@ class Performance
...
         if done_training?
+          wear_hand_armor_and_put_away_instrument
           exit
         else
-          play_song?(@settings, @song_list, is_instrument_worn, skip_clean)
+          DRC.play_song?(@settings, song_list, is_instrument_worn, skip_clean)
         end
       when /Perhaps you should find somewhere drier/
-        exit
+        if done_training?
```
rpherbig, line 78: Why is this now checking if `done_training?` - it used to always exit. Is it even possible to keep training when this message is seen?

performance.lic
```diff
@@ -2,43 +2,54 @@
...
     @settings = get_settings
-    @song_list = get_data('perform').perform_options
+    song_list = get_data('perform').perform_options
+    @mindstate_goal = args.mindstate_goal.to_i
     skip_clean = args.noclean
     check_song_only = args.checksong
 
-    instrument = @settings.instrument
-    is_instrument_worn = instrument.nil?
+    worn_instrument = @settings.worn_instrument
```
rpherbig, line 28: I'm confused by the number of vars we're using here. Is my understanding right?
* `worn_instrument` - string, the instrument name which can be played while worn, optional
* `@instrument` - string, the instrument name which must be gotten before playing, optional
* `is_instrument_worn` - bool, whether or not the instrument can be played while worn, required
* `@instrument_out` - bool, whether or not we got an instrument, required
* `@hand_armor` - string, the hand armor to be removed if the instrument is _not_ worn, optional
* `@hand_armor_removed` - bool, whether or not the hand armor was removed, required

performance.lic
```diff
@@ -52,37 +63,64 @@ class Performance
...
     Flags['ct-song'] = true
   end
 end
 
 before_dying do
   fput('stop play')
-
   instrument = get_settings.instrument
-  DRC.bput("stow #{instrument}", 'You put') if instrument
+  DRCI.stow_item?(instrument) if instrument && DRCI.in_hands?(instrument)
```
rpherbig, line 123: Is this still necessary now that we're cleaning up in `wear_hand_armor_and_put_away_instrument`?

performance.lic
```diff
@@ -52,37 +63,64 @@ class Performance
...
+        if done_training?
+          wear_hand_armor_and_put_away_instrument
+          exit
+        end
       end
     end
   end
 
+  def remove_hand_armor_and_get_instrument
+    if DRCI.remove_item?(@hand_armor)
```
rpherbig, line 87: If either of these steps fail, should we exit the script?

performance.lic
MahtraDR: Would the command line arg take precedence over the yaml setting in that OR? My understanding is that it wouldn't then be clear which one we're using.

MahtraDR: Nevermind, @asechrest explained this to me.

performance.lic
MahtraDR: For the case when you kill performance manually, this makes sense, no?

rpherbig: I'm not sure. I mean, most scripts aren't designed to be killed at any arbitrary point in their execution (or at least, it's assumed they'll leave things in an unknown state). Most scripts don't try to clean up anything beyond Flags if they're killed. 

Should we re-wear our hand armor too? We don't have access to the instance var (I think), so it'd be a matter of either wearing it no matter what or checking if we're already wearing it and then trying to wear it?

performance.lic
```diff
@@ -52,37 +63,64 @@ class Performance
...
           exit
         else
-          play_song?(@settings, @song_list, is_instrument_worn, skip_clean)
+          DRC.play_song?(@settings, song_list, is_instrument_worn, skip_clean)
         end
       when /Perhaps you should find somewhere drier/
-        exit
+          wear_hand_armor_and_put_away_instrument
+          exit
+        end
```
rpherbig, line 80: Is this an extra `end`?

PR #5496, opened 2022-02-03
[script][combat-trainer] reworking the prayer mat routine to use commons

MahtraDR:
Preparing for a different fix for using portals as theurgy_containers, this came up. Works well now, and have tested it extensively on my cleric, as well as one other tester reporting success. Fingers crossed...

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2687,43 +2687,45 @@ class TrainerProcess
   end
 
   def pray_mat(game_state)
-    unless @prayer_mat && @theurgy_supply_container
-      message '*** Cannot run PrayerMat; need prayer_mat and theurgy_supply_container defined'
+    ready_to_mat = (
+      DRCI.exists?(@prayer_mat, @prayer_mat_container) &&
+      DRCI.exists?("wine", @theurgy_supply_container)
+      )
+
+    unless ready_to_mat
+      DRC.message "*** Cannot run PrayerMat; need prayer mat in prayer_mat_container or in theurgy_supply_container, and wine in theurgy_supply_container"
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2687,43 +2687,45 @@ class TrainerProcess
...
 
-    bput("roll #{@prayer_mat}",
-         'carefully gather up',
-         'need to be holding that first', 'not on the ground')
-    bput("put #{@prayer_mat} in my #{@prayer_mat_container}",
-         'You put', 'What were you referring to')
+    DRC.bput("kneel #{@prayer_mat}", 'You humbly kneel', 'You kneel in the water')
+    DRC.bput("kiss #{@prayer_mat}", 'You bend forward to kiss')
+
+    unless DRCI.get_item?("wine", @theurgy_supply_container)
```
rpherbig, line 2715:
```suggestion
    if DRCI.get_item?("wine", @theurgy_supply_container)
        DRC.bput("pour wine on #{@prayer_mat}", 'You quietly pour', 'Pour what', 'You set aside')
        DRCTH.empty_cleric_hands(@theurgy_supply_container)
    else
        DRC.message("*** Finishing PrayerMat early; Cannot get wine from theurgy_supply_container")
        @training_abilities.delete('PrayerMat')
    end
```
and then the shared cleanup only needs to be done once

MahtraDR: How's this version?

rpherbig: Looks good, except I don't think return on line 2721 should be there. Won't that short-circuit and not do the cleanup?

MahtraDR: Good catch! Thank you.

Issue #5244, opened 2021-11-21
combat trainer unable to get flint lighter from Eddy


Valherun:
coding for the meraud_commun method needs to be updated to work with eddies
starts around line 2620

```
fput "get #{@flint_lighter}", 'You get a', 'You are already'
    if bput("get incense from my #{@theurgy_supply_container}", 'You get', 'You are already', 'referring to') == 'referring to'
      fput "stow #{@flint_lighter}", 'You put your', 'Stow what'
      @equipment_manager.wield_weapon(game_state.weapon_name, game_state.weapon_skill)
      @training_abilities.delete('Meraud')
      return
    end
    fput 'light my incense with flint' # get messaging
    pause
    waitrt?
    fput "wave incense at #{checkname}"
    fput 'snuff incense'
    bput("put incense in my #{@theurgy_supply_container}", 'You put', 'What were you referring to')
    fput "stow #{@flint_lighter}", 'You put your', 'Stow what'
```
generic fputs here won't get the item from a portal, possibly upgrade to using equipment manager?

MahtraDR:
@Valherun I'm looking into this one now, if we make it use the function that looks in eddies, you'll have to specify the eddy as your theurgy container. Is that what you expect to happen here?

PR #5497, opened 2022-02-03
[script][combat-trainer] Refactored Meraud commune in combat-trainer

MahtraDR:
Tested this and the prayer mat together. Both work reliably. Fingers crossed.

Also resolves https://github.com/rpherbig/dr-scripts/issues/5244.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2648,7 +2648,7 @@ class TrainerProcess
       reset_ability(game_state, 'Meraud')
       return
     end
-    if /meraud/i =~ bput('commune sense', 'roundtime', 'Meraud', 'Meraud\'s')
+    if /meraud/i =~ DRC.bput('commune sense', 'roundtime', 'Meraud', 'Meraud\'s')
       reset_ability(game_state, 'Meraud', 60)
       game_state.blessed_room = true
       return
@@ -2656,25 +2656,25 @@ class TrainerProcess
     waitrt?
     @equipment_manager.stow_weapon(game_state.weapon_name)
     unless DRCI.exists?(@water_holder, @theurgy_supply_container,)
-      DRC.message "Cannot get #{@water_holder}, skipping Meraud commune"
+      DRC.message("Cannot get #{@water_holder}, skipping Meraud commune")
       @equipment_manager.wield_weapon?(game_state.weapon_name, game_state.weapon_skill)
-      @training_abilities.delete('Meraud') # so this does one...
+      @training_abilities.delete('Meraud')
       return
     end
     unless DRCTH.has_holy_water?(@theurgy_supply_container, @water_holder)
-      DRC.message "Cannot get holy water, skipping Meraud commune"
+      DRC.message("Cannot get holy water, skipping Meraud commune")
       @equipment_manager.wield_weapon?(game_state.weapon_name, game_state.weapon_skill)
-      @training_abilities.delete('Meraud') # so this does one...
+      @training_abilities.delete('Meraud')
       return
     end
     unless DRCTH.sprinkle_holy_water(@theurgy_supply_container, @water_holder, checkname)
-      DRC.message "Cannot sprinkle holy water, skipping Meraud commune"
+      DRC.message("Cannot sprinkle holy water, skipping Meraud commune")
       @equipment_manager.wield_weapon?(game_state.weapon_name, game_state.weapon_skill)
       @training_abilities.delete('Meraud')
       return
     end
     unless DRCTH.wave_incense?(@theurgy_supply_container, @flint_lighter, checkname)
-      DRC.message "Cannot get incense to wave, skipping Meraud commune"
+      DRC.message("Cannot get incense to wave, skipping Meraud commune")
       @equipment_manager.wield_weapon?(game_state.weapon_name, game_state.weapon_skill)
       @training_abilities.delete('Meraud')
       return
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2655,7 +2655,7 @@ class TrainerProcess
     end
     waitrt?
     @equipment_manager.stow_weapon(game_state.weapon_name)
-    unless DRCI.exists?(@water_holder, @theurgy_supply_container,)
+    unless DRCI.exists?(@water_holder, @theurgy_supply_container)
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2648,41 +2648,44 @@ class TrainerProcess
...
       return
     end
-    fput 'light my incense with flint' # get messaging
-    pause
-    waitrt?
-    fput "wave incense at #{checkname}"
-    fput 'snuff incense'
-    bput("put incense in my #{@theurgy_supply_container}", 'You put', 'What were you referring to')
-    fput "stow #{@flint_lighter}", 'You put your', 'Stow what'
+    unless DRCTH.sprinkle_holy_water(@theurgy_supply_container, @water_holder, checkname)
```
rpherbig, line 2670: So this does return a bool, but it's only the result of putting away the holy water (not getting or sprinkling). Maybe that's ok, or maybe `sprinkle_holy_water` needs to be changed. What do you think?

combat-trainer.lic
```diff
@@ -2648,41 +2648,44 @@ class TrainerProcess
...
       reset_ability(game_state, 'Meraud', 60)
       game_state.blessed_room = true
       return
     end
     waitrt?
     @equipment_manager.stow_weapon(game_state.weapon_name)
-    fput "get #{@water_holder} from my #{@theurgy_supply_container}"
-    if /that/i =~ bput("sprinkle #{@water_holder} on #{checkname}", 'You sprinkle', 'Sprinkle what?', 'Sprinkle that')
-      fput "put #{@water_holder} in my #{@theurgy_supply_container}"
+    unless DRCI.exists?(@water_holder, @theurgy_supply_container,)
```
rpherbig, line 2658: Can the trailing comma be removed?

combat-trainer.lic
rpherbig: To put that differently, there's some redundancy/duplication in these three checks.

[DRCI.exists?](https://github.com/rpherbig/dr-scripts/blob/master/common-items.lic#L390) taps the water holder to make sure it's in the container.

[DRCTH.has_holy_water?](https://github.com/rpherbig/dr-scripts/blob/master/common-theurgy.lic#L40) gets the water holder from the container (which is already a predicate function), checks if there is water inside, then puts the water holder away.

[DRCTH.sprinkle_holy_water](https://github.com/rpherbig/dr-scripts/blob/master/common-theurgy.lic#L145) get the water holder form the container (again), calls `sprinkle?`, then puts the water holder away (again).

-----

I feel like what we really want here is "were we able to sprinkle the holy water?", which is _almost_ what `sprinkle_holy_water` does on its own. My only concern is that `sprinkle_holy_water` returns whether or not the item was put away, _not_ whether or not we could sprinkle the holy water.

Now, all of that said, `sprinkle_holy_water` is only used in two places in `commune.lic`, so I think what I'd like is to make a `sprinkle_holy_water?` which returns a more meaningful value and use it.

What do you think about that?

combat-trainer.lic
```diff
@@ -2648,41 +2648,44 @@ class TrainerProcess
...
-    bput("put incense in my #{@theurgy_supply_container}", 'You put', 'What were you referring to')
-    fput "stow #{@flint_lighter}", 'You put your', 'Stow what'
+    unless DRCTH.sprinkle_holy_water(@theurgy_supply_container, @water_holder, checkname)
+      DRC.message("Cannot sprinkle holy water, skipping Meraud commune")
+      @equipment_manager.wield_weapon?(game_state.weapon_name, game_state.weapon_skill)
+      @training_abilities.delete('Meraud')
+      return
+    end
+    unless DRCTH.wave_incense?(@theurgy_supply_container, @flint_lighter, checkname)
+      DRC.message("Cannot get incense to wave, skipping Meraud commune")
```
rpherbig, line 2677: There's an opportunity to combine all of this cleanup logic, something like:

```suggestion
unless DRCTH.sprinkle_holy_water?(@theurgy_supply_container, @water_holder, checkname) && DRCTH.wave_incense?(@theurgy_supply_container, @flint_lighter, checkname)
      DRC.message("Cannot sprinkle incense  and wave incense, skipping Meraud commune")
      @equipment_manager.wield_weapon?(game_state.weapon_name, game_state.weapon_skill)
      @training_abilities.delete('Meraud')
      return
  end
```

What do you think about that?

combat-trainer.lic
MahtraDR: Actually, I found a bug in  [[DRCTH.has_holy_water?](https://github.com/rpherbig/dr-scripts/blob/master/common-theurgy.lic#L40)](https://github.com/rpherbig/dr-scripts/blob/master/common-theurgy.lic#L40) for the case where you can't get the water holder, but you can tap the water inside it. The function returns true, even though you can't get the holder to sprinkle yourself with. I'm finding a few weird odds and ends with common-theurgy, and more as I look.

Let me try and clean this up, but also work on common-theurgy on the side.

rpherbig:
Looks good ✔️ 

Remind me to merge this in a few days once the common* changes have been picked up by everyone

combat-trainer.lic
```diff
@@ -2648,41 +2648,27 @@ class TrainerProcess
...
       game_state.blessed_room = true
       return
     end
     waitrt?
     @equipment_manager.stow_weapon(game_state.weapon_name)
-    fput "get #{@water_holder} from my #{@theurgy_supply_container}"
-    if /that/i =~ bput("sprinkle #{@water_holder} on #{checkname}", 'You sprinkle', 'Sprinkle what?', 'Sprinkle that')
-      fput "put #{@water_holder} in my #{@theurgy_supply_container}"
+    unless DRCTH.sprinkle_holy_water?(@theurgy_supply_container, @water_holder, checkname) &&
+    DRCTH.wave_incense?(@theurgy_supply_container, @flint_lighter, checkname)
```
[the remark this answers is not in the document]
MahtraDR: Good point. Moved it to the same line with the && operator.

rpherbig (approved):

MahtraDR:
@rpherbig This should be good to merge.

I know I could, just checking first. :) If you're ok, I'll push the merge button to see what it looks like?

rpherbig:
I appreciate you not taking the initiative to merge it, I like keeping that power for myself.

That said, go ahead and merge this one to see it :)

PR #5501, opened 2022-02-03
[script][common-theurgy] Making has_holy_water? a tad more reliable.

MahtraDR:
This is for the case when your water holder is not in your theurgy container, but is somewhere you can tap, so you tap the water, without being able to get the container itself.

### DIFF
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -38,7 +38,7 @@ module DRCTH
   ] unless defined?(DEVOTION_LEVELS)
 
   def has_holy_water?(theurgy_supply_container, water_holder)
-    DRCI.get_item?(water_holder, theurgy_supply_container)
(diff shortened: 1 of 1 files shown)


common-theurgy.lic
```diff
@@ -38,7 +38,7 @@ module DRCTH
   ] unless defined?(DEVOTION_LEVELS)
 
   def has_holy_water?(theurgy_supply_container, water_holder)
-    DRCI.get_item?(water_holder, theurgy_supply_container)
+    return unless DRCI.get_item?(water_holder, theurgy_supply_container)
```
rpherbig, line 41: Should this be `return false unless ...` since the function is a predicate?

MahtraDR: I swear I am just making working for you, even if this is an amazing learning opportunity for me... Thank you!

rpherbig: I'm happy to help :)

PR #5502, opened 2022-02-03
[script][common-theurgy] First pass at sprinkle_holy_water?

MahtraDR:
First time making this kind of function... Does this work?

### DIFF
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -143,12 +143,12 @@ module DRCTH
   end
 
   def sprinkle_holy_water?(theurgy_supply_container, water_holder, target)
-    empty_cleric_hands(theurgy_supply_container)
     unless DRCI.get_item?(water_holder, theurgy_supply_container)
       DRC.message("Can't get #{water_holder}, skipping Meraud commune.")
       return false
     end
     unless sprinkle?(water_holder, target)
+      empty_cleric_hands(theurgy_supply_container)
       DRC.message("Couldn't sprinkle holy water, skipping Meraud commune.")
       return false
     end
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -145,11 +145,11 @@ module DRCTH
   def sprinkle_holy_water?(theurgy_supply_container, water_holder, target)
     empty_cleric_hands(theurgy_supply_container)
     unless DRCI.get_item?(water_holder, theurgy_supply_container)
-      DRC.message("Ccan't get #{water_holder}")
+      DRC.message("Can't get #{water_holder}, skipping Meraud commune.")
       return false
     end
     unless sprinkle?(water_holder, target)
-      DRC.message("Couldn't sprinkle holy water")
+      DRC.message("Couldn't sprinkle holy water, skipping Meraud commune.")
       return false
     end
     empty_cleric_hands(theurgy_supply_container)
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -142,10 +142,28 @@ module DRCTH
     DRCI.put_away_item?(DRC.left_hand, container)
   end
 
+  def sprinkle_holy_water?(theurgy_supply_container, water_holder, target)
+    did_sprinkle_holy_water = (
+      DRCI.get_item?(water_holder, theurgy_supply_container) &&
+      sprinkle?(water_holder, target) &&
+      DRCI.put_away_item?(water_holder, theurgy_supply_container)
+    (
+    return did_sprinkle_holy_water
+  end
+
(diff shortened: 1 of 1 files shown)


common-theurgy.lic
```diff
@@ -142,12 +142,30 @@ module DRCTH
     DRCI.put_away_item?(DRC.left_hand, container)
   end
 
+  def sprinkle_holy_water?(theurgy_supply_container, water_holder, target)
+    did_sprinkle_holy_water = (
+      DRCI.get_item?(water_holder, theurgy_supply_container) &&
+      sprinkle?(water_holder, target) &&
+      DRCI.put_away_item?(water_holder, theurgy_supply_container)
+    (
```
KatoakDR, line 150: 🧹  need closing paren
```suggestion
    )
```

KatoakDR: This looks to be doing exactly the same as the existing `sprinkle_holy_water` method but just returns a `true` if all three actions are true, else `false`.

Is it necessary to have the new method or simply add the `&&` to the existing methods?

MahtraDR: I don't intend to keep both. I'll clean up when we have these, and remove the non-booleans.

common-theurgy.lic
```diff
@@ -142,12 +142,30 @@ module DRCTH
     DRCI.put_away_item?(DRC.left_hand, container)
   end
 
+  def sprinkle_holy_water?(theurgy_supply_container, water_holder, target)
+    did_sprinkle_holy_water = (
```
rpherbig, line 146: Capturing the result and returning it isn't necessary unless you feel the var name adds expressiveness

The "Ruby way" would be
```suggestion
def sprinkle_holy_water?(theurgy_supply_container, water_holder, target)
      DRCI.get_item?(water_holder, theurgy_supply_container) && sprinkle?(water_holder, target) && DRCI.put_away_item?(water_holder, theurgy_supply_container)
end
```

(or add some newlines if you prefer that)

Ruby functions implicitly return the last thing in the function

MahtraDR: You and @asechrest said the same thing :) Great minds and all that. Let me swap that. I originally thought the variable name made things more explicit, but if the name of the function isn't explicit enough... that's a different problem.

MahtraDR: Like this?

common-theurgy.lic
```diff
@@ -142,12 +142,24 @@ module DRCTH
...
   def sprinkle_holy_water(theurgy_supply_container, water_holder, target)
     DRCI.get_item?(water_holder, theurgy_supply_container)
     sprinkle?(water_holder, target)
     DRCI.put_away_item?(water_holder, theurgy_supply_container)
   end
 
+  def sprinkle_holy_oil?(theurgy_supply_container, target)
+      DRCI.get_item?('holy oil', theurgy_supply_container) &&
+      sprinkle?('oil', target) &&
+      DRCI.put_away_item?('holy oil', theurgy_supply_container) if DRCI.in_hands?('oil')
```
rpherbig, line 160: Sorry, I didn't see this the first time I looked at the diff.

I don't think I've seen an `if` in a chained conditional like this... Have you tested this helper to make sure it puts away oil if it's in hand and skips (but returns true!) if not?

MahtraDR: How's the current version? I've tried to copy the format from `wave_incense?`

MahtraDR:
When we have this worked out, I'll make an oil version too.

common-theurgy.lic
```diff
@@ -142,6 +142,20 @@ module DRCTH
     DRCI.put_away_item?(DRC.left_hand, container)
   end
 
+  def sprinkle_holy_water?(theurgy_supply_container, water_holder, target)
+    empty_cleric_hands(theurgy_supply_container)
```
rpherbig, line 146: Why empty hand first?

common-theurgy.lic
```diff
@@ -142,6 +142,20 @@ module DRCTH
...
+    unless DRCI.get_item?(water_holder, theurgy_supply_container)
+      DRC.message("Can't get #{water_holder}, skipping Meraud commune.")
+      return false
+    end
+    unless sprinkle?(water_holder, target)
+      DRC.message("Couldn't sprinkle holy water, skipping Meraud commune.")
+      return false
```
rpherbig, line 153: Should we empty hands here?

common-theurgy.lic
MahtraDR: I thought we needed to, but we don't. It works just the same on the left hand. Tested just now.

common-theurgy.lic
MahtraDR: Added,

PR #5503, opened 2022-02-04
[script][common-theurgy] Making sprinkle? return true or false

MahtraDR:
Original code returns:

```
 >;has-holy-water                                                                                                                                                                                                                             
 --- Lich: has-holy-water active.                                                                                                                                                                                                             
 [has-holy-water]>sprinkle chalice on Floofles                                                                                                                                                                                                
 You sprinkle yourself with some holy water.                                                                                                                                                                                                  
 >                                                                                                                                                                                                                                            
 [has-holy-water: result is You sprinkle]                                                                                                                                                                                                     
 [has-holy-water: ]                                                                                                                                                                                                                           
 --- Lich: has-holy-water has exited.                        
```

new code returns:
```
 >;has-holy-water                                                                                                                                                                                                                             
 --- Lich: has-holy-water active.                                                                                                                                                                                                             
 [has-holy-water]>sprinkle chalice on Floofles                                                                                                                                                                                                
 You sprinkle yourself with some holy water.                                                                                                                                                                                                  
 >                                                                                                                                                                                                                                            
 [has-holy-water: true]                                                                                                                                                                                                                       
 --- Lich: has-holy-water has exited.     
```
or
```
 >;has-holy-water                                                                                                                                                                                                                             
 --- Lich: has-holy-water active.                                                                                                                                                                                                             
 [has-holy-water]>sprinkle chalice on Floofles                                                                                                                                                                                                
 Sprinkle that?  I don't think so.                                                                                                                                                                                                            
 USAGE: SPRINKLE (item) on (person|creature|item|ROOM)                                                                                                                                                                                        
 >                                                                                                                                                                                                                                            
 [has-holy-water: false]                                                                                                                                                                                                                      
 --- Lich: has-holy-water has exited.           
```

### DIFF
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -155,8 +155,12 @@ module DRCTH
   end
 
   def sprinkle?(item, target)
-    result = DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle what', 'What were you referring to')
-    result =~ /'You sprinkle'/
+    case DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle what', 'What were you referring to')
+    when /^You sprinkl/
+      return true
+    else
+      return false
+    end
   end
 
   def apply_jalbreth_balm(theurgy_supply_container, target)
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -155,8 +155,8 @@ module DRCTH
   end
 
   def sprinkle?(item, target)
-    case DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle what', 'What were you referring to')
-    when /^You sprinkl/
+    case DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle (what|that)', 'What were you referring to')
+    when /^You sprinkle/
       return true
     else
       return false
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -155,12 +155,8 @@ module DRCTH
   end
 
   def sprinkle?(item, target)
-    case DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle (what|that)', 'What were you referring to')
-    when /^You sprinkle/
-      return true
-    else
-      return false
-    end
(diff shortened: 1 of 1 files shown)


common-theurgy.lic
```diff
@@ -155,8 +155,12 @@ module DRCTH
   end
 
   def sprinkle?(item, target)
-    result = DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle what', 'What were you referring to')
-    result =~ /'You sprinkle'/
+    case DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle (what|that)', 'What were you referring to')
+    when /^You sprinkle/
```
rpherbig, line 159: This is one way to fix the problem. However, I think the original code was cleaner.

What do you think about keeping the original approach (using `=~`), but just fixing the bug in the regular expression (dropping the `''`, optionally adding the `^`)?

MahtraDR: Sorry, I was trying to wrap my head around it, and whilst I accept it's neater to have the two-liner... it's taken hours to wrap my head around what it does, or how to use it. @asechrest has very helpfully worked it out with me, and we think we have a good handle on it now, but honestly, that's iffy on my part.

My inclination is that all our predicates should be, well, predicates and return explicit true false. Even getting to where I can test this has been an effort. I'll make it how you're suggesting if you're set on it, but honestly, I prefer something more easily understood by someone with a lower skill level. Either the case bput, or even making it an `if` that explicitly returns true, would be much more accessible for folks coming into the project. 

Does that make sense?

rpherbig: I see your point re: explicit true/false return values. I'm not generally a fan of relying on ambiguous truthiness.

Clarity > conciseness, but I wonder if there's a way to get both. What do you think about this?

```
def sprinkle?(item, target)
    result = DRC.bput("sprinkle #{item} on #{target}", 'You sprinkle', 'Sprinkle what', 'What were you referring to')
    result == 'You sprinkle'
end
```

MahtraDR: Let me test it out a bit today. That makes a lot more sense to me than =~

MahtraDR: Tests right. Works for me.

PR #5506, opened 2022-02-04
[script][performance] Fix to_i conversion

asechrest:
Fixes to_i conversion causing unexpected behavior. Ruby converts a nil value to 0 when using to_i.

@rpherbig


asechrest:
Thanks @KatoakDR for helping me identify.

PR #5507, opened 2022-02-04
[script][performance] Second try at to_i fix

asechrest:
Flubbed the previous fix. Second try.

@rpherbig My apologies for pinging you so much today.


rpherbig:
I should've caught that :(

asechrest:
> I should've caught that :(

I'm totally docking your pay for the week! Oh wait....

No sweat. I should have caught it, too.

PR #5510, opened 2022-02-05
[script][common]Adding a common method for ouutputting atmospheric messages

MahtraDR:
I removed this method from tarantula.lic in the recent rewrite, and someone asked me to put it back in. Makes more sense here, than in the tarantula script itself. Thoughts?

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -750,6 +750,11 @@ module DRC
     DRC.bput("stance set #{stance}", /Setting your/)
   end
 
+  def atmo(text)
+    string = "<pushStream id=\"atmospherics\"/>" + text
+    _respond(string, "<popStream id=\"atmospherics\" /><prompt time =\"#{Time.now.to_i}\">&gt;</prompt>")
+  end
+
   def message(text)
     string = ''
     $fake_stormfront ? string.concat("\034GSL\r\n ") : string.concat("<pushBold\/>")


rpherbig:
I think it makes sense if any other scripts will want to do this.

MahtraDR:
Another option is to expand DRC.message to allow output to different windows. MoonWindow,percWindow, expWindow and so on.

rpherbig:
I don't really have a stake in this. What's your plan to move this PR forward?

MahtraDR:
Let me pin down @KatoakDR  today and see where we go. I like the idea of scalability, and expanding DRC.message, but that's also a bigger change. Too many things going. I may end up just putting the lines back and deferring this problem. :)

common.lic
```diff
@@ -757,6 +757,11 @@ module DRC
     DRC.bput("stance set #{stance}", /Setting your/)
   end
 
+  def atmo(text)
+    string = "<pushStream id=\"atmospherics\"/>" + text
+    _respond(string, "<popStream id=\"atmospherics\" /><prompt time =\"#{Time.now.to_i}\">&gt;</prompt>")
```
KatoakDR, line 762: 🧹 My stylistic suggestion is to format like this and avoid creating the `string` variable
```ruby
def atmo(text)
  _respond(
    "<pushStream id=\"atmospherics\"/>" + text,
    "<popStream id=\"atmospherics\" /><prompt time=\"#{Time.now.to_i}\">&gt;</prompt>"
  )
end
```

MahtraDR:
@rpherbig If the new code looks ok to you, we should be good to merge. Thanks @KatoakDR for the advice, and discussion!

KatoakDR (approved):
I tested with sending `DRC.atmo('hi')` and it sent the message to my atmospherics window in Genie.

PR #5511, opened 2022-02-06
[script][common-arcana]adding methods for checking symbiotic research and releasing magical research

Valherun:
update common-arcana to support 2 methods needed by upcoming researcher.lic and check-symbiosis.lic

PR #5480, opened 2022-02-01
[script] [common-healing] support tending internal bleeders

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5479

### Background
* Part of the solution for https://github.com/rpherbig/dr-scripts/issues/5440
* Courtesy of @Kaesken

### Changes
* Update `skilled_to_tend_wound?` method to check if you're skilled to tend an external or internal bleeder

PR #5512, opened 2022-02-06
[script] [researcher] [check-symbiosis] adding magical research scripts

Valherun:
Thanks to Crannach the original creator of researcher.lic on the in-game repo
Modified to add in all missing research options for normal and symbiotic research. Also added in guild specific research.

check-symbiosis.lic relies on research.lic and will make sure your symbiotic research is both active and the one you want.

Will require https://github.com/rpherbig/dr-scripts/pull/5511 to update common-arcana

### DIFF
--- a/check-symbiosis.lic
+++ b/check-symbiosis.lic
@@ -1,0 +1,97 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#check-symbiosis
+=end
+
+custom_require.call(%w[common common-arcana drinfomon spellmonitor])
+
+class SymbiosisCheck
+
+  def initialize
+    arg_definitions = [
+      [
+        {
+          name: 'symbiosis_type',
+          optional: true,
+          options: %w[activate avoid cast discern endure examine explore harness harvest heal impress learn perform remember resolve spell spring strengthen watch],
+          description: "Which symbiosis do you want to check? Argument -> Yaml -> defaults to harness"
+        }
+      ]
+    ]
+    args = DRC.parse_args(arg_definitions)
+
+    symbiosis_type = args.symbiosis_type || get_settings.symbiotic_research_type || 'harness'
+    validate_known_feat_for_symbiosis(symbiosis_type)
+    research_symbiosis(symbiosis_type)
+  end
+
+  def research_symbiosis(new_symbiosis)
+    current_symbiosis = perc_symbiotic_research # DRCA
+
+    # is the correct symbiotic research already up?
+    if current_symbiosis.downcase != new_symbiosis.downcase
+      # wrong symbiotic research is active, release it
+      if current_symbiosis
+        DRC.message("Wrong symbiotic research active, correcting")
+        DRC.message("Releasing: #{current_symbiosis}")
+        release_magical_research # DRCA
+      end
+      DRC.message("Researching: #{new_symbiosis}")
+      DRC.wait_for_script_to_complete('researcher', ['symbiosis', new_symbiosis])
+    else
+      DRC.message("Symbiotic research for '#{new_symbiosis}' is already active")
+    end
+  end
+
+  def validate_known_feat_for_symbiosis(symbiosis_type)
+    feat_type = get_feat_for_symbiosis(symbiosis_type)
+
+    if feat_type.nil?
+      DRC.message("Unknown symbiosis type: #{symbiosis_type}")
+      DRC.message("To add support for this symbiosis, please submit an issue on GitHub:")
+      DRC.message("https://github.com/rpherbig/dr-scripts/issues")
+      exit
+    end
+
+    # check that you have the correct spell feat
+    if !DRSpells.known_feats[feat_type]
+      DRC.message("You don't know the '#{feat_type}' feat for '#{symbiosis_type}' symbiosis")
+      DRC.message("See https://elanthipedia.play.net/Magical_feats for more info")
+      exit
+    end
+  end
+
+  def get_feat_for_symbiosis(symbiosis_type)
+    # different feats needed for different types
+    feat_for_symbioses = {
+      'Symbiotic Research' => ['activate', 'case', 'harness'],
+      'Mental Matrices' => ['discern', 'impress', 'remember', 'resolve'],
+      'Physical Matrices' => ['avoid', 'endure', 'spring', 'strengthen'],
+      'Scholar' => ['examine', 'learn', 'perform'],
+      'Survivalist' => ['explore', 'harvest', 'heal', 'watch']
+    }
+    # which feat is needed?
+    feat_for_symbioses
+      .select { |feat, symbioses| symbioses.include?(symbiosis_type.downcase) }
+      .map { |feat, symbioses| feat }
+      .first
+  end
+
+  # What, if any, symbiotic research is active?
+  # TODO move this method to common-arcana
+  def perc_symbiotic_research
+    case DRC.bput('perceive', /combine the weaves of the (\w+) symbiosis/, /Roundtime/)
+    when /combine the weaves of the (\w+) symbiosis/
+      Regexp.last_match(1)
+    else
+      nil
+    end
+  end
+
+  # TODO move this method to common-arcana
+  def release_magical_research
+    2.times { DRC.bput("release symbiosis", "Are you sure", "You intentionally wipe", "But you haven't") }
+  end
+
+end
+
+SymbiosisCheck.new
--- a/check-symbiosis.lic
+++ b/check-symbiosis.lic
@@ -28,7 +28,10 @@ class SymbiosisCheck
     current_symbiosis = DRCA.perc_symbiotic_research
 
     # is the correct symbiotic research already up?
-    if current_symbiosis.downcase != new_symbiosis.downcase
+    if current_symbiosis.downcase == new_symbiosis.downcase
+        DRC.message("Symbiotic research for '#{new_symbiosis}' is already active")
+        return
+    end
       # wrong symbiotic research is active, release it
       if current_symbiosis
         DRC.message("Wrong symbiotic research active, correcting")
--- a/check-symbiosis.lic
+++ b/check-symbiosis.lic
@@ -25,7 +25,7 @@ class SymbiosisCheck
   end
 
   def research_symbiosis(new_symbiosis)
-    current_symbiosis = perc_symbiotic_research # DRCA
+    current_symbiosis = DRCA.perc_symbiotic_research
 
     # is the correct symbiotic research already up?
     if current_symbiosis.downcase != new_symbiosis.downcase
@@ -33,7 +33,7 @@ class SymbiosisCheck
       if current_symbiosis
         DRC.message("Wrong symbiotic research active, correcting")
         DRC.message("Releasing: #{current_symbiosis}")
-        release_magical_research # DRCA
+        DRCA.release_magical_research
       end
       DRC.message("Researching: #{new_symbiosis}")
       DRC.wait_for_script_to_complete('researcher', ['symbiosis', new_symbiosis])
@@ -75,23 +75,6 @@ class SymbiosisCheck
       .map { |feat, symbioses| feat }
       .first
   end
-
-  # What, if any, symbiotic research is active?
-  # TODO move this method to common-arcana
-  def perc_symbiotic_research
-    case DRC.bput('perceive', /combine the weaves of the (\w+) symbiosis/, /Roundtime/)
-    when /combine the weaves of the (\w+) symbiosis/
-      Regexp.last_match(1)
-    else
-      nil
-    end
-  end
-
-  # TODO move this method to common-arcana
-  def release_magical_research
-    2.times { DRC.bput("release symbiosis", "Are you sure", "You intentionally wipe", "But you haven't") }
-  end
-
 end
 
 SymbiosisCheck.new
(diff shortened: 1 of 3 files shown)


Valherun:
Will require https://github.com/rpherbig/dr-scripts/pull/5511 as mentioned in the comments there. 

I swear I'll figure out PR's eventually

KatoakDR:
@Valherun thansk for putting these together. Would you please update this PRs title to be "[script] [researcher] add magical research script from Lich repo". And then in the PR description mention what other PRs this one depends on, for example like https://github.com/rpherbig/dr-scripts/pull/5480

Thanks

check-symbiosis.lic
```diff
@@ -0,0 +1,80 @@
...
+          name: 'symbiosis_type',
+          optional: true,
+          options: %w[activate avoid cast discern endure examine explore harness harvest heal impress learn perform remember resolve spell spring strengthen watch],
+          description: "Which symbiosis do you want to check? Argument -> Yaml -> defaults to harness"
+        }
+      ]
+    ]
+    args = DRC.parse_args(arg_definitions)
+
+    symbiosis_type = args.symbiosis_type || get_settings.symbiotic_research_type || 'harness'
```
KatoakDR, line 22: Just remembered, please add a new entry to base.yaml for `symbiotic_research_type: harness`

This ensures everyone gets the default and then here in code you can remove the `|| 'harness'` because the default is now defined in settings.

check-symbiosis.lic
```diff
@@ -0,0 +1,80 @@
...
+    # is the correct symbiotic research already up?
+    if current_symbiosis.downcase != new_symbiosis.downcase
+      # wrong symbiotic research is active, release it
+      if current_symbiosis
+        DRC.message("Wrong symbiotic research active, correcting")
+        DRC.message("Releasing: #{current_symbiosis}")
+        DRCA.release_magical_research
+      end
+      DRC.message("Researching: #{new_symbiosis}")
+      DRC.wait_for_script_to_complete('researcher', ['symbiosis', new_symbiosis])
```
KatoakDR, line 39: I wonder if we need to wait until that script to complete or simply start it and then let this script end naturally. 

If "check-symbiosis" doesn't need to still be running while "researcher" runs we could start the script via `start_script('researcher', ['symbiosis', new_symbiosis])`

The method "start_script" won't have a module prefix to it because that method is defined in core lich.rbw

It doesn't hurt to let this script wait, but it's also not necessary. 

Here's an example for reference, https://github.com/rpherbig/dr-scripts/blob/90f796710ed2ae3bc4812c182100df055ac92260/locksmithing.lic#L29

researcher.lic
```diff
@@ -0,0 +1,183 @@
+=begin
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#check-symbiosis
```
KatoakDR, line 2:
```suggestion
  Documentation: https://elanthipedia.play.net/Lich_script_repository#researcher
```

researcher.lic
```diff
@@ -0,0 +1,183 @@
...
+=end
+begin
+  custom_require.call %[common common-arcana drinfomon events]
+rescue
+  echo "Dependency is not running! This script requires dependency to be installed and running."
```
KatoakDR, line 7: Now that this script will be part of our Lich framework, lines 4-9 can be replaced by just the custom_require line; the rest aren't necessary

researcher.lic
```diff
@@ -0,0 +1,183 @@
...
+  exit
+end
+
+
+class Researcher
+  include DRC
```
KatoakDR, line 13: I know this is how the script was written originally, now that we're bringing it in then we're trying to refactor code to not use `include` statements. 

Please remove the include and then add the 'DRC.` prefix in the file for methods like bput or message or anything else that comes from common.lic

Thanks!

check-symbiosis.lic
Valherun: can do

check-symbiosis.lic
Valherun: Either method works, I was just more familiar with ``wait_for_script_to_complete``. But, since there is no reason to leave this running I'll switch.

researcher.lic
Valherun: wasn't sure there, so I just left it, will make the change

check-symbiosis.lic
Valherun: removed harness default from check-symbiosis.lic
added new base.yaml entry

researcher.lic
```diff
@@ -0,0 +1,178 @@
...
+    end
+  end
+
+  def convert_guild_skills(skill)
+    case skill
+      when 'astro', 'astrology', 'plane'
+        'plane'
+      when 'summon', 'summoning', 'planes'
+        'planes'
+      when 'therugy', 'road'
```
KatoakDR, line 41: Is there a typo in theurgy?

Valherun: good catch, fixing

asechrest:
@KatoakDR Will we have a collision with a same-named script from the other repo? I seem to recall something like that with `log.lic`.

check-symbiosis.lic
Valherun: > I wonder if we need to wait until that script to complete or simply start it and then let this script end naturally.
> 
> If "check-symbiosis" doesn't need to still be running while "researcher" runs we could start the script via `start_script('researcher', ['symbiosis', new_symbiosis])`
> 
> The method "start_script" won't have a module prefix to it because that method is defined in core lich.rbw
> 
> It doesn't hurt to let this script wait, but it's also not necessary.
> 
> Here's an example for reference,
> 
> https://github.com/rpherbig/dr-scripts/blob/90f796710ed2ae3bc4812c182100df055ac92260/locksmithing.lic#L29

reverting the changes from this.. realized it breaks t2

KatoakDR:
> @KatoakDR Will we have a collision with a same-named script from the other repo? I seem to recall something like that with `log.lic`.

Yes, the copy in this repo will overwrite what anyone has downloaded if they have the same name.

I don't think it should be an issue unless:
1. This version isn't backwards compatible with what's in Lich repo, or
2. The Lich repo version is actively maintained there but not here and the divergence will be hostile

KatoakDR:
> `wait_for_script_to_complete` is required as this script is meant to be used within t2.

@Valherun good catch, I hadn't thought about someone trying to run this script in t2 to block other actions

check-symbiosis.lic
```diff
@@ -0,0 +1,80 @@
...
+    symbiosis_type = args.symbiosis_type || get_settings.symbiotic_research_type
+    validate_known_feat_for_symbiosis(symbiosis_type)
+    research_symbiosis(symbiosis_type)
+  end
+
+  def research_symbiosis(new_symbiosis)
+    current_symbiosis = DRCA.perc_symbiotic_research
+
+    # is the correct symbiotic research already up?
+    if current_symbiosis.downcase != new_symbiosis.downcase
```
rpherbig, line 31: I have a _minor_ preference for turning this into a guard clause

```suggestion
    if current_symbiosis.downcase == new_symbiosis.downcase
        DRC.message("Symbiotic research for '#{new_symbiosis}' is already active")
        return
    end
```

researcher.lic
```diff
@@ -0,0 +1,178 @@
...
+          options: %w[
+            astrology astro plane
+            summoning summon planes
+            theurgy road
+          ],
+          description: "which guild skill?"
+        }
+      ]
+    ]
+    DRC.parse_args(arg_definitions)
```
rpherbig, line 173: I have a _minor_ preference for `parse_args` to always be the first (or second) thing done in initialize, for the sake of consistency across scripts. If we extracted the arrays into something named, do you think the args would be clear enough that we could move them?

researcher.lic
```diff
@@ -0,0 +1,178 @@
...
+      @current_topic = "symbiosis #{args.sym_type}" if args.symbiosis
+      @current_topic = convert_guild_skills(args.guild_skill) if args.guild
+      check_research
+      loop do
+        pause 5
+        check_research
+        break unless researching
+      end
+    else
+      echo "Using YAML settings is not supported at the moment, please specify a research target, or type ';researcher help' for a list of targets"
```
rpherbig, line 30: I have  _minor_ preference for turning this into a guard clause

```suggestion
    unless args.skill || args.symbiosis || args.guild_skill
      echo "Using YAML settings is not supported at the moment, please specify a research target, or type ';researcher help' for a list of targets"
      exit
  end
```

researcher.lic
```diff
@@ -0,0 +1,178 @@
...
+      set_researching(true)
+    when 'Usage:', 'You do not know how to research'
+      DRC.message("Invalid research option")
+      set_researching(false)
+      exit
+    end
+  end
+
+  def check_gaf
+    return if DRSpells.active_spells.include?('Gauge Flow')
```
rpherbig, line 129: Should we also check its duration? What happens if GAF wears off mid-research?

researcher.lic
```diff
@@ -0,0 +1,178 @@
...
+
+    research_start_fail_msgs = [
+      'You cannot begin',
+      'Usage:', 
+      'You do not know how to research'
+    ]
+
+
+    check_gaf
+    @current_topic = 'stream' if @current_topic == 'attunement'
```
rpherbig, line 114: I suggest moving this up with line 22 where `@current_topic` is assigned (then it can be treated as read-only for the rest of the script).

check-symbiosis.lic
```diff
@@ -0,0 +1,83 @@
...
+          name: 'symbiosis_type',
+          optional: true,
+          options: %w[activate avoid cast discern endure examine explore harness harvest heal impress learn perform remember resolve spell spring strengthen watch],
+          description: "Which symbiosis do you want to check? Argument -> Yaml -> defaults to harness"
+        }
+      ]
+    ]
+    args = DRC.parse_args(arg_definitions)
+
+    symbiosis_type = args.symbiosis_type || get_settings.symbiotic_research_type
```
jandersson, line 22: Just nitpicking but you could downcase it here as a sort of input processing stage. Then you don't have to do it downstream.

researcher.lic
Valherun: Sounds fine to me. Done

researcher.lic
Valherun: Not a bad idea, plus means that I need to learn about DRSpells.. So, bonus! :)

Will work on it here shortly

researcher.lic
```diff
@@ -0,0 +1,181 @@
+=begin
```
jandersson, line 1: Thanks for adding this to the repo!

researcher.lic
```diff
@@ -0,0 +1,180 @@
...
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#researcher
+=end
+
+custom_require.call(%w[common common-arcana drinfomon events])
+
+
```
KatoakDR, line 7: 🧹 Deleting extra blank line
```suggestion
```

researcher.lic
```diff
@@ -0,0 +1,180 @@
...
+      when *@moon_mage_research
+        'plane'
+      when *@warrior_mage_research
+        'planes'
+      when *@cleric_research
+        'road'
+    end 
+  end
+
+
```
KatoakDR, line 55: 🧹 Deleting extra line
```suggestion
```

Valherun:
@KatoakDR don't know why it won't let me comment on your example for comment for switching to buff.

The buff script would be good to switch to here. However, GAF is the only spell that needs to be used. There is no real "set" that research would use, so the only spell that would ever be available would be GAF and then you'd likely have people doing similar to this.

``` 
waggle_set:
  researcher: 
    << : *gaf
  gaf:
    << : *gaf
```

Or maybe I'm the only one that setups  my individual spells as waggles in addition to sets?

KatoakDR:
> Or maybe I'm the only one that setups my individual spells as waggles in addition to sets?

I do individual spells too, but my waggles are named after the full spell name and not the abbrev because I can't remember the abbrevs, so this still gives people some flexibility in naming their waggles sets.

> -- edit -- Was working on the edits here and realized that you do bring up a valid point when I was thinking about it. I could see a lower level cleric wanting to cast SAP to allow for slightly higher GAF casts for instance. (thus allowing slightly reduced research times)

Exactly! I should have clarified in my original comment that this enabled people to add more spells that may assist their researching. 

Thanks

researcher.lic
```diff
@@ -0,0 +1,180 @@
...
+  end
+
+  def researching
+    return false if Flags['research-partial']
+    return false if Flags['research-complete']
+    return false unless DRSpells.active_spells.include?('Gauge Flow')
+    UserVars.researcher['researching']
+  end
+
+  def add_flags
```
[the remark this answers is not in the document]
KatoakDR: @Valherun this comment is resolved but I didn't readily see the new "before_dying" block among the code changes. Is it perhaps not committed yet?

researcher.lic
KatoakDR: @Valherun this comment is marked resolved but the extra new line still appears in the file. Perhaps the commit hasn't been pushed? Thanks

Kiriawen:
@Valherun @MahtraDR 

Valherun, is this PR something you still intend to pursue or is it good to close out?

jandersson:
Or merge?

jandersson:
The original script has been deleted, the diff is just now introducing a whole new script. I will give this a viking funeral

jandersson:
image

PR #5514, opened 2022-02-07
Update trade.lic

GreyhallowDR:
tiny change but makes a huge difference with training for the few people using this script

When appraising a pouch solely for exp gain purposes it is always best to APPRAISE MY POUCH QUICK. This is true at any rank from 0-1750 for a pouch of any value. (26 years of Trader training to back this up)

### DIFF
--- a/trade.lic
+++ b/trade.lic
@@ -1341,7 +1341,7 @@ class Trade
   end
 
   def appraise_worn_pouch?
-    case bput("appraise my #{@gem_pouch_adjective} pouch", 'Roundtime', 'You can.t appraise the', 'Appraise what')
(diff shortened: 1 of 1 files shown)


KatoakDR (approved):
Wow, great catch! I did some testing myself and the appraising quick/normal/careful didn't seem to have any influence on the learning rate gained.

In fact, in one scenario the appraise quick gave me a +3 learning rate and the careful only did +2, ha!

MahtraDR:
Note that appraisal.lic already does `quick`

@GreyhallowDR it'd be awesome to have someone clean up our trader stuff... none of the active contributors play traders. :)

PR #5522, opened 2022-02-07
[script] [common] avoid race condition of drinfomon updating DRStats

KatoakDR:
### Background
* `drinfomon` will update `DRStats.encumbrance` to be the text value from `encumbrance` command
* We want `DRC.check_encumbrance` to give us the numeric result (0 to 11)
* Since `drinfomon` runs as a separate thread then there's a race condition between running the command and when DRStats gets updated where sometimes the method returns the outdated value

### Changes
* Don't wait for `drinfomon` before this method returns a value; go ahead and parse it to ensure we get the latest value

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -525,8 +525,15 @@ module DRC
   end
 
   def check_encumbrance(refresh = true)
-    bput('encumbrance', 'Encumbrance : .*') if refresh
-    $ENC_MAP[DRStats.encumbrance]
+    encumbrance = DRStats.encumbrance
+    if refresh
+      encumbrance_pattern = /(Encumbrance)\s:\s(?<encumbrance>.*)/
+      case bput('encumbrance', encumbrance_pattern)
+      when encumbrance_pattern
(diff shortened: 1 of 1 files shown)


common.lic
```diff
@@ -525,8 +525,15 @@ module DRC
...
-    $ENC_MAP[DRStats.encumbrance]
+    encumbrance = DRStats.encumbrance
+    if refresh
+      encumbrance_pattern = /(Encumbrance)\s:\s(?<encumbrance>.*)/
+      case bput('encumbrance', encumbrance_pattern)
+      when encumbrance_pattern
+        encumbrance = Regexp.last_match[:encumbrance]
+      end
+    end
+    $ENCUMBRANCE_MAP[encumbrance]
```
rpherbig, line 536: Should this be `$ENC_MAP`?

KatoakDR: Yes... forgot I renamed that variable locally 

```suggestion
    $ENC_MAP[encumbrance]
```

PR #5525, opened 2022-02-07
[script] [common-travel] add method to ask merchant for something

KatoakDR:
### Background
* While working on `hunting-buddy` script, saw an opportunity to create `DRCT.ask_for_item?(room, npc, item)` to consolidate logic for getting bundling ropes and gem pouches

### Changes
* Add `DRCT.ask_for_item?(room, npc, item)` method to complement the existing `buy_item`, `order_item` and `sell_item` methods

### DIFF
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -8,6 +8,8 @@ custom_require.call(%w[common common-items common-arcana events drinfomon])
 module DRCT
   module_function
 
+  # Visit a merchant and sell an item.
+  # Usually to sell junk at pawn shops.
   def sell_item(room, item)
     return false unless DRCI.in_hands?(item)
 
@@ -32,6 +34,8 @@ module DRCT
     end
   end
 
+  # Visit a merchant and buy an item by offering their asking price.
+  # Usually for restocking ammunition or other misc items.
   def buy_item(room, item)
     walk_to(room)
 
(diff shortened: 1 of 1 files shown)


common-travel.lic
```diff
@@ -69,6 +73,31 @@ module DRCT
...
+    ]
+
+    failure_patterns = [
+      /does not seem to know anything about that/,
+      /All I know about/,
+      /To whom are you speaking/,
+      /Usage: ASK/
+    ]
+
+    case DRC.bput("ask #{name} for #{item}", *success_patterns, *failure_patterns)
```
rpherbig, line 92: I still prefer the `==` syntax

```suggestion
    DRC.bput("ask #{name} for #{item}", *success_patterns, *failure_patterns) == 'hands you
```

But it's not a big deal

KatoakDR: I hear ya. A part of me likes the consistency of the `case/when` syntax, particularly for people new to the repo that are just looking for a "copy/paste" of something that's _similar_ to what they need to do. If these types of conditions follow the same format then it reduces some cognitive effort for a newbie to transform from the `==` to knowing they need either an `if/else` or a `case/when` and can see, "oh, I just need to add another `when` condition.

At least that's the fairy tale I tell myself

PR #5524, opened 2022-02-07
[script] [hunting-buddy] [7] ask for bundling rope

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5525
* https://github.com/rpherbig/dr-scripts/pull/5515
* https://github.com/rpherbig/dr-scripts/pull/5517
* https://github.com/rpherbig/dr-scripts/pull/5518
* https://github.com/rpherbig/dr-scripts/pull/5519
* https://github.com/rpherbig/dr-scripts/pull/5520
* https://github.com/rpherbig/dr-scripts/pull/5521

### Background
* Part of a series of updates I'm bringing to `hunting-buddy`
* As dependent PRs are merged then this one's diff will get smaller

### Changes
* Use `DRCT.ask_for_item?(room, npc, item)` to get bundling rope to reduce duplicate logic

PR #5526, opened 2022-02-07
[script] [hunting-buddy] [8] refactor parsing multiple hunt configs

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5524

### Background
* Part of a series of updates I'm bringing to `hunting-buddy`
* As dependent PRs are merged then this one's diff will get smaller
* The `format_hunting_info` method manipulates the hunting infos for a multi-hunt config with the goal of preventing the user from needing to leave their current room if the next hunt would be in the same zone. However, the implementation is not intuitive to follow to know how to add more hunting variables like `stop_on_no_moons` or `stop_on_encumbrance`; and it's manipulating many variables when the goal is to simply not search for a new room under certain conditions.

### Changes
* Remove `format_hunting_info` method
* Add new method `current_hunting_zone` to tell us which zone we're currently standing in so that we can use this info in subsequent hunting loop iterations to know if we should search for a new room or not

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -11,20 +11,33 @@ class HuntingBuddy
     args = parse_args(arg_definitions, true)
 
     @settings = get_settings
+
+    @town_data = get_data('town')[@settings.hometown]
+    @hunting_data = get_data('hunting')
+    @escort_zones = @hunting_data.escort_zones
+    @hunting_zones = @hunting_data.hunting_zones
+
     @stop_on_familiar_drag = @settings.stop_on_familiar_drag
-    data = get_data('hunting')
-    @escort_zones = data.escort_zones
-    @hunting_zones = data.hunting_zones
+    echo("  @stop_on_familiar_drag: #{@stop_on_familiar_drag}") if $debug_mode_hunting
+
     @stop_on_low_threshold = @settings.stop_on_low_threshold
+    echo("  @stop_on_low_threshold: #{@stop_on_low_threshold}") if $debug_mode_hunting
+
     @stop_on_high_threshold = @settings.stop_on_high_threshold
+    echo("  @stop_on_high_threshold: #{@stop_on_high_threshold}") if $debug_mode_hunting
+
     @magic_exp_threshold = @settings.magic_exp_training_max_threshold
     @magic_skills = ['Arcane Magic', 'Holy Magic', 'Life Magic', 'Elemental Magic',\
                      'Lunar Magic', 'Attunement', 'Arcana', 'Targeted Magic', 'Inner Fire',\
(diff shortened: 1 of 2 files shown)


hunting-buddy.lic
```diff
@@ -391,8 +531,13 @@ before_dying do
     stop_script(script_name) if Script.running?(script_name)
   end
   DRCA.release_cyclics(get_settings.cyclic_no_release)
-  Flags.delete('room-check')
-  Flags.delete('familiar_drag')
+  flags_to_delete = [
+    'hunting-buddy-room-check', # there are multiple flags with this prefix, one per friend
+    'hunting-buddy-familiar-drag'
+  ]
+  Flags.flags.keys
```
jandersson, line 538: This feels a bit over-engineered, if you compare it to the two lines on the left of the diff.

KatoakDR: Note that multiple dynamic flags get created for each buddy/friend in your yaml and the original "two lines" did not delete them all. This update ensures all the hunting buddy flags are deleted. Lingering flags consume resources because they are always trying to match incoming text to their patterns.

jandersson: Right understood, I missed that. Where are the buddy flags created?

jandersson: Looks like here? https://github.com/rpherbig/dr-scripts/pull/5526/files#diff-a6df990badd4d65a33f8e58f3337c06a688d7e8921a6029832f5fc79fbaf58d0R299

KatoakDR: > Looks like here? https://github.com/rpherbig/dr-scripts/pull/5526/files#diff-a6df990badd4d65a33f8e58f3337c06a688d7e8921a6029832f5fc79fbaf58d0R299

Yep, that's it! Yeah, I just stumbled upon it one day randomly checking what flags I had and noticed some residue so set out to clean 'em up. I say "because of performance" but it's probably 99% my OCD and 1% performance" =)

KatoakDR: > This feels a bit over-engineered, if you compare it to the two lines on the left of the diff.

And good to question these things, I've been known to over-engineer things and it helps to have a fresh pair of eyes review the code and solution approach.

Thanks

MahtraDR:
I've tested this one, so as to get all the PRs at once. This one tests very well. I'd be happy to merge it now.

hunting-buddy.lic
KatoakDR: Closing the loop, the deletion of the flags has been simplified as follows:
```ruby
  Flags.flags.keys
    .select { |flag_name| flag_name.start_with?('hunting-buddy-') }
    .each   { |flag_name| Flags.delete(flag_name) }
```

MahtraDR (approved):
Looks good to me. Happy to merge now?

PR #5533, opened 2022-02-08
[script][common-moonmage] Get/store divination tool predicates

MahtraDR:
It made more sense in my mind to break them up, because divination bones, and divination tools have a different format for storage... why keep things simple? ...

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-moonmage
 =end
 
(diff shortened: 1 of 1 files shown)


common-moonmage.lic
```diff
@@ -97,6 +97,26 @@ module DRCMM
     store_bones(storage)
   end
 
+  def get_div_tool?(tool)
+    if tool['tied']
+      DRCI.untie_item?(tool['name'], tool['container'])
```
rpherbig, line 102: `tool['tied']`?

PR #5534, opened 2022-02-08
[script] [bankbot] Deleting script from repo

MahtraDR:
Anyone can still get the script from the history, but we don't maintain it anymore either.


rpherbig:
Oof, this hurts

KatoakDR:
Why delete it?

Kaesken:
Perhaps put a warning and reference to the Simu policy as an echo when first starting it.  Then it becomes "Use at your own risk" without having to delete it.

MahtraDR:
I like @Kaesken's suggestion with a note that my overarching concern here is PR related. I want us not to be seen as condoning actions that are against Simu policy. 

Note that is deleting the file isn't the same as deleting from the repo history. Someone would still be able to come here and click through and find the file. It just means we can say we don't host, or support it anymore.

MahtraDR:
> Thinking about it more, my comment was maybe a bit more absolute than it strictly needed to be. Simu's change in policy that "killed" Bankbot was now considering it to violate the F2P restriction on coins deposited. So I guess if all characters are subscribed and the bankbot deposits regularly, it might still be ok.

Honestly, that's less concerning to me than what i thought might be the case. If it's not against Simu's rules for subscribed characters to use this, there is also a PR, or issue, asking us to track F2P status. Maybe we work on that, then keep bankbot, and add a check to not work on F2P chars. That would completely satisfy my concerns here.
 
> Part of me doesn't want to delete this since people _could_ still use it, and part of me wants to keep it because we (well, I) invested so much time in it. But I think it's probably best to delete it because it's a pretty risky script to be using. And if we do decide no one should be using it, we should delete it so it's one less thing for us to maintain (and the same probably goes for the common* functions only used here).

I'm sorry, I did not mean to push a button here. You have literally contributed more than all of us combined. Your opinion on this counts for everything. :)

asechrest:
I'm inclined to keep it just for sentimental value. :)

rpherbig:
What do you think about just adding a message at startup that the user needs to be aware of Simu's policy and make sure they are abiding by it, then pause the script until they unpause or kill it?

asechrest:
> What do you think about just adding a message at startup that the user needs to be aware of Simu's policy and make sure they are abiding by it, then pause the script until they unpause or kill it?
> 
> We do that with another dangerous script or two, I just can't remember the which ones offhand.

This gets my vote.

MahtraDR:
We just discovered that the player command gives reliable subscription data. How about restricting it to exclude f2p accounts once that is in drinfomon?

MahtraDR:
> We do that with another dangerous script or two, I just can't remember the which ones offhand.

We do that with burgle.lic and force users to use burgle start for it to actually run.

asechrest:
The diff is showing the entire file here.

MahtraDR:
> The diff is showing the entire file here.

Fixed. I also learned how to edit what commits are in a given PR. :)

asechrest:
I'm thinking something like this, like I put in `offload-items` to warn the user:

https://github.com/rpherbig/dr-scripts/blob/c5806e363df12f28c4e5db2830c1d51af1e06427/offload-items.lic#L29

Call a warn method (check end of file for method contents). Output the popbolded warning messages, advise them the script is pausing and to unpause to continue, or `;k` to quit.

MahtraDR:
> I'm thinking something like this, like I put in `offload-items` to warn the user:
> 
> https://github.com/rpherbig/dr-scripts/blob/c5806e363df12f28c4e5db2830c1d51af1e06427/offload-items.lic#L29
> 
> Call a warn method (check end of file for method contents). Output the popbolded warning messages, advise them the script is pausing and to unpause to continue, or `;k` to quit.

Very nice method. Thank you! How do we make that a common? DRC.warn_before_continuing(text) and it'll preface every line  with a `WARNING: ` and pause the script.

bankbot.lic
```diff
@@ -427,6 +438,9 @@ arg_definitions = [
 
 args = parse_args(arg_definitions)
 
+# Warn the script user that this is designed to offload items off your person
```
rpherbig, line 441: This comment seems a bit misleading

KatoakDR: Copy/paste error

MahtraDR: https://github.com/rpherbig/dr-scripts/pull/5551

Issue #5538, opened 2022-02-08
[script][dependency] - Issue aggregation for new dependency


asechrest:
New dependency was a big update. We've had some issues reported in Discord. Aggregating here. I will be fairly broad listing these issues. Some may end up being unrelated.

1. New dependency appears to enforce case sensitivity in yaml files. Many (most?) of us assumed old dependency already enforced `<YourChar>-setup.yaml` case sensitivity. That idea is supported by the sample files all adhering to that case, as well as the wiki yaml basics specifying that the first letter in the character name should be uppercase. But reports in Discord make it clear that folks had all lowercase yaml files and they were pulling without issue.
2. `drinfomon` is not properly pulling some components of character data. Confirmed with testing. Uninterrupted auto-loading of drinfomon shows that the guild and race, at minimum, are not being pulled. Some data _is_ being pulled, like circle. drinfomon then thinks you're a commoner and runs every 30 seconds. A manual `info` fixes the every 30 second spam.
3. `drinfomon` output is not being properly squelched. I believe this 99% to be unrelated to dependency, we just happen to be really noticing it since drinfomon is now running every 30 seconds. I can put in a PR for this.

Will add more as they come up.

asechrest:
Related to #2 above:

Reports on Discord that drinfomon isn't tracking mana properly until a forced manual `info`.

https://discord.com/channels/745675889622384681/745675890242879671/940699495006629950

asechrest:
Pulling `drinfomon` from a few PRs ago and testing it with _new_ dependency exhibits the same `drinfomon` issues. I'll confirm that behavior later.

If that's the case, we have some hiccup with the way new dependency interacts with, or handles, drinfomon. Also important to note that a manual pull of `info` _fixes_ the drinfomon data loading, so it's not necessarily that drinfomon is broken.

PR #5539, opened 2022-02-08
[script][drinfomon] Removing troublesome pauses

MahtraDR:
These four pauses (probably not all four, but none are needed, so removing all four) causes drinfomon not to populate info data, in what is the most obnoxious issue we have seen, I think.

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -411,7 +411,6 @@ class DRSkill
   def self.refresh_exp_mods
     @@silence_exp_mods_hook = true
     fput('exp mods')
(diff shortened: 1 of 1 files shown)


MahtraDR:
Also resolves part of https://github.com/rpherbig/dr-scripts/issues/5538

rpherbig:
Looks good to me ✔️

MahtraDR:
It's not a complete solution. It works for me, but it looks to be a race condition between the capturing loop and the game response to the fputs. We're working on a proper solution now.

PR #5540, opened 2022-02-09
[script][common]Removing the pipe from def message

MahtraDR:
Looking at the history, seems the pipe was a tad contentions even then https://github.com/rpherbig/dr-scripts/pull/3599

However, for the then-intended use case of debug messages, it made sense in a way.

Today we use DRC.message for all kinds of messages, and the pipe comes up as a topic quite a bit.

@rpherbig How would you feel about removing the pipe?


rpherbig:
> How would you feel about removing the pipe?

I have no opinion. I'm fine with whatever most people prefer.

KatoakDR:
+1 to get rid of the pipe

image

asechrest:
I'm ok to get rid of the

image

PR #5541, opened 2022-02-09
[script][common-moonmage] Duplicate moonie defs into common moonmage [1 of 3]

MahtraDR:
Copying these over from common-arcana, and building them so they can pass settings along.

Building up to being able to pass settings to these buffs to be able to store telescopes correctly when casting IOTS.

@KatoakDR Does this look right? It does work locally. :)


KatoakDR (commented):
Looks good to me

Issue #1450, opened 2017-03-08
Add assess_teach and check_listening to common


brycole:
I think it would be a good idea to assess_teach and check_listening methods to common (or some form of them). I pulled them from crossing-training and stripped some stuff out to make them work in a before: script.

It should allow a bit more leverage for other scripts and anyone who is creating a training method outside crossing-training.

```
 def assess_teach
    case DRC.bput('assess teach', 'is teaching a class', 'No one seems to be teaching', 'You are teaching a class')
    when 'No one seems to be teaching', 'You are teaching a class'
      waitrt?
      return {}
    end
    results = reget(20, 'is teaching a class')
    waitrt?

    results.each_with_object({}) do |line, hash|
      line.match(/(.*) is teaching a class on (.*) which is still open to new students/) do |match|
        teacher = match[1]
        skill = match[2]
        # Some classes match the first format, some have additional text in the 'skill' string that needs to be filtered
        skill.match(/.* \(compared to what you already know\) (.*)/) { |m| skill = m[1] }
        hash[teacher] = skill
      end
    end
  end


  def check_listening
    return if DRRoom.pcs.empty?
    classes = assess_teach
    @last_teacher = classes
                    .reject { |t, s| t.nil? || s.nil? }
                    .sort_by { |_t, s| [DRSkill.getxp(s), DRSkill.getrank(s)] }
                    .find { |t, _s| DRC.listen?(t) }
                    .first

    @listen_timer = @last_teacher ? nil : Time.now
  end

```

skyleronken:
@intangibles trying making a fork, branch, and test against dependent scripts. If it works fine, submit it as a PR and lets get it committed!

intangibles:
Will do! I don't quite know how to fork or branch yet, bit of a learning curve still with GIT still.

jandersson:
Just as an update, `assess_teach` was moved to `common`. But `check_listening` is still in `crossing-training`

MahtraDR:
@asechrest Would you be keen on this one?

MahtraDR:
I initially started looking into adding check_listening and have since decided against it. This issue is five years old now and we couldn't think of a good use case for this today. 

Closing.

PR #5542, opened 2022-02-09
[script][common.lic] Adding check_listening into common.lic. 

MahtraDR:
Resolves https://github.com/rpherbig/dr-scripts/issues/1450

Can clean up other from craft and crossing training after.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -471,6 +471,18 @@ module DRC
     end
   end
 
+  def check_listening
+    return if DRRoom.pcs.empty?
+    classes = assess_teach
+    @last_teacher = classes
+                    .reject { |t, s| t.nil? || s.nil? }
+                    .sort_by { |_t, s| [DRSkill.getxp(s), DRSkill.getrank(s)] }
+                    .find { |t, _s| listen?(t) }
+                    .first
+
+    @listen_timer = @last_teacher ? nil : Time.now
+  end
+
   def hide?(hide_type = 'hide')
     unless hiding?
       case bput(hide_type, 'Roundtime', 'too busy performing', 'can\'t see any place to hide yourself', 'Stalk what', 'You\'re already stalking', 'Stalking is an inherently stealthy', 'You haven\'t had enough time', 'You search but find no place to hide')


common.lic
```diff
@@ -471,6 +471,18 @@ module DRC
     end
   end
 
+  def check_listening
+    return if DRRoom.pcs.empty?
+    classes = assess_teach
+    @last_teacher = classes
```
KatoakDR, line 477: common isn't a class with instance variables. Need to remove the `@`

to confirm, `assess_teach` is a method defined in common or drinfomon? I'm unfamiliar with it either way.

common.lic
```diff
@@ -471,6 +471,18 @@ module DRC
...
+                    .reject { |t, s| t.nil? || s.nil? }
+                    .sort_by { |_t, s| [DRSkill.getxp(s), DRSkill.getrank(s)] }
+                    .find { |t, _s| listen?(t) }
+                    .first
+
+    @listen_timer = @last_teacher ? nil : Time.now
```
KatoakDR, line 483: Same comment as above about instance variables. This line should be removed. If tracking is helpful then the calling script probably should handle it is my preference to help keep the common methods stateless.

common.lic
rpherbig: Agreed re: removing the instance vars. I suggest `@last_teacher` be owned by the calling script and returned from `check_listening`. The same with @listen_timer`.

MahtraDR: Thanks for the feedback. Updated the def.

KatoakDR (commented):
Code looks good to me, thanks

What testing has been done to verify the change?

common.lic
```diff
@@ -471,6 +471,16 @@ module DRC
     end
   end
 
+  def check_listening(last_teacher)
+    return if DRRoom.pcs.empty?
+    classes = assess_teach
+    last_teacher = classes
```
rpherbig, line 477: Is this assigning to the input var? I thought Ruby passed this by value, not ref. In either case, for clarity I would much prefer it to be returned instead.

KatoakDR: > I thought Ruby passed this by value, not ref.
It does, I think this is still residue from when it was copied from the original file.

I think this can become:
```ruby
# Find the first teacher who is teaching a class
# for a skill we need to prioritize learning then listen.
classes
  .reject { |teacher, skill| teacher.nil? || skill.nil? }
  .sort_by { |teacher, skill| [DRSkill.getxp(skill), DRSkill.getrank(skill)] }
  .find { |teacher, skill| listen?(teacher) }
```

MahtraDR:
@KatoakDR @rpherbig I'm sorry. I'm just going to trash this. 

I immensely appreciate the time and effort you two make in these reviews, and I feel I'm wasting your time with this PR.

I've been thinking a lot about the idea that sometimes we're making changes that aren't perhaps needed, and the more I stare at the code here, the more I think that applies here. 

The issue this is meant to resolve is from 2017, and hasn't had anyone clamouring for this since it was opened.I'm inclined to close the issue as stale, and leave it at that.

rpherbig:
> I've been thinking a lot about the idea that sometimes we're making changes that aren't perhaps needed, and the more I stare at the code here, the more I think that applies here.

Better you realize it now than after we spend even more time changing/maintaining! 😄 

It's a sign of an advanced developer that they recognize when changes _aren't_ needed.

> I immensely appreciate the time and effort you two make in these reviews, and I feel I'm wasting your time with this PR.

No worries, sometimes this happens and it's perfectly fine.

PR #5543, opened 2022-02-09
[script][common-arcana] Calling our DRCMM functions from here [2 of 3]

MahtraDR:
Conveniently has settings to pass to the DRCMM functions too.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -3,7 +3,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#common-arcana
 =end
 
-custom_require.call(%w[common common-items events spellmonitor drinfomon])
+custom_require.call(%w[common common-items common-moonmage events spellmonitor drinfomon])
 
 $MANA_MAP = {
   'weak' => %w[dim glowing bright],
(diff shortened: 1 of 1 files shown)


KatoakDR:
For ease of reference, for multi step PRs would you mind adding the dependent PRs in the description. For this one it depends on https://github.com/rpherbig/dr-scripts/pull/5541#pullrequestreview-877643485

Example where I do this is with the hunting-buddy PRs that are open right now. 

Thanks!

rpherbig:
I was moving too fast and merged the wrong PR, sorry!

rpherbig:
Why not also delete DRCA's version of the function in this PR?

KatoakDR:
> I was moving too fast and merged the wrong PR, sorry!

@MahtraDR other pro tip with multi step PRs is to put into draft mode the ones that are waiting on other PRs to merge so that it's also clearer to folks merging that conditions aren't yet ready. :)

MahtraDR:
> Why not also delete DRCA's version of the function in this PR?

I plan to do that after all these are merged. Just in case I miss something, I at least won't break existing things that way.

Issue #2653, opened 2018-05-07
Track if a character is f2p or not.


Sarvatt:
I was thinking this could be done at login with drinfomon, something like doing an inv check and matching /[13]00 items/ or teach open and matching You don't have access to teaching classes because you don't have a subscription. The inv check might be preferred because it's actually useful info to have spammed at login.

A few uses would be automatically using the xala climb_practice target if you are f2p instead of being sent to undergondola at 540 ranks, and not taking the trails to other provinces in ;scouting

KatoakDR:
`drinfomon` now identifies your account subscription in `UserVars.subscription`, thanks

PR #5490, opened 2022-02-03
[script] [drinfomon] track `exp mods`

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5458

### Background
* I would like `skill-recorder` to be able to leverage my actual `exp mods` amounts for Athletics checks
* I want to take advantage of my full skill buffs that are above the default 10% that `skill-recorder` awards

### Changes
* Add a new downstream hook that parses `exp mods` using same design pattern I did for `spellmonitor`
* Add `DRSkill.getmodrank(skill)` method that returns the effective ranks for that skill (I almost named the method `geteffrank` but that didn't look good to me, so went with `getmodrank` since the command is `exp mods`).
* If you are not under the influence of any modifiders then `getmodrank` will return same value as `getrank`
* Add new setting `drinfomon_passive_delay` that is the number of seconds to wait between checks. Previously 30 seconds was hardcoded.

### Config
_base.yaml_
```yaml
# The number of seconds the `drinfomon` script waits
# between checking your info and experience.
drinfomon_passive_delay: 30
```

### Tests
_athletics without modifiers_
```
> exp mods
The following skills are currently under the influence of a modifier:

+84(20%) Perception (506 effective ranks)

> ,e echo DRSkill.getrank('Athletics') ; echo DRSkill.getmodrank('Athletics')

--- Lich: exec1 active.

[exec1: 493]

[exec1: 493]

--- Lich: exec1 has exited.
```
_buff athletics_
```
> khri flight

Knowing that a dose of paranoia is healthy for any aspiring Thief, your mind fixates on every possible avenue of escape available to you.
Roundtime: 3 sec.

(wait the passive delay)

> ,e echo DRSkill.getrank('Athletics') ; echo DRSkill.getmodrank('Athletics')

--- Lich: exec1 active.

[exec1: 493]

[exec1: 571]

--- Lich: exec1 has exited.

> exp mods
The following skills are currently under the influence of a modifier:

+78(16%) Athletics (571 effective ranks)
+84(20%) Perception (506 effective ranks)
```
_release buff_
```
> khri stop flight

You attempt to relax your mind from some of its meditative states.
You feel mentally fatigued as your heightened paranoia ceases to enhance your knowledge of nearby escape routes.

(wait the passive delay)

> ,e echo DRSkill.getrank('Athletics') ; echo DRSkill.getmodrank('Athletics')

--- Lich: exec1 active.

[exec1: 493]

[exec1: 493]

[...]

PR #5548, opened 2022-02-09
[script] [drinfomon] refactor downtsream hooks for performance

KatoakDR:
### Background
* Some mix of the https://github.com/rpherbig/dr-scripts/pull/5458 and https://github.com/rpherbig/dr-scripts/pull/5490 updates merged recently has highlighted a race condition with how `drinfomon` parses info.
* Fixes https://github.com/rpherbig/dr-scripts/issues/5538

### Changes
* Refactored the downstream hooks that parse `info`, `exp all`, `exp mods`, and `played` to all follow the same pattern.
* Switched from parsing `ltb info` to `played` for easier parsing
* Created distinct variables for each hook to control whether that hook should or should not squelch output without interferring with what the other hooks are doing.
* Reworked regular expressions and used named captured groups where possible to simplify extracting data.
* If you're playing The Fallen, then `UserVars.premium` will be set to `'true'` for you (requested by Briarleigh)
* Scripts can now detect your subscription level via `UserVars.subscription`. The values can be F2P, Basic, Premium or Platinum

Many thanks to @MahtraDR @asechrest for helping triage and troubleshoot with players on Tuesday to bring this PR together.


MahtraDR:
Note this closes https://github.com/rpherbig/dr-scripts/issues/2653 when merged

KatoakDR:
> Note this closes #2653 when merged

Thanks. What does `played` command output for standard subscription and F2P subscriptions?

asechrest:
> I'm not really sure how to review this one. @KatoakDR @asechrest @MahtraDR ? Is "we thoroughly tested it" the merge condition?

This is a completely fair question. What caught us on the `dependency` update is, I think, too many moving parts too close together. After begging for testers I think we had 6 folks willing to test dependency. After some tweaks by Katoak via testing it was running beautifully. But, my belief is most of us stopped testing and then `drinfomon` had a (quite reasonable) PR submitted that played unexpectedly poorly with new dependency, and that's what caught us.

All that being said, these are necessary PRs to move forward. Dependency's performance was degrading as the project grew due to race conditions, size of yamls, etc. And drinfomon's PRs are good ones, too.

I think we just test this one as much as we can and see where we're at. But, Robert, we still appreciate your look-through. I know this is a full refactoring so it's not easy.

drinfomon.lic
```diff
@@ -340,6 +335,8 @@ class DRSkill
       skill.exp = exp.to_i
       skill.percent = percent.to_i
       skill.current = rank.to_i + (percent.to_i / 100.0)
+    else
+      DRSkill.new(name, rank, exp, percent)
```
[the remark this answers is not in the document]
KatoakDR: Yeah, it's a strange paradigm for sure. The constructor creates a new instance **and then** adds it to a `@@list` class variable. So at this point, we don't care about the return value because the constructor is already adding the reference to a list that's available class-wide.

```ruby
  def self.update(name, rank, exp, percent)
    skill = self.find_skill(name) # this returns the record from `@@list`
    if skill
      # updates the variable properties
    else
      DRSkill.new(name, rank, exp, percent) # adds a new instance to `@@list`
    end
end
```

KatoakDR:
> I'm not really sure how to review this one. @KatoakDR @asechrest @MahtraDR ? Is "we thoroughly tested it" the merge condition?

Great question. I agree with @asechrest answer.

Additionally, I've put out a [request for testers](https://discord.com/channels/745675889622384681/745675890242879671/941043081971662868) in Discord

If we don't have any takers by end of the day then we may need to move forward with whatever testing we can do ourselves. I believe this is still a step in the right direction even if it turns out this isn't 100% perfect.

MahtraDR:
> > Note this closes #2653 when merged
> 
> Thanks. What does `played` command output for standard subscription and F2P subscriptions?

Premium (I think you have this)
```
> Note this closes #2653 when merged

Thanks. What does `played` command output for standard subscription and F2P subscriptions?
```
Standard
```
 >played                                                                                                                                                                                                                                      
 Account Info for REDCACTED:                                                                                                                                                                                                                
   Current Account Status: Basic     
```
f2p
```
 >played                                                                                                                                                                                                                                      
 Account Info for REDCACTED:                                                                                                                                                                                                                 
   Current Account Status: F2P                                                                                                                                                                                                                
         Basic Start Date: 11/10/2020
```

MahtraDR:
> I'm not really sure how to review this one. @KatoakDR @asechrest @MahtraDR ? Is "we thoroughly tested it" the merge condition?

This will be long. I'm sorry.

So, "we" didn't test drinfomon that went live. Or, at least *I* didn't.

I did test dependency. And what I say I tested dependency... I run 4 accounts normally, can stretch to 8. I ran 8 accounts, across 10 characters for 3 weeks. We found early on that buff_nonspells was bugged for non-magic-users. We found no other meaningful bugs during this time. 

We also had some other power users running, folks with 10+ characters. We had 3 of these. They reported no issues.

even with ~50 total yamls running that dependency, we still missed the capitalization issue. Namely, windows file operations aren't case sensitive, and ALL of these yamls were properly capitalized. So dependency went live, and we had missed the fact some windows users have downcased yaml names, and this isn't something we ever considered. I'd say this was well tested. No testing is perfect.

when we had bescort and the skates snafu. I tested on my 6, Katoak tested on his, Gildaren tested on his. It hit live, and it just blew up. We found out we had a lot of success/failure patterns missing, and despite all our testing, the logic for putting on/taking off skates was still flawed when it hit live. I'd still say the original was well tested...

The predicate stuff I created, specially the telescope stuff, I test on one character (the moonie in the case of the telescope) and accept that building up the list of success/failure patterns is a matter of time, and I won't have them all. I'd say this is reasonably well tested. 

So, coming back to this PR. I ran this overnight on my accounts. We have already found that the regex matches for skills aren't quite there yet. I think soak testing on 5-10 accounts over 24-48 hours is "well tested" by any standard really, and we just have to accept sometimes things won't get caught even then. That said, specially for large changes, like drinfomon/dependency, maybe we should make a point of having a slightly more prolonged testing process. 2-3-4-5 days with a good sample size would make sense to me. 

Sorry for the wall of text.

KatoakDR:
> So, "we" didn't test drinfomon that went live. Or, at least _I_ didn't.

I test everything I put out, but my usage doesn't cover all possible play styles or configs.

MahtraDR:
> > So, "we" didn't test drinfomon that went live. Or, at least _I_ didn't.
> 
> I test everything I put out, but my usage doesn't cover all possible play styles or configs.

I mentioned this in Discord as well. I worded that poorly. I was not trying to suggest you didn't test. Just to make clear the massive testing effort was around dependency, and not around drinfomon.

rpherbig:
I want to be absolutely clear that I'm proud of the testing work y'all have done on this and other PRs (dependency, drinfomon, bescort/skates, etc.). Given the complexity of this system, we'll never catch everything. And, frankly, it wouldn't be worth the time investment given how many person-hours it would take. This is a free/volunteer hobby about a game, and people understand we need to push a fix every now and then.

All I want to make sure is that we're giving these changes the same thought as we have those PRs, and it sounds like y'all are on it. When you say this has been tested sufficiently, I believe you. :)

asechrest:
Request from Briarleigh in Discord, which is a good one:

> Briarleigh | Leecya — Today at 4:37 PM
> So,  while ya'll are tinkering with drinfomon and dependency,  for folks in TF can you have it auto set the premium var to true?  ;vars set premium=true  Fang Cove is open to non-premiums here and that gates them from being able to use go2 without manually setting it to true

drinfomon.lic
```diff
@@ -939,11 +1097,33 @@ exp_mods_check_time = Time.now - exp_mods_check_interval
 # Waiting until now to silence the script avoids that ambiguity.
 silence_me
 
+# Now that the hooks are registered,
+# do an initial parse of all the info!
+
+unless dead?
+  squelch_info_output = !UserVars.drinfomon_debug
+  fput('info')
+  pause 0.5
```
KatoakDR, line 1106: @MahtraDR @asechrest You know, I don't know the purpose of the pauses either. Worth testing with and without them

KatoakDR:
> I want to be absolutely clear that I'm proud of the testing work y'all have done on this and other PRs (dependency, drinfomon, bescort/skates, etc.). Given the complexity of this system, we'll never catch everything. And, frankly, it wouldn't be worth the time investment given how many person-hours it would take. This is a free/volunteer hobby about a game, and people understand we need to push a fix every now and then.
> 
> All I want to make sure is that we're giving these changes the same thought as we have those PRs, and it sounds like y'all are on it. When you say this has been tested sufficiently, I believe you. 😄

Thanks @rpherbig! We appreciate your support and for laying down the groundwork for us to have Lich as a fun, safe place to tinker after "work" 😉

drinfomon.lic
MahtraDR: Without pauses it seemed to work better for me, but it's still race condition/latency based, I think.

drinfomon.lic
```diff
@@ -2,6 +2,8 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#drinfomon
 =end
 
+$DRINFOMON_VERSION = '2.0.0'
```
KatoakDR, line 5: @MahtraDR you'd be proud!

drinfomon.lic
KatoakDR: I've removed the pauses, no negative effects "yet", and things are speedier :)

asechrest (commented):
Confirming fix of race condition squelch leaking for `exp mods`.

Before:
image

After:
image

KatoakDR:
@rpherbig this is ready for merge, thanks

MahtraDR:
We have tested this very widely and feel very confident in this version. 

Fingers crossed. 😀

PR #5552, opened 2022-02-09
[data][base-hunting] Adding cliff rooms as a separate area

MahtraDR:
Adding by request.

We considered just adding the rooms to the original archons entry, but chose against it because these rooms act a bit weird (which is why they were left out in the first place.) 

Just giving folks a few more options.

Also sorted the archons entry numerically since I was there anyway.


rpherbig:
I like adding it as a separate hunting zone

PR #5559, opened 2022-02-11
[script][common.lic, tarantula.lic] Re-added bolding

BinuDR:
I added a method to bold text into common.lic and added the command into the atmo command in tarantula.lic so that the tarantula use message is bolded again.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -764,6 +764,10 @@ module DRC
     )
   end
 
+  def bold(text)
+    result = "<pushBold/>#{text}<popBold/>"
+  end
+
   def message(text)
     string = ''
     $fake_stormfront ? string.concat("\034GSL\r\n ") : string.concat("<pushBold\/>")
--- a/tarantula.lic
+++ b/tarantula.lic
@@ -116,7 +116,7 @@ class Tarantula
 
     case DRC.bput("rub my #{@tarantula_noun}", /The .* comes alive in your hand/, /(\d+) roisae?n to generate enough venom/, /But you currently aren.t learning any/, /You need to vary which skillset/, /You should stop practicing your Athletics/)
     when /The .* comes alive in your hand/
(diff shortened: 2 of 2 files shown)


MahtraDR (approved):
Ran the changed code, works as intended. The two blank spaces I looked at twice, but I'm not fussy on that point. 

bold doesn't look any different visually than DRC.message, but the way we have DRC.message, I couldn't nest them, sadly. So adding bold it is. (I tested all variations of nesting and this is the only version that worked for me)

BinuDR:
> Ran the changed code, works as intended. The two blank spaces I looked at twice, but I'm not fussy on that point.
> 
> bold doesn't look any different visually than DRC.message, but the way we have DRC.message, I couldn't nest them, sadly. So adding bold it is. (I tested all variations of nesting and this is the only version that worked for me)

I know, that's why I did it this way in the first place.

BinuDR:
2 blank spaces?  Do you mean `DRC.atmo( DRC.bold...) )` those around the DRC.bold statement?

It's for readability.

rpherbig:
This would need to be 2 PRs to avoid breaking things for users of `tarantula.lic`

BinuDR:
> This would need to be 2 PRs to avoid breaking things for users of `tarantula.lic`

common.lic first and then tarantula a few days later?

rpherbig:
Yep. Sounds good!

rpherbig:
Unless you wanted to make `atmo` do the bold itself. I don't know if atmo should _always_ bold, though.

BinuDR:
> Unless you wanted to make `atmo` do the bold itself. I don't know if atmo should _always_ bold, though.

I think it's good the way it is.  Here are the new PRs.
Common: https://github.com/rpherbig/dr-scripts/pull/5564
Tarantula: https://github.com/rpherbig/dr-scripts/pull/5565

Issue #3937, opened 2019-10-02
Sigil Location Crowd Sourcing


GreyhallowDR:
Has the idea been floated to crowd source mapping sigil locations with Lich? I'm assuming there is a way to have scripts report back info to a central database right?

rpherbig:
Yes, that's how stealing works today. W could do something similar for
sigils.

GreyhallowDR:
exactly, something like base-stealing.yaml file

Sarvatt:
i think E would need to create a new db on firebase for it, is there enough bandwidth for that? we were hitting the free limits but not sure how it is after disabling the trader shop tracking

Sarvatt:
the base stealing yaml file is manually populated from the db on firebase that gets updated in real time. steal.lic updates the data. it would take someone who cares a lot of work to do this, just a heads up it wont be easy if you want to implement it :)

rpherbig:
I'm pretty sure I can make the new fire base data store if someone wanted
to run with this.

MahtraDR:
Closing as resolved by #5562

PR #5562, opened 2022-02-11
[script][sigilrecorder]Sigil mapper

Nyctus:
Creates a YAML which stores the location and season of sigils for use by other scripts.


sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+	end #check_room_data
+	
+	def unstuck_me(target_room)
+		exit_list = Map[target_room].wayto
+		exit_list.each do |rnum, dir|
+		  Flags.add('room-detected', /Room Number: \d+/)
+		  fput("peer #{dir}")
+		  pause
+		  if Flags['room-detected']
+		    fput('retreat')
```
MahtraDR, line 108: We have a nice common for retreating DRC.retreat. Can use that here instead of the fput? (search and replace all)

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+
+class SigilRecorder
+  include DRC
+  include DRCA
+  include DRCI
+  include DRCT
+
+	def initialize
+		require 'yaml'
+		@data = YAML.load(File.read("./scripts/data/base-sigils.yaml"))
```
MahtraDR, line 11: The path won't work like this, because the script will live in scripts/ already. Change to data/base-sigils.yaml?

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+			respond("Travel Debug: #{travel_result} | Room Target: #{roomid}")
+			if Room.current.id == nil
+			  respond("Adding #{roomid} to nogo due to lack of visible room number in LOOK.")
+			  add_to_nogo(roomid)
+			  unstuck_me(roomid)
+			  next
+			end
+			if travel_result == true
+				@current_room = Room.current.id
+				DRCA.do_buffs(@settings, 'outdoors')
```
MahtraDR, line 32: As an example, since you're prefacing do_buffs with DRCA. the include DRCA isn't necessary above. 

We prefer no includes, and explicit method calls like this one, since it makes life easier for someone coming after you, reading the script, to know exactly where a method comes from. 

This is just a long note to say, this line is good. :)

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+		@settings = get_settings
+		get_region
+		get_season #make sure we're recording sigils to the correct season
+		region_room_list #generate the walk_list array
+		generate_nogo_list
+		
+		Flags.add('danger-room', /.*begins to advance on you.*|You weave back and forth in a dance-like motion|You must be closer to use tactical abilities on your opponent/)
+		@walk_list.each do |roomid|
+			@data['nogo'].include?(roomid) ? next : (travel_result = walk_to(roomid, false))
+			respond("Travel Debug: #{travel_result} | Room Target: #{roomid}")
```
MahtraDR, line 23: This is just a suggestion, but check out DRC.message. It'll make the text easier to read than respond alone does.

I personally prefer DRC.message over respond, but that's just a preference thing on my part.

MahtraDR (changes requested):
Hey mate, thanks for submitting this. I know you've been working on it for awhile now. It's a great idea! Thanks for contributing!

sigilrecorder.lic
rpherbig: Try `get_data('sigils')`

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+		@data = YAML.load(File.read("./scripts/data/base-sigils.yaml"))
+		@city_list = ['shard', 'crossing', 'riverhaven', 'other']
+		@sigil_list = ['abolition','congruence','induction','permutation','rarefaction', 'antipode','ascension','clarification','decay','evolution','integration','metamorphosis','nurture','paradox','unity']
+		@settings = get_settings
+		get_region
+		get_season #make sure we're recording sigils to the correct season
+		region_room_list #generate the walk_list array
+		generate_nogo_list
+		
+		Flags.add('danger-room', /.*begins to advance on you.*|You weave back and forth in a dance-like motion|You must be closer to use tactical abilities on your opponent/)
```
rpherbig, line 20: Please prefix the flag name with something unique to this script

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+		@walk_list.each do |roomid|
+			@data['nogo'].include?(roomid) ? next : (travel_result = walk_to(roomid, false))
```
rpherbig, line 22:
```suggestion
			next if @data['nogo'].include?(roomid)
			travel_result = walk_to(roomid, false)
```

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+			respond("Travel Debug: #{travel_result} | Room Target: #{roomid}")
+			if Room.current.id == nil
```
rpherbig, line 24: Is `roomid` always the current room? What if travel failed (e.g. `travel_result` is false)?

sigilrecorder.lic
rpherbig, line 22: Why check if the roomid is in `nogo` if you do this on line 177?

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+  include DRC
+  include DRCA
+  include DRCI
+  include DRCT
+
+	def initialize
+		require 'yaml'
+		@data = YAML.load(File.read("./scripts/data/base-sigils.yaml"))
+		@city_list = ['shard', 'crossing', 'riverhaven', 'other']
+		@sigil_list = ['abolition','congruence','induction','permutation','rarefaction', 'antipode','ascension','clarification','decay','evolution','integration','metamorphosis','nurture','paradox','unity']
```
rpherbig, line 13: We already have this list defined in https://github.com/rpherbig/dr-scripts/blob/6d087f0ebb4dcf449ab854ace7afe0b5f0d884b0/attunement.lic#L14 - is there a way to avoid duplication?

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+	def get_season
+		/^It is currently (\w+) and it is.*$/ =~ bput('time', /^It is currently.*/)
+		@season = $1
+	end #get_season
+	
+	
+	def get_region
+		case Room.current.image.downcase
+		when /ilithi|shard/
+		  @region = 'shard'
```
rpherbig, line 63: Rather than settings instance vars (`@`), I'd much prefer the function to return a value that can then be passed along later. It helps keep track of where vars are coming from, instead of having a bunch of instance vars that anything can modify.

What do you think?

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+		respond("Unsuitable room for sigil harvesting. Adding Room #{banroom} to no-go list.")
+		Flags.reset('danger-room')
+		pause 0.2
+	end #add_to_nogo
+	
+	
+	def generate_nogo_list
+	  respond("Pruning walk list to avoid combat. Current list size: #{@walk_list.length}")
+	  nogo_list = []
+	  @avoids = get_data('hunting').hunting_zones.values
```
rpherbig, line 174: This var doesn't appear to be used

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+		@data['nogo'].uniq
+		respond("Unsuitable room for sigil harvesting. Adding Room #{banroom} to no-go list.")
+		Flags.reset('danger-room')
+		pause 0.2
+	end #add_to_nogo
+	
+	
+	def generate_nogo_list
+	  respond("Pruning walk list to avoid combat. Current list size: #{@walk_list.length}")
+	  nogo_list = []
```
rpherbig, line 173: This array is never populated

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+		  @data[@region][sigil][@season].push(@current_room)
+		  @data[@region][sigil][@season].uniq
+		  #respond("Adding to YAML: #{@data['shard'][sigil][@season]}")
+		next
+		end
+		
+		File.open("./scripts/data/base-sigils.yaml", "w") { |file| file.write(@data.to_yaml) }
+	end #write_to_file
+	
+end #SigilFinder
```
rpherbig, line 198: Should this be `SigilRecorder`? Incidentally, this is why I'm not a fan of the end comments - they're too easy to get stale.

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
...
+			if Room.current.id == nil
+			  respond("Adding #{roomid} to nogo due to lack of visible room number in LOOK.")
+			  add_to_nogo(roomid)
+			  unstuck_me(roomid)
+			  next
+			end
+			if travel_result == true
+				@current_room = Room.current.id
+				DRCA.do_buffs(@settings, 'outdoors')
+				if check_room_data < 2
```
rpherbig, line 33: What is the significance of "2" here?

data/base-sigils.yaml
```diff
@@ -0,0 +1,5503 @@
...
+    - 386
+    - 387
+    - 390
+    summer: 
+    winter: 
+  unity:
+    spring: 
+    summer: 
+    winter: 
+other:
```
KatoakDR: What is `other` section for? There doesn't seem to be any rooms listed. Is this necessary in the file?

data/base-sigils.yaml
```diff
@@ -0,0 +1,5503 @@
...
+    - 14075
+    - 14077
+    - 14078
+    summer: 
+    winter:
+    - 825
+    - 717
+    - 1040
+    - 1041
+riverhaven:
```
KatoakDR: 🧹  Would you mind spelling the town names the same was as they are listed in https://github.com/rpherbig/dr-scripts/blob/master/data/base-town.yaml as a consistency thing?

sigilrecorder.lic
KatoakDR: Suggestions:
`sigil-danger-room`
`sigilrecorder-danger-room`

sigilrecorder.lic
Nyctus: I was seeing weirdness when using get_data and then trying to write to the file, but I think I might've been using the wrong syntax to write to the file when paired with get_data. I'll look a few things up and see if I can get it working with get_data instead

sigilrecorder.lic
Nyctus: I am removing the includes and using the DRCx explicit calls. I like it better as then I know where to look if I need to understand how a function works :)

sigilrecorder.lic
```diff
@@ -0,0 +1,205 @@
+custom_require.call(%w[common common-arcana common-items common-travel drinfomon events spellmonitor equipmanager])
+
+class SigilRecorder
+  include DRC
```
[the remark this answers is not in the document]
Nyctus: Swapping everything to explicit calls

sigilrecorder.lic
Nyctus: added 'sigil-" to all my flags

data/base-sigils.yaml
Nyctus: updated the capitalization of the cities

data/base-sigils.yaml
Nyctus: I can probably get rid of it. The idea was to have a bucket to put rooms in if the script was run in a weird place that wasn't detected as one of the main three cities/regions I am currently supporting.

The better solution might be, for now, to only support Crossing, Shard, and Riverhaven, and only when the associated map image file references either the city name or province name.

sigilrecorder.lic
Nyctus: I am not making a distinction between primary or secondary sigils because when you need a sigil you don't really care which it is and you still have to spam 'perc sigil' until the one you want appears. I could generate the list used by my script by adding the two constants together. That way if new sigils are added in the future my script should automatically start saving data for the new sigils.

Nyctus: Looking at Attunement I am also not sure how to convert those regex patterns to arrays. Arrays are what I am working with. Ideally it would be nice to automagically update my script if new sigils are added, though.

sigilrecorder.lic
Nyctus: Oops. I think I changed my approach here and forgot to update this line. Going to test it without this check, relying only on the pruned list.

Nyctus: Based on the other comment regarding this line I am removing this check entirely

sigilrecorder.lic
Nyctus: Makes sense, I'll see if I can untangle this and have get_region return the region name instead.

MahtraDR:
Just a note that this resolves https://github.com/rpherbig/dr-scripts/issues/3937

sigilrecorder.lic
Nyctus: Figured out my synax issues and am now using get_dat

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+	end #check_room_data
+	
+	#attempts to navigate out of a room lacking important data
+	def unstuck_me(target_room)
+		#Check the exits to the room we think we're in in Mapdb. Store the exits in an array then loop through it.
+		exit_list = Map[target_room].wayto
+		exit_list.each do |rnum, dir|
+		  #flag to listen for room numbers appearing in 'peer' info
+		  Flags.add('sigil-room-detected', /Room Number: \d+/)
+		  bput("peer #{dir}", /You peer/, /You see/, /\[Usage:/)
```
asechrest, line 145: This one needs a DRC prefix.

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		add_to_nogo(target_room, @data)
+		exit
+	end #unstuck_me
+	
+	#uses 'perc sigil' repeatedly until both sigils are discovered
+	def find_sigils
+		room_sigils = []
+		while room_sigils.length < 2
+			#weave is a 0 RT way of detecting enemies in the room. It will auto face enemies but will not harass players or NPCs. After weaving check if the danger-room flag has tripped
+			bput('weave', /You weave back and forth in a dance-like motion/, /You must be closer to use tactical abilities on your opponent/, /There is nothing else to face/)
```
asechrest, line 166: This one needs a DRC prefix.

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+			if Flags['sigil-danger-room']
+			  add_to_nogo(Room.current.id, @data) unless Room.current.id == nil
+			  return nil
+			end 
+
+			#main 'perc sigil' bput and case statement. When sigils are detected they're added to an array
+			case line = bput('perceive sigil', @sigil_list, 'Roundtime', /^You recall.*/, /^Having recently been searched/, /You are too distracted/, /Something about|This is not an appropriate place for that/)
```
asechrest, line 173: This one needs a DRC prefix.

sigilrecorder.lic
Nyctus: IDK how those got mixed after the find all but I did it a second time and then double-checked every bput in the script. (I think I just have too many tabs open in VSCode )

sigilrecorder.lic
Nyctus: Fixed

sigilrecorder.lic
Nyctus: Fixed

MahtraDR (approved):
All my comments have been addressed. Looks good to me.

KatoakDR:
🧹 @Nyctus would you mind replacing the "tabs" with 2 spaces in the file for consistency with the other scripts? thanks!

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		#pull data and settings and store the current region and season
+		@data = get_data('sigils')
+		@settings = get_settings
+		@region = get_region #parse Room.current.image for which region/city we're in
+		@season = get_season #make sure we're recording sigils to the correct season
+		
+		#generate an array of all the room numbers associated with the current room's Room.current.image value
+		region_list = region_room_list
+		#subtract known hunting spots and any other rooms in the YAML's nogo list to avoid combat and rooms with known mapping issues
+		walk_list = generate_nogo_list(region_list)
```
KatoakDR, line 25: ❓ Is the variable `walk_list` the rooms NOT to travel to or the rooms TO travel? I'm not sure because the method is named `generate_nogo_list`

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		
+		#debug info
+		DRC.message("Region detected: #{@region}")
+		DRC.message("Season detected: #{@season.capitalize}")
```
KatoakDR, line 29: 💡 Should `get_season` method be responsible for capitalizing the value? I notice that you don't call that method on `@region` in the line above.

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		#flag known patterns that suggest the room is unsuitable for sigil hunting or harvesting
+		Flags.add('sigil-danger-room', /.*begins to advance on you.*|You weave back and forth in a dance-like motion|You must be closer to use tactical abilities on your opponent/)
+
+		#main travel loop, travels to rooms and then kicks off the sigil hunting process
+		walk_list.each do |roomid|
+			
+			#check if the room title is in the 'banpatterns' array before travelling there. Skip and add to nogo if a match is found
+			banflag = false
+			@data.SigilInfo['banpatterns'].each do |pattern|
+				if Map[roomid].title.to_s.include? pattern
```
KatoakDR, line 44: 🧹 Stylistically I like parenthesis around method calls, thanks
```suggestion
				if Map[roomid].title.to_s.include?(pattern)
```

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		region_list = region_room_list
+		#subtract known hunting spots and any other rooms in the YAML's nogo list to avoid combat and rooms with known mapping issues
+		walk_list = generate_nogo_list(region_list)
+		
+		#debug info
+		DRC.message("Region detected: #{@region}")
+		DRC.message("Season detected: #{@season.capitalize}")
+
+		#perform any outdoorsmanship buffs as this is the skill that governs sigil hunting
+		DRC.message("Checking for \"outdoors\" waggle set...")
```
KatoakDR, line 32: 💡  As done on line 46, you can use the single quote here to not have to escape the double-quotes within double-quotes
```suggestion
		DRC.message("Checking for 'outdoors' waggle set...")
```

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+				if Map[roomid].title.to_s.include? pattern
+					add_to_nogo(roomid, @data)
+					DRC.message("Match found in 'banpatterns'. Adding #{roomid} to nogo.")
+					banflag = true
+				end
+			end
+			next if banflag
+
+			#store whether the attempt to travel to the target room was successful
+			if check_room_data(roomid) >= 2 #Checking for known sigils for this room. If 2 are known, skip room.
```
KatoakDR, line 53: ❓ What is the significance of 2 over any other number?

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+				pause 0.1
+				next
+			else
+				travel_result = DRCT.walk_to(roomid, false)
+			end
+
+			#check buffs in the event they're running low
+			DRCA.do_buffs(@settings, 'outdoors')
+			#debug info
+			DRC.message("Travel Success? #{travel_result} | Room Target: #{roomid}")
```
KatoakDR, line 64: ❓ Are these "debug info" messages throughout the script intended to always be printed to the user? Or should they be conditional based on a debug setting like `DRC.message("something here") if $debug_mode_sigil` as is done in scripts like `crossing-repair`?

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		Flags.add('sigil-danger-room', /.*begins to advance on you.*|You weave back and forth in a dance-like motion|You must be closer to use tactical abilities on your opponent/)
+
+		#main travel loop, travels to rooms and then kicks off the sigil hunting process
+		walk_list.each do |roomid|
+			
+			#check if the room title is in the 'banpatterns' array before travelling there. Skip and add to nogo if a match is found
+			banflag = false
+			@data.SigilInfo['banpatterns'].each do |pattern|
+				if Map[roomid].title.to_s.include? pattern
+					add_to_nogo(roomid, @data)
```
KatoakDR, line 45: 🧹 The method `add_to_nogo` is always called with the same second argument, `@data`. Thoughts on removing the second argument from the method calls and just reference it in the method itself?

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		return room_matches
+	end #check_room_data
+	
+	#attempts to navigate out of a room lacking important data
+	def unstuck_me(target_room)
+		#Check the exits to the room we think we're in in Mapdb. Store the exits in an array then loop through it.
+		exit_list = Map[target_room].wayto
+		exit_list.each do |rnum, dir|
+		  #flag to listen for room numbers appearing in 'peer' info
+		  Flags.add('sigil-room-detected', /Room Number: \d+/)
```
KatoakDR, line 144: 🚧 The output `Room Number: #` is provided by another script, `roomnumbers`. Not everyone runs that script. Could you instead grab the room number from `Room.current.id`?

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		end
+
+		#generate the array
+		roomlist = []
+		data.each do |room|
+		  roomlist.push(room.id)
+		end
+
+		#return the full list of rooms
+		return roomlist
```
KatoakDR, line 208: This loop is fine and how we have to do it most programming languages. I want to share another way it can be done in Ruby just for knowledge share.
```ruby
Map.list.select { |room| room.image == Room.current.image }.map(&:id)
```
**Output**
```
,e echo Map.list.select { |room| room.image == Room.current.image }.map(&:id)

--- Lich: exec1 active.

[exec1: [1746, 1747, 1748, 1749, 1750, 1751, 1752, 1753, 1754, 1755, 1756, 1757, 1758, 1759, 1760, 1761, 1762, 1763, 1764, 1765, 1766, 1767, 1768, 1769, 8439, 8440, 8441, 8442, 8443, 8444, 8445, 8446, 8447, 8448, 8449, 8450, 8451, 8452, 8453, 8454, 8455, 8456, 8457, 8458, 8459, 8460, 8461, 8462, 8463, 8464, 8465, 8466, 8467, 8468, 8469, 8470, 8471, 8472, 8473, 8474, 8475, 8476, 8477, 8478, 8479, 8480, 8481, 8482, 8483, 8484, 8485, 8486, 8487, 8488, 8489, 8490, 8491, 8492, 8493, 8494, 14105, 14106, 14107, 14108, 14109, 14110, 14111, 14112, 14113, 14114]]

--- Lich: exec1 has exited.
```

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+	end #region_room_list
+	
+	#writes a roomid to the nogo list which is subtracted from the walk_list when the script is run in the future
+	def add_to_nogo(banroom, file_data)
+		#initialize a new array if nogo is empty
+		if file_data.SigilInfo['nogo'] == nil
+		  file_data.SigilInfo['nogo'] = []
+		end
```
KatoakDR, line 216: 💡 A ruby-ish way to do this is the conditional assignment: `file_data.SigilInfo['nogo'] ||= []`

If the variable is nil then it's assigned the right-hand side value, else the variable's value is not re-assigned.

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+
+		#push the roomid to the array and remove any duplicates
+		file_data.SigilInfo['nogo'].push(banroom)
+		file_data.SigilInfo['nogo'].uniq
```
KatoakDR, line 220: The `uniq` method **returns** the uniq set but doesn't modify the variable it's called on. You'd want to use `uniq!` with the exclamation point to modify the list in-place like `file_data.SigilInfo['nogo'].push(banroom).uniq!` or conditionally add to the array only if the value doesn't exist `file_data.SigilInfo['nogo'].push(banroom) unless file_data.SigilInfo['nogo'].include?(banroom)`

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		#subtract array of potentially dangerous rooms from the room list that was passed in
+		room_list -= ban_list
+		DRC.message("Walk list has been pruned. New list size: #{room_list.length}")
+		pause 1
+		return room_list #pruned, safer roomlist is returned
+	end #generate_nogo_list
+
+	#remove any rooms in the nogo list from recorded sigil locations before the script exits
+	def prune_final_list
+		ban_list = get_data('hunting').hunting_zones.values.flatten.uniq
```
KatoakDR, line 244: 💡 There's a couple places that `get_data('hunting').hunting_zones.values.flatten.uniq` is being made; I suggest moving this to an instance variable and load it once in `initialize` method

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+
+		#loop through the sigils and place them in the correct spot in @data
+		found_sigils.each do |sigil|
+		  #if the array doesn't exist yet, initialize a new one
+		  if file_data.SigilInfo[@region][sigil][@season] == nil
+		    file_data.SigilInfo[@region][sigil][@season] = []
+		  end
+		  #push the current roomid into the hash/array under the correct region, sigil, and season keys
+		  file_data.SigilInfo[@region][sigil][@season].push(Room.current.id)
+		  file_data.SigilInfo[@region][sigil][@season].to_set
```
KatoakDR, line 270: 🤔 I don't think calling `to_set` transforms the variable it's called on in-place. You may need to do a variable re-assignment. What's your goal with this line?

KatoakDR (changes requested):
Thanks for working on this script; I can tell you've put a lot of work in to it. I left a few questions and suggestions for code changes I'd like to see before we merge. Thanks!

sigilrecorder.lic
```diff
@@ -0,0 +1,290 @@
...
+		exit_list = Map[target_room].wayto
+		exit_list.each do |rnum, dir|
+		  #flag to listen for room numbers appearing in 'peer' info
+		  Flags.add('sigil-room-detected', /Room Number: \d+/)
+		  DRC.bput("peer #{dir}", /You peer/, /You see/, /\[Usage:/)
+		  pause 0.5
+		  #if we detected a room nubmer, move in that direction and return
+		  if Flags['sigil-room-detected']
+		    DRC.retreat
+		    fput(way)
```
KatoakDR, line 150: ❓ What variable does `way` refer to? I don't readily see it defined in this method `unstuck_me`

sigilrecorder.lic
Nyctus: I am renaming this "prune_walk_list" instead to have it be more consistent with what it's doing (it is also compiling the nogo list but it then returns the modified walk list sans anything in nogo).

sigilrecorder.lic
Nyctus: I only capitalize it here to make it look nicer to the user. Everywhere else I work with the lower case. The captured data and strings held in @season_list are all lower case.

Nyctus: I am going to go ahead and put a .downcase on the return $1 in get_season just in case Simu decides to change capitalization of the season names someday.

sigilrecorder.lic
Nyctus: Done!

sigilrecorder.lic
Nyctus: Done!

sigilrecorder.lic
Nyctus: Each room that has sigils has exactly 2 without exception. check_room_data returns the number of instances of a roomid for that season and roomid. If it is 2 or greater we know that all the sigils in that room are already known and it does not need to be mapped. (If it's greater than 2 we have too many sigils stored for that room so I could look into putting in some error messaging in the event that happens?)

sigilrecorder.lic
Nyctus: I tend to like somewhat verbose scripts but I can look into putting them behind a debug flag. Ruby makes that super easy thankfully ;)

sigilrecorder.lic
Nyctus: I did this to avoid modifying instance variables outside the initialize function. The function takes in @data, modifies it, then returns it, so that @data is only modified in initialize.

This does mean that in record_sigil the YAML file is more up-to-date than the data in @data but the out of date portions are the portions the script has already gone over. But I'd be worried about changing the size of arrays/hashes in @data while other parts of the script are looping through indices/keys.

sigilrecorder.lic
Nyctus: Ok debug flag implemented and tested. I was very aggressive about putting 'if debug' proceeding things so everything but the error when the script cannot find an exit and when the script detects an issue with the seasons is now squelched unless that flag is used.

sigilrecorder.lic
Nyctus: Sorcery! Hope you didn't do that in town.

sigilrecorder.lic
Nyctus: Hm ok I am going to retool this a bit, then. I am going to ditch the flag and have it random-move and then check for Room.current.id != nil. I'll maybe cap it at like 3-5 attempts so I don't inadvertently get people into significant trouble.

This is primarily to get the user out of dead-end rooms that have mapping issues which are usually actual dead-ends but occasionally they're part of larger unmapped buildings/features, or areas with day/night issues. I thought about spamming mapping requests with *every* room I found with day-night issues but there's already so many sigil rooms there's no actual reason to bother other people so these rooms can be included. Even with multiple people sigil hunting there should be plenty of rooms to go around for all but the rarest sigils types in a region.

Nyctus: I completely redid this function. I now store the 'start room' where the user began the script as a sort of anchor.

If the script ends a movement in a room without a roomid but did not "fail" to move to the intended room (as far as go2 is concerned) it then trusts that go2 got it to the desired room (since go2 threw no errors) but that the roomid is not visible for some reason. After adding the bad room to nogo, using this assumed roomid, I generate a path_to between the assumed roomid and the start room. Then, using wayto lookups and assumed room ids I slowly backtrace my steps toward the start room. At each stop along the way I check if the current room has a visible roomid, if so, I abort the loop, and the script continues to the next room in walk_list.

If the backtrace fails, the script will exit. Though I tested this in Under Wharfs in Riverhaven and it was able to avoid getting stuck down there so it's fairly robust (definitely not perfect).

sigilrecorder.lic
Nyctus: Added the sorcery to my script. I'll need to do some Googling to truly understand it 🤣

sigilrecorder.lic
Nyctus: Noice. Implemented.

Nyctus: Added!

sigilrecorder.lic
Nyctus: Updated all pushes to use the unless method. If I run into issues with gaps in the arrays I can just do a single cleanup at the end or just fix the bugs causing gaps.

sigilrecorder.lic
Nyctus: Moved hunting data to an instance variable

sigilrecorder.lic
Nyctus: I was trying another way of doing .uniq

I am moving this to 'unless' and 'include?' method as well.

sigilrecorder.lic
Nyctus: This was removed and a new method of backing out of rooms that go2 doesn't recognize was implemented.

KatoakDR (commented):
Looks like the file reverted to tabs again. Would you mind replacing tabs with two spaces please? It's a style consistency to match the other scripts. Thanks

sigilrecorder.lic
```diff
@@ -0,0 +1,297 @@
...
+	def initialize
+    arg_definitions = [
+      [
+        { name: 'debug', regex: /debug/i, optional: true, description: 'Show debug messaging' }
+      ]
+    ]
+
+    args = parse_args(arg_definitions)
+
+    args.debug == 'debug' ? @debug = true : @debug = false
```
KatoakDR, line 19: 🧹 I believe this can be simplified to this assignment.
```suggestion
    @debug = args.debug
```
If `parse_args` sees `debug` as an argument then `args.debug` will have the value "debug", which in Ruby is truthy because it's neither nil nor false.

If `parse_args` does not see `debug` as an argument then `args.debug` will be nil, which is falsey.

sigilrecorder.lic
KatoakDR: Did this get renamed?

sigilrecorder.lic
```diff
@@ -0,0 +1,297 @@
...
+
+		#main travel loop, travels to rooms and then kicks off the sigil hunting process
+		walk_list.each do |roomid|
+			
+			#check if the room title is in the 'banpatterns' array before travelling there. Skip and add to nogo if a match is found
+			banflag = false
+			@data.SigilInfo['banpatterns'].each do |pattern|
+				if Map[roomid].title.to_s.include?(pattern)
+					prune_walk_list(roomid, @data)
+					DRC.message("Match found in 'banpatterns'. Adding #{roomid} to nogo.") if @debug
```
KatoakDR, line 58: 💡 Based on the wording of the debug message, perhaps these two lines should be swapped?

sigilrecorder.lic
```diff
@@ -0,0 +1,297 @@
...
+			case line = DRC.bput('perceive sigil', @sigil_list, 'Roundtime', /^You recall.*/, /^Having recently been searched/, /You are too distracted/, /Something about|This is not an appropriate place for that/)
+			when /^You recall.*/
+			  DRC.message('Room searched too recently') if @debug
+			  return nil
+			when /You are too distracted/
+			  prune_walk_list(Room.current.id, @data)
+			  return nil
+			when /Something about|This is not an appropriate place for that/
+			  prune_walk_list(Room.current.id, @data)
+			  return nil
```
KatoakDR, line 195: These do the same actions, should line 193 combine the match string from 190?

sigilrecorder.lic
```diff
@@ -0,0 +1,297 @@
...
+		#return the array of all sigils discovered in the room (should be 2 unless something went wrong)
+		return room_sigils
+	end
+	
+	#generate the list of all rooms that contain the current room's Room.current.image data
+	def region_room_list
+		data = Map.list.select do |room|
+		  room.image == Room.current.image
+		end
+
```
KatoakDR, line 213: This is now redundant of line 215 and the `data` variable isn't used; these lines can be removed
```suggestion
```

sigilrecorder.lic
```diff
@@ -0,0 +1,297 @@
...
+		room_list -= ban_list
+		DRC.message("Walk list has been pruned. New list size: #{room_list.length}") if @debug
+		pause 1
+		return room_list #pruned, safer roomlist is returned
+	end #generate_nogo_list
+
+	#remove any rooms in the nogo list from recorded sigil locations before the script exits
+	def prune_final_list
+		ban_list = @hunting_zones
+		ban_list += @data.SigilInfo['nogo']
```
KatoakDR, line 252: Line 240 conditionally adds this only if it's not nil. Should this do the same thing?

Also, another low-code way to remove nil values from an array is to use  `compact` method:
```ruby
some_list = [ listA, listB, listC ].flatten.compact
```
Using `compact`, any nil elements are removed

sigilrecorder.lic
Nyctus: That melts my brain a bit but it also sorta makes sense haha. I'll change it over and test it out!

sigilrecorder.lic
Nyctus: Updated. Also swapped the order of the sentences from a readability standpoint (i.e. put the "What I am doing"  before the "Why I am doing it.")

sigilrecorder.lic
Nyctus: Updated and combined branches in the case statement. I also did this for the two possible errors that can be received when a room has been searched to recently.

sigilrecorder.lic
Nyctus: Deleted redundant code

sigilrecorder.lic
Nyctus: Collapsed to a single line: `ban_list = [@hunting_zones, @data.SigilInfo['nogo']].flatten.compact`

rpherbig (approved):

PR #5565, opened 2022-02-11
[script][tarantula.lic] - Re-adding bolding of tarantula use messaging

BinuDR:
I added a method to bold text into common.lic and re-added it into the atmo command in tarantula.lic so that the tarantula use message is bolded again.


KatoakDR (commented):
Code looks good to me, thanks

KatoakDR:
@rpherbig @BinuDR Other updates I've just submitted for `common.lic` in https://github.com/rpherbig/dr-scripts/pull/5574 I believe supercede the need for this PR. `DRC.atmo(text)` will by default be bold. To make it not bold then use `DRC.atmo(text, false)`.

Thanks

PR #5572, opened 2022-02-12
[script] [common-moonmage] Bring over more defs from common-arcana we need for other spells

MahtraDR:
We tested so much... with IOTS. Not with COL.

Broke COL casting without these...

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -351,5 +351,4 @@ module DRCMM
     pause 0.5 while UserVars.moons.empty?
   end
 
-
 end


common-moonmage.lic
```diff
@@ -320,4 +320,36 @@ module DRCMM
     DRC.message("Could not set planet data. Cannot cast #{data['abbrev']}")
   end
 
+  def set_moon_data(data, settings = nil)
```
KatoakDR, line 323: Are moon mages the only guild that needs this info? Do traders have spells that rely on specific moons?

common-moonmage.lic
```diff
@@ -320,4 +320,36 @@ module DRCMM
...
+    return if Script.running?('moonwatch')
+
+    echo 'moonwatch is not running. Starting it now'
+    UserVars.moons = {}
+    custom_require.call('moonwatch')
+    echo "Run `#{$clean_lich_char}e autostart('moonwatch')` to avoid this in the future"
+    pause 0.5 while UserVars.moons.empty?
+  end
+
+
```
KatoakDR, line 354: 🧹 delete extra newline
```suggestion
```

common-moonmage.lic
rpherbig: @KatoakDR should this comment hold up merging?

KatoakDR: Nope, was just curious!

MahtraDR: Can we please commit this now, because some spells are broken for moonies. And I'll clean up after and check the trader question? 

I need to make a trader.

MahtraDR: Katoak and I chatted on discord. Resolving.

common-moonmage.lic
```diff
@@ -320,4 +320,35 @@ module DRCMM
     DRC.message("Could not set planet data. Cannot cast #{data['abbrev']}")
   end
 
+  def set_moon_data(data, settings = nil)
```
rpherbig, line 323: No sooner did I click merge than I realized something. Why does this function take `settings` as an arg? It doesn't seem to use it.

PR #5573, opened 2022-02-13
Remove forcing wyvern mountain from undergondola arg

Hiinky:
There is a known DC/freeze bug that randomly effects folks trying to train athletics on wyvern mountain that often results in character death.  Removing the default behavior to force anyone over 850 athletics (even though its best training) allows the user to avoid unnecessary deaths while a visible warning alerts the user that their current method is less effective.

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -337,7 +337,7 @@ class Athletics
     if UserVars.athletics < 540
       override_location_and_practice('undergondola_branch')
     else
-      DRC.message("Warning: ;athletics undergondola with more than 850 athletics is not best useand you may consider ;athletics wyverns instead.") if UserVars.athletics > 850
+      DRC.message("Warning: ;athletics undergondola with more than 850 athletics is not best use and you may consider ;athletics wyverns instead.") if UserVars.athletics > 850
       walk_to(9607)
       walk_to(9515)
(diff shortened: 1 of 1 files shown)


athletics.lic
```diff
@@ -336,7 +336,8 @@ class Athletics
   def climb_branch
     if UserVars.athletics < 540
       override_location_and_practice('undergondola_branch')
-    elsif UserVars.athletics < 850
+    else
+      DRC.message("Warning: ;athletics undergondola with more than 850 athletics is not best use and you may consider ;athletics wyverns instead.") if UserVars.athletics > 850
```
KatoakDR, line 340: Love the anti-death change ☠️ 🐉 

💡 For Genie users where semicolons are not shown, suggest using `#{$lich_char}`. Not a blocker for merge, though.
```suggestion
      DRC.message("Warning: '#{$lich_char}athletics undergondola' with more than 850 athletics is not best use and you may consider '#{$lich_char}athletics wyverns' instead.") if UserVars.athletics > 850
```

rpherbig: Good catch, but since this involved character deaths I'm going to merge it immediately. I would really like to see the cleanup PR, though.

PR #5578, opened 2022-02-13
[data] [base-hunting] Removing unnecessary reference

MahtraDR:
This isn't a useful reference in this file.


rpherbig:
Wtf? That's been in there for 4 years? Sigh...

PR #5588, opened 2022-02-14
[script] [tdps] Removing explicit exp all to populate drinfomon

MahtraDR:
No longer needed since the rewrite. Happy to wait a day or three for it, but it's also safe to merge now.

### DIFF
--- a/tdps.lic
+++ b/tdps.lic
@@ -37,9 +37,6 @@ stats = [DRStats.strength, DRStats.reflex, DRStats.agility, DRStats.charisma, DR
 race = DRStats.race.downcase.sub(' ', '').sub("'", '')
 mods = races.fetch(race)
 
-put 'exp all'
-waitfor('EXP HELP for more information')
-
 # Capture and extract data
 data = reget(39)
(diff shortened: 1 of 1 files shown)


KatoakDR:
> No longer needed since the rewrite. Happy to wait a day or three for it, but it's also safe to merge now.

Agree, let's give this another day or so to ensure people have logged out/in to get on the latest drinfomon, thanks!

KatoakDR (approved):

tdps.lic
```diff
@@ -37,9 +37,6 @@ stats = [DRStats.strength, DRStats.reflex, DRStats.agility, DRStats.charisma, DR
 race = DRStats.race.downcase.sub(' ', '').sub("'", '')
 mods = races.fetch(race)
 
-put 'exp all'
-waitfor('EXP HELP for more information')
-
 # Capture and extract data
 data = reget(39)
 data.each { |line| exp << line if (/\(.?.\/34\)/ =~ line) }
```
KatoakDR, line 42: I think line 40 is actually needed because line 44 parses the output, are you sure this needs to be deleted?

PR #5590, opened 2022-02-15
[script] [drinfomon] don't squelch on startup, wait til needed

KatoakDR:
### Background
* Reported by [Xelten in Lich Discord](https://discord.com/channels/745675889622384681/745675890242879671/942934618661871717) that _"if you type exp mods, it doesn't show any response at all, but then if you type it a second time it shows the skills"_
* Upon investigation, the variables that control whether output should be squelched default to `!UserVars.drinfomon_debug` so if you didn't have the debug on then by default the downstreams hooks would squelch the first time info/exp/etc were parsed regardless who initiated the command (player or script)

```
>exp mods

>bounce
You bounce around happily.

>exp mods
The following skills are currently under the influence of a modifier:
  None
```

### Changes
* Don't default the squelch variables to true; instead wait til a script calls the appropriate function that requests for background parsing of info/exp/etc, at which point the squelch variable will be set accordingly

### Tests
* @MahtraDR @asechrest confirmed in Wrayth and Profanity
* I confirmed in Genie

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#drinfomon
 =end
 
-$DRINFOMON_VERSION = '2.0.13'
+$DRINFOMON_VERSION = '2.0.14'
 
 no_kill_all
 no_pause_all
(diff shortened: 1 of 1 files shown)


MahtraDR (approved):
Tested locally, and works as expected.

asechrest (approved):
Looks good. Tested locally. Fixes improper squelch of the first manual `exp mods`.

PR #5594, opened 2022-02-15
[script] [safe-room] [data] [base-town] Chyolvea addition for healing and banking

XTyrsin:
Another option for healing and banking in the Ilithi area.
Added line for completion of healing by npc healer Elys
Added town information for banking and npc healing in Chyolvea Tayeu'a

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -60,6 +60,7 @@ class SafeRoom
         'Come along.  Leave room for the others.',
         "I think you don't really need healing",
         "Now move along, I can't do anything more for you now",
+        "You're in good enough shape, #{Regexp.quote(checkname)}",
         'Gaktrhn bends down as if to touch you',
         'After a quick diagnosis')
       Flags.add('moved', 'grabs you and drags you', 'grabs your arm and drags you .* with')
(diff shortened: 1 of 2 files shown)


safe-room.lic
```diff
@@ -60,6 +60,7 @@ class SafeRoom
         'Come along.  Leave room for the others.',
         "I think you don't really need healing",
         "Now move along, I can't do anything more for you now",
+        "You're in good enough shape, #{Regexp.quote(checkname)}",
```
MahtraDR, line 63: You don't need to Regexp.quote here, you can match directly on #{checkname}

The output of the two is the same.

```
>;e echo DRC.message("#{checkname}")
Mahtra
```
```
 >;e echo DRC.message("#{Regexp.quote(checkname)}")
 Mahtra
 ```

MahtraDR: Oh, I see you're just sticking with the existing convention already in the file.

Looks good.

Thanks for the contribution!

MahtraDR (approved):
Tested, works as expected.

MahtraDR:
Checked room numbers, and conversion rate. All looks good.

Issue #5593, opened 2022-02-15
Combat trainer + incense issue when flint is unavailable or missing


Kaesken:
Turns out my flint was missing but not sure if you guys wanted this info or not.

```
[combat-trainer]>glance

A storm bull snorts, its breath forming a misty cloud in the air.

> 
[combat-trainer]>get my icesteel cutlass 

You glance down at your empty hands.

> 
You get an icesteel cutlass with a lapis lazuli hilt from inside your leather pack.

> 
[combat-trainer]>get my incense from my haversack

You get some burnt incense from inside your haversack.

> 
[combat-trainer]>light my incense with my flint

What were you referring to?

> 
Your body below the waist is continuing to drip water everywhere.

> 
A musical note swells in the air as a storm bull brushes its horn against its leg.

> 
The storm bull closes to pole weapon range on you!

> 
You feel fully attuned to the mana streams again.

> 
You feel fully prepared to cast your spell.

> 
A storm bull drums its silvery hooves upon the ground.

> 
[combat-trainer]>light my incense with my flint

What were you referring to?

> 
The storm bull closes to melee range on you!

> 
A storm bull paws at the ground with a silvery hoof.

A storm bull drums its silvery hooves upon the ground.

> 
[combat-trainer]>light my incense with my flint

What were you referring to?

> 
A storm bull drums its silvery hooves upon the ground.

> 
[combat-trainer]>light my incense with my flint

What were you referring to?

> 
A storm bull tosses its head, pearly horns gleaming in the moonlight.

> 
Misty tendrils rise from the pale hide of a storm bull as it gazes alertly around the area.

> 
A storm bull thunders into the area with a grunting snort, razor-sharp hooves digging into the ground.

A storm bull stomps a hind hoof against the ground before returning to the serious business of eating.

The storm bull begins to advance on you!
The storm bull is still a distance away from you and is closing steadily.

> 
A storm bull drums its silvery hooves upon the ground.

> 
[combat-trainer]>light my incense with my flint

What were you referring to?

> 
The storm bull closes to pole weapon range on you!

> 
A musical note swells in the air as a storm bull brushes its horn against its leg.

The storm bull closes to melee range on you!

> 
You feel fully attuned to the mana streams again.

> 
A storm bull drums its silvery hooves upon the ground.

> 
Can't light your incense for some reason. Tried 5 times, giving up.

--- Lich: error: wrong number of arguments (given 0, expected 1)
[...]

PR #5595, opened 2022-02-15
[script] [common-theurgy] Adding theurgy_supply_container to empty_cleric_hands call

MahtraDR:
It should be there to make the function work.

Resolves https://github.com/rpherbig/dr-scripts/issues/5593

### DIFF
--- a/common-theurgy.lic
+++ b/common-theurgy.lic
@@ -214,7 +214,7 @@ module DRCTH
       lighting_attempts += 1
       if (lighting_attempts >= 5)
         DRC.message("Can't light your incense for some reason. Tried 5 times, giving up.")
-        empty_cleric_hands
+        empty_cleric_hands(theurgy_supply_container)
         return false
       end
     end


asechrest:
> > I don't have a cleric so sort of flying blind, but in reference to #5593, why didn't the script recognize there was no `flint_lighter` and `return false`. Instead it ended up trying to light incense without a lighter 5 separate times. It looks to me like line 200 is designed to recognize when we don't have a flint lighter and fail gracefully.
> 
> flint_lighter is a metal weapon. It strikes the flint, which itself never leaves your theurgy_container(which can't be a portal)

Ah! I see, now.

Thoughts on adding the flint check in this PR? Seems of a similar vein.

MahtraDR:
> > > I don't have a cleric so sort of flying blind, but in reference to #5593, why didn't the script recognize there was no `flint_lighter` and `return false`. Instead it ended up trying to light incense without a lighter 5 separate times. It looks to me like line 200 is designed to recognize when we don't have a flint lighter and fail gracefully.
> > 
> > 
> > flint_lighter is a metal weapon. It strikes the flint, which itself never leaves your theurgy_container(which can't be a portal)
> 
> Ah! I see, now.
> 
> Thoughts on adding the flint check in this PR? Seems of a similar vein.

I rather just merge this as quick fix, since it's needed no matter what.

I will make a simple have_flint? predicate which I can add as a check instead of tacking that onto this PR.

KatoakDR (approved):

PR #5598, opened 2022-02-16
common-crafting - Logic fix in find_empty_crucible

BinuDR:
Current behavior:

If you're already in a room with a crucible when this method is called, it only tries to tap a crucible to verify that one is there.  It does not also check to make sure it's empty.


KatoakDR (approved):
Thanks for the fix 👍

@MahtraDR @asechrest do either of y'all use crafting to help test?

PR #5606, opened 2022-02-20
Fix common healing to work when diseased but no wounds

PatricCampbell:
I found that common healing does not handle the case when you are diseased but have no wounds and the infection is not dormant.

It looks like this in game:
`Your body feels at full strength.
 
Your wounds are infected.`

### DIFF
--- a/common-healing.lic
+++ b/common-healing.lic
@@ -24,7 +24,7 @@ module DRCH
     diseased = false
     poisoned = false
 
-    DRC.bput('health', 'You have', 'You feel somewhat tired and seem to be having trouble breathing')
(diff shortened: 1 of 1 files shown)


MahtraDR:
Hi mate. Thanks for the submission! 

Could you please post a log with before and after this change? 

It's simple enough in code, I'd just like to see the difference in game. 

Thanks!

PatricCampbell:
Before the change it tries to use the health command then times out and dumps the log since there's no match. After the change it just moves to the next section so there's nothing to show in game.

Sorry should have been more clear in the PR.

KatoakDR (approved):
Thanks

PR #5607, opened 2022-02-20
[script] [drinfomon] [exp-monitor] Track learning rate increases only

KatoakDR:
### Background
* The `exp-monitor` script is for echoing when you increased your learning rate in a skill so you can easily correlate an action to ROI
* The `drinfomon` script, which monitors for exp pulses, pushes data to `DRSkill.gained_skills` that `exp-monitor` relies on.
* However, the `drinfomon` script was not distinguishing between an exp pulse for new experience vs drained experience, which made it difficult to understand what the output even meant.

### Changes
* Update `drinfomon` to only push data to `DRSkill.gained_skills` if the learning rate indeed increased, and also pass along what that delta change was.
* Update `exp-monitor` to show the experience learning rate change along with the skill (which also mirrors how Genie's EXPTracker plugin does it too)

### Test
```
> collect rock
Roundtime: 15 sec.

Gained: Outdoorsmanship(+2), Perception(+1)
```

I'm a Genie user, so I'm testing the change by ensuring the output from EXPTracker plugin (though I may be on a customized variant of the plugin) matches `exp-monitor` script. So far, it's spot on.

image

image

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#drinfomon
 =end
 
-$DRINFOMON_VERSION = '2.0.15'
+$DRINFOMON_VERSION = '2.0.16'
 
 no_kill_all
 no_pause_all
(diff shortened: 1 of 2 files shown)


KatoakDR:
in draft while recruit testers

MahtraDR (approved):
Soaked, and works well. I suggest we merge this, and deal with any new problems as new. 

This one is amazing. The change to exp-monitor is extremely satisfying.

PR #5608, opened 2022-02-21
[script] [drinfomon] mitigate thread-unsafe `Regexp.last_match` in procs

KatoakDR:
### Dependencies
* https://github.com/rpherbig/dr-scripts/pull/5607

### Background
* In rare scenarios, a downstream hook in `drinfomon` will fail due to failing to parse a named group from Regexp.last_match. For example, `--- Lich: DownstreamHook: undefined group name reference: rate`
* After much testing and investigation, we have reason to believe it is due to Regexp.last_match being **thread-unsafe** when used in a proc, which is what downstream hooks are.
  - https://github.com/psu-libraries/psulib_traject/issues/132
  - https://github.com/traject/traject/issues/225
  - https://stackoverflow.com/questions/9004245/is-regexp-last-match-thread-safe#:~:text=last_match%20is%20equivalent%20to%20reading,last_match%20is%20thread%2Dsafe. says it's thread-local and method-safe, but that doesn't necessarily mean that it's proc-safe

### Changes
* In the downstream hooks, refactor from using `Regexp.last_match` and instead get a `MatchData` variable locally and reference it, thus avoiding race conditions that could cause Regexp.last_match to unexpectedly change and cause script error

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#drinfomon
 =end
 
-$DRINFOMON_VERSION = '2.0.15'
+$DRINFOMON_VERSION = '2.0.18'
 
 no_kill_all
 no_pause_all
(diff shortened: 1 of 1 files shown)


MahtraDR (approved):
Soaked, and works well. I suggest we merge this, and deal with any new problems as new.

PR #5613, opened 2022-02-22
[script] [bescort] add logic to help swim to cave trolls

KatoakDR:
### Dependencies
* Depending on your athletics and whether you have a flying mount, this change is best supported with https://github.com/dragon-realms/dr-lich/pull/69

### Background
* In Dirge, there is one area of [cave trolls](https://elanthipedia.play.net/Cave_troll) that are only accessible by swimming through a fast current, neck deep, stream.
* This PR adds support for swimming the stream as we do for swimming rivers.
* With 505 athletics, it's difficult to swim against the current and requires a lot of retries.
* With 585 athletics, it's navigable and retries are few and far between.
* Note, this area is pitch black; you need a light source.

### Changes
* Add new option `;bescort cave_trolls [enter|exit]` to swim you into, or out of, the cave troll area
* For performance, load settings and equipment manager once per script run

The red-circled section at bottom are the cave trolls accessible by swimming the stream.
image

### Notes
To take advantage of this update, I need to request @Hiinky to help make two room updates in the map:

```diff
  {
    "id": 15759,
    "title": [
      "[[Wicked Burrow Mine, The Depths]]"
    ],
    "description": [
      "A cacophonous rumbling echos all around from the rapid movement of water somewhere above.  The tunnel narrows deeper into the mine."
    ],
    "paths": [
      "Obvious exits: east."
    ],
    "wayto": {
      "15760": "east",
      "15758": "go bank",
+      "19089": ";e start_script('bescort', ['cave_trolls', 'exit']); wait_while{running?('bescort')};"
    },
    "timeto": {
      "15760": 0.2,
      "15758": 0.2,
+      "19089": ";e if UserVars.athletics > 539 && Script.exists?('bescort') then 0.2 else 20.0 end"
    },
    "image": "Zoluren, Wicked Burrow Mine-1.png",
    "image_coords": [
      509,
      468,
      526,
      487
    ]
  },

  ...

  {
    "id": 19089,
    "title": [
      "[[Wicked Burrow Mine, Stream Bank]]"
    ],
    "description": [
      "Water drips and pours from every inch of stone, all headed down the bank and into the fast-moving stream below.  A small bit of light shines in from the west, barely illuminating the barren walls."
    ],
    "paths": [
      "Obvious exits: west."
    ],
    "wayto": {
      "19088": "west",
      "19198": "go bank",
      "12036": ";e multifput('go passage','go hole')",
      "12156": "northwest",
+      "15759": ";e start_script('bescort', ['cave_trolls', 'enter']); wait_while{running?('bescort')};"
    },
    "timeto": {
      "19088": 0.2,
      "19198": 0.2,
      "12036": 0.2,
      "12156": 0.2,
+      "15759": ";e if UserVars.athletics > 539 && Script.exists?('bescort') then 0.2 else 20.0 end"
    },
    "image": "Zoluren, Wicked Burrow Mine-1.png",
    "image_coords": [
      304,
      392,
      321,
      411
    ]
  },
```

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -226,8 +226,12 @@ class Bescort
       ],
       [
         { name: 'galley', regex: /galley/i, description: "Take the M'riss-Mer'Kresh galley." },
-        { name: 'mode', options: %w[mriss merkresh], description: "mriss or merkresh?  Can also be started on the galley."  }
-      ]  
+        { name: 'mode', options: %w[mriss merkresh], description: "mriss or merkresh?  Can also be started on the galley." }
+      ],
+      [
+        { name: 'cave_trolls', regex: /cave_trolls/i, description: "Enter the cave trolls hunting area beyond the stream." },
+        { name: 'mode', options: %w[enter exit], description: "Enter or exit?" }
+      ],
     ]
 
     args = parse_args(arg_definitions)
@@ -305,6 +309,8 @@ class Bescort
       currach(args.mode)
     elsif args.galley
       take_m_m_galley(args.mode)
+    elsif args.cave_trolls
+      cave_trolls(args.mode)
     end
   end
 
@@ -350,6 +356,32 @@ class Bescort
         exit
       end
     end
+  end
+
+  def cave_trolls(mode)
+    flying_mount = @settings.flying_mount
+    @equipment_manager.empty_hands
+    have_changed_gear = flying_mount ? false : @equipment_manager.wear_equipment_set?('swimming')
+    case mode.downcase
+    when /enter/
+      start_room = 19089
+      moveset = %w[southeast southeast southeast]
+    when /exit/
+      start_room = 15759
+      moveset = %w[northwest northwest northwest]
+    end
+    manual_go2(start_room)
+    move('go stream bank')
+    if flying_mount
+      use_flying_mount(flying_mount, 'mount')
+      move(moveset.first)
+    else
+      moveset.each { |dir| swim(dir) }
+    end
+    use_flying_mount(flying_mount, 'dismount') if flying_mount
+    move('go stream bank')
+    @equipment_manager.empty_hands
+    @equipment_manager.wear_equipment_set?('standard') if have_changed_gear
   end
 
   def take_m_m_galley(mode)
--- a/bescort.lic
+++ b/bescort.lic
@@ -359,9 +359,6 @@ class Bescort
   end
 
   def cave_trolls(mode)
-    flying_mount = @settings.flying_mount
-    @equipment_manager.empty_hands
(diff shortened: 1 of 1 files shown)


MahtraDR (commented):
Looks good to me. I'll run my dude there tomorrow and test it.

Hiinky:
This is a locking change and crazy low priority due to the fact that the critters in here are 170-250ish and the athletics check is significantly higher than that.  Until Simu fixes that, I don't see that this is worth a lock.

KatoakDR:
> This is a locking change and crazy low priority due to the fact that the critters in here are 170-250ish and the athletics check is significantly higher than that. Until Simu fixes that, I don't see that this is worth a lock.

Gotcha. Yeah, I was surprised the athletics barrier is so much higher than the critters inside.

If by chance you're keeping a queue of updates to roll in when the map will be locked, I appreciate if this can be tucked in there. In the meantime, I can adjust my map locally to support it. Thanks!

Hiinky:
I think it does make sense to go ahead and merge this change. that way if people really want in there with something like T2, they can still manage it or with a multi.  ;multi 1,;go start room,;bescort things,;go room on the other side.  Pretty sure the rooms on the other end work so a custom handler to get past the wonky room should be easy enough until the next locking list comes up.

bescort.lic
```diff
@@ -349,12 +358,38 @@ class Bescort
...
+      moveset = %w[southeast southeast southeast]
+    when /exit/
+      start_room = 15759
+      moveset = %w[northwest northwest northwest]
+    end
+    manual_go2(start_room)
+    move('go stream bank')
+    if flying_mount
+      use_flying_mount(flying_mount, 'mount')
+      move(moveset.first)
```
KatoakDR, line 377:
```suggestion
      move(moveset.first) # mount travels all three rooms at once
```

MahtraDR:
Ran this today. The darkness on the troll side is annoying, and causes a 30 second wait before bescort puts your armor back on. I'd prefer not waiting, since the go bank has no skill check, it's a .2 second movement.

We also need a warning the area is dark. I wasn't prepared for that. It works well otherwise. :D

rpherbig (approved):

KatoakDR:
> Ran this today. The darkness on the troll side is annoying, and causes a 30 second wait before bescort puts your armor back on. I'd prefer not waiting, since the go bank has no skill check, it's a .2 second movement.
>
> We also need a warning the area is dark. I wasn't prepared for that. It works well otherwise. :D

This PR now removes/wears gear while you're in the stream, which is a safe room and lit.

To handle the fact that you're in a dark room if you don't have a light source is handled via https://github.com/dragon-realms/dr-lich/pull/70

PR #5614, opened 2022-02-23
[script] [drinfomon] tighten up regex patterns; fix logic bug in IF condition

KatoakDR:
### Background
* When `UserVars.track_exp = true` then the script appends the amount of ranks gained per skill this session to the `server_string`.
* This modified string, due to a logic bug with the IF condition, is evaluated and parsed as if it was output from `exp all` (which it wasn't) and calls `DRSkill.update(...)` with an incorrect mindstate value.
* This incorrect mindstate  value causes `exp-monitor` to think that your prior mindstate was 0 and whatever your current mindstate is is the delta and so each pulse prints "Skill Name(+34)" or some erroneously high number.

```
 Gained: Arcana(+34), Performance(+34)
```

### Changes
* First, tighten up the regex patterns to match skill names as letters and spaces, don't mistakenly pick up numbers (which would come from the injected rank gain when `UserVars.track_exp = true`
* Second, fix the flawed IF condition to check if `parsing_exp_output` is the value true, not assign it to true then check

### Tests
@MahtraDR and I collaborated in Discord and tested together.

### DIFF
--- a/drinfomon.lic
+++ b/drinfomon.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#drinfomon
 =end
 
-$DRINFOMON_VERSION = '2.0.18'
+$DRINFOMON_VERSION = '2.0.19'
 
 no_kill_all
 no_pause_all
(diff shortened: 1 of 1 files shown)


MahtraDR (approved):

MahtraDR (approved):
Tested locally, fixed the issue we found.

asechrest:
Nice work on this.

PR #5615, opened 2022-02-23
[script][common-arcana] Made prep_time explicit, and added it as an option for ritual spells

MahtraDR:
Prior to this, two things:

1. ritual spells didn't honor `prep_time`
2. `prep_time` was only respected if it was shorter than `waitcastrt?`

With this PR, ritual spells can have `prep_time`, and it is always explicitly honored.

I also normalized the variable call names between the two methods.

I tested this locally using
```
  bloodthorns: &bloodthorns
    Bloodthorns:
      prep_time: 40
      mana: 500
      recast: 5
      focus: emerald.charm
      worn_focus: true
      before:
      - message: spell stance 130 100 70
      after:
      - message: spell stance 130 100 70
```
for rituals, and
```
  rits: &rits
    River in the Sky:
      mana: 23
      prep_time: 40
      cambrinth: [23]
      recast: 2
      cast: cast low
      before:
      - message: spell stance 100 130 70
      after:
      - message: spell stance 100 100 100
``` 
for non-rituals
and setting the `prep_time` tunable to various values.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -612,13 +612,8 @@ module DRCA
     cast_lifecycle_lambda.call('post-prep', data, settings) if cast_lifecycle_lambda != nil
 
     if data['prep_time']
-      echo("data['prep_time'] is #{data['prep_time']}")
-      echo("Time.now is #{Time.now}")
-      echo("prepare_time is #{prepare_time}")
-      echo("Time.now - prepare_time is #{Time.now - prepare_time}")
       pause until Time.now - prepare_time >= data['prep_time']
     else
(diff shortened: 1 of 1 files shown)


BinuDR:
I tested it out and it's working fine for me.

KatoakDR (commented):
Code looks good to me.  Will test later today, thanks

asechrest (commented):
Tested prep_time and it works fine. I don't currently have a ritual spell so will ask @KatoakDR or @BinuDR to verify that side of it.

image

PR #5617, opened 2022-02-24
[scripts] [common-healing] Adding match for oozing sores

MahtraDR:
Similar vein to https://github.com/rpherbig/dr-scripts/pull/5606, we already have a `when` for this pattern, but it isn't in the bput expected output.

### DIFF
--- a/common-healing.lic
+++ b/common-healing.lic
@@ -24,7 +24,7 @@ module DRCH
     diseased = false
     poisoned = false
 
-    DRC.bput('health', 'You have', 'You feel somewhat tired and seem to be having trouble breathing', 'Your wounds')
+    DRC.bput('health', 'You have', 'You feel somewhat tired and seem to be having trouble breathing', 'Your wounds', "Your body")
     pause 0.5
     health_lines = reget(50).map(&:strip).reverse
 
--- a/common-healing.lic
+++ b/common-healing.lic
@@ -24,7 +24,7 @@ module DRCH
     diseased = false
     poisoned = false
 
-    DRC.bput('health', 'You have', 'You feel somewhat tired and seem to be having trouble breathing', 'Your wounds', "Your body")
+    DRC.bput('health', 'You have', 'You feel somewhat tired and seem to be having trouble breathing', 'Your wounds', 'Your body')
     pause 0.5
     health_lines = reget(50).map(&:strip).reverse
 


common-healing.lic
```diff
@@ -24,7 +24,7 @@ module DRCH
     diseased = false
     poisoned = false
 
-    DRC.bput('health', 'You have', 'You feel somewhat tired and seem to be having trouble breathing', 'Your wounds')
+    DRC.bput('health', 'You have', 'You feel somewhat tired and seem to be having trouble breathing', 'Your wounds', "Your body")
```
asechrest, line 27: The double quotes mixed with single quotes is making me twitch. =P

rpherbig:
> Similar vein to #5606, we already have a `when` for this pattern, but it isn't in the bput expected output.

I can't recall bput's exact behavior: does it return the entire line or just the matched portion? I.e. will the when actually trigger?

asechrest:
> > Similar vein to #5606, we already have a `when` for this pattern, but it isn't in the bput expected output.
> 
> I can't recall bput's exact behavior: does it return the entire line or just the matched portion? I.e. will the when actually trigger?

Just the matched portion.

```
>;e echo DRC.bput('look gildaren', 'are Bee Keeper G')
--- Lich: exec1 active.
[exec1]>look gildaren
You are Bee Keeper Gildaren Spiritlore of Elanthia, an Elf.
<SNIP>
>
[exec1: are Bee Keeper G]
```

However, in this case, I think the `bput` is just to generate the block of information given after a `health` command in-game. We then use `reget` and parse line-by-line, which is why I think the when statements can be different than the bput options.

MahtraDR:
> > Similar vein to #5606, we already have a `when` for this pattern, but it isn't in the bput expected output.
> 
> I can't recall bput's exact behavior: does it return the entire line or just the matched portion? I.e. will the when actually trigger?

Thank you for looking, and asking! I had to test and verify.

So,  your actual question, bput returns just the matched portion. Such that 
```ruby
case DRC.bput("glance", "You glance")
when /^You glance down/
  DRC.message("matched You glance down")
when/^You glance/
  DRC.message("matched You glance")
end
```
returns
```
 >;test-bput                                                                                                                                                                                                                                  
 --- Lich: test-bput active.                                                                                                                                                                                                                  
 [test-bput]>glance                                                                                                                                                                                                                           
 You glance down at your empty hands.                                                                                                                                                                                                         
 >                                                                                                                                                                                                                                            
 matched You glance                                                                                                                                                                                                                           
 --- Lich: test-bput has exited. 
```

However, that's not what we're matching here, as @asechrest said before I could finish this post. :)

KatoakDR (approved):

asechrest (approved):

PR #5623, opened 2022-02-26
[scripts] [crossing-repair] Adding force_repair_town override

MahtraDR:
Now that all repairers repair everything... I'd like to spread the load on Ylono a bit, and try a quieter place to repair.

Tested with
```yaml
hometown: Shard
force_repair_town: Steelclaw Clan
```
Works for me. Does not accept the short aliases the args accepts, however, but it's in line with the other force_*_town overrides in that respect.

### DIFF
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -62,7 +62,7 @@ class CrossingRepair
     amount, denom = settings.sell_loot_money_on_hand.split(' ')
     @keep_copper = DRCM.convert_to_copper(amount, denom)
     @repair_withdrawal_amount = settings.repair_withdrawal_amount
-    @hometown_data = get_data('town')[settings.hometown]
+    @hometown_data = get_data('town')[settings.force_repair_town] || get_data('town')[settings.hometown]
 
     # Start performance script for some sweet TDPs.
     start_script('performance', ['noclean']) unless settings.instrument
--- a/crossing-repair.lic
+++ b/crossing-repair.lic
@@ -17,8 +17,8 @@ class CrossingRepair
 
     args = parse_args(arg_definitions)
     $debug_mode_crossing_repair = args.debug || UserVars.crossing_repair_debug
-    town = DRC.get_town_name(args.town || settings.force_repair_town)
     settings = get_settings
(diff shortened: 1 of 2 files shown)


crossing-repair.lic
```diff
@@ -62,7 +62,7 @@ class CrossingRepair
     amount, denom = settings.sell_loot_money_on_hand.split(' ')
     @keep_copper = DRCM.convert_to_copper(amount, denom)
     @repair_withdrawal_amount = settings.repair_withdrawal_amount
-    @hometown_data = get_data('town')[settings.hometown]
+    @hometown_data = get_data('town')[settings.force_repair_town] || get_data('town')[settings.hometown]
```
KatoakDR, line 65: 💡 This reminds me that [sell-loot](https://github.com/rpherbig/dr-scripts/blob/6531ea5aea2c4c47013a5c3465c333516531733a/sell-loot.lic#L39) does something similar and that perhaps we should follow that model. In which case [line 20](https://github.com/rpherbig/dr-scripts/pull/5623/files#diff-c6a8fe5c06adac610cb24575dd42aa478124886363fd89b715026f2070e03a64R20) in crossing-repair is where the consideration of the `force_repair_town` would take place.

```ruby
town = DRC.get_town_name(args.town || settings.force_repair_town)
```

Doing it this way also means that the values specified in `force_repair_town` can be anything that `DRC.get_town_name` supports (official and colloquial names).

asechrest (approved):

crossing-repair.lic
```diff
@@ -17,8 +17,8 @@ class CrossingRepair
 
     args = parse_args(arg_definitions)
     $debug_mode_crossing_repair = args.debug || UserVars.crossing_repair_debug
-    town = DRC.get_town_name(args.town)
     settings = get_settings
+    town = DRC.get_town_name(args.town || settings.force_repair_town)
```
KatoakDR, line 21: Please also add `force_repair_town` to `base.yaml`, thanks!

KatoakDR:
Works as advertised, thanks!

I tested with the following to also test that shorthand/colloquial names worked, too
```yaml
force_repair_town: Fang
```

PR #5624, opened 2022-02-27
[scripts] [theurgy] fix prayer mat, and use common-methods

MahtraDR:
Trying to pin down some prayer mat issues where my cleric is stuck with an unrolled mat, trying to leave, and can't.

Also noticed we didn't even call common-theurgy... This needs a proper cleanup/rewrite someday.

depends on https://github.com/rpherbig/dr-scripts/pull/5678

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -326,8 +326,10 @@ module DRCMM
     moon = visible_moons.first
     if moon
       data['cast'] = "cast #{moon}"
+    elsif !moon && data['name'].downcase == 'cage of light'
+      data['cast'] = "cast ambient"
     else
-      echo "No moon available to cast #{data['abbrev']}"
+      echo "No moon available to cast #{data['name']}"
       data = nil
     end
     data


theurgy.lic
```diff
@@ -319,7 +319,7 @@ class TheurgyActions
     complete_or_interrupt_research
     cast_spell({ 'abbrev' => 'bless',
                  'mana' => 1,
-                 'prep_time' => 2,
+                 'prep_time' => 4,
```
KatoakDR, line 322: Should this delegate to a waggle so players can account for nuanced prep time/etc for them to cast the spell?

theurgy.lic
```diff
@@ -519,7 +529,7 @@ class TheurgyActions
       complete_or_interrupt_research
       cast_spell({ 'abbrev' => 'bless',
                    'mana' => 1,
-                   'prep_time' => 2,
+                   'prep_time' => 4,
```
KatoakDR, line 532: This is a second place that casts bless, thoughts to move this to its own method?

theurgy.lic
```diff
@@ -756,18 +766,18 @@ class TheurgyActions
...
     return unless DRCI.get_item?("wine", @theurgy_supply_container)
     if @prayer_mat
-      fix_standing
-      bput("kneel #{@prayer_mat}", 'You humbly kneel')
+      DRC.fix_standing
+      DRC.bput("kneel #{@prayer_mat}", 'You humbly kneel')
     end
     complete_or_interrupt_research
-    cast_spell({ 'abbrev' => 'bless',
+    DRCA.cast_spell({ 'abbrev' => 'bless',
```
KatoakDR, line 773: Casting bless again :)

theurgy.lic
```diff
@@ -201,7 +201,7 @@ class TheurgyActions
 
   def last_favor_aspect
     favor_re = /Your most recent favor was granted by (.*)\./
-    match = bput('favor', 'You are not currently favored', favor_re)
+    match =DRC.bput('favor', 'You are not currently favored', favor_re)
```
rpherbig, line 204:
```suggestion
    match = DRC.bput('favor', 'You are not currently favored', favor_re)
```

theurgy.lic
MahtraDR: Turns out there is a method to quick-cast bless in common-theurgy...

MahtraDR:
I've expanded the scope a bit, and went full hog removing includes, and adding all the prefixes. I'll leave it running on my now-two clerics for a few days.

rpherbig:
> I've expanded the scope a bit, and went full hog removing includes, and adding all the prefixes

I understand the desire to improve the script, but doing everything at once makes the review much harder

MahtraDR:
> > I've expanded the scope a bit, and went full hog removing includes, and adding all the prefixes
> 
> I understand the desire to improve the script, but doing everything at once makes the review much harder

Fair enough. I can do one include at a time? What's a morsel size that is reasonably disgetible?

rpherbig:
> > > I've expanded the scope a bit, and went full hog removing includes, and adding all the prefixes
> > 
> > 
> > I understand the desire to improve the script, but doing everything at once makes the review much harder
> 
> Fair enough. I can do one include at a time? What's a morsel size that is reasonably disgetible?

I'd even be fine with one PR for the common* changes and one for the real changes. To me, the difference is something fairly mechanical (common* changes) vs "interesting" (theurgy changes)

theurgy.lic
```diff
@@ -445,21 +435,18 @@ class TheurgyActions
...
+    case DRC.bput('commune tamsine',
                  'completed this commune too recently',
                  'You struggle to commune',
                  'you have attempted a commune too recently in the past',
                  'You feel warmth spread throughout your body')
-    DRCI.put_away_item?(@water_holder, @theurgy_supply_container)
-
-    match == 'You feel warmth spread throughout your body'
+    when /You feel warmth spread throughout your body/
+      return true
```
rpherbig, line 448: Doesn't this need `false` at the end to be explicit? I'm not super clear on what the return value of a case/when is when no branch was taken...

theurgy.lic
```diff
@@ -511,10 +498,7 @@ class TheurgyActions
     elsif DRSpells.known_spells.include? 'Bless'
       DRCI.get_item_safe('some oil', @theurgy_supply_container)
       complete_or_interrupt_research
-      DRCA.cast_spell({ 'abbrev' => 'bless',
-                   'mana' => 1,
-                   'prep_time' => 2,
-                   'cast' => 'cast oil' }, @settings)
+      DRCTH.quick_bless_item('oil', @settings)
```
rpherbig, line 501: Should `DRCTH` also have a quick_gg function? See https://github.com/rpherbig/dr-scripts/blob/64940ff0d16f7246a032fe711b42b83af2219bbc/theurgy.lic#L463

I think that would let us remove `common-arcana` from this file, too.

MahtraDR: Two thoughts. How about a quick-cast method in DRCA where we can specify the spell? 

and/or (and this is something I planned to add here anyway) we allow for users to define their own quick* waggles, since the main thing I was trying to solve here anyway, is the fact that some clerics can't always quick-cast at 2 seconds even at 1 mana (low concentration, or other reasons) that would allow users to work around our defaults if they aren't suited.

Issue #1100, opened 2016-12-02
Hib Hometown


wscherer420:
Below you'll find the roomnumbers and info to setup hometown for Hib.  Let me know if you need anything else.  

HIB MAPPING

```
Hib:
  currency: dokora
  leather_repair:
    id: 11211
    name: Ladar
  metal_repair:
    id: 11211
    name: Ladar
  deposit:
    id: 11214
  exchange:
    id: 11213
  gemshop:
    id: 8899
    name: rasiya
  tannery:
    id: 8885
    name: Vandorf
  npc_empath:
    id: 8881
  locksmithing:
    id: 
  theurgy_supplies:
    id: 
```

PR #5625, opened 2022-02-27
[script] [hunting-buddy] Add option to pick a room to go to after hunting-buddy ends instead of default safe-room

PatricCampbell:
Does just what the title says, adds a new field force_after_hunt_room to overwrite where hunting-buddy goes when it ends.

Q: Do I need to add something to validator for this to check it's a number?

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -156,8 +156,8 @@ class HuntingBuddy
       execute_actions(after_actions)
     end
 
-    DRC.message("***STATUS*** Returning to safe room: #{@settings.after_hunt_room}")
-    DRCT.walk_to(@settings.after_hunt_room)
+    DRC.message("***STATUS*** Returning to safe room: #{@after_hunt_room}")
+    DRCT.walk_to(@after_hunt_room)
     EquipmentManager.new(@settings).wear_equipment_set?('standard')
   end
 
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -25,6 +25,7 @@ class HuntingBuddy
 
     # Will deprecate prehunt_buffs tag as a room
     @buff_room = @settings.prehunt_buffing_room || @settings.prehunt_buffs
+    @after_hunt_room = @settings.force_after_hunt_room || @settings.safe_room
 
     hunting_info = []
     if @settings.hunting_file_list || args.flex.any?
@@ -155,8 +156,8 @@ class HuntingBuddy
       execute_actions(after_actions)
     end
 
-    DRC.message("***STATUS*** Returning to safe room: #{@settings.safe_room}")
-    DRCT.walk_to(@settings.safe_room)
+    DRC.message("***STATUS*** Returning to safe room: #{@settings.after_hunt_room}")
+    DRCT.walk_to(@settings.after_hunt_room)
     EquipmentManager.new(@settings).wear_equipment_set?('standard')
   end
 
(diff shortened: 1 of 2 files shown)


KatoakDR:
👋 Thanks for the submission and following the PR title convention. Great idea to make where characters go after a hunt configurable.

hunting-buddy.lic
```diff
@@ -26,6 +26,8 @@ class HuntingBuddy
     @stop_on_high_threshold = @settings.stop_on_high_threshold
     echo("  @stop_on_high_threshold: #{@stop_on_high_threshold}") if $debug_mode_hunting
 
+    @after_hunt_room = @settings.force_after_hunt_room || @settings.safe_room
```
KatoakDR, line 29: 💡 Please add an echo after line 29 like like 27 to show what the variable value ended up being. Helpful for troubleshooting. Thanks

profiles/base.yaml
```diff
@@ -537,6 +537,9 @@ safe_room_tip_amount:
 safe_room_empath:
 safe_room_empaths:
 
+# override the room you go to when hunting-buddy ends
+force_after_hunt_room:
```
KatoakDR, line 541: 💡 We also have a "prehunt_buffing_room" setting, thoughts if this new setting were named "posthunt_safe_room"?

I see why you would have chosen "force", and I suppose to me the force variables tend to be isolated to overriding the hometown.

profiles/base.yaml
```diff
@@ -537,6 +537,9 @@ safe_room_tip_amount:
 safe_room_empath:
 safe_room_empaths:
 
+# override the room you go to when hunting-buddy ends
```
rpherbig, line 540: I suggest adding a note that the default if this is not provided is `safe_room`. But it's not a big deal either way.

rpherbig (approved):

KatoakDR:
> Do I need to add something to validator for this to check it's a number?

That would be helpful; assuming there exists validation logic for other room settings then you should be able to copy/paste most of what they do, thanks

KatoakDR:
Hi @PatricCampbell, haven't heard any follow up from you on the code comments. Going to switch this back to draft status. When/if you're ready to continue simply follow up on the open comments. Thanks!

Dartellum:
`full_waiting_room:` is what you are looking for:
```yaml
---
#combat_buff_waggle:
hunting_info:
- :zone:
  - cabalist #1100-1400
  - cabalist #1100-1400
  args:
  #- r3
  - d0
  full_waiting_room:
  - 10989
  before:
  #:duration: 140
  stop_on:
  #- Skinning
  #- Targeted Magic
  stop_on_low:
  - Athletics
  during:
  after:

#combat_buff_waggle:

buff_nonspells:
  pathway focus quick: 300
  slap cabalist: 60
  exhale: 63
  fill black pouch with lootpouch: 60
  # retreat: 596
  # retreat: 598
  # summon air domain: 600
  # slam left: 60
  ```

Dartellum:
Why would you not want to go to your safe room?

Promithius-DR:
> Why would you not want to go to your safe room?

An example would be, I've got 7 different hunts multiple near Boar clan, some near Shard, and a few near Crossing.  Running back to shard each time only to head back to either Boar or Crossing is a huge waste of time.   Going to a safe room near your last hunt and then checking where it needs to go next just saves time

MahtraDR:
Closing as abandoned.

Issue #5637, opened 2022-03-05
Necro combat with wounds


Kaesken:
Unfortunately I have to sometimes manually have her eat herbs to heal (long story short, extremely old toon, trained poorly so she doesn't have healing spells yet).  I forgot to check her herb intake and got this during her hunting, figured you guys would like a copy of it.

```
[combat-trainer]>wield my scimitar

Your right hand is too injured to draw the scimitar!
> 
* A sickly blightwater nyad feints low at you.  You partially block with a leather shield reinforced with riveted burgundy leather straps.  
[You're solidly balanced with opponent in good position.]
The blightwater nyad moves into a position to parry.
> 
* A sickly blightwater nyad thrusts a rotting driftwood staff at you.  You barely block with a leather shield reinforced with riveted burgundy leather straps.  
[You're solidly balanced with opponent in superior position.]
* A sickly blightwater nyad sidesteps and bashes at you.  You barely block with a leather shield reinforced with riveted burgundy leather straps.  
[You're solidly balanced with opponent in very strong position.]
You've gained a new rank in your perceptual acuity.
> 
The blightwater nyad moves into a position to parry.
> 
[combat-trainer: *** No match was found after 15 seconds, dumping info]

[combat-trainer: messages seen length: 12]

[combat-trainer: message: The blightwater nyad moves into a position to parry.]

[combat-trainer: message: You've gained a new rank in your perceptual acuity.]

[combat-trainer: message: [You're solidly balanced with opponent in very strong position.]]

[combat-trainer: message: * A sickly blightwater nyad sidesteps and bashes at you.  You barely block with a leather shield reinforced with riveted burgundy leather straps.  ]

[combat-trainer: message: [You're solidly balanced with opponent in superior position.]]

[combat-trainer: message: * A sickly blightwater nyad thrusts a rotting driftwood staff at you.  You barely block with a leather shield reinforced with riveted burgundy leather straps.  ]

[combat-trainer: message: Butcher's Eye  (2 roisaen)]

[combat-trainer: message: Obfuscation  (18 roisaen)]

[combat-trainer: message: The blightwater nyad moves into a position to parry.]

[combat-trainer: message: [You're solidly balanced with opponent in good position.]]

[combat-trainer: message: * A sickly blightwater nyad feints low at you.  You partially block with a leather shield reinforced with riveted burgundy leather straps.  ]

[combat-trainer: message: Your right hand is too injured to draw the scimitar!]

[...]

PR #5639, opened 2022-03-05
[script][equipmanager.lic] - Fix wield weapon match failure with hand injury

asechrest:
Reported via #5637 

This fixes the match hang. I do wonder, though, whether the `wield` verbs balks if you have a bad hand injury, but the `get` verb would still work to get a weapon in your hand?

### DIFF
--- a/equipmanager.lic
+++ b/equipmanager.lic
@@ -243,7 +243,12 @@ class EquipmentManager
   def get_item?(item)
     return true if DRCI.in_hands?(item)
     if item.wield
-      bput("wield my #{item.short_name}", 'You draw', 'You deftly remove', 'You slip', 'Wield what', 'With a flick of your wrist you stealthily unsheathe') != 'Wield what'
+      case bput("wield my #{item.short_name}", 'You draw', 'You deftly remove', 'You slip', 'With a flick of your wrist you stealthily unsheathe', 'Wield what', 'Your left hand is too injured', 'Your right hand is too injured')
+      when 'Your left hand is too injured', 'Your right hand is too injured', 'Wield what'
+        return false
+      else
+        return true
+      end
     elsif item.transforms_to
       item = item_by_desc(item.transforms_to)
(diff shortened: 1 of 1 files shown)


MahtraDR:
I'm not sure just fixing the match so there is t a hang is appropriate here. This needs more thinking to address the "ok too damaged to hold weapons, what do I do now?" Question. Do they even stay in combat then? Necro make things complicated, they can't just go get healed by an empath.

KatoakDR:
This script is about handling your gear regardless the context, so I lean towards improving match strings to avoid hang ups.

A script like hunting-buddy or combat-trainer whose context is combat and goal is to keep the character alive should be monitoring the character's health and escaping as appropriate.

For example, if combat-trainer tries to wield a weapon but equipment-manager returns false, then that's a signal that either (a) they aren't able to hold it, or (b) their gear config is wrong. Whether the combat script stays in combat or not I think is the responsibility of the calling script, not equipment-manager.

My two kronars :)

MahtraDR (approved):

asechrest:
> For example, if combat-trainer tries to wield a weapon but equipment-manager returns false, then that's a signal that either (a) they aren't able to hold it, or (b) their gear config is wrong. Whether the combat script stays in combat or not I think is the responsibility of the calling script, not equipment-manager.
> 
> My two kronars :)

The only additional thing I'm thinking about here is this. Is it a product of DR's silly parser that you can't `wield` a weapon when you have an injured hand but you can `get` a weapon instead, and still function properly for combat training?

Because if that's the case, we probably need to handle it in equipment manager and not in the calling script, since under our current predicate function paradigm, the calling script only knows whether we did or did not successfully wield the weapon, and has no understanding of the granularity of failing to wield; e.g. whether the weapon doesn't exist or we're too injured to wield.

rpherbig (approved):
I think fixing the hang is the first step and the PR is fine as is.

If a volunteer wants to look into the other ideas, all the better, but that shouldn't hold up this fix.

PR #5640, opened 2022-03-05
Update base-hunting.yaml

asechrest:
Remove room 16538 from cloud_rats. Terrible spawn rate. One or two mobs per 10 minutes in my estimation. Slow enough to be nearly worthless for combat training.

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -413,7 +413,7 @@ hunting_zones:
   - 1446
   - 1447
   - 1448
-  silverfish:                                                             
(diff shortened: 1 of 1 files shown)


MahtraDR (approved):
I was a bit taken aback by the white space changes, so ended up testing it. Looks good.

rpherbig:
I like commenting it out so it isn't accidentally reintroduced. It might be nice to also include the 'why' so w know if the issue gets fixed and the room can be uncommented in the future.

PR #5644, opened 2022-03-08
[data] [base-hunting] Updating note for cave trolls, and cloud rats

MahtraDR:
Spoke with a GM about these rooms.

Confirmed Cave Troll room is not meant to be a spawn room. Keep the comment, updated the note.

Cloud Rats had the room reinstated. That room was erroneously set as a non-spawn room.

Convo link here https://discordapp.com/channels/619301383451181075/831336521620652032/950521974369681419

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -1054,7 +1054,7 @@ hunting_zones:
   - 19195
   - 19196
   - 19204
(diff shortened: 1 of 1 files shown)


asechrest:
Nice work. Thank you.

asechrest (approved):

Issue #5649, opened 2022-03-10
Combat Trainer missing weather messaging


mwest152:
I'm pretty sure this is missing messaging.  This happened hunting in Cabalists.
---------------------------------
[combat-trainer]>weather

You glance outside.
It is completely overcast.

[combat-trainer: message: It is completely overcast.]

[combat-trainer: message: You glance outside.]

[combat-trainer: checked against [/inside/i, /You glance up at the sky./i]]

[combat-trainer: for command weather]

asechrest:
Easy fix for later this evening when I'm home. Thanks for the report.

asechrest:
Closed via #5652

asechrest:
Closed via #5652

PR #5652, opened 2022-03-11
[script][combat-trainer] Fix weather check match hang

asechrest:
Reported by @mwest152 via #5649 

The combat-trainer weather check is missing a possible game output from the `weather` command. Added.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1910,7 +1910,7 @@ class SpellProcess
       unless moon
         echo "No moon available to cast #{data['abbrev']}" if $debug_mode_ct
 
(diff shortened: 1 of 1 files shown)


rpherbig (approved):

MahtraDR:
Should we just match on `You glance` ?

asechrest:
> Should we just match on `You glance` ?

I opted not to, to avoid the rare possibility of this catching any other glance like glancing at your hands, etc.

MahtraDR:
> > Should we just match on `You glance` ?
> 
> I opted not to, to avoid the rare possibility of this catching any other glance like glancing at your hands, etc.

Very reasonable. Thanks. I hadn't thought of that.

PR #5668, opened 2022-03-17
[scripts] [go2] Merging go2 from lich5 to avoid conflicts during the transision

MahtraDR:
In preparation to moving to lich5, our go2 has features lich5's doesn't have. I've merged those into lich5's `go2` in https://github.com/elanthia-online/lich-5/pull/129

as we're testing new DR installs, we noticed that dependency will overwrite their go2, creating a problem for users of both games.

This is the lich5 go2, with our changes merged in, so users with both games don't lose functionality.

Once we're using the lich5 installer, I intend to delete this file from our repo, and we use the lich5 version (which has all our changes anyway)

### DIFF
--- a/go2.lic
+++ b/go2.lic
@@ -1,72 +1,86 @@
 =begin
 
-	Fork of go2- to fix dragging in DR
-	attempts to find the shortest route between any two rooms in the game
-	requires a map database (;repository download-mapdb)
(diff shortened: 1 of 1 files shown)


MahtraDR:
Note I have been using this version since my original PR to the lich5 repo 8 days ago. No issues.

rpherbig (approved):

PR #5670, opened 2022-03-18
[scripts] [dependency] Dependency kill drinfomon before dying

MahtraDR:
if we don't kill drinfomon, we can start it again when scripts need it. This resolves the obnoxious bootstrap hanging issues.


rcuhljr:
Specifically because the next run of dependency will wipe out all known script constants bootstrap thinks drinfomon isn't running and then fails to start it as it is in fact still running thus hanging forever.

PR #5674, opened 2022-03-18
[scripts] [Bootstrap] fix typo

MahtraDR:
It wouldn't catch often, if at all, but we found it whilst debugging, and figured it was worth fixing.


bootstrap.lic
```diff
@@ -65,7 +65,7 @@ until scripts_to_run.empty?
   respond("BS:#{script_to_run}:#{scripts_to_run}") if UserVars.bootstrap_debug
```
rpherbig, line 65: Can you fix this line, too?

PR #5685, opened 2022-03-22
[script][bescort] Add support for asketi's mount traversing (Change 1 of 3)

asechrest:
Adds support for traversing asketi's mount up and down in the Black Marble Gargoyles area via `bescort`.

1. Fix some travel failures resulting from the move from Lich4 to Lich5 which arise due to Lich4's overloading of the FalseClass method and Lich5's refusal to do so. These travel failures are best fixed in the map stringprocs but, in the meantime, we'll build in a fix. These fixes will live in `base.yaml` and their implementation will be handled in `;go2` and pass through a DR game and Lich 5+ version check before being deployed, since those still on Lich4 don't need the fix.

This PR sets the following:

1. Add args and methods to `bescort.lic` to traverse asketi's mount. Subsequent PRs will utilize this method.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -365,10 +365,10 @@ class Bescort
   end
 
   def asketis_mount(mode)
-    if mode.include?('up')
+    if mode == "up"
       while bput("climb up", /^You work your way/i, /^You climb over/, /^You can't do that/i) =~ /^You work your way/i
       end
-    elsif mode.include?('down')
+    elsif mode == "down"
       while bput("climb down", /^You work your way/i, /^You climb down/, /^You can't do that/i) =~ /^You work your way/i
       end
     end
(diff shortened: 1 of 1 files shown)


MahtraDR (approved):
Ran up and down, both worked clean, as expected.

bescort.lic
```diff
@@ -358,6 +364,16 @@ class Bescort
     end
   end
 
+  def asketis_mount(mode)
+    if mode.include?('up')
```
rcuhljr, line 368: Minor nitpicks, `mode` feels like an odd term as opposed to direction or something similar. Also include? feels like an odd choice here for matching a string when the fact that up/down is coming through parse args provides us some fairly strict guarantees.

asechrest: I thought `mode` was a little strange but it's what's used in all the other arg sets that take a second arg. I modeled my method partially off of [this one](https://github.com/rpherbig/dr-scripts/blob/7bff1eb59a7e84bb968d611678886c9974b04486/bescort.lic#L516).

You thinking something like
```ruby
if mode == /up/i
```

rcuhljr: `.match?` if using a regex is best practice at this point I believe, but you can just use `== "up"` I'm fairly certain arg parser forces exact matches and down cases them.

PR #5688, opened 2022-03-22
[script][go2.lic] - Add support for custom map wayto stringproc overrides (Change 3 of 3)

asechrest:

1. Fix some travel failures resulting from the move from Lich4 to Lich5 which arise due to Lich4's overloading of the FalseClass method and Lich5's refusal to do so. These travel failures are best fixed in the map stringprocs but, in the meantime, we'll build in a fix. These fixes will live in base.yaml and their implementation will be handled in ;go2 and pass through a DR game and Lich 5+ version check before being deployed, since those still on Lich4 don't need the fix.

This PR sets the following:

1. Adds support for processing of custom map wayto stringproc overrides
2. Does the following checks: is game DR? If so, and base.yaml stringproc overrides exist, is user's Lich version 5+?
3. Merges the base overrides and personal overrides into a single hash, favoring the user's personal overrides in the case of identical hash keys.

### DIFF
--- a/go2.lic
+++ b/go2.lic
@@ -161,6 +161,46 @@ show_help = proc {
    respond output
 }
 
+# For Dragonrealms only
+# A change in behavior from Lich4 to Lich5 caused certain ;go2 map wayto movements to fail.
+# This is because of the way the map stringprocs are structured, and the method used to evaluate
+# certain map moves, which involved an arguably dubious overloading of FalseClass on Lich4.
+# The transition to Lich5 makes it necessary to fix some stringprocs as an interim solution, at
+# least until the map can be divested from current owner control and handled appropriately. This code
+# also allows the flexibility to pull custom map wayto overrides from the standard yaml hierarchy.
+if XMLData.game =~ /^DR/
+  # Get yaml settings
+  settings = get_settings
+
+  # Get base wayto overrides. These base map wayto overrides were created to fix some travel hiccups
+  # from converting from Lich4 to Lich5. They are not needed for Lich4 users, so we'll check Lich
+  # version shortly.
+  base_wayto_overrides = settings.base_wayto_overrides
+  personal_wayto_overrides = settings.personal_wayto_overrides
+
+  # Notify user we've detected and are using personal map stringproc overrides
+
+  # Merge the two hashes into a single hash, favoring personal overrides for duplicate keys.
+  # The reason for favoring personal overrides is that any user who has the ability to create
+  # custom map stringprocs can be trusted to have their customization favored vs. the base overrides.
+  # If they break the functionality, that's on them.
+  if LICH_VERSION[0].to_i >= 5
+    wayto_overrides = base_wayto_overrides.merge(personal_wayto_overrides)
+    echo "Using base.yaml's Lich5+ map wayto stringproc overrides."
+    unless personal_wayto_overrides.empty?
+      echo "Using your personal map wayto stringproc overrides, which take precendence over base."
+    end
+  else
+    wayto_overrides = personal_wayto_overrides
+  end
+
+  # Iterate through the aggregated map wayto overrides from above and set new stringprocs
+  wayto_overrides.each do | key, values |
+    Map.list[values['start_room'].to_i].wayto["#{values['end_room']}"] = StringProc.new("#{values['str_proc']}")
+    echo values['str_proc']
+  end
+end
+
 change_map_vaalor_shortcut = proc { |use_shortcut|
    unless Map.list.any? { |room| room.timeto.any? { |adj_id,time| time.class == Proc and time._dump =~ /$go2_use_vaalor_shortcut/ } }
       if use_shortcut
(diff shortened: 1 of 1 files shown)


asechrest:
NOTE: We expect there will be more travel failures as a result of the root cause of the failure mentioned here. I've tried to design this solution such that simple one-room-to-next failures can be fixed by adding to base.yaml and calling it a day. Nothing more needs to be known than the start room, the end room, and the proper stringproc.

go2.lic
```diff
@@ -161,6 +161,44 @@ show_help = proc {
...
+# certain map moves, which involved an arguably dubious overloading of FalseClass on Lich4.
+# The transition to Lich5 makes it necessary to fix some stringprocs as an interim solution, at
+# least until the map can be divested from current owner control and handled appropriately. This code
+# also allows the flexibility to pull custom map wayto overrides from the standard yaml hierarchy.
+if XMLData.game =~ /^DR/
+  # Get yaml settings
+  settings = get_settings
+
+  # Get base wayto overrides. These base map wayto overrides were created to fix some travel hiccups
+  # from converting from Lich4 to Lich5. They are not needed for Lich4 users, so we'll check Lich
```
rcuhljr, line 176: Is there not already a lich version check down below this?

go2.lic
```diff
@@ -161,6 +161,44 @@ show_help = proc {
...
+  # version shortly.
+  base_wayto_overrides = settings.base_wayto_overrides
+  personal_wayto_overrides = settings.personal_wayto_overrides
+
+  # Merge the two hashes into a single hash, favoring personal overrides for duplicate keys.
```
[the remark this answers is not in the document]
asechrest: I'm working on my over-commenting addiction. =P

go2.lic
```diff
@@ -161,6 +161,44 @@ show_help = proc {
...
+  # The reason for favoring personal overrides is that any user who has the ability to create
+  # custom map stringprocs can be trusted to have their customization favored vs. the base overrides.
+  # If they break the functionality, that's on them.
+  if LICH_VERSION[0].to_i >= 5
+    wayto_overrides = base_wayto_overrides.merge(personal_wayto_overrides)
+    echo "Using base.yaml's Lich5+ map wayto stringproc overrides."
+    # Notify user we've detected and are using personal map stringproc overrides
+    unless personal_wayto_overrides.empty?
+      echo "Using your personal map wayto stringproc overrides, which take precendence over base."
```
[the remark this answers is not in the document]
asechrest: I think that's fair. Will remove.

go2.lic
```diff
@@ -161,6 +161,44 @@ show_help = proc {
...
+  # from converting from Lich4 to Lich5. They are not needed for Lich4 users, so we'll check Lich
+  # version shortly.
+  base_wayto_overrides = settings.base_wayto_overrides
+  personal_wayto_overrides = settings.personal_wayto_overrides
+
+  # Merge the two hashes into a single hash, favoring personal overrides for duplicate keys.
+  # The reason for favoring personal overrides is that any user who has the ability to create
+  # custom map stringprocs can be trusted to have their customization favored vs. the base overrides.
+  # If they break the functionality, that's on them.
+  if LICH_VERSION[0].to_i >= 5
```
rpherbig, line 185: Do we need to check the lich version? In other words, why not _always_ take overrides?

rpherbig (approved):

asechrest:
By the way, here is a log of these 3 PRs working together to traverse Asketi's Mount on my Lich5 setup:

```
>;go2 31491
--- Lich: go2 active.
[go2: Using base.yaml's Lich5+ map wayto stringproc overrides.]
[go2: start_script('bescort', ['asketis_mount', 'up']);wait_while{running?('bescort')};]
[go2: start_script('bescort', ['asketis_mount', 'down']);wait_while{running?('bescort')};]
[go2: ETA: 0:00:00 (3 rooms to move through)]
[go2]>east
You trek east.
[Hag's Crag, Broad Ledge]
Encrusted with ice and snow, a sheer granite cliff towers above.  Heavy with moisture, vaporous clouds encircle the area like ghostly fogbanks upon a churning sea, hindering movement and making travel hazardous.  
Obvious paths: west.
Room Number: 27556
>
[go2]>climb cliff
You work your way about 22 feet up the cliff.

[Asketi's Mount, Treacherous Cliff]
Dense clouds obscure the view below, while an incredibly steep cliff soars out of sight into the icy mists overhead.  At eye level are slippery mountain walls, marred with craggy holes left by pitons and gouges left by ice axes.  Whistling winds rush down from the summit, trailing plumes of frost crystals that threaten to freeze the breath in one's lungs.  
Obvious paths: none.
Roundtime: 10 seconds.
Room Number: 27558
>
--- Lich: bescort active.
--- Lich: spellmonitor active.
--- Lich: events active.
--- Lich: equipmanager active.
--- Lich: equipmanager has exited.
--- Lich: skill-recorder active.
>
[bescort: In asketi's mount bescort super new]
[bescort]>climb up
You work your way about 20 feet up the cliff.
Roundtime: 10 seconds.
>
[spellmonitor]>spells
You need to concentrate on your climbing, not that.
>
[bescort]>climb up
You work your way about 20 feet up the cliff.
Roundtime: 10 seconds.
```

go2.lic
asechrest: I think my hesitancy to force it on Lich4 users is that it's a change in behavior that is not necessary except for the Lich5 switch. And I'm gun-shy about forcing change in behavior after having been resoundingly scolded for it previously. :)

But stepping back for a moment from this specific Asketi traversing issue that cropped up when testing with Lich5, this should be backward-compatible with Lich4 and maybe we don't need the version check, which would open this functionality up to all Lich version users.

MahtraDR (approved):

PR #5690, opened 2022-03-23
[script] [dependency] skip drinfomon autostart, and remove from autostart if found

MahtraDR:
Since we now start drinfomon as script zero directly from dependency, we no longer need it autostarted.

Some folks still have that autostart, which causes weird "drinfomon is already running" messages.

With this change, we make a note, and remove drinfomon from autostarts

```
drinfomon found in autostarts. Attempting to remove it now. If unsuccessful, it can be safely removed with `;e stop_autostart('drinfomon')`
```
This was successful.
```
>;e echo list_autostarts
--- Lich: exec1 active.
[exec1: ["sanowret-crystal", "almanac", "esp", "smartlisten", "play", "log", "tarantula", "afk", "sorter", "exp-monitor", "textsubs", "roomnumbers", "tdps"]]
--- Lich: exec1 has exited.    
```

This is similar logic to what we did with lich5's autostart for dependency in https://github.com/elanthia-online/lich-5/pull/179

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -330,7 +330,7 @@ class ScriptManager
     repo_scripts.each { |(name, _)| get_script(name, force) }
     autostarts.each do |script|
       if script == 'drinfomon'
-        respond "\n---\ndrinfomon found in autostarts. Attempting to remove it now. If unsuccessful, it can be safely removed with `;e stop_autostart('drinfomon')`\n---\n"
+        respond "\n---\ndrinfomon found in autostarts. Attempting to remove it now. If unsuccessful, it can be safely removed with `#{$clean_lich_char}e stop_autostart('drinfomon')`\n---\n"
         stop_autostart('drinfomon')
         next
       end
--- a/dependency.lic
+++ b/dependency.lic
@@ -330,7 +330,7 @@ class ScriptManager
     repo_scripts.each { |(name, _)| get_script(name, force) }
     autostarts.each do |script|
       if script == 'drinfomon'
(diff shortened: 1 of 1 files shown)


dependency.lic
```diff
@@ -329,6 +329,11 @@ class ScriptManager
     @versions = nil
     repo_scripts.each { |(name, _)| get_script(name, force) }
     autostarts.each do |script|
+      if script == 'drinfomon'
+        respond "\n---\ndrinfomon found in autostarts. Attempting to remove it now. If unsuccessful, it can be safely removed with `;e stop_autostart('drinfomon')`\n---\n"
```
rpherbig, line 333: Should `;e` be `#{$clean_lich_char}e`?

MahtraDR: I've changed it, but just to note that genie, the only FE I know of that uses a different character, now also accepts `;`

MahtraDR:
> > Since we now start drinfomon as script zero directly from dependency, we no longer need it autostarted.
> 
> Does this also mean we can remove `drinfomon` from all `custom_require` calls?

I think not, actually.

What we've done is prevent a double start at login. 

custom_require calls would still apply if for some reason drinfomon gets killed mid-session, no?

rpherbig:
> > > Since we now start drinfomon as script zero directly from dependency, we no longer need it autostarted.
> > 
> > 
> > Does this also mean we can remove `drinfomon` from all `custom_require` calls?
> 
> I think not, actually.
> 
> What we've done is prevent a double start at login.
> 
> custom_require calls would still apply if for some reason drinfomon gets killed mid-session, no?

Makes sense to me

rpherbig (approved):

PR #5712, opened 2022-04-02
[script] [dependency] - Foundation for robust help output

asechrest:
These changes to dependency provide part of a foundation for robust in-game help output for all curated scripts. Up-front thanks to @rcuhljr who gave me a huge assist with the core code.

Help output will consist of 3 elements and be provided via a `;<script-name> help` command:

1. A script summary (new)
2. An example of how to call the script and the arguments, and a list of the arguments with descriptions (existing)
3. A list of the yaml settings used in the script, including a general description of the setting as well as a specific description of how the settings is used in the script. (new)

Example of help output: 
image

Display of this help output is fully driven by ArgParser. The structure for help output was already there so I felt like it was a good fit to broaden within it.

The new script summary will pull via ArgParser from a script argument called `script_summary`. Example from my locally edited circlecheck script:

```yaml
def main
  arg_definitions = [
    [
      { name: 'debug', regex: /debug/i, optional: true, description: 'Runs the script with debug messages.' },
      { name: 'mode', options: %w[brief short next], optional: true, description: 'Only display requirements you don\'t meet' },
      { name: 'target', regex: /^\d+$/, optional: true, description: 'See requirements for a specific circle' },
      { name: 'script_summary', optional: true, description: 'Checks your circle and skills to next circle. Optionally checks skills for a target circle.'}
    ]
  ]
```

The new section listing yaml settings used in the script will pull from a "yaml of yamls" called `base-help.yaml` via a `get_data` call. The information here needs to be built manually. We discussed trying to automate the pulling of yaml settings help information by building some parser like ArgParser but for yaml settings, but this "yaml of yaml" approach was determined to be the least intensive for existing script changes while also allowing full backward compatibility, _as well as_ doing little harm if I fall off the face of the Earth and no one cares to update the help information in the future.

Note: all of these changes will be fully backward compatible. If no `script_summary` argument exists, no problem, script functions as-is. If no yaml settings data is built out in base-help.yaml for a script, no sweat, the information just won't be available in the help output.

One final thing of note: because the help output is driven by ArgParser, scripts that don't take an argument and thus don't call `parse_args` will have no help output available unless we implement a `script_summary` argument and a call to ArgParser. I'd like to do that within those scripts, though I recognize it could be considered a superfluous call just for the sake of help output, so I'll take ideas here if you have another idea.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -85,7 +85,7 @@ class ArgParser
 
       # Display help output for settings used in the script. Relies on base-help.yaml.
       yaml_data = get_data('help').to_h
-      if !yaml_data.select { |field, info| info["referenced_by"].include?(Script.current.name)}.empty?
+      unless yaml_data.select { |field, info| info["referenced_by"].include?(Script.current.name)}.empty?
         respond " YAML SETTINGS USED:"
         yaml_data.select { |field, info| info["referenced_by"].include?(Script.current.name)}
                   .each do | field, info |
--- a/dependency.lic
+++ b/dependency.lic
@@ -50,7 +50,7 @@ class ArgParser
 
   def display_args(data)
     # Display help output for script arguments
-    unless Script.current.name == "bootstrap"
+    if Script.current.name != "bootstrap"
       data.each do |def_set|
         respond ''
         def_set.each { |x| respond " SCRIPT SUMMARY: #{x[:description]} " if x[:name].to_s == "script_summary"; }
--- a/dependency.lic
+++ b/dependency.lic
@@ -53,7 +53,9 @@ class ArgParser
     if Script.current.name != "bootstrap"
       data.each do |def_set|
         respond ''
-        def_set.each { |x| respond " SCRIPT SUMMARY: #{x[:description]} " if x[:name].to_s == "script_summary"; }
+        def_set
+          .select { |x| x[:name].to_s == "script_summary" }
+          .each { |x| respond " SCRIPT SUMMARY: #{x[:description]} " }
         respond ''
         respond " SCRIPT CALL FORMAT AND ARG DESCRIPTIONS:"
         respond "  ;#{Script.current.name} " + def_set.map { |x| format_item(x) }.join(' ')
--- a/dependency.lic
+++ b/dependency.lic
@@ -49,7 +49,26 @@ class ArgParser
   end
 
   def display_args(data)
-    # Display help output for script arguments
+    # This section displays help output for a script. It recognizes when a script call has been given
+    # with the wrong number or incorrect arguments, or when a script has been called via
+    # ;script-name help. It outputs an example script call, including arguments, with optional
+    # arguments enclosed in brackets. If an optional script_summary argument has been specified within
+    # the script, it also outputs a summary of the script. Further, if yaml settings have been
+    # outlined in base-help.yaml, it will output help information for the yaml settings used.
+    # Example output when all help settings have been specified:
+    #
+    # >;circlecheck help
+    #   SCRIPT SUMMARY: Checks your circle and skills to next circle. Optionally checks skills for a target circle.
+
+    #   SCRIPT CALL FORMAT AND ARG DESCRIPTIONS (arguments in brackets are optional):
+    #    ;circlecheck [debug] [mode] [target] [script_summary]
+    #     debug        Runs the script with debug messages.
+    #     mode         Only display requirements you don't meet [brief, short, next]
+    #     target       See requirements for a specific circle
+
+    #   YAML SETTINGS USED:
+    #     circlecheck_prettyprint: Tells circlecheck to use the pretty print formatting for output.
+
     if Script.current.name != "bootstrap"
       data.each do |def_set|
         respond ''
@@ -57,7 +76,7 @@ class ArgParser
           .select { |x| x[:name].to_s == "script_summary" }
           .each { |x| respond " SCRIPT SUMMARY: #{x[:description]} " }
         respond ''
-        respond " SCRIPT CALL FORMAT AND ARG DESCRIPTIONS:"
+        respond " SCRIPT CALL FORMAT AND ARG DESCRIPTIONS (arguments in brackets are optional):"
         respond "  ;#{Script.current.name} " + def_set.map { |x| format_item(x) }.join(' ')
         def_set
           .reject { |x| x[:name].to_s == "script_summary" }
--- a/dependency.lic
+++ b/dependency.lic
@@ -59,15 +59,15 @@ class ArgParser
     #
     # >;circlecheck help
     #   SCRIPT SUMMARY: Checks your circle and skills to next circle. Optionally checks skills for a target circle.
-
+    #
     #   SCRIPT CALL FORMAT AND ARG DESCRIPTIONS (arguments in brackets are optional):
     #    ;circlecheck [debug] [mode] [target] [script_summary]
     #     debug        Runs the script with debug messages.
     #     mode         Only display requirements you don't meet [brief, short, next]
     #     target       See requirements for a specific circle
-
+    #
     #   YAML SETTINGS USED:
-    #     circlecheck_prettyprint: Tells circlecheck to use the pretty print formatting for output.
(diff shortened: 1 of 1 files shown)


dependency.lic
```diff
@@ -49,10 +49,26 @@ class ArgParser
...
+        def_set.each { |x| respond " SCRIPT SUMMARY: #{x[:description]} " if x[:name].to_s == "script_summary"; }
+        respond ''
+        respond " SCRIPT CALL FORMAT AND ARG DESCRIPTIONS:"
+        respond "  ;#{Script.current.name} " + def_set.map { |x| format_item(x) }.join(' ')
+        def_set.each { |x| respond "   #{(x[:display] || x[:name]).ljust(12)} #{x[:description]} #{x[:options] ? '[' + x[:options].join(', ') + ']' : ''}" unless x[:name].to_s == "script_summary"; }
+        respond ''
+      end
+
+      # Display help output for settings used in the script. Relies on base-yaml.yaml.
+      yaml_data = get_data('yaml').to_h
```
rpherbig, line 65: What do you think about `base-help.yaml` instead?

dependency.lic
```diff
@@ -49,10 +49,26 @@ class ArgParser
   end
 
   def display_args(data)
-    data.each do |def_set|
-      respond ''
-      respond "  ;#{Script.current.name} " + def_set.map { |x| format_item(x) }.join(' ')
-      def_set.each { |x| respond "   #{(x[:display] || x[:name]).ljust(12)} #{x[:description]} #{x[:options] ? '[' + x[:options].join(', ') + ']' : ''}"; }
+    # Display help output for script arguments
+    unless Script.current.name == "bootstrap"
```
rpherbig, line 53:
```suggestion
    if Script.current.name != "bootstrap"
```

dependency.lic
```diff
@@ -49,10 +49,26 @@ class ArgParser
...
+      data.each do |def_set|
+        respond ''
+        def_set.each { |x| respond " SCRIPT SUMMARY: #{x[:description]} " if x[:name].to_s == "script_summary"; }
```
rpherbig, line 56:
```suggestion
        def_set
        .select { |x| x[:name].to_s == "script_summary" }
        .each { |x| respond " SCRIPT SUMMARY: #{x[:description]} " }
```

dependency.lic
```diff
@@ -49,10 +49,26 @@ class ArgParser
...
+        respond ''
+        respond " SCRIPT CALL FORMAT AND ARG DESCRIPTIONS:"
+        respond "  ;#{Script.current.name} " + def_set.map { |x| format_item(x) }.join(' ')
+        def_set.each { |x| respond "   #{(x[:display] || x[:name]).ljust(12)} #{x[:description]} #{x[:options] ? '[' + x[:options].join(', ') + ']' : ''}" unless x[:name].to_s == "script_summary"; }
```
rpherbig, line 60:
```suggestion
        def_set
        .reject { |x| x[:name].to_s == "script_summary" }
        .each { |x| respond "   #{(x[:display] || x[:name]).ljust(12)} #{x[:description]} #{x[:options] ? '[' + x[:options].join(', ') + ']' : ''}" }
```

dependency.lic
asechrest: I like this better. Will change it.

dependency.lic
asechrest: Thanks for the suggestion. I still struggle through these and end up having to do quite a bit of live testing to get anything right. Let me digest this one.

asechrest:
I've changed this from draft to ready to review. Since @rpherbig approved it, I've just adding some code comments and slightly tweaked the heading text in the help output. No functional change since then.

rcuhljr:
My only comment may not be strictly code related, but base-yaml.yml related. I think each yaml field might benefit from a data type descriptor or example, this could be part of the description or an additional field for easy formatting. Something like this seems more useful

```
   YAML SETTINGS USED:
     circlecheck_prettyprint: Tells circlecheck to use the pretty print formatting for output. [true/false]
```

Something to tell me what I actually set it to. Debatable if it should also display the default value.

MahtraDR:
Does this need base-help for it to work? or is this harmless to merge, and base-help comes when it comes? Does it make sense to have both in the same PR?

asechrest:
> Does this need base-help for it to work? or is this harmless to merge, and base-help comes when it comes? Does it make sense to have both in the same PR?

Thanks for asking. It's harmless to merge and is fully backward compatible but I've got a couple additional local edits. Converted back to draft until I can push those.

dependency.lic
```diff
@@ -49,10 +49,51 @@ class ArgParser
...
+          .reject { |x| x[:name].to_s == "script_summary" }
+          .each { |x| respond "   #{(x[:display] || x[:name]).ljust(12)} #{x[:description]} #{x[:options] ? '[' + x[:options].join(', ') + ']' : ''}" }
+        respond ''
+      end
+
+      # Display help output for settings used in the script. Relies on base-help.yaml.
+      yaml_data = get_data('help').to_h
+      unless yaml_data.select { |field, info| info["referenced_by"].include?(Script.current.name)}.empty?
+        respond " YAML SETTINGS USED:"
+        yaml_data.select { |field, info| info["referenced_by"].include?(Script.current.name)}
```
KatoakDR, line 89: It looks like this is calling `yaml_data.select` twice, to mitigate the duplicated code I suggest assigning that to a variable before line 87 then referencing it.

Something like:

```ruby
      yaml_settings = yaml_data.select { |field, info| info["referenced_by"].include?(Script.current.name)}
      unless yaml_settings.empty?
        respond " YAML SETTINGS USED:"
        yaml_settings.each do |field, info|
            ...
        end
        ...
      end
```

dependency.lic
```diff
@@ -49,10 +49,51 @@ class ArgParser
...
+                  .each do | field, info |
```
KatoakDR, line 90: 🧹 Purely a stylistic comment, I think the majority of block arguments in our codebase don't include the space between the `|` and the variable name.

If that's the code convention, then this line perhaps should become `.each do |field, info|` ?

dependency.lic
```diff
@@ -49,10 +49,51 @@ class ArgParser
...
+    # with the wrong number or incorrect arguments, or when a script has been called via
+    # ;script-name help. It outputs an example script call, including arguments, with optional
+    # arguments enclosed in brackets. If an optional script_summary argument has been specified within
+    # the script, it also outputs a summary of the script. Further, if yaml settings have been
+    # outlined in base-help.yaml, it will output help information for the yaml settings used.
+    # Example output when all help settings have been specified:
+    #
+    # >;circlecheck help
+    #   SCRIPT SUMMARY: Checks your circle and skills to next circle. Optionally checks skills for a target circle.
+
```
KatoakDR, line 62: 🧹 nitpick, would you be ok to add `#` on the empty lines 62 and 68? I think it'd help show that all three comment blocks are part of the same example.

asechrest:
> My only comment may not be strictly code related, but base-yaml.yml related. I think each yaml field might benefit from a data type descriptor or example, this could be part of the description or an additional field for easy formatting. Something like this seems more useful
> 
> ```
>    YAML SETTINGS USED:
>      circlecheck_prettyprint: Tells circlecheck to use the pretty print formatting for output. [true/false]
> ```
> 
> Something to tell me what I actually set it to. Debatable if it should also display the default value.

I have now implemented this in base-help.yaml. Great suggestion. Thanks.

dependency.lic
asechrest: Good catch.

dependency.lic
asechrest: Done.

dependency.lic
asechrest: Another good cosmetic catch.

asechrest:
This is ready.

I'll separately roll in another commit of `base-help.yaml` in the next week or so (the changes here don't break if that file doesn't exist). I'll PR `base-help.yaml` in stages as I build it out with script settings information. A third step is adding a `script_summary` argument to all scripts, and in those scripts that don't currently do so, adding an parse_args call so that the settings help output can function on those.

rpherbig (approved):

PR #5721, opened 2022-04-06
[scripts] [alchemy] Remove includes, add prefixes.

MahtraDR:
As per title.

### DIFF
--- a/alchemy.lic
+++ b/alchemy.lic
@@ -5,11 +5,6 @@
 custom_require.call(%w[common common-crafting common-items common-travel common-arcana])
 
 class Alchemy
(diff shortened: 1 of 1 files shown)


alchemy.lic
```diff
@@ -5,11 +5,6 @@
 custom_require.call(%w[common common-crafting common-items common-travel common-arcana])
```
rpherbig, line 5: Can `common-arcana` be removed? I don't see where it is used

MahtraDR: Thank you!

rpherbig (approved):

PR #5722, opened 2022-04-06
[scripts] [autocontingency] Remove includes, add prefixes.

MahtraDR:
As per title.

### DIFF
--- a/autocontingency.lic
+++ b/autocontingency.lic
@@ -3,9 +3,6 @@
 =end
 custom_require.call(%w(common common-arcana events spellmonitor drinfomon common-travel))
 class Contingency
-  include DRC
-  include DRCA
-  include DRCT
 
   def initialize
     arg_definitions = [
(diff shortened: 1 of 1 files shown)


autocontingency.lic
```diff
@@ -3,9 +3,6 @@
 =end
 custom_require.call(%w(common common-arcana events spellmonitor drinfomon common-travel))
```
rpherbig, line 4: Can `common` be removed? I don't see where it is used...

rpherbig, line 4: My memory is fuzzy, did we make a change to `dependency` such that we no longer need to `custom_require(... drinfomon ...)` or is that still necessary?

MahtraDR: In theory, dependency always starts drinfomon as script 0.1 (dependency itself being script zero) in practice, someone can still kill drinfomon themselves. Calling it here means if for some weird reason drinfomon isn't already running like it should be, it gets called again. I think it's harmless at worst, and at best it'll start drinfomon in edge cases.

MahtraDR: I think we can remove `events` and `drinfomon` too since `spellmonitor` will depends on `drinfomon` anyway.

MahtraDR: I am sure we can remove drinfomon here, because  of spell monitor. 

Events looks like we only need if we have flags, which we don't here. I am not sure how that one works though.

rpherbig (approved):

PR #5728, opened 2022-04-10
[script] [bescort] Remove common includes, add prefixes

MahtraDR:
As per title.

Creating as draft, because I want to test hiding a bit more.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -421,7 +421,7 @@ class Bescort
       #Not there, but at the transport waiting room
       elsif Room.current.id == 6555 || Room.current.id == 6656
         echo 'waiting for transport'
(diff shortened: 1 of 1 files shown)


asechrest:
Having trouble reviewing because the diff is so large. Is there anything in here other than just adding prefixes? If so, what do you think about breaking it into two PRs?

MahtraDR:
> Having trouble reviewing because the diff is so large. Is there anything in here other than just adding prefixes? If so, what do you think about breaking it into two PRs?

Nothing but prefixes. It's just a big file.

MahtraDR:
Tested hiding. Looks good.

rpherbig (approved):

PR #5731, opened 2022-04-11
[script][validate.lic] - Fix hunting_file_list block

asechrest:
Via discussion in Discord, it was discovered that `;validate` was not functioning properly as a result of the dependency thread-safety rewrite. After diagnosis with @MahtraDR and @rcuhljr, we determined that the check in `;validate` to determine if a user has specified a `hunting_file_list` was checking against a nil value. The dependency rewrite now favors any specification in `base-empty.yaml`, and within that file `hunting_file_list` is specified as empty.

Because `;validate` was checking against a nil value, and the absence of a user specifying `hunting_file_list` is never nil, validate was never getting to the `else` block wherein a user's primary `<Username>-setup.yaml` would be validate.

This fix reorders the process a bit. It will now always validate the user's primary yaml file, and also validate a user-specified list of yaml files after that.

### DIFF
--- a/validate.lic
+++ b/validate.lic
@@ -17,10 +17,23 @@ class DRYamlValidator
 
     setup_data
 
-    if settings.hunting_file_list
-      respond("  Validating settings for #{settings.hunting_file_list.size} settings files")
+    respond("")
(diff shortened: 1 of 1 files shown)


MahtraDR (approved):
Tests well. Thanks!

PR #5732, opened 2022-04-12
[data] [base-sorting] Add weapon nouns, add negative lookbehind for "star"

MahtraDR:
Adding to weapons list
```
- morning star
- pasabas
- hhr'ata
- nightstick
```
Changed in jewlery list
```
- star
```
to
```
- (?<!morning) star
```

So it won't match a `morning star`

Tested before:
```
In the hip pouch:                                                                                                                                                                                                                            |
jewelry (1):                                                                                                                                                                                                                                 |
    morning star    
```
after
```
In the hip pouch:                                                                                                                                                                                                                            |
weapon (7):                                                                                                                                                                                                                                      
    morning star
 ```

### DIFF
--- a/data/base-sorting.yaml
+++ b/data/base-sorting.yaml
@@ -72,7 +72,7 @@ jewelry:
 - pendant
 - pin
 - ring
-#- (?<!morning\s)star
+- /(?<!morning\s)star/
 
 crafting:
 - (?:aldamdin|audrualm|ball-peen|Blacksmith's|cross-peen|damite|diagonal-peen|forging|straight-peen|Silversteel-faced|tomiek) (?:hammer|mallet)
--- a/data/base-sorting.yaml
+++ b/data/base-sorting.yaml
@@ -72,7 +72,7 @@ jewelry:
 - pendant
 - pin
 - ring
-- (?<!morning )star
+- (?<!morning) star
 
 crafting:
 - (?:aldamdin|audrualm|ball-peen|Blacksmith's|cross-peen|damite|diagonal-peen|forging|straight-peen|Silversteel-faced|tomiek) (?:hammer|mallet)
--- a/data/base-sorting.yaml
+++ b/data/base-sorting.yaml
@@ -72,7 +72,7 @@ jewelry:
 - pendant
 - pin
 - ring
-#- star
+- (?<!morning )star
 
 crafting:
 - (?:aldamdin|audrualm|ball-peen|Blacksmith's|cross-peen|damite|diagonal-peen|forging|straight-peen|Silversteel-faced|tomiek) (?:hammer|mallet)
--- a/data/base-sorting.yaml
+++ b/data/base-sorting.yaml
@@ -72,6 +72,7 @@ jewelry:
 - pendant
 - pin
 - ring
+- (?<!morning) star
 
 crafting:
 - (?:aldamdin|audrualm|ball-peen|Blacksmith's|cross-peen|damite|diagonal-peen|forging|straight-peen|Silversteel-faced|tomiek) (?:hammer|mallet)
--- a/data/base-sorting.yaml
+++ b/data/base-sorting.yaml
@@ -72,7 +72,7 @@ jewelry:
 - pendant
 - pin
 - ring
-- star
+#- star
 
 crafting:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: data/base-sorting.yaml [hunk 1]
```diff
@@ -72,7 +72,7 @@ jewelry:
 - pendant
 - pin
 - ring
-- star
+- (?<!morning) star
```
KatoakDR, line 75: Does the space between morning and star need to be in the right parenthesis to avoid matching " star" instead of "star"?

```suggestion
- (?<!morning )star
```

I haven't tested, just asking

MahtraDR: That was my first version too, and the regex match fails that way. I tried `(?<!morning\s)star` and `(?<!morning )star` And neither matched morning star. This way works for morning star, and I don't have a plain star to test. My original thought was to delete `star` altogether. :) Whenif I ever get a `star` I'll test it though.

asechrest: > Does the space between morning and star need to be in the right parenthesis to avoid matching " star" instead of "star"?
> 
> I haven't tested, just asking

Need moar Katoak!

MahtraDR:
Added a few more entries. All of these work for me as they are here.

data/base-sorting.yaml
rpherbig: I'm confused - is the goal to match `morning star` or not morning star?

MahtraDR: I want to match morning star. There is an entry for "morning star" under weapons. This works. 

There is an entry for plain "star" under jewelry. I want this entry to only match "star" without "morning " in front of it. The latter I can't seem to make happen.

rpherbig: `^(?!morning )star` seems to work for me:

* morning star = no match
* star = match

https://rubular.com/r/9D7IFASCbS8h7q

MahtraDR: Of course. I tell @rcuhljr this all the time... you know stuff I can't conceive of. 

I use the syntax I'm trying in profanity and it didn't occur to me there was another.

MahtraDR: Thanks for that link too. Looks super useful!

MahtraDR: So, this doesn't work either. 
Screenshot from 2022-04-13 10-44-11

```
vimes:~/lich/scripts/data$ egrep morning base-sorting.yaml 
- ^(?!morning )star
```

I'm inclined to just remove `star` :)

MahtraDR: So, clearly the parsing has some issue. I don't want to tackle that right now, but I do want my morning star to match correctly.

How about we compromise on changing `star` to `stars` which will accomplish what I want, and looking at `https://drservice.info/Plaza/` suggests would also catch most of the jewelry correctly.

MahtraDR: https://rubular.com/r/9D7IFASCbS8h7q says the lookahead I coded up front should work too, but it doesn't.

Any which way I've tried, if I have the entry under `jewelry` it seems to overwrite the `morning star` entry and the morning star gets categorised under jewelry.

MahtraDR: Still doesn't work with the prefix. This is driving me mad. We have other regex in there, and those seem to work just fine.

MahtraDR: the entry under weapons is a regex, note, and that works (If I comment out the entry under jewelry) `(morning|throwing) star` works. We have all kinds of positive lookaheads in here. Seems just the negative doesn't work?

MahtraDR: Oh my goodness, it works now! `/(?<!morning )star/` was the magic incantation.

MahtraDR: Bah, declared victory too early. Even with `/(?<!morning )star/` if I comment out the weapon line, my morning star registers as jewelry. :( I'm sad.

MahtraDR: Hah, disregard. I was a tad careless. With `/(?<!morning )star/` it does work!  It had the morning star under `other` which I just noticed when I double checked.

asechrest (approved):
I don't have a morning star, but the proposed changes look good after working out the regex.

PR #5736, opened 2022-04-12
[scripts] [appraisal] Adding count argument, and low-value detection

MahtraDR:
Allows for pouches to be checked against count, so pouches with 1-499 gems in the full pouch container can be put into a separate container for sorting.

Also allows for custom values set in yaml, below which, pouches can sorted out of the full pouch container.

I have hundreds of pouches, and only want to keep ones above a certain value, is my use case.

### DIFF
--- a/appraisal.lic
+++ b/appraisal.lic
@@ -13,7 +13,8 @@ class Appraisal
         { name: 'item', regex: /\w+/i, description: 'Item to use appraise focus with.' }
       ],
       [ 
-        { name: 'nonspecific', regex: /nonspecific/i, optional: true, description: 'Toggle off specific gem_pouch_adjective' }
+        { name: 'nonspecific', regex: /nonspecific/i, optional: true, description: 'Toggle off specific gem_pouch_adjective' },
+        { name: 'count', regex: /count/i, optional: true, description: 'Count gem pouches as they are appraised' }
       ]
     ]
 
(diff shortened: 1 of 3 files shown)


appraisal.lic
```diff
@@ -29,6 +30,16 @@ class Appraisal
       pouch_adjective = settings.gem_pouch_adjective
     end
 
+    if args.count
+      if settings.low_value_gem_pouch_container.empty? || settings.low_value_gem_pouch_container.nil?
```
rpherbig, line 34: This would also make a good validator

appraisal.lic
```diff
@@ -29,6 +30,16 @@ class Appraisal
...
+        DRC.message("\"count\" passed as an argument, but no low_value_gem_pouch_container defined. Exiting") 
+        exit
```
rpherbig, line 36: Is it better to exit if unable to count, or do as much as possible (i.e. appraise without counting)?

We could do this check on line 106-ish, output a message if we can't count, and then set `@count = false` (roughly). What do you think?

MahtraDR: Agreed. See current iteration. It's loud, but it works :) (It's loud in that it messages at every iteration that it can't count)

appraisal.lic
MahtraDR: See proposed validate.lic change?

rpherbig (approved):

PR #5737, opened 2022-04-14
[scripts] [common] def play_song? store instrument noun

MahtraDR:
Store instrument noun as a UserVar, and when we detect changes, reset song variables.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -594,7 +594,7 @@ module DRC
     if UserVars.instrument.nil?
       UserVars.instrument = instrument
     elsif UserVars.instrument != instrument
-      message("New instrument #{instrument} detected; old instrument: #{UserVars.instrument}. Resetting stored song data")
+      message("New instrument #{instrument} detected; old instrument: #{UserVars.instrument}. Resetting stored song data.")
       UserVars.song = nil
       UserVars.climbing_song = nil
       UserVars.instrument = instrument
--- a/common.lic
+++ b/common.lic
@@ -595,7 +595,7 @@ module DRC
       UserVars.instrument = instrument
     end
 
-    UserVars.instrument = instrument unless UserVars.instrument == instrument
+    UserVars.instrument = instrument if UserVars.instrument != instrument
  
     if UserVars.instrument != instrument
       message("New instrument #{instrument} detected; old instrument: #{UserVars.instrument}. Resetting stored song data")
--- a/common.lic
+++ b/common.lic
@@ -593,11 +593,7 @@ module DRC
 
     if UserVars.instrument.nil?
       UserVars.instrument = instrument
-    end
-
-    UserVars.instrument = instrument unless UserVars.instrument != instrument
- 
-    if UserVars.instrument != instrument
+    elsif UserVars.instrument != instrument
       message("New instrument #{instrument} detected; old instrument: #{UserVars.instrument}. Resetting stored song data")
       UserVars.song = nil
       UserVars.climbing_song = nil
--- a/common.lic
+++ b/common.lic
@@ -591,7 +591,7 @@ module DRC
   def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
     if UserVars.instrument != instrument
-      message("Instrument change detected. Resetting stored song data")
+      message("New instrument (#{instrument) detected; old instrument: #{UserVars.instrument}. Resetting stored song data")
       UserVars.song = nil
       UserVars.climbing_song = nil
     end
--- a/common.lic
+++ b/common.lic
@@ -595,7 +595,7 @@ module DRC
       UserVars.instrument = instrument
     end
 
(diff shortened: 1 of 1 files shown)


common.lic
```diff
@@ -590,6 +590,12 @@ module DRC
 
   def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
+    if UserVars.instrument != instrument
```
rpherbig, line 593: Won't this force a reset for every user the first time they play a song after this is merged? Because `UserVars.instrument` will be `nil`, thus seen as a new instrument?

How about
```suggestion
    UserVars.instrument = instrument unless UserVars.instrument
    if UserVars.instrument != instrument
        message("Instrument change detected. Resetting stored song data")
        UserVars.song = nil
        UserVars.climbing_song = nil
        UserVars.instrument = instrument
    end
```

common.lic
```diff
@@ -590,6 +590,12 @@ module DRC
...
+      message("Instrument change detected. Resetting stored song data")
+      UserVars.song = nil
+      UserVars.climbing_song = nil
+    end
+    UserVars.instrument = instrument unless UserVars.instrument == instrument
```
rpherbig, line 598: Won't this force a reset for every user the first time they play a song after this is merged? Because `UserVars.instrument` will be `nil`, thus seen as a new instrument?

How about
```suggestion
    UserVars.instrument = instrument unless UserVars.instrument
    if UserVars.instrument != instrument
        message("Instrument change detected. Resetting stored song data")
        UserVars.song = nil
        UserVars.climbing_song = nil
        UserVars.instrument = instrument
    end
```

rpherbig: Or...

```suggestion
    if !UserVars.instrument # maybe this is UserVars.instrument.nil?
        UserVars.instrument = instrument 
    elsif UserVars.instrument != instrument
        message("Instrument change detected. Resetting stored song data")
        UserVars.song = nil
        UserVars.climbing_song = nil
        UserVars.instrument = instrument
    end
```

common.lic
```diff
@@ -590,6 +590,12 @@ module DRC
 
   def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
+    if UserVars.instrument != instrument
+      message("Instrument change detected. Resetting stored song data")
```
rpherbig, line 594: What do you think about mentioning the old & new instruments?

```suggestion
      message("New instrument (#{instrument) detected; old instrument: #{UserVars.instrument}. Resetting stored song data")
```

MahtraDR: Great suggestion. Thank you!

common.lic
MahtraDR: In principle I like this, however, the reason I am making this change in the first place is that I changed my instruments, went from zills > cowbell, and my song wasn't working, and performance wasn't updating it. I had to manually delete the vars, and re-run performance. 

If we accept nil, someone in my situation would still have to manually delete the var, or change instruments again to reset the song. 

I do accept it might be a good middle-step though, to not force everyone to reset their song when this is pushed... So, all in all, it's a better way forward.

MahtraDR: How's the current version look?

common.lic
```diff
@@ -590,6 +590,19 @@ module DRC
 
   def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
+
+    if UserVars.instrument.nil?
+      UserVars.instrument = instrument
+    end
+
+    UserVars.instrument = instrument unless UserVars.instrument == instrument
```
rpherbig, line 598: I don't think this will work. Isn't this the equivalent to:

```suggestion
    UserVars.instrument = instrument if UserVars.instrument != instrument
```

Which means they'll always be set to equal if they are not already equal, i.e. I don't think the condition on line 600 can ever happen.

MahtraDR: Hmmmm Ok, back to the thinking cap!

common.lic
```diff
@@ -590,6 +590,19 @@ module DRC
 
   def play_song?(settings, song_list, worn = true, skip_clean = false, climbing = false)
     instrument = worn ? settings.worn_instrument : settings.instrument
+
+    if UserVars.instrument.nil?
+      UserVars.instrument = instrument
+    end
+
+    UserVars.instrument = instrument if UserVars.instrument != instrument
```
MahtraDR, line 598:
```suggestion
    UserVars.instrument = instrument unless UserVars.instrument != instrument
```

common.lic
MahtraDR: With the new block from 600, do we need this at all now?

MahtraDR: Check current version. it tests nice and clean. Tested by deleting the var, ran performance, the var was set, no fuss.

Tested by renaming the var, and running performance, picked up the change, set the var, picked song again.

rpherbig (approved):

PR #5740, opened 2022-04-14
[scripts] [appraisal] Fixing counting logic

MahtraDR:
Quick fix to stop it spamming on every loop.


MahtraDR:
@asechrest @rpherbig Merging my own PR as a one-off, since it's a breaking change.

PR #5746, opened 2022-04-18
[script] [go2] fix: `timeto` must be specified else ignored in pathfinding algorithm

KatoakDR:
### Background
* I'm trying to provide a custom route through some caves, but every time I tried to go to the destination room I received error that there was no path from where I was (Crossing) to where I wanted to go (some distant caves).
* I stumbled upon this issue because my overrides are creating a new pathway between two rooms that didn't previously exist so there was no preexisting `timeto` property.

```
[go2: error: failed to find a path between your current room (1900) and destination room (15760)]
```

```yaml
personal_wayto_overrides:
  cave_trolls_stream_enter:
    start_room: 19089
    end_room: 15759
    str_proc: start_script('bescort', ['cave_trolls', 'enter']); wait_while{running?('bescort')};
  cave_trolls_stream_exit:
    start_room: 15759
    end_room: 19089
    str_proc: start_script('bescort', ['cave_trolls', 'exit']); wait_while{running?('bescort')};
```

### Changes
* When applying custom pathing overrides, ensure `timeto` is populated so that the pathing is considered in routing algorithm

### DIFF
--- a/go2.lic
+++ b/go2.lic
@@ -183,7 +183,14 @@ if XMLData.game =~ /^DR/
 
   # Iterate through the aggregated map wayto overrides from above and set new stringprocs
   wayto_overrides.each do | key, values |
-    Map.list[values['start_room'].to_i].wayto["#{values['end_room']}"] = StringProc.new("#{values['str_proc']}")
+    start_room_id = values['start_room'].to_i
+    end_room_id = values['end_room'].to_i
+    travel_command = StringProc.new("#{values['str_proc']}")
+    travel_time = 0.2
+
+    start_room = Map.list[start_room_id]
+    start_room.wayto["#{end_room_id}"] = travel_command
+    start_room.timeto["#{end_room_id}"] = travel_time
   end
 end
 
--- a/go2.lic
+++ b/go2.lic
@@ -185,12 +185,28 @@ if XMLData.game =~ /^DR/
   wayto_overrides.each do | key, values |
     start_room_id = values['start_room'].to_i
     end_room_id = values['end_room'].to_i
-    travel_command = StringProc.new("#{values['str_proc']}")
-    travel_time = values['travel_time'].to_f || 0.2
 
     start_room = Map.list[start_room_id]
-    start_room.wayto["#{end_room_id}"] = travel_command
-    start_room.timeto["#{end_room_id}"] = travel_time
+
+    old_wayto = start_room.wayto["#{end_room_id}"]
+    old_timeto = start_room.timeto["#{end_room_id}"]
+
+    new_wayto = old_wayto
+    new_timeto = old_timeto
+
+    if values['str_proc']
+      new_wayto = StringProc.new("#{values['str_proc']}")
+    end
+
+    if values['travel_time']
+      new_timeto = values['travel_time'].to_f
+    end
+
+    start_room.wayto["#{end_room_id}"] = new_wayto
+    start_room.timeto["#{end_room_id}"] = new_timeto
+
+    echo "old_wayto=#{old_wayto}, new_wayto=#{new_wayto}"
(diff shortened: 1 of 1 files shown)


go2.lic
```diff
@@ -183,7 +183,14 @@ if XMLData.game =~ /^DR/
...
   wayto_overrides.each do | key, values |
-    Map.list[values['start_room'].to_i].wayto["#{values['end_room']}"] = StringProc.new("#{values['str_proc']}")
+    start_room_id = values['start_room'].to_i
+    end_room_id = values['end_room'].to_i
+    travel_command = StringProc.new("#{values['str_proc']}")
+    travel_time = 0.2
+
+    start_room = Map.list[start_room_id]
+    start_room.wayto["#{end_room_id}"] = travel_command
+    start_room.timeto["#{end_room_id}"] = travel_time
```
KatoakDR, line 193: The `wayto` defines _what_ command to perform to get from one room to the other.

The `timeto` defines _how long_ it takes to move between the rooms, and is required by the routing algorithm. If there's no `timeto` then the algorithm assumes there is no route, despite what `wayto` says.

MahtraDR: Are we better off adding timeto to our custom string procs, or hard coding .2?

asechrest:
This is interesting! I didn't do a whole lot of research into `timeto` for this fix, and I definitely didn't realize a `timeto` was necessary to have the pathing work at all. I know what the `timeto` is and I know that changing its value can effect our pathing I believe because this effects the weight of one node to another for the Dijkstra algorithm that the map/travel is built upon.

So, a couple of questions for the group:

1. Would it make sense to add the ability for a user to specify a custom `timeto` in the yaml? The only use case I can think of here is for a power user to be able to effect which paths are taken by customizing the timeto. BUT that might be a non-useful use case. Just not sure.
2. If so, would we want to set up a hierarchical default? If yaml specifies a `timeto`, use that. If none specified in yaml, use the base specified timeto. If no timeto exists anywhere, default to 0.2?

Let me know thoughts!

MahtraDR:
Thinking about this more, having custom timetos makes more sense to me. 

I can then blacklist routes if
I want to by setting the timeto very high for a place I don't want to go to.

KatoakDR:
I agree, custom `timeto` makes sense, I'll send up a change, thanks

KatoakDR:
This is ready for review again.

I tested in Crossing where I could move NE -> S, SE -> N, or just E from my current room. The original simplest path to go from my room to the room east of me was just to move E. I tinkered with the `str_proc` and `travel_time` options (none set, one set, both set) to have me take less ideal paths then the ideal path from my original location.

go2.lic
```diff
@@ -183,7 +183,27 @@ if XMLData.game =~ /^DR/
...
+    new_wayto = old_wayto
+    new_timeto = old_timeto
+
+    if values['str_proc']
+      new_wayto = StringProc.new("#{values['str_proc']}")
+    end
+
+    if values['travel_time']
+      new_timeto = values['travel_time'].to_f
+    end
```
KatoakDR, line 203: Conditionally override the `wayto` and `timeto`. This lets you make a path less desirably by increasing the `timeto` without necessarily needing to override the way in which you move between the two rooms, and vice versa. Or, you can override both!

```yaml
personal_wayto_overrides:
  test:
    start_room: 12245
    end_room: 12240
    travel_time: 0.1
    str_proc: move ne; move sw; move e
```
```yaml
personal_wayto_overrides:
  test:
    start_room: 12245
    end_room: 12240
    travel_time: 50
```
```yaml
personal_wayto_overrides:
  test:
    start_room: 12245
    end_room: 12240
    str_proc: move ne; move sw; move e
```

asechrest (approved):

MahtraDR (approved):

PR #5753, opened 2022-04-23
Update base-crafting.yaml

Raykyn55:
added parts found in bin in Hib to boar clan and hib part-room.


MahtraDR (approved):

asechrest:
Thank you for the first PR!

PR #5763, opened 2022-04-26
Update forge.lic

Raykyn55:
Attempts to adjust tongs before going for shovel. If it adjusts, works as shovel, and reverts.

Not sure if different adjustables have different messaging, if so, could add?


Raykyn55:
Forgot a line.

PR #5769, opened 2022-04-29
[script] [carve-lockpicks] Create carve-lockpicks.lic

Raykyn55:
Lockpick carving script for thieves.

### DIFF
--- a/carve-lockpicks.lic
+++ b/carve-lockpicks.lic
@@ -15,6 +15,7 @@ class CarveLockpicks
   def initialize
     @settings = get_settings
     @bag = @settings.crafting_container
+    @bag_items = settings.crafting_items_in_container
     @belt = @settings.engineering_belt
     @lockpick_carve_settings = @settings.lockpick_carve_settings
 	  @grandbag = @lockpick_carve_settings['grand_container']
@@ -24,7 +25,7 @@ class CarveLockpicks
 
     DRC.wait_for_script_to_complete('buff', ['carve'])
     DRC.bput('sit', 'You sit', 'You are already sitting', 'You rise', 'While swimming?') unless sitting?
-    DRCC.get_crafting_item("carving knife", @crafting_container, @crafting_container_items, @belt)
+    DRCC.get_crafting_item("carving knife", @bag, @bag_items, @belt)
     if DRCI.exists?("keyblank pocket")
       get_keyblank
     else
--- a/carve-lockpicks.lic
+++ b/carve-lockpicks.lic
@@ -55,7 +55,6 @@ class CarveLockpicks
   def carve_keyblank
     case DRC.bput('carve my keyblank with my knife', /proudly glance down at a grandmaster/, /begin to carefully carve/, /proudly glance down at a master/, /but feel your knife slip/)
     when /proudly glance down at a grandmaster/
-      initial_grand if @initial
 	    stow_item("grand")
     when /begin to carefully carve/
       carve_lockpick
@@ -83,12 +82,6 @@ class CarveLockpicks
     when /iron lockpick in your hands/
       stow_item("trash")
     end
-  end
-
-
-  def initial_grand
-    DRC.bput "carve my lockpick with my knife", 'With the precision and skill'
-    stow_item("grand")
   end
 
   def stow_item(name)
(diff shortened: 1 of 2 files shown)


carve-lockpicks.lic
```diff
@@ -0,0 +1,114 @@
+=begin
+	Need these settings in your yaml:
+		#Stow locations for Grandmaster's, Master's, and anything less.
+    grand_container: backpack
```
rpherbig, line 4: Please add any new settings to `base.yaml`

carve-lockpicks.lic
```diff
@@ -0,0 +1,114 @@
...
+
+  def initialize
+    @settings = get_settings
+    @bag = @settings.crafting_container
+    @belt = @settings.engineering_belt
+	  @grandbag = @settings.grand_container
+	  @masterbag = @settings.master_container
+	  @trashbag = @settings.trash_container
+	  @initial = @settings.initial
+    @tool = "carving knife"
```
rpherbig, line 23: Why use an instance var when this is a constant string?

carve-lockpicks.lic
```diff
@@ -0,0 +1,114 @@
+=begin
+	Need these settings in your yaml:
+		#Stow locations for Grandmaster's, Master's, and anything less.
+    grand_container: backpack
+    master_container: haversack
+    trash_container: cloak
+    #Whether to initial Grandmaster's with an extra carve.
+    initial: true
```
rpherbig, line 8: A lot of these setting names are kind of generic. A few ideas:

* Rename them to indicate they're for carving lockpicks
* Nest them under a parent setting which indicates they're for carving lockpicks
* Something else?

carve-lockpicks.lic
```diff
@@ -0,0 +1,114 @@
...
+    @bag = @settings.crafting_container
+    @belt = @settings.engineering_belt
+	  @grandbag = @settings.grand_container
+	  @masterbag = @settings.master_container
+	  @trashbag = @settings.trash_container
+	  @initial = @settings.initial
+    @tool = "carving knife"
+    DRC.wait_for_script_to_complete('buff', ['carve'])
+    DRC.bput('sit', 'You sit', 'You are already sitting', 'You rise', 'While swimming?') unless sitting?
+    DRCC.get_crafting_item("carving knife", @crafting_container, @crafting_container_items, @belt)
```
rpherbig, line 26: If you do keep `@tool`, should it be used here?

carve-lockpicks.lic
```diff
@@ -0,0 +1,114 @@
...
+      stow_item("master")
+    when /but feel your knife slip and the iron keyblank snaps like a twig/
+      get_keyblank
+    end
+  end
+
+  def carve_lockpick
+    case DRC.bput('carve my lockpick with my knife', /proudly glance down at a grandmaster/, /proudly glance down at a master/, /into the side of a/, /find a creature to carve/, /It would be better/, /Working with extreme caution/, /but feel your knife slip/)
+    when /proudly glance down at a grandmaster/
+      initial_grand if @initial
```
rpherbig, line 71: What do you think about moving the conditional into a guard clause in `initial_grand`?

carve-lockpicks.lic
```diff
@@ -0,0 +1,114 @@
...
+    when /proudly glance down at a master/
+      stow_item("master")
+    when /but feel your knife slip/
+      get_keyblank
+	  return
+    when /find a creature to carve/, /It would be better/
+      stow_item("trash")
+    when /Working with extreme caution/
+      carve_lockpick
+    when /iron lockpick in your hands./
```
rpherbig, line 82: Is the `.` needed in the regex?

carve-lockpicks.lic
Raykyn55: I think I understand what you mean. Updated, it's now under stow_item, where it carves again IF @initial_grand = true

carve-lockpicks.lic
Raykyn55: nested under lockpick_carve_settings

carve-lockpicks.lic
Raykyn55: removed.

carve-lockpicks.lic
Raykyn55: removed.

Raykyn55:
Thank you for the comments and suggestions, still have a lot to learn!

carve-lockpicks.lic
Raykyn55: Which removes that method, forgot that part. Second updated version is correct.

rpherbig: What I meant was a guard clause in the `initial_grand` function:

```
def initial_grand
  return unless @initial
  ...
end
```

But I'm fine with the current approach, as long as you remove line 71 (which calls the now removed function).

Raykyn55: It's being stubborn, but I think it's gone now.

rpherbig (approved):

Dartellum:
I thought someone made one. Checking.... I have a make-lockpicks.lic script already so it exists. Perhaps it on the repo (not GitHub)?

PR #5770, opened 2022-04-29
[scripts] [combat-trainer] Add special handling for cage of light

MahtraDR:
As per title. Leaving in draft because I have a couple of folks who say they want to test too.


combat-trainer.lic
```diff
@@ -1907,13 +1907,15 @@ class SpellProcess
     if data['moon']
       DRCMM.check_moonwatch
       moon = UserVars.moons['visible'].first
-      unless moon
+      if !moon && data['abbrev'].downcase == 'col'
+        data['cast'] = "cast ambient"
```
KatoakDR, line 1911: Seems helpful. I modified my yaml to enforce always `cast ambient` to get around this, too

```yaml
    Cage of Light: &cage_of_light
      name: Cage of Light
      abbrev: CoL
      mana: 15
      cambrinth: [10, 10]
      # Disable casting on first visible moon because it may be too weak
      # to learn any skill from. Instead, `cast ambient` which uses any available light.
      # Note, we have to explicitly set the moon property to override base-spells
      # so that common-arcana doesn't change the cast command to `cast {moon}`.
      moon: false
      cast: cast ambient
```

MahtraDR: Oh It needs to be fixed in common-moonmage too. I am just starting here. If this method works, I'll expand it there.

combat-trainer.lic
```diff
@@ -1907,13 +1907,15 @@ class SpellProcess
...
         echo "No moon available to cast #{data['abbrev']}" if $debug_mode_ct
 
-        @weather ||= bput('weather', 'inside', 'You glance up at the sky.', 'You glance outside')
-        if @weather =~ /inside/
-          message "*** You're inside and there are no available moons. You're going to have a hard time casting #{data['abbrev']}"
-        end
+        @weather ||= DRC.bput('weather', 'inside', 'You glance up at the sky.', 'You glance outside')
+          if @weather =~ /inside/
+            DRC.message "*** You're inside and there are no available moons. You're going to have a hard time casting #{data['abbrev']}"
+          end
```
asechrest, line 1918:
```suggestion
        @weather ||= DRC.bput('weather', 'inside', 'You glance up at the sky.', 'You glance outside')
         if @weather =~ /inside/
           DRC.message "*** You're inside and there are no available moons. You're going to have a hard time casting #{data['abbrev']}"
         end
```

combat-trainer.lic
MahtraDR: I've been chasing this down a bit, and waiting for GM confirmation... but it looks like there is zero drawback for `cast ambient` always. At which point I'd change scrap this PR, and suggest we instead change base-spells to disable moon lookups, and always `cast ambient` by default. Thoughts?

MahtraDR:
@KatoakDR Any objections to merging this now, and working on common-arcana separately?

PR #5772, opened 2022-04-30
[script] [common-crafting] Changes 'tap anvil' to 'look on anvil'

Raykyn55:
Changes tap anvil to look on anvil, unless to if, in order to stop attempting to use occupied anvils.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -55,7 +55,7 @@ module DRCC
   end
 
   def find_anvil(hometown)
-    return unless DRC.bput('tap anvil', 'I could not', 'You tap.*anvil') == 'I could not'
+    return if DRC.bput('look on anvil', 'I could not', 'ready for forging', 'On the iron anvil you see') == 'ready for forging'
     anvils = get_data('crafting')['blacksmithing'][hometown]['anvils']
     idle_room = get_data('crafting')['blacksmithing'][hometown]['idle-room']
     DRCT.find_sorted_empty_room(anvils, idle_room, proc { (DRRoom.pcs - DRRoom.group_members).empty? && clean_anvil? })
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -55,7 +55,7 @@ module DRCC
   end
 
   def find_anvil(hometown)
-    return if DRC.bput('tap anvil', 'I could not', /You tap.*anvil/) =~ /You tap.*anvil/ && clean_anvil?
+    return if DRC.bput('tap anvil', 'I could not', /You tap.*anvil/) =~ /You tap.*anvil/ && (DRRoom.pcs - DRRoom.group_members).empty? && clean_anvil?
     anvils = get_data('crafting')['blacksmithing'][hometown]['anvils']
     idle_room = get_data('crafting')['blacksmithing'][hometown]['idle-room']
     DRCT.find_sorted_empty_room(anvils, idle_room, proc { (DRRoom.pcs - DRRoom.group_members).empty? && clean_anvil? })
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -55,7 +55,7 @@ module DRCC
   end
 
   def find_anvil(hometown)
-    return if DRC.bput('look on anvil', 'I could not', 'ready for forging', 'On the iron anvil you see') == 'ready for forging'
+    return if DRC.bput('tap anvil', 'I could not', /You tap.*anvil/) =~ /You tap.*anvil/ && clean_anvil?
     anvils = get_data('crafting')['blacksmithing'][hometown]['anvils']
     idle_room = get_data('crafting')['blacksmithing'][hometown]['idle-room']
     DRCT.find_sorted_empty_room(anvils, idle_room, proc { (DRRoom.pcs - DRRoom.group_members).empty? && clean_anvil? })


MahtraDR:
We really want to return if looking at the anvil returns "ready for forging" ? It would make sense to me to return unless it's ready, no?

Raykyn55:
> We really want to return if looking at the anvil returns "ready for forging" ? It would make sense to me to return unless it's ready, no?

The method is lookig for an open anvil. If it finds one, it ends the method. The previous iteration returned on simply finding an anvil(unless <anvil not found>). This version returns If <anvil found> only if <anvil clear>. It can find an anvil, but if NOT clear, keeps looking.

rpherbig:
I see why you want to make this change, but I wonder if the better approach is simply to add the current room to the list of possible anvil rooms... the logic for "is this the room we want to stay in?" already exists on line 61. IIRC line 58 was added as a sort of optimization: if the current is ok, don't search. Putting the current room at the front of the search list would do the same thing, right?

Raykyn55:
> I see why you want to make this change, but I wonder if the better approach is simply to add the current room to the list of possible anvil rooms... the logic for "is this the room we want to stay in?" already exists on line 61. IIRC line 58 was added as a sort of optimization: if the current is ok, don't search. Putting the current room at the front of the search list would do the same thing, right?

I think the issue arises when the training person (yalda for example) is in the current room, so it doesn't look for an EMPTY room, being already IN one of those rooms? An alternative could be to move to an idle room? Any room without an anvil, so the search would check for empty. The problem here, is when the room is empty, but the anvil isnt. So if someone started a project, left it on the anvil, and departs, you have an empty room with an anvil, but crafting hangs because, while it found an anvil in an empty room, it didn't check if the anvil itself was empty.

rpherbig:
> I think the issue arises when the training person (yalda for example) is in the current room

I'm pretty sure that the logic ignores NPCs, only checking for PCs. Is that not what you're seeing?

> So if someone started a project, left it on the anvil, and departs, you have an empty room with an anvil, but crafting hangs because, while it found an anvil in an empty room, it didn't check if the anvil itself was empty.

I think that's what I was saying. The logic on 61 will attempt to clean the not-empty anvil and move on if cleaning fails. Why not use that same logic for the initial room check?

Raykyn55:
I have never seen it do anything but tap. In particular, never seen it actually clean an anvil(and I'm not sure that would be desirable behavior). Moreover, is there ever a time when we want smith/workorders to find an anvil and NOT check if it's clear?

Looking through my logs, it's always tap, never look, never clean.

Dartellum:
If the room is empty and the anvil is not clean, why not clean it off? The person left a project on the anvil. I have also seen, early on mind you, people putting stuff on the anvils to break Lich scripts.

Raykyn55:
Well, you might do, but a couple problems occurred to me. First, how do we know they haven't just stepped away to buy a hilt? Do we add a wait? or just scrub it instantly? Could have disconnected briefly, coming right back, etc.

Second, ingot's won't clean. have to move them manually.

Third, do we check for metal type? damite/glaes/etc?

I've run into this issue occasionally, most often in hib, and script never seems to clean. It just hangs, so if the script is supposed to do this, it's not doing it, at least that I can tell.
[log occupied anvil.txt](https://github.com/rpherbig/dr-scripts/files/8598088/log.occupied.anvil.txt)

Dartellum:
Typically you buy the hilt first but if not scripting the person should remove the item from the anvil if on the hilt part. Anvil not required for that. 

As for ingots, I think that is what people put on to mess up the scripts - brass ingots if I recall. Can update the clean method to remove the ingot and even check if it is the type you would keep. 

As for checking for metal type, meh. 

As for the disconnected part, that is a tough one. I really feel bad if the person did disconnect due to Internet but how long is the proper length of time to wait? The person may not come back on for hours or days depending on the situation.

Raykyn55:
So I took another look at this.  Without this change, we see tap. If tap finds an anvil, the method is done, and it's returned. Interestingly, if we DON'T find an anvil, it goes down the list of anvil rooms, and in each it checks first if the room is empty of players, THEN looks on the anvil. The initial return condition short-circuits what this method normally does. 

In fact, if you look at find_empty_crucible, the method cleverly incorporates tap with empty_crucible?. That, it seems to me, would be the solution that satisfied both of our concerns. Keeping it as IF, but mirroring crucible behavior.

Raykyn55:
So, before this change, if we get a workorder, or cue up ;smith, while already at an anvil, _find_anvil_ would tap, and if it found an anvil in the current room, it would start crafting (without checking anvil surface).

```
>l on anv
On the iron anvil you see a steel ingot.

...

[smith]>tap anvil
>
You tap an iron anvil.
>
--- Lich: forge active.

```

_Smith_ runs find_anvil, finds an anvil with tap, and cues _forge_ (which doesn't have any mechanism for cleaning/checking anvils, that's smith/CC territory). So, _forge_ fires up, reads book, studies recipe, then:


```
[forge]>get my steel ingot
You get a steel ingot from inside your rucksack.
>
[forge]>put my ingot on anvil
The anvil already has a steel ingot on it.
```

At which point Forge hangs, skips to the next step, getting tools, loops a couple times, then exits.

```
[forge]>put my book in my rucksack
You put your book in your rucksack.
>
[forge]>get my steel ingot
You get a steel ingot from inside your rucksack.
>
[forge]>put my ingot on anvil
The anvil already has an unfinished steel crown on it.
>
>
[forge: *** No match was found after 15 seconds, dumping info]
[forge: messages seen length: 6]
[forge: message: You feel fully prepared to cast your spell.]
[forge: message: Gauge Flow  (12 roisaen)]
[forge: message: Noumena  (21 roisaen)]
[forge: message: Gauge Flow  (12 roisaen)]
[forge: message: Noumena  (22 roisaen)]
[forge: message: The anvil already has an unfinished steel crown on it.]
[forge: checked against [/You put your/i]]
[forge: for command put my ingot on anvil]
[forge]>untie my blunt hammer from my forger's belt
You untie a glaes blunt forging hammer from the straps of your forger's belt.
>
[forge]>untie my tongs from my forger's belt
You fumble with the ties, quickly realizing that you need a free hand.
>
[forge: *** No match was found after 15 seconds, dumping info]
[forge: messages seen length: 4]
[forge: message: Gauge Flow  (11 roisaen)]
[forge: message: Noumena  (21 roisaen)]
[forge: message: [General] Your mind hears Leayne thinking, "the anatomy ones?"]
[forge: message: You fumble with the ties, quickly realizing that you need a free hand.]
[forge: checked against [/You (remove|untie)/i, /Untie what/i, /Your wounds hinder your ability to do that./i]]
[forge: for command untie my tongs from my forger's belt]
[forge]>untie my tongs from my forger's belt
You fumble with the ties, quickly realizing that you need a free hand.
>
You pause for a moment as the details of the Chaos symbiosis fade from your mind.
>
[forge: *** No match was found after 15 seconds, dumping info]
[forge: messages seen length: 6]
[forge: message: You pause for a moment as the details of the Chaos symbiosis fade from your mind.]
[forge: message: Gauge Flow  (11 roisaen)]
[forge: message: Noumena  (21 roisaen)]
[forge: message: Gauge Flow  (11 roisaen)]
[forge: message: Noumena  (21 roisaen)]
[forge: message: You fumble with the ties, quickly realizing that you need a free hand.]
[forge: checked against [/You (remove|untie)/i, /Untie what/i, /Your wounds hinder your ability to do that./i]]
[forge: for command untie my tongs from my forger's belt]
[forge]>get my tongs
You should untie the muracite tongs from the forger's belt first.

[forge: *** ERROR TRYING TO CRAFT, EXITING ***]
[forge]>tie my blunt hammer to my forger's belt
You attach a glaes blunt forging hammer to a strap on your forger's belt.
>
[forge]>tie my tongs to my forger's belt
You must be holding the muracite tongs to tie to the belt.
>
You feel fully rested.
>
--- Lich: forge has exited.
```

With this change, given the same parameters, the result is more along the lines of what the rest of the method does. This is a very small script i wrote to test:

```
custom_require.call(%w[common common-crafting])

class Test

    def initialize
        @settings = get_settings
        
        DRCC.find_anvil(@hometown)
        exit
    end
end

Test.new
```
First, an occupied anvil:
```
>;test

--- Lich: test active.

[test]>tap anvil

You tap an iron anvil.

[test]>look on anvil

On the iron anvil you see an unfinished steel crown.

[test]>clean anvil

You drag the waste bucket close and prepare to clean off the anvil.

[Do you really want to clean off the anvil?  Clean it one more time to destroy the contents.]

[test]>clean anvil

You stop working and clean off the anvil's surface, discarding an unfinished steel crown into the waste bucket.
Roundtime: 2 sec.

--- Lich: test has exited.
```
Unoccupied:
```
[test]>tap anvil

You tap an iron anvil.

[test]>look on anvil

The anvil's surface looks clean and ready for forging.

--- Lich: test has exited.
```

Finally, with an ingot:

```
>;test

--- Lich: test active.

[test]>tap anvil

You tap an iron anvil.

[test]>look on anvil

On the iron anvil you see a steel ingot.

[test]>clean anvil

You glance at the anvil's contents, and realize you should be able to remove them yourself.

[test]>get ingot from anvil

You get a steel ingot from atop an iron anvil.

[test]>put ingot in bucket

You drop a steel ingot in a large waste bucket.

--- Lich: test has exited.
```

Raykyn55:
The one concern i do have, though, is that IF we cue smith in a room with an anvil, but someone is already in the room, theres no check. Perhaps making this:

```
return if DRC.bput('tap anvil', 'I could not', /You tap.*anvil/) =~ /You tap.*anvil/ && (DRRoom.pcs - DRRoom.group_members).empty? && clean_anvil?
```
So that if you come into a room with an anvil, even if there IS an anvil, if theres someone there, it won't use that anvil. The player check occurs before the clean, so we don't destroy someone else's work.

asechrest:
This looks pretty good to me and you're right, it does mirror `find_empty_crucible` in _most_ ways except that the crucible methods don't appear to actually clean the crucible if it's not empty (if cleaning a crucible is even a thing). And thank you for the in-comment logs!

I've literally never used an anvil so I don't have too much skin in the game on this one. But it appears to improve script behavior.

PR #5775, opened 2022-04-30
[script] [combat-trainer] Barb Research - auto choose magic skill to train

KatoakDR:
# 🚧 On Hold -- Retesting Existing Script

### Background
* Javac [confirmed](https://discord.com/channels/619301383451181075/619330237418831917/970012705287180328) that Debilitation is not trainable via `meditate research` because it's a combat skill and you must be contested against an opponent.
* Since `meditate research` supports Augmentation, Warding, and Utility, I'd like the training ability to research the most appropriate skill that needs training based on learning rate/rank.

### Changes
* Remove support for `Barb Research Debilitation` (it doesn't work, see above)
* Add support for training ability `Barb Research` which will auto select the magic skill to meditate on.
* Maintain support for `Barb Research Augmentation`, `Barb Research Warding`, and `Barb Research Utility` for backwards compatibility and player-choice.

### Coniguration
_Only enable one of these variants_
```yaml
training_abilities:
  Barb Research: 300
  # Barb Research Warding: 300
  # Barb Research Augmentation: 300
  # Barb Research Utility: 300
```


combat-trainer.lic
```diff
@@ -2555,14 +2556,28 @@ class TrainerProcess
...
-      bput('meditate research turtle', 'You clear ')
-    when /^Barb Research Debilitation$/i
-      bput('meditate research earthquake', 'You clear ')
-    when /^Barb Research Utility$/i
-      bput('meditate research prediction', 'You clear ')
+    when /^Barb Research/i
+      # Note, the above regex is a 'starts with' match, don't append $ character to it
+      # so that we handle all scenarios of 'Barb Research' and 'Barb Research <magic skill>'.
+      #
+      # Debilitation cannot be learned via meditate research
```
rpherbig, line 2563: What do you think about adding a validator that Debilitation doesn't work now? I only suggest it because it is a change in behavior

combat-trainer.lic
```diff
@@ -2351,8 +2351,8 @@ class TrainerProcess
       'App Quick' => 'Appraisal',
       'App' => 'Appraisal',
       'Astro' => 'Astrology',
+      'Barb Research' => 'Magic', # Auto-select magic skill to train
       'Barb Research Augmentation' => 'Augmentation',
```
rpherbig, line 2355: Do we still need the old researches? I.e. does this new one encompass the rest?

KatoakDR: Good question. Although technically if the player had "Barb Research Augmentation: 200" in their yaml then this code change would handle it, but hat training ability would run regardless if the Augmentation skill needed to be trained then or not because the word "Magic" isn't a skill so it's learning rate will be 0 and that will always be less than the training abilities target rate.

If we keep "Barb Research <skill>" => "<skill>" then if a player were to be using those specific variants then combat-trainer would only attempt the research if that skill needed to be trained, otherwise it'd pick a different training ability to do.

Because of that nuance in choosing which training ability to perform, I chose to leave it in case someone really wanted that behavior.

rpherbig: Does that mean `Barb Research` (no specified skill) will always run, even if all skills are above the threshold?

KatoakDR: > Does that mean `Barb Research` (no specified skill) will always run, even if all skills are above the threshold?

Hrmm... yes, yes it would... 🤔 

**Unless** the `check_ability` method gets a bit smarter to let us specify one or more skills in that map

Thoughts on doing this differently?

combat-trainer.lic
KatoakDR: Done via https://github.com/rpherbig/dr-scripts/pull/5775/commits/ee8ed8a9d815a3c7a5f27a9b492db876ea57fbe5

combat-trainer.lic
KatoakDR: Actually... looking at `check_ability`... I think it _might_ support multiple skills already, stand by, will test!

MahtraDR:
I no longer have a barb to test this with. :(

combat-trainer.lic
KatoakDR: Thanks for the questions, I've updated `Barb Research` to specify the three magic skills it's relevant for, and `check_ability` will skip that if those three skills don't need training 👍 

Resolved in https://github.com/rpherbig/dr-scripts/pull/5775/commits/96171ece29d0b04bf4894dd061e076d305d19ca7

KatoakDR:
> I no longer have a barb to test this with. :(

For shame!

KatoakDR:
@rpherbig @MahtraDR I think I'm an idiot....... if I had only listed all three `Barb Research <skill>` in my yaml then that would have accomplished the exact same thing.

I've re-invented the wheel 🤦 

Ok, I'm going to go back and test with the following config, and if that works, I'll close this PR **unless you all think it makes sense to consolidate it down to a single Barb Research entry??**
```yaml
training_abilities:
  Barb Research Warding: 300
  Barb Research Augmentation: 300
  Barb Research Utility: 300
```

PR #5776, opened 2022-05-01
[script] [mine] Add adjustable tong support

Raykyn55:
adds the ability to specify tongs as mining implement, for those with adjustable tongs.

### DIFF
--- a/mine.lic
+++ b/mine.lic
@@ -116,6 +116,34 @@ class Mine
     else
       bput("get my #{@mining_implement}", 'You get', 'You are already')
     end
+    if @mining_implement = 'tongs'
+      case DRC.bput("adjust my tongs", 'With a yank you fold the shovel', 'You lock the tongs into', 'You cannot adjust', 'You have no idea how')
+      when 'You cannot adjust', 'You have no idea how'
+        DRC.message('Tongs are not adjustable, reverting. Please change yaml to reflect adjustable_tongs: false')
+        @adjustable_tongs = false
+        store_mining_tool
+        if exists?('shovel')
+          @mining_implement = 'shovel'
+          if @forging_belt['items'].grep(/shovel/i).any?
+            bput("untie my shovel", 'You remove', 'You untie')
+          else
+            bput("get my shovel", 'You get', 'You are already')
+          end
+        elsif exists?('pickaxe')
+          @mining_implement = 'pickaxe'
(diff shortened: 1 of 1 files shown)


mine.lic
```diff
@@ -116,6 +116,34 @@ class Mine
     else
       bput("get my #{@mining_implement}", 'You get', 'You are already')
     end
+    if @mining_implement = 'tongs'
+      case DRC.bput("adjust my tongs", 'With a yank you fold the shovel', 'You lock the tongs into', 'You cannot adjust', 'You have no idea how')
+      when 'You cannot adjust', 'You have no idea how'
+        DRC.message('Tongs are not adjustable, reverting. Please change yaml to reflect adjustable_tongs: false')
+        @adjustable_tongs = false
```
rpherbig, line 123: Why is `@adjustable_tongs` set to `false`? I don't see it used anywhere else

mine.lic
```diff
@@ -116,6 +116,34 @@ class Mine
     else
       bput("get my #{@mining_implement}", 'You get', 'You are already')
     end
+    if @mining_implement = 'tongs'
+      case DRC.bput("adjust my tongs", 'With a yank you fold the shovel', 'You lock the tongs into', 'You cannot adjust', 'You have no idea how')
+      when 'You cannot adjust', 'You have no idea how'
+        DRC.message('Tongs are not adjustable, reverting. Please change yaml to reflect adjustable_tongs: false')
```
rpherbig, line 122: Should this instead say `@mining_implement` should not be 'tongs'`?

mine.lic
```diff
@@ -116,6 +116,34 @@ class Mine
...
+          @mining_implement = 'pickaxe'
+          if @forging_belt['items'].grep(/pickaxe/i).any?
+            bput("untie my pickaxe", 'You remove', 'You untie')
+          else
+            bput("get my pickaxe", 'You get', 'You are already')
+          end
+        else
+          DRC.message('No usable mining tool present')
+          exit
+        end
```
rpherbig, line 142: I'm not a big fan of re-implementing `get_mining_tool` inside itself again, twice.

What about something like this?
```suggestion
        if exists?('shovel')
          @mining_implement = 'shovel'
          get_mining_tool
          return
        elsif exists?('pickaxe')
          @mining_implement = 'pickaxe'
          get_mining_tool
          return
        else
          DRC.message('No usable mining tool present')
          exit
        end
```

PR #5777, opened 2022-05-01
[script] [tinker] Add adjustable tong support

Raykyn55:
adds adjustable tong support for tinker.

### DIFF
--- a/tinker.lic
+++ b/tinker.lic
@@ -143,9 +143,27 @@ class Tinker
       echo "Didn't find a room with a gear press!"
       exit
     end
-    get_item('shovel')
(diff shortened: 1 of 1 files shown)


tinker.lic
```diff
@@ -143,9 +143,27 @@ class Tinker
       echo "Didn't find a room with a gear press!"
       exit
     end
-    get_item('shovel')
-    bput('push fuel with my shovel', 'furnace')
-    stow_item('shovel')
+    if @adjustable_tongs
```
rpherbig, line 146: Where is `@adjustable_tongs` defined?

PR #5778, opened 2022-05-01
[script] [combat-trainer] Barb Research and Berserk Training Abilities

KatoakDR:
### Background
* Supersedes https://github.com/rpherbig/dr-scripts/pull/5775
* Javac [confirmed](https://discord.com/channels/619301383451181075/619330237418831917/970012705287180328) that Debilitation is not trainable via `meditate research` because it's a combat skill and you must be contested against an opponent.
* The inner fire check to do Berserk Landslide or Berserk Avalanche are _after_ that ability has been chosen as what needs to be trained next. If your inner fire is below your threshold, then you can enter a state that starves out other eligible training abilities while you wait for your inner fire to recuperate.

### Changes
* Remove support for `Barb Research Debilitation` (it doesn't work, see above)
* Move inner fire check when doing berserk training to the `check_ability` method so a different ability will be picked if you don't have enough.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2351,10 +2351,9 @@ class TrainerProcess
       'App Quick' => 'Appraisal',
       'App' => 'Appraisal',
       'Astro' => 'Astrology',
-      'Barb Research Augmentation' => 'Augmentation',
-      'Barb Research Debilitation' => 'Debilitation',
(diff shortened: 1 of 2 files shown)


asechrest (approved):
I don't have a barb to test this, but the code looks good.

rpherbig (approved):

PR #5781, opened 2022-05-01
[scripts] [appraisal] Fix small regression with arguments

MahtraDR:
https://github.com/rpherbig/dr-scripts/pull/5780 broke appraisal without any args, which is a common usage method. This fixes that regression.


MahtraDR:
Self merging because it's a regression fix. @asechrest @rpherbig

PR #5784, opened 2022-05-02
[scripts] [workorders] Add support for additional material

Raykyn55:
Allows those with hobby/career or in higher ranks to select more difficult materials, eg silk, for workorders to speed up training. Fewer items, more mindstates, less time.  Not wedded to the names, open to suggestions!

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -898,6 +898,7 @@ workorders_materials:
   wood_type: maple
   stone_type: alabaster
   bone_type: wolf
+  metal_type: bronze
   knit_type: yarn
 
 #CARAVAN SETTINGS
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -893,6 +893,10 @@ workorder_cash_on_hand:
 retain_crafting_materials: false
 # Whether you want to repair your crafting tools after completing workorders or not. For those with tools setup under gear, set to false in your yaml.
 workorders_repair: true
+workorders_materials:
+  fabric_type: burlap
+  wood_type: maple
+  stone_type: alabaster
 
 #CARAVAN SETTINGS
 # container holding caravan contracts for ;trade
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -897,6 +897,7 @@ workorders_materials:
   fabric_type: burlap
   wood_type: maple
   stone_type: alabaster
+  bone_type: wolf
   knit_type: yarn
 
 #CARAVAN SETTINGS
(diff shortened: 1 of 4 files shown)


rpherbig (commented):
I'm not a big fan of a big switch/case statement that overrides the content of `base-crafting.yaml`. I'd rather have the stock number/volume/etc directly in there.

What do you think about that?

A few approaches I see:
* Make `sew-stock-number`/etc. a hash (linen -> 7, etc.)
* Make several entries: `burlap-stock-number`, `linen-stock-number`, etc.
* Something else?

profiles/base.yaml
```diff
@@ -893,6 +893,10 @@ workorder_cash_on_hand:
 retain_crafting_materials: false
 # Whether you want to repair your crafting tools after completing workorders or not. For those with tools setup under gear, set to false in your yaml.
 workorders_repair: true
+workorders_materials:
```
rpherbig, line 896: This setting is named to be specific to `workorders`, but the overridden things (`sew-stock-number`, etc.) are also used elsewhere (e.g. `trade`). Would other scripts want this functionality?

Raykyn55: Trade uses workorders, so this change gives that functionality to Trade. I think Craft would have to be rewritten with new ranges in order to use this materials adjustment. I don't use Craft, but if someone who does wants this, I'm all ears on a tactic for that.

Workorders is kind of a special case for this kind of change, in that it has a set number of items it's going to do. This makes changing the materials a niche adjustment. Craft, for example, can just do 2.5 items and be efficient (this is why some prefer it, that don't care about prestige).

Raykyn55:
I have an idea for making the change inside base-crafting. I'll submit later today.

Raykyn55:
In summary, this adds another block in base-crafting called "stock". All of the volume and order numbers are persistent, so having them individually listed under each region is unnecessary, imo. As far as i can tell, workorders is the only script that uses this information, so i went ahead and removed it under each hometown. 

Info now contains home-town specific information, same as before, but information regarding specific materials is collected under materials_info. This normalizes stock names/numbers (instead of sew-stock, bone-stock, etc.)

I also added, though didn't implement, stock value. This might be useful down the road for common money if we wanted to dial in more exact figures for cash on hand. it's in Kronars, so no matter the denomination, you'll always have enough. 

I cleaned up the carving sections some, removing if/else for bone/stone, and moved that to the initial setup.

Raykyn55:
Resubmitted here: https://github.com/rpherbig/dr-scripts/pull/5894

PR #5787, opened 2022-05-03
[script] [workorders] Remove bad prefixes...

MahtraDR:
We were a tad aggressive in https://github.com/rpherbig/dr-scripts/pull/5783 

It's frustrating when scripts have local defs that have the same name as common-defs...

### DIFF
--- a/workorders.lic
+++ b/workorders.lic
@@ -198,7 +198,7 @@ class WorkOrders
     end
   end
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
@rpherbig @asechrest self-merging to fix a regression.

MahtraDR:
@Raykyn55 just FYI. It's fixed now.

PR #5791, opened 2022-05-05
[scripts] [hunting-buddy] Extra YAMLs and configurable prehunt_buff waggle

bishoprook:
Allows additional YAMLs to be specified as flex args and honored. Also introduces a new `prehunt_buff_waggle` setting that defaults to `prehunt_buffs` so different hunts can use different waggle sets.

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -10,7 +10,7 @@ class HuntingBuddy
     arg_definitions = [[]]
     args = parse_args(arg_definitions, true)
 
-    @settings = get_settings
+    @settings = get_settings(args.flex)
 
     @town_data = get_data('town')[@settings.hometown]
     @hunting_data = get_data('hunting')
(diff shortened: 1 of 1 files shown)


MahtraDR:
I quite like the `args.flex` to me that fixes what is effectively a bug.

I don't really see a purpose for yet more waggles? As is, I do this:

```yaml
  default: &default
    <<: *stw     # Dual load, +range
    <<: *sott    # +Perc, +Ref
    <<: *oath    # +Melee
    <<: *bloodthorns
  prehunt_buffs:
    <<: *cs
    <<: *default
```
On a different character it's even simpler
```yaml
  default: &default
    <<: *mef
    <<: *regen
    <<: *gol
    <<: *vig
    <<: *heal
  prehunt_buffs: *default
```

MahtraDR:
Oh, but if you make it a setting, you can have a different prehunt buff waggle, without having to define all your waggles again eh?

KatoakDR:
> Oh, but if you make it a setting, you can have a different prehunt buff waggle in a different hunt file, without having to define all your waggles again eh?

That's my understanding

KatoakDR (commented):
👍 Code looks good to me

PR #5792, opened 2022-05-05
[script] [combat-trainer] Don't face/engage when using Innocence

bishoprook:
Adds a new `is_engaging_allowed?` gamestate which is honored by preventing any action that would require a `face` or `engage`. This prevents empaths from breaking Innocence (e.g. when manipulating and/or using GS)


combat-trainer.lic
```diff
@@ -2507,7 +2508,7 @@ class TrainerProcess
       waitrt?
       fix_standing
     when /^Tactics$/i
-      bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty?
+      bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty? || !game_state.is_engaging_allowed?
```
rpherbig, line 2511: I'm not super familiar with Innocence (I just read its entry on elanthipedia), so maybe this is a dumb question. Why would someone using Innocence also be training Tactics?

combat-trainer.lic
```diff
@@ -2230,6 +2230,7 @@ class AbilityProcess
     timer = game_state.cooldown_timers['Battle Cry']
     return unless !timer || (Time.now - timer).to_i > @battle_cry_cooldown
     return unless game_state.npcs.length > 0
+    return unless game_state.is_engaging_allowed?
```
rpherbig, line 2233: Battle cries are Bardic. Would a Bard ever be using Innocence?

combat-trainer.lic
```diff
@@ -2762,6 +2763,7 @@ class TrainerProcess
     return if game_state.npcs.empty?
     return if fail_count > @analyze_retry_count
     return if DRSkill.getxp('Tactics') >= @combat_training_abilities_target
+    return unless game_state.is_engaging_allowed?
```
rpherbig, line 2766: I'm not super familiar with Innocence (I just read its entry on elanthipedia), so maybe this is a dumb question. Why would someone using Innocence also be using Analyze?

combat-trainer.lic
```diff
@@ -3407,7 +3411,7 @@ class AttackProcess
   end
 
   def dance(game_state)
-    if game_state.npcs.empty?
+    if game_state.npcs.empty? || !game_state.is_engaging_allowed?
```
rpherbig, line 3414: Do dance actions break Innocence? If not, should this check be moved to line 3420?

combat-trainer.lic
```diff
@@ -2567,7 +2568,7 @@ class TrainerProcess
     when /^Collect$/i
       game_state.sheath_whirlwind_offhand
       retreat
-      put('engage') unless game_state.npcs.empty? || game_state.retreating?
+      put('engage') unless game_state.npcs.empty? || game_state.retreating? || !game_state.is_engaging_allowed?
```
rpherbig, line 2571: What do you think about making a function on game_state (`engage`?) that encapsulates this repeated set of conditionals? See lines 2621, 2812, , 3358, etc.

combat-trainer.lic
bishoprook: It's fair, there are a couple "should never happen" examples here -- see also training Astrology, since Innocence is an empath signature spell. However, it also felt like `is_engaging_allowed?` could potentially be broader than whether or not you have Innocence cast, so I tried to cover pretty much any case that would lead to a `face` or `engage` command being issued. If that doesn't make sense I could also see this being restricted to only cases that would actually happen for an empath using the spell.

combat-trainer.lic
bishoprook: Yes, the `engage` on line 3416 will break Innocence. Any `face` or `engage` will. (But for whatever reason, `support` won't, and implicitly facing the next target with a `target` `bob` etc won't...)

But dance commands like `weave` and `circle` require you to be at pole range or closer, which is incompatible with Innocence which keeps enemies entirely disengaged if it works.

combat-trainer.lic
bishoprook: Yeah, I thought about that. Also considered making `is_engaging_allowed?` defined as "not cast Innocence and not retreating"

There are some cases where we'll completely skip some actions when retreating that we should be able to do while using Innocence, so they're not quite the same thing, but there is a lot of repeat/overlap.

Maybe it would make sense to also include `game_state.npcs.empty?` in the check too. In that case, it doesn't _hurt_ anything to issue a `face` or `engage` command but it's still pointless.

bishoprook: This could also address a number of the other concerns. "Why would an empath..." "Why would someone using Innocence..."

Generalizes more to "prevent us from issuing a `face` or `engage` command when it would be either bad or wrong"

MahtraDR:
How about treating Innocence like we treat Absolution? Currently the `undead` argument triggers checks for Absolution, and prevents attacks without the spell cast. It seems consistent to keep that workflow? The it's a more proactive action on the part of the user, setting the argument in their `hunting_info` section like we would `construct` or `undead` and not just "Oh, is Innocence up?"

bishoprook:
> The it's a more deliberate action on the part of the user, setting the argument in their `hunting_info` section like we would `construct` or `undead` and not just "Oh, is Innocence up?"

That makes a lot of sense. When I'm defining the hunt I should know that I plan to be using Innocence and I don't want to be dancing to break it.

combat-trainer.lic
```diff
@@ -2507,7 +2508,7 @@ class TrainerProcess
       waitrt?
       fix_standing
     when /^Tactics$/i
-      bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty?
+      bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') if game_state.can_engage?
```
bishoprook, line 2511: Note: as written this now prevents this action when `retreating?` which would be a new change. I'm not 100% sure what the `retreating?` flag is about, so maybe this is incorrect. But if we're retreating we want to stay at ranged, and we can't do tactics actions at ranged, right?

combat-trainer.lic
```diff
@@ -2759,7 +2760,7 @@ class TrainerProcess
   end
 
   def analyze(game_state, fail_count)
-    return if game_state.npcs.empty?
+    return unless game_state.can_engage?
```
bishoprook, line 2763: Likewise this now covers `retreating?` where it didn't before, but we have to be at pole range or closer to analyze.

combat-trainer.lic
```diff
@@ -3981,6 +3989,14 @@ class GameState
     !@clean_up_step.nil?
   end
 
+  def can_engage?
+    return false if @innocence_mode
+    return false if retreating?
+    return false if npcs.empty?
+    return false if danger
```
bishoprook, line 3996: Not certain it makes sense for this `danger` check to be here too. Will we always be `retreating?` if we're in danger?

combat-trainer.lic
bishoprook: Seems correct to restrict this when we're `retreating?` since this CAN engage

bishoprook:
Any other thoughts here?

MahtraDR:
> Any other thoughts here?

If you can clean/resolve any pending comments, and if it's in a place to be re-reviewed, tag us for review?

combat-trainer.lic
```diff
@@ -3407,7 +3409,7 @@ class AttackProcess
   end
 
   def dance(game_state)
-    if game_state.npcs.empty?
+    unless game_state.can_engage?
```
rpherbig, line 3412: Minor, but I'd prefer to swap the two branches so we can use `if` instead of `unless` here

combat-trainer.lic
```diff
@@ -3981,6 +3988,19 @@ class GameState
     !@clean_up_step.nil?
   end
 
+  def can_engage?
+    return false if @innocence_mode
+    return false if retreating?
+    return false if npcs.empty?
```
rpherbig, line 3994:
```suggestion
    return false unless can_face?
    return false if retreating?
```

Thoughts?

rpherbig (approved):
Only minor comments

combat-trainer.lic
MahtraDR: Sorry following up on this one. What do you mean? I'll fix the code and get this merged as there's interest from a couple others.

rpherbig:
```
if game_state.can_engage?
    line 3415...
else
    pause 1
end
```

combat-trainer.lic
MahtraDR: Misconfiguration? It should be safe to return here regardless, as the two are mutually exclusive.

combat-trainer.lic
MahtraDR: @rpherbig Is it good now?

combat-trainer.lic
```diff
@@ -3407,17 +3409,15 @@ class AttackProcess
...
-      pause 1
-    else
+    if game_state.can_engage?
       game_state.set_dance_queue
-      case bput(game_state.next_dance_action, 'You must be closer', 'There is nothing else', 'What are you trying', /.*/)
+      case DRC.bput(game_state.next_dance_action, 'You must be closer', 'There is nothing else', 'What are you trying', /.*/)
       when 'You must be closer', 'There is nothing else', 'What are you trying'
-        fput('engage')
+        DRC.bput('engage', 'You begin', 'You are already')
         pause 2
```
MahtraDR, line 3417:
```suggestion
```

MahtraDR:
I'll smoke test this today and if no errors, will merge.

Kiriawen:
@MahtraDR Is this one still brewing or has it gone flat?

MahtraDR:
> @MahtraDR Is this one still brewing or has it gone flat?

I plan to merge this imminently, now that HE is over, I can actually test it. :D

combat-trainer.lic
```diff
@@ -2896,7 +2896,7 @@ class TrainerProcess
       waitrt?
       DRC.fix_standing
     when /^Tactics$/i
-      DRC.bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') unless game_state.npcs.empty? || game_state.retreating?
+      DRC.bput($tactics_actions.sample, 'roundtime', 'There is nothing else', 'Face what', 'You must be closer', 'You must be standing', 'Strangely, you don\'t feel like fighting right now', 'flying too high for you to attack') if game_state.can_face?
```
mdr55, line 2899: shouldn't this be can_engage?  The original logic checks for retreat? which means we should be using retreat?'s replacement, can_engage?  otherwise this can pull someone back in.

combat-trainer.lic
```diff
@@ -3880,18 +3882,14 @@ class AttackProcess
   end
 
   def dance(game_state)
-    return if game_state.retreating?
-
-    if game_state.npcs.empty? || game_state.retreating?
-      pause 1
-    else
+    if game_state.can_engage?
```
mdr55, line 3885: original logic gets out of dance immediately when retreating.  The new logic's !engage results in pause.  Is that desireable?

combat-trainer.lic
```diff
@@ -3220,7 +3220,7 @@ class TrainerProcess
     if DRSkill.getrank('Astrology') <= 120
       DRCMM.predict('weather')
       waitrt?
-      game_state.engage
+      game_state.engage if game_state.can_engage?
```
mdr55, line 3223: mahtra - why not lock the check inside game_state.engage? for futureproofing / belts and braces
Any side-effects?  game_state.engage is conditionally called at the start of AttackProcess' execute.

MahtraDR:
Closing. Implemented in #7103.

PR #5797, opened 2022-05-11
[data] [base-spells] Add new Empath spell Calculated Rage

MahtraDR:
Tested on my empath with
```yaml
  cr: &cr
    Calculated Rage:
      use_automana: true
```

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -922,6 +922,13 @@ spell_data:
     recast: 1
     mana_type: arcane
     prep: prepare
+  Calculated Rage:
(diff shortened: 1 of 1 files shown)


Kiriawen:
now we just need to add in some handling for sending RAGE.  Would be nice if we could add it for IZ too

MahtraDR:
> now we just need to add in some handling for sending RAGE. Would be nice if we could add it for IZ too

Something like a smart `training_abilities` that checks if the spell is up, and rages/touches on a timer? Chat on discord about this one.

asechrest (approved):

PR #5806, opened 2022-05-27
Island Crafting Update

Hiinky:
Finally mapped the new societies.  So sorry if I accidentally send someone back to crossing.


MahtraDR (approved):
Yaml is valid, data looks sensical.

PR #5807, opened 2022-05-28
Update base-spells.yaml

Raykyn55:
Adds base settings for Iridius Rod, new trader spell, to base-spells. Put it directly under TRC.


MahtraDR:
This spell is going to trip people up. It requires a bright moon, or starlight aura. 

Do we just merge as is? Other trader spells are like this I suppose, and no mention is made.

MahtraDR (approved):
Works as is. Merging now.

Issue #5814, opened 2022-05-31
hunting-buddy fails to exit combat-trainer when stop_on_low fires at beginning of hunt


Tirost:
A situation occurs where my tarantula eats a skill listed under stop_on_low just as hunting-buddy/combat-trainer are beginning a hunt. The `***STATUS*** Stopping because skills dropped below learning threshold` messaging triggers, but the hunt doesn't stop. The hunt just continuously progresses without ever checking either time, stop_on or stop_on_low limits.

Log below:

https://pastebin.com/HxsL0uuk

MahtraDR:
Could you please post your `hunting_info` settings for this hunt as well? Thanks!

Tirost:
`hunting_info:
- :zone: black_apes
  :duration: 40
  stop_on:
  - First Aid
  - Staves
  stop_on_low:
  - Shield Usage
  - Evasion`
  
Thanks for looking into it, Mahtra.

asechrest:
The following hunch was discussed in Discord, with Briarleigh and @vtcifer contributing:

In this instance, `hunting-buddy` immediately notices a `stop_on_low` threshold and calls the `$COMBAT_TRAINER.stop` method to set instance variable `@stop = true`. However, this may be _before_ combat trainer's initial setup, which sets `@stop = false` in preparation for the rest of the combat trainer run.

The current best idea, also courtesy of @vtcifer , is to build in logic in combat-trainer's `stop` method to wait until the script is initialized and `@running` before setting the stop variable so that it's not overwritten.

MahtraDR:
After a bit of digging, this is the same problem as described here https://github.com/rpherbig/dr-scripts/pull/4849#issuecomment-811440402

PR #5816, opened 2022-06-01
[script][combat-trainer] - Better stop process

asechrest:
Addresses #5814 via @Tirost , wherein `combat-trainer` will sometimes fail to stop when `hunting-buddy` sends a remote stop request extremely early in the process.

This attempts to wait for combat-trainer to init and setup before setting the stop boolean, since the setup process sets the boolean to false and may overwrite.

Credit to @vtcifer , this is an exact copy/paste of his proposed solution.

Settings as draft for the moment and I'll grab a couple testers if I can.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3553,10 +3553,14 @@ class CombatTrainer
 
   def stop
     echo 'Received stop signal' if $debug_mode_ct
+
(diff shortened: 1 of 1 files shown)


asechrest:
@Tirost confirms this fix didn't work. Still doesn't recognize the stop command and combat-trainer runs indefinitely.

I'll need to dig in a little more with some testing of my own.

PR #5819, opened 2022-06-04
[script][combat-trainer] - Simple roar helm functionality

asechrest:
Adds simple roar helm functionality. By request in Discord.

Will require 1 new yaml setting. I'll submit separately after some testing of this PR.

Draft for now while I get some testers.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2087,9 +2087,6 @@ class AbilityProcess
     @battle_cry_cooldown = settings.battle_cry_cooldown
     echo("  @battle_cry_cooldown: #{@battle_cry_cooldown}") if $debug_mode_ct
 
-    @roar_helm = settings.roar_helm
-    echo("  @roar_helm: #{@roar_helm}") if $debug_mode_ct
-
     @roar_helm_noun = settings.roar_helm_noun
     echo("  @roar_helm_noun: #{@roar_helm_noun}") if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


asechrest:
Note for myself: probably need to be sure noun is specified if `roar_helm` is true. In `;validate` or in CT code? Probably the latter.

combat-trainer.lic
```diff
@@ -2248,8 +2254,16 @@ class AbilityProcess
 
     if battle_cry_data
       echo "Selected battle cry: #{battle_cry_data}" if $debug_mode_ct
+      echo "Using roar helm: #{@roar_helm}, with roar helm noun: #{@roar_helm_noun}" if $debug_mode_ct
+
       command = battle_cry_data['command']
       command = (command + " at " + battle_cry_data['target_enemy']) if battle_cry_data['target_enemy']
+
+      # Use roar helm if specified
+      fput("scream #{@roar_helm_noun}") if @roar_helm
```
KatoakDR, line 2263: Can we avoid two new settings and have this "if" check based on `roar_helm` nil or not?

asechrest: I believe we need to know the noun of the helm in case `helm` isn't it. We could use one setting, perhaps, called `roar_helm_noun`, and just check if that's not empty?

asechrest: Or are you thinking to grab the noun from our currently worn helm somehow, maybe via EquipManager?

asechrest: PS - There is also a more complex version of implementing roar helms that involves something like your idea: adding an `:is_roar: true` element to gear list, then making sure we're wearing that helm before we try the `scream helm`, which would prevent failed screams when cycling armor and the helm isn't being worn.

I don't 100% know how to do that and would need to do a bit of study. Thoughts?

asechrest:
Some chat in Discord suggest most barbs are not cycling head armor when using a roar helm because there is a built-in 10 second RT to remove a roar helm of any type. Elanthipedia corroborates the RT aspect.

So I'm inclined not to get more complex with gear cycling and a new Class Item accessor for `:is_roar:` and recognizing when we remove and then don a roar helm, etc., if nearly all barbs are not cycling head armor when intending to use a roar helm.

KatoakDR:
@asechrest Looks good to me. I agree with [Robert](https://github.com/rpherbig/dr-scripts/pull/5819#discussion_r891462353) that this PR just needs `roar_helm_noun` added to `base.yaml` then good to go.

asechrest:
#5819 adds `roar_helm_noun` to base.yaml.

I've also separately had at least 2 barbs testing an iteration of this change in Discord and they say it works fine. They also say no real use for the complexity of considering armor cycling. Worst case scenario is a zero-RT attempt to `scream helm`, and barbs who want to use this functionality are unlikely to be cycling helms anyway.

Changing this from draft to ready. Though I think we need to wait for the base.yaml merge?

PR #5435, opened 2022-01-17
[script] [common] major refactor of identify noun regex

KatoakDR:
# 🚧 Work in progress 🚧 

### Background
* Increasingly, I'm needing to detect the "noun" of an item as it's parsed from LOOK, RUMMAGE, game output, inventory list, etc and maintaining the current regex is not scalable
* In collaboration with @asechrest and @MahtraDR, we have been working on new heuristics and modularizing the regex to reliably identify the noun
* It's a daunting task, but one I enjoy working on, so here we are!

### Changes
* Rework the regex in `remove_flavor_text` method to more reliably identify nouns of items
* Evolution of the proposed regular expression and our tests are at https://regex101.com/r/4lGY6u/latest
  - The items in the "Test String" field in the bottom half of the site are items that don't parse yet
  - In the left side-bar, there's a "Unit Tests" tab (can take 10-20 minutes for them to complete) that shows current progress towards 100% compatibility
* We have over 1,700 tests on regex101, and we continually add more
* Data is sourced from our personal vaults, trader tables, and elanthipedia

PR #5824, opened 2022-06-07
[script][common-items.lic] - Update to support new Steelclaw Clan bin

MazMedias:


common-items.lic
```diff
@@ -356,6 +356,7 @@ module DRCI
 
     trashcans = DRRoom.room_objs
                       .reject { |obj| obj =~ /azure \w+ tree/ }
+                      .map { |obj| obj.gsub('small bin concealed with some nearby brush', 'small bin') }
```
rpherbig, line 359: Does `DRC.get_noun` not work on "small bin concealed with some nearby brush"?

MazMedias: It returns `concealed` so I implemented a workaround suggestion form @asechrest

asechrest: Confirming it doesn't pare it down to the correct noun. (Interestingly, @KatoakDR 's noun regex refactor works perfectly on it, but I don't know that it's ready for release).

We threw this around in Discord quite a bit today. I suggested this solution as a way to just target this specific bin and use `gsub` to get it into a form where the noun regex would work on it, and then the rest of it flows fine.

I welcome better approaches, though.

Also appreciate @MazMedias submitting the PR!

KatoakDR: I think I have a fix for the current noun parsing regex at https://github.com/rpherbig/dr-scripts/pull/5827

Interim fix until I can spend more time to get https://github.com/rpherbig/dr-scripts/pull/5435 ready for primetime

rpherbig:
Overcome by #5827

MazMedias:
It is working fine as far as I can tell. I believe the change to 372 still needs to be added to make it see 'small bin' rather than just 'bin,' but the current behavior leads to 'put my X in bin' which technically works just fine.

PR #5829, opened 2022-06-08
[data][base.yaml] - Add roar helm noun

asechrest:
#5819  relies on this.

Add `roar_helm_noun` to support scream helm functionality being added to combat-trainer for barbs.


MahtraDR:
Don't do this as a separate PR. Folks can end up with one and not the other. Put both in the same PR, then one merge will have all the necessary bits, and there can be no instance where someone has this and not the other, and vice versa.

PR #5830, opened 2022-06-10
[scripts] [inventory-manager] Included feature to save deed register entries

GreyhallowDR:
new arg "register" to get deed register, turn it to contents page, read register and save deed entries to inventory


rpherbig (approved):
I'm not super familiar with this script, but the changes look reasonable

PR #5833, opened 2022-06-15
Update base-recipes.yaml

Raykyn55:
Adds two missing parts fields for Saddle Blanket and Saddle Shabrack in outfitting section.


asechrest:
Thanks for the PR!

PR #5835, opened 2022-06-17
Minor edit to sew

Raykyn55:
There was some funky ordering in sew that had sew default to getting needles after assembly, when that wasn't always the next step, but using the pins command, so the script hangs.

Moved tool retrieval to the top of the command action, moved the command definition to the needles condition, removed the action to get needles after assembly. appears to be working.

### DIFF
--- a/sew.lic
+++ b/sew.lic
@@ -90,7 +90,7 @@ class Sew
     DRCC.stow_crafting_item(item, @bag, @belt)
     room = Room.current.id
     DRCT.order_item(@hometown_data['tool-room'], 5)
-    DRCC.stow_crafting_item('pin', @bag, @belt)
+    DRCC.stow_crafting_item('pins', @bag, @belt)
     DRCT.walk_to(room)
     DRCC.get_crafting_item(item, @bag, @bag_items, @belt)
     Flags.reset('sew-pins')
@@ -111,6 +111,7 @@ class Sew
 
   def leather
     check_thread
+    check_pins
     DRCA.crafting_magic_routine(@settings)
     study_tailoring(@chapter, @recipe_name)
     DRCC.get_crafting_item("#{@mat_type} leather", @bag, @bag_items, @belt)
@@ -119,6 +120,7 @@ class Sew
 
   def sew
     check_thread
+    check_pins
     DRCA.crafting_magic_routine(@settings)
     study_tailoring(@chapter, @recipe_name)
     DRCC.get_crafting_item("#{@mat_type} cloth", @bag, @bag_items, @belt)
@@ -142,6 +144,7 @@ class Sew
 
   def lighten
     check_thread
+    check_pins
     study_tailoring(1, 'tailored armor lightening')
     DRCC.get_crafting_item('scissors', @bag, @bag_items, @belt)
     @noun = DRC.right_hand.noun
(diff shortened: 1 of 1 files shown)


Raykyn55:
When restocking thread, it would drop the item being crafted on the ground, which is never a good idea. Then it would attempt to resume crafting, but wouldn't know what to do, so it just dumped blank commands and hung. Now it stows in the bag and goes back to sewing.

Removed includes, added prefixes.

Removed get/stow methods, used common's methods instead.

sew.lic
```diff
@@ -261,18 +249,16 @@ class Sew
 
   def do_sew(command, tool, lighten = false)
     waitrt?
-    crafting_magic_routine(@settings)
-    stow_item(left_hand) if checkleft
+    DRCA.crafting_magic_routine(@settings)
+    DRCC.stow_crafting_item(DRC.left_hand, @bag, @belt) unless DRC.left_hand == (tool)
     if @last_command == "cut my #{@mat_type} cloth with my scissors"
```
rpherbig, line 254: Should `cloth` be `@mat_noun`?

sew.lic
```diff
@@ -54,16 +48,18 @@ class Sew
...
-    wait_for_script_to_complete('buff', ['sew'])
+    DRC.wait_for_script_to_complete('buff', ['sew'])
 
     Flags.add('sew-assembly', 'another finished \S+ shield (handle)', 'another finished wooden (hilt|haft)', 'another finished (long|short|small|large) leather (cord|backing)', 'another finished (small|large) cloth (padding)', 'another finished (long|short) wooden (pole)')
     Flags.add('sew-pins', 'The pins is all used up, so you toss it away.')
 
     if args.type == 'knitting'
       knit
     elsif args.type == 'sewing'
+      @mat_noun = 'cloth'
```
rpherbig, line 59: I see several references to `cloth` and `leather` still in the script. And the single use of `@mat_noun` is in a common helper whose calling function already knows if it is cloth or leather.

What do you think about adding a `material_type` argument to `do_sew`?

sew.lic
rpherbig, line 254: Should `cloth` be `@mat_noun`?

sew.lic
```diff
@@ -78,72 +74,65 @@ class Sew
...
   end
 
   def leather
     check_thread
-    crafting_magic_routine(@settings)
+    check_pins
+    DRCA.crafting_magic_routine(@settings)
     study_tailoring(@chapter, @recipe_name)
-    get_item("#{@mat_type} leather")
+    DRCC.get_crafting_item("#{@mat_type} leather", @bag, @bag_items, @belt)
```
rpherbig, line 117: Should `leather` be `@mat_noun`?

sew.lic
rpherbig, line 254: Why do we store and check the `@last_command` at the top of `do_sew`, instead of doing this logic at the end of the previous execution of `do_sew`?

sew.lic
Raykyn55: So far as i could tell, there was only one spot i missed. One reason NOT to add that argument is that as soon as you cut it with scissors, it's no longer "cloth" or "leather", but is now "cloak" or "gloves". submitting the change.

sew.lic
Raykyn55: Yes, submitted.

Raykyn55: I suppose it doesn't matter if that's how we're going to do it. An alternative would be checking feet slot? not sure how to do that. Could also put something like this in cleanup, instead of eatly in the process? Maybe extend the setup for both sewing and leather?

MahtraDR (approved):
I've had this running the last few days and looks good. Merging.

PR #5837, opened 2022-06-17
[data] [base-town] Making Wolf Clan explicit, and small cleanup

MahtraDR:
Using anchors for repairers so we can re-use them and know which we're reusing where.

Fleshing out Wolf Clan so folks can be explicit about repairing at Randal in WC, versus Catrox in Crossing city.

We should allow named repairer calls eventually, or work out better logic so we can use Fang Cove repairers, but that's larger scope...


rpherbig (approved):

MahtraDR:
Any objection to making Randal exclusive to Wolf Clan, and just force Catrox for Crossing repairs?

MahtraDR:
Made Crossing exclusive to Catrox.

Added entry for Rangu.

Made Randal exclusive to Wolf Clan.

Issue #5566, opened 2022-02-11
shape script


Kaesken:
Script tried to apply glue twice and bonked.
```
[shape]>get my glue

You get some wood glue from inside your duffel bag.

> 
[shape]>apply my glue to my shortbow

The shortbow appears ready to be reinforced with some backer material.

You apply a layer of glue to the pieces of your shortbow.
Roundtime: 19 sec.

R> 
The world takes on a glimmering red monotone briefly, but soon returns to normal.

R> 
The black rat on Damiza's shoulder snuffles a bit at her head.  It's not likely to find anything too interesting on that expanse of scalp!
[shape]>apply my glue to my shortbow

The shortbow is ready to be laminated with backer material.  You must assemble the backer with the shortbow before proceeding.

> 
New patches of dark red begin to appear on Damiza's chestwrap, slowly oozing up from the deeper layers of fabric to seep through to the surface with a glistening, sanguine dampness.
Rangu lays a well-oiled set of rifflers on the counter and packs them snugly away in a piece of burlap.

> 
Talia consults with several attendants, directing them to organize various areas of the room.

> 
[shape: *** No match was found after 15 seconds, dumping info]

[shape: messages seen length: 13]

[shape: message: Talia consults with several attendants, directing them to organize various areas of the room.]

[shape: message: There are 2 members in your group.]

[shape: message:   You (Leader): Healthy.]

[shape: message:   Damiza: Healthy.]

[shape: message: Members of your group:]

[shape: message: Murrula's Flames  (OM)]

[shape: message: Divine Radiance  (OM)]

[shape: message: Centering  (2 roisaen)]

[shape: message: Persistence of Mana  (OM)]

[shape: message: Osrel Meraud  (74%)]

[shape: message: Rangu lays a well-oiled set of rifflers on the counter and packs them snugly away in a piece of burlap.]

[shape: message: New patches of dark red begin to appear on Damiza's chestwrap, slowly oozing up from the deeper layers of fabric to seep through to the surface with a glistening, sanguine dampness.]

[shape: message: The shortbow is ready to be laminated with backer material.  You must assemble the backer with the shortbow before proceeding.]

[...]

PR #5839, opened 2022-06-19
[script][shape] - Fix glue application

Raykyn55:
Small change stemming from issue #5566 

added missing language for the flag

match messaging was present in the case statement, but there was nothing done with it. This, in part, caused the problem we saw in the referenced issue. 

For composite bows, what we needed during initial assembly was a strict ordering of setting up for glue, assembling the backer, applying the glue, then matching that application with the clamps step. 

However, during lamination for any bow, the ordering was applying glue, then setting up for clamps, assembling the backer, then applying clamps.

These two issues were resolved moving the glue success messaging into the clamps section (conditional?), and adding "appears ready to be" messaging to the flag. The last WHEN statement was added as a backstop in case the Flag gets skipped.

[shape log.txt](https://github.com/rpherbig/dr-scripts/files/8934620/shape.log.txt)

[shape log.txt](https://github.com/rpherbig/dr-scripts/files/8934650/shape.log.txt)

[shape log.txt](https://github.com/rpherbig/dr-scripts/files/8934676/shape.log.txt)

I suspect bolts/arrows might need similar love, unless it's been updated since the last time I tried making them.

### DIFF
--- a/shape.lic
+++ b/shape.lic
@@ -53,7 +53,7 @@ class Shape
     @training_spells = @settings.crafting_training_spells
 
     wait_for_script_to_complete('buff', ['shape'])
-    Flags.add('shaping-assembly', 'appears ready to be reinforced with some (backer)', 'You need another bone, wood or horn (backing) material', 'another finished bow (string)', 'another finished (long|short) wooden (pole)', 'another .* (leather (?:strip|cord))', 'You must assemble the (backer)', 'You must assemble the (strips)', 'another arrow (flights)', 'another .* (arrowhead)')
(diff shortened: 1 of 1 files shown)


asechrest:
@Raykyn55 

Appreciate you taking on some cleanup of the crafting scripts. These scripts are complex, interwoven, old (meaning old code), and can be difficult to follow given they have few code comments and often lack debug messages. And they're not as commonly used as some of the other curated scripts that are used constantly by all Lich users.

I know you've been waiting for me to test some of the changes in other crafting scripts and I apologize for not having the time to do so. What would really help me review these better is if you pasted succinct logs within the comment itself rather than linking to full text files. I appreciate the full text files too, so I don't mean to be picky, but you've obviously got some ability to help clean these scripts up so I trust you to identify the exact spot of the failure, post the log in the comment, fix the failure, then post the log of the success in the comment, and then the reviewers can check the code out to confirm, and comment or merge.

It would just make it easy to see that things were working without getting a bevy of testers to test these PRs, which is hard to do on the crafting scripts. So something like this:

```
Here is the log of the failure.
```
_Here is what I did to fix._
```
Here is success after my fix.
```
Really appreciate you contributing.

asechrest (approved):

Raykyn55:
Copy, will do.

asechrest:
@MahtraDR I'm good to merge here if you review.

rpherbig (approved):

PR #5844, opened 2022-06-19
[script][attunement] - Add magic training to attunement walking 

Hiinky:
Since we can't play worn instruments during perceive walking of any variety, it would be worthwhile to weave magic casting. Since attunement walking can take a long time, depending on the type of walk, it is quite possible run out of mana with these big no-camb casts. For empaths (some still health walk) in particular, this could be fatal so I opted to add a toggle here to prevent users being surprised by a sudden lack of mana.


attunement.lic
```diff
@@ -146,13 +149,23 @@ class Attunement
     waitrt?
   end
 
+  def magic_cleanup
+    return if @training_spells.empty?
```
rpherbig, line 153: If this is the only use of `@training_spells`, then I suggest instead storing `@training_spells.empty?` once.

On second thought, why not make this conditional on `@magic_filler`?

MahtraDR:
Closing as abandoned. Happy to revisit in a fresh/up to date PR.

PR #5853, opened 2022-06-21
[script] [combat-trainer] Add match for redeemed necro dissecting

MahtraDR:
A redeemed necro is kinda sorta not a necro in some ways.

This allows a redeemed necro to keep a safe, and blank rituals section, and just dissect like everybody else...


rpherbig (approved):

rpherbig:
"and just dissect like everybody else"

Who are you hanging out with? 😁

PR #5854, opened 2022-06-21
[scripts] [combat-trainer] Make arranging when dissecting optional

MahtraDR:
Reordering the sequence of skinning/dissecting, to make the decision to always arrange optional.

Some creatures teach Skinning when arranged, depending on your skill level, and the range they teach to. When this happens, it's beneficial to always arrange.

When that doesn't happen... arranging to dissect is a waste of time.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -387,7 +387,6 @@ class LootProcess
     # Setting to determine whether to always arrange, regardless of skinning, or dissecting
     # Depending on the creature level, and the character's skinning skill, simply arranging will teach skinning
     # Defaulting to true to keep existing behaviour.
-    # @always_arrange = skinning['always_arrange'] || true
     @always_arrange = skinning['arrange_for_dissect'].nil? ? true : skinning['arrange_for_dissect']
     echo("  @arrange_for_dissect: #{@arrange_for_dissect}") if $debug_mode_ct
 
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -804,6 +804,7 @@ class LootProcess
       return
     end
     return if Time.now - @loot_timer < @loot_delay
+    @already_arranged = false
 
     game_state.mob_died = true
     waitrt?
@@ -897,7 +898,7 @@ class LootProcess
 
     arrange_mob(mob_noun, game_state) unless @already_arranged
     @already_arranged = false
-    
+
     pause 0.25
     waitrt?
     if game_state.need_bundle
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -861,8 +861,8 @@ class LootProcess
   #not necromancer related - not to be confused with the necro rituals.  According to GMs this may change but for now they are two separate functions.
   def dissected?(mob_noun, game_state)
     arrange_mob(DRRoom.dead_npcs.first, game_state) if @always_arrange
-    case bput("dissect #{mob_noun}","You'll gain no insights from this attempt", 'You succeed in dissecting the corpse','What exactly are you trying to dissect',"You'll learn nothing further",'While likely a fascinating study','You cannot dissect','would probably object',"should be left alone.","That's going to be hard","That'd be a waste of time.",'A skinned creature is worthless','You do not yet possess the knowledge', 'This ritual may only be performed on a corpse', /You learn something/i, 'A failed or completed ritual has rendered')
-    when 'You succeed in dissecting the corpse', /You learn something/i
+    case bput("dissect #{mob_noun}","You'll gain no insights from this attempt", 'You succeed in dissecting the corpse','What exactly are you trying to dissect',"You'll learn nothing further",'While likely a fascinating study','You cannot dissect','would probably object',"should be left alone.","That's going to be hard","That'd be a waste of time.",'A skinned creature is worthless','You do not yet possess the knowledge')
+    when 'You succeed in dissecting the corpse'
       return true
     when 'While likely a fascinating study',"That'd be a waste of time.",'You do not yet possess the knowledge','This ritual may only be performed on a corpse', 'A failed or completed ritual has rendered'
       game_state.undissectable(mob_noun)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -897,7 +897,6 @@ class LootProcess
     return unless game_state.skinnable?(mob_noun)
 
     arrange_mob(mob_noun, game_state) unless @already_arranged
-    @already_arranged = false
 
     pause 0.25
     waitrt?
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -387,7 +387,7 @@ class LootProcess
     # Setting to determine whether to always arrange, regardless of skinning, or dissecting
     # Depending on the creature level, and the character's skinning skill, simply arranging will teach skinning
     # Defaulting to true to keep existing behaviour.
-    @always_arrange = skinning['arrange_for_dissect'].nil? ? true : skinning['arrange_for_dissect']
+    @arrange_for_dissect = skinning['arrange_for_dissect'].nil? ? true : skinning['arrange_for_dissect']
     echo("  @arrange_for_dissect: #{@arrange_for_dissect}") if $debug_mode_ct
 
     @arrange_all = skinning['arrange_all'] || false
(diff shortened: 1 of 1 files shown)


Hiinky:
Seems redundant as this is already controlled by the arrange fields.  
```
skinning:
  skin: false
  arrange_all: false   #<<<<
  arrange_count: 0     #<<<<
  tie_bundle: true
  dissect: true
  dissect_priority: First Aid
  ```

**Comment by MahtraDR** [comment]
> Seems redundant as this is already controlled by the arrange fields.
> 
> ```
> skinning:
>   skin: false
>   arrange_all: false   #<<<<
>   arrange_count: 0     #<<<<
>   tie_bundle: true
>   dissect: true
>   dissect_priority: First Aid
> ```

That would stop all arranging. I am simply making arranging before dissecting optional. I still want to arrange before skinning.

File: combat-trainer.lic [hunk 1]
```diff
@@ -854,6 +859,7 @@ class LootProcess
   end
 
   def dissected?(mob_noun, game_state)
+    arrange_mob(DRRoom.dead_npcs.first, game_state) if @always_arrange
     case bput("dissect #{mob_noun}","You'll gain no insights from this attempt", 'You succeed in dissecting the corpse','What exactly are you trying to dissect',"You'll learn nothing further",'While likely a fascinating study','You cannot dissect','would probably object',"should be left alone.","That's going to be hard","That'd be a waste of time.",'A skinned creature is worthless','You do not yet possess the knowledge', 'This ritual may only be performed on a corpse', /You learn something/i, 'A failed or completed ritual has rendered')
```
MahtraDR, line 863:
```suggestion
    case DRC.bput("dissect #{mob_noun}","You'll gain no insights from this attempt", 'You succeed in dissecting the corpse','What exactly are you trying to dissect',"You'll learn nothing further",'While likely a fascinating study','You cannot dissect','would probably object',"should be left alone.","That's going to be hard","That'd be a waste of time.",'A skinned creature is worthless','You do not yet possess the knowledge', 'This ritual may only be performed on a corpse', /You learn something/i, 'A failed or completed ritual has rendered')
```

MahtraDR:
@rpherbig Would you mind glancing at this one again? In particular, I added logic to prevent double-arranging, if we arranged after dissecting, and failed to dissect. It was arranging again before skinning.

combat-trainer.lic
```diff
@@ -384,6 +384,12 @@ class LootProcess
     @dissect_priority = skinning['dissect_priority']
     echo("  @dissect_priority: #{@dissect_priority}") if $debug_mode_ct
 
+    # Setting to determine whether to always arrange, regardless of skinning, or dissecting
+    # Depending on the creature level, and the character's skinning skill, simply arranging will teach skinning
+    # Defaulting to true to keep existing behaviour.
+    @always_arrange = skinning['always_arrange'].nil? ? true : skinning['always_arrange']
```
rpherbig, line 390: I saw this used to be `@always_arrange = skinning['always_arrange'] || true`, which I think is easier to read (and we use the same style below). Why was it changed?

combat-trainer.lic
```diff
@@ -854,7 +861,8 @@ class LootProcess
   end
 
   def dissected?(mob_noun, game_state)
-    case bput("dissect #{mob_noun}","You'll gain no insights from this attempt", 'You succeed in dissecting the corpse','What exactly are you trying to dissect',"You'll learn nothing further",'While likely a fascinating study','You cannot dissect','would probably object',"should be left alone.","That's going to be hard","That'd be a waste of time.",'A skinned creature is worthless','You do not yet possess the knowledge', 'This ritual may only be performed on a corpse', /You learn something/i, 'A failed or completed ritual has rendered')
+    arrange_mob(DRRoom.dead_npcs.first, game_state) if @always_arrange
```
rpherbig, line 864: The new instance var storing state and where all it gets read/written is hard for me to follow.

What do you think about moving all of that into `check_skinning`?

I'm picturing something like this:

```ruby
def check_skinning(mob_noun, game_state)
    already_arranged = false
    if @dissect && game_state.dissectable?(mob_noun)
      should_dissect = (@dissect_priority == 'First Aid' && DRSkill.getxp('First Aid') <= 32) ||
                                  (@dissect_priority == 'Skinning' && DRSkill.getxp('Skinning') >= 32) ||
                                  (DRSkill.getxp('First Aid') < DRSkill.getxp('Skinning')) ||
                                  (!game_state.skinnable?(mob_noun) && DRSkill.getxp('First Aid') <= 32) 
        if should_dissect
            if @always_arrange
                already_arranged = true
                arrange_mob(DRRoom.dead_npcs.first, game_state) 
            end
            return if dissected?(mob_noun, game_state)
        end
    end

    ... code ...

    arrange_mob(mob_noun, game_state) unless already_arranged
    already_arranged = false
```

rpherbig (commented):
The new instance var storing state and where all it gets read/written is hard for me to follow.

combat-trainer.lic
rpherbig: On second thought, I don't think that final `already_arranged = false` is even needed

combat-trainer.lic
MahtraDR: I want to say "because Seped told me to" but we found that this statement was always truthy.

combat-trainer.lic
MahtraDR: The reason for the instance var is that my next PR is to split the prioritization logic  out of `check_skinning` I've been working with Gildaren and Seped to setup a `should_skin_or_dissect` routine. I'm happy to put it as a local var in `check_skinning` for this PR, but then it'll change again.

MahtraDR:
> The new instance var storing state and where all it gets read/written is hard for me to follow.

The premise is that `@already_arranged = false` when we enter the loot routine. Presumably we only ever enter this routine once per corpse, so it always starts out false.

Then we enter the should skin, or dissect routine. If we have always_arrange, we arrange in the dissect routine, and set `@already_arranged = true` in there. 

Now if dissection fails, we can still skin. Since we have already set `@already_arranged = true`, we don't arrange again.

Having entered the skinning routine, `@already_arranged` has served its purpose, whatever happens from here, we set `@already_arranged = false` again in case we can't skin for some reason. Although, thinking about it, we don't even need it here... We can set it to false when we enter `dispose_body`  and just not arrange again in `check_skinning`

combat-trainer.lic
MahtraDR: > I think the final `already_arranged = false` can be dropped in that approach?

It can be dropped in any case, Because the corpse can't exist beyond this point. It was either skinned or dissect, and we only come back to this point again in a new corpse, at which point `@already_arranged` is off again from `dispose_body`

Removed it.

rpherbig (approved):

combat-trainer.lic
```diff
@@ -873,7 +872,10 @@ class LootProcess
   end
 
   def check_skinning(mob_noun, game_state)
+    already_arranged = false
     if @dissect && game_state.dissectable?(mob_noun)
+      arrange_mob(mob_noun, game_state) if @always_arrange
```
rpherbig, line 877: I think the initialization of `@always_arrange` was accidentally removed?

rpherbig, line 877: I think this _could_ change behavior, since we aren't guaranteed to dissect at this point, right?

MahtraDR: It was, re-added, renamed.

MahtraDR: I don't think so, because either way, we are in the dissect code, and even if we arrange here, and don't dissect, we just do it early for the skinning routine further down, which won't arrange again.

combat-trainer.lic
```diff
@@ -873,7 +878,10 @@ class LootProcess
   end
 
   def check_skinning(mob_noun, game_state)
+    already_arranged = false
     if @dissect && game_state.dissectable?(mob_noun)
+      arrange_mob(mob_noun, game_state) if @arrange_for_dissect
```
rpherbig, line 883: Should the conditional use `@always_arrange`? Or should line 390 be named `@arrange_for_dissect`?

MahtraDR: I left the old name in. Sorry about that, and thank you!

rpherbig (approved):

PR #5860, opened 2022-06-26
[scripts] [Athletics] Change shard climbing options

MahtraDR:
This allows for faster athletics in Shard, and safer, as the Zaldeni was never safe at that skill range, and this is more dynamic than practicing at the embrasure the whole time.


MahtraDR:
In addition to working for my baby necro, also tested by user Teldarion#2602 on discord. Looks good.

athletics.lic
```diff
@@ -231,10 +231,14 @@ class Athletics
 
   def shard_athletics
     until done_training?
-      if UserVars.athletics > 20 && UserVars.athletics <= 90
-        swim_zaldeni
-      elsif UserVars.athletics < 240
-        override_location_and_practice('shard_embrasure')
+      if UserVars.athletics < 240
```
asechrest, line 234: Should this be an `elsif`? I'm not sure of the code implications but this starts a new if/elsif block at this point.

MahtraDR: No, the stuff in red is what is being removed. :)

asechrest: Ah hell LOL. My fault. You can tell I'm tired.

PR #5864, opened 2022-06-27
[script][smith] Update, add prefixes, remove includes, simplified.lic

Raykyn55:
Removed includes.
Added prefixes.

Changed handling of oil since "of oil" only ever gets the crafting oil. Each smith cycle will consume a maximum of 3 parts of oil (one for initial craft, one for temper, one for enhancement) meaning if we have 3 when smith starts, we always have enough. Redundant oil checks removed.

The four enhancement methods all included a "get my noun" line, but forge doesn't stow, so this isn't necessary either.

As a result, five of the methods (craft_item, hone, balance, temper, lighten) were just running a script, so i just put the script calls in the smith method with the same conditional as before and removed those other methods.

I never used smith to stamp(forge w/ yaml setting), but I'm pretty sure it was broken as written. This version should work. 

Bumped coin_on_hand to 3 gold, figuring 1000 for oil, leaving 2000 for ingot/parts. Removed repeated coin checks, so one trip to the bank per cycle.

### DIFF
--- a/smith.lic
+++ b/smith.lic
@@ -52,7 +52,7 @@ class Smith
     DRCC.find_anvil(@hometown)
     DRC.wait_for_script_to_complete('buff', ['smith'])
     DRC.wait_for_script_to_complete('forge', [recipe['type'], recipe['chapter'], recipe['name'], material, recipe['noun']])
-    stamp_item(recipe) if stamp
+    stamp_item(recipe['noun']) if stamp
     DRC.wait_for_script_to_complete('forge', ['temper', recipe['noun']])  if temper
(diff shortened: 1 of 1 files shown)


Raykyn55:
@BinuDR suggested reverting oil handling from grabbing the first and trusting the user to keep other oils out of their crafting bag, TO getting the forging oil from wherever it lies within the crafting bag, sifting through any other oils that might be present. I'm not wedded to either idea, zero problem changing back. 

I'd say Binu's method is far more intelligent than mine.

Raykyn55:
So the oil issue is being resolved by moving it into a centrally handled method in common crafting. Closing this and resubmitting with additional updates.

PR #5866, opened 2022-06-27
[scripts] [combat-trainer] Remove includes, add prefixes to LootProcess class [1 of 3]

MahtraDR:
As per title.

PR #5867, opened 2022-06-27
[scripts] [combat-trainer] Removing fputs from LootProcess class [2 of some]

MahtraDR:
Depends on https://github.com/rpherbig/dr-scripts/pull/5866

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -364,13 +364,7 @@ class SetupProcess
 end
 
 class LootProcess
-  include DRC
-  include DRCI
-  include DRCS
-  include DRCH
-  include DRCMM
-
-  def initialize(settings, equipment_manager)
+   def initialize(settings, equipment_manager)
     @equipment_manager = equipment_manager
     echo('New LootProcess') if $debug_mode_ct
     skinning = settings.skinning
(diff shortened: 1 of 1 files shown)


MahtraDR:
Note I have the final version of this from https://github.com/rpherbig/dr-scripts/pull/5869 running on my 8, no errors, or problems. They all perform exactly as I expect with FA/Skin/Thano priority, and predictable behaviour.

asechrest:
@MahtraDR Also have leading spaces in this PR for every DRC prefix.

MahtraDR:
Note this introduces a behaviour change. Currently, if you have too many items on your person, it tries to just dump a skin from your bundle.

There are at least two problems with this approach, nevermind that it's done with fputs... bundles can have other items than skins, and... too many items takes you to the automated junkyard now. Better we turn off skinning at that stage, and let the user manage their item count. No need to note here, the game will be very loud about it...

combat-trainer.lic
```diff
@@ -862,11 +862,7 @@ class LootProcess
       game_state.undissectable(mob_noun)
       return false
     when 'would probably object', "should be left alone."
-      fput ('dissect')
-    when "That's going to be hard"
```
rpherbig, line 866: Should `"That's going to be hard"` no longer be in the bput matches on line 856?

rpherbig (approved):
Approved, aside from the 1 comment

combat-trainer.lic
MahtraDR: Removed. Let's find out if we need it or not! :)

PR #5870, opened 2022-06-28
[script][combat-trainer] Apparent fix for load/aim loop

Raykyn55:
Inverted timer function, added incremental reduction in time-to-fire to break check_firing_time loops.

added @firing_check resets after each instance of game_state.loaded becoming false, in order that we not skip the initial condition on the first time check (which sets the disparity between now and next fire) when we start aiming after a fresh load.

added some debug lines to help with testing.

```
```
This line was leaving ct with nothing to do when full-aim message arrived, but the conditionals weren't passing the firing command(still not sure why) so i moved the flag opposite our timer. As long as EITHER the full aim message has arrived, OR the time has expired, we get a fire command, but until then, we keep cycling fillers.

Best way I've found to recreate the bug was commenting out aim_fillers entirely (4425 in the current version), do ranged weapons only, no offhand attacks, and only training abilities was combat-maneuvers.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -307,6 +307,7 @@ class SetupProcess
     # as soon as you equip another ranged weapon before you've had
     # a chance to load it because the game_state thinks you're still holding a loaded weapon.
     game_state.loaded = false
+    @firing_check = 0
 
     # Prepare the next weapon
     if next_summoned
@@ -2999,7 +3000,7 @@ class AttackProcess
 
     @firing_delay = settings.firing_delay
     echo(" @firing_delay: #{@firing_delay}") if $debug_mode_ct
-    @firing_timer = Time.now - @firing_delay
+    @firing_timer = Time.now
     echo("@firing_timer: #{@firing_timer}") if $debug_mode_ct
     @firing_check = 0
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
I've got this going locally, will report back.

combat-trainer.lic
```diff
@@ -3225,7 +3234,10 @@ class AttackProcess
                   'shoot'
                 end
       shoot_aimed(command, game_state)
+      game_state.clear_aim_queue
       game_state.loaded = false
+      @firing_check = 0
+      echo("@firing_check reset normal fire: #{@firing_check}") if debug_mode_ct
```
MahtraDR, line 3240:
```suggestion
      echo("@firing_check reset normal fire: #{@firing_check}") if $debug_mode_ct
```

combat-trainer.lic
```diff
@@ -3525,6 +3542,8 @@ class AttackProcess
           shoot_aimed('shoot', game_state)
         end
         game_state.loaded = false
+        @firing_check = 0
+        echo("@firing_check reset on maneuver fire: #{@firing_check}") if debug_mode_ct
```
MahtraDR, line 3546:
```suggestion
        echo("@firing_check reset on maneuver fire: #{@firing_check}") if $debug_mode_ct
```

asechrest:
Thanks for the PR! I'll pull this down this evening and run it, though I've never seen the bug this tries to fix. So I'll just test for no adverse issues.

Raykyn55:
> Thanks for the PR! I'll pull this down this evening and run it, though I've never seen the bug this tries to fix. So I'll just test for no adverse issues.

Most probably wouldnt, but with certain conditions, you can force it. Run blank aiming_trainables in the old one and youll see it more often, especially after powershot without dual load.

XTyrsin:
Working fine in general and appears to have successfully resolved the load-aim-shoot bug.

asechrest:
This has been through some testing and appears to fix the issue and not cause others. Will merge and monitor.

PR #5878, opened 2022-06-30
[scripts] [common-titems] Add function to count necro stackers

MahtraDR:
As per title. This is a dependency of an upcoming PR for supporting necro stackers in combat-trainer.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -550,8 +550,7 @@ module DRCI
 
   # Takes in the noun of the configured necro material stacker, and returns the current material item count.
   def count_necro_stacker(necro_stacker)
-    count = DRC.bput("study my #{necro_stacker}", /currently holds \d+ items/).scan(/\d+/).first.to_i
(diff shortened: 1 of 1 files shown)


common-items.lic
```diff
@@ -548,6 +548,12 @@ module DRCI
     DRC.rummage('B', container)
   end
 
+  # Takes in the noun of the configured necro material stacker, and returns the current material item count.
+  def count_necro_stacker(necro_stacker)
+    count = DRC.bput("study my #{necro_stacker}", /currently holds \d+ items/).scan(/\d+/).first.to_i
```
rpherbig, line 553: You could return the `bput` directly, without declaring count

rpherbig (approved):
Optional comment, fine as-is

common-items.lic
MahtraDR: Thank you! :) Done it that way.

PR #5882, opened 2022-07-01
[data] [base.yaml] Set health_threshold default to zero

Kiriawen:
setting health_threshold: to zero so afk.lic doesn't auto log.  this should be an option people opt into, not a default.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -1386,7 +1386,7 @@ listen_skills:
 waggle_sets:
 
 # Vitality/spirit level when combat-trainer will log you or afk will start warning about low health/spirit when you get halfway between 100 and this number.
-health_threshold: 50
+health_threshold: 0
 depart_on_death: false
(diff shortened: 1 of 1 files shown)


MahtraDR:
I agree with this, and personally happy to merge. 

What happens to folks who depend on this today?

Kiriawen:
As discussed on discord, we'll add a message in afk.lic that if the health_threshold: is set to 0 that the default has been changed and if they want the previous functionality they'll have to add it to their setup yaml.  explicit opt in.

After a few weeks we can remove the message.  By that time people should know about having to add it to the yaml to opt in, want it at 0 or don't use the afk script as is so don't care.

MahtraDR:
Discord discussion was here https://discord.com/channels/745675889622384681/745675890242879671/992273607101268078

MahtraDR (approved):

PR #5883, opened 2022-07-01
[script] [afk] adding messaging about the default health_threshold: 

Kiriawen:
The new messaging should be removed after 7/15/22.  By that time people should be used to the new default or have it set in their own yaml

This is in concert with #5882


MahtraDR (approved):

MahtraDR:
image

PR #5889, opened 2022-07-03
[script][common] - Log to window

asechrest:
Logs a message to specified window. Come to find out window names are case sensitive, so specifically not forcing case here.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -778,7 +778,6 @@ module DRC
   # Sends a message to the specified window.
   # By default the message is bold.
   def log_window(text, window_name, make_bold = true)
-    # window_name = window_name.downcase
     _respond(
       "<pushStream id=\"#{window_name}\"/>" + (make_bold ? bold(text) : text),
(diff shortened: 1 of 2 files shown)


MahtraDR:
Opinions on deprecating the window specific atmo method, and just changing it to use this method instead?

asechrest:
> Opinions on deprecating the window specific atmo method, and just changing it to use this method instead?

If [this search](https://github.com/rpherbig/dr-scripts/search?q=atmo) is correct and we've only got one script using the `DRC.atmo`, agreed.

asechrest:
Some final changes, and now ready. Options to create the window if it doesn't exist (tested on Wrayth and Genie), as well as pre-clear the window before logging if the user desires.

Also updated the only script that uses `atmo()`, as well as changing the atmo def to simply call `log_window`.

asechrest:
Screenshot 2022-07-04 121111

MahtraDR (approved):
Looks good to me. Tested on all my windows, I have the tarantula script going too, that also works exactly the same. Neat change!

PR #5890, opened 2022-07-03
[script][shape] Change continue to resume

Raykyn55:
Minor edit to standardize arguments among scripts. Think it's resume everywhere but in shape.

### DIFF
--- a/arrows.lic
+++ b/arrows.lic
@@ -33,14 +33,14 @@ class Arrows
           description: 'Get tools from clerk?'}
       ],
       [
-        { name: 'continue', regex: /continue/i, variable: true }
+        { name: 'resume', regex: /resume/i, variable: true }
       ]
     ]
 
     args = parse_args(arg_definitions)
 
     if args.continue
-      DRC.wait_for_script_to_complete('shape', ['continue', 'arrows'])
+      DRC.wait_for_script_to_complete('shape', ['resume', 'arrows'])
       exit_tasks
       exit
     end
--- a/shape.lic
+++ b/shape.lic
@@ -22,8 +22,8 @@ class Shape
         { name: 'noun', regex: /\w+/i, variable: true }
       ],
       [
-        { name: 'continue', regex: /continue/i, variable: true },
-        { name: 'noun', regex: /\w+/i, variable: true }
+        { name: 'resume', regex: /resume/i, variable: true, description: 'Attempts to continue an ongoing project, must be holding the item' },
+        { name: 'noun', regex: /\w+/i, variable: true, description: 'NOUN of the item you wish to continue.' }
       ],
       [
         { name: 'recipe_name', display: 'enhancement', options: %w[laminate lighten cable], description: 'Enhancements to crafted bows' },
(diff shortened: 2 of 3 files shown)


asechrest:
Will this break things for existing users who have `shape continue` hardcoded in T2, for example? Thinking about backward compatibility, which is tough with these kinds of changes.

Thoughts?

Raykyn55:
> Will this break things for existing users who have `shape continue` hardcoded in T2, for example? Thinking about backward compatibility, which is tough with these kinds of changes.
> 
> Thoughts?

I think even if these issues arise, small end user edits arent asking a lot. If any scripts use continue specifically for shape, and i cant think of any that do(trade maybe) then those scripts would be doing special handling for shape that could be removed and brought in line with sew, etc. The closer these scripts resemble one another in function, the easier it is to handle them as a group(workorders/craft), and they are more approachable to new users. This effort largely predates me, im just continuing that work.

MahtraDR:
> Will this break things for existing users who have `shape continue` hardcoded in T2, for example? Thinking about backward compatibility, which is tough with these kinds of changes.
> Thoughts?

I am not fussed by changing this. This example isn't one I can see someone actually doing. Having shape continue in t2 would be weird. 😀

asechrest:
Not a deal-breaker for me, either. Just bringing it out in the open to discuss. That's my style, and you know what happens when we introduce a behavior change... :)

But if this is unlikely to be hard-coded in folks' T2, I like the change.

Raykyn55:
You were right, it was a single line in Trade.

MahtraDR:
https://github.com/rpherbig/dr-scripts/blob/2dccd9a25edbed6c27633ce977e725d271a4b545/arrows.lic#L43 As well.

Raykyn55:
> https://github.com/rpherbig/dr-scripts/blob/2dccd9a25edbed6c27633ce977e725d271a4b545/arrows.lic#L43
> 
> As well.

Shape makes arrows now.

Raykyn55:
Easy change, though, ill grab it after dinner if nobody beats me to it.

MahtraDR:
> > https://github.com/rpherbig/dr-scripts/blob/2dccd9a25edbed6c27633ce977e725d271a4b545/arrows.lic#L43
> 
> > As well.
> 
> Shape makes arrows now.

Can you check if we can deprecate the arrows script then?

Raykyn55:
Ill look, i may plunder it first. Theres a lot of convenience in arrows that i didnt put into shape. Simpler for now to just edit that line, i think, regards this PR.

arrows.lic
```diff
@@ -33,14 +33,14 @@ class Arrows
...
       ],
       [
-        { name: 'continue', regex: /continue/i, variable: true }
+        { name: 'resume', regex: /resume/i, variable: true }
       ]
     ]
 
     args = parse_args(arg_definitions)
 
     if args.continue
```
asechrest, line 42: Per @Dartellum , needs to change to `if args.resume`.

PR #5893, opened 2022-07-04
[script] [combat-trainer] Add setting to ignore weapon mindstates [2 of 2]

MahtraDR:
~~Depends on https://github.com/rpherbig/dr-scripts/pull/5892~~

Adds setting `combat_trainer_ignore_weapon_mindstate` which when set, tells CT to ignore mindstates, and just work on action counts. Allows farming with capped weapons.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -28,7 +28,8 @@ class SetupProcess
     echo("  @cycle_regalia: #{@cycle_regalia}") if $debug_mode_ct
     @default_armor = settings.default_armor_type
     echo("  @default_armor: #{@default_armor}") if $debug_mode_ct
-
+    @ignore_weapon_mindstate = settings.combat_trainer_ignore_weapon_mindstate
+    echo("  @ignore_weapon_mindstate: #{@ignore_weapon_mindstate}") if $debug_mode_ct
     @gearsets = settings.gear_sets
     @warhorn = settings.warhorn
     echo("  @warhorn: #{@warhorn}") if $debug_mode_ct
@@ -197,6 +198,7 @@ class SetupProcess
 
   def determine_next_to_train(game_state, weapon_training, ending_ranged)
     return unless game_state.skill_done? || !weapon_training[game_state.weapon_skill] || ending_ranged
+    return if @ignore_weapon_mindstate
     #check if all weapons are maxed exp to prevent rapid cycling
     if weapon_training.reject { | skill,_ |  DRSkill.getxp(skill) == 34 }.empty?
       echo 'all weapons locked, not switching' if $debug_mode_ct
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -4,10 +4,9 @@
 custom_require.call(%w[common common-arcana common-healing common-items common-summoning common-theurgy common-moonmage common-travel drinfomon equipmanager events spellmonitor])
 
 class SetupProcess
-  include DRC
-  include DRCS
 
   def initialize(settings, equipment_manager)
+    $debug_mode_ct = true
```
MahtraDR, line 9:
```suggestion
```

profiles/base.yaml
```diff
@@ -307,6 +307,8 @@ zombie:
   behavior: defensive
 # a hash of weapon skills and the weapons used to train them. They must be defined in gear
 weapon_training:
+# Tells Ct to ignore mindstates, allowing one to farm whilst mindlocked, or with capped weapons
+combat_trainer_ignore_weapon_mindstate:
```
rpherbig, line 311: Should this default to `false`?

combat-trainer.lic
```diff
@@ -197,12 +198,16 @@ class SetupProcess
...
-      echo 'all weapons locked, not switching' if $debug_mode_ct
-      return
+    
+    unless @ignore_weapon_mindstate
+      #check if all weapons are maxed exp to prevent rapid cycling
+      if weapon_training.reject { | skill,_ |  DRSkill.getxp(skill) == 34 }.empty?
+        echo 'all weapons locked, not switching' if $debug_mode_ct
+        return
+      end
     end
```
rpherbig, line 208: Some suggestions (any of these can be done or not done independently):
1. Extract the `reject` call into a named var for clarity
2. Use `all?` instead of `reject {...}.empty?`
3. Merge the conditionals

```suggestion
    #check if all weapons are maxed exp to prevent rapid cycling
    no_weapons_to_train = weapon_training.all? { | skill,_ |  DRSkill.getxp(skill) == 34 }
    if no_weapons_to_train && !@ignore_weapon_mindstate
        echo 'all weapons locked, not switching' if $debug_mode_ct
        return
    end
```

profiles/base.yaml
rpherbig, line 311: Should this default to `false`?

combat-trainer.lic
MahtraDR: Initially it didn't make sense to me, but I added echoes, and tested, and this works well. Thank you for the suggestion.

profiles/base.yaml
MahtraDR, line 311:
```suggestion
combat_trainer_ignore_weapon_mindstate: false
```

MahtraDR: Yes. :) Thank you.

MahtraDR:
@rpherbig Are you happy with this now? I've got it running and it works as intended at least.

rpherbig (approved):

PR #5895, opened 2022-07-04
[script][forge] Rewrite

Raykyn55:
The good:
- Everything works. Enhancements, Temper, RESUME, but especially armor crafting.
- Backwards compatibility. Any script that used forge previously, uses this script with the same arguments to the same effect.
- Supports adjustable tongs on the new method more intelligently
- debugging argument with a good bit of information to see what's actually going on. Also monitors debug_mode.
- tools handling is a little cleaner, more in line with how shape handles them. There are some differences by necessity, but in large part, tool changes are done with the right hand, and items are held in your left. 

The bad:
- It's more complicated. A lot of the methods in forge have been subsumed into work or prep method. Not as easy to follow the logic.
- Likely instance variables that could be local, especially since settings is passed. Fresh eyes would be helpful.

The ugly:
- work method is big, easy to get lost in. If something is broken there, no aspect of this script will work properly. Before, even if say hone didn't work, temper would. If armor didn't work, weaponsmithing would. In this, either they all work, or none do.

### DIFF
--- a/forge.lic
+++ b/forge.lic
@@ -2,163 +2,169 @@
   Documentation: https://elanthipedia.play.net/Lich_script_repository#forge
 =end
 
-custom_require.call(%w[common common-arcana common-crafting common-travel events drinfomon])
+custom_require.call(%w[common common-arcana common-crafting common-items common-travel events drinfomon])
 
 class Forge
 
   def initialize
+
     arg_definitions = [
       [
         { name: 'book_type', display: 'book type', options: %w[black armor weapon], description: 'What smithing type is this item.' },
         { name: 'chapter', regex: /\d+/i, variable: true, description: 'Chapter containing the item.' },
         { name: 'recipe_name', display: 'recipe name', regex: /^[A-z\s\-\']+$/i, variable: true, description: 'Name of the recipe, wrap in double quotes if this is multiple words.' },
         { name: 'metal', regex: /\w+/i, variable: true, description: 'Type of metal ingot to use.' },
-        { name: 'noun', regex: /\w+/i, variable: true }
+        { name: 'noun', regex: /\w+/i, variable: true, description: 'name of item being crafted' },
+        { name: 'debug', regex: /debug/i, optional: true, description: 'displays debug information' }
       ],
       [
         { name: 'book_type', display: 'book type', options: %w[black armor weapon], description: 'What smithing type is this item.' },
         { name: 'instruction', regex: /\w+/i, variable: true, description: 'instruction noun' },
         { name: 'metal', regex: /\w+/i, variable: true, description: 'Type of metal ingot to use.' },
-        { name: 'noun', regex: /\w+/i, variable: true }
+        { name: 'noun', regex: /\w+/i, variable: true, description: 'name of item being crafted' },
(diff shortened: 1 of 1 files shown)


MahtraDR:
I've got this going on mine. Will revert.

Raykyn55:
Plural spikes and you get need, not needs. works fine without those two words.

MahtraDR (approved):
This tests well for me.

MahtraDR:
Encountered an issue last night where it wasn't pulling oil from crafting container, and used up all my holy oil instead. We're working through these.

Raykyn55:
Change 'of oil' to 'oil' to match crafting container settings. If you're holding more than one type of oil, you'll want to have 'oil' listed under that crafting_items_in_container.

MahtraDR:
> Change 'of oil' to 'oil' to match crafting container settings. If you're holding more than one type of oil, you'll want to have 'oil' listed under that crafting_items_in_container.

Yep yep, this is the expected behaviour. :)

Raykyn55:
> > Change 'of oil' to 'oil' to match crafting container settings. If you're holding more than one type of oil, you'll want to have 'oil' listed under that crafting_items_in_container.
> 
> Yep yep, this is the expected behaviour. :)

With that in mind, changed how smith handles this as well, so it doesn't go and dump your holy oil.

BinuDR:
> > > Change 'of oil' to 'oil' to match crafting container settings. If you're holding more than one type of oil, you'll want to have 'oil' listed under that crafting_items_in_container.
> > 
> > 
> > Yep yep, this is the expected behaviour. :)
> 
> With that in mind, changed how smith handles this as well, so it doesn't go and dump your holy oil.

The way I had this setup before was that it would `look oil` in your crafting container, and if it wasn't the correct type it would go along the ordinals till it found the right one and get it out.  When done with it, be sure to put it back into the crafting container.

This is what it should look like:
```R>l oil
A thick and syrupy substance that easily coats and lubricates metal.```

Holy oil looks like this:
```>l holy oil
You see nothing unusual.```

BinuDR:
Actually, this was in the smith script, not forge.  Here's the code:

```  def check_oil(discipline)
    $ORDINALS.each do |ord|
      case bput("look my #{ord} oil in #{@container}", /thick and syrupy substance that easily coats and lubricates metal/, /You see nothing unusual/, /I could not find what you were referring to/, /There appears to be something written/, /(coal|lamp|fish) oil/, /The oil bottle is full/)
      when /thick and syrupy substance that easily coats and lubricates metal/
        return if ord == 'first'
        bput("get my #{ord} oil in #{@container}", /You get/)
        bput("put oil in my #{@container}", 'You put')
        return
      when /I could not find what you were referring to/
        break
      end
    end```

MahtraDR:
Current version works well on my cleric. Errors of yesterday are gone. Check Binu's suggestion too?

Raykyn55:
@BinuDR For the purposes of this PR, I think the current method accords with smith in either configuration. For smith, I'm going to comment on that PR (https://github.com/rpherbig/dr-scripts/pull/5864)

MahtraDR:
Ran multiple iterations, didn't have a problem. Merging. We fix issues whenif we find them...

PR #5905, opened 2022-07-12
combat-trainer - Warhorn/egg bput fix

BinuDR:
Changing the bput matches to regexes instead of strings because we found out that the string matches get converted into case insensitive in the bput code.

Also added anchors to the matches to make it as specific as possible.


MahtraDR (approved):
Tests well for me.

MahtraDR:
Merging now. If we have problems, we address them as we go.

PR #5907, opened 2022-07-13
[script][forge] Add Master Crafting books

Raykyn55:
Works for the worn master crafting book, requires yaml setting:
```
master_crafting_book: true
```
Tested by Chyral:
```
[forge]>turn my book to discipline armorsmithing

You turn the book to the section on armorsmithing.

[forge]>turn my book to chapter 4

You turn your book to chapter 4, entitled "Shield Design".

[forge]>read my book
```
Couple additional changes:
1. First, turn_to method wasn't being used anymore, so went ahead and deprecated.
2. book_type is ONLY used by the end user to start forge. Otherwise, smith was submitting the full name (armorsmithing) when running, which forge picked up anyways. This allowed me to dump book_type right into the master crafting book portion in a way that would work when forge was run alone, and when run by smith.
3. Only gets or stows the book when it's NOT a master crafting book.

If anyone has a master crafting book that isn't worn, they may run into trouble.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -132,15 +132,15 @@ module DRCC
     end
   end
 
-  def find_recipe(chapter, match_string)
-    case DRC.bput("turn my book to chapter #{chapter}", 'You turn', 'You are too distracted to be doing that right now', 'The book is already turned')
+  def find_recipe(chapter, match_string, book = 'book')
+    case DRC.bput("turn my #{book} to chapter #{chapter}", 'You turn', 'You are too distracted to be doing that right now', 'The .* is already turned')
     when 'You are too distracted to be doing that right now'
       echo '***CANNOT TURN BOOK, ASSUMING I AM ENGAGED IN COMBAT***'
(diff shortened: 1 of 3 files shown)


BinuDR:
Awesome addition!  One thing though.  I think you're going to need to change your variable from true/false to have the book type because my master crafting book won't work with this change.

`an ornate codex of master crafting instructions`

I'd suggest `master_crafting_book: ornate codex`
Maybe if the setting is true, have it default to `book` but otherwise read the book adj+noun from the setting.

BinuDR:
Either that or add a second setting for the name of the book.

Raykyn55:
Last edits were to handle alternate nouns. So the yaml setting master_crafting_book still defaults to false, but if you have a book, you would put the adjective(if exists) and noun.
```
master_crafting_book: book
master_crafting_book: codex
etc.
```

BinuDR:
I tested this out and it's working well.

PR #5910, opened 2022-07-14
,attunement - moons fix

GreyhallowDR:
If you have a moonstone gem in an open container the PERC MOONS command will fail. PERC MOONS IN SKY is a more robust way to insure that can't happen


MahtraDR:
I wish we could observe spider in the sky too :)

MahtraDR (approved):
Tested, looks good.

PR #5922, opened 2022-07-17
[script][combat-trainer] - Refactor egg and warhorn use

asechrest:
Refactors egg and warhorn use, allows rotating of each.

~~Draft until final cleanup.~~ Ready for review.

A request in Discord by user Sharkbait requested the ability to rotate warhorn and egg use in combat. Legacy use of a warhorn or egg existed in `class GameState` and relied on a game state of `dancing` in order to fire. You could only use one or the other. It relied on a single yaml setting: `warhorn:`. 

After much discussion we settled upon a few things:

- A warhorn effect is a room effect that lasts 10 minutes
- Warhorn and egg use should be independent. Since the room effect is every 10 minutes, with egg cooldown being 15 minutes and warhorn cooldown being 15 - 20 minutes, rotation of the two would allow a permanent room effect every 10 minutes.
- Warhorn cooldown should be configurable. Most are 20 minutes, but some are 15 minutes.
- Warhorn/egg use should not rely on a game state of `dancing`
- We should move the code logic from `class GameState` to `class AbilityProcess`. This functions essentially like an ability. A paladin's use of a prayer badge is code-analogous and that exists in AbilityProcess.

The logic was also heavily refactored. The code allows the legacy use of `warhorn: egg` in order for backward compatibility. Timers were moved to UserVars to allow persistence between CT sessions. Allows rotation of a warhorn and egg and picks the one with the shortest timer to start. Has failsafes if the warhorn or egg can't be used, and removes them from the array for that CT session. Considers both the room effect cooldown of 10 minutes as well as each individual warhorn/egg cooldown, attempting to use the effect on timers as tightly as possible.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -424,6 +424,8 @@ custom_loot_type:
 last_rites: false
 # define the noun to turn on warhorn usage every 5 minutes in combat
 warhorn:
+# Set the cooldown of your warhorn in seconds. Default is 20 minutes. Some horns are 15.
+@warhorn_cooldown: 1200
 # noun of the prayer mat to use in training_abilities
 prayer_mat:
 # optional container to bypass default mat storage in theurgy_supply_container
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -422,10 +422,12 @@ tk_ammo:
 custom_loot_type:
 # perform last_rites or not?
 last_rites: false
-# define the noun to turn on warhorn usage every 5 minutes in combat
+# define the noun to turn on warhorn usage in combat.
 warhorn:
 # Set the cooldown of your warhorn in seconds. Default is 20 minutes. Some horns are 15.
 warhorn_cooldown: 1200
+# Define the noun to turn on egg usage in combat. Cooldown is 15 minutes.
+egg:
 # noun of the prayer mat to use in training_abilities
 prayer_mat:
 # optional container to bypass default mat storage in theurgy_supply_container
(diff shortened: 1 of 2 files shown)


profiles/base.yaml
```diff
@@ -424,6 +424,8 @@ custom_loot_type:
 last_rites: false
 # define the noun to turn on warhorn usage every 5 minutes in combat
 warhorn:
+# Set the cooldown of your warhorn in seconds. Default is 20 minutes. Some horns are 15.
+@warhorn_cooldown: 1200
```
MahtraDR, line 428:
```suggestion
warhorn_cooldown: 1200
```

MahtraDR (changes requested):

asechrest:
First tests by `sharkbait` on Discord are positive. He has an egg, as well as an auction warhorn with a 15 minute cooldown. It's cycling for him, with fairly tight timers. It's honoring the 10 minute room effect and cycling the egg and warhorn between 10 and 13 minutes after the last effect.

Waiting for additional testers who have only an egg, only a warhorn, and also someone using the legacy `warhorn: egg` setting.

asechrest:
@MahtraDR reports some issues with legacy use of `warhorn: egg` in this PR. It sometimes does nothing at all.

Also some funkiness with the cycle timers reported by @sharkbait where it skips the 20 minute mark, which should be a point where the timers work when cycling.

So this needs a bit more work.

asechrest:
@rpherbig @MahtraDR 

Ready for final review. Testers were: Gildaren (warhorn only), Mahtra (egg only), Sharkbait (egg and warhorn rotation, plus egg only on a thief).

Updating the original comment now with a summary of changes.

PR #5948, opened 2022-07-20
[data] [base.yaml] Merge *_no_use_rooms into one list, and use anchors.

MahtraDR:
<pushBold/>Failed to activate due to room settings.<popBold/>
<pushBold/>Recommend you add room number: 9639 to wand_watcher_no_use_rooms in your yaml.<popBold/>

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -326,8 +326,10 @@ module DRCMM
     moon = visible_moons.first
     if moon
       data['cast'] = "cast #{moon}"
+    elsif !moon && data['name'].downcase == 'cage of light'
+      data['cast'] = "cast ambient"
     else
-      echo "No moon available to cast #{data['abbrev']}"
+      echo "No moon available to cast #{data['name']}"
       data = nil
     end
     data


Dartellum:
You might want to add it to `pattern_hues_no_use_rooms` as well.

MahtraDR:
> You might want to add it to `pattern_hues_no_use_rooms` as well.

And probably to buff-watcher too. Should we maintain a common-list and just anchor it in?

Dartellum:
> > You might want to add it to `pattern_hues_no_use_rooms` as well.
> 
> And probably to buff-watcher too. Should we maintain a common-list and just anchor it in?

That would make maintaining easier.

MahtraDR:
> > > You might want to add it to `pattern_hues_no_use_rooms` as well.
> > 
> > 
> > And probably to buff-watcher too. Should we maintain a common-list and just anchor it in?
> 
> That would make maintaining easier.

See the current version. Whatcha think?

Dartellum (approved):
Looks good. Perhaps nothing, but I do `<< : *anchor` not `<<: *anchor`.

MahtraDR:
> Looks good. Perhaps nothing, but I do `<< : *anchor` not `<<: *anchor`.

They're both valid yaml. The important thing isn't the space, or the colon, but the `<<` which is a [merge key](https://yaml.org/type/merge.html)

MahtraDR:
@Kiriawen @BinuDR I'd love your feedback here. Having the lists all merged, makes things much simpler to maintain. There are place here where an almanac could be used, where the magic stuff doesn't work, but I reckon this is a reasonable compromise, and the almanac will get used as soon as they leave that room anyway.

MahtraDR:
@Kiriawen @BinuDR I'd love your feedback here. Having the lists all merged, makes things much simpler to maintain. There are place here where an almanac could be used, where the magic stuff doesn't work, but I reckon this is a reasonable compromise, and the almanac will get used as soon as they leave that room anyway.

Dartellum:
> > Looks good. Perhaps nothing, but I do `<< : *anchor` not `<<: *anchor`.
> 
> They're both valid yaml. The important thing isn't the space, or the colon, but the `<<` which is a [merge key](https://yaml.org/type/merge.html)

Again, just something we may want to standardize one.

rpherbig (commented):
I'm not sure I have an opinion on this change - I haven't played in a while so I don't know how it would impact people.

I see the obvious 'pro' that a single list is easier to maintain; with the 'con' (as someone pointed out) that this is a little over-generalized and perhaps there are a few scripts in the list which could be used in a few rooms - so it's not without any functional change.

A few thoughts come to mind:

* are any of these rooms ones people are likely to be sitting in for a while, i.e. they would even notice the inability to use those scripts in these rooms?
* are any of these scripts ones which would have a severe impact if they couldn't be called for a while?
* are rooms likely to be added to this list in the future rooms which should be added for all affected scripts?
* do we have any knowledge of how/if people are overriding this setting (since it is, after all, only a default here)?

That last point especially could inform us how/if to change direction. Can someone poll lnet/discord/slack?

rpherbig:
As for the code itself, the changes look fine (assuming this is the direction we want to go). I would suggest adding a validator that detects overrides and warns that the defaults changed (or maybe even that the individual settings are deprecated...).

PR #5949, opened 2022-07-20
[script][base-crafting] Implement leather for workorders materials

Raykyn55:
Opens the option to use your preferred society leather for workorders under 
```yaml
workorders_materials:
  leather_type: rat/cougar
```
**Editted above from original (fabric_type: rat/cougar) which is no longer accurate.

Also updated the other materials full name to make them interactive if necessary, and then added handling in the grunt scripts to scrub that information.

Default fabric_type is still going to remain burlap, this is something the end user will need to configure via recipe whitelisting items according to their fabric_type preference eg if you whitelist leather items, you'll need to change fabric_type to a valid leather option, and vice versa.


Dartellum:
> ```yaml
> workorders_materials:
>   fabric_type: rat/cougar
> ```

This still applies?

Raykyn55:
No, thats leather type now.

Dartellum:
I believe we will need the volume for each recipe as well before letting loose in the wild.

Dartellum:
> i think there might be a better way to make these substitutions rather than editing base-crafting to have cloth/leather

Can go by chapter, 7 - 10 are leather. Volume is there; I must have overlooked it when scrolling.

Raykyn55:
> > i think there might be a better way to make these substitutions rather than editing base-crafting to have cloth/leather
> 
> Can go by chapter, 7 - 10 are leather. Volume is there; I must have overlooked it when scrolling.

There would need to be an additional variable defined for tailoring material type, not a hard change, but necessary since sew_items is handling these materials and had cloth hardcoded in before.

Raykyn55:
pushed that, lmk what you think @Dartellum

Dartellum:
> pushed that, lmk what you think @Dartellum

I like it. We also need to review `craft.lic` and adjust as needed.

PR #5959, opened 2022-07-26
Create heroic-tattoo.lic

Raykyn55:
Basic script you'd run at startup, calculates downtime for heroic tattoo based on your arcana, then attempts to keep it up at all times. Testing well for me.

### DIFF
--- a/heroic-tattoo.lic
+++ b/heroic-tattoo.lic
@@ -5,7 +5,7 @@ no_kill_all
 downtime = DRSkill.getrank('Arcana') < 1400 ? (60 - DRSkill.getrank('Arcana')/24)*60 : 0
 DRC.message("Approximate downtime on your Heroic Tattoo is : #{downtime/60} minutes")
 pause 1
-DRC.message("Attempts will be made every 5 minutes to active a tattoo on cooldown, or 5 minutes plus your downtime if tattoo currently active")
+DRC.message("Attempts will be made every 5 minutes to activate a tattoo on cooldown, or 5 minutes plus your downtime if tattoo currently active")
 pause 5
 timer = 3600 + downtime
 next_invoke = Time.now
--- a/heroic-tattoo.lic
+++ b/heroic-tattoo.lic
@@ -43,6 +43,7 @@ loop do
         next_invoke = Time.now + [300,downtime].min
         DRC.message("Next attempt in #{(next_invoke - Time.now).to_i/60} minutes")
     when 'already under a heroic effect'
+        break if args.once
         next_invoke = Time.now + (300 + downtime)
         DRC.message("Next attempt in #{(next_invoke - Time.now).to_i/60} minutes")
     when 'Closing your eyes, you carefully bend'
--- a/heroic-tattoo.lic
+++ b/heroic-tattoo.lic
@@ -1,0 +1,38 @@
+custom_require.call(%w[common])
+no_pause_all
+no_kill_all
+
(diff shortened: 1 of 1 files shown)


Raykyn55:
```--- Lich: afk paused.
[tattoo]>invoke my tattoo
You brace yourself as you activate your tattoo and feel empowered, as its magic washes over you.
>
Next attempt in 106 minutes
--- Lich: afk unpaused.
Gained: Arcana(+27)
```
chugging along nicely.

heroic-tattoo.lic
```diff
@@ -0,0 +1,57 @@
...
+if args.cycle
+    DRC.message("Heroic Tattoo uptime is 60 minutes per activation")
+    DRC.message("Approximate downtime on your Heroic Tattoo is : #{downtime/60} minutes")
+    DRC.message("So total cycle from activation to next possible activation is #{(downtime/60)+60} minutes")
+    exit
+end
+
+pause 5
+loop do
+    while next_invoke > Time.now || no_use_scripts.any? { |name| Script.running?(name) } || ( no_use_rooms.any? { |room| room === DRRoom.title.to_s()[2..-3] || room == Room.current.id} )
```
asechrest, line 28: Can you help me out on what's going on in this complex while loop? Just trying to understand it.

heroic-tattoo.lic
```diff
@@ -0,0 +1,57 @@
...
+
+no_use_rooms = get_settings.wand_watcher_no_use_rooms
+no_use_scripts = ['burgle','go2','athletics']
+downtime = DRSkill.getrank('Arcana') < 1400 ? (60 - DRSkill.getrank('Arcana')/24)*60 : 0
+timer = 3600 + downtime
+next_invoke = Time.now
+
+arg_definitions = [[
+      { name: 'once', regex: /once/i, optional: true, description: "One activation, however long that takes"},
+      { name: 'try', regex: /try/i, optional: true, description: "One ATTEMPT at activation, regardless of success or failure, and does not continue attempting"},
```
asechrest, line 14: What is the purpose of an invoke attempt that accepts a failure to invoke and doesn't care?

heroic-tattoo.lic
```diff
@@ -0,0 +1,57 @@
...
+
+    case DRC.bput("invoke my tattoo",
+        'as its magic washes over you',
+        'Its magic appears depleted',
+        'already under a heroic effect',
+        'Closing your eyes, you carefully bend')
+    when 'as its magic washes over you'
+        next_invoke = Time.now + timer
+        DRC.message("Next attempt in #{(next_invoke - Time.now).to_i/60} minutes")
+        break if args.once
```
asechrest, line 41: Are we properly unpausing the scripts that were paused when we `break` for args.once?

asechrest: Answering my own question: I think we do. It's a break not an exit. Thanks.

heroic-tattoo.lic
Raykyn55: First check is timer, second is scripts running, third is room id/description. If any of those are true(timer not expired, script is running we dont want to interfere with, or we're in a room thats on the blacklist) we pause and check again in 15 seconds. All have to be false to pass to the loop.

heroic-tattoo.lic
Raykyn55: T2 if folks dont want to run it all the time, but want to trigger it at the start of combat. T2 would wait for it to end, which could be a long time.

heroic-tattoo.lic
Raykyn55: actually, this is a problem, it leaves the whole loop. works on try, not on once.

PR #5961, opened 2022-07-27
[script][textsubs] Add tool durability subs in.

Dartellum:
When the durability subs were removed due to Simu granting the ability via toggle appraisal, tools were inadvertently effected. This change brings back the numbers to durability for tools.


Raykyn55:
Testing well on all my tools.

PR #5718, opened 2022-04-05
[data] [base-help] Create base-help.yaml

asechrest:
Second component part of robust help output for scripts. Again, thanks to @rcuhljr for help fleshing out this idea.

This new file works in concert with:

- #5712 

This file serves as a help file of yaml settings. The purpose of this file is to provide the base of information that a call to `;<script-name> help` will pull yaml settings help output from. The aforementioned change to `dependency`, separately, provides the structure that parses and outputs the data when a user calls the help argument.

**Background:**
Why do this? DR Lich's total number of yaml settings has grown significantly. One of the most common questions we have from users is: what yaml settings are used in this script? We of course have a substantially long file called `base.yaml` that contains most of our yaml settings, with some, but not all. of them code-commented. But many users don't know where to access that file, and it's long enough that it's difficult to reference. Further, Ruby doesn't parse code comments, so there is no way to draw on those comments to output help information.

**The advantages:**
Creating this help file of yaml settings, along with the changes to dependency that are backward-compatible, allow us to build this file over time into a complete yaml settings help document that can be drawn upon in-game to output help information. Further, this requires no changes in the scripts themselves. Rather, we're building the information in a separate file, manually. While that has downsides, the upsides is that we don't need to change every script, and, this project can be abandoned without any detriment to the project, if in the future the devs decide to do so.

**Examples of what this will look like:**

Within base-help.yaml we'll build our yaml settings like this, giving a general description, a list of all scripts that reference this yaml setting, and a specific description of how the yaml setting works with each script:

```yaml
# crafting_container: # This is the actual name of the setting as specified in base/character yamls
#   description: # This is the general description of the setting itself
#   referenced_by: # Here you should specify which scripts, by their script name, use this setting
#     - clean-leather
#     - carve-bead
#   specific_descriptions: # Here you should specify, by script name, the specific use of the setting
#     clean-leather: Scraper, preservatives and cleaned hides go here.
#     carve-bead: Carving knife and shaper go here.

hometown:
  description: Town you wish to live and work in.
  referenced_by:
    - clean-leather
  specific_descriptions:
    clean-leather: Used to look up Stock vendor to buy from and banking settings.
```

Doing so will allow us to have robust yaml settings output like this:

image

PR #5967, opened 2022-07-29
[script][card-collector] Create script card-collector

Raykyn55:
Useful fire-and-forget script if you're collecting cards. Doesn't take arguments, requires yaml settings:
```yaml
card_bags:
  fresh: bag1
  duplicates: bag2
```
Takes recently acquired cards from whatever bag you designate and attempts to file them in your collector's case. If they won't fit, puts them into a second bag you designate for duplicates.

### DIFF
--- a/card-collector.lic
+++ b/card-collector.lic
@@ -22,7 +22,7 @@ class Cards
     cards_add(fresh,duplicates)
   end
 
-  def cards_add(fresh,duplicates)
+  def cards_add(fresh, duplicates)
     unless /referring/i =~ DRC.bput("get my collector case", 'You get', 'You are already', 'What were you referring to')
       DRC.bput('swap', 'You move') unless DRC.right_hand == "collector's case"
       loop do
--- a/card-collector.lic
+++ b/card-collector.lic
@@ -10,13 +10,13 @@ class Cards
     ]
 
     args = parse_args(arg_definitions)
-    @settings = get_settings
+    settings = get_settings
     unless @settings.card_bags
       DRC.message("No bags defined in your yaml, please run this script with argument 'help' for more information.")
       exit
     end
-    fresh = @settings.card_bags['fresh']
-    duplicates = @settings.card_bags['duplicates']
+    fresh = settings.card_bags['fresh']
+    duplicates = settings.card_bags['duplicates']
 
     DRCI.stow_hands
     cards_add(fresh,duplicates)
--- a/card-collector.lic
+++ b/card-collector.lic
@@ -19,6 +19,10 @@ class Cards
 
     args = parse_args(arg_definitions)
     @settings = get_settings
+    unless @settings.card_bags
+      DRC.message("No bags defined in your yaml, please run this script with argument 'help' for more information.")
+      exit
+    end
     @card_bag = @settings.card_bags
     
     DRCI.stow_hands
--- a/card-collector.lic
+++ b/card-collector.lic
@@ -5,15 +5,7 @@ class Cards
   def initialize
     arg_definitions = [
       [
-        { name: 'script_summary', optional: true, description:'***
-    YAML SETTINGS:
-    card_bags:
-      fresh: bag1
-      duplicates: bag2
-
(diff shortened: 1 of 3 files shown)


Dartellum:
I run 4 collection cases presently and use a macro to do the card stuff. I have 4 different locations each with a collection case. It starts with the first one, moves any not entered into the first case to the next location and repeats.

asechrest:
@Raykyn55 

Thanks for doing the `script_summary` arg. I see you identified yaml settings within that. I got something for you, that you can help me with. Take a look at this file called `base-help.yaml`: https://github.com/rpherbig/dr-scripts/blob/master/data/base-help.yaml

I've got a nice set of comments at the top of the file that show you how to use it. You can also look at some examples. If you add to this file your script and settings, when anyone does a `;card-collector help` in-game, not only will they get the script summary, but they'll get a printout of the yaml settings used in the script.

asechrest:
PS - Here's how that works: https://github.com/rpherbig/dr-scripts/pull/5718

asechrest:
PPS - Here is the code in `dependency` that gives the help output: https://github.com/rpherbig/dr-scripts/blob/bb9cebcccf1909a55095d71d4f957066e89aace6/dependency.lic#L51

Dartellum:
I'd like to see the option for nth level of collection cases. No, not because I have four collections going on. Nope, certainly not that. Nothing to see here, move along. :smile:

Raykyn55:
> I'd like to see the option for nth level of collection cases. No, not because I have four collections going on. Nope, certainly not that. Nothing to see here, move along. 😄

I mean, could cycle through additional cases if it doesnt fit in one, not a bad idea tbh. Adding a second would be as hard/easy as adding a twenty-second.

card-collector.lic
```diff
@@ -0,0 +1,42 @@
...
+    settings = get_settings
+    unless @settings.card_bags
+      DRC.message("No bags defined in your yaml, please run this script with argument 'help' for more information.")
+      exit
+    end
+    fresh = settings.card_bags['fresh']
+    duplicates = settings.card_bags['duplicates']
+
+    DRCI.stow_hands
+    cards_add(fresh,duplicates)
```
asechrest, line 22:
```suggestion
    cards_add(fresh,duplicates)
```
```suggestion
    cards_add(fresh, duplicates)
```

card-collector.lic
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def cards_add(fresh,duplicates)
```
asechrest, line 25:
```suggestion
  def cards_add(fresh, duplicates)
```

data/base-help.yaml
```diff
@@ -917,3 +917,12 @@ water_holder:
...
     carve-bead:
+
+card_bags:
+  description: Bags you use to collect and store cards
+  example: red backpack
+  referenced_by:
+    - card-collector
+  specific_descriptions:
+    fresh: Fresh bag contains unsorted, newly collected cards. This is where script gets cards for sorting.
+    duplicates: Duplicates bag(bag2) holds anything that doesn\'t fit in the collector\'s case. This also holds collector\'s case.
```
asechrest, line 928:
```suggestion
    card-collector:
```

This section is designed to list the scripts that use the setting, and then the specific way in which it uses the setting. Granted, in this case, your setting is designed solely for a single script. You're welcome to leave this section blank if the general description fully encapsulates the use of the setting.

But in the case of other settings, many scripts use them. So we give a top-level general description, and then this `specific-descriptions` section again lists each script that uses the setting, and then allows you to describe the specific way in which it uses it.

See this settings example for `waggle_sets` in `base-help.yaml`: https://github.com/rpherbig/dr-scripts/blob/1cd4361c3e186bcc8d3892377811e94f83a81f43/data/base-help.yaml#L889

asechrest:
Should we add the new setting to `base.yaml` with a description?

After those comments are resolved, this looks good.

Raykyn55:
> Should we add the new setting to `base.yaml` with a description?
> 
> After those comments are resolved, this looks good.

Added blank, so script will still proc the error if they're not filled in.

PR #5971, opened 2022-08-01
base.yaml - added burgle to tarantula no use list

BinuDR:
Nothing should ever pause burgle and a smart pause all was recently added to tarantula.lic, so I added burgle to the tarantula no use script list.


profiles/base.yaml
```diff
@@ -2146,6 +2146,7 @@ tarantula_noun: biomechanical tarantula
 tarantula_startup_delay: 15
 tarantula_skip_alternate: false
 tarantula_no_use_scripts:
+- burgle
```
MahtraDR, line 2149:
```suggestion
  - burgle
```

Issue #5974, opened 2022-08-02
base-spells successful cast


Raykyn55:
So as it stands, successful casts are caught with the phrase
```
^You gesture
```
What does this mean?
- Nothing below that line hits the match table.

Why is this a problem?
- backfires, low attunement messages, other failures don't register.

What does this affect?
- Buff can backfire a critical buff, but buff doesn't know, just carries on like it worked.
- if you fail a symbiosis cast, your scripts think it was successful, don't realize you still have symb up, and if your next spell isn't set up for symbiosis, you're gonna have a bad time.

How can this be fixed?
- I can see two options for sorting this.
  1. Spell cast messaging filled in for all new, and replacement cast messages.
  2. reget for a couple lines, and if we don't catch failure, we assume success.

2 is probably the easier, though less robust method, but it looks like 1 was the goal initially. This would have to be completed in its entirety before removing You gesture from the match table.

Tirost:
Scripting for Genie, I had tried something similar to solution one - and it worked well, until I was in a room with someone else using the same spells. "You" must be in the successful cast regex if possible; otherwise, other people in the same area might trigger it.

Ultimately, I went back to "^You gesture" to determine most successful casts, and had to rely on another method for catching backfires.

Raykyn55:
> Scripting for Genie, I had tried something similar to solution one - and it worked well, until I was in a room with someone else using the same spells. "You" must be in the successful cast regex if possible; otherwise, other people in the same area might trigger it.
> 
> Ultimately, I went back to "^You gesture" to determine most successful casts, and had to rely on another method for catching backfires.

Right now, for personal use, I'm using a tiny passive loop with a flag for failure. That might be option 3.
```ruby
Flags.add('failed-spell', 'You strain, but are too mentally fatigued', 'amount of mana slip away from you')
no_pause_all
no_kill_all

loop do
    until Flags['failed-spell']
        pause 0.25
    end
    Flags.reset('failed-spell')
    fput('release mana')
end
```
work that into CT and buff, perhaps.

Raykyn55:
oh, think i found where i can add this.

Raykyn55:
the common method already checks for failure flags on symb casts, but doesn't check for low attunement failure, so wouldn't catch that and wouldn't dismiss symb for that failure.

PR #5975, opened 2022-08-02
[script][common-arcana] Add low attunement failure to Flag['spell-fail']

Raykyn55:
My previous fix for this problem wasn't effective because the match for this messaging was being hijacked by the match for You gesture, which was occurring a line above. Backfires and the like are handled with a flag already, adding this messaging should give it the same release treatment.

```
[buff]>cast
You gesture.
You strain, but are too mentally fatigued to finish the pattern, and it slips away.
>
[buff]>release mana
```


Raykyn55:
resolves https://github.com/rpherbig/dr-scripts/issues/5974

PR #5977, opened 2022-08-02
[data] [base-hunting.yaml] Update Maiden Tresses

Hiinky:
TT127 bolstered spawn and split the area into Maiden Tresses and slightly harder Matron Tresses.


asechrest:
Thanks, Kali!

PR #5983, opened 2022-08-06
[scripts] [equipmanager] Add prefixes, remove includes [1 of 2]

MahtraDR:
As per title.


MahtraDR:
Note this keeps `include DRC` for now, as I can't seem to get rid of needing that for `Item.new`

Issue #5988, opened 2022-08-08
carve-lockpicks.lic: missing messaging


mwest152:
[carve-lockpicks]>put my lockpick in my wristcuff in my backpack

**You put a grandmaster's iron lockpick in your lockpick wristcuff.**

[carve-lockpicks: *** No match was found after 15 seconds, dumping info]

[carve-lockpicks: messages seen length: 26]

[carve-lockpicks: message: You put a grandmaster's iron lockpick in your lockpick wristcuff.]

[carve-lockpicks: for command put my lockpick in my wristcuff in my backpack]

mwest152:
=begin
	Need these settings in your yaml:
    #Header
    Lockpick_carve_settings:
      grand_container: #bag for complete grandmaster's lockpicks
      master_container: #bag for completed master's lockpicks
      trash_container: #bag for anything less than master/grandmaster
      initial_grand: #true/false as to whether to initial your grandmasters' lockpicks
=end

custom_require.call(%w[common common-items common-crafting])

class CarveLockpicks

  def initialize
    @settings = get_settings
    @bag = @settings.crafting_container
    @bag_items = @settings.crafting_items_in_container
    @belt = @settings.engineering_belt
    @lockpick_carve_settings = @settings.lockpick_carve_settings
	  @grandbag = @lockpick_carve_settings['grand_container']
	  @masterbag = @lockpick_carve_settings['master_container']
	  @trashbag = @lockpick_carve_settings['trash_container']
	  @initial = @lockpick_carve_settings['initial_grand']

    DRC.wait_for_script_to_complete('buff', ['carve'])
    DRC.bput('sit', 'You sit', 'You are already sitting', 'You rise', 'While swimming?') unless sitting?
    DRCC.get_crafting_item("carving knife", @bag, @bag_items, @belt)
    if DRCI.exists?("keyblank pocket")
      get_keyblank
    else
      DRCC.stow_crafting_item("carving knife", @bag, @belt)
      exit
	end
  end


  def get_keyblank
    case DRC.bput "get keyblank from my keyblank pocket", 'You get', 'What were you referring to?'
    when /What were you referring to?/
      case DRC.bput "count my keyblank pocket", /There's nothing inside the keyblank pocket/, /It looks like there/, /I could not find what you were referring to./
      when /There's nothing inside the keyblank pocket/
        empty_pocket
      when /It looks like there/
        DRC.bput "open my keyblank pocket", 'You open a'
[...]

PR #5989, opened 2022-08-08
[script][carve-lockpicks] \shorten match phrase

Raykyn55:


asechrest:
Addresses #5988

asechrest:
I've never looked at this script before. Couple things.

1. I hate the way these `bput`s are structured. Can we use convention of `bput` call followed by parentheses?
2. Can we use regex instead of string matches?
3. Once converted to regex, these are now unanchored, incredibly short regex matches, which is a recipe for false matches. Let's try to anchor them and retain enough of the words in our intended match to guard against false matches.

This is just off the cuff but:

`/^You put (\w+) lockpick/`

  vs.

`/You put/`

Raykyn55:
> 

I'll add this to the list for rewrites then. Works fine as is unless you're putting picks in a bag inside a bag, so no urgency.

PR #5990, opened 2022-08-08
Crafting hometown override

Raykyn55:
Simple enough change, if a bit wide-ranging. Idea here is to let folks pick a different town to do their crafting in, if the specific town doesn't support some or all of the crafts, or if the town is getting a little crowded (eg crossing)

Draws from the same information as hometown, only applies to the crafting scripts.

Change is pretty straightforward:
```ruby
```
becomes
```ruby
@hometown = @settings.crafting_hometown || @settings.hometown
```
chooses your crafting hometown first. If undefined (as in base) sets it to your normal hometown. Yaml setting as follow:
```yaml
crafting_hometown: Shard
```

### DIFF
--- a/carve.lic
+++ b/carve.lic
@@ -11,7 +11,7 @@ class Carve
 
   def initialize
     @settings = get_settings
-    @hometown = @settings.crafting_hometown || @settings.hometown
+    @hometown = @settings.crafting_hometown['engineering'] || @settings.crafting_hometown || @settings.hometown
     @bag = @settings.crafting_container
     @bag_items = @settings.crafting_items_in_container
     @belt = @settings.engineering_belt
--- a/carve.lic
+++ b/carve.lic
@@ -11,6 +11,7 @@ class Carve
 
   def initialize
     @settings = get_settings
+    @hometown = @settings.crafting_hometown || @settings.hometown
     @bag = @settings.crafting_container
     @bag_items = @settings.crafting_items_in_container
     @belt = @settings.engineering_belt
--- a/craft.lic
+++ b/craft.lic
@@ -15,7 +15,7 @@ class Craft
     args = parse_args(arg_definitions)
 
     @settings = get_settings
-    @hometown = @settings.hometown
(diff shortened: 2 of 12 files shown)


MahtraDR:
Are there any scenarios where I might want a different hometown for just one craft? I can see wanting to go from Shard to Crossing to rent a forge, but not want to go to Crossing for tailoring. 

Could we make these per craft?

Raykyn55:
> Are there any scenarios where I might want a different hometown for just one craft? I can see wanting to go from Shard to Crossing to rent a forge, but not want to go to Crossing for tailoring.
> 
> Could we make these per craft?

We could make it a hash:
```yaml
crafting_hometown(s):
  Forging: Crossing
  Outfitting: Shard
  etc
```
Workorders will require a little more work, but def doable. Think that's preferable?

MahtraDR:
> > Are there any scenarios where I might want a different hometown for just one craft? I can see wanting to go from Shard to Crossing to rent a forge, but not want to go to Crossing for tailoring.
> > Could we make these per craft?
> 
> We could make it a hash:
> 
> ```yaml
> crafting_hometown(s):
>   Forging: Crossing
>   Outfitting: Shard
>   etc
> ```
> 
> Workorders will require a little more work, but def doable. Think that's preferable?

Sorry, didn't get to these sooner.

Yes, that sounds better to me. Thanks!

Raykyn55:
New town by town yaml pattern option:
```yaml
crafting_hometown:
  engineering: Crossing
  forging: Crossing
  alchemy: Crossing
  outfitting: Shard
  enchanting: Fang Cove
```
or, if you just want one hometown for all of your crafts:
```yaml
crafting_hometown: Shard
```
If you intend to go town by town, you need to define every craft you intend to use. If you have:
```yaml
crafting_hometown:
  engineering: Shard
```
and attempt to run an enchanting script, you will error out. In this case, you'd just do:
```yaml
crafting_hometown: Shard
```
The idea is to allow folks to designate a town for each craft, or one town for all crafts, or ignore this setting and just use their hometown.

Issue #5987, opened 2022-08-08
Pick.lic: missing messaging


mwest152:
You get a rotting mahogany chest from inside your tapestry carpetbag.
s!> 
[pick]>glance my mahogany chest

It looks like there is 1 trap left on the chest.
Looking more closely you see...
Whoever the alchemist was that made this trap deserves to be gutted.  The clay and vial can only mean this is the infamous shockwave trap!

It looks like there is 1 lock left on the chest.
Looking more closely you see...
A sturdy block of metal, the hole for the key is very wide.  This big lock obviously is meant for a correspondingly large key.
s!> 
[pick]>disarm my mahogany chest identify

**You're in no shape to be disarming anything!**

[pick: *** No match was found after 15 seconds, dumping info]

[pick: messages seen length: 16]

**[pick: message: You're in no shape to be disarming anything!]**

[...]

PR #5991, opened 2022-08-09
[scripts] [common-items] Add match for hiro bearskin gloves

MahtraDR:
As per title. Weird messaging on wear for this item.


MahtraDR:
Addresses part of https://github.com/rpherbig/dr-scripts/issues/5987

PR #5995, opened 2022-08-09
[scrip][common-arcana] Rework magic routine for new(old) yaml pattern

Raykyn55:
So as things stand, yaml settings for spells in waggle-sets, buffs, etc, are all set up one way:
```yaml
  Membrach's Greed:
    mana: 20
    cambrinth: [25,25]
```
while training spells are organized based on skill:
```yaml
  Augmentation:
    abbrev: meg
    symbiosis: true
    mana: 17
```
Normally no biggie, but the problem comes in when you're using a sorcery spell. Common-arcana grabs SOME information from your yaml, and the REST from base-spells, but since the magic routine is designed to just pick up where you left off if you have a spell prepped, it doesn't use your yaml for the cast. It searches base-spells for information on the spell you have prepped, and if*(edit) theres a mismatch, it doesn't find anything, and doesn't cast. 

Specifically, here:
```ruby
spell = XMLData.prepared_spell
data = training_spells[skill]
skill = get_data('spells').spell_data[spell]['skill']
```
So we get a spell name, lets say Aura of Tongues, a popular sorcery spell, which is actually a utility spell. AoT is prepped, so spell = Aura of Tongues. Then we go into our yaml and get the skill for that spell, this would be Sorcery, if you're casting it sorcerously. So data would contain all our information from that entry from our yaml (abbrev, symb, mana). Then we go to base-spells and look for a match of THAT spell (aot) with our YAML skill (sorcery) and find...nothing.

This went unnoticed for some time due to the fact that if you have a spell like AoT as a sorcery spell, and ALSO have a block for utility, it uses the AOT  spell for the prep, but the Utility block for the cast. I noticed this when using HW for utility on my empath, because AoT kept trying to cast skin. When I commented out the utility block (HT sungobund on discord), magic routine didn't even prep it.

Changing how we write training spells in our yamls is a significant change, but this brings it in line with how other spell groups are handled. This would allow for anchors to work with training spell lists.

### DIFF
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -815,15 +815,21 @@ module DRCA
   end
 
   def crafting_magic_routine(settings)
-    training_spells = settings.crafting_training_spells
+    # declare array and hash, respectively, unless already populated
+    spells ||= []
+    skill_to_spell ||= {}
+    # get spells and settings, push a list of spell names into spells array
+    # retains information in training_spells for later use
+    training_spells = get_settings.crafting_training_spells.each { |k,v| spells << k } 
+    # take spells array and make it a hash, with keys for skills learned based on yaml settings so we can target spell based on skill needing training
+    spells.each { |x| skill_to_spell.store(training_spells[x]['skill'], x) } 
 
     return if training_spells.empty?
     return if DRStats.mana <= settings.waggle_spells_mana_threshold
 
     if !XMLData.prepared_spell.eql?('None') && checkcastrt == 0
       spell = XMLData.prepared_spell
-      skill = get_data('spells').spell_data[spell]['skill']
-      data = training_spells[skill]
+      data = training_spells[spell]
       crafting_cast_spell(data, settings)
     end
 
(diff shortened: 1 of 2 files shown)


Dartellum:
So, we would have:
```yaml
  Membrach's Greed:
    skill: Sorcery
    mana: 20
    cambrinth: [25,25]
```
:question:

Raykyn55:
> So, we would have:
> 
> ```yaml
> training_spells:
>   Membrach's Greed:
>     skill: Sorcery
>     mana: 20
>     cambrinth: [25,25]
> ```
> 
> ❓

Correct. Would function like wagglesets does now, with the addition of a required skill field to overwrite the one in base spells for sorcery.

Raykyn55:
Changes to magic-training are myriad
1. change handling of yaml information to bring it in line with the new pattern
2. simplified methods, removed unnecessary load_settings method
3. Now passes args/default list to train_magics? method, instead of using instance variable.
4. settings reloads are done after each round of training, and basically all of our work is in one method, so now that method gets the settings, and passes as needed to health method. Otherwise, keeps things local.

PR #6015, opened 2022-08-15
base-hunting - typo fix for malchata zone

BinuDR:
Zone was labeled `ssgui_malchata`
Now it's `ssugi_malchata`


Kiriawen:
shouldn't we have done an anchor/alias?  incase people were using the incorrectly spelled label?
we've done it here already
image

asechrest:
Fair point.

BinuDR:
Yeah, I was thinking about that, but I decided to let people learn the hard way instead!  As Kssarh would say, "Fly, bitch!"

I figured not many people use that zone and also, they should learn to use `fz` to find zones.

PR #6016, opened 2022-08-15
[data][base-hunting] - Plug in backward compatibility after Binu's typo fix

asechrest:
Adding backward compat in case people are using the typo.


asechrest:
Self merging this simple change so as not to screw folks up.

PR #6018, opened 2022-08-15
[script][combat-trainer] Safe Sorcery

Raykyn55:
Allows yaml setting:
```yaml
safe_sorcery: true
```
to prevent combat-trainer from stowing your hands for a sorcery cast. Defaults to false. This is a setting for those who are certain their sorcerous casts won't result in backlash over tier 0, and thus won't lose a hand, drop a weapon, etc..

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1188,6 +1188,9 @@ class SpellProcess
     @runestone_storage = settings.runestone_storage
     echo("  @runestone_storage: #{@runestone_storage}") if $debug_mode_ct
 
+    @safe_sorcery = settings.safe_sorcery
+    echo("  @safe_sorcery: #{@safe_sorcery}") if $debug_mode_ct
+
     @perc_health_timer = Time.now
     @aura_timer = 0
     @avtalia_timer = Time.now - 290
@@ -1594,6 +1597,7 @@ class SpellProcess
   end
 
   def spell_is_sorcery?(spell_data)
+    return false if @safe_sorcery
     return true  if spell_data['sorcery']
     return false unless spell_data['mana_type']
     return false if DRStats.native_mana.casecmp(spell_data['mana_type']).zero?
--- a/data/base-help.yaml
+++ b/data/base-help.yaml
@@ -925,16 +925,3 @@ water_holder:
     - carve-bead
   specific_descriptions:
     carve-bead:
-
-card_bags:
-  description: Bags you use to collect and store cards
-  example: red backpack
-  referenced_by:
-    - card-collector
-  specific_descriptions:
-    card-collector: Sets bag to draw cards from (fresh) and bag to place duplicates and your case(duplicates)
-    
-safe_sorcery:
-  description: Prevents combat-trainer from stowing weapons before sorcery casts. Advanced option, edit with maximum caution.
-  referenced_by:
-    - combat-trainer
--- a/data/base-help.yaml
+++ b/data/base-help.yaml
@@ -790,6 +790,11 @@ safe_room:
   specific_descriptions:
     athletics: If specified, walks to your safe room first for athletics training.
 
+safe_sorcery:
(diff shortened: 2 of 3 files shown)


Raykyn55:
```
[combat-trainer]>stow my nightstick

You put your nightstick in your enshrouding clouds.

[combat-trainer]>get my smokewhorl.whip

You get a braided smokewhorl whip with fire whirl rubies set into a long tapered handle from inside your enshrouding clouds.

[combat-trainer]>cast

You gesture.
Your body tingles as the brilliant silver glow fades from around you.
A brilliant silver glow enwreathes your body.
Emotional pain and exhaustion lingers in your mind, even after you regain rational control.

[combat-trainer]>whip

Gained: Sorcery(+2)

The fluorescent pink and amethyst paint on a silver-grey cloud rat guides your attack.
You whip your smokewhorl whip at a silver-grey cloud rat.  The smokewhorl whip lands a hard hit (5/23) to a silver-grey cloud rat's left foreleg.
[You're bruised, winded, incredibly balanced and overwhelming opponent.]
Roundtime: 2 sec.
```

MahtraDR:
We don't need this. If your sorcery is safe, just set the spell's native mana type to your native mana type.

MahtraDR:
> ```yaml
>   Substratum:
>     mana: 5
>     cambrinth: [30,30]
>     symbiosis: false
>     sorcery: true
> ```

> would check first, and mana type wouldn't factor in. Importantly, this is a perfectly valid yaml setting, likely some of our users have this setting. This means that even setting mana_type: life, for a ranger, when I cast this spell as a buff in combat, I will still stow my weapon.

In this scenario, the user explicitly set the `sorcery: true` flag, since that isn't from base-spells.yaml... They could not set that if they don't want the spell to be treated as sorcery...

> I'm wary of situations where we lie to scripts to accomplish functions that are otherwise guarded against. This was the problem we had, in part, with the aim bug. We were telling ct that our weapon was not loaded as a patch for a critter having died during aim. While that particular workaround would often successfully reacquire a target, occasionally it would trigger an aim loop. It seems to me a better solution is giving ct accurate information, and coding it to react to that information, rather than getting the results by tricking the script.

I accept the comment about the aim bug, but this is a very different thing, to me. We already overwrite default spell info, and to me, this is no different. We can maybe document it better.

> Here, we have a special case where users can opt-in to a risky setting, and personally, I think this should be it's own setting, rather than overwriting accurate information with false information to achieve the desired ends.

I disagree with the characterization of this being "false information" it's no different than using cambrinth, or a worn charm. They're just changes to the default spell data everyone uses when casting spells with lich.

Dartellum:
> We don't need this. If your sorcery is safe, just set the spell's native mana type to your native mana type.
> 
> edit: Expanding a bit, the code for this check is very basic.
> 
> ```ruby
>   def spell_is_sorcery?(spell_data)
>     return true  if spell_data['sorcery']
>     return false unless spell_data['mana_type']
>     return false if DRStats.native_mana.casecmp(spell_data['mana_type']).zero?
>     return false if spell_data['mana_type'].casecmp('ap').zero?
>     true
>   end
> ```
> 
> Just set the mana type of the spell to be your native mana type. We don't need additional code, or settings for this to work.

I am not sure I follow this - also past midnight presently - but why would you change mana type to native type? I cast many sorcery spells in combat and out of combat. The only spell I found troublesome is HOT as it destroys the left hand so swapping armor, weapons, and picking up stuff is problematic. I cast refresh, Eagle's Cry, Bless, Clear Vision, Protection, DRUM, etc. I do not change the mana type for them. 

Solution: Do not add the `sorcery: true` to the spells. I do not have that on any of my sorcery spells. I do have `skill: Sorcery` for the TM spells. But for buffs I do not have that or sorcery:
```yaml
refresh: &refresh
  Refresh:
    mana: 0
    prep_type: invoke
    symbiosis: true
    runestone_name: quartz runestone
    cambrinth:
    - - 27
 ```

PR #6021, opened 2022-08-16
[script][magic-training] localize variables

Raykyn55:
Additionally, add yaml setting 
```yaml
magic_training_room: #
```
To set a room for training, be it outside justice, or just better mana.

### DIFF
--- a/data/base-help.yaml
+++ b/data/base-help.yaml
@@ -265,6 +265,14 @@ burgle_settings:
   specific_descriptions:
     burgle: "Set the following: room, entry_type, max_priority_mindstate, rope_adjective, hometown, use_lockpick_ring, lockpick_container, max_search_count, retry, loot, loot_container, safe_mode, room_blacklist, item_whitelist, before, after."
 
+card_bags:
+  description: Bags you use to collect and store cards
+  example: red backpack
+  referenced_by:
+    - card-collector
+  specific_descriptions:
+    card-collector: Sets bag to draw cards from (fresh) and bag to place duplicates and your case(duplicates)
+
 circlecheck_prettyprint:
   description: Tells circlecheck to use the pretty print formatting for output.
   example: true/false
@@ -713,6 +721,12 @@ lumber_use_packet:
   specific_descriptions:
     chop-wood:
 
+magic_training_room:
+  description: Designated room for training magic.
+  example: 2866
+  referenced_by:
+    - magic-training
+
 mark_crafted_goods:
   description: Specify whether to mark your crafted goods.
   example: true/false
@@ -925,17 +939,3 @@ water_holder:
     - carve-bead
   specific_descriptions:
     carve-bead:
-
-card_bags:
-  description: Bags you use to collect and store cards
-  example: red backpack
-  referenced_by:
-    - card-collector
-  specific_descriptions:
-    card-collector: Sets bag to draw cards from (fresh) and bag to place duplicates and your case(duplicates)
-    
-magic_training_room:
-  description: Designated room for training magic.
-  example: 2866
-  referenced_by:
-    - magic-training
--- a/data/base-help.yaml
+++ b/data/base-help.yaml
@@ -933,3 +933,9 @@ card_bags:
     - card-collector
   specific_descriptions:
     card-collector: Sets bag to draw cards from (fresh) and bag to place duplicates and your case(duplicates)
+    
+magic_training_room:
+  description: Designated room for training magic.
+  example: 2866
+  referenced_by:
+    - magic-training
--- a/magic-training.lic
+++ b/magic-training.lic
@@ -47,7 +47,7 @@ class MagicTraining
     DRCT.walk_to(settings.magic_training_room)
     exp_threshold = settings.magic_exp_training_max_threshold
     force_cambrinth = settings.waggle_force_cambrinth
-    training_spells = settings.training_spells
+    training_spells = settings.training_spells || settings.magic_training
 
     magic_skills = training_spells.keys
     skills_to_train = magic_skills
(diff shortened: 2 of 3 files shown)


magic-training.lic
```diff
@@ -18,58 +18,42 @@ class MagicTraining
...
         { name: 'max_time', regex: /\d+/, optional: true, description: 'Max time, in minutes, spent training. (Optional)' },
-        { name: 'training_skills', regex: /^((Augmentation|Warding|Utility|Sorcery)((\s*,\s*)(Augmentation|Warding|Utility|Sorcery))*)$/i, optional: true, description: 'Override which magic skills to train, as a comma-separated list. (Optional)' }
+        { name: 'training_skills', regex: /^((Augmentation|Warding|Utility|Sorcery)((\s*,\s*)(Augmentation|Warding|Utility|Sorcery))*)$/i, optional: true, description: 'Override which magic skills to train, as a comma-separated list. (Optional)' },
+        { name: 'script_summary', optional: true, description:'Trains magic skills defined in Yaml setting training_skills' }
       ]
     ]
 
     args = parse_args(arg_definitions)
+    max_time = ((args.max_time || 60).to_i * 60) # convert to seconds
+    training_skills = args.training_skills ? args.training_skills.split(',').map { |skill| skill.strip.downcase.capitalize } : %w[Utility Warding Augmentation Sorcery]
```
asechrest, line 28: While unlikely to happen, this no longer protects against `args.training` skills being empty, since empty is "truthy" in Ruby.

Raykyn55: Defined is truthy, but args not sent are nil. Been using this version for a while.

magic-training.lic
```diff
@@ -18,58 +18,42 @@ class MagicTraining
...
-
-    @health_threshold = @settings.health_threshold # vitaltiy percentage
-    @saferoom_health_threshold = @settings.saferoom_health_threshold # wound severity threshold
-  end
-
-  def train_magics?
-    load_settings # This is done purposely so you can adjust settings on the fly, it's great for finding mana values
+  def train_magics?(training_skills)
+    settings = get_settings
+    DRCT.walk_to(settings.magic_training_room)
```
asechrest, line 47: I believe `common-travel` needs to be added to `custom_require`.

magic-training.lic
Raykyn55:
```ruby
class CaseTest

    def initialize

        arg_definitions = [
            [
            { name: 'laminate', regex: /^laminate/i, optional: true, description: 'Optional setting to skip restocking con' },
            { name: 'lighten', regex: /^lighten/i, optional: true, description: 'Optional setting to skip restocking con' },
            { name: 'temper', regex: /^temper/i, optional: true, description: 'Optional setting to skip restocking con' },
            { name: 'pickle', regex: /^pickle/i, optional: true, description: 'Optional setting to skip restocking con' }
            ]
        ]
        training_skills = args.laminate ? args.laminate.split(',').map { |skill| skill.strip.downcase.capitalize } : %w[Utility Warding Augmentation Sorcery]
        echo(training_skills)
        exit

      
    end
end

CaseTest.new
```
>;casetest
--- Lich: casetest active.
[casetest: ["Utility", "Warding", "Augmentation", "Sorcery"]]
--- Lich: casetest has exited.
>

>;casetest laminate
--- Lich: casetest active.
[casetest: ["Laminate"]]
```
Defining it, even with an empty definition re: your test, is what gives it "value" as true. Since parse_args doesn't define, only match, unpassed args remain undefined, and thus false.

Comment (asechrest) [reply magic-training.lic:28]: I _think_ that's fair. I was essentially just noting that prior behavior checked for nil _and_ empty. Whether that was a superfluous check or not, I wasn' tsure.

Comment (Raykyn55) [reply magic-training.lic:28]: > I _think_ that's fair. I was essentially just noting that prior behavior checked for nil _and_ empty. Whether that was a superfluous check or not, I wasn' tsure.

It cant be empty in this case. But supposing you could, that would be saying "I want you to train nothing". Which is the behavior you would get.

File: data/base-help.yaml [hunk 3]
```diff
@@ -933,3 +933,9 @@ card_bags:
     - card-collector
   specific_descriptions:
     card-collector: Sets bag to draw cards from (fresh) and bag to place duplicates and your case(duplicates)
+    
+magic_training_room:
```
asechrest, line 937: Could you throw this setting alphabetically in this file? I missed that card-collector was put at the end, too. We can PR separately for that. This helps searching within the file itself, but it also allows a `;script-name help` to list the settings used alphabetically as well. Just keeps it kind of ordered. TY.

magic-training.lic
asechrest, line 47: What happens if `magic_training_room` isn't specified? Real question, don't know how `DRCT.walk_to("")` responds.

data/base-help.yaml
Raykyn55: Yes, apologies, i didnt realize they were alphabetical. Ill get it when i get back. Sorry about that

magic-training.lic
Raykyn55: Checked that, it just returns, doesnt even open go2

Raykyn55: Return unless room, iirc

data/base-help.yaml
asechrest: > Yes, apologies, i didnt realize they were alphabetical. Ill get it when i get back. Sorry about that

No apology needed. It's a new thing, so it's ok.

magic-training.lic
asechrest: > 

I don't use this script, even historically. What was prior behavior. Did it just train in the room you were standing in?

Raykyn55: Yes.

data/base-help.yaml
MahtraDR: Is this comment resolved?

magic-training.lic
MahtraDR: Is this comment resolved?

PR #6029, opened 2022-08-18
[profile][base] Adding variables for Fang Cove/alternate hometown fixes

Smaptastic:
Fang Cove has issues with several scripts. The revised versions of these scripts read the new variables and treat a city other than hometown: as their hometown, if set.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -2,6 +2,18 @@
 # See https://github.com/rpherbig/dr-scripts/wiki/YAML-Settings for documentation
 
 hometown: Crossing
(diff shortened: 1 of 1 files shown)


MahtraDR:
Don't do this. Instead, for the PRs that add this functionality in the other scripts, add these changes (the relevant ones per script) into those PRs, please.

Smaptastic:
> Don't do this. Instead, for the PRs that add this functionality in the other scripts, add these changes (the relevant ones per script) into those PRs, please.

Sorry about that. I did warn you that I'd do something stupid in my first attempt at making commits.

PR #6032, opened 2022-08-18
[script][crossing-repair] Fang Cove fix - crossing-repair

Smaptastic:
Applies the variables from PR #6029 to crossing-repair.lic, causing it to be able to use fang_cove_override_town

### DIFF
--- a/almanac.lic
+++ b/almanac.lic
@@ -36,8 +36,11 @@ class Almanac
     end
 
     # Pause scripts to prevent interference
-    scripts_to_unpause = []
-    scripts_to_unpause = DRC.smart_pause_all
+    until scripts_to_unpause = DRC.safe_pause_list
+      echo("Cannot pause, trying again in 30 seconds.")
+      pause 30
+    end
+
     waitrt?
 
     # Wait for any last output from paused scripts to resolve
@@ -53,11 +56,11 @@ class Almanac
     unless DRCI.get_item_if_not_held?(@almanac) && DRCI.in_hands?(@almanac)
       if DRCI.exists?(@almanac)
         DRC.message('Hands full, will try again later')
-        DRC.unpause_all_list(scripts_to_unpause)
+        DRC.safe_unpause_list(scripts_to_unpause)
         return
       else
         DRC.message('Almanac not found, exiting')
-        DRC.unpause_all_list(scripts_to_unpause)
+        DRC.safe_unpause_list(scripts_to_unpause)
         exit
       end
     end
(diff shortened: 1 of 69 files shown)


MahtraDR:
This one I don't understand. If FC is your hometown, you can already override that with `force_repair_town` What's this toggle adding that doesn't exist today? Thanks!

Smaptastic:
> This one I don't understand. If FC is your hometown, you can already override that with `force_repair_town` What's this toggle adding that doesn't exist today? Thanks!
fang_cove_override_town toggles the town to Fang Cove for multiple scripts that do not work in Fang Cove, all at once. Setting it sets athletics, burgle, crossing-repair, checkfavors, and favor to Fang Cove, all at once.

Raykyn55:
What is the advantage to using fang cove as your hometown?

Smaptastic:
> What is the advantage to using fang cove as your hometown?

Availability from anywhere. No need to change safe rooms when moving to a new hunting zone/town. Public crafting areas aren't crowded. Healer is better (though I know you can just force it as a heal town, I like having everything set up in one place as much as possible).

Raykyn55:
might need updates in:
mining-buddy
forestry-buddy
herb-stock
restock

Could also run into issues where scripts are writing @hometown variable using hometown: x, replacing your custom hometown. Can't think of an example offhand, but something to look out for.

Smaptastic:
> might need updates in: mining-buddy forestry-buddy herb-stock restock

I don't use those so I didn't add them in, but I'm happy to do so. I'll try to get the PR for them in tonight.

> Could also run into issues where scripts are writing @hometown variable using hometown: x, replacing your custom hometown. Can't think of an example offhand, but something to look out for.

A couple of the scripts in this batch did that. See checkfavors. I was able to solve that issue.

Raykyn55:
line 65:
```ruby
@hometown_data = get_data('town')[settings.hometown]
```
Could present a problem.

Smaptastic:
> line 65:
> 
> ```ruby
> @hometown_data = get_data('town')[settings.hometown]
> ```
> 
> Could present a problem.

I THINK that only kicks in if force_repair_town and fang_cove_override_town are blank. In the original, force_repair_town worked properly when set, and it doesn't appear there. I think these lines basically change settings.hometown to one of the above:

```town = DRC.get_town_name(args.town || settings.force_repair_town || settings.fang_cove_override_town) (line 21)
settings['hometown'] = town if town (line 37)```

I can tell you that I've been running this for several repair cycles (hometown: Fang Cove, fang_cove_override_town: Crossing) and it has worked as expected.

**Comment by Raykyn55** [comment]
> > line 65:
> > ```ruby
> > @hometown_data = get_data('town')[settings.hometown]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could present a problem.
> 
> I THINK that only kicks in if force_repair_town and fang_cove_override_town are blank. In the original, force_repair_town worked properly when set, and it doesn't appear there. I think these lines basically change settings.hometown to one of the above:
> 
> ```
> town = DRC.get_town_name(args.town || settings.force_repair_town || settings.fang_cove_override_town) (line 21)
> 
> settings['hometown'] = town if town (line 37)
> ```
> 
> I can tell you that I've been running this for several repair cycles (hometown: Fang Cove, fang_cove_override_town: Crossing) and it has worked as expected.

It fetches information like currency type, among other things. Its not an alternative to hometown, its drawing information about your hometown directly from base-town.

**Comment by Smaptastic** [comment]
If that’s the case, force_repair_town never worked as advertised, since all I’m doing here is adding another variable with the exact same function. 

In practice, it gets currency and works fine. Give it a test run.

**Comment by Raykyn55** [comment]
> If that’s the case, force_repair_town never worked as advertised, since all I’m doing here is adding another variable with the exact same function.
> 
> In practice, it gets currency and works fine. Give it a test run.

You are right, I didn't read 37 correctly. I wonder if that wouldn't be more appropriate in the original town definition, appended after your override variable?
```ruby
town = DRC.get_town_name(args.town || settings.force_repair_town || settings.fang_cove_override_town || settings.hometown)
```
then @hometown_data would be:
```ruby
@hometown_data = get_data('town')[town]
```

Smaptastic:
That seems like one way to do it, sure. It's more similar to the way other scripts did it. But I don't know enough about this stuff to make significant modifications like that, so all I was doing was appending a variable (fang_cove_override_town) to a list that contained an override that I already knew worked (force_repair_town).

I think the function would be identical either way, but I don't know enough to say that with any degree of certainty.

PR #6033, opened 2022-08-18
[script][checkfavors] Fang Cove fix - checkfavors

Smaptastic:
checkfavors did not work properly with Fang Cove as hometown. Added fang_cove_override_town as a multi-script town override, and favor_town as a checkfavors-specific town override.

### DIFF
--- a/checkfavors.lic
+++ b/checkfavors.lic
@@ -9,7 +9,7 @@ class CheckFavors
   def initialize
     @settings = get_settings
     town_data = get_data('town')
-    @character_hometown = @settings.favor_hometown || @settings.fang_cove_override_town || @settings.hometown
+    @character_hometown = @settings.favor_town || @settings.fang_cove_override_town || @settings.hometown
     @hometown = town_data[@character_hometown]
     check_favors
   end
--- a/checkfavors.lic
+++ b/checkfavors.lic
@@ -9,7 +9,8 @@ class CheckFavors
   def initialize
     @settings = get_settings
     town_data = get_data('town')
-    @hometown = town_data[@settings.hometown]
+    @character_hometown = @settings.favor_hometown || @settings.fang_cove_override_town || @settings.hometown
+    @hometown = town_data[@character_hometown]
     check_favors
   end
 
--- a/checkfavors.lic
+++ b/checkfavors.lic
@@ -9,8 +9,8 @@ class CheckFavors
   def initialize
     @settings = get_settings
     town_data = get_data('town')
-    @character_hometown = @settings.favor_town || @settings.fang_cove_override_town || @settings.hometown
-    @hometown = town_data[@character_hometown]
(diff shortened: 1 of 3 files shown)


checkfavors.lic
```diff
@@ -9,7 +9,8 @@ class CheckFavors
   def initialize
     @settings = get_settings
     town_data = get_data('town')
-    @hometown = town_data[@settings.hometown]
+    @character_hometown = @settings.favor_town || @settings.fang_cove_override_town || @settings.hometown
```
Raykyn55, line 12: If this is the only use of this variable  (character_hometown) it need not be an instance variable. treat it like town_data. In fact, with a script this small, you can pass all of this, and none of them need be an instance variable.  That would eliminate any chance of interactions between scripts. See recent changes to magic-training for examples.

checkfavors.lic
```diff
@@ -9,7 +9,8 @@ class CheckFavors
...
+    @hometown = town_data[@character_hometown]
     check_favors
```
Raykyn55, line 14: this would become
```ruby
check_favors(settings, hometown)
```
and the receiving method:
```ruby
def check_favors(settings, hometown)
```

Smaptastic: > this would become
> 
> ```ruby
> check_favors(settings, hometown)
> ```
> 
> and the receiving method:
> 
> ```ruby
> def check_favors(settings, hometown)
> ```

I don't fully understand. It sets @character_hometown to the first of favor_town, fang_cove_override_town, and hometown. It then sets @hometown to town_data[@character_hometown]. The only time a hometown is used in the execution is on line 27: 

`DRCT.walk_to(@hometown['favor_altar']['id'])`

Raykyn55: The point is that @hometown is accessible in other scripts, hometown is not. But that also means its not accessible outside that method, so you have to pass it. Making it a local variable eliminates interaction outside this script.

checkfavors.lic
Smaptastic: Re: character_hometown, I pulled that trick from another script with a similar use of @hometown and an existing override. I think it was sell-loot.

checkfavors.lic
Smaptastic: It is @hometown in the original script. I wasn’t looking to alter anything about the original that wasn’t necessary to add in my variables.

checkfavors.lic
Raykyn55: It always works until it doesnt, at which point a lot of people become upset, and suddenly youre scrambling to figure out why. This suggestion is one way to hedge against that.

Raykyn55: I'm wrong about this. carry on! Well, not about the scrambling, but def the variables.

Smaptastic: Changed @character_hometown to a local variable.

PR #6034, opened 2022-08-18
[script][favor] Fang Cove fix - favor

Smaptastic:
favor did not work properly with Fang Cove as hometown. Added fang_cove_override_town as a multi-script town override, and favor_town as a favor-specific town override.


asechrest:
Thanks for adding the settings to `base-help.yaml`.

PR #6042, opened 2022-08-20
[script][workorders] Crafting hometown override part 2

Raykyn55:
Lets try this again. Problem with the original was craft method wasn't transferred for tailoring.

Additionally, some pieces of a different PR crept in.


Smaptastic:
I think maybe removing the town settings (engineering: Crossing, etc.) in base.yaml would be best so people who don't use this in their setup yaml don't end up running all over the world to craft. Assuming I understand properly how that works.

Raykyn55:
> I think maybe removing the town settings (engineering: Crossing, etc.) in base.yaml would be best so people who don't use this in their setup yaml don't end up running all over the world to craft. Assuming I understand properly how that works.
> 
> Edit: I don't mean removing them, sorry. Just leaving them blank rather than having a town pre-set.

Theyre commented out in base, so inoperable by default.

Smaptastic:
Ah shoot. Was looking for the hashtag after the section (engineering: # Crossing), not before. My bad.

Raykyn55:
> Ah shoot. Was looking for the hashtag after the section (engineering: # Crossing), not before. My bad.

np, more eyes is always better on this stuff, even if it's a whiff.

PR #6044, opened 2022-08-20
[script][buff-watcher] New pause, local variables

Raykyn55:
Updated to remove unnecessary instance variables, to forestall overlaps with other scripts.
Uses new safe_pause_list/safe_unpause_list
requires https://github.com/rpherbig/dr-scripts/pull/6043


Dartellum:
Testing with #6043. I changed the echo to DRC.message in 6043. Need to see the message and not lose it in a sea of text.

PR #6046, opened 2022-08-20
[script][pattern-hues] Refactored (removed include) and added new pause method.

Dartellum:
Removed include. Added `base-help section`. May need help fixing the settings to list better. 
Added new pause routine safe_pause_list/safe_unpause_list
requires https://github.com/rpherbig/dr-scripts/pull/6043

### DIFF
--- a/pattern-hues.lic
+++ b/pattern-hues.lic
@@ -5,7 +5,6 @@
 custom_require.call(%w(common))
 
 class PatternHues
-  include DRC
 
   def initialize
     arg_definitions = [
@@ -56,14 +55,14 @@ class PatternHues
       /You should stop practicing/,
       /You find it hard to concentrate on anything/
     ]
-    scripts_to_unpause = DRC.smart_pause_all
-    bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
+    pattern_hues_pause_list ||= DRC.safe_pause_list
+    DRC.bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
     if @settings.pattern_hues['gesture'] == "random"
-      bput("gesture #{@adj.sample} #{@color.sample}", *match_messages)
+      DRC.bput("gesture #{@adj.sample} #{@color.sample}", *match_messages)
     else
-      bput("gesture #{@settings.pattern_hues['gesture']}", *match_messages)
+      DRC.bput("gesture #{@settings.pattern_hues['gesture']}", *match_messages)
     end
-    DRC.unpause_all_list(scripts_to_unpause)
+    DRC.safe_unpause_list(pattern_hues_pause_list)
   end
 
   def cast_now
--- a/pattern-hues.lic
+++ b/pattern-hues.lic
@@ -46,10 +46,14 @@ class PatternHues
   end
 
   def cast_pattern_hues
-    match_messages = [
+    match_successes = [
       /You trace the complex sigils of the Pattern Hues cantrip around yourself./,
       /You already have/,
       /You gesture/,
+      /You are now prepared to cast the Pattern Hues cantrip./,
+
+    ]
+    match_failures = [
       /Something in the area is interfering/,
       /You try, but the dolphin choose/,
       /You should stop practicing/,
@@ -57,18 +61,19 @@ class PatternHues
       /you suddenly realize that there is no custom color/,
       /You'll interrupt your research!/
     ]
-    scripts_to_unpause = DRC.safe_pause_list
-    unless scripts_to_unpause
+
+    @scripts_to_unpause = DRC.safe_pause_list
+    unless @scripts_to_unpause
       DRC.message("Cannot pause, trying again shortly.")
       return
     end
-    DRC.bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
+    DRC.bput("prep cantrip pattern hues", *match_successes, *match_failures)
     if @settings.pattern_hues['gesture'] == "random"
-      DRC.bput("gesture #{@adj.sample} #{@color.sample}", *match_messages)
+      DRC.bput("gesture #{@adj.sample} #{@color.sample}", *match_successes, *match_failures)
     else
-      DRC.bput("gesture #{@settings.pattern_hues['gesture']}", *match_messages)
+      DRC.bput("gesture #{@settings.pattern_hues['gesture']}", *match_successes, *match_failures)
     end
-    DRC.safe_unpause_list(scripts_to_unpause)
+    DRC.safe_unpause_list(@scripts_to_unpause)
   end
 
   def cast_now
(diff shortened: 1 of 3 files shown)


pattern-hues.lic
```diff
@@ -56,14 +55,14 @@ class PatternHues
...
+    DRC.bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
     if @settings.pattern_hues['gesture'] == "random"
-      bput("gesture #{@adj.sample} #{@color.sample}", *match_messages)
+      DRC.bput("gesture #{@adj.sample} #{@color.sample}", *match_messages)
     else
-      bput("gesture #{@settings.pattern_hues['gesture']}", *match_messages)
+      DRC.bput("gesture #{@settings.pattern_hues['gesture']}", *match_messages)
     end
-    DRC.unpause_all_list(scripts_to_unpause)
+    DRC.safe_unpause_list(pattern_hues_pause_list)
```
MahtraDR, line 65: The new method doesn't work like this, and you have to have a check for the return value of the method.

See https://github.com/rpherbig/dr-scripts/pull/6044

MahtraDR, line 65:
```suggestion
    pattern_hues_pause_list = DRC.safe_unpause_list
    unless pattern_hues_pause_list
      echo("Cannot pause, trying again shortly.")
      return
    end
```

Dartellum: DRC.safe_unpause_list(scripts_to_unpause) is in 6044.

MahtraDR: I'm sorry, I misread.

pattern-hues.lic
```diff
@@ -56,14 +55,14 @@ class PatternHues
       /You should stop practicing/,
       /You find it hard to concentrate on anything/
     ]
-    scripts_to_unpause = DRC.smart_pause_all
-    bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
+    pattern_hues_pause_list ||= DRC.safe_pause_list
```
MahtraDR, line 58:
```suggestion    
    pattern_hues_pause_list = DRC.safe_unpause_list
   
    unless pattern_hues_pause_list
      echo("Cannot pause, trying again shortly.")
      return
    end
```

MahtraDR: This is where I meant to make that suggestion to.

Dartellum: You mean safe_pause_list here.

Dartellum: Need ||= or an error occurs:
```
2022-08-20 01:23:30 -0400:--- Lich: error: wrong number of arguments (given 0, expected 1)
2022-08-20 01:23:30 -0400:	common:784:in `safe_unpause_list'
2022-08-20 01:23:30 -0400:	pattern-hues:69:in `cast_pattern_hues'
```

MahtraDR: Test it out. I am just writing from memory. :) Just pointing out you need to check return values.

Dartellum: Changing name from pattern_hues.... to scripts_to_unpause.

MahtraDR: > 

This error is because I had suggested the wrong method. DRC.safe_pause_list has no arguments.

Dartellum:
Updated it.

Raykyn55:
Testing! Noticed a missing match when using custom, but you don't have any custom colors available.
```
As you begin to trace the complex sigils of the Pattern Hues cantrip around yourself, you suddenly realize that there is no custom color available to you.  With a brief surge of energy, your cantrip fails.
Roundtime: 3 seconds.
```
That's the message you get.

Raykyn55:
```
[pattern-hues]>prep cantrip pattern hues
You need to concentrate on your climbing, not that.
>
[pattern-hues: *** No match was found after 15 seconds, dumping info]
```

Raykyn55:
```
[pattern-hues: Pausing ["t2", "afk", "athletics"] to run pattern-hues]
[pattern-hues]>prep cantrip pattern hues
Are you sure you want to do that?  You'll interrupt your research!
[Enter your command again if you want to.]
```
want a match for that too, I think, perhaps a list of failure messages for a reset

pattern-hues.lic
```diff
@@ -54,16 +53,22 @@ class PatternHues
...
+      /You'll interrupt your research!/
     ]
-    scripts_to_unpause = DRC.smart_pause_all
-    bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
+    scripts_to_unpause = DRC.safe_pause_list
+    unless scripts_to_unpause
+      DRC.message("Cannot pause, trying again shortly.")
+      return
+    end
+    DRC.bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
```
Raykyn55, line 65: I believe this is the bput that we want those missing matches added, and we'll want to stop before gesturing if we get failure, eg climbing, or researching, etc. perhaps a small case statement? when *failure => unpause and go back to the loop, else, gesture x or y.

pattern-hues.lic
```diff
@@ -54,16 +53,22 @@ class PatternHues
...
     if @settings.pattern_hues['gesture'] == "random"
-      bput("gesture #{@adj.sample} #{@color.sample}", *match_messages)
+      DRC.bput("gesture #{@adj.sample} #{@color.sample}", *match_messages)
     else
-      bput("gesture #{@settings.pattern_hues['gesture']}", *match_messages)
+      DRC.bput("gesture #{@settings.pattern_hues['gesture']}", *match_messages)
     end
-    DRC.unpause_all_list(scripts_to_unpause)
+    DRC.safe_unpause_list(scripts_to_unpause)
```
Raykyn55, line 71: If we are returning to the loop on a failure, we might want to move unpause down to the loop.

pattern-hues.lic
```diff
@@ -47,23 +46,34 @@ class PatternHues
...
     ]
-    scripts_to_unpause = DRC.smart_pause_all
-    bput("prep cantrip pattern hues", "You are now prepared to cast the Pattern Hues cantrip.","You already have", "You should stop practicing")
+
+    @scripts_to_unpause = DRC.safe_pause_list
+    unless @scripts_to_unpause
+      DRC.message("Cannot pause, trying again shortly.")
+      return
+    end
+    DRC.bput("prep cantrip pattern hues", *match_successes, *match_failures)
```
Raykyn55, line 70: So this prevents a hang, but it bulls through to the gesture, even if it hits failure. I've tried this:
```ruby
unless /You are now prepared/i =~ DRC.bput("prep cantrip pattern hues", *match_successes, *match_failures)
  DRC.safe_unpause_list(scripts_to_unpause)
  return
end```

Which, as I understand it, accepts only a successful prep, otherwise, unpauses, then returns to wait the 600 second duration before another attempt.

Dartellum: I can review this tonight, maybe.

Dartellum: Since the pattern for successes could contain more than one, I changed to a case statement:
```ruby
case DRC.bput("prep cantrip pattern hues", *match_successes, *match_failures)
    when *match_failures
        if "That is not a cantrip you can cast"
          exit
        else
          DRC.safe_unpause_list(scripts_to_unpause)
          return
        end
    end
    ```
    Tested on one char who does not have pattern hues yet. Will test on Dartellum after his hunt.

Dartellum: @Raykyn55 please try the update and let me know.

MahtraDR:
Closing as abandoned.

PR #6048, opened 2022-08-21
[script][pay-debt] Fang Cove exchange messaging added.

Smaptastic:
Using the exchange in Fang Cove generated messaging unrecognized by pay-debt. This should resolve that.


asechrest:
Thanks for the PR.

PR #6056, opened 2022-08-21
[scripts] [almanac] Switch to safe_pause with mutex

MahtraDR:
As per title.


MahtraDR:
I've tested this on mine, along with wand-watcher, smartlisten, tessera, healer, tarantula, buff-watcher, and burgle.

PR #6058, opened 2022-08-21
[scripts][first-aid] Add scholarship to requisite skill check to break first-aid early.

Smaptastic:
Currently, first-aid ends early if your First Aid mindstate is over 32. However, first-aid is also used to train scholarship. This adds a Scholarship mindstate check in addition to the First Aid mindstate check, both of which must be passed for an early termination.

If preferred, I could create a new script (fa-scholarship or something) out of this instead, if it's determined that editing this would mess too much with people's First Aid training.

### DIFF
--- a/first-aid.lic
+++ b/first-aid.lic
@@ -70,7 +70,7 @@ class FirstAid
             DRC.bput("get my #{@booktype}", 'You get', 'You are already holding')
           end
           waitrt?
-          break if DRSkill.getxp('First Aid') >= 32
+          break if DRSkill.getxp('First Aid') >= 32 && DRSkill.getxp('Scholarship') >= 32
         end
       end
   end


Dartellum:
Scholarship is a side product of reading the first aid material. You could get into a situation where you might read all the charts, if using those, or to a section to hard in a textbook to advance further. You may end up in a loop of `first-aid.lic` because both conditions are not met. .

Smaptastic:
> Scholarship is a side product of reading the first aid material. You could get into a situation where you might read all the charts, if using those, or to a section to hard in a textbook to advance further. You may end up in a loop of `first-aid.lic` because both conditions are not met. .

The conditions only kill first-aid early. They don’t cause a loop if not met. 

Conversely, if using first-aid to train scholarship and first aid in T2, you can get into an infinite-ish loop in its current form, since if your FA is high from hunting, it’ll kill first-aid early without meeting the scholarship necessary to pass that section.

MahtraDR:
I don't like this change as is, on the basis that it changes existing behaviour in a way folks might not expect. Currently, I can say "read 100 charts" and it'll stop reading at 32FA. With this change it'll read a lot longer/more charts. I'd prefer this to be an opt-in, so add a second conditional on a setting in their yaml? (and add said setting into base.yaml and base-help?)

asechrest:
@Smaptastic 

Any response to @MahtraDR 's comment?

Smaptastic:
> @Smaptastic
> 
> Any response to @MahtraDR 's comment?

I wouldn’t mind having it opt-in but I am not confident with my ability to do that.

asechrest:
Would the options mentioned here allow better scholarship training? I don't use this script, so unsure: https://github.com/rpherbig/dr-scripts/wiki/First-Aid-Strategy#advanced-options

asechrest:
Outside of that, an easy way to handle this would be to create an `args` section that allowed you to call this script with a `skill_focus` argument, which might accept `First Aid`, `Scholarship`, or `Both`. The script would then handle accordingly depending on how the user called it.

Smaptastic:
Looks like a good idea with the args. I just have no idea how to do it. Would be happy if someone else wanted to make that edit.

asechrest:
See PR here if you want to see what I implemented: https://github.com/rpherbig/dr-scripts/pull/6116

Smaptastic:
> See PR here if you want to see what I implemented: #6116

Nice. Thanks, that’s a good solution.

PR #6065, opened 2022-08-22
[scripts] [jail-buddy] Switch to safe_pause with mutex

MahtraDR:
As per title, also switched some fputs to common methods.


MahtraDR:
I've tested this on mine, along with almanac, smartlisten, tessera, healer, tarantula, wand-watcher, and burgle.

PR #6067, opened 2022-08-22
[scripts] [tarantula] Switch to safe_pause with mutex

MahtraDR:
As per title.


MahtraDR:
I've tested this on mine, along with almanac, smartlisten, tessera, healer, buff-watcher, wand-watcher, and burgle.

PR #6070, opened 2022-08-22
[script][craft] Updated craft to do T10 and T11 enchanting items. Checks for sigils first as these are hunted and scribed and not store bought.

Dartellum:
Added examples of `;sigilhunter` syntax to hunt for the sigils. Also fixed one enchanting recipe that could use any sigil so made it purchase two from Society.

### DIFF
--- a/craft.lic
+++ b/craft.lic
@@ -389,7 +389,7 @@ class Craft
       DRC.wait_for_script_to_complete('enchant', [6, 'dispel runestone', 'runestone'])
       DRCI.dispose_trash('runestone', @worn_trashcan, @worn_trashcan_verb)
     elsif rank <= 850 # Tier 9 - Intricate
-      Buy a induction and rarefaction sigils and basic runestone
+      # Buy a induction and rarefaction sigils and basic runestone
       DRCM.ensure_copper_on_hand(4000, @settings)
       DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], 1) unless DRCI.exists?("induction sigil-scroll")
       DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], 5) unless DRCI.exists?("rarefaction sigil-scroll")
--- a/craft.lic
+++ b/craft.lic
@@ -293,147 +293,160 @@ class Craft
     rank = DRSkill.getrank('Enchanting')
 
     if rank <= 25 # Tier 1  Extremely Easy
-      # Buy an abolition sigil and bone totem
+      # Buy an abolition sigil and bone totem for Radiant Trinket
+      recipe = get_data('recipes').crafting_recipes.find { |recipe| recipe['name'] =~ /a radiant trinket/ }
       DRCM.ensure_copper_on_hand(2000, @settings)
-      DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], 2) unless DRCI.exists?("abolition sigil-scroll")
-      DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], 6) unless DRCI.exists?("bone totem")
+      DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], recipe['enchant_stock1']) unless DRCI.exists?("abolition sigil-scroll")
+      DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], recipe['item']) unless DRCI.exists?("bone totem")
       DRCC.stow_crafting_item(DRC.right_hand, @bag, @belt)
       DRCC.stow_crafting_item(DRC.left_hand, @bag, @belt)
       DRCC.fount(crafting_data['artificing'][@hometown]['tool-room'],1,crafting_data['artificing'][@hometown]['fount'],2,@bag, @bag_items, @belt)
       DRCT.walk_to(@training_room)
-      DRC.wait_for_script_to_complete('enchant', [2, 'a radiant trinket', 'totem'])
+      DRC.wait_for_script_to_complete('enchant', [recipe['chapter'], recipe['name'], recipe['noun'].split.last])
       DRCI.dispose_trash('totem', @worn_trashcan, @worn_trashcan_verb)
     elsif rank <= 50 # Tier 2 - Very Easy
-      # Buy an induction sigil and bone totem
+      # Buy an induction sigil and bone totem for Flash Trinket
+      recipe = get_data('recipes').crafting_recipes.find { |recipe| recipe['name'] =~ /a flash trinket/ }
       DRCM.ensure_copper_on_hand(2000, @settings)
-      DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], 1) unless DRCI.exists?("induction sigil-scroll")
-      DRCT.order_item(crafting_data['artificing'][@hometown]['stock-room'], 6) unless DRCI.exists?("bone totem")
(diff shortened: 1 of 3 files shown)


asechrest:
@Dartellum Would like some visual spacing here in the code. It's wall-of-text-ish at the moment.

I'll need @Raykyn55 to help evaluate this as well. I'm not good with crafting scripts.

craft.lic
```diff
@@ -401,10 +401,38 @@ class Craft
...
+    elsif rank <= 1175 # Tier 10 - Difficult - Need to use sigilhunter or similar to gather the sigils
+      # Examples for sigilhunter:
+      # ;sigilharvest crossing evolution 15 2
+      # ;sigilharvest crossing metamorphosis 15 2
+      # ;sigilharvest crossing unity 15 2
+      # or: ;multi 3,;sigilharvest crossing evolution 15 2,;sigilharvest crossing metamorphosis 15 2,;sigilharvest crossing unity 15 2
+      DRCM.ensure_copper_on_hand(4000, @settings)
+      exit if !DRCI.exists?("evolution sigil-scroll")
+      exit if !DRCI.exists?("metamorphosis sigil-scroll")
+      exit if !DRCI.exists?("unity sigil-scroll")
```
asechrest, line 413: We seem to have a fair amount of `exists?` checks here. Can we break this out into a method that helps us check?

Also, do we need some cleanup here in case we have tools in our hands or what have you? Or is a brute-force exit here appropriate?

Raykyn55:
Shouldn't be any tools in our hand at this point(re cleanup)

I think the exit is appropriate because we for sure don't want to START crafting without having everything on hand for fear of wasting a sigil we have to farm, but moving it into a separate method for these checks is probably a good idea. Reason being, you have information there that the user might want(the commented information specifically), and if the script just dies, they might know they're missing one of the sigils, but not the whole list. What about sending the needed parts to a method that checks for them, and if any fail, it spits out that helpful set of comments you have in the code?

Something like this in place of the tap checks:
```ruby
sigils_needed = ['evolution', 'metamorphosis', 'unity']
check_sigils(sigils_needed)
```
method could be something like:
```ruby
def check_sigils(sigils)
  missing = []
  missing << sigils.select { |sigil| DRCI.exists?("#{sigil} sigil-scroll") }
  unless missing.empty?
    DRC.message("Missing #{missing.to_s} sigil-scroll(s). IF you have them, they need to be loose in your bag, if you do not, these cannot be purchased, and must be harvested. We recommend using our sigilhunter script. Run ;sigilhunter help for more information.")
    exit
  end
end
```

craft.lic
Dartellum: You can only harvest these sigils. 

> Shouldn't be any tools in our hand at this point(re cleanup)
> 
> I think the exit is appropriate because we for sure don't want to START crafting without having everything on hand for fear of wasting a sigil we have to farm, but moving it into a separate method for these checks is probably a good idea. Reason being, you have information there that the user might want(the commented information specifically), and if the script just dies, they might know they're missing one of the sigils, but not the whole list. What about sending the needed parts to a method that checks for them, and if any fail, it spits out that helpful set of comments you have in the code?
> 
> Something like this in place of the tap checks:
> 
> ```ruby
> sigils_needed = ['evolution', 'metamorphosis', 'unity']
> check_sigils(sigils_needed)
> ```
> 
> method could be something like:
> 
> ```ruby
> def check_sigils(sigils)
>   missing = []
>   missing << sigils.reject { |sigil| DRCI.exists?("#{sigil} sigil-scroll") }
>   unless missing.empty?
>     DRC.message("Missing #{missing.to_s} sigil-scroll(s). IF you have them, they need to be loose in your bag, if you do not, these cannot be purchased, and must be harvested. We recommend using our sigilhunter script. Run ;sigilhunter help for more information.")
>     exit
>   end
> end
> ```

I am working on a `common-crafting` way. I have it working with `workorders` and can use here (my plan was for that all along).

Dartellum:
With the new method in DRCC:
```
>;craft enchanting
--- Lich: craft active.
[craft]>open my embroidered backpack
That is already open.
[craft]>open my party sack
That is already open.
[craft]>open my watery portal
[Assuming you mean a swirling eddy of incandescent light bound by a gold-striated coralite frame.]
You can't do that.
[craft]>open my green backpack
That is already open.
[craft]>open my blue backpack
That is already open.
[craft]>open my red backpack
That is already open.
[craft]>open my duffel bag
That is already open.
[craft]>open my zillinen pouch
That is already open.
[craft]>open my thigh bag
That is already open.
[craft]>open my striking bag
That is already open.
[craft]>open my hip pouch
That is already open.
[craft]>open my champion's belt
That is already open.
[craft]>open my wyvern-hide lootpouch
That is already open.
[craft]>open my canvas sack
What were you referring to?
[craft]>wealth kronars
  9 gold, 5 silver, 46 bronze, and 40 copper Kronars (10,000 copper Kronars).
[craft]>rummage /C sigil-scroll in my embroidered backpack
You rummage through an embroidered backpack depicting a large town by the side of a river looking for something similar to "sigil-scroll":
crafting (7): 
    flawless paradox sigil-scroll comprised of broad strokes (3)
    flawless integration sigil-scroll comprised of broad strokes (2)
    flawless evolution sigil-scroll comprised of broad strokes (2)
[craft]>rummage /C sigil-scroll in my embroidered backpack
You rummage through an embroidered backpack depicting a large town by the side of a river looking for something similar to "sigil-scroll":
crafting (7): 
    flawless paradox sigil-scroll comprised of broad strokes (3)
    flawless integration sigil-scroll comprised of broad strokes (2)
    flawless evolution sigil-scroll comprised of broad strokes (2)
Not enough metamorphosis sigil-scroll(s). You need to harvest 1 more. We recommend using our sigilhunter script. Run ;sigilhunter help for more information.
[craft]>rummage /C sigil-scroll in my embroidered backpack
You rummage through an embroidered backpack depicting a large town by the side of a river looking for something similar to "sigil-scroll":
crafting (7): 
    flawless paradox sigil-scroll comprised of broad strokes (3)
    flawless integration sigil-scroll comprised of broad strokes (2)
    flawless evolution sigil-scroll comprised of broad strokes (2)
Not enough unity sigil-scroll(s). You need to harvest 1 more. We recommend using our sigilhunter script. Run ;sigilhunter help for more information.
--- Lich: craft has exited.
```
image

Dartellum:
Switching to draft so common-crafting PR can get merged first. I will create that one after a few more test runs.

Dartellum:
Relies on #6071.

Dartellum:
I want to move all the other sections of enchanting to point to the new def.

Dartellum:
Added lots of comments. I also incorporated the find training room change from #6082.

Dartellum:
:question: Should I change `;craft enchanting` to `;craft artificing`? Two reasons: first, it would then match `;workorders artificing` and second, when invoking and binding are released enchanting will not work.

Raykyn55:
> ❓ Should I change `;craft enchanting` to `;craft artificing`? Two reasons: first, it would then match `;workorders artificing` and second, when invoking and binding are released enchanting will not work.

I would argue no, given the nature of this script. The goal here is to train the skill, the script decides the best way to go about it. This is different from workorders, where you're deciding what discipline to do your work in.

Dartellum:
> > question Should I change `;craft enchanting` to `;craft artificing`? Two reasons: first, it would then match `;workorders artificing` and second, when invoking and binding are released enchanting will not work.
> 
> I would argue no, given the nature of this script. The goal here is to train the skill, the script decides the best way to go about it. This is different from workorders, where you're deciding what discipline to do your work in.

Okay, fair enough.

PR #6071, opened 2022-08-23
[script][common-crafting] New method added to handle checking for sigils and purchasing, if it can.

Dartellum:


common-crafting.lic
```diff
@@ -346,5 +371,23 @@ module DRCC
       end
     end
   end
-    
+
+  def check_for_existing_sigil?(sigils, stock_number, quantity, bag, belt, info)
```
Raykyn55, line 350: Are you sending this a list of sigils?

Dartellum: It should only receive one at a time by each or called with the enchant_name, etc, directly. See the updated `craft` script. `Workorders` is forth coming too. All working as I expect.

Raykyn55: If it's just checking one sigil, 'sigils' shouldn't be plural.

common-crafting.lic
```diff
@@ -346,5 +346,30 @@ module DRCC
...
-    
+  
+  def check_for_existing_sigil?(sigil, stock_number, quantity, bag, belt, info)
+    merged = []
+    primary = []
+    secondary = []
+    primary = ["abolition", "congruence", "induction", "permutation", "rarefaction"]
+    secondary = ["antipode", "ascension", "clarification", "decay", "evolution", "integration", "metamorphosis", "nurture", "paradox", "unity"]
+    merged.concat primary
+    merged.concat secondary
```
asechrest, line 357: Can this entire section be consolidated to something like:
```ruby
primary = ["abolition", "congruence", "induction", "permutation", "rarefaction"]
secondary = ["antipode", "ascension", "clarification", "decay", "evolution", "integration", "metamorphosis", "nurture", "paradox", "unity"]
merged = primary + secondary
```
I assume you're keeping the two primary and secondary arrays just so it's easy to see which sigils are primary vs. secondary when reading the code, otherwise I think this section could be consolidated to a single line.

Dartellum: I tried that and it did not work. I can try again later. The only way I could get them together and flattened was how I did it. Should I add the Guild sigils now even though they are not released?

Raykyn55: Try
```
merged = [primary + secondary].flatten
```
That declares the array at the same time you populate it. 

Larger question: if all of these are now purchaseable, whats the sense in having two lists?

asechrest: There is no need to flatten two 1-D arrays that are simply concatenated, which is what the `+` operator does.

image


If we don't need the two arrays for some code readability/understanding sake, this whole section turns into a one-line array assignment.

Dartellum: Today `merged = primary + secondary` worked. Last night it was giving me them combined but like this: `[["abolition", "congruence", "induction", "permutation", "rarefaction"], ["antipode", "ascension", "clarification", "decay", "evolution", "integration", "metamorphosis", "nurture", "paradox", "unity"]]` :man_shrugging: 

Anyway, should I add in the Guild ones now?

Dartellum:
Leaving guild planned sigils out for now.

asechrest:
Dart, this is a little bit picky but I think it greatly improves code readability if we try to adhere to the Rubocop/Ruby [Style Guideline](https://github.com/rubocop/ruby-style-guide#empty-lines-between-methods) that suggests we should break up methods into internal "paragraphs" that make logical sense to be grouped together.

That can be a little nebulous but in this case some visual spacing really helps. So the main if/else block should have a space between the things before it to set it apart. Perhaps also set apart the array variable assignments from some of the handling below it. Things like that.

asechrest:
Thanks for the spacing. Looks much better. I lied about a final comment. @rpherbig notes that this is perhaps the 3rd or 4th place in code where we list the sigils. Shall we turn the sigil lists into a global?

Dartellum:
> Thanks for the spacing. Looks much better. I lied about a final comment. @rpherbig notes that this is perhaps the 3rd or 4th place in code where we list the sigils. Shall we turn the sigil lists into a global?

I can take a look in doing that tonight and tomorrow night. If it makes sense, sure.

Dartellum:
> > Thanks for the spacing. Looks much better. I lied about a final comment. @rpherbig notes that this is perhaps the 3rd or 4th place in code where we list the sigils. Shall we turn the sigil lists into a global?
> 
> I can take a look in doing that tonight and tomorrow night. If it makes sense, sure.

A quick search on sigil in `common-crafting` only shows 9 matches all within my new method. Two of those are referencing `sigilhunter`.

Dartellum:
You mean like `$primary = ["abolition", "congruence", "induction", "permutation", "rarefaction"]` in `common-crafting` at the beginning?

asechrest:
> You mean like `$primary_sigils = ["abolition", "congruence", "induction", "permutation", "rarefaction"]` in `common-crafting` at the beginning?

Yes, that's my thought. Then we can utilize those globals directly here, in the 3 other scripts, and in any others.

Dartellum:
I will test.

Dartellum:
Those are not arrays in those scripts. I mean, not a bad idea to have them global since used in multiple scripts. So, I would need to take `PRIMARY_SIGILS_PATTERN = /\b(?:abolition|congruence|induction|permutation|rarefaction) sigil\b/` and then make that into an array to check against the incoming. 

The only real reason to keep the check if the sigil is in the list is for when Guild or new sigils are released that are not store bought. I would like to keep the script in the mode that some may not go on sale at the Society store in the future, just like secondaries were not for years.

Dartellum:
> Those are not arrays in those scripts. I mean, not a bad idea to have them global since used in multiple scripts. So, I would need to take `PRIMARY_SIGILS_PATTERN = /\b(?:abolition|congruence|induction|permutation|rarefaction) sigil\b/` and then make that into an array to check against the incoming.
> 
> The only real reason to keep the check if the sigil is in the list is for when Guild or new sigils are released that are not store bought. I would like to keep the script in the mode that some may not go on sale at the Society store in the future, just like secondaries were not for years.

I found a way to do it. Testing will begin!

asechrest:
@Dartellum How's this looking? Thanks for adding the sigil globals. I'm showing this PR has conflicts to resolve.

Dartellum:
> @Dartellum How's this looking? Thanks for adding the sigil globals. I'm showing this PR has conflicts to resolve.

Some updates need added. Working well for me and the few testing from Discord have not posted anything wrong.

Dartellum:
Resolved conflicts. Added a catch in case the same challenge occurs. Although, it did not in initial testing but rather safe than sorry.

PR #6074, opened 2022-08-24
[data] [base-recipes] Switching item order number

MahtraDR:
TT131 which just dropped moved bone totem from number 6, to number 16 in the order list. This accounts for that.


Smaptastic:
Line 8958 also needs the change I believe. Otherwise looks good.

MahtraDR:
> Line 8958 also needs the change I believe. Otherwise looks good.

Well spotted. Thank you. Fixed.

PR #6084, opened 2022-08-25
[script][common-crafting] Cost calculator for common-crafting

Raykyn55:
Feed in your recipe, parts, and stock material if applicable, and this will spit out a cost calculation for crafting said item, or n items, in the currency of your hometown. Base values are in kronars.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -196,6 +196,34 @@ module DRCC
     end
     true
   end
+  
+  def crafting_cost(recipe, parts, quantity, hometown, material = false)
+    currency = DRCM.town_currency(hometown)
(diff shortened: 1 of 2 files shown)


Dartellum:
:question: Costs are the same in all Providences?

Raykyn55:
> ❓ Costs are the same in all Providences?

they are, just the currency varies, though I do believe they round up, hence .ceil.

Raykyn55:
Put in some comments regarding use of the method, so folks don't have to figure out what's required to use the method. Plugged in the stock-numbers for the new sigils, and set up method to allow nil for parts and quantity, for recipes with no parts, or where quantity isn't defined, method assumes quantity of 1.

PR #6094, opened 2022-08-26
[data][base-spells] Add Unyielding and Seek meditations

Smaptastic:
Adds the Unyielding and Seek meditations to base-spells. Unyielding has no end messaging.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -451,6 +451,15 @@ barb_abilities:
     start_command: meditate staunch
     activated_message: ^You .* to meditate
     expired_message: heart quickens and blood rushes back
+  Unyielding:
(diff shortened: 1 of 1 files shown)


data/base-spells.yaml
```diff
@@ -451,6 +451,15 @@ barb_abilities:
     start_command: meditate staunch
     activated_message: ^You .* to meditate
     expired_message: heart quickens and blood rushes back
+  Unyielding:
```
jbrock24, line 454: Unyielding requires an expired message so it can set the flag appropriately to reuse when worn off.

Smaptastic: It doesn’t have one that I can find. See initial comment.

PR #6095, opened 2022-08-26
[data][base-lumber] Fix for bad room in White Laurel Grove

Smaptastic:
Not sure what happened in the initial mapping, but I got a weird bad room number in White Laurel Grove. Fixed here.

Edit: Ok this room is bizarre. It changed back to the original room ID which I thought was bad. Commented it out to resolve the issue.


Smaptastic:
Ok apparently the room that was giving trouble was double-mapped and has now been fixed. This is now redundant, as the initial mapping is in line with the fixed number.

PR #6101, opened 2022-08-30
[script][common-crafting] - Fix argument order

asechrest:
Fix order of args. Defaults must come at the end, and can't be intermixed with other args that don't have a default.


asechrest:
Self-merging to fix breaking change.

PR #6102, opened 2022-08-30
base-spells - aethrolysis was miscategorized

BinuDR:
Aethrolysis is a WM spell, not Necro.


Kiriawen:
my bad,  I was going from memory on a lot of those

PR #6103, opened 2022-08-30
base-spells - corrections and updates

BinuDR:
Will O' Wisp cantrip is Warrior Mage, not Bard.
Rite of Forbearance was misspelled.
Ghoulflesh was missing.  Can someone with a necromancer check and make sure I got all the needed info?
Smite Horde is a Paladin spell, not Cleric.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -864,17 +864,6 @@ spell_data:
     prep: prepare
     recast: 2
     guild: Bard
-  Will O Wisp:
-    skill: cantrip
-    abbrev: C W O W
-    prep_time: 0
-    cast: gesture
-    harmless: true
-    mana:
-    recast: 1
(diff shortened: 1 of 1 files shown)


BinuDR:
Sidholt's Flaying is not technically a Necromancer spell.  It is sorcerous for Necromancers to cast too.
We might need to handle this differently.

MahtraDR:
> Sidholt's Flaying is not technically a Necromancer spell.  It is sorcerous for Necromancers to cast too.
> 
> We might need to handle this differently.

The way we check sorcery in common-arcana is by comparing native spell type. Could we just set that to a special mana type?

Kiriawen:
I apparently don't know every spell in the game!  :D I did the majority of these from memory and only spot checked a couple I wasn't sure of.  Apparently I should have checked more of them

PR #6109, opened 2022-09-02
[script][clean-lumber] Update, add table saw, use deed packets

Raykyn55:
Removed includes, add prefixes
implemented table_saw arg for using table saw rather than manually
packet rather than stack it


clean-lumber.lic
```diff
@@ -5,81 +5,91 @@
...
-  end
-
-  def stow_item(name)
-    stow_crafting_item(name, @bag, @belt)
+    %w[stick log branch limb deed].each { |size| 
+      loop do
+        break unless process_size(args, size)
+        cut_and_stack(args)
+      end
+      }
```
asechrest, line 31: For a multi-line `.each`, the `.each do |size|` without curly braces is preferred.

PR #6115, opened 2022-09-04
[script][bescort] - Fix edge case for Ice Road collect rocks

asechrest:
Fixes an edge case where we fail to collect rocks and thus move too quickly.

Prior behavior would not pause if we somehow failed to collect rocks. This could be because your hands were full, ~~or you were playing an instrument~~, etc. This change favors a brute-force `fput('collect rock')`, and then detecting whether we're in RT. If so, we're collecting and we kick the pile and move on. If not, it pauses 15 (the same duration as a collect), so that we don't move quickly and get injured.

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -1067,9 +1067,12 @@ class Bescort
       move_fast = false if Flags['bescort-move-slow']
 
       unless move_fast
-        DRC.collect('rock')
-        DRC.kick_pile? # don't litter the ice road
-        waitrt?
+        fput('collect rock')
+        if waitrt?
+          DRC.kick_pile? # don't litter the ice road
+        else
+          pause 15
+        end
       end
     end
 


asechrest:
Hmm, I did discover a downside to using `fput`. If you're running `performance`, and you can't don your skates, previous version uses `bput` which handles `stop play` and then retries. `fput` doesn't have such handling, so if you're playing an instrument it won't successfully collect, but it _will_ pause 15.

bescort.lic
```diff
@@ -1067,9 +1067,12 @@ class Bescort
       move_fast = false if Flags['bescort-move-slow']
 
       unless move_fast
-        DRC.collect('rock')
-        DRC.kick_pile? # don't litter the ice road
-        waitrt?
+        fput('collect rock')
```
MahtraDR, line 1070:
```suggestion
        fput('stop play')
        fput('collect rock')
```

MahtraDR: It's an fput. Super cheap, no fuss. :) You could even add in a `DRC.fix_standing` in there.

Kiriawen:
@asechrest @MahtraDR 

Is this still in progress or ready for review?  If you don't intend to continue with it please close it out.

Thanks!

PR #6116, opened 2022-09-04
[script][first-aid] - Allow training granularity for FA/Scholarship/Both

asechrest:
Adds args to allow you to focus on first aid, scholarship, or both. Also adds settings to base-help.yaml.

Defaults to checking `First Aid` skill if none is specified, to retain prior behavior.

### DIFF
--- a/data/base-help.yaml
+++ b/data/base-help.yaml
@@ -218,6 +218,20 @@ bescort_hide:
   specific_descriptions:
     bescort: Hides while on ferries.
 
+bleed_bot:
+  description: The name of a character you're using as a first aid trainer bleed bottom.
+  example: Gildaren
+  referenced_by:
+    - first-aid
+  specific_descriptions:
+
+bleed_bot_room:
+  description: The room number where your bleed bot character lives.
+  example: 1234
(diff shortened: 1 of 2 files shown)


Dartellum:
As a side note, the `attunement` script has an arg for scholarship too:
```ruby
{
          name: 'sigil',
          regex: /sigil/,
          optional: true,
          description: 'Do sigil walking (scholarship training) instead of regular attunement walking'
        }
```

Smaptastic:
Using `;first-aid first aid` seems to break it for me, but that's not a big deal since the default targets the First Aid skill so it's kinda a redundant argument. Works fine for me with default, scholarship, and both.

asechrest:
> 

Thanks for testing, and I know why. Having a script argument with a space, and without enclosing quotes, will cause the script to think the second word is an additional argument. Since this script only has one argument, it fusses at you because it thinks you have two. Will work on a fix.

Smaptastic:
The change looks good. There wasn't a need for first aid as a setting, given that it's default. Works perfectly for both and scholarship. Ship it.

PR #6119, opened 2022-09-05
[script][appraisal] Fix app focus

Raykyn55:
Couple changes.
1. add wait arg, since research is interrupted by any number of actions, no point doing focus without waiting, but if you have a followup sequence that doesn't affect, or accounts for, research (athletics?) you can opt in or out of the waiting.
2. Reworked focus:
  - Doesn't fetch the item, since that's unnecessary
  - Added a check for current research/research unavailable at the top of each focus section.
  - Moved the checks into it's own method.

Currently using this in conjunction with a modified version of athletics, but works as a standalone if managed properly, or if wait arg is used.

### DIFF
--- a/appraisal.lic
+++ b/appraisal.lic
@@ -10,8 +10,8 @@ class Appraisal
     arg_definitions = [
       [
         { name: 'focus', regex: /focus/i, description: 'Perform appraise focus on an item.' },
-        { name: 'item', regex: /\w+/i, description: 'Item to use appraise focus with.' },
-        { name: 'wait', regex: /^wait/i, optional: true, description: 'Waits for research to complete before exiting'},
+        { name: 'item', regex: /\w+/i, description: 'Item to use appraise focus with, use double quotation marks for multi-word items(eg "warring axe").' },
+        { name: 'wait', regex: /^wait/i, optional: true, description: 'Waits for focus project to complete before exiting'},
         { name: 'script_summary', optional: true, description: 'Trains the Appraisal skill by appraising your gear, zills, bundles, gem pouches, and studying the art in the Crossing art gallery.'}
       ],
       [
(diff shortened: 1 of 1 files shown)


asechrest:
@Raykyn55 - This script appears to be broken.

```
;apprais focus magic
--- Lich: appraisal active.
[appraisal]>appraise focus check
You feel ready for any sort of appraisal focus.
>
[appraisal]>appraise focus my magic
You can't seem to find much to focus your appraisal that way.
>
Too recently completed a project
--- Lich: appraisal has exited.
```
I believe that's improper game syntax for appraise focus of a skill.

For PRs marked ready for review, we need to make sure these _exact_ edits are running smoothly for you locally, else marking them as draft is probably best.

Raykyn55:
So, it's not clear to me that you ever need to append the appraisal focus with "my", in which case it's unnecessary to differentiate between concepts and items. Using appraisal focus on anything does not require that you hold said thing. You can focus on a wall for athletics, a weapon for that skill, a concept like khri or magic, and you needn't hold any of those things, nor specify your own objects. The risk, if it exists, of focusing the wrong item(eg axe on the ground over axe in your hand) would be rare, and in 99% of those rare cases, inconsequential. The axe example would risk focusing on the wrong weapon skill (LE vs 2he vs SE), or a forging hammer rather than a throwing hammer, but in any case where items may interfere, the addition of an adjective by the user would obviate this risk.

This also opens up the use of, say, a coin for thievery, which wouldn't be possible as the script was written. This eliminates problems with things like exo armor, which have a cooldown for removal. This also means you can focus any item even if your hands are full.

The only limitation here is items you store in your eddy. I don't have one, so I'm not 100% on how they work, but my understanding is the items aren't "there" until you look in the bag, or only there for a time after you look in the bag. Someone with an eddy would have to confirm, but perhaps just don't use eddy items for focus, or keep your focus items elsewhere.

PR #6120, opened 2022-09-05
[script][athletics] Add skip arg, send hib hometown to shard

Raykyn55:
Adds skip arg to athletics to allow for focus appraisal in t2. Could plausibly work that into alternative routines (lockpicking?)
Lets hib hometown use shard's superior athletics training options. The river is just a bad/dangerous way to train athletics, #changemymind

NOT choosing skip doesn't affect default behavior, so this is just an added feature geared towards appraisal focus.


Raykyn55:
going a different route with this, I think.

PR #6121, opened 2022-09-05
[script][buff] Add charge arg

Raykyn55:
Adds optional 'charge' arg to buff for warrior mages who use elemental weapons. This can be triggered via the argument, or a yaml setting:
```yaml
buff_max_elemental_charge: true
```
Allows standalone max of your charge with ;buff charge only(or really any word that isn't a waggle_set), complementary chargeup AFTER buffing with ;buff charge, or ;buff <waggle_set> charge. Finally, if the yaml is set, tacks on a max chargeup at the end of any use of buff script.

### DIFF
--- a/buff.lic
+++ b/buff.lic
@@ -20,9 +20,9 @@ class Waggle
 
     setname = args.spells || 'default'
 
-    unless settings.waggle_sets[setname] || args.charge
+    unless settings.waggle_sets[setname]
       DRC.message("No waggle set found for name: #{setname}")
-      exit
+      exit unless args.charge
     end
 
     if args.force
--- a/buff.lic
+++ b/buff.lic
@@ -9,6 +9,7 @@ class Waggle
   def initialize
     arg_definitions = [
       [
+        { name: 'charge', regex: /charge/i, optional: true, description: 'Max out your elemental charge by summoning admittances' },
         { name: 'spells', regex: /\w+/, optional: true, description: 'Spell list to use, otherwise default' },
         { name: 'force', regex: /force/i, optional: true, description: 'Recast spells even if currently active' }
       ]
@@ -19,7 +20,7 @@ class Waggle
 
     setname = args.spells || 'default'
 
-    unless settings.waggle_sets[setname]
+    unless settings.waggle_sets[setname] || args.charge
       DRC.message("No waggle set found for name: #{setname}")
       exit
     end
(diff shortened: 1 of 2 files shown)


buff.lic
```diff
@@ -19,7 +20,7 @@ class Waggle
 
     setname = args.spells || 'default'
 
-    unless settings.waggle_sets[setname]
+    unless settings.waggle_sets[setname] || args.charge
```
asechrest, line 23: If the user doesn't have a `waggle_set` for `setname`, but does call the script with a charge arg, this evaluates to true, and the script doesn't exit. What, then, happens, when we get to line 34's `DRCA.do_buffs`?

Raykyn55: script ends. line 20 sets a default. If run as ;buff charge only, or ;buff charge pickledickle (or any string that's not a waggle set) do_buffs just returns:
```ruby
  def do_buffs(settings, set_name)
    # takes a waggle and sends it to the appropriate helper
    return unless settings.waggle_sets[set_name]
```
and the script ends.

asechrest: We don't want a silent exit. Current behavior warns if we have neither the proper `setname` NOR a proper `default`.

Can you help me understanding the reason for adding the conditional `|| args.charge` check here`? It seems to me that the vast, vast majority of users will at the very least have a `default` waggle, which seems to make the `|| args.charge` check here useful only in the instance that a user has passed in a waggle in the script call that doesn't exist OR doesn't have a `default` waggle.

Raykyn55: If you run buff, and all of your buffs are up, you already get the same silent exit. For this, the only way to get that silent exit is after charge does it's thing:
```
[buff]>summon admittance

You continue meditating, further increasing your proximity to the Elemental Plane of Air.
Roundtime: 13 sec.
Finishing your meditation, you feel that you have reached your limit.

--- Lich: buff has exited.
```
Regards conditional, this is to allow you to run charge only. Absent that arg, you will still get the message, and MOST users won't even know this change occurred. They'll still get the warning about no waggle named blah unless they named their waggle 'charge' exactly.

Raykyn55: Perhaps a better answer to your question is that without that added conditional, we can't skip buffing with spells. If we don't send a valid waggle the script just exits. In any use case where you don't use the charge arg, you'll still get the same exit. If you do use the charge arg, AND you send an invalid waggle, you'll get ONLY the charging behavior. In that case, you'd be sending an invalid waggle on purpose. 

I suppose if we wanted to notify the user in all cases if they've send an invalid waggle, we can move the conditional to the exit stage:
```ruby
unless settings.waggle_sets[setname]
  DRC.message("No waggle set found for name: #{setname}")
  exit unless args.charge
end
```
That would permit the same behavior, but still send the message even if all you wanted was to charge up.

Dartellum:
I thought a public script existed for this already. I believe one is in `;repos`.

PR #6127, opened 2022-09-08
[script][go2] - Store Start Room

asechrest:
This is a really simple change. It stores the start room that you're in when calling `;go2 <target>` in a UserVar for future use. ~~I have a simple companion script called `;goback` that takes you back to the room you started in. Sometimes I'll need to head to a place and then go back right from whence I came. It's just kind of a little QOL script thing.~~

Now implements a `goback` feature within `;go2` itself.

### DIFF
--- a/go2.lic
+++ b/go2.lic
@@ -450,6 +450,7 @@ for var in script.vars[1..-1]
       target_search_array.push(var)
    end
 end
+
 target_search_string = target_search_array.join(' ')
 
 #
@@ -538,6 +539,17 @@ end
 unless start_room = Room.current
    echo 'error: your current room was not found in the map database'
    exit
+end
+
+if target_search_string =~ /goback/i
+  last_start_room = UserVars.go2_start_room.to_s
+  if last_start_room
+    target_search_string = last_start_room
+    echo "Attempting to return to last start room of: #{last_start_room}"
+  else
+    echo "go2 does not have a stored start room from the last time it ran."
+    exit
+  end
 end
 
 #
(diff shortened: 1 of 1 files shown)


bwfisher82:
Does it make sense to have this 'go back' in go2 rather than another script? We have a lot and this feature is a bit tiny to get a new script?

Maybe ',go2 last' takes you to the last room go2 was launched from?

If so, when you do go back, you should set the start room again to where you returned from. So doing ,go2 last two or more times puts you in a loop. I'm not sure why you'd want that but it explains what I'm thinking. If this was already in the other goback script, good job... I haven't looked at it yet.

Also a way to toggle this off and customize 'last' to another word in settings in case you want to or already are using 'last' in custom tags (even though I doubt it).

Two other ideas sprang to mind unrelated to this request / feature. A ,go2 history [list | number] sounds nice, with some configurable for history size. A ,go2 stats that would show your top rooms. Both are other features however, and not a part of this.

What do you think of the rest of the above comments?

asechrest:
@bwfisher82 
It's a good thought. @Raykyn55 asked me the same thing the other day. I think you're both probably right. I find `go2`'s code to be a big legacy so it's not what I'm used to looking at, but let me play with it a bit.

And yes, the idea here, and the functionality implemented, stores the room you start in each time you start `;go2`.

asechrest:
This is now ready for review.

`;go2 goback` will return you to the room `;go2` last started in.

asechrest:
@MahtraDR 

Ticked the version. Added language to help. Thanks.

PR #6134, opened 2022-09-12
[scripts][tarantula] Fix bad anchoring of regex.

MahtraDR:
In adding the anchoring, I missed that there was a `[` as the first character.

The actual line is 
```
[You need to vary which skillset you select with every use.  Magic was your last used skillset.]
```


MahtraDR:
Self-merging since it's a regression fix.

PR #6149, opened 2022-09-19
[script][heroic-tattoo] Update

Raykyn55:
Makes script a class
Adds a Flag to catch when tattoo falls, giving a more exact time for reup. This changes the maximum time between _actually refreshable_ and _next attempt to refresh_ from a possible 5 minutes plus 15 seconds, to 15 seconds.

### DIFF
--- a/data/base-help.yaml
+++ b/data/base-help.yaml
@@ -647,6 +647,12 @@ herb_container:
   specific_descriptions:
     alchemy: Get herbs from here for alchemy.
 
+heroic_tattoo_downtime_override:
+  description: Allows custom downtime (in seconds) for heroic tattoo cycle, for folks who have a reliable arcana buff.
+  example: 220 #seconds
+  referenced_by:
+    - heroic-tattoo
+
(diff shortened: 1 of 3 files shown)


Raykyn55:
Lets invoke timers persist between logins with UserVars.herotatt_next_invoke mirroring next_invoke updates within the script. So logging out and in again doesn't trigger an attempt at reup unless at least the timer has expired.

Raykyn55:
Added:
```yaml
# Optional setting if you ALWAYS or RELIABLY have an arcana buff up, to reduce the calculated downtime modifier to a custom downtime.
# This will be a function of (60 - <arcana modified ranks>/24)*60, added on to the end of the timer from an expired heroic buff.
heroic_tattoo_downtime_override: # In seconds
```
to base.yaml. This is an optional setting you can use if you have a reliable arcana buff, to manually input a downtime.

asechrest:
Thanks. Could you throw the new setting into `base-help.yaml`?

Will merge after that.

PR #6150, opened 2022-09-20
[script][combat-trainer] - Add Earthquake improved whirlwinds

mikejcook:
Added tracking of the Barbarian Earthquake berserk improved Whirlwind trigger. When an improved Whirlwind is triggered, prioritize this attack over other possibilities.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4121,6 +4121,7 @@ class GameState
     if DRStats.barbarian?
       Flags.add('ct-barbarian-whirlwind', 'Your inner fire roils with a rising desire to unleash a whirlwind attack!')
       Flags.add('ct-barbarian-whirlwind-expired', 'The whirlwind moment passes.')
+      echo "Improved whirlwind flags reset" if $debug_mode_ct
     end
 
     @dedicated_camb_use = settings.dedicated_camb_use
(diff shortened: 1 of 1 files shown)


Raykyn55:
HI!

So, currently, as a condition for whirlwind, CT checks:
```ruby
(npcs.length > 2) && !balance_low?
```
If we're triggering whirlwind more often, would it be appropriate to check those two conditions _in addition_ to the flags?

mikejcook:
> HI!
> 
> So, currently, as a condition for whirlwind, CT checks:
> 
> ```ruby
> (npcs.length > 2) && !balance_low?
> ```
> 
> If we're triggering whirlwind more often, would it be appropriate to check those two conditions _in addition_ to the flags?

When the improved whirlwind fires, it's a balance neutral attack that does improved damage and has a 2 second RT. So when it fires, it's a super great thing to take advantage of. That's the newly added check.

In addition, it can be a good thing to use when you have a lot of enemies and have good balance, which is what the second condition checks, which is what we've always done.

I'm hoping that clears it up!

Raykyn55:
> When the improved whirlwind fires, it's a balance neutral attack that does improved damage and has a 2 second RT. So when it fires, it's a super great thing to take advantage of. That's the newly added check.
> 
> In addition, it can be a good thing to use when you have a lot of enemies and have good balance, which is what the second condition checks, which is what we've always done.
> 
> I'm hoping that clears it up!

I see, so it's just hands-down the best move if it's available, even if there's only one enemy and your balance isn't top notch. Makes sense. 

My little barb doesn't have WW yet, so I can't test, but suggest adding some debug messaging for those who can. Something to indicate CT has chosen WW because of the flag, and perhaps a message indicating flags being reset. thinking lines 4123/4351 ish.

asechrest:
Think we can recruit at least one tester in Discord? Maybe try pinging sharkbait?

asechrest:
Thanks for the PR. Thanks, Sharkbait, for testing.

PR #6151, opened 2022-09-26
[scripts][common-moonmage] Use "cast ambient" when no moons available  for Cage of Light

MahtraDR:
Caters specifically to moonie spell Cage of Light which can be cast at a moon, but can also be "cast ambient" as an alternative when no moons are available.

### DIFF
--- a/common-moonmage.lic
+++ b/common-moonmage.lic
@@ -326,8 +326,10 @@ module DRCMM
     moon = visible_moons.first
     if moon
       data['cast'] = "cast #{moon}"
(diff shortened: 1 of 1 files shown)


MahtraDR:
Tested multiple times on mine. With moons CoL gets cast at a moon, without any moons, it gets cast ambient every time.

MahtraDR:
Works for more than just me :) 

https://discord.com/channels/745675889622384681/745676388056432731/1024086360916762705

PR #6152, opened 2022-09-26
[scripts][combat-trainer] SK with Electrogenesis

XTyrsin:
Added aim messaging for Syamelyo Kuniyo while under the effect of Electrogenesis metaspell.


asechrest:
Thanks for the PR.

PR #6153, opened 2022-09-27
[data][base-recipes] advanced AP ritual focus fix

SekmehtDR:
Odd one, maybe a one-off.

This recipe requires a secondary sigil to begin, then ANY sigil to complete it.

The issue is that as is, the script purchases congruence, but then looks for decay (secondary) to begin.

By adding enchant_stock1 to Decay, and enchant_stock2 to Congruence, this work order can be completed as expected.

Technically this PR could have been completed by just adding sigil 10, but I wanted to change the order to reflect Decay (Secondary) and #1 and Congruence (or any) as #2.

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -9236,7 +9236,8 @@ crafting_recipes:
   type: artificing
   chapter: 3
   work_order: false
-  enchant_stock1: 3
+  enchant_stock1: 10
+  enchant_stock2: 3
   item: 16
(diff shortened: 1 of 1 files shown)


SekmehtDR:
I'll fiddle more with the recipes, but this one is just weird because it requires a secondary sigil to begin and any after.

SekmehtDR:
I've tested this modification and I can complete this item/work order without issues now.

PR #6154, opened 2022-09-30
[script][forge] Minor edit to resume

Raykyn55:
Commands defined in find_item were being overwritten in set_defaults. This fixes that.


Raykyn55:
@asechrest

PR #6168, opened 2022-10-05
[script][forge] Add missing match

Raykyn55:
When resuming a shield you've started, but stopped at the assembly stage, analyze misses the match for assembly. This is what's required.


Raykyn55:
@asechrest Ready, very small edit.

PR #6173, opened 2022-10-07
[script][common-crafting] Add in new way to handle enchanting items.

Dartellum:


common-crafting.lic
```diff
@@ -460,4 +463,27 @@ module DRCC
     end
   end
 
+  def check_for_existing_sigil?(sigil, stock_number, quantity, bag, belt, info)
+    merged = Regexp.union($PRIMARY_SIGILS_PATTERN, $SECONDARY_SIGILS_PATTERN)
+
+    tmp_count = 0
+    more = 0
+    tmp_count = DRCI.count_items_in_container("#{sigil} sigil-scroll", bag)
```
Raykyn55, line 471: This method in common items strips out any word but the last, so we only get a count of "sigil"
```ruby
  def count_items_in_container(item, container)
    contents = DRC.bput("rummage /C #{item.split.last} in my #{container}", /^You rummage .*/, /That would accomplish nothing/)
    # This regexp avoids counting the quoted item name in the message, as
    # well as avoiding finding the item as a substring of other items.
    contents.scan(/ #{item}\W/).size
  end
  ```
might have to just use an edited version locally, eg rummage container, same match (You rummage) and use scan on that against the specific sigil?

File: common-crafting.lic [hunk 2]
```diff
@@ -460,4 +463,27 @@ module DRCC
...
+      return true
+    else
+      if merged.match?("#{sigil} sigil")
+        more = quantity - tmp_count
+        # Found a weird challenge that made the temp_part_count equal 1 even though no "sigil" was in container
+        more += 1 if reget(3, "but there is nothing in there like that")
+        DRCC.order_enchant(info['stock-room'], more, stock_number, bag, belt)
+        return true
+      else
+        DRC.message("Not enough #{sigil} sigil-scroll(s). You can purchase or harvest #{more} more. We recommend using our sigilhunter script. Run ;sigilhunter help for more information.")
```
Raykyn55, line 483: This is likely deprecated, since they're all purchaseable.

common-crafting.lic
Dartellum: So far it works as intended. The same method is used for herb-stock and restock.

common-crafting.lic
Dartellum: Except for guild specific and any new ones they want to throw in. I would like to go proactive on this instead of reactive.

Dartellum:
This was tested by the community and I believe ready to go.

PR #6185, opened 2022-10-16
Allow setting for workorders.lic to use a different town than hometown.

cfrisby77:


Kiriawen:
you'll need to test this extensively.  There are many sub scripts that get called and you'd need to ensure that none of them will use the hometown instead of the new setting as well.

cfrisby77:
Possible further testing required.

PR #6196, opened 2022-10-28
[scripts][combat-trainer] Only list status of egg_or_warhorn during d…

MahtraDR:
…ebug if defined

Was debugging an unrelated issue on a character with no egg, or warhorn and kept getting spammed.

This removes the spam.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -852,12 +852,12 @@ class LootProcess
   end
 
   def dissected?(mob_noun, game_state)
-    if @dissect_for_thanatology 
+    if @dissect_for_thanatology
       return false if (DRSkill.getxp('Thanatology') == 34 && DRSkill.getxp('First Aid') == 34)
     else
       return false if DRSkill.getxp('First Aid') == 34
     end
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2125,7 +2125,7 @@ class AbilityProcess
     check_paladin_skills
     check_nonspell_buffs(game_state)
     check_battle_cries(game_state)
-    DRC.message("warhorn_or_egg array is: #{@warhorn_or_egg}")  if $debug_mode_ct
+    DRC.message("warhorn_or_egg array is: #{@warhorn_or_egg}")  if (@warhorn_or_egg && $debug_mode_ct)
```
MahtraDR, line 2128:
```suggestion
```

MahtraDR:
My original PR added a check if the egg was defined. I don't think this line is even needed, and is a vestige of a debug line instead.

asechrest:
IIRC I used this to track the warhorn and egg rotation. I'm fine with removing it.

PR #6198, opened 2022-10-29
[scripts][combat-trainer] Use DRCI for getting bundling rope

MahtraDR:
There was a rope on the ground...
```
[combat-trainer: message: You pick up some bundling rope.]
[combat-trainer: checked against [/You get/i, /What were you referring to/i, /You need a free hand/i]]
[combat-trainer: for command get bundling rope]
```

As reported on Discord https://discord.com/channels/745675889622384681/745675890242879671/1035916880948502538

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -915,7 +915,7 @@ class LootProcess
         @equipment_manager.stow_weapon(game_state.weapon_name)
       end
 
-      if DRC.bput('get bundling rope', 'You get', 'What were you referring to', 'You need a free hand') == 'You get'
+      if DRCI.get_item_if_not_held?('bundling rope')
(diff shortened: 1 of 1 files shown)


Raykyn55:
Would it make sense to add a container list via storage_containers yaml setting?

MahtraDR:
> Would it make sense to add a container list via storage_containers yaml setting?

I thought about doing that, but the routine that gets rope just does a stow, so there's no telling where it'll end up.

https://github.com/rpherbig/dr-scripts/blob/8b2b16cfcdc1766ea3675f23a31c2b3efc9b2e4b/hunting-buddy.lic#L252

We can do a simple `DRCI.get_item?` or this, which at least lets you already have the rope in hands.

PR #6199, opened 2022-10-31
Add base-healingherbs and update herb-stock to utilize it

mikejcook:
Added base-healingherbs.yaml with Crossing and Shard healing herb store information. Updated herb-stock.lic to utilize new data and cleaned up the script.

- Flattened the default Crossing herb data. Added Shard herb data.
- Added functionality to pull the herb data in for the appropriate town. Supported towns are Crossing and Shard. Riverhaven will go to Crossing if athletics are high enough. I'm unclear on if the islands offer herb restock options. If they do, we'd want to add that to the base data.
- Removed validate data method, since it's now being provided in a data file
- Prior script was setup to use herbs defined in the player's yaml or those found in consumables (looks like it originally used restock.lic as a baseline). In theory, the herbs could have been added to base-consumables and we could have just used restock, but that ship sailed long ago and this avoids breaking backwards compatibility. So, I removed all of the code around pulling herbs from the player's yaml and dropped the parsing of the data to figure out if it wanted to get them from the store defined in the yaml or from base-consumables.
- Overall cleanup of the script
- heal-remedy uses herbs to heal if the herbs setting in the player's yaml has data. So, once someone gets these changes, that will continue to work. Moving forward, using herbs to heal will simply require herbs: true in the yaml. I call herb-stock as part of my T2, which will work without any yaml changes. There could be a future opportunity to incorporate this into hunting-buddy.

### DIFF
--- a/data/base-healingherbs.yaml
+++ b/data/base-healingherbs.yaml
@@ -1,0 +1,255 @@
+herb_stocks:
+  Crossing:
+    - name: jadice flower
+      size: 6
+      stackable: true
+      room: 8259
+      price: 812
+      quantity: 12
+    - name: plovik leaf
+      size: 6
+      stackable: true
+      room: 8259
+      price: 812
+      quantity: 12
+    - name: nilos salve
+      size: 6
+      stackable: true
+      room: 8259
+      price: 812
+      quantity: 12
+    - name: hulnik grass
+      size: 6
+      stackable: true
+      room: 8259
+      price: 812
+      quantity: 12
+    - name: nemoih root
+      size: 6
+      stackable: true
+      room: 8259
+      price: 875
+      quantity: 12
+    - name: georin salve
+      size: 6
+      stackable: true
+      room: 8259
+      price: 875
+      quantity: 12
+    - name: sufil sap
+      size: 6
+      stackable: true
+      room: 8259
+      price: 875
+      quantity: 12
+    - name: yelith root
+      size: 6
+      stackable: false
+      room: 8259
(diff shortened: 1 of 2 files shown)


Dartellum:
Missing all the Riverhaven ones, particularly ones you cannot get in Crossing or Shard. The following three are ones I use all the time:

```yaml
  #GRETZKY THE HERBALIST'S COTTAGE
    lujeakave_elixir:
      hometown: Riverhaven
      name: lujeakave elixir
      size: 6
      stackable: false
      room: 15019
      price: 870
      quantity: 9
    eghmok_potion:
      hometown: Riverhaven
      name: eghmok potion
      size: 6
      stackable: false
      room: 15019
      price: 720
      quantity: 9
    aevaes_solution:
      hometown: Riverhaven
      name: aevaes solution
      size: 6
      stackable: false
      room: 15019
      price: 740
      quantity: 9
```

mikejcook:
Added the three Riverhaven herbs along with handling for herbs that have location-overrides. If the location is Riverhaven, I check to make sure the player has enough athletics skill.

MahtraDR:
I don't use this script today.

How backwards compatible is this?

mikejcook:
> I don't use this script today.
> 
> How backwards compatible is this?

I pulled in all of the herbs found in the sample herbs user yaml. So, if you are just using the default herbs from Crossing or Shard, it's 100% backwards compatible. You wouldn't know any different. Dart shared some custom herbs that I added to the commit.

PR #6214, opened 2022-11-07
Create taskmaster.lic

Raykyn55:


Raykyn55:
yaml settings:
```yaml
task_loot_bag: backpack
task_use_boost: true
```
can run for a single task with just the name of the task giver as the argument, as a loop of that task giver with 'repeat' arg, and as a cycle randomly choosing the task giver with 'cycle'

Tools you will need: 
zasele (outfitting)
knitting needles
sewing needles
awl
yardstick
slickstone
scissors

bradyn (engineering)
bone saw
chisels
shaper
rasp
carving knife
drawknife

poltu (forging)
hammer
tongs
shovel
stirring rod

associated consumables will be handled by the scripts, excess materials and instructions are discarded, sacks are cleaned of coins and gems and stowed, regardless of what is inside. There is messaging for notable finds, but it saves every sack.

Raykyn55:
Needs handling for using fang cove, room selections and the like. Right now it's hardcoded for crossing

Dartellum:
You should make `tool_drop` and `tool_pickup` optional.

Dartellum:
> > You should make `tool_drop` and `tool_pickup` optional.
> 
> 
> 
> so, it's either repair through shop, or repair on your own. Do you mean, making repairing at all optional?

Yes, because I have the skip repair option set. Also, I thought this section was to store the tools with the clerk.

Raykyn55:
> > > You should make `tool_drop` and `tool_pickup` optional.
> > 
> > 
> > so, it's either repair through shop, or repair on your own. Do you mean, making repairing at all optional?
> 
> Yes, because I have the skip repair option set. Also, I thought this section was to store the tools with the clerk.

negative, clerk-tools would be your go-to on that. I'm pushing an edit to give you the option of not repairing at all.
```yaml
task_no_repair: true
```

Raykyn55:
@Dartellum See comments at top of this commit

PR #6218, opened 2022-11-08
[script][forge] ingot fetching

Raykyn55:
Specified fetching ingots from your crafting container. Smith stows it there, as does the new taskmaster, but the DRCC.get_crafting_item doesn't fetch from there unless it's on the list of items to get from your bag.
```ruby
DRCC.stow_crafting_item('ingot', @bag, @forging_belt)
```
This makes it a specific fetch from the same bag it should be in, instead of just from anywhere on your person.

### DIFF
--- a/forge.lic
+++ b/forge.lic
@@ -151,7 +151,7 @@ class Forge
     end
     swap_tool(@home_tool)
     if @metal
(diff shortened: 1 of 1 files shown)


Raykyn55:
doing something different for this

PR #6224, opened 2022-11-10
[script][sew] Grabs the raw material from your bag

Raykyn55:
Previously it was grabbing the first it found on your person unless you added the particular cloth to your bag. This edit quickly stuffs that material into the bag_items variable before sending the command off to get_crafting_item, thereby appending the command with "from my #{crafting_container}". 

Also dumps the book stow, unnecessary extra step.


Raykyn55:
Issue here is stow feet puts the cloth/leather in your default bag, which for some folks is different from their crafting bag. Need to sort out how to get around this.

Raykyn55:
Same as ingot handling PR, this should be done with either a yaml setting (not ideal) or some kind of tracking with dependency.

PR #6229, opened 2022-11-12
[script][forge] Ingot handling

Raykyn55:
Adds the metal-type ingot to your bag items before fetching, so that it always fetches from your crafting container, instead of grasping at any ol' metal-type ingot you might be carrying.

### DIFF
--- a/forge.lic
+++ b/forge.lic
@@ -151,8 +151,9 @@ class Forge
     end
     swap_tool(@home_tool)
     if @metal
+      @bag_items << "#{@metal} ingot"
       DRCC.get_crafting_item("#{@metal} ingot", @bag, @bag_items, @forging_belt)
-      DRC.bput('put my ingot on anvil', 'You put your')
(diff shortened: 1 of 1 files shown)


Raykyn55:
same issue as sew, though likely unfixable. Game is coded to automatically default stow any spare metal after the first pound into default bag.

Dartellum:
> same issue as sew, though likely unfixable. Game is coded to automatically default stow any spare metal after the first pound into default bag.

Perhaps you can capture where the game stowed the metal and then use `transfer-items` or just code it in to move that metal back to the crafting container.

Raykyn55:
> 

not a bad idea, catch the bag iwth a flag maybe, and compare the flag with the crafting container, resolve in finish method?

Raykyn55:
Eventually I think we'll need to determine what your default bag is, and make that information available to these scripts. Either a yaml setting (pita) or ideally something in dependency.

PR #6230, opened 2022-11-12
[scripts][circlecheck] Don't discard zero-rank skills on new characters

MahtraDR:
I need to figure out how to do this in drinfomon too...

### DIFF
--- a/circlecheck.lic
+++ b/circlecheck.lic
@@ -543,12 +543,12 @@ def display_requirements(requirements, level_target, brief, pretty_print)
     # This outputs the beginning section of circle check, which
     # gives your current circle and your TDPs at next level.
     DRC.message("___________________________________________________________")
-    _respond " Current Circle:  <pushBold/>#{DRStats.circle}<popBold/>" + 
+    _respond " Current Circle:  <pushBold/>#{DRStats.circle}<popBold/>" +
       " | Next Level TDPs:  #{DRStats.tdps} <pushBold/>+ #{tdps}<popBold/> = #{DRStats.tdps + tdps}"
     respond("")
-    
+
     # Set column headings for later display for prettyprint
-    column_headings =   
+    column_headings =
       "SKILL".center(column_spacing[1]) +
       "CATEGORY".center(column_spacing[2]) +
       "CIR".rjust(column_spacing[3]) +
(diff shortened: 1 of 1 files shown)


MahtraDR:
image

asechrest:
I don't quite understand the output here. Should we just suppress the output and message for new characters?

MahtraDR:
No, the top is "before" and the actual display is "after" this change. 

Sorry, I could have conveyed that better.

asechrest:
> 

I still don't understand the display "after". While it doesn't error out, it doesn't display meaningful data for most of those lines. What I'm asking is if we should detect this for newbies, and output something different?

MahtraDR:
> I still don't understand the display "after". While it doesn't error out, it doesn't display meaningful data for most of those lines. What I'm asking is if we should detect this for newbies, and output something different?

Oh, no, I don't think so.  She's a level 1 moonie, those lines are correct.

The script doesn't run at all without this change, for level 1 characters otherwise. Any skills they don't have are just nil. All this is doing is taking that nil.to_i which makes them zero, so there is a comparison point for the circle requirements.

MahtraDR:
Clarifying:

With this change:
image

Without this change:
image

asechrest:
Sorry, maybe I'm not being clear, either. While this fixes the script to work, what I'm getting at is this meaningless output:

image

MahtraDR:
> Sorry, maybe I'm not being clear, either. While this fixes the script to work, what I'm getting at is this meaningless output:
> 
> image

I don't understand what you mean! :D

This is reflecting the skills the character has. She's a level 1 moon mage, and doesn't have those skills at all. These are the only skills she has:

image

So circlecheck is correctly telling me I have zero in a skill I need points in to circle... The script is correctly identifying skills needed to circle, but since the game hasn't given her any points in those skills, they are nil. This change makes `nil.to_i` effectively, which converts them to zero.

asechrest:
Haha, maybe we're totally talking around each other. Let me pose you this question:

image

Is this a meaningful line in any way? It doesn't indicate what skill is being referenced, and says we have enough skill for circle 0 (duh?). Therefore, it's a worthless output. I'm asking why we're giving worthless output, though I'm granting you that the script not breaking is an improvement.

MahtraDR:
Back in business.

image

Level 0/1 characters have a lot of zeroes... we shouldn't discard them. No sense discarding any skills here, even...

PR #6231, opened 2022-11-13
tarantula - Possible fix for infinite pause bug

BinuDR:
Please test and let me know if this works.

It'll delay your tarantula use by 20 seconds when it would otherwise get stuck in an infinite loop.


Dartellum:
I implemented this and guess it works as expected. I have not had any loops.

PR #6249, opened 2022-11-23
[scripts][common-healing] Add dead status to output

MahtraDR:
Allows us to detect whether our touch target is dead or not, and perhaps act accordingly.

```
2022-11-23-3.log:"[exec1: {\"wounds\"=>{}, \"parasites\"=>{}, \"poisoned\"=>false, \"diseased\"=>false, \"dead\"=>true, \"score\"=>0}]"
2022-11-23-3.log:"[exec1: {\"wounds\"=>{}, \"parasites\"=>{}, \"poisoned\"=>false, \"diseased\"=>false, \"dead\"=>false, \"score\"=>0}]"
```

In my case, I use it to determine whether I want to link unity (which will kill my empath if the person being healed is a corpse...) or take all quick, which won't kill my empath ever.

### DIFF
--- a/common-healing.lic
+++ b/common-healing.lic
@@ -99,7 +99,7 @@ module DRCH
   def perceive_health_other(target)
     return unless DRStats.empath?
 
-    case DRC.bput("touch #{target}", 'Touch what', 'feels cold and you are unable to sense anything', 'avoids your touch', 'You sense a successful empathic link has been forged between you and (?<name>\w+)\.')
+    case DRC.bput("touch #{target}", 'Touch what', 'feels cold and you are unable to sense anything', 'avoids your touch', 'You sense a successful empathic link has been forged between you and (?<name>\w+)\.', /^(He|She) is dead\./)
     when 'avoids your touch', 'feels cold and you are unable to sense anything', 'Touch what'
       return nil
     when /You sense a successful empathic link has been forged between you and (?<name>\w+)\./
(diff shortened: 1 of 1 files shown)


common-healing.lic
```diff
@@ -138,17 +138,24 @@ module DRCH
       /^[\w]+ (?:body|skin) is covered (?:in|with) open oozing sores/
     ])
 
+    dead_regex = Regexp.union([
+      /^(He|She) is dead/
+    ])
+
```
asechrest, line 144: What is the purpose of `Regexp.union` here, given you're only specifying one regex pattern? `.union` will give a union of multiple patterns.

MahtraDR: In an earlier draft I had the two lines separate and followed the existing convention. I don't think we'll ever had to add a line here though. I'll fix in a bit.

MahtraDR: Actually, I rather keep it like this. It's all the same for the result, but this syntax leaves open simple additions for this regex (even if I can't think of any today) and is simple to read.

Dartellum:
This update for when using `empathylink`?

MahtraDR:
> This update for when using `empathylink`?

I hadn't considered that use case, actually. Should work too. 

I use it for random healing with an alias. If they're dead I `take all quick` if they are alive, I `link unity`.

PR #6252, opened 2022-11-23
[scripts][moonwatch] Add check and notice if used on any server than Prime

MahtraDR:
image

As per title.

I asked @Kiriawen to test on The Fallen, and works as intended there.


Kiriawen:
image

PR #6253, opened 2022-11-23
[scripts][corn-maze]Flags to bput, and other updates

MahtraDR:
Updated to common methods, remove flags for long maze, swapping with much more reliable bput case when.

Addresses specific issue brought up by Remyngton where the flag capture with the simple `else` made a person running through seem like a kill task.

Have run 70+ long maze runs with this version.

### DIFF
--- a/corn-maze.lic
+++ b/corn-maze.lic
@@ -298,18 +298,6 @@ class CornMaze
   def wait_for_command
     reset_flags
     Flags.reset('task_done')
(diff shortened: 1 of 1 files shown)


asechrest:
Does this force logging status messages to a window, and if so is that a change in behavior?

MahtraDR:
> Does this force logging status messages to a window, and if so is that a change in behavior?

Yes technically, but it's an addition, not a change in behavior as such.

asechrest:
> 

Let's see who complains. :)

PR #6256, opened 2022-11-27
[scripts][combat-trainer] Patching out monkey ostruct

MahtraDR:
Ruby 3, here we come....


MahtraDR:
Self merging because it's a regression.

PR #6263, opened 2022-11-29
[scripts][heal-remedy]Remove the pause

MahtraDR:
It doesn't need the pause.

It could do with a proper cleanup, which I am not keen to do right this minute.


Dartellum:
What do you mean it does not need the pause? What if you run it outside of combat-trainer?

MahtraDR:
> What do you mean it does not need the pause? What if you run it outside of combat-trainer? I have no problems with it pausing and unpausing scripts.

I mean I can't think of a context in which it needs to pause other scripts to do its business. It's not an opportunistic script like burgle, or a script that runs all the time like tarantula, or almanac. It gets called explicitly, which in turn means it runs, then finishes its business. Combat-trainer calls it with `DRC.wait_for_script_to_complete` as does find-dark.lic, which means those scripts themselves wait for heal-remedy to run before they continue.

Dartellum:
Go ahead and remove the pause. We discussed this and it is run with `DRC.wait_for_script_to_complete`. I just removed the pause from my copy to test.

Dartellum (approved):

MahtraDR:
@asechrest can you please merge?

PR #6264, opened 2022-11-30
common-crafting - prefer exact match when multiple recipes found

mikejcook:
When finding a matching recipe, if multiple matches are found, prefer an exact match, should it exist. If an exact match is not found, use previous logic to require user to select recipe.

I tested a couple 'normal' recipes without issue. Tested the specific problematic recipe by adding the following yaml:
workorder_recipes:
  weaponsmithing:
    - a metal pike

Here's the issue that sent me down this path, for reference. Note that upon getting 'a metal pike' from the forge person, smith is unsure which recipe I am trying to craft. As a result, smith hangs until it receives input.

```
[wo_forge]>ask Yalda for challenging weaponsmithing work
Yalda shuffles through some notes and says, "Alright, this is an order for a metal pike. I need 2 of superior quality, made from any material and due in 115 roisaen.  Please complete the items, bundle them with your logbook and then give me the logbook to complete this order.  Good luck!"

[smith: Please select desired recipe ;send #]
    1: a metal pike
    2: a metal pike staff
```

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -152,6 +152,12 @@ module DRCC
     when 1
       recipes.find { |x| x['name'] =~ /#{item_name}/i }
     else
+      exact_matches = recipes.map { |x| x['name'] }.select { |x| x == item_name }
+
+      if exact_matches.length == 1
+        return recipes.find { |x| x['name'] == item_name }
+      end
+
       echo('Please select desired recipe ;send #')
       match_names.each_with_index { |x, i| respond "    #{i + 1}: #{x}" }
       line = get until line.strip =~ /^([0-9]+)$/
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -158,6 +158,7 @@ module DRCC
         return recipes.find { |x| x['name'] == item_name }
       end
 
+      DRC.message("Using the full name of the item you wish to craft will avoid this in the future (e.g. 'a metal pike' vs 'metal pike')")
       echo('Please select desired recipe ;send #')
       match_names.each_with_index { |x, i| respond "    #{i + 1}: #{x}" }
       line = get until line.strip =~ /^([0-9]+)$/


Raykyn55:
so the concern I would have is the need to match exactly. so "coarse leathers" would miss, right? you'd need to run ;sew with "some coarse leathers" or ;forge with "a metal pike" instead of metal pike. Would be a problem for workorders if you're using a whitelist.

Do you have any logs of what happens when you drop "a" or "some" ?

mikejcook:
> so the concern I would have is the need to match exactly. so "coarse leathers" would miss, right? you'd need to run ;sew with "some coarse leathers" or ;forge with "a metal pike" instead of metal pike. Would be a problem for workorders if you're using a whitelist.
> 
> Do you have any logs of what happens when you drop "a" or "some" ?

So, it might help to look at the method as a whole, below. I'm not actually changing anything with how passing coarse leathers would have worked before. If coarse leathers returns a single match, then we don't even get to the new 'exact match' check. The exact match check only happens in the event that we find more than one match. In the else portion of the check, which occurs only when we find more than one recipe matching our item_name, I perform a check to see if there's an exact match for the item_name. If there is, we use that. If there's not, we fall back and use the old logic, which requires the user to send which recipe they want to make.

`  def recipe_lookup(recipes, item_name)
    match_names = recipes.map { |x| x['name'] }.select { |x| x =~ /#{item_name}/i }
    case match_names.length
    when 0
      echo("No recipe in base-recipes.yaml matches #{item_name}")
      nil
    when 1
      recipes.find { |x| x['name'] =~ /#{item_name}/i }
    else
      exact_matches = recipes.map { |x| x['name'] }.select { |x| x == item_name }

      if exact_matches.length == 1
        return recipes.find { |x| x['name'] == item_name }
      end

      echo('Please select desired recipe ;send #')
      match_names.each_with_index { |x, i| respond "    #{i + 1}: #{x}" }
      line = get until line.strip =~ /^([0-9]+)$/
      item_name = match_names[Regexp.last_match(1).to_i - 1]
      recipes.find { |x| x['name'] =~ /#{item_name}/i }
    end
  end`

MahtraDR:
What if we did something like this instead maybe?
```ruby
    when 1
      recipes.find { |x| x['name'] =~ /#{item_name}$/i }
      ```

**Comment by mikejcook** [comment]
> What if we did something like this instead maybe?
> 
> ```ruby
>     match_names = recipes.map { |x| x['name'] }.select { |x| x =~ /#{item_name}$/i }
> ```

I'm not sure what you mean? The line you added is already in the existing code. That's the code that finds the possible matches. Because it does fuzzy matching, it currently finds a metal pike and a metal pike staff.

**Comment by asechrest** [comment]
> ```ruby
> #{item_name}$
> ```

I haven't looked closely at this PR, but note that what Mahtra posted uses the `$` character, which asserts position at the end of a line.

**Comment by MahtraDR** [comment]
This is the difference, based on the example given here.
image
versus
image

**Comment by mikejcook** [comment]
Oh. I'm not really sure. That would change the current behavior, and I'm not sure if that would negatively impact anyone. If someone currently tries to craft something with a partial word, such as 'sleeve' vs 'sleeves' it would work today. With your suggested change, it would no longer work. So that might not be desirable?

some rugged leather sleeves is a recipe. Today, they could type "rugged leather sleeve" and it would work, after this change, it would fail to find a match at all. Obviously not a forging recipe, but you get the idea. So I would be hesitant to change the behavior of the method in that way. With the change I made, it should have no impact on existing usage, except that if they've typed the exact name of a recipe (or workorders is providing the exact name), then it'll choose that specific recipe.

**Comment by MahtraDR** [comment]
> Oh. I'm not really sure. That would change the current behavior, and I'm not sure if that would negatively impact anyone. If someone currently tries to craft something with a partial word, such as 'sleeve' vs 'sleeves' it would work today. With your suggested change, it would no longer work. So that might not be desirable?
> 
> some rugged leather sleeves is a recipe. Today, they could type "rugged leather sleeve" and it would work, after this change, it would fail to find a match at all. Obviously not a forging recipe, but you get the idea. So I would be hesitant to change the behavior of the method in that way. With the change I made, it should have no impact on existing usage, except that if they've typed the exact name of a recipe (or workorders is providing the exact name), then it'll choose that specific recipe.

In effect, what you're describing is the same problem as `a metal pike` matching `a metal pike staff` the reason that happens is the same reason that allows `rugged leather sleeve` to match `rugger leather sleeves` you can't have one without the other as such.

In either case, whatever we do, will require a change by the user, either user input, or an edit of their existing yaml (in the case of `rugger leather sleeve` when they want `rugged leather sleeves`) I have no horse here, just pointing out the root cause of the original problem statement. :)

**Comment by mikejcook** [comment]
I don't think I see what you're saying. What issue does the change I made introduce? What change would the user need to make? There is currently an issue where the script requires input. My change would remove that need. Other situations should continue working as they do today.

**Comment by MahtraDR** [comment]
> I don't think I see what you're saying. What issue does the change I made introduce? What change would the user need to make? There is currently an issue where the script requires input. My change would remove that need. Other situations should continue working as they do today.

With the change you're proposing, what happens if a user has `a metal pik` in their recipes list, versus what happens before the change?

**Comment by mikejcook** [comment]
Right now, a metal pik would match to both a metal pike and a metal pike staff. They would be presented with a request to send 1 or 2 to select one of the recipes. With the change, that would remain the same, because a metal pik wouldn't be an exact match to either recipe, so it would skip the change I made and continue using the old logic.

The only change here is in bold. If and only if multiple matches are found, if one of the recipes matches exactly, it'll choose that instead of requiring user input. I think maybe there's some confusion from Raykyn55's original question, where I think they thought the change was requiring an exact match period? But that's not the case, it just uses the exact match as an escape valve if it needs it.

```
def recipe_lookup(recipes, item_name)
    match_names = recipes.map { |x| x['name'] }.select { |x| x =~ /#{item_name}/i }
    case match_names.length
    when 0
      echo("No recipe in base-recipes.yaml matches #{item_name}")
      nil
    when 1
      recipes.find { |x| x['name'] =~ /#{item_name}/i }
    else
      **exact_matches = recipes.map { |x| x['name'] }.select { |x| x == item_name }

      if exact_matches.length == 1
        return recipes.find { |x| x['name'] == item_name }
      end**

      echo('Please select desired recipe ;send #')
      match_names.each_with_index { |x, i| respond "    #{i + 1}: #{x}" }
      line = get until line.strip =~ /^([0-9]+)$/
      item_name = match_names[Regexp.last_match(1).to_i - 1]
      recipes.find { |x| x['name'] =~ /#{item_name}/i }
    end
  end
```

**Comment by MahtraDR** [comment]
> Right now, a metal pik would match to both a metal pike and a metal pike staff. They would be presented with a request to send 1 or 2 to select one of the recipes. With the change, that would remain the same, because a metal pik wouldn't be an exact match to either recipe, so it would skip the change I made and continue using the old logic.
> 
> The only change here is surrounded by asterisks (apparently you can't bold in the code block). If and only if multiple matches are found, if one of the recipes matches exactly, it'll choose that instead of requiring user input. I think maybe there's some confusion from Raykyn55's original question, where I think they thought the change was requiring an exact match period? But that's not the case, it just uses the exact match as an escape valve if it needs it.
> 
> ```
> def recipe_lookup(recipes, item_name)
>     match_names = recipes.map { |x| x['name'] }.select { |x| x =~ /#{item_name}/i }
>     case match_names.length
>     when 0
>       echo("No recipe in base-recipes.yaml matches #{item_name}")
>       nil
>     when 1
>       recipes.find { |x| x['name'] =~ /#{item_name}/i }
>     else
>       **exact_matches = recipes.map { |x| x['name'] }.select { |x| x == item_name }
> 
>       if exact_matches.length == 1
>         return recipes.find { |x| x['name'] == item_name }
>       end**
> 
>       echo('Please select desired recipe ;send #')
>       match_names.each_with_index { |x, i| respond "    #{i + 1}: #{x}" }
>       line = get until line.strip =~ /^([0-9]+)$/
>       item_name = match_names[Regexp.last_match(1).to_i - 1]
>       recipes.find { |x| x['name'] =~ /#{item_name}/i }
>     end
>   end
> ```

I understand what you're proposing. I don't oppose it. I am simply pointing out that the source of the problem is the loose regex in the first place. If we just anchor the regex to the end of the `item_name` it'll solve the problem you're presenting, with a smaller code change. It does have the side effect of requiring people to end their recipes correctly (no `metal piks` for example.) Yours has the benefit of moving the solution forward, without impacting existing users of metal piks though.

**Comment by mikejcook** [comment]
I think the loose regex was probably actually intentional to better facilitate use when a person is calling smith directly to craft something, as opposed to just being a part of a workorder routine. Allow a little room for abbreviations, omitting 'some' or 'a' etc.

Here are some addition logs for comparison.

**Before the change (current behavior):**

>;smith steel "a metal pike"
[smith: Please select desired recipe ;send #]
    1: a metal pike
    2: a metal pike staff

>;smith steel "a metal pike staff"
[smith: {"name"=>"a metal pike staff", "noun"=>"staff", "volume"=>11, "type"=>"weaponsmithing", "work_order"=>true, "chapter"=>8}]

>;smith steel "coarse leathers"
[smith: {"name"=>"some coarse leathers", "noun"=>"leathers", "volume"=>55, "type"=>"tailoring", "work_order"=>false, "chapter"=>9, "part"=>["large padding"]}]

>;smith steel "a metal pik"
[smith: Please select desired recipe ;send #]
    1: a metal pike
    2: a metal pike staff

**After the change:**

>;smith steel "a metal pike"
>
[smith: {"name"=>"a metal pike", "noun"=>"pike", "volume"=>15, "type"=>"weaponsmithing", "work_order"=>true, "chapter"=>7, "part"=>["long pole"]}]
>;smith steel "a metal pike staff"
[smith: {"name"=>"a metal pike staff", "noun"=>"staff", "volume"=>11, "type"=>"weaponsmithing", "work_order"=>true, "chapter"=>8}]

>;smith steel "coarse leathers"
>
[smith: {"name"=>"some coarse leathers", "noun"=>"leathers", "volume"=>55, "type"=>"tailoring", "work_order"=>false, "chapter"=>9, "part"=>["large padding"]}]
>;smith steel "coarse leather"
[smith: Please select desired recipe ;send #]
    1: a coarse leather aventail
    2: a coarse leather mask
    3: a coarse leather cap
    4: a coarse leather helm
    5: a coarse leather tasset
    6: a coarse leather cowl
    7: a coarse leather jerkin
    8: a coarse leather mantle
    9: some coarse leather sleeves
    10: a coarse leather vest
    11: a coarse leather robe
    12: a coarse leather coat
    13: some coarse leathers

>;smith steel "a metal pik"
[smith: Please select desired recipe ;send #]
    1: a metal pike
    2: a metal pike staff


**The Regex change:**

>;smith steel "a metal pike"
[smith: {"name"=>"a metal pike", "noun"=>"pike", "volume"=>15, "type"=>"weaponsmithing", "work_order"=>true, "chapter"=>7, "part"=>["long pole"]}]


>;smith steel "a metal pike staff"
[smith: {"name"=>"a metal pike staff", "noun"=>"staff", "volume"=>11, "type"=>"weaponsmithing", "work_order"=>true, "chapter"=>8}]


>;smith steel "coarse leathers"
[smith: {"name"=>"some coarse leathers", "noun"=>"leathers", "volume"=>55, "type"=>"tailoring", "work_order"=>false, "chapter"=>9, "part"=>["large padding"]}]


>;smith steel "a metal pik"
[smith: No recipe in base-recipes.yaml matches a metal pik]
[smith: ]
>

**Comment by MahtraDR** [comment]
I personally prefer that last one and someone trying to craft a metal pik should fix their spelling. 😀

It's your PR. I just wanted the discussion. You and @Raykyn55 decide. Happy to merge as is.

**Comment by Raykyn55** [comment]
Revisiting this because I ran into a similar problem with metal spear and light metal spear, which the anchor wouldn't catch. 
```
>;smith steel "metal spear" temper balance
>
[smith: Please select desired recipe ;send #]
    1: a metal spear
    2: a light metal spear
>;k smith
>;smith steel "a metal spear" temper balance
[smith]>wealth kronars
  3 platinum Kronars (30,000 copper Kronars).
>
--- Lich: go2 active.
--- Lich: 'textsubs' has been stopped by go2.
[go2: ETA: 0:00:14 (25 rooms to move through)]
```
I like it. Worst case, if folks are using whitelists with items that have multiple matches, they'll need to fill in the full name from base-recipes if they don't want select to pop up. Only thing I'd possibly suggest is adding an additional note to indicate this possiblity.

```
DRC.message("Using the full name of the item you wish to craft will avoid this in the future")
echo('Please select desired recipe ;send #')
```
or something to that effect.

mikejcook:
I like that. I'll add the messaging to the PR tomorrow or Monday.

PR #6266, opened 2022-12-01
[script][go2] - Change to Script.current.name

asechrest:
Change from local `script.name` to `Script.current.name`. More robust, matches GS.


MahtraDR:
Why not just merge the core lich script here instead? It has all our features anyway and we wouldn't have to maintain both.

PR #6277, opened 2022-12-12
Create slim-remedy.lic

Raykyn55:
Super basic version of remedy, designed for workorders/single-use crafting. Doesn't resume, assumes you have what you need.

Lifted some of the tools logic, but stripped out purchasing and item generation handling (because alcohol/water is ridiculously cheap).

### DIFF
--- a/slim-remedy.lic
+++ b/slim-remedy.lic
@@ -19,12 +19,13 @@ class SlimRemedy
 
         @recipes = get_data('recipes')['remedies']
         recipe = DRCC.recipe_lookup(@recipes, args.recipe_name)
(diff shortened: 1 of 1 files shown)


asechrest:
Thanks for the draft PR. Any reason to make this a separate script, vs. allowing the existing `remedy.lic` to handle a "slim" option?

Raykyn55:
> Thanks for the draft PR. Any reason to make this a separate script, vs. allowing the existing `remedy.lic` to handle a "slim" option?

variable overlap and flags, primarily. It's also just easier visually, and for debug purposes. I was considering doing it as a rewrite, but there was a lot of functionality in the remedy script that craft, for example, needs remedy to handle.

Raykyn55:
requires #6280

Raykyn55:

PR #6280, opened 2022-12-13
[data][base-recipes] New pattern, sort + collapse

Raykyn55:
Reworked under discipline headers, parts collected into parts array.


Raykyn55:
changed blue flowers to crushed flowers, to reflect the actual item you'd need. requires update to crafting calc to specify unpurchaseable.

PR #6282, opened 2022-12-13
[script][workorders] Rewrite

Raykyn55:
requires #6280


mikejcook:
Hey, so I have my own custom workorder that I hacked for forging because of stock parts, but based on your statement that you can train 0-cap in every craft using store-bought materials, I have a question.

Bronze bars are 5 volume, so workorders, your new one included, will go through and buy one bronze bar for each quantity item needed, correct? So once an item you want to craft is over 5, it'll stop working. I believe that's why most people switch to using their own ingots. Is that correct, or am I off-base on that? My take has always been that stock workorders only really works when you are first getting started.

Raykyn55:
> Hey, so I have my own custom workorder that I hacked for forging because of stock parts, but based on your statement that you can train 0-cap in every craft using store-bought materials, I have a question.
> 
> Bronze bars are 5 volume, so workorders, your new one included, will go through and buy one bronze bar for each quantity item needed, correct? So once an item you want to craft is over 5, it'll stop working. I believe that's why most people switch to using their own ingots. Is that correct, or am I off-base on that? My take has always been that stock workorders only really works when you are first getting started.

You can switch to steel pretty early, which goes to 10. 50 workability, low quality, it'll get you there. Some of the folks near cap started using low steel or even high steel, but only for optimizing. 

The switch tonyour own ingots is partly that, also partly folks who ad value to their ingots with plat/gold/niniam etc. for better workorder payouts. Also, armorsmiths.

Raykyn55:
@mikejcook 
This is the setting you might be looking for:
```yaml
workorders_materials:
  fabric_type: linen
  leather_type: cougar
  metal_type: steel
  wood_type: pine
  stone_type: marble
  bone_type: wolf
  knit_type: yarn
  ```
Then just set up a whitelist using:
```yaml
workorder_recipes:
  blacksmithing:
    - lockpick ring
    - razor
    - flask
    - clippers
    - brazier
    - crown
    - rasp
    - pliers
    - slender metal awl
    - mask
    - torque
    - headdress
    - sieve
    ```

mikejcook:
Oh, I'm picking up the disconnect. So people generally whitelist the recipes that are 10 volume or under. I remember now, I did upgrade to steel eventually, but ran into an issue when I got over 10 volume items, there wasn't a solution at the time (big volume steel now, maybe). So that's why I tweaked mine to create my own ingots. But that's fair, you could just whitelist all of the recipes. You know, it might be nice to add a setting workorder_max_volume or something where if the item being requested for the workorder is over that amount, it just automatically gets skipped over. It could default to 99. Then people could just set that to 5 or 10, based on their metal type, and not have to worry about the recipe list.

Raykyn55:
> Oh, I'm picking up the disconnect. So people generally whitelist the recipes that are 10 volume or under. I remember now, I did upgrade to steel eventually, but ran into an issue when I got over 10 volume items, there wasn't a solution at the time (big volume steel now, maybe). So that's why I tweaked mine to create my own ingots. But that's fair, you could just whitelist all of the recipes. You know, it might be nice to add a setting workorder_max_volume or something where if the item being requested for the workorder is over that amount, it just automatically gets skipped over. It could default to 99. Then people could just set that to 5 or 10, based on their metal type, and not have to worry about the recipe list.

As they expand the big ingots in more forges, the most likely result will be transitioning over to those for larger orders. That's the strategy I went with for the taskmaster script when they released them in crossing.

Raykyn55:
To utilize your own materials, eg not purchased lumber/cloth/bone, you'll bring the workorders_materials setting into your yaml and add a custom setting under workorders_materials like the following:
```yaml
workorders_materials:
  custom:
    lumber: teak
    ingot: steel
  fabric_type: linen
  leather_type: cougar
  metal_type: steel
  wood_type: pine
  stone_type: marble
  bone_type: wolf
  knit_type: yarn
  ```
Does not work for stone or yarn, and obviously not either remedies or artificing. Can add as many or as few as you like, and if you are out of the custom materials, it will revert to stock materials as defined by type below custom.
  
This removes the need for (and use of) the setting use_own_ingot_type.

Raykyn55:
New argument 'train' allows you to skip logging, selects a recipe from a curated list, calculates a quantity based on current mindstates(25+ it doesn't even start), and disposes of the finished items. This would take the place of craft.lic for those not interested in prestige.

Raykyn55:
Added training_monitor method similar to how hunting-buddy works, which will monitor experience in a craft while crafting, and exit upon mindlock. 

tweaked materials handling

Raykyn55:
going to repost, another one got dragged into my bad githubbing.

PR #6284, opened 2022-12-13
[script][shape] args + finish

Raykyn55:
requires #6282


Raykyn55:

PR #6285, opened 2022-12-14
Create new script to install the latest Prime map into a non-Prime in…

jstrobush:
…stance.

Blatant copy and paste job of the map management logic in repo with needed short circuiting.

Will grab the latest Prime map from repo, and install it into the data dir for the instance you're logged into.
Will bail if you try to run it in Prime (no need to have this tool where the auto-update works properly)

Tested in:
* Platinum, Mac
* The Fallen, Windows

### DIFF
--- a/download-prime-map.lic
+++ b/download-prime-map.lic
@@ -1,0 +1,275 @@
+
+custom_require.call(%w[common])
+
+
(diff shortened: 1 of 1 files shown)


Kiriawen:
image
image
Tested in TF.  Passed all checks.  I'll call this one good and Mergable.

PR #6289, opened 2022-12-14
[script][common-crafting] Update parts list

Raykyn55:
requires #6280 

changes 'blue flowers' to 'crushed flowers'. Still can't purchase, but now matches new recipes.

### DIFF
--- a/common-crafting.lic
+++ b/common-crafting.lic
@@ -264,10 +264,12 @@ module DRCC
       total += (stock_to_order*material['stock-value'])
     end
 
-    if parts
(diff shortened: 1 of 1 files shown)


Dartellum:
You an purchase red flowers and they turn into crushed flowers but it is red flowers to purchase. Test, test again, test more, and when you think you are done testing, test again. :smile:

Raykyn55:
> You can purchase red flowers and they turn into crushed flowers but it is red flowers to purchase. Test, test again, test more, and when you think you are done testing, test again. 😄

Red flowers become dried red flowers and go in the mortar, blue flowers become crushed flowers and go in the bowl. Two different processes, is there a recipe that flips that?

Dartellum:
> 

Now I need to remember if one does that. I am pretty sure one did use red flowers in mortar.

Raykyn55:

PR #6295, opened 2023-01-02
[script][combat-trainer] Tweak warhorn use

Raykyn55:
Stopping/restarting CT after successful warhorn use can lead to a time-wasting loop where you perform one action, then attempt to blow the warhorn over and over.

The changes:

1. Eliminate the loop. Failing to use the warhorn(lungs too tired) triggers the cooldown.
2. Move stow/wield for offhand weapon into warhorn get/stow.

We don't want to sheath/wield our offhand unless we're actually going to use the warhorn, and we don't want to do it at all if we're using the egg (because you can invoke it without holding it). Moving that into warhorn use, the stow before getting the warhorn, and the retrieve after stowing the warhorn, means we won't stow/retrieve weapons just to check if we're ready to use warhorn/egg.


Raykyn55:
I've been using this since I put it into draft. working well.

PR #6302, opened 2023-01-13
[scripts][magic-training] Isolating class variables into the actual class

MahtraDR:
This is required for Ruby3, as it's a lot more fussy about class variables outside of a class... No change in functionality, and it works as is like this.

### DIFF
--- a/magic-training.lic
+++ b/magic-training.lic
@@ -6,13 +6,34 @@ custom_require.call(%w[common common-arcana common-healing common-moonmage drinf
 
 # Track how long we've been training and how many mindstates we increase at the end.
 # These are defined with @@ so that they can be referenced in before_dying block.
-@@time_in = Time.now
-@@aug_in = DRSkill.getxp('Augmentation')
-@@ward_in = DRSkill.getxp('Warding')
-@@util_in = DRSkill.getxp('Utility')
-@@sorc_in = DRSkill.getxp('Sorcery')
 
 class MagicTraining
+
+  @@time_in = Time.now
+  @@aug_in = DRSkill.getxp('Augmentation')
+  @@ward_in = DRSkill.getxp('Warding')
+  @@util_in = DRSkill.getxp('Utility')
+  @@sorc_in = DRSkill.getxp('Sorcery')
+
+  def self.time_in
+    @@time_in
+  end
+
+  def self.aug_in
+    @@aug_in
+  end
+
+  def self.ward_in
+    @@ward_in
+  end
+
+  def self.util_in
+    @@util_in
+  end
+
+  def self.sorc_in
+    @@sorc_in
+  end
 
   def initialize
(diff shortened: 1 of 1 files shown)


MahtraDR:
@rpherbig Is this a sane/reasonable way to access class variables from outside the class?

Raykyn55:
What's the difference between:
```ruby
  def self.time_in
    @@time_in
  end
  ```
and
```ruby
  def self.time_in
    Time.now
  end
```
if we're just using the method from outside the class?

MahtraDR:
> What's the difference between:
> 
> ```ruby
>   def self.time_in
>     @@time_in
>   end
> ```
> 
> and
> 
> ```ruby
>   def self.time_in
>     Time.now
>   end
> ```
> 
> if we're just using the method from outside the class?

The first instance sets Time.now to the time the variable was initialized, and any time the def is called, it returns that fixed time.

The second instance return `Time.now` at the time the def is called each time.

Raykyn55:
> > What's the difference between:
> > ```ruby
> >   def self.time_in
> >     @@time_in
> >   end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```ruby
> >   def self.time_in
> >     Time.now
> >   end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > if we're just using the method from outside the class?
> 
> The first instance sets Time.now to the time the variable was initialized, and any time the def is called, it returns that fixed time.
> 
> The second instance return `Time.now` at the time the def is called each time.

Ah, because the script is still running, so it doesn't run a new version and redefine the variable. In that case, does it need to be a global variable? If you're fetching it from a class, could it be a class variable?

Raykyn55:
something like:
```ruby
  @time_in = Time.now
  @aug_in = DRSkill.getxp('Augmentation')
  @ward_in = DRSkill.getxp('Warding')
  @util_in = DRSkill.getxp('Utility')
  @sorc_in = DRSkill.getxp('Sorcery')

  attr_reader :time_in, :aug_in, :ward_in, :util_in, :sorc_in
  ```

MahtraDR:
> something like:
> 
> ```ruby
>   @time_in = Time.now
>   @aug_in = DRSkill.getxp('Augmentation')
>   @ward_in = DRSkill.getxp('Warding')
>   @util_in = DRSkill.getxp('Utility')
>   @sorc_in = DRSkill.getxp('Sorcery')
> 
>   attr_reader :time_in, :aug_in, :ward_in, :util_in, :sorc_in
> ```

This is an instance variable like this. 

The original, and my version, are both class variables. 

I just changed where they're stored. I wanted to make the smallest change needed, whilst keeping the same functionality.

PR #6307, opened 2023-01-17
[data][base-recipes] New pattern/collapse (repost)

Raykyn55:
Reworked under discipline headers, parts collected into parts array. Resubmit of #6280


Raykyn55:

PR #6315, opened 2023-01-20
[multiple scripts] Renamed Scouting to Instinct

MahtraDR:
As per title. Just a simple rename/search and replace. No functionality changed.


Kiriawen:
Merge it

asechrest:
Wondering about renaming `scouting.lic`. I guess we shouldn't incase folks have it named as such in their yamls?

MahtraDR:
> Wondering about renaming `scouting.lic`. I guess we shouldn't incase folks have it named as such in their yamls?

There is a way to rename the script and create a backwards compatible alias for scouting to point to instinct. If we can additionally add a deprecation notice, that would make it clean. Or add a note now to scouting.lic that it'll be renamed on X date and we just rename it then.

asechrest:
Now that I typed that, I bet the command to use a trail is still `scout trail`, so there is an argument that the script is designed to scout a trail and `scouting.lic` isn't a misnomer.

PR #6319, opened 2023-01-20
Update stale.yml

asechrest:
After discussion in dev group chat, turning off auto-close for now.

Also changed the label when marking as stale, and changed the auto-comment to prod the issuer to comment if the problem still occurs.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -2,14 +2,14 @@
 daysUntilStale: 185
 
 # Number of days of inactivity before a stale issue is closed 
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -2,25 +2,24 @@
 daysUntilStale: 185
```
MahtraDR, line 2:
```suggestion
daysUntilStale: false
```

MahtraDR:
I've disabled this script for now.

PR #6332, opened 2023-01-24
[scripts][moonwatch] Add uservar for moonbot mode

MahtraDR:
Lets user set a uservar for correct mode, so you can autostart moonwatch for that mode (since autostarts don't normally allow for args)

### DIFF
--- a/moonwatch.lic
+++ b/moonwatch.lic
@@ -170,6 +170,6 @@ loop do
   update_moon_info(get_all_moon_data)
   update_sun_info(get_all_moon_data)
   update_moon_window if CharSettings['moon_window']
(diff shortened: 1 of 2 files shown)


moonwatch.lic
```diff
@@ -170,6 +170,6 @@ loop do
   update_moon_info(get_all_moon_data)
   update_sun_info(get_all_moon_data)
   update_moon_window if CharSettings['moon_window']
-  check_for_new_moons if args.correct
+  check_for_new_moons if args.correct || UserVars.moonwatch_correct
```
MahtraDR, line 173:
```suggestion
  check_for_new_moons if args.correct || UserVars.moonwatch_correct || get_settings.moonwatch_correct
```

Issue #3869, opened 2019-09-16
Enchant.lic - imbue isn't recast when it backfires


dakre18:
While enchanting, my first imbue spell backfired, and it didn't attempt to recast it. Might want to make sure it does on the second cast as well, since it will not finish the enchantment without successfully casting imbue.

I do have 'use_harness_when_arcana_locked: true` set, so it charges my cambrinth as well as harnesses mana, but it should still cast it at 100 mana (as i'm trying to do).

************

[enchant]>prep imbue 25
You trace a careful sigil in the air, shaping the pattern of the Imbue spell.

[enchant]>charge my watersilk bag 25
You harness a moderate amount of energy and attempt to channel it into your watersilk bag.
You are able to channel all the energy into the bag.
The watersilk bag absorbs all of the energy.
Roundtime: 3 sec.

[enchant]>charge my watersilk bag 25
...wait 1 seconds.

[enchant]>charge my watersilk bag 25
You harness a moderate amount of energy and attempt to channel it into your watersilk bag.
You are able to channel all the energy into the bag.
The watersilk bag absorbs all of the energy.
Roundtime: 3 sec.

[enchant]>charge my watersilk bag 25
...wait 1 seconds.

[enchant]>charge my watersilk bag 25
You harness a moderate amount of energy and attempt to channel it into your watersilk bag.
You are able to channel all the energy into the bag.
The watersilk bag absorbs all of the energy.
Roundtime: 3 sec.

[enchant]>invoke my watersilk bag 75
You feel fully prepared to cast your spell.

The watersilk bag pulses with Elemental energy.  You reach for its center and forge a magical link to it, attempting to ready 75 energy lines for your use.  Your efforts are mostly successful, but you accidentally attune yourself to 76 of the lines.
Roundtime: 1 sec.

[enchant]>harness 25
You tap into the mana from twenty-five of the surrounding streams and attempt to keep it channeling in a stream around you.
Roundtime: 2 sec.

[enchant]>harness 25
You tap into the mana from twenty-five of the surrounding streams and add it to the appreciable amount already streaming around you.
Roundtime: 2 sec.

[enchant]>harness 25
...wait 1 seconds.

[enchant]>harness 25
You tap into the mana from twenty-five of the surrounding streams and add it to the vast amount already streaming around you.
Roundtime: 2 sec.

You sense a tiny amount of mana slip away from you.

[enchant]>cast basic runestone on brazier
You gesture at a basic basalt runestone.
[...]

PR #6333, opened 2023-01-24
[scripts][buff] Add strict arg to retry till the buff lands

MahtraDR:
Causes buff.lic to retry casting  a buff until the buff lands. If you're casting on the edge of your skill, sometimes concentration (or arcana, in the case of a bad cambrinth invoke) can cause the cast to fail. This will retry until the buff lands.

### DIFF
--- a/buff.lic
+++ b/buff.lic
@@ -13,7 +13,7 @@ class Waggle
         { name: 'list', regex: /list/, optional: true, description: 'list of defined spell sets' },
         { name: 'spells', regex: /\w+/, optional: true, description: 'Spell list to use, otherwise default' },
         { name: 'force', regex: /force/i, optional: true, description: 'Recast spells even if currently active' },
-        { name: 'strict', regex: /strict/i, optional: true, description: 'Keep trying to spells until they stick. BE CAREFUL' }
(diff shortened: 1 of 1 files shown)


MahtraDR:
Would help situations like https://github.com/rpherbig/dr-scripts/issues/3869

buff.lic
```diff
@@ -12,7 +12,8 @@ class Waggle
         { name: 'set', regex: /(set=(\w+))/, optional: true, description: 'list of defined spell sets' },
         { name: 'list', regex: /list/, optional: true, description: 'list of defined spell sets' },
         { name: 'spells', regex: /\w+/, optional: true, description: 'Spell list to use, otherwise default' },
-        { name: 'force', regex: /force/i, optional: true, description: 'Recast spells even if currently active' }
+        { name: 'force', regex: /force/i, optional: true, description: 'Recast spells even if currently active' },
+        { name: 'strict', regex: /strict/i, optional: true, description: 'Keep trying to spells until they stick. BE CAREFUL' }
```
asechrest, line 16:
```suggestion
        { name: 'strict', regex: /strict/i, optional: true, description: 'Keep trying to cast buffs until they stick. BE CAREFUL' }
```

PR #6335, opened 2023-01-24
[scripts][hunting-buddy] Stop combat to go burgle on cooldown

MahtraDR:
As per description.

This allows the user to opt-in to having their hunting buddy session cut short if the burgle timer hits.

### DIFF
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -471,6 +471,9 @@ priority_skills_lower_limit: 6
 # define one or multiple hunting zones (see base-hunting in scripts/data) and the behaviors for those hunts.
 # https://github.com/rpherbig/dr-scripts/wiki/YAML-Basics:-Part-2---Creating-Your-Own-YAML
 hunting_info:
+# Toggle to cause hunting buddy to stop a hunt short when the burgle
+# timer hits.
+stop_to_burgle: false
 # a list of possible messages to say when you enter a room and don't see anyone.  If someone responds you will leave the room.
 empty_hunting_room_messages:
 - Is anyone else in here?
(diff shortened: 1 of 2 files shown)


profiles/base.yaml
```diff
@@ -471,6 +471,9 @@ priority_skills_lower_limit: 6
 # define one or multiple hunting zones (see base-hunting in scripts/data) and the behaviors for those hunts.
 # https://github.com/rpherbig/dr-scripts/wiki/YAML-Basics:-Part-2---Creating-Your-Own-YAML
 hunting_info:
+# Toggle to cause hunting buddy to stop a hunt short when the burgle
+# timer hits.
+stop_to_burgle: false
```
jbrock24, line 476: Since this is only used by hunting_buddy, we could make it a part of `hunting_info` like we do with `boxes` and `stop_on` and `stop_on_low`, etc.

jbrock24 (changes requested):
The setting for this should really be in hunting_info, it isn't used anywhere else and functions like other hunting_buddy-related settings.

jbrock24:
Honestly, quality change. I've been handling it using stop_on_low, but I have a specific routine where this works and this is just better.

profiles/base.yaml
MahtraDR: That's reasonable. I originally coded it just for myself, and didn't give the setting name/location too much thought. :)

MahtraDR: Ahhh. I remember now why I did it this way. I have this setting in my include-main file, which all my characters include in their yamls. One file change, versus 16 (since all my characters have more than one hunting-buddy config file.) This makes me a little conflicted on the notion of putting this in hunting_info, tbh.

MahtraDR: How would you feel about having both options? I'd like to keep my universal setting, but that defaults to false, and I can add a toggle to `hunting_info` as well.

MahtraDR: Have a look at the current iteration? I've kept my global toggle, but added a per-hunt toggle as well.

MahtraDR:
@jbrock24 Any opinions on this version? I can merge as is if not.

PR #6341, opened 2023-01-24
[scripts][burgle] Switching to guard clause instead of integer to string comparison

MahtraDR:
As per title.

Integer to string comparisons no longer work in ruby3.2. I originally was going to just change to `@burgle_room.to_s` but @Raykyn55 suggested a guard clause and it's a better solution.

### DIFF
--- a/burgle.lic
+++ b/burgle.lic
@@ -112,7 +112,7 @@ class Burgle
       DRC.message("You have empty burgle_settings.  These must be set before running.")
       return false
     end
-    if @burgle_room == nil || @burgle_room =~ /\D/
+    if @burgle_room == nil || @burgle_room.to_s =~ /\D/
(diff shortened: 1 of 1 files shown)


Raykyn55:
Does this compare two numbers, or just check if it's a number?

Raykyn55:
If the latter, would the following work in ruby 3?

```ruby
unless @burgle_room.is_a?(Integer)
  msg
end
```

MahtraDR:
> Does this compare two numbers, or just check if it's a number?

It's just checking if it's a number.

> If the latter, would the following work in ruby 3?
> 
> ```ruby
> unless @burgle_room.is_a?(Integer)
>   msg
> end
> ```

image


That is a very good idea. Thank you. I was really narrowly focused on the compatibility, and you're right, that's a better check for this.

PR #6344, opened 2023-01-24
[script][burgle] Fix for bank/dokt rooms

Raykyn55:
If you're in the bank or at dokts' (or perhaps other select rooms) burgle hangs when trying to open your bag:
```
[burgle: Pausing ["t2", "afk"] to run burgle]

[burgle]>open my scavenger's pack

This is probably not the time nor place for that.
You notice your augmented reflex slip away.

[burgle: *** No match was found after 15 seconds, dumping info]
```
The original code was missing the match, but even adding it just means you get the wrong error and a failed burgle:
```
You do not have a burgle_settings:loot_container set/set to container you have.  Loot must have a place to be stored prior to exiting the house, even if dropping loot.
It is very important that you check the documentation before running: https://elanthipedia.play.net/Lich_script_repository#burgle 
This is a dangerous script to run if it's not understood.  If you get caught there are very high fines, and the loss of all your items is possible if you can't pay your debt.
```
Moved the if/or into a case statement where if you're in one of these rooms, moves you to your burgle room and checks settings again:
```
>;burgle start rope
--- Lich: burgle active.
[burgle: Pausing [] to run burgle]
[burgle]>open my scavenger's pack
This is probably not the time nor place for that.
>
--- Lich: go2 active.
[go2: ETA: 0:00:03 (17 rooms to move through)]
...TRAVEL...
[go2: travel time: 0:00:01]
--- Lich: go2 has exited.
[burgle]>open my scavenger's pack
That is already open.
>
[burgle]>burgle recall
You take a moment to think, you believe...
You should wait at least 16 roisaen for the heat to die down.
>
[burgle: Unpausing [], burgle has finished.]
--- Lich: burgle has exited.
```
Post fix normal burgle not in burgle room:
```
>;burgle start rope
--- Lich: burgle active.
[burgle: Pausing [] to run burgle]
[burgle]>open my scavenger's pack
That is already open.
>
[burgle]>burgle recall
You take a moment to think, you believe...
You should wait at least 7 roisaen for the heat to die down.
>
[burgle: Unpausing [], burgle has finished.]
--- Lich: burgle has exited.
```

### DIFF
--- a/burgle.lic
+++ b/burgle.lic
@@ -124,7 +124,7 @@ class Burgle
       case DRC.bput("open my #{@loot_container}", 'already open', '^You.+open', '^You spread your arms, carefully holding your bag well away from your body', 'Please rephrase that command', 'What were you referring', 'You can\'t do that', 'This is probably not the time nor place for that')
       when 'This is probably not the time nor place for that'
         if @follow
-          DRC.message("Couldn't verify your bag due to room restrictions. Allowing the script to continue assuming your bag is open and available.")
+          DRC.message("Couldn't verify your bag due to room restrictions. Allowing the script to continue assuming your #{@loot_container} is open and available.")
           return true
         else
           travel_to_burgle_room
--- a/burgle.lic
+++ b/burgle.lic
@@ -121,7 +121,11 @@ class Burgle
       return false
     end
     if @burgle_settings['max_search_count'] > 0
-      if /(already open|^You.+open|^You can't do that)/ !~ DRC.bput("open my #{@loot_container}", 'already open', '^You.+open', '^You spread your arms, carefully holding your bag well away from your body', 'Please rephrase that command', 'What were you referring', 'You can\'t do that')
+      case DRC.bput("open my #{@loot_container}", 'already open', '^You.+open', '^You spread your arms, carefully holding your bag well away from your body', 'Please rephrase that command', 'What were you referring', 'You can\'t do that', 'This is probably not the time nor place for that')
+      when 'This is probably not the time nor place for that'
+        DRCT.walk_to(@burgle_room)
+        return valid_burgle_settings?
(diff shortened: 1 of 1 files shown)


asechrest:
I don't know if force traversing to the burgle room is the right way to go, here. Farther down in the script we actually avoid walking to the burgle room if we're in a group. I've never used burgle in a group, though. We also have a check to make sure the traverse was successful in that section. We could duplicate those checks here, I guess.

https://github.com/rpherbig/dr-scripts/blob/27fffc9580c5f99d1eb623e5ab5fe13ea496f84d/burgle.lic#L232

Thoughts?

Raykyn55:
> 

I agree, that wouldn't work, but I can't see this being a scenario that is ever a problem. It will only send you to your burgle room if the settings check occurs while you're in a bank or other non-permitted room, which will never be an actual burgle room. 

In the unlikely event that your group dashes to the bank, then you run burgle follow, and it yoinks you out of the group and sends you to your burgle room, I'd say lesson learned, don't run burgle until you're ready to burgle.

If, though, we did want to sort that out, we could just skip the bag check altogether if we're following, and if we fail to stow, burgle can handle that. The problem is, minus moving, there's no way to resolve this. You just can't open your bag in these rooms.

Raykyn55:
Failure to travel could be teased out into it's own method, instead of duplicating the code. I'll put a commit together for that.

Raykyn55:
Updated to address both concerns @asechrest . This time the creep is your fault :D

asechrest:
Can I get an example room where this message happens? I tried the Shard bank, Crossing bank, Dokts, The Healerie.

Raykyn55:
> Can I get an example room where this message happens? I tried the Shard bank, Crossing bank, Dokts, The Healerie.

https://github.com/rpherbig/dr-scripts/pull/6344#issue-1555527948

Might be limited to the hitman's style backpacks.

PR #6346, opened 2023-01-24
[script][burgle] Reasons

Raykyn55:
So, the original logic here only used the argument if it was defined. converting whatever is there to an integer turns nil into 0, so it was setting our burgle room to 0. This only converts if the argument is defined, otherwise uses burgle room from our yaml settings.


MahtraDR:
Next time let me make my teeny change that doesn't break anything. :D

I just saw lnet, that's pretty funny.

Raykyn55:
> 

This is so much more instructive, though!

Issue #3989, opened 2019-10-25
Mining buddy pickaxe vs shovel


DryIce41:
Would it be possible and simple to set up a way to use a pickaxe when clearing rooms with veins that are not in your keep list, but a shovel when the vein is? This will make clearing rooms via mining buddy to go faster and reset spawns.

MahtraDR:
@Raykyn55 whatcha think?

MahtraDR:
Closing as very stale. Happy to revisit if it still exists today.

PR #6347, opened 2023-01-25
Pickaxe for clearing

Raykyn55:
response to #3989 

Idea here was adding the option to use a pickaxe to clear out rooms, and a shovel if we find a material we want to keep. As usual, at least when I poke at these things, what started as a small change balloons into a partial rewrite. Thankfully, most of the code already present was accomidating, and while the changelog looks extensive, the edits were pretty minor. Aside from repair, most of the changes were ordering and rearranging. Default behavior for current users remains unchanged in form/function, albeit quicker.

Changes to mine.lic:
1. add implement arg, optionally allowing you to define a mining implement when you run mine, or allowing mining-buddy to select an implement based on your settings and the results of the prospect.
2. moved flag creation for 'resource-level' to mining buddy, if we're running mining buddy, so skip the (now) redundant second prospect when mine starts. default resource level to 2 when just running mine by itself.

Changes to mining-buddy.lic:
1. add flag from mine, so that we pick up resource level when we prospect there, and mine can use that information.
2. add yaml setting 'pickaxe' where user can define their pickaxe for clearing operations
3. checks for defined pickaxe on our person, and sets to default implement if not found
4. reworked repair to accept an array of tools. Previously, we only used one implement, but since we might be using one, or two, we want to repair intelligently.
5. added logic to select a tool based on the presence or absence of a metal/stone we want to keep.

Added notation for changes where appropriate.
Added messaging for repair status and exits


Raykyn55:
So, all of the above, converted to a loop instead of calling the same method from within itself.  Additionally, added 'clear' arg, optional, to skip prospect careful when we're just trying to clear out a room. Reworked mining-buddy to make use of that.

Raykyn55:
Making this a draft until repair is sorted.

Raykyn55:
requires #6349 and #6363

PR #6348, opened 2023-01-26
[scripts][common-items] Fix DRCI.dispose_trash bugs

MahtraDR:
I want to deprecate the duplicate that is DRCT.dispose...

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -358,7 +358,7 @@ module DRCI
       when /^Perhaps you should be holding that first/
         return (DRCI.get_item?(item) && DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb))
       when *@@drop_trash_retry_patterns
-        return true if DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb)
+        return DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb)
       when *@@drop_trash_success_patterns
         if worn_trashcan_verb
           DRC.bput("#{worn_trashcan_verb} my #{worn_trashcan}", *@@worn_trashcan_verb_patterns)
--- a/common-items.lic
+++ b/common-items.lic
@@ -288,7 +288,7 @@ module DRCI
     /^You slowly open/,
     /^The .* opens/,
     /^You unbutton/,
-    /is already open/,
+    /(It's|is) already open/,
     /^You spread your arms, carefully holding your bag well away from your body/
   ]
 
--- a/common-items.lic
+++ b/common-items.lic
@@ -413,8 +413,8 @@ module DRCI
     end
 
     # No trash bins or not able to put item in a bin, just drop it.
-    case DRC.bput("drop my #{item}", @@drop_trash_success_patterns, @@drop_trash_failure_patterns, @@drop_trash_retry_patterns, /^Perhaps you should be holding that first/)
-    when /^Perhaps you should be holding that first/
+    case DRC.bput("drop my #{item}", @@drop_trash_success_patterns, @@drop_trash_failure_patterns, @@drop_trash_retry_patterns, /^Perhaps you should be holding that first/, /^But you aren't holding that/)
+    when /^Perhaps you should be holding that first/, /^But you aren't holding that/
       return (DRCI.get_item?(item) && DRCI.dispose_trash(item))
     when *@@drop_trash_retry_patterns
       return dispose_trash(item)
--- a/common-items.lic
+++ b/common-items.lic
@@ -350,7 +350,7 @@ module DRCI
 
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
-  return unless DRCI.get_item_if_not_held?(item)
+    return unless DRCI.get_item_if_not_held?(item)
 
     if worn_trashcan
       case DRC.bput("put my #{item} in my #{worn_trashcan}", @@drop_trash_retry_patterns, @@drop_trash_success_patterns, @@drop_trash_failure_patterns, /^Perhaps you should be holding that first/)
--- a/common-items.lic
+++ b/common-items.lic
@@ -348,7 +348,7 @@ module DRCI
   # TRASH ITEM
   #########################################
 
-  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil, trash_room = nil)
+  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
 
     DRCT.walk_to(trash_room) unless trash_room.nil?
--- a/common-items.lic
+++ b/common-items.lic
@@ -349,7 +349,7 @@ module DRCI
   #########################################
 
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
-    return if item.nil?
+    return unless item
     return unless DRCI.get_item_if_not_held?(item)
 
     if worn_trashcan
--- a/common-items.lic
+++ b/common-items.lic
@@ -350,7 +350,6 @@ module DRCI
 
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
-
         return unless DRCI.get_item_if_not_held?(item)
 
     if worn_trashcan
--- a/common-items.lic
+++ b/common-items.lic
@@ -985,7 +985,7 @@ module DRCI
   #########################################
 
   def open_container?(container)
-    case DRC.bput("open my #{container}", @@open_container_success_patterns, @@open_container_failure_patterns)
+    case DRC.bput("open #{container}", @@open_container_success_patterns, @@open_container_failure_patterns)
     when *@@open_container_success_patterns
       return true
     end
--- a/common-items.lic
+++ b/common-items.lic
@@ -5,7 +5,7 @@
 
 # This module should be 'bottom-level' and only depend on common.
 # Any modules that deal with items and <something> should be somewhere else
-custom_require.call(%w[common])
+custom_require.call(%w[common common-travel])
 
 module DRCI
   module_function
@@ -288,7 +288,7 @@ module DRCI
     /^You slowly open/,
     /^The .* opens/,
     /^You unbutton/,
-    /(It's|is) already open/,
+    /is already open/,
     /^You spread your arms, carefully holding your bag well away from your body/
   ]
 
@@ -348,8 +348,10 @@ module DRCI
   # TRASH ITEM
   #########################################
 
-  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
+  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil, trash_room = nil)
     return if item.nil?
+
+    DRCT.walk_to(trash_room) unless trash_room.nil?
 
     if worn_trashcan
       case DRC.bput("put my #{item} in my #{worn_trashcan}", @@drop_trash_retry_patterns, @@drop_trash_success_patterns, @@drop_trash_failure_patterns, /^Perhaps you should be holding that first/)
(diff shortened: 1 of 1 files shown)


MahtraDR:
Tested with 
Goes to 9459, runs DRCI.dispose_trash and puts stuff in the hollow.
```
;eq settings=get_settings; DRCI.dispose_trash('grass', nil, nil, 9459)
```
and
Goes to 9459, runs DRCI.dispose_trash and puts stuff in your worn bucket.
```
;eq settings=get_settings; DRCI.dispose_trash('grass', settings.worn_trashcan, settings.worn_trashcan_verb, 9459)
```
and
stays in place, and trashes the grass
```
;eq settings=get_settings; DRCI.dispose_trash('grass', settings.worn_trashcan, settings.worn_trashcan_verb, nil)
```

Raykyn55:
What about having it test the worn trashcan before moving?

MahtraDR:
> What about having it test the worn trashcan before moving?

The calling script can test. The method just does as it's told. If there is a bucket it'll get used, wherever you tell the script to use it.

MahtraDR:
> What about having it test the worn trashcan before moving?

I can also see a scenario where you would want to go somewhere else to use the worn trashcan. It's quite noisy.

Raykyn55:
> > What about having it test the worn trashcan before moving?
> 
> The calling script can test. The method just does as it's told. If there is a bucket it'll get used, wherever you tell the script to go to use it.

Thinking, perhaps, move the walk_to below the worn_trashcan portion, such that if the trashcan fails, we go to whatever room (if defined) then perform the normal trash routine. So: trashcan, or travel+roomcan, or roomcan where we are, or drop.

MahtraDR:
> > > What about having it test the worn trashcan before moving?
> 
> > 
> 
> > The calling script can test. The method just does as it's told. If there is a bucket it'll get used, wherever you tell the script to go to use it.
> 
> 
> 
> Thinking, perhaps, move the walk_to below the worn_trashcan portion, such that if the trashcan fails, we go to whatever room (if defined) then perform the normal trash routine. So: trashcan, or travel+roomcan, or roomcan where we are, or drop.

I'm trying to picture that. It would nullify my scenario. You wouldn't then be able to define that you want to go to another room to use the bucket.

Raykyn55:
> > > > What about having it test the worn trashcan before moving?
> > 
> > 
> > > 
> > 
> > 
> > > The calling script can test. The method just does as it's told. If there is a bucket it'll get used, wherever you tell the script to go to use it.
> > 
> > 
> > Thinking, perhaps, move the walk_to below the worn_trashcan portion, such that if the trashcan fails, we go to whatever room (if defined) then perform the normal trash routine. So: trashcan, or travel+roomcan, or roomcan where we are, or drop.
> 
> I'm trying to picture that. It would nullify my scenario. You wouldn't then be able to define that you want to go to another room to use the bucket.

could add another arg, like force_travel:
```ruby
def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil, trash_room = nil, force_travel = false)
  DRCT.walk_to(trash_room) if force_travel
```
walk_to already ignores nil.

MahtraDR:
> walk_to already ignores nil.

That's cool. I did not realize that.

I like it as is. We haven't checked if the worn can ever worked before, and that function is pretty robust. I use my bucket a LOT, if it was doing an `exists?` check every time it would actually double how long it takes.

MahtraDR:
I found a bonus bug.
Before this PR:
image

After this PR:
image

Before this PR:
image

After this PR:
image

common-items.lic
```diff
@@ -983,7 +985,7 @@ module DRCI
   #########################################
 
   def open_container?(container)
-    case DRC.bput("open #{container}", @@open_container_success_patterns, @@open_container_failure_patterns)
+    case DRC.bput("open my #{container}", @@open_container_success_patterns, @@open_container_failure_patterns)
```
MahtraDR, line 988:
```suggestion
    case DRC.bput("open #{container}", @@open_container_success_patterns, @@open_container_failure_patterns)
```

common-items.lic
```diff
@@ -691,7 +693,7 @@ module DRCI
   # http://forums.play.net/forums/DragonRealms/Discussions%20with%20DragonRealms%20Staff%20and%20Players/Game%20Master%20and%20Official%20Announcements/view/1899
   def get_item_from_eddy_portal?(item, container)
     # Ensure the eddy is open then look in it to force the contents to be loaded.
-    return false unless DRCI.open_container?('my eddy') && DRCI.look_in_container(container)
+    return false unless DRCI.open_container?('eddy') && DRCI.look_in_container(container)
```
MahtraDR, line 696:
```suggestion
    return false unless DRCI.open_container?('my eddy') && DRCI.look_in_container(container)
```

common-items.lic
```diff
@@ -288,7 +288,7 @@ module DRCI
     /^You slowly open/,
     /^The .* opens/,
     /^You unbutton/,
-    /(It's|is) already open/,
+    /is already open/,
```
MahtraDR, line 291:
```suggestion
    /(It's|is) already open/,
```

common-items.lic
```diff
@@ -348,15 +348,17 @@ module DRCI
...
+    DRCT.walk_to(trash_room) unless trash_room.nil?
+
     if worn_trashcan
       case DRC.bput("put my #{item} in my #{worn_trashcan}", @@drop_trash_retry_patterns, @@drop_trash_success_patterns, @@drop_trash_failure_patterns, /^Perhaps you should be holding that first/)
       when /^Perhaps you should be holding that first/
-        return false unless DRCI.get_item?(item) && DRCI.dispose_trash(item)
+        return (DRCI.get_item?(item) && DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb))
       when *@@drop_trash_retry_patterns
-        return true if dispose_trash(item, worn_trashcan, worn_trashcan_verb)
+        return true if DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb)
```
MahtraDR, line 361:
```suggestion
        return DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb)
```

MahtraDR:
Something is wrong with custom_require. DRCI and DRCT have each other in custom_require, and that causes them to hang with bootstrap.

Raykyn55:
> I found a bonus bug. Before this PR: image
> 
> After this PR: image
> 
> Before this PR: image
> 
> After this PR: image

DRCT.dispose had a get, DRCI.dispose_trash always lacked the get. Why not get_item_if_not_held at the top, for both scenarios?

Raykyn55:
> > walk_to already ignores nil.
> 
> That's cool. I did not realize that. I'll leave it as is in any case, purely because it makes clear what is happening when that value is nil though. I've been caught out too many times with a function that wasn't clear, and this is clear. :)
> 
> I like it as is. We haven't checked if the worn can ever worked before, and that function is pretty robust. I use my bucket a LOT, if it was doing an `exists?` check every time it would actually double how long it takes.

I don't mean check if it works/exists, i mean just use it if it's defined (if worn_trashcan), meaning unless you've got that defined, it's going to skip anyways, but if you do, it's going to use it. YOu'd have a move above the trashcan if you force_travel, and one below if you don't:

```ruby
  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil, trash_room = nil, force_travel = false)
    return if item.nil?

    DRCT.walk_to(trash_room) if force_travel

    if worn_trashcan
      case DRC.bput("put my #{item} in my #{worn_trashcan}", @@drop_trash_retry_patterns, @@drop_trash_success_patterns, @@drop_trash_failure_patterns, /^Perhaps you should be holding that first/)
      when /^Perhaps you should be holding that first/
        return false unless DRCI.get_item?(item) && DRCI.dispose_trash(item)
        return (DRCI.get_item?(item) && DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb))
      when *@@drop_trash_retry_patterns
        return true if dispose_trash(item, worn_trashcan, worn_trashcan_verb)
        return DRCI.dispose_trash(item, worn_trashcan, worn_trashcan_verb)
      when *@@drop_trash_success_patterns
        if worn_trashcan_verb
          DRC.bput("#{worn_trashcan_verb} my #{worn_trashcan}", *@@worn_trashcan_verb_patterns)
          DRC.bput("#{worn_trashcan_verb} my #{worn_trashcan}", *@@worn_trashcan_verb_patterns)
        end
        return true
      end
    end
    
    DRCT.walk_to(trash_room)
  ```
Just to clarify what I meant.

**Comment by asechrest** [comment]
> > I found a bonus bug. Before this PR: image
> > After this PR: image
> > Before this PR: image
> > After this PR: image
> 
> DRCT.dispose had a get, DRCI.dispose_trash always lacked the get. Why not get_item_if_not_held at the top, for both scenarios?

I see a get in `DRCI.dispose_trash`:

https://github.com/rpherbig/dr-scripts/blob/f99da932afc9a98f206029cd7fe6e66668f9c10e/common-items.lic#L403

**Comment by asechrest** [comment]
Oh, I think I see the bug. This line should be passing back in the `worn_trashcan` and `worn_trashcan_verb` params when re-calling `dispose_trash`.

https://github.com/rpherbig/dr-scripts/blob/f99da932afc9a98f206029cd7fe6e66668f9c10e/common-items.lic#L357

**Comment by asechrest** [comment]
Whatever we end up doing, I _don't_ at all like a second dispose method in DRCT just because it travels to dispose. I think `DRCI.dispose_trash` is quite robust and I'd like to see that be our singular dispose method and start moving scripts toward using that. Doing that facilitates some other thoughts in my mind, like allowing a yaml-based dispose blacklist, which `DRCI.dispose_trash` would check against before disposing, and allow another layer of protection against fancy item loss. Even outside of that idea, it makes sense to have one dispose method, not two or more.

File: common-items.lic [hunk 5]
```diff
@@ -5,7 +5,7 @@
 
 # This module should be 'bottom-level' and only depend on common.
 # Any modules that deal with items and <something> should be somewhere else
-custom_require.call(%w[common])
+custom_require.call(%w[common common-travel])
```
MahtraDR, line 8:
```suggestion
custom_require.call(%w[common])
```

common-items.lic
```diff
@@ -348,15 +348,17 @@ module DRCI
   # TRASH ITEM
   #########################################
 
-  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
+  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil, trash_room = nil)
```
MahtraDR, line 351:
```suggestion
  def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
```

common-items.lic
```diff
@@ -351,12 +351,14 @@ module DRCI
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
 
+    DRCT.walk_to(trash_room) unless trash_room.nil?
```
MahtraDR, line 354:
```suggestion
        return unless DRCI.get_item_if_not_held?(item)
```

common-items.lic
```diff
@@ -351,12 +351,14 @@ module DRCI
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
```
MahtraDR, line 353:
```suggestion
```

common-items.lic
```diff
@@ -350,13 +350,14 @@ module DRCI
 
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
+        return unless DRCI.get_item_if_not_held?(item)
```
MahtraDR, line 353:
```suggestion
  return unless DRCI.get_item_if_not_held?(item)
```

common-items.lic
```diff
@@ -350,13 +350,14 @@ module DRCI
 
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
+  return unless DRCI.get_item_if_not_held?(item)
```
MahtraDR, line 353:
```suggestion
    return unless DRCI.get_item_if_not_held?(item)
```

MahtraDR:
Removed the travel. We can't have DRCT in DRCI. I'll add a method to DRCT.

Added a `DRCI.get_item_if_not_held?` to the top. It'll obsolete some of the functionality downstream in the method, but shouldn't break anything. Hasn't yet for me, anyway.

common-items.lic
```diff
@@ -350,13 +350,14 @@ module DRCI
 
   def dispose_trash(item, worn_trashcan = nil, worn_trashcan_verb = nil)
     return if item.nil?
```
MahtraDR, line 352:
```suggestion
    return unless item
```

PR #6349, opened 2023-01-26
[script][common-crafting] Add shop repair to CC

Raykyn55:
A common method, could be used in:
workorders
craft
taskmaster
mining-buddy
forestry-buddy
clean-lumber
clean-leather

Perhaps others? Splits up drop and pickup for flexibility, default is to pickup right after dropping off. Pickup method looks for the tool on your belt settings, finds the belt where it belongs and ties. If no belt found, just stows it in your bag.


common-crafting.lic
```diff
@@ -409,6 +409,36 @@ module DRCC
     end
   end
 
+  def repair_tools(tools, bag, belt, town, settings, pickup = true)
+    tools = tools.to_a
+    repair_info = get_data('town')[town]['metal_repair']
+    DRCT.walk_to(repair_info['id'])
+    tools.each do |tool_name|
+      get_crafting_item(tool_name, bag, nil, belt, true)
+      case DRC.bput("give #{repair_info['name']}", /I don't repair those here/, /What is it/, /There isn't a scratch on that/, /Just give it to me again/, /I will not/, /I can't fix those/)
```
MahtraDR, line 418: How about using DRCI.give_item?, and expanding the messaging there to account for NPCs?

common-crafting.lic
```diff
@@ -409,6 +409,36 @@ module DRCC
...
+        stow_crafting_item(tool_name, bag, belt) if DRCI.in_hands?(tool_name)
+      end
+    end
+    pickup_tools(bag, town, settings) if pickup
+  end
+  
+  def pickup_tools(bag, town, settings)
+    repair_info = get_data('town')[town]['metal_repair']
+    DRCT.walk_to(repair_info['id'])
+    while /You get/ =~ DRC.bput("get my #{repair_info['name']} ticket", /^You get.*#{repair_info['name']}.*ticket/, /^What were you referring to/)
```
MahtraDR, line 432: Why not use DRCI.get_item? here?

MahtraDR:
Man, this functionality, drop off now, go do stuff, pick up later, has been requested of crossing-trainer forever... Want to do it there instead? Maybe we add a flag for `tool:true` for equipment manager instead? We could have categories even. Default to everything, but args for tools, weapons, armor? Or something.

common-crafting.lic
Raykyn55: yeah, that's a much better idea, i didn't think about the stow issue. I don't know what you mean by dealing with npcs thought.

Raykyn55:
> Man, this functionality, drop off now, go do stuff, pick up later, has been requested of crossing-trainer forever... Want to do it there instead? Maybe we add a flag for `tool:true` for equipment manager instead? We could have categories even. Default to everything, but args for tools, weapons, armor? Or something.

Assuming you mean crossing-repair, that's not a bad idea, could expand the pickup to figure it out automatically. Something like, is it something that belongs on a belt? y/n, is it a tool? y/n, is it in gear/worn? y/n, stow.

What about keeping the repair_tools method in CC, but moving the pickup to CI and expanding it, then keeping the drop for crossing-trainer as is, with a toggle. That would require an edit to crossing-trainer, so maybe you meant crossing-trainer after all.

common-crafting.lic
Raykyn55: good idea, I'll change that

common-crafting.lic
Raykyn55: meant this comment on the next review item. This is doable, though..

Raykyn55:
requires #6363

PR #6350, opened 2023-01-26
[script][common-items] Repair routines in common items

Raykyn55:
First method functions the same as crossing-repair script, but allows us to opt out of picking up immediately, so can be worked into scripts like training-manager.

Notably theres no check for cash, though that could be worked in?


Raykyn55:
This is too much for common-items. . .

PR #6351, opened 2023-01-27
[script][combat-trainer][common-arcana][discern] - auto_mana for sorcery

mikejcook:
Changes to combat-trainer, common-arcana, and discern to implement auto_mana for sorcery using the existing approach to symbiosis.

Overall, the changes are pretty minor, since I'm using the existing logic used for symbiosis. Most changes are just adding an additional check for sorcery.

I did add one additional change to combat-trainer: if the spell fails to cast, we decrease the amount of mana and so that we can try again next time with less mana. If we have to do this, I set a flag on the spell ('failed'). I now check the 'failed' property when checking to see if we should increase the mana going into the spell because of the learning threshold. This addresses an issue where casting the spell provides less than the learning threshold, but any more mana causes the spell to fail. The script will get into a cycle where it constantly increases, decreases, and then again increases the mana. If the spell fails, the script will cease increasing for this session.

I've also extended discern in a couple ways:
set: this is a new argument that allows the user to set their mana at a reasonable level. Mahtra noted that the tuning for auto-mana for symbiosis (and now sorcery) is super slow, so this allows the user to set a baseline.
check: this is a simple argument that allows the user to output the discern values they have saved for a particular spell, rather than reading through the entire variable with ;vars list

I've been testing on two different characters - one with a sorcery spell set to auto_mana, one without. I've also tested with auto_mana symbiosis. All test cases worked as I expected, but I am going to evaluate a bit more.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1495,10 +1495,12 @@ class SpellProcess
     if game_state.is_offense_allowed? || @prepared_spell['harmless']
       snapshot = DRSkill.getxp(@skill)
       success = cast?(@custom_cast, @symbiosis, @before, @after)
-      if @symbiosis && @use_auto_mana
+      if (@symbiosis || spell_is_sorcery?(@prepared_spell)) && @use_auto_mana
         if !success
           UserVars.discerns[@abbrev]['more'] = [UserVars.discerns[@abbrev]['more'] - 1, 0].max
-        elsif DRSkill.getxp(@skill) - snapshot < @symbiosis_learning_threshold
+          @prepared_spell['failed'] = true
+          DRC.message("Casting spell #{@abbrev} failed - will no longer automatically increase")
+        elsif DRSkill.getxp(@skill) - snapshot < @symbiosis_learning_threshold && !@prepared_spell['failed']
           UserVars.discerns[@abbrev]['more'] = UserVars.discerns[@abbrev]['more'] + 1
         end
       end
@@ -1855,7 +1857,7 @@ class SpellProcess
   def prepare_spell(data, game_state, force_cambrinth = false)
     return unless data
 
-    data = check_discern(data, @settings) if data['use_auto_mana']
+    data = check_discern(data, @settings, spell_is_sorcery?(data)) if data['use_auto_mana']
     game_state.cast_timer = Time.now
     @prep_time = data['prep_time']
     echo("prepare spell: #{data}") if $debug_mode_ct
--- a/common-arcana.lic
+++ b/common-arcana.lic
@@ -689,17 +689,17 @@ module DRCA
     true
   end
 
-  def check_discern(data, settings)
+  def check_discern(data, settings, spell_is_sorcery = false, more_override = nil)
     UserVars.discerns = {} unless UserVars.discerns
     discern_data = UserVars.discerns[data['abbrev']] || {}
-    if data['symbiosis']
-      if discern_data.empty? || discern_data['min'].nil?
+    if data['symbiosis'] || spell_is_sorcery
+      if discern_data.empty? || discern_data['min'].nil? || more_override
         DRC.retreat
         /requires at minimum (\d+) mana streams/ =~ DRC.bput("discern #{data['abbrev']}", 'requires at minimum \d+ mana streams')
         discern_data['mana'] = Regexp.last_match(1).to_i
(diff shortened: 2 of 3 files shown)


MahtraDR:
I like the idea. I don't like adding more code for sorcery detection. 

We can change/add documentation, but all you have to do to have `spell_is_sorcery?` fall to false is set the spell's `mana_type` to the character's native mana type. 

Like so:

```yaml
  Utility: &utility
    abbrev: Bless
    mana: 66
    mana_type: lunar
    symbiosis: true
    ```

That simplifies the code quite a bit.

**Comment by mikejcook** [comment]
> 



> I like the idea. I don't like adding more code for sorcery detection.
> 
> We can change/add documentation, but all you have to do to have `spell_is_sorcery?` fall to false is set the spell's `mana_type` to the character's native mana type.
> 
> Like so:
> 
> ```yaml
>   Utility: &utility
>     abbrev: Bless
>     mana: 66
>     mana_type: lunar
>     symbiosis: true
> ```
> 
> That simplifies the code quite a bit.

I'm not entirely sure what you mean. I didn't really dig into that method, I just glanced at it and it seems to identify the spells at sorcery, so I just sort of accepted it. I would be pretty nervous to change how we detect sorcery - that would be a pretty big breaking change, I feel like?

**Comment by MahtraDR** [comment]
> I'm not entirely sure what you mean. I didn't really dig into that method, I just glanced at it and it seems to identify the spells as sorcery, so I just sort of accepted it. I would be pretty nervous to change how we detect sorcery - that would be a pretty big breaking change, I feel like?

Ah, no, this is the sorcery check:

```ruby
  def spell_is_sorcery?(spell_data)
    return true  if spell_data['sorcery']
    return false unless spell_data['mana_type']
    return false if DRStats.native_mana.casecmp(spell_data['mana_type']).zero?
    return false if spell_data['mana_type'].casecmp('ap').zero?
    true
  end

```

If you set mana_type to your native spell type in your yaml, your spell won't be treated as sorcery. (this is today, without any code change.)

mikejcook:
> 
> Ah, no, this is the sorcery check:
> 
> ```ruby
>   def spell_is_sorcery?(spell_data)
>     return true  if spell_data['sorcery']
>     return false unless spell_data['mana_type']
>     return false if DRStats.native_mana.casecmp(spell_data['mana_type']).zero?
>     return false if spell_data['mana_type'].casecmp('ap').zero?
>     true
>   end
> ```
> 
> If you set mana_type to your native spell type in your yaml, your spell won't be treated as sorcery. (this is today, without any code change.)

I looked at this with fresh eyes, but I'm not clear on what you're suggesting I change. Could you clarify, please? Thanks!

mikejcook:
I've been running this now on multiple characters for several days and it's working well. I haven't run into any issues with the new code or backwards compatibility issues for synergy or sorcery.

MahtraDR:
Thanks for the update.

Can you give me a simplest usable scenario I can test with? Any chance you could make a brief wiki article on how to use this script with symbiosis learning?

mikejcook:
Sure thing.

Setup a section in your YAML (probably offensive_spells) with a Sorcery spell and set use_auto_mana to true. Choose a character where your allowable prep is going to be at least a bit more than the minimum. 100+ ranks of Sorcery should do it, I would think.

Run discern by itself, and it should discern the sorcery spell and add it to your discerns user var with the values of min = the minimum spell prep and more = 0.

If you go and run combat-trainer, each time it casts that spell, it's going to check your learning rate against the threshold (which is 2 by default - symbiosis_learning_threshold). Assuming it doesn't hit 2, and it shouldn't, since we're starting at min prep, it'll increase your 'more' value by 1 and update the actual cast settings (mana and cambrinth).

It will keep increasing until you hit a level that trains more than the threshold (or is 1 less than a level that backfires).

You can also use ;discern set calm 18  (for example, if the spell you wanted to update was calm), to increase the mana cost without having to step 1-by-1 through combat-trainer. This runs the exact same logic as combat-trainer (which just calls common-arcana-discern, which knows to perform special handling on symbiosis spells (and now also sorcery spells)) and sets the more cost to whatever you specify (minus the base mana cost) and then sets the mana and cambrinth values appropriately.

I'll go ahead and create a wiki article describing use_auto_mana in general.

mikejcook:
https://github.com/rpherbig/dr-scripts/wiki/Auto-Mana

PR #6361, opened 2023-02-02
[script][common-items] Expand give_item

Raykyn55:
Per suggestion by @MahtraDR for use with repair, expanded get_item? method in common-items to include handling of repairing gear, returning repair tickets. Base functionality retained.

PR #6362, opened 2023-02-02
Create slim-repair.lic

Raykyn55:
A repair script similar to crossing-repair, but with some added utility, minus certain details.

Added:
1. accepts arguments drop, pickup, or no arguments at all. 
2. Drop will drop your gear off at the repair shop, then save that repair location in UserVars, and exit.
3. Pickup will use your last repair location info from UserVars to pick up repaired gear, and exit.
4. Town argument is only used for the drop. If it has a repair location saved from a previous drop, pickup will ignore this and go to that saved town (assuming you have a ticket from there).
5. last town information is reset upon successful pickup, or upon finding no repair tickets for that location.
6. Gear stowing is a little more detailed to handle tools. If you normally repair your tools via regular repair shop (instead of after crafting), the stow method will attempt to secure it to a crafting belt, or put it in a crafting container, before simply stowing it like any other piece of gear.

Not in slim-repair but present in crossing-repair:
1. Does not repair by heuristic, eg checking condition of key pieces. Unless you set a timer, this will attempt to repair every time it runs.
2. Does not deposit excess coins between drop and pickup
3. No performance training.

This is a different type of repair script than crossing-repair, more of a t2 style script that gives you more control over how you choose to repair. 

requires #6361 as this expanded method is used for both drop and pickup.


MahtraDR:
I have mixed feelings here. Does this effectively replace crossing-repair? Should we maybe just call it `repair.lic` or would it make sense to add this functionality into crossing-repair? It's been a long requested feature to split drop off and collection.

MahtraDR:
Can we also capture and store the duration of the longest repair item, then when we call the pickup portion provide an option to skip that call until that longest duration has passed.

Raykyn55:
> I have mixed feelings here. Does this effectively replace crossing-repair? Should we maybe just call it `repair.lic` or would it make sense to add this functionality into crossing-repair? It's been a long requested feature to split drop off and collection.

Short answer, no. I don't think this replaces crossing-repair. The heuristic, for one, I imagine is still used by some people. Yaml settings like force, override, the split between leather/metal repair. Folks who aren't interested in the split, or don't use t2, will probably want to stick with crossing-repair.

> Can we also capture and store the duration of the longest repair item, then when we call the pickup portion provide an option to skip that call until that longest duration has passed.

Could, but I don't think we should. For example, how I'm using it for my yaml:
```yaml
- skill:
  - Evasion
  - Small Edged
  start: 35
  scripts:
  - checkfavors
  - burgle start 
  - multi 1,app my green pouch quick
  - hunting-buddy main
  - slim-repair drop
  - safe-room
  - burgle start
  - sell-loot
  - slim-repair pickup
  - appraisal focus shield
  - athletics max wyvern
  - outdoorsmanship
```

So, finish the hunt, drop off, but i want that gear before the appraisal focus, and certainly before I leave this block. Additionally, folks who repair tools via gear are going to want those tools back before their crafting. I'm not sure having the script not pick up your gear is a good idea, even with an opt-in like that. 

That said, not saying no to either. If the consensus is a flat rewrite of crossing-repair, I can load in what's missing. If the people want a pickup timer, that's not impossible either.

Raykyn55:
see #6570

PR #6363, opened 2023-02-03
[script][common-crafting] get/stow methods return success/failure

Raykyn55:
Updating get/stow methods to return true/false instead of nil on successful operation:
```
>;e echo DRCC.get_crafting_item('basic stamp', 'clouds', nil, nil, true)
--- Lich: exec1 active.
[exec1]>get my basic stamp
You get a basic glaes stamp from inside your featherlette pack.
>
[exec1: true]
```
Also handles belt mismatches, where you have a tool listed but that belt won't accept that tool:
```
>;e echo DRCC.stow_crafting_item("basic stamp", 'clouds', get_settings.outfitting_belt)
--- Lich: exec1 active.
[exec1]>tie my basic stamp to my outfitter's belt
The stamp doesn't seem to fit.
>
[exec1]>put my basic stamp in my clouds
You put your stamp in your enshrouding clouds.
>
[exec1: true]
```
When a belt has run out of room:
```
>;e echo DRCC.stow_crafting_item("set bellows", 'clouds', get_settings.engineering_belt)
--- Lich: exec1 active.
[exec1]>tie my set bellows to my agonite toolstrap
There's no more free ties on the agonite toolstrap for the set of bellows.  Try untying something first!
>
[exec1]>put my set bellows in my clouds
You put your bellows in your enshrouding clouds.
>
[exec1: true]
```
Finally, when missing a tool (and skip_exit is passed):
```
>;e echo DRCC.get_crafting_item('set bellows', 'clouds', nil, get_settings.forging_belt, true)
--- Lich: exec1 active.
[exec1]>untie my set bellows from my forger's belt
Untie what?
>
[exec1]>get my set bellows
What were you referring to?
>
Merchant Prince Nyrhin just arrived.
>
[exec1: ]
[exec1: You seem to be missing: set bellows]
[exec1: false]
```


MahtraDR:
I had these methods on my radar actually. Even with this, these are reinventing the wheel. We have tie and untie methods, and we have get/stow methods in DRCI. If we must have additional versions of these in DRCC, they should at least make use of the existing methods.

Raykyn55:
> 

True, but in some cases, we want more than just true/false from our bput. I did collect a portion of stow into put_away_item where appropriate(line 263). The only purpose of this PR is to tell the calling script whether the job was completed or not. Did we get the item? Did we stow the item?

There's a variety of instances in which get/untie would return false, but which get_crafting_item has a solution, so replacing most of what's there with generic common-items methods would be a regression. Retaining the function would require edits to those common methods that would break their utility for other scripts/methods. Moreover, loading that function into common-items would require more than just what's present in the common module. Take untie:

```ruby
  def tie_item?(item, container)
    case DRC.bput("tie my #{item} to my #{container}", @@tie_item_success_patterns, @@tie_item_failure_patterns)
    when *@@tie_item_success_patterns
      return true
    else
      return false
    end
  end
```

If we return a string instead of false, in order to catch that information for our crafting method, methods that were expecting a false are going to get a truthy response. If we add handling for failure patterns inside common-items eg too wounded, we'd need travel.

Raykyn55:
requires #6374

Raykyn55:
I believe this is closer to what you were looking for @MahtraDR ? Strips out the fluff, uses items methods where possible.

Raykyn55:
ah poo, have to redo this one haha

PR #6365, opened 2023-02-03
[script][common-crafting] Get/stow passthrough

Raykyn55:
Since the new methods answer a question, they ought to reflect that. However, changing the method name before changing all instances of the method's use to the new name would be a problem. This method is used a LOT, both on and off the repo, so adding the pass through would make it backwards compatibible while giving the method a name that matches it's new attributes.

Making this a separate PR as a draft suggestion. Implementation of the new method would be done on a case-by-case basis where appropriate, so that instead of just changing every instance over, we can evaluate how best to utilize the *fact* of success/failure in each script.

An example of it's use:
```ruby
DRCC.get_crafting_item(tool_name, bag, bag_items, belt, true) #attempts to fetch the next tool, with the option (true) to continue if fetch fails
next unless DRC.right_hand #if we don't get the next tool for whatever reason, we move on to the next tool.
```
Here we call the get, but without knowing whether it succeeded, we have to follow up with verification from another method. Most uses of the original method just assumed it succeeded, which was why a hard exit was written into the method to begin with. Once those uses are brought over to the new method, at some point the hard exit won't be necessary.


Raykyn55:
use of the old method name, new method pass through:
```
>;e echo DRCC.stow_crafting_item("knitting needles", 'pack', nil)
--- Lich: exec1 active.
[exec1]>put my knitting needles in my pack
You put your needles in your scavenger's pack.
>
[exec1: ]
```
use of the new method name:
```
>;e echo DRCC.stow_crafting_item?("knitting needles", 'pack', nil)
--- Lich: exec1 active.
[exec1]>put my knitting needles in my pack
You put your needles in your scavenger's pack.
>
[exec1: true]
```
new method name, not in our hands:
```
>;e echo DRCC.stow_crafting_item?("knitting needles", 'pack', nil)
--- Lich: exec1 active.
[exec1: true]
```

PR #6368, opened 2023-02-05
[script][help_me] - Create script help_me

asechrest:
Pulls the `help_me` functionality out of dependency and into a script. Also makes a couple of changes to original functionality:

No longer requires lnet. If lnet isn't running, it'll echo the pastebin URL to the game window. Alternatively, you can call the script with an arg that specifies a name to private message the URL to on LNET.

### DIFF
--- a/help_me.lic
+++ b/help_me.lic
@@ -1,0 +1,83 @@
+require 'net/http'
+
+class HelpMe
+
+  def initialize
+
+    arg_definitions = [
+      [
+        { name: 'user', regex: /\b\w+\b/i, optional: true, description: 'Send help_me to a specific user.' },
+        { name: 'script_summary', optional: true, description: 'Script that sends installation and yaml information for user help purposes.'}
+      ]
+    ]
+    args = parse_args(arg_definitions)
+
+    @paste_bin_token = 'dca351a27a8af501a8d3123e29af7981'
+    @paste_bin_url = 'https://pastebin.com/api/api_post.php'
+
+    help_me(args.user)
+  end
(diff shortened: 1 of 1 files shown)


asechrest:
I'll remove the help me stuff from dependency after this is merged.

MahtraDR:
Can we include autostarts here? Both lich autostarts, and dependency autostarts?

asechrest:
> Can we include autostarts here? Both lich autostarts, and dependency autostarts?

I'm aware of that request. I'd like to work on it as a separate PR. It's more difficult than it seems. `autostart.lic` is ancient and not easily accessible to a DR script. I even tried stealing code from it to rebuild the core auto start list for display in this script but it doesn't work the same way because this is a DR script inside a class.

So I'll have to do some more exploration. There is also a desire by the Lich5 folks to make core auto starts part of core Lich.

MahtraDR:
> > Can we include autostarts here? Both lich autostarts, and dependency autostarts?
> 
> 
> 
> I'm aware of that request. I'd like to work on it as a separate PR. It's more difficult than it seems. `autostart.lic` is ancient and not easily accessible to a DR script. I even tried stealing code from it to rebuild the core auto start list for display in this script but it doesn't work the same way because this is a DR script inside a class.
> 
> 
> 
> So I'll have to do some more exploration. There is also a desire by the Lich5 folks to make core auto starts part of core Lich.

Suits me.

MahtraDR:
Tested. Looks good to me. Merging.

PR #6374, opened 2023-02-06
[script][common-items] flesh out tie/untie

Raykyn55:
Added some missing matches, removed returns, modified untie to accept just the item, for general untie, or item + container for specific untie. This is necessary for the proposed changes to get_crafting_items.


MahtraDR:
The returns don't change anything. They just make things explicit. 

What items can you general untie without specifying what they're tied to? That's new to me. 

The extra matches look good.

Raykyn55:
> 

I can put the returns back in if you'd prefer them, but method ends with those actions anyways. 

Actually, everything can be untied without designating the location. I ran into this issue with pliers before I started keeping two sets, since they're used in forging and engineering.

PR #6375, opened 2023-02-06
[script][common-crafting] Get/stow crafting item rewrite

Raykyn55:
Slimmed down, uses items methods, gives us true/false. As an added benefit, allows the use of eddy portals for crafting storage.


Raykyn55:
le sigh

PR #6377, opened 2023-02-06
[script][help-me][validate] - Add dev mode indication

asechrest:
- Changed script name from `help_me` to `help-me` to match script naming convention
- Added indication of whether user is on dev mode to the `help-me` output
- Added warning to `;validate` if user is on dev mode

### DIFF
--- a/help-me.lic
+++ b/help-me.lic
@@ -42,6 +42,8 @@ class HelpMe
     message_body += "\n"
     message_body += "Game Client: #{$frontend}"
     message_body += "\n"
+    message_body += "Dev Mode: #{$manager.developer}"
+    message_body += "\n"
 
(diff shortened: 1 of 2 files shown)


MahtraDR:
Keep in mind that we don't delete files ever, so this won't rename existing scripts, it'll simply add a new help-me script. Do you want to rename the file even so? (I'm happy to merge, note)

asechrest:
> Keep in mind that we don't delete files ever, so this won't rename existing scripts, it'll simply add a new help-me script. Do you want to rename the file even so? (I'm happy to merge, note)

Welllllll, I don't know, then. Hmm.

MahtraDR:
@asechrest Keep or ditch?

PR #6378, opened 2023-02-07
[script][combat-trainer] support all magical training via combat_spell_training

mikejcook:
This is a change that allows all magical training with combat-trainer to happen via combat_spell_training, disregarding offensive_spells completely.

Why make this change?
I found that my magical training rates would be inconsistent. I would also have to tweak the combat_spell_timer to get things to be 'just right' - and they almost never were.

What are the results of the changes?
With these changes, all spells can live in combat_spell_training. offensive_spells can be left completely blank. When setup this wait, combat-trainer will do its best to just rotate through each skill, training the most appropriate selection each time.

What was actually changed?
I exclude TM, Debil, and combat-based Sorcery if there are no NPCs present when selecting the spell to train in check_training, otherwise it'll try to cast a spell without a target. Additionally, I had to add a reset to ct-spelllost in that method, otherwise if the target e.g. died before the spell could be cast, things would get screwy and it would try to prepare another spell, add more mana, and then usually backfire. There's a variable training_spells_combat_sorcery that defaults to false so that non-combat-based Sorcery spells will not check for NPCs.

There should be no backwards compatibility issues. For characters setup using offensive_spells, everything will continue to work the same as it used to. You would need to opt-in by adding the appropriate spells to combat_spell_training. Additionally, training_spells_combat_sorcery prevents any issues with people who are casting non-combat sorcery spells as part of their training routine today.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1595,8 +1595,7 @@ class SpellProcess
     needs_training = ['Warding', 'Utility', 'Augmentation', 'Sorcery', 'Debilitation', 'Targeted Magic']
                      .select { |skill| @training_spells[skill] }
                      .select { |skill| game_state.is_offense_allowed? || @training_spells[skill]['harmless'] }
-                     .select { |skill| ['Warding', 'Utility', 'Augmentation'].include?(skill)
-                                        || !@training_spells_combat_sorcery && skill == 'Sorcery' || !game_state.npcs.empty? }
+                     .select { |skill| ['Warding', 'Utility', 'Augmentation'].include?(skill) || !@training_spells_combat_sorcery && skill == 'Sorcery' || !game_state.npcs.empty? }
                      .select { |skill| DRSkill.getxp(skill) < @magic_exp_training_max_threshold }
                      .select { |skill| Time.now > (@training_spells[skill]['cyclic'] ? @training_cyclic_timer : @training_cast_timer) }
                      .select { |skill| @training_spells[skill]['night'] ? UserVars.sun['night'] : true }
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1595,7 +1595,8 @@ class SpellProcess
     needs_training = ['Warding', 'Utility', 'Augmentation', 'Sorcery', 'Debilitation', 'Targeted Magic']
                      .select { |skill| @training_spells[skill] }
                      .select { |skill| game_state.is_offense_allowed? || @training_spells[skill]['harmless'] }
-                     .select { |skill| ['Warding', 'Utility', 'Augmentation'].include?(skill) || !@training_spells_combat_sorcery && skill == 'Sorcery' || !game_state.npcs.empty? }
+                     .select { |skill| ['Warding', 'Utility', 'Augmentation'].include?(skill)
+                                        || !@training_spells_combat_sorcery && skill == 'Sorcery' || !game_state.npcs.empty? }
                      .select { |skill| DRSkill.getxp(skill) < @magic_exp_training_max_threshold }
                      .select { |skill| Time.now > (@training_spells[skill]['cyclic'] ? @training_cyclic_timer : @training_cast_timer) }
                      .select { |skill| @training_spells[skill]['night'] ? UserVars.sun['night'] : true }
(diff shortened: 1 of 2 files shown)


BinuDR:
This can already be done with `offensive_spells` as follows:

```
offensive_spells:
- skill: Targeted Magic
  name: Fire Ball
  mana: 7
  cast_only_to_train: true
- skill: Utility
  name: Air Bubble
  mana: 8
  harmless: true
  symbiosis: true
  use_auto_mana: false
  cast_only_to_train: true
- skill: Augmentation
  name: Substratum
  mana: 8
  harmless: true
  use_auto_mana: false
  symbiosis: true
  cast_only_to_train: true
- skill: Warding
  name: Gam Irnan
  mana: 8
  harmless: true
  use_auto_mana: false
  symbiosis: true
  cast_only_to_train: true
- skill: Sorcery
  name: Athleticism
  mana: 1
  harmless: true
  use_auto_mana: false
  cast_only_to_train: false
  symbiosis: true

```

mikejcook:
The method that casts offensive spells won't run if there are no NPCs
present. So it has the opposite problem.

Kiriawen:
it actually doesn't.  it _will_ cast spells even if no critters are present.  If you are finding yourself not casting make sure you tag your aug/ward/util and non-offensive sorc spells as harmless.

mikejcook:
> it actually doesn't. it _will_ cast spells even if no critters are present. If you are finding yourself not casting make sure you tag your aug/ward/util and non-offensive sorc spells as harmless.
> 
> I use `offensive_spells:` to train all my magics and have over a number of charcters/guilds.
> 
> Only downside is you have to manage your TM/Debil spells. If you're capping the critters and those aren't gaining mindstates you'll have to comment them out temporarily or move up to a harder critter to keep all yer magics moving.

Are you using a custom version of combat-trainer? Because the third line in check_offensive returns if there aren't any NPCs. Hence why using offensive_spells for all magic training isn't possible, unless you're okay with it not casting when there aren't NPCs present. I am guessing that your scripts just don't cast spells when there aren't NPCs around and you just didn't realize it.

```
  def check_offensive(game_state)
    echo "Checking offensive spells:" if $debug_mode_ct
    return if game_state.casting
```

Hiinky:
Personally, this feels like a perfect use of a self-edit preserved in the custom folder.  Purely because the window where having training and offensive spells together being more effective is very small.  Second, if you find yourself with zero mobs very often, you really need to question your characters life choices and pick a better mob or kill slower as its a net loss overall to just standing there casting magic.  At the end of the day it's a fairly small edit, I just cringe to see ;combat getting even more complicated.

mikejcook:
> Personally, this feels like a perfect use of a self-edit preserved in the custom folder. Purely because the window where having training and offensive spells together being more effective is very small. Second, if you find yourself with zero mobs very often, you really need to question your characters life choices and pick a better mob or kill slower as its a net loss overall to just standing there casting magic. At the end of the day it's a fairly small edit, I just cringe to see ;combat getting even more complicated.

What do you mean when you say the window where having them together being more effective is very small? What's the advantage to splitting them out? I don't really see a reason to have training spells and offensive spells split out. What value does this provide or having them merged into one group?

Regarding the zero mobs issue, it tends to be an issue in earlier hunting grounds. I think they tend to be questionably tuned for situations where there's only one or two people in a zone. Premie Gargs can be great or super slow, Silver-backed bears is the same - it's often great, but there can be periods where the bears just aren't spawning for a few minutes.

combat-trainer.lic
```diff
@@ -1592,10 +1595,13 @@ class SpellProcess
     needs_training = ['Warding', 'Utility', 'Augmentation', 'Sorcery', 'Debilitation', 'Targeted Magic']
                      .select { |skill| @training_spells[skill] }
                      .select { |skill| game_state.is_offense_allowed? || @training_spells[skill]['harmless'] }
+                     .select { |skill| ['Warding', 'Utility', 'Augmentation'].include?(skill)
+                                        || !@training_spells_combat_sorcery && skill == 'Sorcery' || !game_state.npcs.empty? }
```
MahtraDR, line 1599:
```suggestion
                     .select { |skill| ['Warding', 'Utility', 'Augmentation'].include?(skill) || !@training_spells_combat_sorcery && skill == 'Sorcery' || !game_state.npcs.empty? }
```

mikejcook:
Yep! Everything is working great.

PR #6384, opened 2023-02-11
[script][combat-trainer] Removed return if @tk_ammo from CT

jbrock24:
It doesn't seem to be warranted, and is used as a fail-safe for no moons. If no moons, these spells never cast. I don't see the point iN not using slivers as well as tk_ammo, the latter being a failsafe if no moons are up. CT will always stow tk_ammo on cleanup, and it uses slivers first when casting the spell, don't see a problem with it using both.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1306,7 +1306,7 @@ class SpellProcess
   def check_slivers(game_state)
     return unless DRStats.moon_mage?
     return if game_state.casting
-    return if @tk_ammo
+    #return if @tk_ammo
     return unless @tk_spell
     return if DRSpells.slivers
     return if UserVars.moons['visible'].empty?
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1308,7 +1308,7 @@ class SpellProcess
     return unless @tk_spell
     return if game_state.casting
     return if DRSpells.slivers
-    if UserVars.moons['visible'].empty?
+    unless UserVars.moons['visible'].any?
       DRC.message "There are NO moons available and you have NO TK_AMMO set! You need to fix this!" if !@tk_ammo && $debug_mode_ct
       return
     end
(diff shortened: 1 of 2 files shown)


jbrock24:
New functionality:

1. Will warn if the user is not using those spells, this will spam given circumstances, we can keep this in as a HEY DUDE PAY ATTENTION, or we can remove it and let them deal with it. 
2. If tkt/tks with `sliver` set to use, it will cast if the slivers (DRSPELLS check) are up or if tk_ammo is set, else it will not cast those spells.
3. Added warning check for validator. If slivers is selected, it will check if tk_ammo is set, if not, it's going to warn the user that no moons = no spells unless tk_ammo is set.

combat-trainer.lic
```diff
@@ -1798,7 +1801,7 @@ class SpellProcess
                    .select { |spell| spell['cyclic'] ? !DRSpells.active_spells[spell['name']] : true }
                    .select { |spell| spell['night'] ? UserVars.sun['night'] : true }
                    .select { |spell| spell['day'] ? UserVars.sun['day'] : true }
-                   .select { |spell| spell['slivers'] ? DRSpells.slivers : true }
+                   .select { |spell| spell['slivers'] ? DRSpells.slivers || @tk_ammo : true }
```
MahtraDR, line 1804:
```suggestion
                   .select { |spell| spell['slivers'] ? (DRSpells.slivers || @tk_ammo) : true }
```

combat-trainer.lic
```diff
@@ -1305,11 +1305,14 @@ class SpellProcess
 
   def check_slivers(game_state)
     return unless DRStats.moon_mage?
-    return if game_state.casting
-    return if @tk_ammo
     return unless @tk_spell
+    return if game_state.casting
     return if DRSpells.slivers
-    return if UserVars.moons['visible'].empty?
+    if UserVars.moons['visible'].empty?
```
MahtraDR, line 1311:
```suggestion
    unless UserVars.moons['visible'].any?
```

jbrock24:
Thanks for the feedback!

Issue #6014, opened 2022-08-15
enchant.lic - getting an augmenting loop when crafting an augmenting loop


Tirost:
Requesting that enchant.lic attempt to get an augmenting loop from the player's craft bag. Currently it uses the command `get my augmenting loop` which tries to get the unfinished augmenting loop from the brazier when crafting one.

I think the modification can be made to line 198 of enchant.lic at the `get_item(@loop)` section.

Log: https://pastebin.com/AuMyxVjY

MahtraDR:
Resolved by https://github.com/rpherbig/dr-scripts/pull/6390

Please reopen if there are any more issues.

PR #6390, opened 2023-02-14
[scripts][enchant]Removed includes, add prefixes, trimmed methods

MahtraDR:
Have tested across several recipes on mine now. Captures some news text from secondary sigils newly available from societies. 

Still todo, move stamp and crucible cleaning to common-crafting.


MahtraDR:
Resolves https://github.com/rpherbig/dr-scripts/issues/6014

Issue #6392, opened 2023-02-15
[script] validate.lic throws an error when Targeted Magic is used as  a stop on


Kiriawen:
When TM is used as a stop_on: and it is not listed in weapon_training:  it throws an error.  It should not.  

![image](https://user-images.githubusercontent.com/77058868/219174007-61901b38-221a-44a7-b672-71d6e78baea2.png)

PR #6393, opened 2023-02-15
[scripts][validate] Exclude TM from weapon skills error

MahtraDR:
As per title.

Resolves https://github.com/rpherbig/dr-scripts/issues/6392


Kiriawen:
Tested.

image

PR #6397, opened 2023-02-17
[script][spellmonitor] Track STC duration

Raykyn55:
Stellar Collector can last a long time, but we didn't have any way of checking the remaining duration, since the unit of measurement could be anlaen, and the xml was different from normal spells:
```
Finesse  (9 roisaen)

Noumena  (8 roisaen)

Iridius Rod  (intact, fading)

Stellar Collector  (0%, 4 anlaen)
```
I added a regex specifically for this spell, since afaik, this is the only type of spell that operates like this.  Before this PR, checking your stellar collector with active_spells was either nil or 1000. After:
```
Stellar Collector  (100%, 3 anlaen)

>;e echo DRSpells.active_spells['Stellar Collector']
--- Lich: exec1 active.
[exec1: 90]
--- Lich: exec1 has exited.
```
Assuming here my math was correct, that an anlaen is 30 Roisaen. In truth, it's largely irrelevant, because by ~60 it reverts to Roisaen, at which point we just get the duration in Roisaen. Might need to make anlan an option, though like I said, I think it reverts to roisaen before then.

The purpose of this change was to allow folks to drop a collector and keep it up with buff-watcher (say recast: 20) which can be done anywhere.

### DIFF
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -190,6 +190,12 @@ active_spells_hook = proc do |server_string|
       # Cyclic spell or Osrel Meraud cyclic spell
       spell = Regexp.last_match[:spell]
       duration = UNKNOWN_DURATION
+    when /Stellar Collector\s+\(\d+%,\s*(?<duration>\d+)\s*(?<unit>(?:roisae?n|anlaen))/
+      # Stellar Collector
+      spell = "Stellar Collector"
+      duration = Regexp.last_match[:duration].to_i
+      unit = Regexp.last_match[:unit]
+      duration = unit == 'anlaen' ? duration*30 : duration
     when /(?<spell>[^<>]+?)\s+\(.+\)/i
       # Spells with inexact duration verbiage, such as with
       # Barbarians without knowledge of Power Monger mastery
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -199,22 +199,15 @@ active_spells_hook = proc do |server_string|
       # Cyclic spell or Osrel Meraud cyclic spell
       spell = Regexp.last_match[:spell]
       duration = UNKNOWN_DURATION
-    when /(?<spell>Stellar Collector)\s+\((?<percentage>\d+)%,\s*(?<duration>\d+)\s*(?<unit>(?:roisae?n|anlaen))/
+    when /(?<spell>Stellar Collector)\s+\((?<percentage>\d+)%,\s*(?<duration>\d+)?\s*(?<unit>(?:roisae?n|anlaen|fading))/
       # Stellar collector special case
       # XML looks like:
       # Stellar Collector  (0%, 4 anlaen)
       spell = Regexp.last_match[:spell]
-      duration = Regexp.last_match[:duration].to_i
+      duration = duration.nil? ? 0 : Regexp.last_match[:duration].to_i
       DRSpells.stellar_percentage = Regexp.last_match[:percentage].to_i
       unit = Regexp.last_match[:unit]
       duration = unit == 'anlaen' ? duration*30 : duration
-    when /(?<spell>[^<>]+?)\s+\(fading\)/i, /(?<spell>[^<>]+?)\s+\(\d+%, fading\)/
-      # Spell fading away
-      # special case for Stellar Collector
-      # XML looks like:
-      # Stellar Collector  (0%, fading)
-      spell = Regexp.last_match[:spell]
-      duration = 0
     when /(?<spell>[^<>]+?)\s+\(.+\)/i
       # Spells with inexact duration verbiage, such as with
       # Barbarians without knowledge of Power Monger mastery
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -203,7 +203,7 @@ active_spells_hook = proc do |server_string|
       # Stellar collector special case
       # XML looks like:
       # Stellar Collector  (0%, 4 anlaen)
-      spell = "Stellar Collector"
+      spell = Regexp.last_match[:spell]
       duration = Regexp.last_match[:duration].to_i
       DRSpells.stellar_percentage = Regexp.last_match[:percentage].to_i
       unit = Regexp.last_match[:unit]
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -17,6 +17,7 @@ class DRSpells
   @@known_spells = {}
   @@known_feats = {}
   @@refresh_data = {}
+  @@stellar_percentage = 0
   @@slivers = false
   @@spellbook_format = nil # 'column-formatted' or 'non-column'
 
@@ -49,6 +50,14 @@ class DRSpells
 
   def self.slivers=(val)
     @@slivers = val
+  end
+  
+  def self.stellar_percentage
+    @@stellar_percentage
+  end
+  
+  def self.stellar_percentage=(val)
+    @@stellar_percentage = val
   end
 
   def self.grabbing_active_spells
@@ -190,12 +199,22 @@ active_spells_hook = proc do |server_string|
       # Cyclic spell or Osrel Meraud cyclic spell
       spell = Regexp.last_match[:spell]
       duration = UNKNOWN_DURATION
-    when /Stellar Collector\s+\(\d+%,\s*(?<duration>\d+)\s*(?<unit>(?:roisae?n|anlaen))/
-      # Stellar Collector
+    when /(?<spell>Stellar Collector)\s+\((?<percentage>\d+)%,\s*(?<duration>\d+)\s*(?<unit>(?:roisae?n|anlaen))/
+      # Stellar collector special case
+      # XML looks like:
+      # Stellar Collector  (0%, 4 anlaen)
       spell = "Stellar Collector"
       duration = Regexp.last_match[:duration].to_i
+      DRSpells.stellar_percentage = Regexp.last_match[:percentage].to_i
       unit = Regexp.last_match[:unit]
       duration = unit == 'anlaen' ? duration*30 : duration
+    when /(?<spell>[^<>]+?)\s+\(fading\)/i, /(?<spell>[^<>]+?)\s+\(\d+%, fading\)/
+      # Spell fading away
+      # special case for Stellar Collector
+      # XML looks like:
+      # Stellar Collector  (0%, fading)
+      spell = Regexp.last_match[:spell]
+      duration = 0
     when /(?<spell>[^<>]+?)\s+\(.+\)/i
       # Spells with inexact duration verbiage, such as with
       # Barbarians without knowledge of Power Monger mastery
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -204,7 +204,7 @@ active_spells_hook = proc do |server_string|
       # XML looks like:
       # Stellar Collector  (0%, 4 anlaen)
       spell = Regexp.last_match[:spell]
-      duration = duration.nil? ? 0 : Regexp.last_match[:duration].to_i
+      duration = Regexp.last_match[:duration].to_i
       DRSpells.stellar_percentage = Regexp.last_match[:percentage].to_i
       unit = Regexp.last_match[:unit]
       duration = unit == 'anlaen' ? duration*30 : duration
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -203,6 +203,7 @@ active_spells_hook = proc do |server_string|
       # Stellar collector special case
       # XML looks like:
       # Stellar Collector  (0%, 4 anlaen)
+      # Stellar Collector  (0%, fading)
       spell = Regexp.last_match[:spell]
       duration = Regexp.last_match[:duration].to_i
       DRSpells.stellar_percentage = Regexp.last_match[:percentage].to_i
--- a/spellmonitor.lic
+++ b/spellmonitor.lic
@@ -2,7 +2,7 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#spellmonitor
 =end
 
-$SPELLMONITOR_VERSION = '2.0.0'
+$SPELLMONITOR_VERSION = '2.0.1'
 
 custom_require.call(%w[drinfomon])
 


MahtraDR:
Paste the xml stream entry for this? It might be like OM from clerics.

Raykyn55:
> Paste the xml stream entry for this? It might be like OM from clerics.

```
<pushStream id="percWindow"/>Trabe Chalice  (intact, fading)

Finesse  (9 roisaen)

Noumena  (8 roisaen)

Iridius Rod  (intact, fading)

Stellar Collector  (0%, 4 anlaen)

<popStream/><prompt time="1676593107">&gt;</prompt>
```

MahtraDR:
> > Paste the xml stream entry for this? It might be like OM from clerics.
> 
> ```
> <pushStream id="percWindow"/>Trabe Chalice  (intact, fading)
> 
> Finesse  (9 roisaen)
> 
> Noumena  (8 roisaen)
> 
> Iridius Rod  (intact, fading)
> 
> Stellar Collector  (0%, 4 anlaen)
> 
> <popStream/><prompt time="1676593107">&gt;</prompt>
> ```

Does it ever have leading zeroes? Is it 1% or 01% ? I assume the former.

MahtraDR:
This is the OM stream, not exactly the same, but same idea in the end:
```xml
"<clearStream id=\"percWindow\"/>"
"Osrel Meraud  (90%)"
```

For OM we just track percentage. I suggest we do the same for STC.

MahtraDR:
We don't need all this. We can track STC duration with just a regex change to https://github.com/rpherbig/dr-scripts/blob/eeed2a2a2947f6b1860d7c1cbfd1a039029dabf5/spellmonitor.lic#L181

https://rubular.com/r/qSXHN1MZs2epCJ

Raykyn55:
> We don't need all this. We can track STC duration with just a regex change to
> 
> https://github.com/rpherbig/dr-scripts/blob/eeed2a2a2947f6b1860d7c1cbfd1a039029dabf5/spellmonitor.lic#L181
> 
> https://rubular.com/r/00jzmvsAqlbIhn

Did you look at the duration match on rubular?

MahtraDR:
> > We don't need all this. We can track STC duration with just a regex change to
> > https://github.com/rpherbig/dr-scripts/blob/eeed2a2a2947f6b1860d7c1cbfd1a039029dabf5/spellmonitor.lic#L181
> > 
> > https://rubular.com/r/00jzmvsAqlbIhn
> 
> Did you look at the duration match on rubular?

image

I wouldn't trust the anlas conversion to be exact, because it's depicting a range. The percentage you can count on. It's how we deal with OM too, so we know it works, and how.

Raykyn55:
> > > We don't need all this. We can track STC duration with just a regex change to
> > > https://github.com/rpherbig/dr-scripts/blob/eeed2a2a2947f6b1860d7c1cbfd1a039029dabf5/spellmonitor.lic#L181
> > > 
> > > https://rubular.com/r/00jzmvsAqlbIhn
> > 
> > 
> > Did you look at the duration match on rubular?
> 
> image
> 
> I wouldn't trust the anlas conversion to be exact, because it's depicting a range. The percentage you can count on. It's how we deal with OM too, so we know it works, and how.

Problem is we need the duration to know when to refresh the spell. 100% stc can have a duration under a rois, and die. The % for stc is in no way tied to whether it stays up.

MahtraDR:
> Problem is we need the duration to know when to refresh the spell. 100% stc can have a duration under a rois, and die. The % for stc is in no way tied to whether it stays up.

OHHHH Ok. Sorry. I don't know the spell. So what's the percentage indicative of?

Raykyn55:
> > Problem is we need the duration to know when to refresh the spell. 100% stc can have a duration under a rois, and die. The % for stc is in no way tied to whether it stays up.
> 
> OHHHH Ok. Sorry. I don't know the spell. So what's the percentage indicative of?

Just how much aura is stored. I'm not sure that's all that useful, since aura is a little bit black boxy. Could capture that as well, but I was more focused on just keeping the spell up.

MahtraDR:
> Just how much aura is stored. I'm not sure that's all that useful, since aura is a little bit black boxy. Could capture that as well, but I was more focused on just keeping the spell up.

There is at least one stc script out there, Urbaj has one I think. Easy enough to capture the percentage anyway.

Are there any other spells that use this format? Regalia maybe? A small regex change could make this generic to any spell using this format, without hardcoding the spell as STC.

https://rubular.com/r/00jzmvsAqlbIhn Keeps the format of the other entries, allows for more spells with this same weird syntax.

Raykyn55:
> Easy enough to capture the percentage anyway.
>
> Are there any other spells that use this format? Regalia maybe? A small regex change could make this generic to any spell using this format, without hardcoding the spell as STC.

I don't use OM, but my understanding is the % is used in lieu of the duration. It doesn't capture additional information.

I can open it up and catch the %, but that would mean changing the output from a string to. . .something else. 

>There is at least one stc script out there, Urbaj has one I think. 

I'd be interested to see that script, could use that in place of what I worked into the regalia draft.

MahtraDR:
I think we can fold this all into the main `when` case and not duplicate lines.

https://rubular.com/r/n9TcRPo6sWvNR2

Urbaj's is here https://github.com/urbaj-dr/dr-scripts/blob/master/stc.lic

MahtraDR:
> > > Problem is we need the duration to know when to refresh the spell. 100% stc can have a duration under a rois, and die. The % for stc is in no way tied to whether it stays up.
> > 
> > 
> > OHHHH Ok. Sorry. I don't know the spell. So what's the percentage indicative of?
> 
> Just how much aura is stored. I'm not sure that's all that useful, since aura is a little bit black boxy. Could capture that as well, but I was more focused on just keeping the spell up.

So, if we capture the % we can do it in a new field, `<capacity>` ? or whatever. Then scripts can make use of that. `if capacity == 100` or whatever the usage of this spell is. It'd be more flexible anyway.

Raykyn55:
> I think we can fold this all into the main `when` case and not duplicate lines.
> 
> https://rubular.com/r/n9TcRPo6sWvNR2
> 
Was just thinking, we could jam that % into a uservar, stellar_collector_perc or something. As far as duping lines, you're call, but this strikes me as a special case. Have a think and lmk, sure I can sort out how to do it inside the other when.

Raykyn55:
> So, if we capture the % we can do it in a new field, `<capacity>` ? or whatever. Then scripts can make use of that. `if capacity == 100` or whatever the usage of this spell is. It'd be more flexible anyway.

We can, but like I said, we'd need to work out some way to store it. As it is, the method ends:
```ruby
DRSpells.active_spells[spell] = duration
```
Which leaves room for only the one field. <spell> sets the variable, duration is the value. All the scripts that use this check over/under against an integer, or truthy.

MahtraDR:
> > So, if we capture the % we can do it in a new field, `<capacity>` ? or whatever. Then scripts can make use of that. `if capacity == 100` or whatever the usage of this spell is. It'd be more flexible anyway.
> 
> We can, but like I said, we'd need to work out some way to store it. As it is, the method ends:
> 
> ```ruby
> DRSpells.active_spells[spell] = duration
> ```
> 
> Which leaves room for only the one field. sets the variable, duration is the value. All the scripts that use this check over/under against an integer, or truthy.

I'm just reading that script in full now, and I agree it's a bit bigger change to do this here. A dedicated script could do it though, and maybe we leave this scope creep out. :)

MahtraDR:
What does the xml look like at the  low end of duration? We have an entry for when spells are fading.

Raykyn55:
> I'm just reading that script in full now, and I agree it's a bit bigger change to do this here. A dedicated script could do it though, and maybe we leave this scope creep out. :)

I think your idea is a good one. If we leave this in its own regex, we can tease out duration like any other spell, and add an attribute for stc %. If we were to make another script, it would have to access the xml which is just duplicating the work spellmonitor.lic already does, or matching against perceive which is clunky and slow. 

This seems like precisely the sort of information this script was designed to collect and make available. 

>What does the xml look like at the low end of duration? We have an entry for when spells are fading.

The same as other spells, but still with the leading %value. Drops to roisan when it's under 60 or 90.

MahtraDR:
> I think your idea is a good one. If we leave this in its own regex, we can tease out duration like any other spell, and add an attribute for stc %. If we were to make another script, it would have to access the xml which is just duplicating the work spellmonitor.lic already does, or matching against perceive which is clunky and slow.
> 
> This seems like precisely the sort of information this script was designed to collect and make available.

I agree, but I don't have the brain cycles to expand this right now. :) 

Let's see if this is a thing people would even use, before we embark on a larger project here. This would be a very large rewrite, and every script that accesses the data would have to change too.

> > What does the xml look like at the low end of duration? We have an entry for when spells are fading.
> 
> The same as other spells, but still with the leading %value. Drops to roisan when it's under 60 or 90.

We need this xml entry to adjust the `# Spell fading away` entry.

Raykyn55:
> > I think your idea is a good one. If we leave this in its own regex, we can tease out duration like any other spell, and add an attribute for stc %. If we were to make another script, it would have to access the xml which is just duplicating the work spellmonitor.lic already does, or matching against perceive which is clunky and slow.
> > This seems like precisely the sort of information this script was designed to collect and make available.
> 
> I agree, but I don't have the brain cycles to expand this right now. :)
> 
> Let's see if this is a thing people would even use, before we embark on a larger project here. This would be a very large rewrite, and every script that accesses the data would have to change too.

Let me submit something tomorrow when I'm fresh, I don't think we're thinking along the same lines on this. 

> We need this xml entry to adjust the `# Spell fading away` entry.

Heh, min cast last 90 minutes, and you can't release the spell. Let's just say I'll try :)

MahtraDR:
This works
```ruby
    when /(?<spell>Stellar Collector)\s+\(\d+%,\s*(?<duration>\d+)\s*(?<unit>(?:roisae?n|anlaen))\)/
      # Stellar collector special case
      # XML looks like:
      # Stellar Collector  (0%, 4 anlaen)
      spell = Regexp.last_match[:spell]
      duration = Regexp.last_match[:duration].to_i
      unit = Regexp.last_match[:unit]
      duration = unit == 'anlaen' ? duration*30 : duration
    when /(?<spell>[^<>]+?)\s+\(fading\)/i, /(?<spell>[^<>]+?)\s+\(\d+%, fading\)/
      # Spell fading away
      # special case for Stellar Collector
      # XML looks like:
      # Stellar Collector  (0%, fading)
      spell = Regexp.last_match[:spell]
      duration = 0
```

Raykyn55:
Stellar Collector  (0%, 88 roisaen)

```
>;e echo DRSpells.active_spells['Stellar Collector']
--- Lich: exec1 active.
[exec1: 88]
--- Lich: exec1 has exited.
>;e echo DRSpells.stellar_percentage
--- Lich: exec1 active.
[exec1: 0]
```

MahtraDR:
> Stellar Collector (0%, 88 roisaen)
> 
> ```
> >;e echo DRSpells.active_spells['Stellar Collector']
> --- Lich: exec1 active.
> [exec1: 88]
> --- Lich: exec1 has exited.
> >;e echo DRSpells.stellar_percentage
> --- Lich: exec1 active.
> [exec1: 0]
> ```

Hah! Awesome out of the box thinking!

spellmonitor.lic
```diff
@@ -190,6 +199,22 @@ active_spells_hook = proc do |server_string|
       # Cyclic spell or Osrel Meraud cyclic spell
       spell = Regexp.last_match[:spell]
       duration = UNKNOWN_DURATION
+    when /(?<spell>Stellar Collector)\s+\((?<percentage>\d+)%,\s*(?<duration>\d+)\s*(?<unit>(?:roisae?n|anlaen))/
+      # Stellar collector special case
+      # XML looks like:
+      # Stellar Collector  (0%, 4 anlaen)
+      spell = "Stellar Collector"
```
MahtraDR, line 206:
```suggestion
      spell = Regexp.last_match[:spell]
```

MahtraDR:
We don't need a whole second `when` for the fading. Just expand the existing one.

MahtraDR:
https://github.com/rpherbig/dr-scripts/blob/eeed2a2a2947f6b1860d7c1cbfd1a039029dabf5/spellmonitor.lic#L185 Needs to be changed to the regex list above, so it can correctly capture the fading line, and set the duration to zero like it does for other spells. See https://github.com/rpherbig/dr-scripts/pull/6397#issuecomment-1434346121

Raykyn55:
> https://github.com/rpherbig/dr-scripts/blob/eeed2a2a2947f6b1860d7c1cbfd1a039029dabf5/spellmonitor.lic#L185
> 
> Needs to be changed to the regex list above, so it can correctly capture the fading line, and set the duration to zero like it does for other spells. See [#6397 (comment)](https://github.com/rpherbig/dr-scripts/pull/6397#issuecomment-1434346121)

Ok, I'm confused. Do you want this on it's own line or part of the other? If we get fading, we don't get a duration (nil). nil converted to an integer is 0, right?

Raykyn55:
```
>;e echo DRC.bput("perc", /You sense(?<pickle>\d+)? the/);echo Regexp.last_match(:pickle).to_i
--- Lich: exec1 active.
[exec1]>perc
You reach out with your senses and see luminous (14/21) streams of deep green and blue Life mana flowing through the area.
Letting your senses extend further, you feel there is shimmering (9/21) mana to the east.
You are being protected by a weak Essence of Yew, which should last for less than a roisan.
You sense the Embed the Cycle spell upon you, which will last for about two roisaen.
You sense the Senses of the Tiger spell upon you, which will last less than a roisan.
You remember a way to combine the weaves of the Avoid symbiosis with another spell.
Roundtime: 3 sec.
>
[exec1: You sense the]
[exec1: 0]
```

spellmonitor.lic
```diff
@@ -190,6 +199,15 @@ active_spells_hook = proc do |server_string|
       # Cyclic spell or Osrel Meraud cyclic spell
       spell = Regexp.last_match[:spell]
       duration = UNKNOWN_DURATION
+    when /(?<spell>Stellar Collector)\s+\((?<percentage>\d+)%,\s*(?<duration>\d+)?\s*(?<unit>(?:roisae?n|anlaen|fading))/
+      # Stellar collector special case
+      # XML looks like:
+      # Stellar Collector  (0%, 4 anlaen)
```
MahtraDR, line 205:
```suggestion
      # Stellar Collector  (0%, 4 anlaen)
      # Stellar Collector  (0%, fading)
```

MahtraDR:
You're right. The current version looks good.

Could you please tick the version on it from `2.0.0` to `2.0.1` and we can merge.

PR #6408, opened 2023-02-19
[scripts][common]Add match for bput when incapacitated.

MahtraDR:
Adds messaging for when incapacitated as per the Ice Patch spell.

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -173,7 +173,7 @@ module DRC
         put message
         timer = Time.now
         next
-      when /^You are still stunned/, /^You can't do that while entangled in a web/
+      when /^You are still stunned/, /^You can't do that while entangled in a web/, /You don't seem to be able to move to do that/
         pause 0.5
         put message
         timer = Time.now


Raykyn55:
hey, so, I put in a fix on this forever ago that G wasn't a fan of (perhaps you as well) that involved moving. One part of that fix was probably ill advised, but this particular change will result in a spammy loop. An alternative option that doesn't involve the moving or the spam is to include the recovery message:
```ruby
      when /^You don't seem to be able to move to do that/
        next unless matches.include?(response)
      when /^After failing to draw a breath for what feels like forever/
        fix_standing
        put message
        timer = Time.now
        next
```

So what happens with this is we get the failure, and it returns to the loop to wait for another match for our time-out interval. once we receive the recovery message, stands you up, sends the original message, resets the time-out, and checks for responses. Recovery usually takes between 5 and 10 seconds. On this, instead of getting upwards of 20 attempts to perform the command, we get one until we see the recovery message, or the timer runs out.

MahtraDR:
The benefit is no spamming, but the drawback is waiting upwards of 15 seconds. There isn't a recovery message for IP that I saw, the effect just ends, and this isn't a condition we can test for. So suppose there was a message for IP, we'd additionally have to find messages for all other spells/things that cause immobilization and capture those. The spam is a bit uncouth, but it is generic enough to work for any spell/situation too. Tradeoffs.

MahtraDR:
I forget you're not on discord, but... we learned a few days ago that @asechrest passed away earlier this month. Sad times.

Raykyn55:
> I forget you're not on discord, but... we learned a few days ago that @asechrest passed away earlier this month. Sad times.

what? jesus, what happened?

MahtraDR:
> hey, so, I put in a fix on this forever ago that G wasn't a fan of (perhaps you as well) that involved moving. One part of that fix was probably ill advised, but this particular change will result in a spammy loop. An alternative option that doesn't involve the moving or the spam is to include the recovery message:
> 
> ```ruby
>       when /^You don't seem to be able to move to do that/
>         next unless matches.include?(response)
>       when /^After failing to draw a breath for what feels like forever/
>         fix_standing
>         put message
>         timer = Time.now
>         next
> ```
> 
> So what happens with this is we get the failure, and it returns to the loop to wait for another match for our time-out interval. once we receive the recovery message, stands you up, sends the original message, resets the time-out, and checks for responses. Recovery usually takes between 5 and 10 seconds. On this, instead of getting upwards of 20 attempts to perform the command, we get one until we see the recovery message, or the timer runs out.

This does seem to work well. Can't think of any drawbacks either. I just added the breath match to the `when /^You must be standing/...` line. I'll PR that shortly.

Issue #3395, opened 2019-01-25
safe-room: handle stunned when using pc empath


jandersson:
```
 * Morrgann joins the adventure.                                                
!SP>                                                                            
[safe-room: ***STATUS*** Bleeding: true, poison: false]                         
[safe-room]>find Empath                                                        
Brave Adventurer in the Realms:                                                 
  Empath.                                                                       
[safe-room]>whisper Empath heal                                                 
You are still stunned.                                                          
[safe-room]>listen to Empath                                                    
You are still stunned.                                                          
!SP>                                                                            
```
The 'You are still stunned' messaging isn't being handled properly. It should what until not stunned then retry.

MahtraDR:
This came up in conversation recently and the problem with resolving this is that the stunned check is often used in combat so we can't make it slower. Closing as won't fix. Nobody is stunned for that long, and they can't do anything whilst stunned anyway. A little spam won't hurt anyone. 😀

rpherbig:
I don't think the idea here was to change stun checking in any common way. My understanding is that the ask was to change `safe-room` to detect stun and wait, e.g. here: https://github.com/rpherbig/dr-scripts/blob/master/safe-room.lic#L270

Even if my understanding is correct, I have no opinion on whether or not this is worth changing.

BinuDR:
We were testing stuns last night.  Lich.rbw has a check for stuns, `stunned?`, and it works most of the time.  Some stuns don't register though, which is a bug on Simu's end.  We can use this relatively reliably to deal with stuns, and request that if anyone finds any other stuns that it doesn't work that they bug it with Simu.

MahtraDR:
As Binu says, we're actually looking into this in a general way. Reopening for now.

PR #6410, opened 2023-02-21
[scripts][common]Improved handling for being immobilized, or stunned, or webbed.

MahtraDR:
As per title, and suggested by Raykyn55 https://github.com/rpherbig/dr-scripts/pull/6408#issuecomment-1438690967

Would it make sense to do something like this for the handling being stunned too? Something like

```ruby
      when /^You are still stunned/
      next if stunned?
```

### DIFF
--- a/common.lic
+++ b/common.lic
@@ -173,12 +173,14 @@ module DRC
         put message
         timer = Time.now
         next
-      when /^You are still stunned/, /^You can't do that while entangled in a web/, /^You don't seem to be able to move to do that/
+      when /^You don't seem to be able to move to do that/
+        next unless matches.include?(response)
+      when /^You are still stunned/, /^You can't do that while entangled in a web/
         pause 0.5
         put message
         timer = Time.now
         next
-      when /^You must be standing/, /^You should stand up first/, /^You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /^You should be sitting up/, /^You really should be standing to play/
+      when /^You must be standing/, /^You should stand up first/, /^You'll need to stand up first/, /^You can't do that while (sitting|kneeling|lying)/, /^You should be sitting up/, /^You really should be standing to play/, /^After failing to draw a breath for what feels like forever/
         fix_standing
         waitrt?
         put message
--- a/common.lic
+++ b/common.lic
@@ -176,7 +176,10 @@ module DRC
       when /^You don't seem to be able to move to do that/
         next unless matches.include?(response)
       when /^You are still stunned/
-        next if stunned?
+        pause 0.5 while stunned?
+        put message
+        timer = Time.now
+        next
       when /^You can't do that while entangled in a web/
         pause 0.5
         put message
--- a/common.lic
+++ b/common.lic
@@ -175,7 +175,9 @@ module DRC
         next
       when /^You don't seem to be able to move to do that/
         next unless matches.include?(response)
-      when /^You are still stunned/, /^You can't do that while entangled in a web/
+      when /^You are still stunned/
+        next if stunned?
+      when /^You can't do that while entangled in a web/
         pause 0.5
         put message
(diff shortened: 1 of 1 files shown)


MahtraDR:
@Raykyn55 Whatcha think?

MahtraDR:
Ah, we'd need a recovery condition. Thinking...

Raykyn55:
This is what I'm running as of this morning:
```ruby
      when /^You can't do that while entangled in a web/
        while webbed? do pause 0.1 end
        put message
        timer = Time.now
        next
      when /^You are still stunned/, 
        while stunned? do pause 0.1 end
        put message
        timer = Time.now
        next
```
For some reason pause 0.1 while webbed? threw an error, which is a new one. This seems to work though.

Raykyn55:
for a more permanent solution, at least specifically for IP, there does appear to be xml indicators for IP:
```
<prompt time="1677007261">IP&gt;</prompt>
```
which when it falls off becomes: 
```
<prompt time="1677007262">P&gt;</prompt>
```
combat indicators by style, but I can't figure out what to do with them wrt capturing. I can't find XMLData to add it, nor would I know how to capture that if I did.

MahtraDR:
> for a more permanent solution, at least specifically for IP, there does appear to be xml indicators for IP:
> 
> ```
> <prompt time="1677007261">IP&gt;</prompt>
> ```
> 
> which when it falls off becomes:
> 
> ```
> <prompt time="1677007262">P&gt;</prompt>
> ```
> 
> combat indicators by style, but I can't figure out what to do with them wrt capturing. I can't find XMLData to add it, nor would I know how to capture that if I did.

That's not an actual xml indicator. What that is updating is the prompt, which not everyone has set. `set StatusPrompt` by contrast, the stunned xml indicator is extremely verbose, and specific (and always happens*)

MahtraDR:
> This is what I'm running as of this morning:
> 
> ```ruby
>       when /^You can't do that while entangled in a web/
>         while webbed? do pause 0.1 end
>         put message
>         timer = Time.now
>         next
>       when /^You are still stunned/, 
>         while stunned? do pause 0.1 end
>         put message
>         timer = Time.now
>         next
> ```
> 
> For some reason pause 0.1 while webbed? threw an error, which is a new one. This seems to work though.

`webbed?` is a status in lich, but from GS. We don't get that in our XML stream in DR. I'd leave the old behaviour for webbing.

Raykyn55:
> 
```
<indicator id='IconWEBBED' visible='y'/>Reonin gestures at you.

<pushStream id="combat" />A thousand tiny spiders crawl out of their hiding places and begin crawling over you, trapping you in a cocoon of sticky webbing!
```

It's the little green symbol that shows up in wrayth. The lich method for capturing both is via indicator(XMLData.indicator shows the full list)

MahtraDR:
> > 
> 
> ```
> <indicator id='IconWEBBED' visible='y'/>Reonin gestures at you.
> 
> <pushStream id="combat" />A thousand tiny spiders crawl out of their hiding places and begin crawling over you, trapping you in a cocoon of sticky webbing!
> ```
> 
> It's the little green symbol that shows up in wrayth. The lich method for capturing both is via indicator(XMLData.indicator shows the full list)

!!! That's new!!! That's amazing! 

One thing to note about these, they're only mostly reliable :) The stun from getting your hand chopped off burgling wasn't registering in the XML for a long time, for example.

MahtraDR:
I've kept the `pause 0.5` line in case we encounter a situation where the stun/webbing isn't registering in the XML, but the game still says we're webbed/stunned. At worse we lose half a second. This should take care of the spam though.

MahtraDR:
This version is running really well now.

MahtraDR:
Resolves https://github.com/rpherbig/dr-scripts/issues/3395 for realsies.

PR #6411, opened 2023-02-22
[scripts][almanac] Rubocop cleanup

MahtraDR:
`rubocop -A` plus manual cleanup of triple equals operator.

### DIFF
--- a/almanac.lic
+++ b/almanac.lic
@@ -1,11 +1,10 @@
-=begin
-  Documentation: https://elanthipedia.play.net/Lich_script_repository#almanac
-=end
+# frozen_string_literal: true
+
+#   Documentation: https://elanthipedia.play.net/Lich_script_repository#almanac
 
 custom_require.call(%w[common common-items])
 
 class Almanac
-
   def initialize
     settings = get_settings
     UserVars.almanac_last_use ||= Time.now - 600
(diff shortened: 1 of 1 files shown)


MahtraDR:
@rpherbig Do we want the frozen literal? Or should I remove that? Would you mind reviewing our https://github.com/rpherbig/dr-scripts/blob/master/.rubocop.yml ? Thanks!

rpherbig:
Sorry, I just now got around to this. In general, the rules make sense.

I see a few disabled rules that I would have expected to be enabled. What do these catch if they are enabled?
* `OptionalBooleanParameter`
* `TrailingWhitespace`
* `AssignmentInCondition`
* `UnifiedInteger`

MahtraDR:
> Sorry, I just now got around to this. In general, the rules make sense.
> 
> I see a few disabled rules that I would have expected to be enabled. What do these catch if they are enabled?
> 
> * `OptionalBooleanParameter`

This one would just trip too often. https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/OptionalBooleanParameter

> * `TrailingWhitespace`

Inherited from Elanthia-Online. I'll disable it for ours now. 

> * `AssignmentInCondition`

Torn on this one. We have scripts that do this. How about we enable it, and allow AllowSafeAssignment?

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/AssignmentInCondition

> * `UnifiedInteger`

We certainly want this one. Thank you. As above, from Elanthia-Online. I'll get them to sync with us too.

PR #6409, opened 2023-02-21
[scripts][common-items]Add gelapod to trash bins

Raykyn55:

PR #6412, opened 2023-02-22
[scripts][drinfomon]Add exception for gelapod at the pond

MahtraDR:
Adding exception for the gelapod at the pond so it's just a regular room object, and not an NPC.


MahtraDR:
Adds an exception for the gelapod to be an object, rather than an NPC at the pond in Crossing.

Relates to https://github.com/rpherbig/dr-scripts/pull/6409

PR #6461, opened 2023-02-26
[scripts][drinfomon] Rubocop

MahtraDR:
As per title.


MahtraDR:
@rpherbig Sorry to bother. This one I would rather not merge without a review. :) It's mostly just `rubocop -A`  but a few of the lines are manual edits, that rubocop couldn't auto-fix.

rpherbig:
> In particular, I don't know how private works, so that is one area I'd love a review for.

My understanding is that `private` works as expected for _instance_ methods (`def foo`) (called with `ClassName.new.foo`), but not for _class_ methods (`def self.foo`) (called with `ClassName.foo`). For class methods we have to use `private_class_method` instead of `private`.

Converting all of these class methods to instance methods seems like the wrong approach, so I'm fine with using `private_class_method` here.

drinfomon.lic
```diff
@@ -827,7 +826,7 @@ exp_hook = proc do |server_string|
   # https://github.com/traject/traject/issues/139
   match_briefexp_on         = regex_flag_briefexp_on.match(server_string)
   match_briefexp_off        = regex_flag_briefexp_off.match(server_string)
-  match_exp_columns         = regex_exp_columns.match(server_string)
+  # match_exp_columns         = regex_exp_columns.match(server_string)
```
rpherbig, line 829: Why was this commented out?

drinfomon.lic
```diff
@@ -1014,7 +1013,7 @@ info_hook = proc do |server_string|
   match_info_gender_age_circle = regex_info_gender_age_circle.match(server_string)
   match_info_stat_value        = regex_info_stat_value.match(server_string)
   match_info_encumbrance       = regex_info_encumbrance.match(server_string)
-  match_info_balance           = regex_info_balance.match(server_string)
+  # match_info_balance           = regex_info_balance.match(server_string)
```
rpherbig, line 1016: Why was this commented out?

drinfomon.lic
```diff
@@ -1439,7 +1438,7 @@ while line = script.gets
         .each_key do |town|
           copper = convert2copper(Regexp.last_match(1).to_i, Regexp.last_match(2))
           CharSettings['bank_accounts'][town] = CharSettings['bank_accounts'][town].to_i + copper
-          break
+          # break
```
rpherbig, line 1441: Why was this commented out?

drinfomon.lic
```diff
@@ -1451,7 +1450,7 @@ while line = script.gets
         .each_key do |town|
           copper = convert2copper(Regexp.last_match(1).to_i, Regexp.last_match(2))
           CharSettings['bank_accounts'][town] = CharSettings['bank_accounts'][town].to_i - copper
-          break
+          # break
```
rpherbig, line 1453: Why was this commented out?

drinfomon.lic
MahtraDR: Rubocop reports:
```
drinfomon.lic:830:3: W: Lint/UselessAssignment: Useless assignment to variable - match_exp_columns. Did you mean match_exp_clear_mindstate?
  match_exp_columns         = regex_exp_columns.match(server_string)
  ^^^^^^^^^^^^^^^^^
```

drinfomon.lic
MahtraDR: Rubocop reports
```
drinfomon.lic:830:3: W: Lint/UselessAssignment: Useless assignment to variable - match_exp_columns. Did you mean match_exp_clear_mindstate?
  match_exp_columns         = regex_exp_columns.match(server_string)
  ^^^^^^^^^^^^^^^^^
```

drinfomon.lic
MahtraDR: Rubocop suggested this loop would only ever run once in any case
```
drinfomon.lic:1437:7: W: Lint/UnreachableLoop: This loop will have at most one iteration.
      bank_titles ...
      ^^^^^^^^^^^
drinfomon.lic:1449:7: W: Lint/UnreachableLoop: This loop will have at most one iteration.
      bank_titles ...
      ^^^^^^^^^^^
```

drinfomon.lic
MahtraDR: Rubocop suggested this loop would only ever run once in any case
```
drinfomon.lic:1437:7: W: Lint/UnreachableLoop: This loop will have at most one iteration.
      bank_titles ...
      ^^^^^^^^^^^
drinfomon.lic:1449:7: W: Lint/UnreachableLoop: This loop will have at most one iteration.
      bank_titles ...
      ^^^^^^^^^^^
```

drinfomon.lic
rpherbig: Let's just delete it, then

drinfomon.lic
rpherbig: Let's just delete it, then

drinfomon.lic
rpherbig: We really should only ever have one item match the `select` a few lines up. The break just makes that explicit.

I suggest changing `select` to `find` (and deleting the `break`), which will also have a meaningless performance improvement since Ruby doesn't need to traverse the entire list.

drinfomon.lic
rpherbig: Same suggestion as the other select/break loop.

drinfomon.lic
```diff
@@ -873,7 +873,7 @@ exp_hook = proc do |server_string|
   when regex_exp_columns
     if parsing_exp_output
       server_string.scan(regex_exp_columns) do |skill_value, rank_value, percent_value, rate_value|
-        rate = learning_rates.index(rate) # convert word to number
+        rate_value = learning_rates.index(rate_value) # convert word to number
```
rpherbig, line 876: I'm not a fan of reassigning to "input" parameters (loop param in this case). Maybe `rate_as_word` on line 875 and `rate_as_number` on line 876?

MahtraDR:
@rpherbig I've refactored the two sections for detecting what town bank we're in, and looks good in testing.

rpherbig (approved):

PR #6532, opened 2023-02-27
[scripts][dependency] Rubocop

MahtraDR:
As per title.


MahtraDR:
I've been running this version, no visible issues, and the changes are all very small.

PR #6189, opened 2022-10-17
[script][inventory-manager] support for shadow servants

GreyhallowDR:
added an arg 'servant', casts PG on servant 
checks DRStats for guild
confirms successful PG cast on servant
sorry about the spacing

PR #6554, opened 2023-02-28
[script][inventory-manager] Servants in Inventory-Manager

Raykyn55:
Reposting a previous PR with some modifications

### DIFF
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -8,7 +8,6 @@ class InventoryManager
   def initialize
     @active_character = Char.name.to_sym
     @item_db_path = "data/inventory.yaml"
-    @pg_prep_time = get_settings.pg_prep_time
 
     arg_definitions = [
       [
@@ -210,36 +209,19 @@ class InventoryManager
       @item_data[@active_character] = []
     end
 
-    # Default behavior is a full prep, which can be modified with yaml setting pg_prep_time.
-    # This can be an integer representing prep time in seconds, nil for default, or anything else for snapcast
-    # DRCA.safe_release # This would be an appropriate place for setting a release block
-    DRCA.prepare?('pg', 5)
-    if @pg_prep_time.is_a? Integer
-      echo "Pausing #{@pg_prep_time} seconds"
-      sleep @pg_prep_time
-    elsif @pg_prep_time.nil?
-      waitcastrt?
+    pg_pause = [300 / DRSkill.getrank("Utility") + 1, 5].min
+
+    DRCA.cast_spell?({'abbrev'=>'pg', 'mana'=>5, 'prep_time'=>pg_pause, 'cast'=>'cast servant'}, get_settings)
+    while (line = get)
+      break if line =~ /Your servant/i
+
+      item = line.lstrip.concat(' (servant)')
+      @item_data[@active_character] << item
     end
 
-    case DRC.bput('cast servant', 'Within the belly', 'The center of your vision', 'backfires')
-    when 'The center of your vision'
-      DRC.message "Your servant is not present. Summon it and retry."
-    when 'backfires'
-      DRC.message "Your prep time is too low, set pg_prep_time a little higher or remove the setting for default long cast behavior"
-      @pg_prep_time = nil
-      add_servant_inv
-    when 'Within the belly'
-      while (line = get)
-        break if line =~ /Your servant/i
-
-        item = line.lstrip.concat(' (servant)')
-        @item_data[@active_character] << item
-      end
-
-      DRC.message "Saving shadow servant data for #{@active_character}!"
-      UserVars.servant_last_updated = Time.now
-      File.open(@item_db_path, 'w') { |file| file.write(@item_data.to_yaml) }
-    end
+    DRC.message "Saving shadow servant data for #{@active_character}!"
+    UserVars.servant_last_updated = Time.now
+    File.open(@item_db_path, 'w') { |file| file.write(@item_data.to_yaml) }
   end
 
   def add_eddy_inv
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -220,7 +220,7 @@ class InventoryManager
       while (line = get)
         break if line =~ /Your servant/i
 
-        item = line.lstrip.concat(' (Shadow Servant)')
+        item = line.lstrip.concat(' (servant)')
         @item_data[@active_character] << item
       end
 
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -196,7 +196,7 @@ class InventoryManager
     DRC.message "Saving register data for #{@active_character}!"
     File.open(@item_db_path, 'w') { |file| file.write(@item_data.to_yaml) }
   end
-  
+
   def add_servant_inv
     unless DRStats.moon_mage?
       DRC.message "You're not a Moon Mage! Try again."
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -204,7 +204,7 @@ class InventoryManager
     end
 
     if @item_data[@active_character]
-      @item_data[@active_character] = @item_data[@active_character].select { |line| !line.include?'servant' }
+      @item_data[@active_character].reject! { |line| line.include?('servant') }
     else
       @item_data[@active_character] = []
     end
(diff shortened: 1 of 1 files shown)


Raykyn55:
cleaned up to pass the rubocop. out for testing, will report back. This is largely a repost of #6189

MahtraDR:
Let's not add a `pg_prep_time` setting. Why not instead call `buff pg` (like the crafting scripts do) error out if there isn't a pg waggle set, or if there's no waggle for pg, use a hard-coded min-prep 5 second cast? I don't mind which, I just don't want a setting for this, as that doesn't make sense to me.

Raykyn55:
> Let's not add a `pg_prep_time` setting. Why not instead call `buff pg` (like the crafting scripts do) error out if there isn't a pg waggle set, or if there's no waggle for pg, use a hard-coded min-prep 5 second cast? I don't mind which, I just don't want a setting for this, as that doesn't make sense to me.

I think adding this as an option, rather than a requirement, makes the most sense. We can add a note in base-help for the setting, even a blank entry in base if so desired, with a description of it's use. This would be a purely optional setting, used only in cases where a person wanted to shorten their prep time in this specific instance. 

Someone might want to have PG up on themselves as a normal buff, but you only really need a min cast to see the servant's contents. But supposing, for instance, we had a moonie who only ever wanted to use PG to cast on their servant, lets call it our best case scenario. That means defining, at minimum, a PG cast in it's own waggle:

```yaml
waggle_sets:
  piercing_gaze:
    Piercing Gaze:
      mana: 5 # optional here
      prep_time: 5 # optional here
      cast: cast servant # required, but can be assumed by the script
```

which would be required for anyone to use this feature. Spell name never changes, cast command never changes, and mana wouldn't need to be anything over 5. the only thing you'd ever "set" here is prep time, and that only if you wanted to snap or near-snap cast it, and only in cases where you're using PG on an item, in this case your servant. That means we're writing out an entire waggle rather than a single line you can add anywhere to set the only variable we occasionally care about.

Ignoring the naming issues (can't call it servant, piercing_gaze would likely already be a waggle), we're asking the user to provide this waggle *just to use the feature* when in actuality, we don't need any of that information, only the prep time(and only if user wanted to set it shorter).

Furthermore, using buff, we're likely to miss the actual contents of the servant while buff script is wrapping up (wait_for_script_to_complete). I haven't tested this, I could be wrong here, but that is my instinct.

MahtraDR:
> > Let's not add a `pg_prep_time` setting. Why not instead call `buff pg` (like the crafting scripts do) error out if there isn't a pg waggle set, or if there's no waggle for pg, use a hard-coded min-prep 5 second cast? I don't mind which, I just don't want a setting for this, as that doesn't make sense to me.
> 
> I think adding this as an option, rather than a requirement, makes the most sense. We can add a note in base-help for the setting, even a blank entry in base if so desired, with a description of it's use. This would be a purely optional setting, used only in cases where a person wanted to shorten their prep time in this specific instance.
> 
> Someone might want to have PG up on themselves as a normal buff, but you only really need a min cast to see the servant's contents. But supposing, for instance, we had a moonie who only ever wanted to use PG to cast on their servant, lets call it our best case scenario. That means defining, at minimum, a PG cast in it's own waggle:
> 
> ```yaml
> waggle_sets:
>   piercing_gaze:
>     Piercing Gaze:
>       mana: 5 # optional here
>       prep_time: 5 # optional here
>       cast: cast servant # required, but can be assumed by the script
> ```
> 
> which would be required for anyone to use this feature. Spell name never changes, cast command never changes, and mana wouldn't need to be anything over 5. the only thing you'd ever "set" here is prep time, and that only if you wanted to snap or near-snap cast it, and only in cases where you're using PG on an item, in this case your servant. That means we're writing out an entire waggle rather than a single line you can add anywhere to set the only variable we occasionally care about.
> 
> Ignoring the naming issues (can't call it servant, piercing_gaze would likely already be a waggle), we're asking the user to provide this waggle _just to use the feature_ when in actuality, we don't need any of that information, only the prep time(and only if user wanted to set it shorter).
> 
> Furthermore, using buff, we're likely to miss the actual contents of the servant while buff script is wrapping up (wait_for_script_to_complete). I haven't tested this, I could be wrong here, but that is my instinct.

Then have them create a second waggle for this script, that has the prep time they want? It doesn't make sense to add a setting in base.yaml for something we can use a waggle for.

Raykyn55:
> Then have them create a second waggle for this script, that has the prep time they want? It doesn't make sense to add a setting in base.yaml for something we can use a waggle for.

If it's that much of an issue, i can hard code a 3-5 second pause for the cast. Or do a skill check to find the appropriate prep time, something like:
```ruby
pg_pause = [300 / DRSkill.getrank("Utility") + 1, 5].min
```
so < 150 gets you 3 seconds, 151-300 gets you 2 seconds, 301+ gets you 1 seconds prep. shouldn't see anything lower than 200 anyways, but if you did, it would cap at 5 second prep. 1 second prep leaves a safe threshold for silly DR magic stuffs with true code-speed snap casts.

MahtraDR:
There's no need to do that. This is all bypassed by them just creating their own waggle.
```ruby
    DRC.wait_for_script_to_complete('buff', ['inventorypg'])
```

You can do a quick min-cast PG as a backup if they don't have the waggle set, this is what theurgy.lic does.
```ruby
    return unless shop['needs_bless'] && DRSpells.known_spells.include?('Bless')

    complete_or_interrupt_research
    DRCA.cast_spell({ 'abbrev'    => 'bless',
                      'mana'      => 1,
                      'prep_time' => 2,
                      'cast'      => "cast my #{item[:name]}" }, @settings)
```

There's no need to make this more complicated than it needs to be. :)

Raykyn55:
> There's no need to make this more complicated than it needs to be. :)

We are definitely operating on different definitions of the word 'complicated'.

Raykyn55:
How about this:
```ruby
    pg_pause = [300 / DRSkill.getrank("Utility") + 1, 5].min    
    DRCA.cast_spell?({'abbrev'=>'pg', 'mana'=>5, 'prep_time'=>pg_pause, 'cast'=>'cast servant'}, get_settings)
```
no new settings, uses common-arcana to do the cast, sets a simple skill-based timer for the prep time.

MahtraDR:
> How about this:
> 
> ```ruby
>     pg_pause = [300 / DRSkill.getrank("Utility") + 1, 5].min    
>     DRCA.cast_spell?({'abbrev'=>'pg', 'mana'=>5, 'prep_time'=>pg_pause, 'cast'=>'cast servant'}, get_settings)
> ```

Is that an established formula? Where does it it come from? Also note, using the buff method, you can ensure the cast succeeds neatly with `buff <waggle_set> strict` As is here, that DRCA.cast? has a return value, but it's never checked, so if someone backfires PG, they won't be able to proceeed.

Raykyn55:
> Is that an established formula? Where does it it come from? Also note, using the buff method, you can ensure the cast succeeds neatly with `buff <waggle_set> strict` As is here, that DRCA.cast? has a return value, but it's never checked, so if someone backfires PG, they won't be able to proceeed.

Mahtra, it's not a buff when used in this fashion. You cast it at the servant, and get a list of items. This is why a waggle is kind of missing the point. Using cast? would show us a backfire, true, but it wouldn't show us, say, a missing servant. We'd get a successful cast, and an endless loop. Same with using buff. if buff <waggle_set> strict *ever* finished, it would be because we cast pg on ourself, and failed to probe the servant.

MahtraDR:
> > Is that an established formula? Where does it it come from? Also note, using the buff method, you can ensure the cast succeeds neatly with `buff <waggle_set> strict` As is here, that DRCA.cast? has a return value, but it's never checked, so if someone backfires PG, they won't be able to proceeed.
> 
> Mahtra, it's not a buff when used in this fashion. You cast it at the servant, and get a list of items. This is why a waggle is kind of missing the point. Using cast? would show us a backfire, true, but it wouldn't show us, say, a missing servant. We'd get a successful cast, and an endless loop. Same with using buff. if buff <waggle_set> strict _ever_ finished, it would be because we cast pg on ourself, and failed to probe the servant.

That's a fair point. I didn't realize it had to be cast at the SS. We'll have to check the  success differently, but everything else still applies. You can have a 
```yaml
cast: cast servant
```
or whatever the correct cast syntax is in the waggle.

Raykyn55:
Going to close this, the changes herein are already incorporated into the rewrite #6556

PR #6556, opened 2023-02-28
[script][inventory-manager] Rewrite

Raykyn55:
So, playing with inventory manager I noticed some redundant code, a few other small issues. Changes:
1. Use common-items for get/stow
2. collect container searching into one method
3. brought in rummage from common-items to allow rummaging vaults
4. collect removal of previous entries into one method
5. renamed a few entries using proper names eg 'vault' to 'Personal Vault', 'portal' to 'Eddy Portal' mostly for visual
6. since personal inventory catches portal, add a removal for portal entries prior to doing personal inventory

This includes the changes from #6554, so we'll leave it in draft until that's sorted.

### DIFF
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -197,7 +197,7 @@ class InventoryManager
   def add_register_inv
     return unless DRCI.get_item?('register')
 
-    remove_previous_entries('register')
+    remove_previous_entries('Deed Register')
 
     DRC.bput("turn my register to contents", 'You flip your deed register', 'already at the table of contents')
 
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -8,7 +8,6 @@ class InventoryManager
   def initialize
     @active_character = Char.name.to_sym
     @item_db_path = "data/inventory.yaml"
-    @pg_prep_time = get_settings.pg_prep_time || 3
 
     arg_definitions = [
       [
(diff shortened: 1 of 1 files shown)


Raykyn55:
Most of that commit-spam is chasing rubocop "offenses", but a few things were changed since inittial comment:
1. Added option to search for an item (or item group, more below) on a specific player
2. Using search for player, you can get a list of items in a grouping eg deed register on a specific player
3. Dropped default prep for PG down to 3, which bumps to 5 on a backfire. I tested snap-casting PG on a moonie with ~110 utility (shadow servant requires ~220 to even cast the spell) and anything short of an actual snapcast fired successfully with no issues. If you can't get it off at 5 second prep, there's likely another reason.
4. added a cleanup routine at the top, to remove old data from the yaml when this goes live
5. narrowed our removal parameters to categories in parenthesis eg (worn) instead of worn, to prevent false positives.

Raykyn55:
The concern for Mahtra is the pg cast, where he believes a waggle ought to be required as a first choice, with the default pause written in as a backup. What's currently implemented is a "hard coded" progressive pause based on Utility skill: 
< 250 3+ seconds between prep and cast of pg
251-500 2 second pause between prep and cast of pg
501+ 1 second pause between prep and cast of pg

I'd be curious what the moonies have to say, if any wanted to chime in.

MahtraDR:
Post for testers on discord?

Raykyn55:
> Post for testers on discord?

posted. Fuss and Crows are already using a version of this.

Raykyn55:
So, it's been brought to my attention by testers that the family vault could be an issue. As previously written, it created a new entry, almost like a new character, for the family vault. Problem here, is if you have two (or more) accounts, you'll end up with a single family vault constantly overwriting the other. On this new version, you'd pick a character to save the family vault, and just update with that person, for that account, and it would be listed under their character. However, it would still be searchable by:
```
;inv search "family vault"
```

Thoughts? Should i revert to the old style or keep this?

MahtraDR:
> posted. Fuss and Crows are already using a version of this.

A version of this, or this version?

MahtraDR:
> So, it's been brought to my attention by testers that the family vault could be an issue. As previously written, it created a new entry, almost like a new character, for the family vault. Problem here, is if you have two (or more) accounts, you'll end up with a single family vault constantly overwriting the other. On this new version, you'd pick a character to save the family vault, and just update with that person, for that account, and it would be listed under their character. However, it would still be searchable by:
> 
> ```
> 
> ;inv search "family vault"
> 
> ```
> 
> 
> 
> Thoughts? Should i revert to the old style or keep this?

What's the old style?

Raykyn55:
> > posted. Fuss and Crows are already using a version of this.
> 
> A version of this, or this version?

Yes.

>What's the old style?

Create a header for family vault. It's in the comment there.

mwest152:
Script has been working great.  Tested registers, servants, and regular vaults.  Don't have a family vault to test.  Also been using the waggle for prep with no issue.  One function I do use a lot that can maybe be added to the help section is the ability to call a full list for the specific vaults for a character, or to call my servant inventory. 

,inv servant Fuss to call my servant inventory from any character
,inv vault (character name)

Without the " character name" argument it doesn't return anything which is good, but not sure if it is intuitive without the help info.

Raykyn55:
> Script has been working great. Tested registers, servants, and regular vaults. Don't have a family vault to test. Also been using the waggle for prep with no issue. One function I do use a lot that can maybe be added to the help section is the ability to call a full list for the specific vaults for a character, or to call my servant inventory.
> 
> ,inv servant Fuss to call my servant inventory from any character ,inv vault (character name)
> 
> Without the " character name" argument it doesn't return anything which is good, but not sure if it is intuitive without the help info.
> 
> Not a big deal, but could it be possible to call up registers in the same way?

so you can call all of those things with search:
```
;inventory-manager search "deed register" fuss
;inv search register fuss
;inv search vault fuss
;inv search servant fuss
```
any of those without the character name will pull up all information for that category:
```
inv search register
```
would pull up register info for all of your characters.

mwest152:
> 

Oh that's dope and way simpler to understand.

mikejcook:
@MahtraDR I've been starting to use inventory-manager, and there are some good features in this. It sounds like the testers gave a thumbs up. Do you know if there's any compelling reason not to look at merging it?

MahtraDR:
> @MahtraDR I've been starting to use inventory-manager, and there are some good features in this. It sounds like the testers gave a thumbs up. Do you know if there's any compelling reason not to look at merging it?

Not really, as such. What could be done in this script that would make it really nice, and easy to add more fields to, is to use the `Lich::Util.issue_command` functionality and rethink how each type of inventory managed is stored. We could have one function that gets passed starting and ending lines, and just process the output. It would drastically simplify the existing functionality in the script, and make it much simpler to add more types of inventory storage. In any case, if this works better than the existing script, submit it again and show some test results and we can merge it.

PR #6562, opened 2023-03-02
Add hunting-buddy and t2 to setupaliases.lic

vtcifer:
as listed.

### DIFF
--- a/setupaliases.lic
+++ b/setupaliases.lic
@@ -16,7 +16,7 @@
 
   # toggles training-manager on or off
   ['tm', "#{$clean_lich_char}eq Script.running?('training-manager') ? kill_script('training-manager') : start_script('training-manager')"],
-  
+ 
   # toggles hunting-buddy on or off
   ['hb', "#{$clean_lich_char}eq Script.running?('hunting-buddy') ? kill_script('hunting-buddy') : start_script('hunting-buddy', %w(\\?))"],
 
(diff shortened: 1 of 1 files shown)


vtcifer:
I don't see the trailing whitespace rubocop is mentioning here... if I read that right it should be on line 19?  Is it complaining about the extra CR/LF that was technically there before?

setupaliases.lic
```diff
@@ -16,6 +16,15 @@
 
   # toggles training-manager on or off
   ['tm', "#{$clean_lich_char}eq Script.running?('training-manager') ? kill_script('training-manager') : start_script('training-manager')"],
+
```
MahtraDR, line 19:
```suggestion

```

MahtraDR, line 19:
```suggestion
 
```

MahtraDR: Line 19 is a space. Not just a newline/carriage return.

setupaliases.lic
```diff
@@ -16,6 +16,15 @@
...
```
MahtraDR, line 19:
```suggestion

```

setupaliases.lic
vtcifer: Ah, when I went back to my repo in the web editor, I must have been on the wrong branch.  Thanks!

MahtraDR: All good. :)

PR #6568, opened 2023-03-04
[data] [base-spells.yaml] updating aewo and care

Kiriawen:
Aewo and Care are non cyclics now.  Aewo is a normal debil and care is a normal buff.


MahtraDR:
Could you please just delete the two cyclic lines? It defaults to false.

Kiriawen:
done

PR #6569, opened 2023-03-05
Create selfrepair.lic

Raykyn55:
Script to use wire and brush to repair tools/weapons/armor. each arg would be the name of a gear item, or in the case of tools, the name of a tools list (forging, outfitting, etc) as defined in your yaml under the corresponding discipline_tools: setting.


Raykyn55:
hm

PR #6570, opened 2023-03-06
Create repair.lic

Raykyn55:
All in one repair script. Can be run with the listed options or any combination of tools, gear, etc.
    Only restricted in pickup, the pickup routine will go to the last repair location, or the location you designate,
    if you have any tickets from that repair person.

  EXAMPLES:
  ;repair forging                    # runs to your hometown repair shop, drops off and picks up your forging tools
  ;repair forging self_repair        # Uses wire brush and oil to repair your own forging tools as defined in yaml
  ;repair forging crossing           # runs to crossing repair shop, drops off and picks up your forging tools
  ;repair forging drop_off shard     # runs to shard, drops your forging tools at the repair shop for pickup at a later time.
  ;repair shield "iron bracer" pilum forging shard drop_off
                                     # runs to shard, drops off your shield, bracer, pilum, and forging tools for pickup at a later time.

  (probably works)

### DIFF
--- a/repair.lic
+++ b/repair.lic
@@ -52,6 +52,11 @@ class Repair
       @toolset_list   << @settings.send(name + '_tools')
     end
 
+    if @town == 'Fang Cove' && fang_closed?
+      DRC.message("Fang Cove repair personnel are sleeping off their latest hangover, and aren't available for repairs at this hour. Try back later")
+      exit
+    end
+
     unless (repair_info = get_data('town')[@town]['metal_repair']) || args.self_repair
       DRC.message("No repair info found for #{@town}, exiting")
       reset_town
@@ -216,6 +221,11 @@ class Repair
 
     Flags.reset('proper-repair')
     UserVars.immune_list.store(gear_item, Time.now + 7000)
+  end
+
+  def fang_closed?    
+    tod = DRC.bput('time', /^It is currently .* and it is .*./)
+    ['evening', 'night', 'sunrise', 'dawn', 'early morning'].any? { |closed| tod.include?(closed) }
   end
 
   def pickup_repaired_items(repair_info)
--- a/repair.lic
+++ b/repair.lic
@@ -128,7 +128,7 @@ class Repair
 
   def repair_gear(gear, info, repair_own = false)
     verify_funds(gear, repair_own)
-    DRCT.walk_to(repair_info['id']) unless repair_own
+    DRCT.walk_to(info['id']) unless repair_own
 
     gear.each do |gear_item|
       unless smart_get_gear(gear_item)
--- a/repair.lic
+++ b/repair.lic
@@ -43,11 +43,11 @@ class Repair
     @repair_timer   = @settings.repair_timer
     @sort_head      = @settings.sort_auto_head
     @disciplines    = ['forging', 'tinkering', 'carving', 'shaping', 'outfitting', 'alchemy', 'enchanting', 'engineering']
-    @toolbelts_list = []
-    @toolset_list   = []
-    @craft_data     = get_data('crafting')['blacksmithing'][@town.capitalize]
+    @craft_data     = get_data('crafting')['blacksmithing'][@town]
     @equipmanager   = EquipmentManager.new
 
+    @toolbelts_list = []
+    @toolset_list   = []
     @disciplines.each do |name|
       @toolbelts_list << @settings.send(name + '_belt')
       @toolset_list   << @settings.send(name + '_tools')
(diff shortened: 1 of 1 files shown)


MahtraDR:
I love it.

We'll need to think about the impact on other scripts. If combat equipment pickup fails, we need to have a check somewhere for that so we don't send folks to their combat routine without armor.

Raykyn55:
> I love it.
> 
> We'll need to think about the impact on other scripts. If combat equipment pickup fails, we need to have a check somewhere for that so we don't send folks to their combat routine without armor.

repair and/or hinderance reduction. They'll thank me later.

Raykyn55:
On a serious note, I don't see this as a problem we need to address. IF the user opts to split up drop and pickup, it's on the user to pick up before going into combat. Running the script without the drop_off argument runs a normal drop/pickup routine. You have to specifically tell the script not to pick up.

MahtraDR:
> On a serious note, I don't see this as a problem we need to address. IF the user opts to split up drop and pickup, it's on the user to pick up before going into combat. Running the script without the drop_off argument runs a normal drop/pickup routine. You have to specifically tell the script not to pick up.

In that scenario I agree. 

I was thinking if a scenario where they run the get too soon, not all armor pieces are ready yet. Does it just wait till they all are?

Raykyn55:
> > On a serious note, I don't see this as a problem we need to address. IF the user opts to split up drop and pickup, it's on the user to pick up before going into combat. Running the script without the drop_off argument runs a normal drop/pickup routine. You have to specifically tell the script not to pick up.
> 
> In that scenario I agree.
> 
> I was thinking if a scenario where they run the get too soon, not all armor pieces are ready yet. Does it just wait till they all are?

Yes, it picks up every single item from that repair shop. Basically just grabs tickets til you're out, it doesn't try to track items.

repair.lic
```diff
@@ -0,0 +1,227 @@
...
+    else
+      DRC.message("Last repair town not defined.")
+    end
+    exit if leave
+  end
+
+  def should_repair?
+    return true if @repair_timer.nil?
+
+    @repair_timer >= (Time.now - UserVars.repair_timer_snap).to_i
```
Raykyn55, line 116: Brain isn't cooperating today, could someone verify this is accurate? I *think* this is saying, if the time difference (in seconds) between the last time you repaired (repair_timer_snap) and the present time, which would be elapsed time. is GREATER THAN your repair timer, return true.

Raykyn55, line 116:
```suggestion
    @repair_timer <= (Time.now - UserVars.repair_timer_snap).to_i
```
No, I was right the first time.

MahtraDR: This is checking if `@repair_timer` is greater than, or equal to `(Time.now - UserVars.repair_timer_snap).to_i` if it is, it'll return `true`.

MahtraDR: Would you still want to respect the repair timer variable from crossing-repair?

Raykyn55:
For example, in my t2, above every hunt:
```
[t2: ***STATUS*** EXECUTE repair pick_up]
--- Lich: repair active.
Last repair town not defined.
[repair]>tap my Catrox ticket 
>
--- Lich: repair has exited.
```

Had there been a ticket, it would pick up every one, waiting on ones that aren't done.

repair.lic
Raykyn55: > Would you still want to respect the repair timer variable from crossing-repair?

Yeah, but only for a full repair. You could run, say, your shield and parry stick every time, and have a general repair you only want to call every 6 hours.

Can even just drop those two off, go do stuff, then it'll grab em next time you do a pickup.

Sanzia:
Basically tested everything and no issues and works great!  Here is what I sent and performed as expected.
;repair help
;repair self_repair shield
;repair drop_off "fang cove" plate (FC was open, so didn't see if the closed break works)
;repair pick_up "fang cove" plate
;repair town     (Was in FC and it reset to YAML setting of Crossing)
;repair drop_off balaclava     (Tvl from FC to Xing)
;repair drop_off "Shard" gloves
;repair pick_up Shard
;repair reset_town
;repair gaunt     (Tvl from Shard to Xing - Also note it matched gaunt for gauntlets.)
;repair     (Did it right...even ran out coin so it got more and picked right back up)
(After my big repair session, tested the force function)
>;repair
--- Lich: repair active.
Last repair was 3 minutes ago, which is less than your repair_timer settings
Currently set to repair only once every 386 minutes. Run with arg force_repair to repair anyways.
--- Lich: repair has exited.
>;repair shield force_repair (Tvl to Xing to Repiar Shield)

Raykyn55:
> Basically tested everything and no issues and works great!

Thanks for testing! Very thorough.

PR #6572, opened 2023-03-07
[data][base-recipes] Wrong part for Celestial Beacon

Raykyn55:
Incorrectly states celestial beacon uses a bead, when it requires a gem, and shouldn't be eligible for workorders.


Raykyn55:
https://elanthipedia.play.net/Item:Celestial_beacon_(crafted)

Raykyn55:
https://discord.com/channels/745675889622384681/745762938991804556/1082600900390027324

PR #6574, opened 2023-03-08
[script][locksmithing] add support for multiple lockboxes

mikejcook:
Added support for multiple lockboxes. Removed lockbox script, as this functionality is covered by locksmithing. Cleaned up locksmithing in general.

@MahtraDR Here's the PR for the locksmithing changes we discussed. I don't have a way to test multiple lockboxes, but it works correctly with one. In theory, it should work with a second, but we'll need someone with 2 or more to check that out. I also don't have a trader to test swapping lockbox to locksmithing in trade.


mikejcook:
Rubocop has an issue with trade unrelated to my changes, but I really don't even know what the code is trying to accomplish, so I'm hesitant to touch it.

Raykyn55 (commented):
I think for consumables we want to do for each, and then loop through that box type until we hit our mindstate, or run out of tha type. So:
```ruby
@consumable_trainers.each do |box|
  break if <exp check>

  while <get this type box>
  <pick this box>
  end
end
```

So we're getting keepsake boxes until we run out, then jewelry  boxes, etc. So that we don't go through one keepsake and one jewelry with low charges while we have a bunch more in our bags.

I see the use of ordinals here, for trainers, but what's missing is some way to handle two different named trainers. Something similar to how we do consumables would make the most sense. In fact, a method that simply iterates over an array of box names, such as in @consumable_trainers would do the trick, 
```ruby
box_list.each do |box|
  break if <exp chek>

  while <get this # type box>
  <pick this box>
  ords.shift if @trainer # something to indicate we're doing training boxes and not consumables
  stow if @trainer
end
```
then call that method for both consuambles and training_boxes

mikejcook:
@Raykyn55 So, I tried to limit how much I changed the script. The majority of your comments are on code that was already in this script. I'm not sure if we want to totally rewrite the script at this stage. I guess we could, but it wasn't really in the scope of what I was trying to do. Thoughts? My goal was adding the ability to support multiple versions of a lockbox. I cleaned up a few things, but I really avoided a total rewrite.

Raykyn55:
> @Raykyn55 So, I tried to limit how much I changed the script. The majority of your comments are on code that was already in this script. I'm not sure if we want to totally rewrite the script at this stage. I guess we could, but it wasn't really in the scope of what I was trying to do. Thoughts? My goal was adding the ability to support multiple versions of a lockbox. I cleaned up a few things, but I really avoided a total rewrite.

Totally up to you! 

I'm only offering suggestions, here. If you disagree, I won't be the least bit offended! I'm the king of creep, every little edit turns into a rewrite, but I treat scripts how a mechanic treats a car. I might be going in for an oil change, but if the belts are frayed and the ball joints are a mess, I may as well sort it while it's on the lift.

mwest152:
What needs testing? I use this script to open a couple hundred boxes a day.  I have dozens of consumables as well.

mikejcook:
@Raykyn55 So, I've been totally rewriting this. Here's the box loop I have, but I'm not sure I love it. Can you look at it and see if you have any thoughts on how to improve it?


```
  def box_loop(lockboxes, container, lockbox_quantity)
    lockboxes.each do |box|
      break if DRSkill.getxp('Locksmithing') >= @exp_max_threshold

      lockbox_quantity.times do |i|
        break if DRSkill.getxp('Locksmithing') >= @exp_max_threshold

        # We will keep getting boxes until we run out, but for dailies, we will exit this loop after the first time
        while DRCI.get_item("#{$ORDINALS[i]} #{box}", container)
          pick_box(box)

          # Only consumables have a container defined, so if container is null, that means we're picking dailies
          if container.nil?
            DRCI.stow_hand('right') if DRC.right_hand && !@worn_lockbox
            break
          else
            DRCI.put_away_item?(box, container)
          end
        end
      end
    end
  end
  ```

I call it with one of these:

```
box_loop(@daily_boxes, nil, @lockbox_quantity)

box_loop(@consumable_trainers, @consumable_container, 1)
```

**Comment by Raykyn55** [comment]
So, poking at this a little on my end, taking into account:
1. multiple consumable boxes, which we want to use first
2. multiple training boxes, which we do not want to discard

```ruby
 lockbox_quantity.times do |i|
 ```
So, what happens if I have three training boxes, two being 'locksmith padlock' while the third is 'locksmith pouch'? 

```ruby
DRCI.stow_hand('right') if DRC.right_hand && !@worn_lockbox
```
If you're using a loose lockpick, your right hand is going to be that loose lockpick.

My little test script:
```ruby
@worn_lockbox = false
@bag = 'pack'
@live_boxes = false
@training_box_get = @worn_lockbox ? :remove_item? : :get_item?
@consumable_lockboxes = ['keepsake box', 'jewelry box']
@training_boxes = ['house shoes', 'jousting shield']

def pick_box(box)
  fput "tap my #{box}"
end

DRC.wait_for_script_to_complete('pick') if @live_boxes

@consumable_lockboxes.each do |box|
  while DRSkill.getxp('Locksmithing') < 34 && DRCI.get_item?(box, @bag)
    pick_box(box)
    DRCI.put_away_item?(box, @bag) if DRCI.in_hands?(box)
  end
end

@training_boxes.each do |box|
  @ords = $ORDINALS.dup
  while DRSkill.getxp('Locksmithing') < 34 && DRCI.send(@training_box_get, "#{@ords.shift} #{box}")
    pick_box(box)
    DRCI.wear_item?(box) if @worn_lockbox
    DRCI.put_away_item?(box, @bag) if DRCI.in_hands?(box)
  end
end
```
```
>;locktest
--- Lich: locktest active.
Picking live boxes
[locktest]>get my keepsake box from my pack
What were you referring to?
>
[locktest]>get my jewelry box from my pack
What were you referring to?
>
[locktest]>remove my first house shoes
You take some gentleman's house shoes of luxurious purple khaddar off your feet.
>
[locktest]>tap my house shoes
You tap some gentleman's house shoes of luxurious purple khaddar that you are holding.
>
[locktest]>wear my house shoes
You put some gentleman's house shoes of luxurious purple khaddar onto your feet.
>
[locktest]>remove my second house shoes
Remove what?
>
[locktest]>remove my first jousting shield
You loosen the straps securing your jousting shield and remove it.
>
[locktest]>tap my jousting shield
You tap a silversteel jousting shield with glaes spikes that you are holding.
>
[locktest]>wear my jousting shield
You slide your left arm through the straps on your jousting shield and flex, gaining a feel for the shield's weight.
>
[locktest]>remove my second jousting shield
Remove what?
>
--- Lich: locktest has exited.
```

Raykyn55:
```ruby
    if @daily_boxes.nil?
      @daily_boxes = Array.new
      @daily_boxes.push(@daily_box)
    end
```
We can require the user list daily boxes as an array:
```yaml
daily_boxes:
  - daily box one
  - daily box two
```
Also, this is very neat, I didn't know you could do this:
```ruby
lockbox_quantity.times do |i|
```
which does fix your 0-5 issue :D

Aislynnial:
Would it be at all possible to prioritize (or add the ability to prioritize) consumable boxes over the daily training box? I would so much rather chew through these dozens of consumable boxes and THEN lean into the training box, rather than the other way around.

mikejcook:
> Would it be at all possible to prioritize (or add the ability to prioritize) consumable boxes over the daily training box? I would so much rather chew through these dozens of consumable boxes and THEN lean into the training box, rather than the other way around.

Sure, that's easy enough and makes sense.

mikejcook:
> ```ruby
>     if @daily_boxes.nil?
>       @daily_boxes = Array.new
>       @daily_boxes.push(@daily_box)
>     end
> ```
> 
> We can require the user list daily boxes as an array:
> 
> ```yaml
> daily_boxes:
>   - daily box one
>   - daily box two
> ```
> 
> Also, this is very neat, I didn't know you could do this:
> 
> ```ruby
> lockbox_quantity.times do |i|
> ```
> 
> which does fix your 0-5 issue :D

I don't want to break backwards compatibility. That's really the only reason I implemented it this way.

mikejcook:
@Raykyn55 
Re: So, what happens if I have three training boxes, two being 'locksmith padlock' while the third is 'locksmith pouch'?

So, you have:
padlock
padlock
pouch

You define:
- locksmith padlock
- locksmith pouch
lockbox_quantity: 2

So it'll try to do first locksmith padlock, second locksmith padlock and first locksmith pouch, second locksmith pouch. Second locksmith pouch won't find anything, and it'll just skip over it. So I think that should work well enough? I think anything more is over complicating it more than we need to worry about?

mikejcook:
@Raykyn55 Okay, thanks for the feedback. I've cleaned this up. I added a setting to prioritize consumables and I fixed up the box loop to not be dumb - I feel better about it now. I just need to test (or find someone to test) consumables.

I also added an optional picking_lockbox_container. If it's set, it'll cleanly keep it in the right place. If not, it falls back to stowing the box. But we shouldn't have any issues with lockpicks in hand.

@Aislynnial Would you be able to test this against your consumables?

You would just need to set this in your yaml:
prioritize_consumable_lockboxes: true

Raykyn55:
@mikejcook Hey, can you msg me on discord: Scrounger#9724

PR #6575, opened 2023-03-09
[script][common-money] Incorrectly checking currency

Raykyn55:
So this has been bugging me for awhile, finally tracked down the problem. When passing a hometown that isn't your yaml hometown to this script's method ensure_copper_on_hand, it's supposed to use the hometown arg if provided:
```ruby
  def withdraw_exact_amount?(amount_as_string, settings, hometown = nil)
    hometown = settings.hometown if hometown == nil
```
 Problem is, in this particular spot:
```ruby
        currency = hometown_currency(settings.hometown)
```
it was getting the currency type from your yaml setting, rather than the hometown arg. Changing this to use the local variable instead of getting your hometown from settings sorts the issue.

### DIFF
--- a/common-money.lic
+++ b/common-money.lic
@@ -176,7 +176,7 @@ module DRCM
       if DRRoom.pcs.include?(settings.bankbot_name)
         amount_convert, type = amount_as_string.split
         amount = convert_to_copper(amount_convert, type)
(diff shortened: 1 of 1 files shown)


Raykyn55:
gah, think i snagged the wrong line

PR #6594, opened 2023-03-16
[scripts][stack-scrolls] Allow usage of multiple B&E stackers

urbaj-dr:
Adds special handling for multiple "worn books" (with different full names) to be used alongside each other. Worn books are retrieved from the stacker container after rummaging in the container and pulling out the ORDINAL + WORN BOOK from it.

### DIFF
--- a/stack-scrolls.lic
+++ b/stack-scrolls.lic
@@ -30,6 +30,8 @@ class ScrollStack
 
     args = parse_args(arg_definitions)
 
+    @worn_book_list = []
+
     unless args.container
       if args.get
         search_scrolls(args.query, stacker_container)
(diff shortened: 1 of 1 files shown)


MahtraDR:
This breaks backwards compatibility it seems.

```
2023-03-17 15:19:37 NZDT: >;stack get "aura of tongues"
2023-03-17 15:19:37 NZDT: [stack-scrolls]>get my leather folder from my duffel bag
2023-03-17 15:19:37 NZDT: You get a deep russet leather-covered folder with silk ribbon straps from inside your duffel bag.
2023-03-17 15:19:37 NZDT: [stack-scrolls]>flip my 
2023-03-17 15:19:37 NZDT: [stack-scrolls]>open my 
```

```yaml
##################################################################
# Scroll Stackers
##################################################################
stacker_container: duffel bag
scroll_stackers:
  - scroll carrier
  - scroll tube
  - leather folder
  - thick tome
```

mikejcook:
This is great. I tested this out and was able to take a ton of the scrolls that have been hanging around in bags and get them into three different worn books. I didn't run into the issue Mahtra did with the backwards compatibility, fwiw.

I then added the worn books with their full names and everything worked great! Once that bug is tracked down, this seems like a great addition.

mikejcook:
Oh, I see the issue. It's because he did a search initially, I did a populate.

When you do a search, it finds the name of the stacker and sends that to get_stacker. It's a string, not a hash, so you don't populate the short_name. So then the next two lines immediately fail because short_name is empty. 

See below, how it passes the name to get_stacker, not the actual stacker variable.
```
    target = UserVars.stackers.find { |stacker| stacker['contents'].find { |data| data.first =~ /#{query}/i } }
    unless target
      echo("No scroll found matching: #{query}")
      exit
    end

    get_stacker(target['name'], stacker_container)
```

urbaj-dr:
Revisiting the code on this sometime later, and it seems like there should be a cleaner/safer way to do this. I might return to it eventually.

PR #6599, opened 2023-03-21
[script][theurgy] Lighter support and loop escape

Raykyn55:


Raykyn55:
waiting on #6600 for rework

PR #6600, opened 2023-03-23
[script][common-items] common pyrotechnics

Raykyn55:
common method to empower your inner pyromaniac
1. If warmie, uses burning touch
2. if you have a lighter, and it's properly defined, use that
3. Otherwise, grabs flint and steel and lights your item with that

Lighter yaml setting:
```yaml
lighter_info:
  name: lava drake
  container: smoking jacket
```

Calling script needs to initialize equipmanager script, and pass appropriately formatted item information for the blade via:
```ruby
EquipmentManager.new.item_by_desc(blade)
```

IF successful, ends the method with the lit item in your right hand and a 'true' response. IF unsuccessful, stows your goodies ( or attempts to) and returns false.

Flint method can be called separately, but need to pass it an initialized equipmanager arg to handle weapon.

### DIFF
--- a/common-items.lic
+++ b/common-items.lic
@@ -1240,7 +1240,7 @@ module DRCI
 
     DRCI.stow_item?('flint') if DRCI.in_hands?('flint')
     equipmanager.return_held_gear
-    DRCI.lift?(item.split.last, stow) # If we are successful, simply lifts the item. otherwise, stows it accordingly
+    DRCI.lift?(item, stow) # If we are successful, simply lifts the item. otherwise, stows it accordingly
     !stow # returns our result
   end
 
--- a/common-items.lic
+++ b/common-items.lic
@@ -382,6 +382,14 @@ module DRCI
     /^What is it you're trying to give/
   ]
 
+  @@light_success_patterns = [
+    /^You touch/,
+    /^That('s| is) already burning/,
+    /bursts into flames/,
+    /quickly catch fire/,
+    /^But .* is already lit/
+  ]
+
   #########################################
   # TRASH ITEM
   #########################################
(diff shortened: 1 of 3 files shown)


Raykyn55:
Implementation, if this is the final version, would mean scripts using a lit object would call this method, then do whatever it does with the lit object. For example, common-theurgy's wave_incense method would be truncated to simply the waving of lit incense, while CT would need to call this method directly before calling wave_incense, to ensure we've got a lit piece of incense.

Smoke might have trouble using it since you're lighting 'tobacco in pipe', so may have to keep it's own lighting logic specifically for pipes.

common-items.lic
```diff
@@ -382,6 +382,14 @@ module DRCI
     /^What is it you're trying to give/
   ]
 
+  @@light_success_patterns = [
```
MahtraDR, line 385:
```suggestion
  LIGHT_SUCCESS_PATTERN = [
```

common-items.lic
```diff
@@ -1185,4 +1193,58 @@ module DRCI
...
+    end
+
+    DRCI.stow_item?('flint') if DRCI.in_hands?('flint')
+    equipmanager.return_held_gear
+    DRCI.lift?(item, stow) # If we are successful, simply lifts the item. otherwise, stows it accordingly
+    !stow # returns our result
+  end
+
+  def ready_set_light(command)
+    Regexp.new(LIGHT_SUCCESS_PATTERNS.join('|')) =~ DRC.bput(command, { 'timeout' => 5, 'suppress_no_match' => true }, @@light_success_patterns)
```
MahtraDR, line 1248:
```suggestion
    Regexp.new(LIGHT_SUCCESS_PATTERNS.join('|')) =~ DRC.bput(command, { 'timeout' => 5, 'suppress_no_match' => true }, LIGHT_SUCCESS_PATTERNS)
```

Raykyn55: I want to say I tried that, and it didn't work  The variable is just an array, not regex. missed the change on the end, though, thanks for the spot.

Raykyn55: I'll test this new variant when I get a chance.

MahtraDR: If you get just the array, don't do the join, you can just do a `Regexp.union` instead, which is much cleaner than the join.

MahtraDR:
```
>;eq echo Regexp.union(DRCI::LIGHT_SUCCESS_PATTERNS)
[exec1: (?-mix:(?-mix:^You touch)|(?-mix:^That('s| is) already burning)|(?-mix:bursts into flames)|(?-mix:quickly catch fire)|(?-mix:^But .* is already lit))]
>;eq echo Regexp.new(DRCI::LIGHT_SUCCESS_PATTERNS.join('|'))
[exec1: (?-mix:(?-mix:^You touch)|(?-mix:^That('s| is) already burning)|(?-mix:bursts into flames)|(?-mix:quickly catch fire)|(?-mix:^But .* is already lit))]
```

common-items.lic
```diff
@@ -1185,4 +1193,58 @@ module DRCI
...
+    end
+
+    DRCI.stow_item?('flint') if DRCI.in_hands?('flint')
+    equipmanager.return_held_gear
+    DRCI.lift?(item, stow) # If we are successful, simply lifts the item. otherwise, stows it accordingly
+    !stow # returns our result
+  end
+
+  def ready_set_light(command)
+    Regexp.new(LIGHT_SUCCESS_PATTERNS.join('|')) =~ DRC.bput(command, { 'timeout' => 5, 'suppress_no_match' => true }, LIGHT_SUCCESS_PATTERNS)
```
MahtraDR, line 1248:
```suggestion
    Regexp.union(DRCI::LIGHT_SUCCESS_PATTERNS) =~ DRC.bput(command, { 'timeout' => 5, 'suppress_no_match' => true }, LIGHT_SUCCESS_PATTERNS)
```

common-items.lic
```diff
@@ -1185,4 +1193,58 @@ module DRCI
...
+  # Any script calling this method needs to call script 'equipmanager' to load it's methods and your settings
+  # blade arg is the EquipmentManager formatted version of a weapon, as defined in gear
+  # fetched with EquipmentManager.new.item_by_desc(blade)
+  # lighter is a hash from yaml setting:
+  #   lighter_info:
+  #     name: lava draked
+  #     container: smoking jacket
+  def light_item_on_fire?(item, blade, lighter, item_container = nil)
+    equipmanager = EquipmentManager.new
+    if DRStats.warrior_mage? && /prepared/ =~ DRC.bput("prep c b t", /^You are now prepared/) && ready_set_light("gesture #{item}")
```
MahtraDR, line 1212:
```suggestion
    if DRStats.warrior_mage? && /prepared/ =~ DRC.bput("prep c b t", /^You are now prepared/) && ready_set_light("gesture #{item}")
```

Why not use common arcana here?
```
;eq    DRCA.cast_spell({ 'abbrev'    => 'C B T', 'cast'      => "gesture my cigar" }, @settings)

```

Raykyn55: To what purpose?

PR #6624, opened 2023-04-13
[script][combat-trainer] Enable whirlwind with additional weapons

Raykyn55:
Current CT on a barb, you're limited to one-handed staves and polearms for whirlwind. When you attempt to whirlwind with a twohander, you get a message saying you need to free up your left hand. What this change does is capture that, and add your current weapon skill to the list of twohanded weapon skills for the remainder of the hunt.

In particular with barbarians and maneuvers, CT wields a weapon from your whirlwind list in the offhand, as you can get an extra attack with your offhand on most maneuvers. However, if you're trying to maneuver with a twohander, you'll get the same message about freeing up your offhand, so we'll catch that and add the current weapon to our twohanded skills list. 

The obvious alternative would be a separate setting, with a list of weapon skills that are two-handed:
```yaml
twohanded_weapon_skills:
- Twohanded Edged
- Twohanded Blunt
- Staves
- Polearms
```
Actually, think I'll add this.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3339,10 +3339,10 @@ class AttackProcess
 
       game_state.sheath_offhand_weapon(skill) if offhand_weapon_name
     end
-
-    # If we didn't try a maneuver, or we did but it wasn't successful, then fallback to normal attack.
+    # If maneuver was successful, register an action taken and return
     return game_state.action_taken if maneuver_success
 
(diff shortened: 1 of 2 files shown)


Raykyn55:
Yeah, so if you set those weapon skills as twohanded in your yaml, the handling should treat them like 2handers from the get, but if you do not, CT will figure it out when it tries to use them onehanded, and they'll be added.

PR #6637, opened 2023-05-15
[scripts][combat-trainer] Don't try to switch pouches if left hand is full

MahtraDR:
As per title. Refactored a bit into its own def, and now we just try later when the left hand is empty.

As is, in edge cases where the pouch is full, and the offhand is also full, this could mess up the CT routine by stowing the left hand, and getting it back, into the right hand later.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -501,21 +501,14 @@ class LootProcess
       fput 'DUMP JUNK'
       @dump_timer = Time.now
     end
-    if @autoloot_container && @autoloot_gems && (Time.now - @autoloot_fill_gem_pouch_timer > @autoloot_fill_gem_pouch_delay)
-      DRCI.fill_gem_pouch_with_container(
-        @gem_pouch_adjective,
-        @gem_pouch_noun,
-        @autoloot_container,
-        @full_pouch_container,
-        @spare_gem_pouch_container,
-        @tie_pouch
-      )
-      @autoloot_fill_gem_pouch_timer = Time.now
-    end
+    fill_pouch_with_autolooter if @autoloot_container && @autoloot_gems
 
     game_state.mob_died = false
+
     dispose_body(game_state)
+
     stow_lootables(game_state)
+
     if (game_state.mob_died || game_state.npcs.empty?) && game_state.finish_killing?
       15.times do
         break unless Flags['using-corpse']
@@ -527,9 +520,26 @@ class LootProcess
       echo('LootProcess::clean_up') if $debug_mode_ct
       game_state.next_clean_up_step
     end
+
     return true if game_state.finish_spell_casting? || game_state.stowing?
 
     false
+  end
+
+  def fill_pouch_with_autolooter
+    return unless (Time.now - @autoloot_fill_gem_pouch_timer > @autoloot_fill_gem_pouch_delay)
+    return unless DRC.left_hand.nil?
+
+    DRCI.fill_gem_pouch_with_container(
+      @gem_pouch_adjective,
+      @gem_pouch_noun,
+      @autoloot_container,
+      @full_pouch_container,
+      @spare_gem_pouch_container,
+      @tie_pouch
+    )
+
+    @autoloot_fill_gem_pouch_timer = Time.now
   end
 
   def stow_loot(item, game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -502,7 +502,7 @@ class LootProcess
       @dump_timer = Time.now
     end
 
-    fill_pouch_with_autolooter if @autoloot_container && @autoloot_gems
+    fill_pouch_with_autolooter(game_state) if @autoloot_container && @autoloot_gems
 
     game_state.mob_died = false
 
@@ -527,9 +527,11 @@ class LootProcess
     false
   end
 
-  def fill_pouch_with_autolooter
-    return unless (Time.now - @autoloot_fill_gem_pouch_timer > @autoloot_fill_gem_pouch_delay)
-    return unless DRC.left_hand.nil?
+  def fill_pouch_with_autolooter(game_state)
+    return unless (Time.now - @autoloot_fill_gem_pouch_timer) > @autoloot_fill_gem_pouch_delay
+    return unless (DRC.left_hand.nil? || game_state.currently_whirlwinding)
+
+    game_state.sheath_whirlwind_offhand if game_state.currently_whirlwinding
 
     DRCI.fill_gem_pouch_with_container(
       @gem_pouch_adjective,
@@ -541,6 +543,8 @@ class LootProcess
     )
 
     @autoloot_fill_gem_pouch_timer = Time.now
+
+    game_state.wield_whirlwind_offhand if game_state.currently_whirlwinding
   end
 
   def stow_loot(item, game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -501,6 +501,7 @@ class LootProcess
       fput 'DUMP JUNK'
       @dump_timer = Time.now
     end
+
     fill_pouch_with_autolooter if @autoloot_container && @autoloot_gems
 
     game_state.mob_died = false


Raykyn55:
If this is only a problem with something in the left hand, with the right hand free, why not gate it with that condition? barbs spend basically all of combat with something in their offhand.

MahtraDR:
> If this is only a problem with something in the left hand, with the right hand free, why not gate it with that condition? barbs spend basically all of combat with something in their offhand.

Hmm ok. The idea is to have one hand free. I suppose which hand doesn't matter. I'll look at how the almanac code handles this. Thanks, barb is one of the few guilds I don't have and don't inherently account for.

MahtraDR:
@Raykyn55 Looking at the Almanac code.
```ruby
    return if DRC.left_hand && !game_state.currently_whirlwinding
```
Is that meant to be an `OR` instead? 

In the def https://github.com/rpherbig/dr-scripts/blob/c03053e2e74285dcaff3a2c61a35d80df74995fa/combat-trainer.lic#L2879 we handle the case where the Offhand is the whirlwind offhand weapon.

I want to rewrite this line to be clearer. What issues might we have with
```ruby
return unless DRC.left_hand.nil? || game_state.currently_whirlwinding
```
? This would then check if we have a free hand, or if we're whirlwinding, since the latter we cater for in the def already.

Raykyn55:
> @Raykyn55 Looking at the Almanac code.
> 
> ```ruby
>     return if DRC.left_hand && !game_state.currently_whirlwinding
> ```
> 
> Is that meant to be an `OR` instead?
> 
> In the def
> 
> https://github.com/rpherbig/dr-scripts/blob/c03053e2e74285dcaff3a2c61a35d80df74995fa/combat-trainer.lic#L2879
> 
> we handle the case where the Offhand is the whirlwind offhand weapon.
> I want to rewrite this line to be clearer. What issues might we have with
> 
> ```ruby
> return unless DRC.left_hand.nil? || game_state.currently_whirlwinding
> ```
> 
> ? This would then check if we have a free hand, or if we're whirlwinding, since the latter we cater for in the def already.

So the condition we're concerned about is strict offhand weapon training. With whirlwind, barbs rarely run this routine, if ever. Perhaps return if weapon skill is offhand? Should be sufficiently rare, while also not broad enough to just skip whenever there's something in our left hand.

MahtraDR:
> > @Raykyn55 Looking at the Almanac code.
> > ```ruby
> >     return if DRC.left_hand && !game_state.currently_whirlwinding
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is that meant to be an `OR` instead?
> > In the def
> > https://github.com/rpherbig/dr-scripts/blob/c03053e2e74285dcaff3a2c61a35d80df74995fa/combat-trainer.lic#L2879
> > 
> > we handle the case where the Offhand is the whirlwind offhand weapon.
> > I want to rewrite this line to be clearer. What issues might we have with
> > ```ruby
> > return unless DRC.left_hand.nil? || game_state.currently_whirlwinding
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ? This would then check if we have a free hand, or if we're whirlwinding, since the latter we cater for in the def already.
> 
> So the condition we're concerned about is strict offhand weapon training. With whirlwind, barbs rarely run this routine, if ever. Perhaps return if weapon skill is offhand? Should be sufficiently rare, while also not broad enough to just skip whenever there's something in our left hand.

I'm not sure that the condition we're concerned with as such. The condition we want is a free hand, doesn't technically matter which. The problem with "strict offhand weapon training" is purely that in that one case we would sheathe, and reget, but the weapon would then be in the right hand. I am not a fan of recreating a new set of conditions for this, when the use case is so very similar to the almanac one. I would prefer to keep the existing almanac use case (even if we gently tweak it as mentioned) and re-use that for pouch filling.

I'm leaning towards this
```ruby
  def fill_pouch_with_autolooter(game_state)
    return unless (Time.now - @autoloot_fill_gem_pouch_timer) > @autoloot_fill_gem_pouch_delay
    return unless (DRC.left_hand.nil? || game_state.currently_whirlwinding)

    game_state.sheath_whirlwind_offhand if game_state.currently_whirlwinding

    DRCI.fill_gem_pouch_with_container(
      @gem_pouch_adjective,
      @gem_pouch_noun,
      @autoloot_container,
      @full_pouch_container,
      @spare_gem_pouch_container,
      @tie_pouch
    )

    @autoloot_fill_gem_pouch_timer = Time.now

    game_state.wield_whirlwind_offhand if game_state.currently_whirlwinding
  end
  ```

Raykyn55:
> > > @Raykyn55 Looking at the Almanac code.
> > > ```ruby
> > >     return if DRC.left_hand && !game_state.currently_whirlwinding
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Is that meant to be an `OR` instead?
> > > In the def
> > > https://github.com/rpherbig/dr-scripts/blob/c03053e2e74285dcaff3a2c61a35d80df74995fa/combat-trainer.lic#L2879
> > > 
> > > we handle the case where the Offhand is the whirlwind offhand weapon.
> > > I want to rewrite this line to be clearer. What issues might we have with
> > > ```ruby
> > > return unless DRC.left_hand.nil? || game_state.currently_whirlwinding
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ? This would then check if we have a free hand, or if we're whirlwinding, since the latter we cater for in the def already.
> > 
> > 
> > So the condition we're concerned about is strict offhand weapon training. With whirlwind, barbs rarely run this routine, if ever. Perhaps return if weapon skill is offhand? Should be sufficiently rare, while also not broad enough to just skip whenever there's something in our left hand.
> 
> I'm not sure that the condition we're concerned with as such. The condition we want is a free hand, doesn't technically matter which. The problem with "strict offhand weapon training" is purely that in that one case we would sheathe, and reget, but the weapon would then be in the right hand. I am not a fan of recreating a new set of conditions for this, when the use case is so very similar to the almanac one. I would prefer to keep the existing almanac use case (even if we gently tweak it as mentioned) and re-use that for pouch filling.
> 
> I'm leaning towards this
> 
> ```ruby
>   def fill_pouch_with_autolooter(game_state)
>     return unless (Time.now - @autoloot_fill_gem_pouch_timer) > @autoloot_fill_gem_pouch_delay
>     return unless (DRC.left_hand.nil? || game_state.currently_whirlwinding)
> 
>     game_state.sheath_whirlwind_offhand if game_state.currently_whirlwinding
> 
>     DRCI.fill_gem_pouch_with_container(
>       @gem_pouch_adjective,
>       @gem_pouch_noun,
>       @autoloot_container,
>       @full_pouch_container,
>       @spare_gem_pouch_container,
>       @tie_pouch
>     )
> 
>     @autoloot_fill_gem_pouch_timer = Time.now
> 
>     game_state.wield_whirlwind_offhand if game_state.currently_whirlwinding
>   end
> ```
> 
> I would then change clause in `use_almanac` to match this.

If this is only a problem during offhand skill training, it only makes sense to gate it there. Seems like you're trying to find "offhand skill" by redefining it.

MahtraDR:
> If this is only a problem during offhand skill training, it only makes sense to gate it there. Seems like you're trying to find "offhand skill" by redefining it.

Huh? I'm reusing the code from `use_almanac` I'm inverting the `if` into an `unless` for readability.

Raykyn55:
> > If this is only a problem during offhand skill training, it only makes sense to gate it there. Seems like you're trying to find "offhand skill" by redefining it.
> 
> Huh? I'm reusing the code from `use_almanac` I'm inverting the `if` into an `unless` for readability.

Almanac is doing the same thing, trying to define the status of training offhand instead of just using game_state.weapon_skill (offhand). The easy solution is gate it there, the hard solution is remembering which hand was holding what, and reequipping appropriately, the bad solution is trying to describe what offhand looks like and hoping we don't get false positives.

MahtraDR:
> Almanac is doing the same thing, trying to define the status of training offhand instead of just using game_state.weapon_skill (offhand). The easy solution is gate it there, the hard solution is remembering which hand was holding what, and reequipping appropriately, the bad solution is trying to describe what offhand looks like and hoping we don't get false positives.

The easy solution to me is to keep the existing code, which works and is well tested, having been around for a very long time now. I'll do that, I think. I've had it running on mine since yesterday now, and it looks good.

PR #6638, opened 2023-05-16
[scripts][common-items] We just need a free hand. Doesn't matter which.

MahtraDR:
As per title.


Raykyn55:
We're not going to have a free hand if we have a weapon in one and a gem in the other.

MahtraDR:
> We're not going to have a free hand if we have a weapon in one and a gem in the other.

Whilst true... so? :) It just means we don't have a free hand right then. Try again later.

PR #6646, opened 2023-07-02
New cast script

bishoprook:
This is a script I've been developing to allow casting spells with a similar ease to using `,buff` but with much more ability to make overrides on a per-cast basis. Saved defaults and aliases go in `UserVars`, but can be exported in YAML format compatible with waggle sets for those who want to continue using `,buff` for its purposes.

Lots and lots of usage documentation in the comment at the top.

Opening as a PR in case other folks might be interested in using it!

### DIFF
--- a/cast.lic
+++ b/cast.lic
@@ -1,4 +1,4 @@
-#quiet
+# quiet
 =begin
 
   A highly configurable helper to cast spells.
@@ -27,7 +27,7 @@
   `prep_scaling_factor` you have configured). Ex:
   * `,cast heal discern` will discern the Heal spell and then cast at some
     fraction of the discerned max prep.
-  
+
   You can use `-d` or `-debug` to turn on debug logging for this run.
 
   OTHER PARAMETERS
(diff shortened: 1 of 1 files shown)


Hiinky:
I haven't looked at any code, but I am curious if the ;buffother script isn't doing what you need it to do?

bishoprook:
> ~I haven't looked at any code, but I am curious if the ;buffother script isn't doing what you need it to do?~ nevermind, I should have looked first!

To be more explicit on this, these were my design goals:

* Focus just on the task of casting a single spell (rather than maintaining a set of them like `,buff`
* Be equally useful for all spell types including debil and TM in combat
* Allow casting of spells by name or abbreviation without having to pre-configure anything either in YAML or vars
* Allow one or more spell configurations to be saved with a script command (rather than modifying YAML)
* Allow configurations to be overridden on a per-cast basis
* Relax `,buffother`'s constraint that the target has to be the full name of a player who is in the room
* Have similar configurable mana/conc thresholds to `,buff` (but adding fatigue which can also hurt spellcasting) but allow per-cast handling of what happens when you're under them: fail immediately, wait to cast, or try and cast anyway

It's basically a laundry list of the things I don't personally enjoy about `,buff` or `,buffother`, but whether other folks share my preferences is up to them!

MahtraDR:
Thanks for the contribution!

PR #6654, opened 2023-07-24
[scripts][T2] Add startup delay.

MahtraDR:
Delays startup of t2 until drinfomon has settled a bit.

This allows one to auto-autostart T2.

It also allows one to start T2 within a close amount of time after logging in. In my case, being on high latency (300ms+) distances from the server, if I start T2 too soon, the game chokes, specially go2. This resolves the issue by just waiting a bit before executing training blocks. I use 120s as a delay, for ereference, and it works wonderfully.

Note: This delta is from drinfomon starting, since that's the best way I could think of to find a delta from when we logged in (it's also the same method used in learned.lic to track "time since login" and it made sense to me.

### DIFF
--- a/t2.lic
+++ b/t2.lic
@@ -34,7 +34,7 @@ class T2
       end
     end
 
-    unless (Time.now - DRSkill.start_time > @settings.t2_startup_delay.to_i) || @args.nodelay
+    if !@args.nodelay && @settings.t2_startup_delay.to_i > 0
       pause until Time.now - DRSkill.start_time > @settings.t2_startup_delay
     end
   end
--- a/t2.lic
+++ b/t2.lic
@@ -9,7 +9,12 @@ custom_require.call(%w[common common-arcana equipmanager])
 
 class T2
   def initialize
-    arg_definitions = [[]]
+    arg_definitions = [
+      [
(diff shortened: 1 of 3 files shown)


MahtraDR:
@rpherbig Sorry to bother. Would you mind taking a minute to glance at this? Thanks!

MahtraDR:
Note, I've had this running for weeks now. It works perfectly that I can tell, and has been tested by two others (one for almost as long as me.)

t2.lic
```diff
@@ -28,6 +33,10 @@ class T2
         fput("avoid #{avoid['type']}")
       end
     end
+
+    unless (Time.now - DRSkill.start_time > @settings.t2_startup_delay.to_i) || @args.nodelay
```
rpherbig, line 37: 💡 This conditional could probably be simplified to something like `if !@args.nodelay && @settings.t2_startup_delay.to_i > 0` since, worst case, it's already been long enough and the next line gets skipped by _it's_ conditional.

I find this version more readable, but it's not a big deal.

rpherbig (approved):

t2.lic
MahtraDR: Thank you. I always appreciate your time, and help.

PR #6657, opened 2023-07-26
[script][setupaliases] update map to use xnarost

mrhoribu:
Update to use xnarost over narost for `map` alias


Kiriawen:
Coulda swore we already did that.   

Might have been just my local copy though.

PR #6658, opened 2023-07-27
[scripts][safe-room] Add support for new empath spell healing plant

MahtraDR:
Adding support for the new empath spell `Embrace of the Vela'Tohr` creates a plant that one can touch to get healed (the plant is like a wound bank) the empath can then come around later and heal the plant.

This PR adds the plant as a healing option to safe-room. By default, I've made the plant take priority over NPC empaths, and player empaths take priority over the plant. 

I haven't decided if I want it opt-in, or opt-out, or not optional?

### DIFF
--- a/safe-room.lic
+++ b/safe-room.lic
@@ -169,7 +169,7 @@ class SafeRoom
     Flags.reset('moved')
 
     if /.*vela.tohr (\w+)/ =~ DRRoom.room_objs.grep(/vela'tohr (plant|thicket)/).to_s
-      fput("touch #{Regexp.last_match(1).to_s}")
+      fput("touch #{Regexp.last_match(1)}")
     end
 
     if Room.current.id == 8393 # Fang Cove Healer fix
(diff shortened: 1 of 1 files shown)


gmfaux:
If the plant is almost at capacity, and you are very wounded and touch it, I believe there is a chance that in taking your wounds it will reach its capacity and 'shut down', but you will still have wounds left over. In that case, you would have to either touch again if a plant/thicket is present, or proceed to other sources of healing?

MahtraDR:
> If the plant is almost at capacity, and you are very wounded and touch it, I believe there is a chance that in taking your wounds it will reach its capacity and 'shut down', but you will still have wounds left over. In that case, you would have to either touch again if a plant/thicket is present, or proceed to other sources of healing?

As is, this change doesn't assume the plant is be all end all, it just tries to touch, in addition to whatever other source of healing it would normally do. In the scenario you describe, it would then proceed to wait for the autopath/empath to finish healing then.

gmfaux:
@MahtraDR what if the room only has a plant and it dies?

MahtraDR:
> @MahtraDR what if the room only has a plant and it dies?

In that scenario you won't get healed, and the logic flow continues as if there'd never been a plant, or empath in there.

gmfaux:
Thanks for adding this!

PR #6659, opened 2023-08-01
Update combat-trainer.lic to add match for using egg while hidden

Tirost:
Add a match for using the war horn ability of the divine egg when hidden.

The match is for: "You cannot stay hidden while using the egg."


MahtraDR:
Thanks for the submission!

PR #6672, opened 2023-08-23
[scripts][dependency] Track warhorn/egg data in firebase (1 of 2)

MahtraDR:
Allows for submission of warhorn/egg data into firebase, and for us to check if a given room has a warhorn timer. 

In a subsequent PR, I'll submit the changes to combat-trainer that allow it to use this feature set.

To set a timer:
```
;eq $hornwatch.submit_horn_update(1900, Time.now)
```
then to query the timer by room
```
>;eq echo $hornwatch.get_horn_data(1900)                                                                                                                                                                                                      
[exec1: 2023-08-23 20:08:42 +1200] 
```

If there is no timer, we return epoch
```
>;eq echo $hornwatch.get_horn_data(190)                                                                                                                                                                                                       
[exec1: 1970-01-01 12:00:00 +1200]
```

This allows for time comparisons.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -10,7 +10,7 @@ require 'ostruct'
 require 'digest/sha1'
 require 'monitor'
 
-$DEPENDENCY_VERSION = '1.6.1'
+$DEPENDENCY_VERSION = '1.6.2'
 $MIN_RUBY_VERSION = '2.5.5'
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
I will revert this change if it results in Firebase charging us. :)

PR #6674, opened 2023-08-23
[scripts][wand-watcher] Wand-Watcher special handling of astrolabes

MahtraDR:
If you have a group-based astrolabe, you might give the luck effect to someone else with a group based astrolabe.

This makes it so you can tell wand-watcher to treat astrolabes as special, and check luck state before invoking yours, deferring by 60 seconds (each time) until you have no luck.

Note there is no messaging about when luck ends, which is annoying... so we have to `info` before each check.

### DIFF
--- a/wand-watcher.lic
+++ b/wand-watcher.lic
@@ -139,11 +139,23 @@ class WandWatcher
       container = @wand_list[wand]['container']
       cooldown = @wand_list[wand]['cooldown'] || 30
       count = @wand_list[wand]['count'] || 2
+      luck_item = @wand_list[wand]['luck_item'] || false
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
```yaml
wands:  # The adjective and noun of your wand.
  brass astrolabe:
    activation message: !ruby/regexp /You feel the magic in the astrolabe swell up|An ornate brass astrolabe remains inert in your hands/
    activate verb: invoke
    container: spidersilk.bag
    cooldown: 16
    count: 2
    luck_item: true
```

MahtraDR:
I've been running this for five days, works as intended.

PR #6677, opened 2023-08-24
Added unhandled match string for dissect in combat-trainer

OdnaZvezda:
The following messages were previously unhandled when a necromancer would attempt to dissect after performing other rituals on a corpse.

The preservation ritual that has been performed on this corpse prevents a meaningful dissection.
The harvesting ritual performed on this corpse prevents a meaningful dissection.


MahtraDR:
Thanks for the submission!

PR #6685, opened 2023-08-30
[scripts][sew]worn_trashcan_verb typo

urbaj-dr:
worn_trashcan_verb was being read from @setting instead of @settings


MahtraDR:
Thanks for the PR!

PR #6696, opened 2023-09-16
[new script] Taisidon.lic

MahtraDR:
Original by @Hiinky. Uploading with permission.

The event is on now, and Tillmen's repo is down.


MahtraDR:
Changes from the original:

Renamed morada.lic to taisidon.lic (Hopefully this reduces the number of questions I get about which script to use for this event!

Rubocop -A

Removed includes, added prefixes.

PR #6700, opened 2023-09-20
task-forage

urbaj-dr:
task-forage completes foraging tasks for Trading EXP from task givers in Crossing (Mags) and Shard (flower peddler). Similar to Magzon, but with some new features (more convenient locations, option to use BOOST TASK, use collect instead of forage, not sending people to die in storm bulls in search of tea leaves, etc.)


task-forage.lic
```diff
@@ -0,0 +1,621 @@
...
+    end
+    kill_time(2) if determine_task_necessary
+  end
+
+  def use_boosts
+    echo 'use_boosts' if @debug
+    fput('info') # Check to see if a boost is already active
+    pause 0.25
+
+    lines = reget(50).reverse
```
Raykyn55, line 377: Could this just be a case statement rather than a reget? activated, not enough credits, or already boosted?

task-forage.lic
```diff
@@ -0,0 +1,621 @@
...
+  end
+
+  def kill_time(repetitions = 1)
+    echo "kill_time: #{repetitions}" if @debug
+    echo "use_boosts: #{@use_boosts}, and current status is #{@boosted}" if @debug
+    echo "never_wait: #{@never_wait}" if @debug
+    use_boosts if @use_boosts && @boosted.nil?
+
+    # No need to kill time if we're using a boost
+    return if !@boosted.nil? && @boosted
```
Raykyn55, line 408: I believe !boosted.nil? and @ boosted here would be the same condition. if it's nil, @ boosted would be falsey, if not, !@boosted.nil? would be true. so could just be return if @ boosted. Also, checking against nil wouldn't catch a false, right? so if @ boosted is false, the first condition would actually pass, whereas second condition would fail.

task-forage.lic
```diff
@@ -0,0 +1,621 @@
...
+    end
+    echo 'known = true' if @debug
+    true
+  end
+
+  # Check if we have a location for the item
+  def mapped_item?
+    echo 'mapped_item?' if @debug
+    rooms = @foraging_data.find { |item| item['item'] == @item }['rooms']
+    return false if rooms.empty? || rooms == ' '
```
Raykyn55, line 467: Why would this ever return a single space?

task-forage.lic
```diff
@@ -0,0 +1,621 @@
...
+
+    echo 'mapped = true' if @debug
+    true
+  end
+
+  # Check if the item is off of a failure cooldown
+  def ready_item?
+    echo 'ready_item?' if @debug
+    if !UserVars.task_forage['item_failures'][@item].nil? && Time.now < UserVars.task_forage['item_failures'][@item] + 7_200
```
Raykyn55, line 476: Easier to read if you simply check that a value exists eg
```ruby
if UserVars.task_forage['item_failures'][@item] && etc..
```

task-forage.lic
```diff
@@ -0,0 +1,621 @@
...
+
+    echo 'UserVars:' if @debug
+    echo UserVars.task_forage.to_yaml if @debug
+    echo 'forage_settings:' if @debug
+    echo @forage_settings.to_yaml if @debug
+
+    while determine_task_necessary
+      find_giver
+      until get_task
+        # loop until we have an appropriate foraging task
```
Raykyn55, line 168: Would it not make more sense to make get_task method a loop internally? replace false with next and true with break.

task-forage.lic
```diff
@@ -0,0 +1,621 @@
...
+    echo 'old find' if @debug
+    DRC.wait_for_script_to_complete('find', [@task_town, @task_giver])
+    return if check_giver
+
+    DRC.message('Unable to locate the wandering task giver.')
+    safe_exit
+  end
+
+  def in_room?(npc)
+    DRRoom.npcs.any? { |room_npc| room_npc =~ /\b#{npc}\b/i }
```
Raykyn55, line 237: DRRoom.npcs.include?(npc.capitalize) is probably faster, and since you only use it once, probably doesn't need it's own method.

Raykyn55 (commented):
Very neat, coding in fast_find is interesting.

task-forage.lic
urbaj-dr: Do you mean case + DRC.bput?
If so, the reason I didn't go that route is that there's no indication to match on if you don't have an active boost. The boost information (assuming there is any) is the last part that shows up in the INFO list. So if there is no boost, it'll hang until the bput 15 sec timeout

task-forage.lic
urbaj-dr: I pulled that function from find.lic. But yeah, you're right.

urbaj-dr:
> Very neat, coding in fast_find is interesting.

I really appreciate the thorough review. It's incredibly helpful to see someone else's perspective on my code, especially as I'm still fairly naive with Ruby.

task-forage.lic
Raykyn55: Fair. Could just suppress the failed match and drop the timer with the bput, but understand the reasoning. I'm just not a fan of reget as it can cause problems if the room is busy. Tried using it in smoke and it kept clipping the wrong info and catching less than what I wanted.

urbaj-dr: Oh cool, I wasn't really familiar with the bput options, but I agree that's a cleaner way to do it.

task-forage.lic
```diff
@@ -371,29 +368,25 @@ class TaskForage
...
-    fput('info') # Check to see if a boost is already active
-    pause 0.25
 
-    lines = reget(50).reverse
-    lines.each do |line|
-      case line
-      when / +Task Master, expiring /
-        echo 'Already boosted' if @debug
+    # Check to see if a boost is already active
+    case DRC.bput('info', ' +Task Master, expiring ', ' +Encumbrance :', { 'timeout' => 0.5, 'suppress_no_match' => true })
```
Raykyn55, line 373: might bump that timer to 2 or 3 to account for lag. I usually don't go lower than 5 for that reason.

urbaj-dr: Yeah, I don't love forcing a lengthy(ish) pause for the good folks with reliable internet connections.  But in hindsight it's probably the thing to do given that boosts are extremely finite, and double-redeeming one because of a missed match is wasteful. I can live with 2 sec.

urbaj-dr: Oh, while considering this I realized my switch from reget to bput was flawed (bput options need to come before matching patterns, which was obscured by the fact that the Encumbrance always match first, and return a "no boost" condition).  I'll do some better testing and submit.

urbaj-dr: I think most people who decide to use boosts will just set the option to true, burn through whatever backlogged boosts they have, and leave the option enabled to consume additional boosts whenever they get earned again. Assuming that 99% of the time there will be be no boosts active, that's a 1-5 second waste each time the script is run (depending on timeout). That might end up being less efficient than just never using boosts at all.

So, a lot of rambling for a pretty minor decision, but I prefer the super-short timeout (0.5 sec) to minimize that, recognizing that lag could cause a false negative on the check. If that does happen, no boost will be activated that particular run, which is not a big deal IMO.

PR #6707, opened 2023-09-27
[scripts][combat-trainer] Include an optional health check for parasites

Nyt3:
Added a ct-parasite flag to look for parasites (line 1059). Currently only looks for blood mites but can be expanded to include other parasites.

If the flag is set, will call the tendme script to tend the parasite (lines 1084 & 1152-1157).

User will have to add a buff_nonspells health command to their yaml to enable this function:
buff_nonspells:
  health: 300

Without a manual health check, this change will be transparent to users.


MahtraDR:
Looks good. Thanks! Which creatures lead to parasites? We could add a note to base-hunting as well?

PR #6708, opened 2023-09-27
[scripts][forge] Properly handle reinforcing armor

Nyt3:
Added string matches to properly recognize and handle armor reinforcing.


MahtraDR:
Thank you!

PR #6711, opened 2023-10-06
[scripts][inventory-manager]search case insensitive pattern

desertkaz:
Currently you are unable to use an item with capitalization to search as args.item comes through lowercase.  by downcasing the yaml value we can search the provided item pattern for keywords that are capitalized.


MahtraDR:
Thanks for the submission! Good change.

PR #6712, opened 2023-10-09
Update base-hunting.yaml to match Elanthipedia

urbaj-dr:
Per the discussion in the Discord, this updated version of base-hunting has skill ranges scraped from Elanthipedia. Some hunting spots were also re-ordered to match their new positions relative to other hunting zones. The update is effectively semantically identical to the previous version, with the only changes being zone ordering and comments (per https://www.yamldiff.com/)

### DIFF
--- a/data/base-hunting.yaml
+++ b/data/base-hunting.yaml
@@ -74,7 +74,7 @@ escort_zones:
     base: 7958
     area: wilds
     enter: leucro2
-  # https://elanthipedia.play.net/Seordhevor_kartais                     180-250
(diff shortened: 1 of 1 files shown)


MahtraDR:
Hi mate, I really appreciate the effort that went into this, but it's hard to review as is, and... elanthiapedia can (and does) change on a whim for folks. If we really want it to match, I'd prefer we just remove the skill ranges from this file altogether, and leave just the links for people to follow and make their own choices.

urbaj-dr:
Fine by me - totally get it your perspective. I have no strong feelings one way or another.

PR #6713, opened 2023-10-09
[scripts][Athletics] Uservars.athletics.to_i 

MahtraDR:
That `nil` for new characters is a problem. This resolves the issue.


MahtraDR:
New characters now should correctly get sent to Arthe instead of getting stuck by madmen.

PR #6717, opened 2023-10-14
[base-shrines] Fixes and p5 shrines

Kiriawen:
Fixed some of the `push:` nouns and added a `Hibarnhvidar:` section for P5.  It's incomplete still but adding what I've got so far.


Kiriawen:
We don't have to wait on this to be completed to merge it.  I'll add additional PR's as I go through them or if someone from discord provides the info requested.

PR #6721, opened 2023-11-03
Updates to sigilharvest and common-items to fix some logic errors in the former and add compatibility to the latter

Nyctus:
common-items: added another match string to make count_item_parts compatible with blank scrolls.

sigilharvest: fixed some logic errors that were preventing the script from buying scrolls when it otherwise should have been

NOTE: The changes to sigilharvest are dependent upon the match string added to common-items.


Nyctus:
Thank you Rubocop for making my code less of a nightmare. :) All the issues should now be fixed!

PR #6731, opened 2023-11-13
Error reporting bug

cbisazza:
Change error when material not found to indicate that the material was not found


MahtraDR:
Help me out here. Can you please show what this change does?

Maybe show script output before and after?

Thanks!

cbisazza:
The original one references an undefined variable called 'name' ... but I'll get you some output if you need that :-)

PR #6750, opened 2023-12-29
Update shape.lic to support worn trashcan

tonyguidice:
Adding support for worn trashcans, copied from sew.lic which already has support


MahtraDR:
Thanks for the submission!

PR #6757, opened 2024-02-06
[scripts][pick] Handle khri safe dodged failures

urbaj-dr:
The old logic assumes that dodging a blown trap with Khri Safe fully disarms the box, and that it's ok to proceed with picking. This doesn't seem to be the case - see the attached log: [khri_safe_dodge.txt](https://github.com/rpherbig/dr-scripts/files/14184625/khri_safe_dodge.txt)
From line 22 onward, all traps remain after dodging the failed disarm.

### DIFF
--- a/pick.lic
+++ b/pick.lic
@@ -540,11 +540,8 @@ class Pick
     when *@disarmed_trap_messages.values
       echo "Identified already disarmed trap: #{disarm_result}" if @debug
       trapped = false
-    when *@disarm_retry
+    when *@disarm_retry, 'Thanks to an instinct provided by your sense of security'
       echo 'Failed to disarm trap' if @debug
-      trapped = true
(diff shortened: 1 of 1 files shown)


MahtraDR:
I don't like this change because it renders those two `when` cases pointless. Might as well move them into another case which already does this.

pick.lic
```diff
@@ -461,8 +461,8 @@ class Pick
 
     case disarm_identify_match
     when 'Thanks to an instinct provided by your sense of security'
-      echo 'Dodged a trap with Safe...'
-      trapped = false
+      echo 'Dodged a trap with Safe...' if @debug
```
MahtraDR, line 464:
```suggestion
      echo 'Detected a trap with Safe...' if @debug
```

MahtraDR:
Merging in now, I'll own fixing if it screws something up. :)

PR #6758, opened 2024-02-09
[scripts][astrology] Add spell books

Algonem:
Adds to Perception for targets


vtcifer:
`perc mana` already covers all of the individual books themselves, and they operate on a shared timer of the individual books.

PR #6780, opened 2024-03-03
[scripts][go2] Re-supporting the ability to use spaces in custom go2 locations

KevenDenen:
An  earlier commit removed the ability to use spaces as part of custom locations. This adds back in that ability.

### DIFF
--- a/go2.lic
+++ b/go2.lic
@@ -14,7 +14,7 @@
 
    changelog:
       1.36 (2024-03-03)
-        Re-supporting the ability to use spaces in custom go2 locations. 
+        Re-supporting the ability to use spaces in custom go2 locations.
       1.35 (2023-02-04)
         Add support for custom named targets with an array of roomnumbers. Finds nearest.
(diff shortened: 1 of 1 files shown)


urbaj-dr (changes requested):
I think that this change to the regex could allow for accidental duplicated custom targets, based on how go2 handles whitespace.
```
go2 save test=1901
go2 save test = 1900
```
These would get saved as "test" and "test ", and only "go2 test" [the target without a trailing space] would be navigable.

It's not a big problem, but stripping the whitespace on [line 360](https://github.com/rpherbig/dr-scripts/pull/6780/files#diff-08b3a5f922bb2f71db0dc4b08d9e7929dbd053ba6289ab73b361f3f28b569ec6R360) should solve it 
```
   target_name = $1.strip
```

KevenDenen:
Added in a strip per @urbaj-dr. Also, removed the extra space from my history comment to appease rubocop.

PR #6787, opened 2024-03-26
[scripts][sanowret-crystal] Add EXHALE option.

Nyt3:
Added an optional argument to "exhale" the sanowret crystal. While the learning is slower, there's less conflicts/backfiring risks when using the exhale option than using the standard gaze option (the concentration hit is immediate, which slows down other scripts like buff.lic which have concentration checks). Without the argument the script will operate as usual.


MahtraDR:
This one looks good too. Thank you!

PR #6788, opened 2024-03-26
fixed excessive call to get_settings in moonwatch

skyleronken:
On very rare occasions lich will hang due to excessive calls to get_settings within moonwatch (every .1 seconds). The issue is probably related to filesystem limitations on WSL, but this call is excessive regardless.


MahtraDR:
Hey mate, I see you have made several commits here, and I'm sorry for the hassle this caused you... I am closing this PR with a comment, and won't merge it as is, because I want to still have that call to get_settings, although I agree it shouldn't be in the loop. I've moved it out of the loop in https://github.com/rpherbig/dr-scripts/pull/6794 instead.

PR #6789, opened 2024-03-27
[scripts][astrology] refactor for better logic.

Algonem:
The astrology script will repeat until you reach 32/34. 
Adds checks to the buff for spells not casted over and over again.
Adds check to skill 32/34 on Astrology
Stops sanowret crystal which just interfered at times with the casting of buffs. (can be removed as a custom script)


MahtraDR:
Thanks for this. Looks good.

PR #6790, opened 2024-03-29
[script][dusk-labyrinth.lic] fix map reset behavior

desertkaz:
Instead of hard failure, handle map reset to continue with fresh init
image


MahtraDR:
Thank you for the submission! I appreciate it.

PR #6812, opened 2024-05-04
[data][base-spells] Adding expire message to BTN and changed SCO to lunar from arcane

hat071af:


MahtraDR:
Thank you for the submission!

PR #6817, opened 2024-05-15
[script] sort-scrolls - some major updates

BinuDR:
I've changed from using loops with `line = get` to `Lich::Util::issue_command()` as it is more efficient and reduces the complexity of the code.

I changed from using bputs to get the cases and scrolls out to `DRCI.get_item_safe?()`.

I added the abbreviation of spells to the big array we build at the start of the script and also added a check for when you can't turn to a full spell name.  Now it will try to turn to the abbreviation if the full name fails.

I updated from using `inv container` and trying to capture any line that has a scroll in it to now using `DRCI.get_scroll_list_in_container(@default_container)` which uses the `rummage /SC` option to find only scrolls in that container.

The yaml settings should no longer need the `scroll_nouns` section, but I've not made any changes to that part yet.

I think that's all.

### DIFF
--- a/sort-scrolls.lic
+++ b/sort-scrolls.lic
@@ -79,7 +79,7 @@ class ScrollSorter
   end
 
   def find_abbrev(spell)
-    return @all_spells.select{ |s, _g| s.casecmp(spell) == 0 }.flatten[2]
+    return @all_spells.select { |s, _g| s.casecmp(spell) == 0 }.flatten[2]
   end
 
   def check_scroll(scroll)
--- a/sort-scrolls.lic
+++ b/sort-scrolls.lic
@@ -11,7 +11,7 @@ class ScrollSorter
       next if v['skill'] == 'cantrip'
 
       v['guild'] = 'Analogous' if v['mana_type'] == 'ap'
-      @all_spells.push([k, v['guild']])
+      @all_spells.push([k, v['guild'], v['abbrev']])
     end
     @settings = get_settings
     @default_container = @settings.default_container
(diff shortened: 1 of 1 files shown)


BinuDR:
This new version will require whichever version of lich where the `Lich::Util::issue_command()` method was added.
I'm not sure which version that is.

MahtraDR:
> This new version will require whichever version of lich where the `Lich::Util::issue_command()` method was added.
> 
> I'm not sure which version that is.

@mrhoribu do you know the answer to this?

Note that as is, dependency enforces Ruby 3.2.2 which came with Lich 5.7.0. We can start enforcing that lich version as well as a floor.

BinuDR:
> > > This new version will require whichever version of lich where the `Lich::Util::issue_command()` method was added.
> > > I'm not sure which version that is.
> > 
> > 
> > @mrhoribu do you know the answer to this?
> 
> It was added in Lich 5.7.0. So should be safe there.
> 
> One question is do you want to anchor the command on a new status prompt or on a static txt output of the command? I see that the previous used `DRC.bput("inv #{@stacker_container}", /^Inside.+you see:/)` and then stopped at `INVENTORY HELP`. If you wanted to do this with issue command non-xml you would do `capture = Lich::Util.issue_command("inv #{@stacker_container}", /^Inside.+you see:/, /INVENTORY HELP/, usexml: false)`
> 
> As for the second usage of it with the stacker, I don't see an end command there and was more of a timeout option. So that one, unless there's always a static end msging when using the command, would be better left with xml and using the status prompt as an end pattern.

I think using 5.7.0 is a good floor version to enforce.

For the first usage change, I did that one more because the code is cleaner.  Ending on status prompt with that one is fine and is equivalent to ending on `INVENTORY HELP` since that's just the last line of the output from the command.

Yeah, the second usage is the one that was a the main reason I switched to issue_command because there is no output I can break on otherwise.  That's why I used the timeout option.  Using the statusprompt there makes this much easier to code.

MahtraDR:
@BinuDR Is this good to go?

BinuDR:
> @BinuDR Is this good to go?

Yeah, I'll just need to announce that if anyone who uses the script is on any lich before 5.7.0 that it won't work for them.

PR #6820, opened 2024-05-19
Update wand-watcher.lic activation_failure_messages

urbaj-dr:
Expanded the ".* remains inert" @activation_failure_messages to account for astrolabes


MahtraDR:
Thank you!

PR #6824, opened 2024-05-23
[scripts][learned] Display skills with zero exp at the bottom

mikejcook:
Added a new 'zero' argument to the learned script that I've found useful primarily when working with newer characters to ensure that all of the skills are moving. If set, it will find only skills that have not had any experience gains in the session and list them.

Example output:

--- Lich: learned active.
    Enchanting     0.00 (0.00/hr, 0.00/day)     Alchemy       0.00 (0.00/hr, 0.00/day)
    Outfitting     0.00 (0.00/hr, 0.00/day)     Forging       0.00 (0.00/hr, 0.00/day)
    First Aid      0.00 (0.00/hr, 0.00/day)
--- Lich: learned has exited.

### DIFF
--- a/learned.lic
+++ b/learned.lic
@@ -22,10 +22,12 @@ class LearnedData
       [{ name: 'lore', regex: /lore/, description: 'Show exp gained in lore skills.' }],
       [{ name: 'guild', regex: /guild/, description: 'Show exp gained in your guild skill.' }],
       [{ name: 'reset', regex: /reset/, description: 'Reset learned experience' }],
+      [{ name: 'zero', regex: /zero/, description: 'Show only skills with nothing learned' }],
       []
     ]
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
Why not just make learned show all skills instead of zero and greater than zero?

mikejcook:
Are you thinking by default, or when called with an argument? As for the overall change, functionally it gets to the same place, so as long as people are cool with the additional skills being listed, it works for me.

MahtraDR:
I do mean change the default. Test it out. It's either remove that one select, or change the `>` to a `>=`

mikejcook:
@MahtraDR Made the change we discussed; just removed the select that was filtering the skills. Have been running this for a day or so, no issues.

Issue #6290, opened 2022-12-15
workorders - new enchant functionality problem with 'rods'


mikejcook:
I ran into an issue where the new functionality got stuck on a rod. Recipe was for 2 a cambrinth retuners. As you can see below, because the imbuement rod (enchanting tool) was in the backpack, it only bought 1 bobcat rod, so the script eventually hung. @Dartellum 

```
[workorders]>rummage /C rod in my backpack
You rummage through a rugged backpack looking for something similar to "rod" and see an imbuement rod.
<pushBold/>need_part with potiental plus 1 is 1.<popBold/>
[workorders]>buy bobcat rod
You decide to purchase the rod, and pay the sales clerk 312 Kronars.
The sales clerk hands you your bobcat rod.
[workorders]>put my bobcat rod in my backpack
You put your rod in your backpack.
[workorders]>tap my fount
```

MahtraDR:
This is expected I think. You need to keep your tools in a different container.

mikejcook:
Negative. Clerk-tools and workorders put your tools in crafting_container, which is also what the new code checks.

problematic code:
```
@bag = @settings.crafting_container

    # Parts #
    tmp_part_count = 0
    need_part = 0
    if recipe['part']
      for p in recipe['part']
        p.to_s 
        tmp_part_count = DRCI.count_items_in_container(p.split.last, @bag)

```

clerk-tools.lic

```
@bag = settings.crafting_container

  def get_tools(tools, roomnumber, repairnpc)
    DRCT.walk_to(roomnumber)
    tools.each do |tool|
      case DRC.bput("ask #{repairnpc} for #{tool}", 'Ah, yes, we have one of your tools like that', 'Ah, yes, we have several of your tools like that', "It doesn't look like we have anything like that")
      when 'Ah, yes, we have one of your tools like that'
        DRCC.stow_crafting_item(tool, @bag, @belt)
      when 'Ah, yes, we have several of your tools like that'
        # #Sack
      end
    end
  end
```

mikejcook:
I don't know what the solution would be. Adding a new parts bag? But that sounds terrible.

Honest question: what problem is this change trying to solve? The sigils make sense, since you can farm your own, but can you get your own parts? I'm not aware of a reason you would have existing parts in your bag already.

Dartellum:
[...]

PR #6825, opened 2024-05-23
[scripts][workorders][data][base-recipes] fix for an issue with rod parts

mikejcook:
Corrected an issue I (and others) have encountered since 2022 when logic to use existing materials for enchanting was added. After digging into it, the biggest source of the issue is that it was passing only the noun into DRCI.count_items_in_container, which prevented that method from checking for the full item name as it should. Additionally, rods were listed with only their partial name as parts in base-recipes. Finally, rods are listed with only their noun in base-crafting, but the script was passing the full name here.

I've tested with both a low and high level enchanter.

This fixes issue #6290

### DIFF
--- a/data/base-recipes.yaml
+++ b/data/base-recipes.yaml
@@ -8893,7 +8893,7 @@ crafting_recipes:
   enchant_stock4:
   item:
   part:
-  - bobcat rod
+  - basalt bobcat rod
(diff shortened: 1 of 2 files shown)


MahtraDR:
Thanks for this!

PR #6829, opened 2024-06-02
[script][remedy] Added stamp and cube functionality

mikejcook:
Adding functionality to stamp crafted remedies from a request in Discord. Also noticed that the script doesn't have the HE cube functionality, so I added that as well.


MahtraDR:
Thank you!

PR #6830, opened 2024-06-03
Update base-spells.yaml

hat071af:
Added catch for case when using tattoo_tm: to target a tm spell after invoking it from a tattoo fails due to lack of target.

### DIFF
--- a/data/base-spells.yaml
+++ b/data/base-spells.yaml
@@ -142,6 +142,8 @@ prep_messages:
 - Please don't do that here
 - You cannot use the tattoo while maintaining the effort to stay hidden
 - As you attempt to prepare the spell, a sense of overwhelming peace washes over you
+# You tried to target a spell but there is nothing left to face
+- There is nothing else to face
 
 cast_messages:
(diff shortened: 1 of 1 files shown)


MahtraDR:
Why the commented line?

hat071af:
I just copied the same bit from the cast_messages: section below to be consistent with the documentation there. 
`# You tried to cast an offensive spell and the target no longer exists
- There is nothing else to face
# If don't know the cast command, like 'barrage'
- Please rephrase that command`

MahtraDR:
Ah so the line being added to the list is a snippet from the comment? Sorry that wasn't clear. If this is the case, please match case.

hat071af:
Uh, no? I'm not sure where the communication error is happening lol. The functional line being added is the in game output amd matches case. You asked why there was a commented line, and that's just for documentation and follows the convention of other documentation comments that already exist in the file.

MahtraDR:
All good. I'll merge this and we fix whatever we have to later if we have to. 😀

PR #6835, opened 2024-06-09
[script][inventory-manager] add support for vault standard

mikejcook:
Adding the arg vault_standard, which calls the command VAULT STANDARD to list vault inventory. Removes support for vault books, which instead will call this command.


MahtraDR:
Thank you!

PR #6836, opened 2024-06-10
[scripts][hunting-buddy] Add combat_trainer_gear_set setting

hat071af:
Added support for `combat_trainer_gear_set:` for use with combat-trainer.lic.  

When hunting-buddy.lic returns to your safe room from a combat-trainer.lic session that uses the `combat_trainer_gear_set:` gearset override, it now keeps you in that gearset rather than trying to swap you to the standard gearset. There is no behavior change if not using `combat_trainer_gear_set:`


MahtraDR:
Thank you!

PR #6848, opened 2024-06-22
Update base-spells.yaml - added anti-magic room checks for invoke/charge/perc

hat071af:
Added matches for when invoking or charging cambrinth, perc, or invoking a tattoo fails due to wild magic anti-magic room


MahtraDR:
I'm happy to merge these, but I do also have a worry because these temporary changes to room magic mechanics will impact other scripts. This isn't inherently relevant to this PR though. Wand watcher and buff watcher for example will need to account for temporary anti-magic rooms.

PR #6849, opened 2024-06-22
Update task-forage.lic

urbaj-dr:
Better handling for all items when the request includes the phrase "bits of" - eghmok moss, green moss, and possibly others were not being read correctly.


MahtraDR:
Nice update. Thanks!

PR #6853, opened 2024-06-26
Update combat-trainer.lic to fix tessera scoping

urbaj-dr:
The previous tessera change used a bad variable scope, trying to access `settings` from invest_in_tessera, rather than defining the variable in TrainerProcess initialization


MahtraDR:
Sorry mate, I fixed this about 10 minutes ago, without noticing this.

PR #6858, opened 2024-07-06
[script][base-spells]move php recast message to cast_messages

desertkaz:
Moved recast message for php to `cast_messages`
image
image


Kiriawen (changes requested):
Instead of adding line 265 why not adjust line 264 and just remove everything after the `,` so that the one line matches multiple messages

PR #6859, opened 2024-07-08
[script][inventory-manager]add vault book,storage book,pocket. adjust vault surface handling for container nesting data

desertkaz:
**inventory-manager.lic:**
- Added storage book logic for new system
  - clean output for storage box like a vault book
- Added vault book logic for new system
  - vault standard fails if rent is past due
- Added logic for rummaging secret pocket containers
- Added logic to retry the `Lich::Util.issue_command` when timing out
- Update eddy method call
- Handle empty register and storage book contents

**base.yaml:**
- moved vault surfaces to separate key to handle tagging nested items
- added pattern for storage book ignores

### DIFF
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -16,9 +16,10 @@ class InventoryManager
       ],
       [
         { name: 'save', regex: /save/i, description: "Save new or update current character's items to the database." },
-        { name: 'vault_book', regex: /vault_book/i, optional: true, description: "DEPRECATED. Will call vault_standard." },
+        { name: 'vault_book', regex: /vault_book/i, optional: true, description: "Uses vault book for inventory." },
         { name: 'vault_regular', regex: /vault_regular/i, optional: true, description: "Rummages your vault to save items NOTE: must be standing by your open vault." },
         { name: 'storage_box', regex: /storage_box/i, optional: true, description: "Rummages caravan storage box for items." },
+        { name: 'storage_book', regex: /storage_book/i, optional: true, description: "Uses storage book for caravan items instead of rummaging." },
         { name: 'family_vault', regex: /family_vault/i, optional: true, description: "Rummages your family vault to save items NOTE: must be standing by your open vault." },
         { name: 'register', regex: /register/i, optional: true, description: "Reads contents of deed register." },
         { name: 'eddy', regex: /eddy/i, optional: true, description: "Save the contents of an eddy (HE 436 Gift)." },
@@ -26,6 +27,7 @@ class InventoryManager
         { name: 'home', regex: /home/i, optional: true, description: "Save home inventory." },
         { name: 'servant', regex: /servant/i, optional: true, description: "Save shadow servant inventory." },
         { name: 'shop', regex: /shop/i, optional: true, description: "Save your Trader shop inventory." },
+        { name: 'pocket', regex: /pocket/i, optional: true, description: "Rummages your secret pocket container to save items." },
         { name: 'vault_standard', regex: /vault_standard/i, optional: true, description: "Uses VAULT STANDARD to get a list of items." }
       ],
       [
@@ -45,12 +47,11 @@ class InventoryManager
     args = parse_args(arg_definitions)
     @settings = get_settings
     @inventory_ignores = @settings.inventory_manager_ignores
-
+    @vault_surfaces = @settings.inventory_manager_vault_surfaces
     setup
 
     if args.vault_book
-      DRC.message('Vault Book support has been removed in favor of VAULT STANDARD.')
-      add_vault_standard
+      add_vault_book
     elsif args.vault_standard
       add_vault_standard
     elsif args.vault_regular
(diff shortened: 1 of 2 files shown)


desertkaz:
output screenshots

image
image
image

inventory-manager.lic
```diff
@@ -223,6 +240,20 @@ class InventoryManager
         else
           container = item.to_s
         end
+      elsif /vault book/.match(command)
```
mikejcook, line 243: Line 257 uses command == 'vault standard', the two new ones use /vault book/.match(command). I don't know this for sure, but .match does a regular expression comparison, which I suspect is more labor intensive than just doing a string compare. It's probably fairly insignificant, but I would think we're better off with consistency, in any event.

mikejcook: One other thought, you should probably combine the vault book and storage book, since the code itself is the same.

mikejcook:
I like the change to retry the command. I've noticed once or twice that if I get stuck in RT, it'll end up putting empty data in until run again. I wish Lich::Util.issue_command had built in support for handling RT. I wonder if it makes sense to create a helper method in common.lic that calls Lich::Util.issue_command similar to DRC.bput. I feel like there is a risk that it could get stuck in a never ending loop (such as when checking the eddy inventory, if you don't have an eddy), but I'm not sure that really needs to be handled. I guess it could do a tap portal before issuing the command.

And good catch on the method name for the eddy. No idea how I didn't notice that.

For the vault surfaces, it looks like if something is in a container on a surface, the surface will get overwritten. I started looking at this, and I realized there was a bug in the existing code with nested containers.

We could do something like this to address the issue:

Add below container = "":
```
sub_container = ""
```

```
      elsif /vault book/.match(command)
        # items in vault book output are prefixed by 6 spaces, nested containers are 8 spaces
        if line =~ /\s{8}(?:a|an|some) /
          item.concat(" (nested container - #{sub_container})")
        elsif line =~ /\s{6}(?:a|an|some) /
          item.concat(" (in container - #{container})")
          sub_container = item.to_s
        else
          container = item.to_s
        end
```

```
      elsif command == 'vault standard'
        # items in a container are prefixed by 11 spaces once the item count is removed, nested containers are 16
        if line =~ /\s{16}(?:a|an|some) /
          item.concat(" (nested container - #{sub_container})")
        elsif line =~ /\s{11}(?:a|an|some) /
          item.concat(" (in container - #{container})")
          sub_container = item.to_s
        else
          container = item.to_s
        end
```

They will then save like this:

```
- a canvas backpack (in container - a bottom drawer) (vault)
- an infuser stone (nested container - a canvas backpack (in container - a bottom drawer)) (vault)
- an infuser stone (nested container - a canvas backpack (in container - a bottom drawer)) (vault)
- an infuser stone (nested container - a canvas backpack (in container - a bottom drawer)) (vault)
```

inventory-manager.lic
desertkaz: My initial thought was just being against"read my vault book" and "read my storage book" as the matches and typically just use regex.  Can change to whatever.

I was following the example of vault and family vault being unique methods when I made the updates.

mikejcook: Oh, not the methods, I just mean these two elsif statements. But just my two cents!

```
      elsif /vault book/.match(command)
        # items in vault book output are prefixed by 6 spaces
        if line =~ /\s{6}(?:a|an|some) /
          item.concat(" (in container - #{container})")
        else
          container = item.to_s
        end
      elsif /storage book/.match(command)
        # items in a storage book are prefixed by 6 spaces
        if line =~ /\s{6}(?:a|an|some) /
          item.concat(" (in container - #{container})")
        else
          container = item.to_s
        end
```

desertkaz:
I wasn't sure if it was my network or what on the Lich::Util call but yeah, sometimes I was seeing up to 10 timeouts before a success.  This was the easiest fix to ensure actual data gets written.  My last update half my characters had empty data which is what prompted these hotfixes and additions.

desertkaz:
combined into single block for the main loop.  updated to use double string evaluation vs regex.

desertkaz:
image

desertkaz:
Output showing nested matching
image

desertkaz:
Output showing nesting in character, vault and storage_box.

image

MahtraDR:
@mrhoribu worth looking into these timeouts separately.

desertkaz:
```
- the head and pelt of an albino wolf with vicious fangs and claws (worn) (character)
- the horned head and hide of a storm bull (closed) (character)
- several serrated loimic horns (worn) (character)
- clusters of vibrant violets (character)
- one half of a tiny amethyst (vault)
- the head and pelt of a tawny gold gryphon (character)
```
`- a motley cloak (in container - the horned head and hide of a storm bull (closed)) (character)` brought to my attention the fact there are some outliers that will cause weird issues nested and the using the match of `a|an|some`.

image
Also combined shadow servant with vault book/storage book since all the output formatting is the same.

Last addition for now until we get the rest of this worked out and merged.

PR #6860, opened 2024-07-10
[script][afk]attempt to resolve justice messaging by releasing any triggers_justice spells active

desertkaz:
image

### DIFF
--- a/afk.lic
+++ b/afk.lic
@@ -53,6 +53,7 @@ loop do
 
   if justice_message_count == 1
     justice_message_count = 0 if DRSpells.active_spells.each { |x,y| fput("release #{get_data("spells")["spell_data"][x]["abbrev"]} spell") if get_data("spells")["spell_data"][x]["triggers_justice"] }
+  end
   if justice_message_count > justice_threshold
     exit_game("It looks like you've run into trouble with the law too many times")
   end
--- a/afk.lic
+++ b/afk.lic
@@ -51,6 +51,10 @@ loop do
   justice_message_count += 1 if line =~ /^"Burn .+!  Burn .+!" .* giving you a wide berth/
   justice_message_count += 1 if line =~ /authorities will try to bring you in for endangering the public/
 
+  if justice_message_count > 0 && /1000/.match(DRSpells.active_spells.to_s)
+    fput("release cyclic") 
+    justice_message_count = 0
+  end
   if justice_message_count > justice_threshold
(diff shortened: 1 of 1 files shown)


afk.lic
```diff
@@ -51,6 +51,9 @@ loop do
   justice_message_count += 1 if line =~ /^"Burn .+!  Burn .+!" .* giving you a wide berth/
   justice_message_count += 1 if line =~ /authorities will try to bring you in for endangering the public/
 
+  if justice_message_count >= 1
+    justice_message_count = 0 if DRSpells.active_spells.each { |x, _y| fput("release #{get_data("spells")["spell_data"][x]["abbrev"]} spell") if get_data("spells")["spell_data"][x]["triggers_justice"] }
```
MahtraDR, line 55:
```suggestion
    fput("release #{get_data("spells")["spell_data"][x]["abbrev"]} spell") if get_data("spells")["spell_data"][x]["triggers_justice"] }
```

MahtraDR: This is still not ideal. What we want to do here is check if any of our active spells trigger justice. If any of them trigger justice, then release all the spells that trigger justice. We don't reset the counter here, because we rely on the counter for the def above, which exits the game if we reach justice threshold.

desertkaz: Without resetting the counter for this "false positive", we would end up checking spells every loop from that moment until script closure.  The goal was just to reset things if a spell was actually released to prevent that.

MahtraDR: Ok, so my concern is that we need `justice_message_count` to stay set so it gets checked properly against `justice_threshold` 

I suppose yeah, what we need here is an additional flag then. So we only end up back into this new def if there was a change.  So we see justice message count go over 1. We entered this def, we released all the offending spells. We set a flag we've done this for this justice_message_count value. We should then only re-enter this def if the justice_message_count changes, otherwise we'd be looping into it over and over.

The reason I'm not comfortable fully resetting justice_message_count is that there are other reasons/ways people might be getting tagged for justice, and if we reset the count, those might get missed.

desertkaz: yeah I'm totally ok with that. and solved my initial reason for it in latest.

desertkaz: or do we `spell_check_count = justice_message_count` instead of incrementing a counter. they shouldn't get out of sync, but why risk it?

MahtraDR:
Thanks for this!

PR #6862, opened 2024-07-12
[script][inventory-manager]make nested container information an option vs default for output for search and list

desertkaz:
**LIST**
image
image

**SEARCH**
image


mikejcook:
@desertkaz This was already merged, but I'm not sure what the purpose of this change is? What's bad about seeing the container of the item?

PR #6863, opened 2024-07-13
Update textsubs.lic -  regex fix for large plat values

hat071af:
When `textsubs_use_plat_grouping: true` is set the output of currency amounts >=1000 plat was not displaying correctly. Namely, it continued digit groupings in sets of 4 which should only apply to the first plat grouping. Fixed the regex to properly sub in such cases.

Before:
 You are certain that the icesteel mask is worth exactly 1156,2500 Kronars.
 You think it is likely that the icesteel hauberk is worth around 1,8749,9987 Kronars.

After:
 You think it is likely that the icesteel mask is worth exactly 1,156,2500 Kronars.
 You estimate that the icesteel hauberk is worth about 18,749,9993 Kronars.

### DIFF
--- a/textsubs.lic
+++ b/textsubs.lic
@@ -823,7 +823,7 @@ end
 
 # Commas when dealing with coins
 if settings.textsubs_use_plat_grouping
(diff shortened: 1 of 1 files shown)


MahtraDR:
Thanks for the submission!

PR #6864, opened 2024-07-13
alchemy.lic and combat-trainer.lic updates

dantiadr:


vtcifer:
Hi @dantiadr 

A couple of comments:

1\ It looks like the editor you're using uses tab indentation instead of spaces.  Ruby style guide is to utilize two spaces instead of the tab character.  Can you please change the setting to use 2 spaces instead of tabs?  If you need assistance with that, we can help you in the discord.

2\ We try to limit PRs to related changes.  Could you please submit two different PRs for these changes as they don't seem related (although it is difficult to tell because of Item 1 above.)

Thanks!

dantiadr:
Thanks! That is crazy helpful feedback. I couldn't figure out why it was showing the entire file changed when I really hadn't!

Good to know about separate PRs, will do. Would it make sense for me to remove this PR and open two individuals when I correct the spacing issue, or continue to roll with this single one?

PR #6867, opened 2024-07-16
[scripts][combat-trainer] Add successful dissect message

dantiadr:
Add additional successful dissect message seen when dissecting telga orek.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -956,8 +956,9 @@ class LootProcess
                   /You learn something/i,
                   'A failed or completed ritual has rendered',
                   'You realize after a few seconds',
-                  'prevents a meaningful dissection')
-    when /You succeed in dissecting the corpse/, /You learn something/i
+                  'prevents a meaningful dissection',
+                  "With less concern than you'd give a fresh corpse")
+    when /You succeed in dissecting the corpse/, /You learn something/i, /With less concern than you'd give a fresh corpse/
       return true
     when /You'll gain no insights from this attempt/
       waitrt?
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -958,7 +958,7 @@ class LootProcess
                   'You realize after a few seconds',
                   'prevents a meaningful dissection',
                   "With less concern than you'd give a fresh corpse")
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -956,8 +956,9 @@ class LootProcess
                   /You learn something/i,
                   'A failed or completed ritual has rendered',
                   'You realize after a few seconds',
-                  'prevents a meaningful dissection')
-    when /You succeed in dissecting the corpse/, /You learn something/i
+                  'prevents a meaningful dissection',
+                  "With less concern than you'd give a fresh corpse")
+    when /You succeed in dissecting the corpse/, /You learn something/i, /With less concern than you'd give a fresh corpse/
```
MahtraDR, line 961:
```suggestion
                  'With less concern than you'd give a fresh corpse')
    when /You succeed in dissecting the corpse/, /You learn something/i, 'With less concern than you'd give a fresh corpse'
```

MahtraDR: I've done two things here.

1. I've moved your double quotes to single quotes. This is merely style in this case. Double quotes would enable string expansion, which we don't use here, so in this case they'd be interchangeable, but I moved to single for style purposes, since all the other non-regex strings use single quotes.

2. Note that I moved your slashes to single quotes, to match the string. If we'd kept the double quotes, we could have switched the slashes to double quotes to match too. The slashes are for regexes. Here it would also not have mattered, note, but this keeps the code consistent.

dantiadr: Thanks @MahtraDR, one question:

1. For the double vs single quotes. I went with double since the string has an apostrophe in it (following the example of one of the previous lines of `"You'll learn nothing",`). Will the apostrophe in the newly added line create an issue with using single quotes?
2. Got it!

vtcifer: Single quotes need to escape the single quote in the string. My preference there is double quotes to not need the escape

MahtraDR: @dantiadr That is spot on by you. My oversight on the apostrophe. Double quotes for both then! :)

MahtraDR:
Thank you!

PR #6868, opened 2024-07-16
[scripts][attunement] Add magic disabled messaging

dantiadr:
With wild magic, a specific room might have magic disabled temporarily. Adding this message will allow the script to continue smoothly.


MahtraDR:
Good add. Thanks!

PR #6869, opened 2024-07-17
[scripts][combat-trainer] Honor stop_hunting_if_bleeding

MahtraDR:
We were defaulting to exiting combat with untendable bleeders. Now we honor the `stop_hunting_if_bleeding` setting, as with weird necros who want to heal in combat...


desertkaz:
nice. I was actually testing this same fix on my local after booting the necro back up with the recent changes.

MahtraDR:
> nice. I was actually testing this same fix on my local after booting the necro back up with the recent changes.

I'm still shocked nobody had reported this till now. I added this section 11 months ago!

PR #6870, opened 2024-07-17
[scripts][crossing-repair] Add Performance check

dantiadr:
Don't try to kick off performance if skill is capped.


MahtraDR:
Thank you!

PR #6878, opened 2024-07-30
[script] [combat-trainer] Devour functionality with necromancer_healing

hat071af:
Adds a fail-safe to not recast Devour if it is still active when using necromancer_healing settings.


MahtraDR:
Good one. Thank you.

PR #6882, opened 2024-07-30
[script] [combat-trainer] Added Call from Within functionality to combat-trainer. 

hat071af:
Added combat-trainer logic for CFW. This is essentially a cut and paste of the CFB code with tweaks made to handle the differences between the spells.

Works independently of CFB and they can both be utilized together using the settings below. The script will prioritize summoning a dormant CFB > creating a CFW > creating a CFB in order of least to most complicated to create in order to get your pets going as quickly as possible.
```
# zombie settings. Making a zombie in combat requires a waggle set called 'zombie' with CFB in it.
zombie:
  make: true
  summon: false
  stance: offense
  behavior: aggressive

# bonebug settings. Making a bonebug in combat requires a waggle set called 'bonebug' with CFW in it.
bonebug:
  make: true
```
Example waggles:
```
  cfb: &cfb
    Call from Beyond:
      mana: 30
  cfw: &cfw
    Call from Within:
      mana: 6

  zombie:
    << : *cfb
  bonebug:
    << : *cfw 
```

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -435,6 +435,9 @@ class LootProcess
     @make_zombie = settings.zombie['make']
     echo("  @make_zombie: #{@make_zombie}") if $debug_mode_ct
 
+    @make_bonebug = settings.bonebug['make']
+    echo("  @make_bonebug: #{@make_bonebug}") if $debug_mode_ct
+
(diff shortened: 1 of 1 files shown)


MahtraDR:
Thank you!

PR #6883, opened 2024-07-30
[scripts][pilgrimage] added town override functionality

mikejcook:
pilgrimage.lic now includes an optional override to specify the town in which you wish to search for shrines.

### DIFF
--- a/data/base-shrines.yaml
+++ b/data/base-shrines.yaml
@@ -62,24 +62,8 @@ shrines:
   - location: The altar at Misenseor Abbey
     start_room: 6138
     push: altar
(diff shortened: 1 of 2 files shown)


MahtraDR:
Could you please split this into the data change, and then the script change? We merge the data first, since that's completely harmless. If the script change is somehow bad, at worse, we could just revert the script change without also reverting the data change.

mikejcook:
> Could you please split this into the data change, and then the script change? We merge the data first, since that's completely harmless. If the script change is somehow bad, at worse, we could just revert the script change without also reverting the data change.

All set.

MahtraDR:
Thank you!

PR #6886, opened 2024-07-31
[script] [tome] Add additional study failure message

dantiadr:
Add failure message for when you try to study your tome in a room where it is not allowed (such as the Crossing Bank Lobby). This allows the script to continue gracefully, without hanging due to no recognized response.


MahtraDR:
Thank you!

PR #6892, opened 2024-08-16
[scripts][combat-trainer] end GS pause loop after 10 seconds

mikejcook:
So far, two complaints on Discord about instances where after casting GS, combat-trainer "hangs" not doing anything. After casting GS, CT pauses until the warrior is found. If the warrior never appears, it'll pause indefinitely. Now, after 10 seconds, we'll move on.

Log of casting the spell and having it immediately fail. combat-trainer then hung for a couple minutes before I manually stopped it.

https://pastebin.com/aYpr1qRH

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1712,7 +1712,13 @@ class SpellProcess
 
     return unless @command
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
Thanks!

PR #6894, opened 2024-08-24
[scripts][sort-scrolls] adjust handling of scrolls with > 2 words in the adjective segment

desertkaz:
three word scrolls were having parsing issues.  unable to get `sanguine paper scroll`.  scroll system puts it all into the 15 characters for adjective, so we can split and use the first + last to successfully grab.  tested with various paper colors.

### DIFF
--- a/sort-scrolls.lic
+++ b/sort-scrolls.lic
@@ -130,11 +130,12 @@ class ScrollSorter
   def sort_scrolls
     scrolls = DRCI.get_scroll_list_in_container(@default_container)
     scrolls.each do |s|
-      next unless get_scroll(s)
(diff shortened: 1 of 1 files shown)


MahtraDR:
Sorry, I somehow missed this. Not all scrolls work with first + last. Have a look at https://github.com/rpherbig/dr-scripts/blob/aa4247d7c0c0cc47d5df4adfd16aab9f08e0a997/common.lic#L344 for examples. I'm happy to merge this if you want, just know it's not the full solution.

MahtraDR:
What are we doing with this one?

desertkaz:
Closing MR, will look into common update vs script for this.

PR #6897, opened 2024-08-28
add SMH cast messaging

bryandistefano:
- Smite Horde no longer uses "You gesture" after a cast
- Added "As you bow your head and chant" to the cast_messages in base-spells.yaml


bryandistefano:
cast message per wiki:
Screenshot_6

sample cast showing the "You gesture" no longer appears:
Screenshot_7

PR #6903, opened 2024-09-04
[data][base-hunting] Update entries for snow goblins

XTyrsin:
Updated snow goblin hunting rooms for updated area.


MahtraDR:
Thank you!

PR #6911, opened 2024-09-05
[scripts] [safe-room] Add support for EV only places in safe_room_empaths

vtcifer:
Adds the ability to set rooms to be checked for EV (Embrace of the Vela'tohr) plants, instead of just PC empaths.

```yaml
safe_room_empaths:
  - name: EV 
    id: 1234
  - name: EV
    id: 4321
```


MahtraDR:
Thank you!

PR #6917, opened 2024-09-12
[scripts][textsubs] Add textsubs for EV plant status

vtcifer:


vtcifer:
I disagree with rubocop this time and feel the alignment makes more sense this way than how it want it.

PR #6919, opened 2024-09-13
Burgle and base-town updates for Aesry

XTyrsin:
Added pawnshop entry and room numbers for Ranger and Trader guild leaders.
Added pawnshop rejection entry.


MahtraDR:
Thank you!

PR #6920, opened 2024-09-13
[scripts] [script-watch] Initial commit of script-watch.lic

vtcifer:
Provides a new script `script-watch.lic` that gives you a window that shows which scripts are running and their state (if paused), options to show script arguments, customize the format/which scripts are hidden from the window, and make the names clickable to pause/unpause.

The idea is to emulate the genie script bar as much as possible, for lich scripts.


MahtraDR:
This looks pretty cool. Thanks!

PR #6936, opened 2024-10-16
[data] [base-herbs] Hisan salve fix 

hat071af:
Moved hisan into skin section rather than nerves which wasn't being referenced.

### DIFF
--- a/data/base-herbs.yaml
+++ b/data/base-herbs.yaml
@@ -2,7 +2,7 @@
 remedies:
   external:
     skin:
-    #- aloe balm #External - not available for purchase in any shop
+    # - aloe balm # External - not available for purchase in any shop
     - lujeakave elixir #Internal - bought at select shops only
     limbs:
     - jadice flower #ext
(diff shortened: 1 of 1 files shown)


data/base-herbs.yaml
```diff
@@ -26,7 +27,8 @@ remedies:
   scars:
     general:
     skin:
-    - cebi root
+    - cebi root #ext
+    - hisan salve #int
```
MahtraDR, line 31:
```suggestion
    - cebi root # External
    - hisan salve # Internal
```

data/base-herbs.yaml
```diff
@@ -2,6 +2,7 @@
 remedies:
   external:
     skin:
+    #- aloe balm #External - not available for purchase in any shop
```
MahtraDR, line 5:
```suggestion
    # - aloe balm # External - not available for purchase in any shop
```

MahtraDR:
Thanks!

PR #6937, opened 2024-10-18
[script] [combat-trainer] Option for Bard cyclic to segue to on combat end instead of releasing

hat071af:
Adds an option for Bards to segue to a chosen cyclic spell when combat ends instead of releasing their cyclic. If the segue fails it will still release the original cyclic if it's not on the cyclic_no_release list. 

These are the settings needed in your character yaml:
```
# Set a cyclic spell to segue to instead of releasing a cyclic when combat-trainer ends. 
# Bard only setting. Must know how to segue. 
# Strongly recommended to use a justice safe spell. Works best when using a spell from cyclic_no_release: 
# Use the cyclic spell abbrevation. E.g. botf. The segue parser doesn't like full spell names.
segue_spell_on_stop: botf
segue_prep_on_stop: 25
```

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1392,6 +1392,9 @@ class SpellProcess
     if game_state.finish_spell_casting?
       echo('SpellProcess::clean_up') if $debug_mode_ct
       game_state.next_clean_up_step
+      if DRStats.guild == "Bard" && @settings.segue_spell_on_stop
+        DRCA.segue?(@settings.segue_spell_on_stop,@settings.segue_prep_on_stop)
+      end
       DRCA.release_cyclics(@settings.cyclic_no_release)
       if checkprep != 'None'
         DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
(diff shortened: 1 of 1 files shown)


MahtraDR:
This looks really neat. Sorry I missed it earlier. It does need those two settings in base-yaml.yaml blanked out though.

hat071af:
> This looks really neat. Sorry I missed it earlier. It does need those two settings in base-yaml.yaml blanked out though.

Hrm. Can you elaborate/clarify? I tested it without them existing in either base.yaml and my character-setup.yaml and it just skips the 3 lines I added. I.e. segue_spell_on_stop doesn't exist so the if statement ` if DRStats.guild == "Bard" && @settings.segue_spell_on_stop` returns false and it moves on to the usual release cyclic check.

MahtraDR:
> > This looks really neat. Sorry I missed it earlier. It does need those two settings in base-yaml.yaml blanked out though.
> 
> Hrm. Can you elaborate/clarify? I tested it without them existing in either base.yaml and my character-setup.yaml and it just skips the 3 lines I added. I.e. segue_spell_on_stop doesn't exist so the if statement ` if DRStats.guild == "Bard" && @settings.segue_spell_on_stop` returns false and it moves on to the usual release cyclic check.
> 
> I can add them to base.yaml if needed, though.

Yes please add them. Weird, without them defined at all it should error, ah well. Good practices chug one! :)

MahtraDR:
Thanks!

PR #6944, opened 2024-10-22
[data][base-town] Flesh out Knife Clan a bit

Kiriawen:
typo fix

### DIFF
--- a/data/base-town.yaml
+++ b/data/base-town.yaml
@@ -262,7 +262,7 @@ Crossing:
 Knife Clan:
   currency: kronars
   npc_empath:
-  id: 6218 # This is dokt
+    id: 6218 # This is dokt
 Darkling Wood:
   currency: dokoras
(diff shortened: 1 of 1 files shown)


data/base-town.yaml
```diff
@@ -27,7 +27,7 @@ Crossing:
     id: 895
   deposit: &crossing_deposit
     id: 1900
-  exchange:
+  exchange: &crossing_exchange
```
MahtraDR, line 30: Why add this anchor when it isn't used?

Kiriawen: I was gonna use it for knife clan but decided not to as it wouldn't be needed.  Having the anchor there wouldn't impact anything by leaving it and I was being lazy.

MahtraDR, line 30:
```suggestion
  exchange:
```

PR #6945, opened 2024-10-22
[scripts][safe-room]Add support for custom healing plants

Tirost:
Added yaml settings to define adjectives and nouns for healing plants and modified safe-room to check for them. Right now, that adds one plant type, "a swaying plant dancing to a melody only heard by the vela'tohr," from the HE store Adaptive Botany, but it will allow for more versatile incorporation of new plant types introduced by fest stores or modified by alterers.


Tirost:
Yes.

MahtraDR:
How about we make this a list, instead of a pseudo-regex stored in yaml? It'll make the code a bit easier to read, and users can just add a line. no pipes, nor guesswork, etc... We can build the regex from the list inside the script. Instead of using the string-sorta-regex thing.

Tirost:
That sounds even better, though perhaps beyond my ability to code. Are you thinking of something that would look like:
```healing_plant_adjective:
- vela'tohr
- swaying```
Or would it be better to keep it as a horizontal entry, just without the quotation marks and pipes, like: `healing_plant_adjective_regex: vela'tohr swaying` ?

**Comment by MahtraDR** [comment]
Have a play with this. If you get stuck, let me know.

base.yaml would look something like this 

```yaml
healing_plant_adjectives:
  - vela'tohr
  - swaying

healing_plant_nouns: 
  - briar
  - bush
  - plant
  - shrub
  - thicket
  - thornbush
```

Please make sure to add two empty entries in base_empty.yaml as well, namely:
```yaml
healing_plant_adjectives: []
healing_plant_nouns: []
```

in safe-room you'd then have
```ruby
@plant_adjectives = settings.healing_plant_adjectives
@plant_nouns = settings.healing_plant_nouns
```

Then we combine each array into a regex of its own.

```ruby
@adjectives_regex = Regexp.union(settings.healing_plant_adjectives)
@noun_regex = Regexp.union(settings.healing_plant_nouns)
@plant_regex = Regexp.new("#{@adjectives_regex} #{@noun_regex}")
```

Tirost:
Made some updates, and this version seems to be working well, and uses settings in the format you mentioned. I'll keep testing to confirm it works with a variety of plant types.

safe-room.lic
```diff
@@ -15,15 +15,20 @@ class SafeRoom
 
     args = parse_args(arg_definitions)
     settings = get_settings
+    @plant_adjectives = settings['healing_plant_adjective_regex']
+    @plant_nouns = settings['healing_plant_noun_regex']
```
MahtraDR, line 19:
```suggestion
```

MahtraDR: Remove these two lines which are no longer used, and this looks really good!

Tirost:
Done! Really appreciate your help on this.

MahtraDR:
Thank you!

PR #6947, opened 2024-10-26
[scripts][dependency][bootstrap] Prep for when drinfomon is in core lich

MahtraDR:
Ideally this should cause exactly zero behaviour changes. The goal here is to be ready for when drinfomon is in core lich, and not try and load drinfomon.lic when it's already there...

### DIFF
--- a/bootstrap.lic
+++ b/bootstrap.lic
@@ -61,7 +61,7 @@ end
 scripts_to_run = args.flex || []
 echo scripts_to_run.to_s if UserVars.bootstrap_debug
 
-if scripts_to_run.include?("drinfomon") && DRINFOMON_IN_CODE_LICH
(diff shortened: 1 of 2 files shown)


bootstrap.lic
```diff
@@ -60,6 +60,12 @@ end
 
 scripts_to_run = args.flex || []
 echo scripts_to_run.to_s if UserVars.bootstrap_debug
+
+if scripts_to_run.include?("drinfomon") && DRINFOMON_IN_CODE_LICH
```
MahtraDR, line 64:
```suggestion
if scripts_to_run.include?("drinfomon") && DRINFOMON_IN_CORE_LICH
```

MahtraDR:
Makes dr-scripts ready for https://github.com/elanthia-online/lich-5/pull/686

PR #6954, opened 2024-10-27
[scripts][herb-stock] Add Riverhaven and Ratha

MikeLC-hub:
Added Riverhaven and Ratha as hometown options.

Set athletics minimum for Riverhaven hometown as 150.


MahtraDR:
Thank you!

Issue #6970, opened 2024-11-08
[Dependency] Add update command function


Kiriawen:
Add a command to check git for updates to facilitate people checking for script or data updates without having to relaunch.

PR #6973, opened 2024-11-08
[scripts][dependency] Add messaging for forced script downloads

MahtraDR:
Adds messaging when scripts are force-checked, and adds the option to force-check all scripts.

Resolves https://github.com/rpherbig/dr-scripts/issues/6970

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -387,6 +387,7 @@ class ScriptManager
   def get_script(filename, force = false)
     filename = filename.dup
     echo("checking:#{filename} force:#{force}") if @debug
+    echo("Downloading #{filename}") if force
     download_script(filename, force)
   end
 
@@ -1638,7 +1639,7 @@ def get_script(script_name)
 end
 
 def force_refresh_scripts
-  $manager.start_scripts
+  $manager.start_scripts(true)
 end
 
 def list_tracked_scripts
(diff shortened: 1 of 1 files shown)


Kiriawen:
usage would be `;eq force_refresh_scripts` ?

MahtraDR:
> usage would be `;eq force_refresh_scripts` ?

There are two commands.

One - Force-gets all scripts (I am not sure we actually want this, we may only want it to CHECK all scripts, and get the ones that are out of date?)
```
;eq force_refresh_scripts
```
Two - force-gets one script
```
;eq get_script('script-name')
```

Note this should resolve https://github.com/rpherbig/dr-scripts/issues/6970 however we end up tuning it.

Kiriawen:
yeah having it check for updates would be better than redownloading everything.  Save on bandwidth at least

Kiriawen:
though having both a way to force a redownload of everything and a way to just check for updated files wouldn't be a bad idea either.

MahtraDR:
> though having both a way to force a redownload of everything and a way to just check for updated files wouldn't be a bad idea either.

The way it works now, it'll check every file and redownload if they're out of date or don't exist. Is that good enough here?

Kiriawen:
yep, that works.  That's what I was expecting.

PR #6977, opened 2024-11-11
[scripts][sew] Use lift instead of stow feet

MahtraDR:
Instead of doing a blanket `stow feet` we will now `lift` and see what we have, then put it away using the crafting stow functions, which should put things back into the crafting container, and not the default stow container.

### DIFF
--- a/sew.lic
+++ b/sew.lic
@@ -228,7 +228,13 @@ class Sew
         swap_tool('sewing needles')
         command = "push my #{@noun} with my sewing needles"
       when 'What were you referring', 'I could not find what you were'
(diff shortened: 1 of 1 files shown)


Raykyn55:
Hey, jus spotted this. The reason for blanket stow feet here was to get us moving if we had to buy more wax/pins/thread. Technically, since we added stow feet to main lich(I think, i forget exactly how that went), that line could be safely removed. Marrying a hand clearing action to a true lift would just stow the project item, plus whatever we picked up, breaking the script once you bought the wax/pins/thread.

PR #6995, opened 2024-11-19
[scripts][tinker]mechanism updates

desertkaz:
set tool for mechanism in proper case. otherwise, `swap_tool` goes to first entry in `tinkering_tools` (bellow in `base.yaml`)

if creating mechanisms, add mechanism to `bag_items` if not present. `get my mechanism` fails when holding one to combine multiples.

image


MahtraDR:
Thank you!!

Issue #6996, opened 2024-12-02
ammo update for combat-trainer


TegCross:
Currently line 3398, could you please add marble as an ammo type from Droughtman's and from the Cup It shop at Hollow Eve? I did a quick check of the ammo from the shop and it looks like everything else was added except for 'tile'. I added both in the line below if you don't care where it goes. Thank you very much!
https://elanthipedia.play.net/Cup_It_(2)

ammo_pattern = "(?<ammo>arrow|bolt|stone shard|shard|rock|sphere|clump|marble|tile|coral|fist|holder|lump|patella|pellet|pulzone|quadrello|quarrel|quill|stone|stopper|verretto|blowgun dart|crumb|spine|mantrap spike|tiny dragon|icicle|fang|scale|grey-black spike|bacon strip|page|naga|thorn|fragment|talon|cork|button|core|pebble|geode|stub|pit|thimble|doorknob|cone|bell|hunk|piece|present|sleighbell|sprig|star|toy|spiral)s?"

MahtraDR:
Closing this as resolved by #6997 Please reopen if that isn't the case.

PR #6997, opened 2024-12-03
[script][combat-trainer]add marble to ammo pattern

desertkaz:
Droughtman 450 introduced marbles as sling ammo


desertkaz:
fixes #6996 suggestion for `tile` also

MahtraDR:
Thank you!

PR #6998, opened 2024-12-03
[script] [combat-trainer] add stow failure

Kiriawen:
This adds a failure to eliminate a hang for an item that cannot be picked up.
```
[combat-trainer]>stow trunk
> 
That can't be picked up.
```


MahtraDR:
Thank you!

PR #7007, opened 2024-12-10
[scripts][athletics] Add buffing

### LINKED ISSUE #7004: [script] [athletics.lic] add buff routine
Can we move the buff routine that checks for the athletics waggle to be done each loop?

I've been having issues with my athletics spell dropping mid climb practice causing the script to hang indefinitely because the climb became too difficult.

Checking the buff each loop would see that it dropped and recast it before attempting again.

MahtraDR:
Will run `buff athletics` before starting the script. This has no bearing as to where the climbing happens, note. Just that it'll do whatever it was going to do, but with a higher skill level.

Resolves #7004

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -34,6 +34,8 @@ class Athletics
 
     args = parse_args(arg_definitions)
 
+    DRC.wait_for_script_to_complete('buff', ['athletics'])
+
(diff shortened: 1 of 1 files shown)


Kiriawen:
I have an update blurb for this already set to post once you merge it.

PR #7021, opened 2024-12-12
[scripts][dependency] Add case for when we update via repository.lic

MahtraDR:
Catches the case where dependency.lic was updated via repository.lic, so the shasum isn't matching. 

Now we get the shasum of the existing file to compare with what's in the GitHub API, and accept that to be the correct file if those match.

### DIFF
--- a/dependency.lic
+++ b/dependency.lic
@@ -495,19 +495,6 @@ class ScriptManager
     start_script(filename)
   end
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
Also removes ironically named, and no-longer used... `def file_outdated?`

PR #7023, opened 2024-12-16
New Rummage.lic & Rewrite of Sorter.lic

ThatRasputin:
- New rummage.lic file to output container contents in table format using the terminal-table gem
- Rewrite of sorter.lic to allow the option of using table format through the terminal-table gem 
      ;sorter table
  - Refactored other code in the script as part of this effort. It should work when looking at ON items as well as IN. Might need to handle more cases like behind (probably pretty rare).

### DIFF
--- a/rummage.lic
+++ b/rummage.lic
@@ -1,0 +1,95 @@
+# Rummage Parser
+
+custom_require.call(%w[common common-items])
+require 'terminal-table'
+
+class Rummage
+  def run(container)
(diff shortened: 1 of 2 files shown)


ThatRasputin:
This is my first PR, and I'm new to Ruby. I welcome any testing and feedback on these.

ThatRasputin:
I'm going to resubmit this as different PRs for each file. Was silly to do it this way. plus I found something I had to fix in sorter.

PR #7027, opened 2024-12-16
Rework of sorter to accept table arg

ThatRasputin:
- Reworked entire script to allow use of the 'terminal-table' gem
- Added ability to call sorter with the arg 'table' to format output as a table using the gem
- Made some changes to the old-school output of sorter, and left it as the default output
- Assured it worked with looking on, behind, and under
- Checked for use with various surfaces to ensure recognition

### DIFF
--- a/sorter.lic
+++ b/sorter.lic
@@ -2,177 +2,132 @@
   Documentation: https://elanthipedia.play.net/Lich_script_development#sorter
 =end
 
-custom_require.call(%w[common])
+custom_require.call(%w[common common-items])
+require 'terminal-table'
 
-@settings = get_settings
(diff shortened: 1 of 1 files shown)


ThatRasputin:
That was a harrowing journey. Everything should be clean now.

MahtraDR:
Got some screenshots or some example changes?

ThatRasputin:
image
image
image

Retention of the current method as the default option:
image

ThatRasputin:
The next bit I'm working on is to tableify shop output with prices as part of it. I realize that is a little superfluous since it already outputs as lines, but sorting it into categories would be something I would like myself.

MahtraDR:
This looks awesome. Thank you!

PR #7032, opened 2024-12-20
[data] [base-crafting] Add polish to Fang Cove

dantiadr:
Add polish room number and order number for purchase.


MahtraDR:
Thank you!

PR #7033, opened 2024-12-20
[data][base-crafting]Fang Cove enchanting Trainer possible room numbers

XTyrsin:
Added center of society.


MahtraDR:
Thank you!

PR #7038, opened 2025-01-03
[common] [arcana] Resolve potential race condition in crafting_magic_routine

vtcifer:


common-arcana.lic
```diff
@@ -836,14 +836,14 @@ module DRCA
     return if training_spells.empty?
     return if DRStats.mana <= settings.waggle_spells_mana_threshold
 
-    if !XMLData.prepared_spell.eql?('None') && checkcastrt == 0
+    if checkcastrt > 0
+      return
+    elsif !XMLData.prepared_spell.eql?('None') && checkcastrt == 0
       spell = XMLData.prepared_spell
       data = training_spells.find { |_skill, info| info['name'] == spell }.last
       crafting_cast_spell(data, settings)
```
MahtraDR, line 844:
```suggestion
    unless XMLData.prepared_spell.eql?('None') && checkcastrt == 0
      return
     else
      spell = XMLData.prepared_spell
      data = training_spells.find { |_skill, info| info['name'] == spell }.last
      crafting_cast_spell(data, settings)
```

MahtraDR: It seems to me we want to return in all cases where we have no spell prepped. If that's correct, I prefer an unless else, versus and if/elsif not.

vtcifer: > It seems to me we want to return in all cases where we have no spell prepped. If that's correct, I prefer an unless else, versus and if/elsif not.

No, we want to fall through to preparing a spell if none are prepared.

MahtraDR: Sorry, this one went awry. Can you please file it against https://github.com/elanthia-online/lich-5/blob/staging/lib/dragonrealms/commons/common-arcana.rb

PR #7049, opened 2025-01-11
Minor update to base-town.yaml

mdr55:
Added Denalis the pawnshop owner to base-town.  This will support ;pawn-items being able to work at Leth Deriel.


MahtraDR:
Thank you!

PR #7050, opened 2025-01-12
[common][travel] failure pattern added

mdr55:
Added pattern on account of some B&E items which currently hangs bput in sell_item, when called from ;pawn-items.

Example:
Cormyn shakes his head and says, "A tall silver desk lamp with an ornate acanth base isn't worth my time."

### DIFF
--- a/common-travel.lic
+++ b/common-travel.lic
@@ -23,7 +23,8 @@ module DRCT
       /I need to examine the merchandise first/,
       /That's not worth anything/,
       /I only deal in pelts/,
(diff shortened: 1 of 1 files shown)


mdr55:
maybe hold off on this for a while, or until someone can grab the patterns from the other pawnshop owners.  Haven and Aesry at least also have their own unique texts:
#Ioun shakes his head and says, "Nope.  Not worth buying."
#Ishh briefly glances at your pajama top and says, "I have no need for that."

PR #7055, opened 2025-01-15
[script][combat-trainer] fix for dissecting constructs

mdr55:
The minimum necessary for bput to handle the !dissect message from DR.


MahtraDR:
Thanks!

PR #7056, opened 2025-01-15
[script][combat-trainer] QOL improvement to use of collect

mdr55:
Mask the use of collect if we have a loaded weapon.  Its likely that the RT from collect will result in the weapon switch window being exceeded and wasting the load & aim time already invested.  Given collect is less time critical and not directly combat related, suggest its ok to drop it in preference to losing an aim slot.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2955,15 +2955,14 @@ class TrainerProcess
       DRCA.activate_barb_buff?('Avalanche')
     when /^Collect$/i
       # only collect if weapon is not loaded, as collect's high RT has a high chance of overrunning weapon switch window
-      if !game_state.loaded then
-        game_state.sheath_whirlwind_offhand
-        DRC.retreat
-        game_state.engage unless game_state.npcs.empty? || game_state.retreating?
-        DRC.collect(@forage_item)
-        waitrt?
-        game_state.wield_whirlwind_offhand
-        DRC.kick_pile?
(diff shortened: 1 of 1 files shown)


combat-trainer.lic
```diff
@@ -2954,13 +2954,16 @@ class TrainerProcess
...
+      # only collect if weapon is not loaded, as collect's high RT has a high chance of overrunning weapon switch window
+      if !game_state.loaded then
+        game_state.sheath_whirlwind_offhand
+        DRC.retreat
+        game_state.engage unless game_state.npcs.empty? || game_state.retreating?
+        DRC.collect(@forage_item)
+        waitrt?
+        game_state.wield_whirlwind_offhand
+        DRC.kick_pile?
+      end
```
MahtraDR, line 2966:
```suggestion
return unless game_state.loaded

game_state.sheath_whirlwind_offhand
DRC.retreat
game_state.engage unless game_state.npcs.empty? || game_state.retreating?
DRC.collect(@forage_item)
waitrt?
game_state.wield_whirlwind_offhand
DRC.kick_pile?
```

MahtraDR: Uh, with the correct indentation, of course...

mdr55: Updated, looks much cleaner now

MahtraDR:
Thanks!

PR #7061, opened 2025-01-17
[script][combat-trainer] Improved handling of constructs

mdr55:
2 changes related to better handling of constructs

The first around line 822 is for a perform butchery on a construct to be gracefully handled - this affects only necros

The second around line 996 is an expansion of PR #7055 to actively flag a construct rather than just ignore the DR response.  This should affect all (part of skin/dissect code not necro ritual code)


MahtraDR:
Thank you!

PR #7063, opened 2025-01-18
[script] [ranger-companion] Added user variable for food containers, …

hat071af:
…modernized some script calls, fixed script exiting after feeding raccoon


MahtraDR:
Looks like a great improvement. Thank you!

PR #7064, opened 2025-01-18
Update base-town.yaml

BinuDR:
Added Shard's nexus point to Hibarnhvidar


BinuDR:
Closing this since we have a setting for `nexus_town` to use instead of automatically forcing people to go to Shard from Hib.

Issue #7052, opened 2025-01-13
;avtalia display window


hat071af:
I'd like a display window for ;avtalia that updates whenever the script updates something. I took a look at ;moonwatch and there's some stuff in there that could probably be used, but I'm not sure how to actually do it since unlike ;moonwatch the entries it tracks are variable. I.e. moons are fixed but which cambrinth a player tracks is not.
`;e echo UserVars.avtalia` displays the info I want but it's super unreadable in its raw form and includes stuff I don't care about.

So for example, instead of:

> [exec1: {"cambrinth prism"=>"mana"=>72, "cap"=>108, "time_seen"=>2025-01-12 16:08#33.1673239 -0800!#, "arek domisaha"=>"mana"=>0, "cap"=>75, "time_seen"=>2025-01-12 16:08#23.1594908 -0800!#, "cambrinth pilonu"=>"mana"=>60, "cap"=>60, "time_seen"=>2025-01-12 16:08#33.1673547 -0800!#, "anklet"=>"mana"=>50, "cap"=>50, "time_seen"=>2025-01-12 16:12#03.328997 -0800!#, "armband"=>"mana"=>50, "cap"=>50, "time_seen"=>2025-01-12 16:13#43.4038298 -0800!#, "skinning knife"=>"mana"=>15, "cap"=>15, "time_seen"=>2025-01-12 16:08#33.1673649 -0800!#, "cambrinth ring"=>"mana"=>5, "cap"=>5, "time_seen"=>2025-01-12 16:08#33.1673682 -0800]
> --- Lich: exec1 has exited.

It'd show:

> cambrinth prism: 72/108
> arek domisaha: 0/75
> cambrinth pilonu: 60/60
> anklet: 50/50
> armband: 50/50
> skinning knife: 15/15
> cambrinth ring: 5/5

urbaj-dr:
Take a look at https://github.com/urbaj-dr/dr-scripts/blob/master/avtalia.lic and see if that does what you're expecting. Run the modified avtalia.lic with the arg `window`

MahtraDR:
Share your changes with the main repo! 😀🙏

MahtraDR:
Closing as resolved by https://github.com/elanthia-online/dr-scripts/pull/7065

PR #7065, opened 2025-01-18
Update avtalia.lic with window

urbaj-dr:
Adds a "window" argument to output mana values of avtalia cambrinth items to a GUI window, similar to moonwatch. (Per issue #7052 request)


MahtraDR:
Thanks!

PR #7071, opened 2025-01-24
[script][combat-trainer] YACTOM - yet another CT optimisation mod - "focus_threshold" 

mdr55:
The idea is that once all weapons being trained reach focus_threshold, CT will begin to focus solely on that single weapon.
This will stop the constant weapon switching that most people take for granted with a low target_increment value, and improve
gain efficiency for the typical CT user who is training everything.
Focus mode will disable once ANY skill drops below the threshold.

Initial testing seems ok, let this bake for a bit and/or take it out for a spin if you'd like.

More details are in the code comments.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4878,25 +4878,25 @@ class GameState
     # The code will allow CT to focus on a single weapon once we know all weapons are draining;
     # the focus means there is no weapon switching costs being incurred.
 
-    if @focus_threshold then
-      if (@focus_threshold > 10) then
-        if !@focus_threshold_active && weapon_training.reject { |skill, _| DRSkill.getxp(skill) > @focus_threshold }.empty? then
-          # all weapons above threshold
-          DRC.message("Focussing on single weapon")
-          @target_action_count = 1000
-          @target_increment = 34
-          @focus_threshold_active = true
-          return true
-        elsif @focus_threshold_active && weapon_training.any? { |skill, _| DRSkill.getxp(skill) < (@focus_threshold / 2) } then
-          # any weapon below threshold/2
-          DRC.message("Spreading attention to all weapons")
-          @target_action_count = settings.combat_trainer_action_count
-          @target_increment = settings.combat_trainer_target_increment
-          @focus_threshold_active = false
-          return true
-        end
-      end
+  if (@focus_threshold > 10)
+    if !@focus_threshold_active && weapon_training.reject { |skill, _| DRSkill.getxp(skill) > @focus_threshold }.empty?
+      # all weapons above threshold
+      DRC.message("Focussing on single weapon")
+      @target_action_count = 1000
+      @target_increment = 34
+      @focus_threshold_active = true
+      return true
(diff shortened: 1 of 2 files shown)


mdr55:
Code working well so far.  I think this can be reviewed now.

combat-trainer.lic
```diff
@@ -4324,6 +4324,10 @@ class GameState
     @weapons_to_train.each { |skill_name, _weapon_name| @no_gain_list[skill_name] = 0 }
     echo("  @no_gain_list: #{@no_gain_list}") if $debug_mode_ct
 
+    @focus_threshold_active = false
+    @focus_threshold = settings.combat_trainer_focus_threshold
```
MahtraDR, line 4328: We should set a default for this (zero?) in base.yaml.

combat-trainer.lic
```diff
@@ -4862,6 +4866,37 @@ class GameState
...
+        elsif @focus_threshold_active && weapon_training.any? { |skill, _| DRSkill.getxp(skill) < (@focus_threshold / 2) } then
+          # any weapon below threshold/2
+          DRC.message("Spreading attention to all weapons")
+          @target_action_count = settings.combat_trainer_action_count
+          @target_increment = settings.combat_trainer_target_increment
+          @focus_threshold_active = false
+          return true
+        end
+      end
+    end
```
MahtraDR, line 4899:
```suggestion
  if (@focus_threshold > 10)
    if !@focus_threshold_active && weapon_training.reject { |skill, _| DRSkill.getxp(skill) > @focus_threshold }.empty?
      # all weapons above threshold
      DRC.message("Focussing on single weapon")
      @target_action_count = 1000
      @target_increment = 34
      @focus_threshold_active = true
      return true
    elsif @focus_threshold_active && weapon_training.any? { |skill, _| DRSkill.getxp(skill) < (@focus_threshold / 2) }
      # any weapon below threshold/2
      DRC.message("Spreading attention to all weapons")
      @target_action_count = settings.combat_trainer_action_count
      @target_increment = settings.combat_trainer_target_increment
      @focus_threshold_active = false
      return true
    else
      return false
    end
  end
```

MahtraDR (changes requested):
Couple of comments here.

We do need a default setting in base.yaml.

I have removed the explicit `then` from your `if` statements. Ruby convention is they're implied.

We also don't need a nested test for `@focus_threshold` (specially since we'll have a default...) hence we can remove one if statement altogether.

combat-trainer.lic
mdr55: That would work, a zero would fail the hard minimum threshold check of 10, effectively disabling it by default.  I can add that to the PR if you'd like.

MahtraDR: Yes please. As a general rule, if we're adding new settings, we add default settings to base.yaml, with comments explaining their use.

mdr55: cheers, will do

mdr55: added into base

PR #7072, opened 2025-01-25
[script][combat-trainer] YACTOM: offhand_gain_check

mdr55:
offhand_gain_check is a feature that is activated by combat_trainer_offhand_gain_check: value in the yaml  (suggested value of 2)

Its the offhand equivalent of the gain_check feature that was merged in PR #7068 - it will switch offhand weapons when they stop gaining.

This PR also contains a fix for a non-fatal edge case affecting gain_check (most weapons blacklisted + remainder at 34)

more details in code comments

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -5348,21 +5348,21 @@ class GameState
       offhand_current_exp = DRSkill.getxp(@last_offhand_skill)
 
       # keep track of no gain occurences
-      if (offhand_current_exp <= @offhand_last_exp) && (offhand_current_exp != 34) then
+      if (offhand_current_exp <= @offhand_last_exp) && (offhand_current_exp != 34)
         @offhand_no_gain_list[@last_offhand_skill] += 1
       else
         @offhand_no_gain_list[@last_offhand_skill] = 0
       end
 
       # blacklist weapon if gain check failed
-      if (@offhand_no_gain_list[last_offhand_skill] > @offhand_gain_check) then
+      unless @offhand_no_gain_list[last_offhand_skill] <= @offhand_gain_check
         DRC.message("WARNING: Suppressing #{last_offhand_skill} in offhand due to skill not training")
         @offhand_blacklist |= [@last_offhand_skill]
       end
       options -= @offhand_blacklist
 
       # temporarily lock out skill if mindlocked
-      options -= [@last_offhand_skill] if offhand_current_exp == 34
+      options -= [@last_offhand_skill] unless offhand_current_exp < 34
     end
 
     echo "determine_aiming_skill::options: #{options}" if $debug_mode_ct
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4384,7 +4384,6 @@ class GameState
     @offhand_no_gain_list = Hash.new(0)
     @aiming_trainables.each { |skill_name| @offhand_no_gain_list[skill_name] = 0 }
     echo("  @offhand_no_gain_list: #{@offhand_no_gain_list}") if $debug_mode_ct
-    
     @focus_threshold_active = false
     @focus_threshold = settings.combat_trainer_focus_threshold
     echo("  @focus_threshold: #{@focus_threshold}") if $debug_mode_ct
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -163,13 +163,16 @@ class SetupProcess
   def determine_next_to_train(game_state, weapon_training, ending_ranged)
     return unless game_state.skill_done? || !weapon_training[game_state.weapon_skill] || ending_ranged
 
-    unless @ignore_weapon_mindstate
+    # skip_all_weapon_max_check is a kludge for a transient condition from blacklisting a skill with
+    # all remaining skills at 34
+    unless @ignore_weapon_mindstate || game_state.skip_all_weapon_max_check
       # check if all weapons are maxed exp to prevent rapid cycling
       if weapon_training.reject { |skill, _| DRSkill.getxp(skill) == 34 }.empty?
         echo 'all weapons locked, not switching' if $debug_mode_ct
         return
       end
     end
+    game_state.skip_all_weapon_max_check = false
 
     echo('new skill needed for training') if $debug_mode_ct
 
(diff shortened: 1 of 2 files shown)


combat-trainer.lic
```diff
@@ -4290,6 +4295,9 @@ class GameState
     @gain_check = settings.combat_trainer_gain_check
     echo("  @gain_check: #{@gain_check}") if $debug_mode_ct
 
+    @offhand_gain_check = settings.combat_trainer_offhand_gain_check
```
MahtraDR, line 4298: Needs a default setting in base.yaml...

combat-trainer.lic
```diff
@@ -4324,6 +4332,14 @@ class GameState
     @weapons_to_train.each { |skill_name, _weapon_name| @no_gain_list[skill_name] = 0 }
     echo("  @no_gain_list: #{@no_gain_list}") if $debug_mode_ct
 
+    @offhand_last_exp = -1
+    @last_offhand_skill = ""
+    @offhand_blacklist = []
```
MahtraDR, line 4337:
```suggestion
    @offhand_last_exp = nil
    @last_offhand_skill = String.new
    @offhand_blacklist = Array.new
```

combat-trainer.lic
```diff
@@ -5312,6 +5334,37 @@ class GameState
...
     options -= @summoned_weapons.map { |info| info['name'] } # remove summoned weapons because they may not exist for offhand use
+
+    # offhand_gain_check code: add in some extra logic to stop training an offhand wielded
+    # weapon that is not gaining any MS at all.  Reduces the opportunity cost.
+    # blacklist on the offhand will not blacklist on mainhand by intent, as offhand has accuracy penalties
+    # the code will modify the options list if active
+    if @offhand_gain_check then
+
+      # if we know answer is empty list, go right to the end and bypass unwanted side effects
+      return nil if ((options - @offhand_blacklist) == [])
```
MahtraDR, line 5345:
```suggestion
      return nil if (options - @offhand_blacklist).empty?
```

combat-trainer.lic
```diff
@@ -5312,6 +5334,37 @@ class GameState
...
+
+      # keep track of no gain occurences
+      if (offhand_current_exp <= @offhand_last_exp) && (offhand_current_exp != 34) then
+        @offhand_no_gain_list[@last_offhand_skill] += 1
+      else
+        @offhand_no_gain_list[@last_offhand_skill] = 0
+      end
+
+      # blacklist weapon if gain check failed
+      if (@offhand_no_gain_list[last_offhand_skill] > @offhand_gain_check) then
```
MahtraDR, line 5358:
```suggestion
      unless (@offhand_no_gain_list[last_offhand_skill] < @offhand_gain_check)
```

combat-trainer.lic
```diff
@@ -5312,6 +5334,37 @@ class GameState
...
+        DRC.message("WARNING: Suppressing #{last_offhand_skill} in offhand due to skill not training")
+        @offhand_blacklist |= [@last_offhand_skill]
+      end
+      options -= @offhand_blacklist
+
+      # temporarily lock out skill if mindlocked
+      options -= [@last_offhand_skill] if offhand_current_exp == 34
```
MahtraDR, line 5365:
```suggestion
      options -= [@last_offhand_skill] unless offhand_current_exp < 34
```

MahtraDR (changes requested):
Small changes, see below. General theme that we need default settings in base.yaml when we introduce new settings.

Double check my logic inversions too, from `if` to `unless`

combat-trainer.lic
mdr55: the -1 is to force a comparison fail, is it good Ruby practice to use an arithmetic comparison with nil?

combat-trainer.lic
mdr55: changed

combat-trainer.lic
mdr55: changed

combat-trainer.lic
mdr55: added to base

combat-trainer.lic
mdr55: string and array done

MahtraDR: Leave the -1. I was wrong there. Thanks!

combat-trainer.lic
```diff
@@ -4369,6 +4377,14 @@ class GameState
...
     echo("  @no_gain_list: #{@no_gain_list}") if $debug_mode_ct
 
+    @offhand_last_exp = -1
+    @last_offhand_skill = String.new
+    @offhand_blacklist = Array.new
+
+    @offhand_no_gain_list = Hash.new(0)
+    @aiming_trainables.each { |skill_name| @offhand_no_gain_list[skill_name] = 0 }
+    echo("  @offhand_no_gain_list: #{@offhand_no_gain_list}") if $debug_mode_ct
+
```
MahtraDR, line 4387:
```suggestion

```

combat-trainer.lic
```diff
@@ -4886,7 +4901,7 @@ class GameState
     echo("last_action_count: #{@last_action_count}") if $debug_mode_ct
 
     # gain_check code: logic to account for a skill that is not gaining any MS at all
-    if @gain_check && (weapon_skill != nil) then
+    if (@gain_check > 0) && (weapon_skill != nil) then
```
MahtraDR, line 4904:
```suggestion
    unless ((@gain_check == 0) && weapon_skill.nil?)
```

Is this maybe better here?

combat-trainer.lic
```diff
@@ -5392,6 +5413,37 @@ class GameState
     options -= weapon_training.select { |_skill, weapon| damaris_weapon_options.include?(weapon) }.keys # remove all weapons part of the same Damaris weapon group since they're the same item
     options -= weapon_training.select { |_skill, weapon| swappable_weapon?(weapon) }.keys # remove swappable weapons because they might be swapped into a two-handed state
     options -= @summoned_weapons.map { |info| info['name'] } # remove summoned weapons because they may not exist for offhand use
+
+    # offhand_gain_check code: add in some extra logic to stop training an offhand wielded
+    # weapon that is not gaining any MS at all.  Reduces the opportunity cost.
+    # blacklist on the offhand will not blacklist on mainhand by intent, as offhand has accuracy penalties
+    # the code will modify the options list if active
+    if @offhand_gain_check > 0
```
MahtraDR, line 5421:
```suggestion
    unless @offhand_gain_check == 0
```

Similarly here, would this be cleaner? Or much of a muchness?

mdr55: i'm thinking @offhand_gain_check > 0 reads like - if feature is active, i.e. greater than zero) do (feature)   - to me its more natural presented as a positive than what looks like a double negative to me.  Its idiosyncratic.

The inverted logic works ok though.

MahtraDR: I like how you describe that. That's not unreasonable.

mdr55: we should never have to deal with negative numbers, but if we do, > 0 will protect!

combat-trainer.lic
mdr55: Similar response to offhand - prefer the > 0 version, I did update the nil to .nil? and removed the then to make them more Rubyish.
The inverted logic should be  unless gain_check == 0 || weapon_skill.nil?   , if you really want it in there.

PR #7075, opened 2025-01-26
[script][combat-trainer] Updates for focus_threshold, offhand_trainables

mdr55:
This is a running list to track items so I don't forget, **please don't merge yet** these are non-fatal issues. 

focus_threshold - min threshold check should be >= 10 rather than > 10 
offhand_trainables - offhand skills selection pool to include .... "Offhand Weapon"
focus_threshold - an errant return false removed, it cause behaviour similar to a focus_threshold of zero

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -208,13 +208,14 @@ class SetupProcess
 
     if @offhand_trainables && game_state.focus_threshold_active
 
-      # define what mainhand skills look like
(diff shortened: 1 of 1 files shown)


MahtraDR:
Converted to a draft so it's clear it's not ready. Flip to ready when it is. :)

mdr55:
lets review and push theses out, the 3rd one is impacting on how focus_threshold is intended to work

PR #7076, opened 2025-01-27
[script][combat-trainer]  Updates for focus_threshold

mdr55:
updates

1.  The hard minimum for focus_threshold may be too conservative.  A value of ~4 may be workable for a weapon tert.  For other skillset placements focus_threshold may need to be higher to account for drain whilst stuck on a weapon going for 34/34 (so you don't end up with 0 ms on a skill with no ability to correct until the 34 is done).  Setting the minimum to say > 1 gives the player freedom to set what is needed for their char and specific circumstances. 
2. Conditional CT break bug fix (if focus_threshold was activated + the bottom of the hysterisis band was reached).  Rookie mistake :(

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -4310,6 +4310,7 @@ class GameState
 
     @target_action_count = settings.combat_trainer_action_count
     echo("  @target_action_count: #{@target_action_count}") if $debug_mode_ct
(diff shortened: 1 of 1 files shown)


mdr55:
ready for review.  #2 is conditionally breaking for users of feature

PR #7086, opened 2025-02-01
[data] [base-spells] Added Sorrow entry and cast message

hat071af:


Kiriawen:
please add minmana, maxmana, minrank and maxrank.

```
  Bonegrinder:
    skill: Targeted Magic
    abbrev: bonegrinder
    prep: target
    mana: 20
    mana_type: arcane
    guild: Necromancer
    minmana: 20
    maxmana: 66
    minrank: 250
    maxrank: 1000
```

PR #7087, opened 2025-02-01
[script][combat-trainer] fix for devour blocking looting 

mdr55:
as above, per issues # 7077

currently testing.


mdr55:
Done as much testing as I could without USOL.  Seems to not be performing consume when devour is active now (which was the root of the problem in 7077)

mdr55:
confirmed as ok with original reporter of the issue, suggest merge

MahtraDR:
Thank you!

PR #7088, opened 2025-02-01
[script][combat-trainer] cleanup to include removing symb prep

mdr55:
For some reason cleanup process was not getting rid of an in-progress spell's symb prep; this is a fix for that.  Also removed an orphan condition.

I've tested this ok on my char but I don't have access to !chaos symbs, which I understand has slightly different behaviour and possibly different messaging.

The main idea is of course to not accidentally wipe out your researched symb when CT ends its run.

Can someone test, verify and confirm with a comment here please?

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1485,9 +1485,11 @@ class SpellProcess
       end
       DRCA.release_cyclics(@settings.cyclic_no_release)
 
-      DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
-      DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
-      DRC.bput('release symb', 'You release the', "But you haven't prepared a symbiosis") if @symbiosis
+      unless checkprep
+        DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
+        DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
+        DRC.bput('release symb', 'You release the', "But you haven't prepared a symbiosis") if @symbiosis
+      end
 
       if @tk_ammo
         waitrt?
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -1484,10 +1484,11 @@ class SpellProcess
         DRCA.segue?(@settings.segue_spell_on_stop, @settings.segue_prep_on_stop)
       end
       DRCA.release_cyclics(@settings.cyclic_no_release)
-      if checkprep != 'None'
-        DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
-        DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
-      end
+
+      DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
+      DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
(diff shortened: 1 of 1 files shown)


mdr55:
The existing behaviour (before PR) is a random crap shoot between different states when clean-up sequence finishes:
1. no spell is being interrupted- no issues, no spell cleanup needed.  prep symb goes away when the spell was successfully cast.  Have not checked behaviour if yaml is set for a symb cast that's too high for char.
2. non-symbiosis spell in prep interrupted - no issues, release magic and release mana takes care of it
3. symbiosis spell in prep interrupted <-- this is it.  Release magic and release mana will still leave the prep symb hanging.
Frequency/likelihood of landing on 3. depends on spells in spell lists and other configuration factors.

MahtraDR:
My only concern is for folks using researched symbiosis losing those. That said, I think the way you've changed this brings it in line with other usages of this idea.

mdr55:
The "if @symbiosis" *SHOULD* prevent losing the research symbiosis?  Or is that not how it works with researched symbs - once you turn it on, you can't turn off without losing the research time investment?
If the latter, I'll close the PR :p

mdr55:
The code tracks the @symbiosis state pretty tightly.  I think if @symbiosis == true, then there's a very very good chance prep symb was performed just before prep spell.

MahtraDR:
> The "if @symbiosis" *SHOULD* prevent losing the research symbiosis?  Or is that not how it works with researched symbs - once you turn it on, you can't turn off without losing the research time investment?
> 
> If the latter, I'll close the PR :p

No it's not the latter. If you have a researched symbiosis and you used prep symb you can then release symb. If you release symb twice in quick succession, you will forget your researched symbiosis though. Which was the concern. Your point is good though and I think we'll be fine here.

combat-trainer.lic
```diff
@@ -1484,10 +1484,11 @@ class SpellProcess
...
       end
       DRCA.release_cyclics(@settings.cyclic_no_release)
-      if checkprep != 'None'
-        DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
-        DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
-      end
+
+      DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
+      DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
+      DRC.bput('release symb', 'You release the', "But you haven't prepared a symbiosis") if @symbiosis
```
MahtraDR, line 1490:
```suggestion
      unless checkprep
            DRC.bput('release spell', 'You let your concentration lapse', "You aren't preparing a spell")
            DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
            DRC.bput('release symb', 'You release the', "But you haven't prepared a symbiosis") if @symbiosis
      end
```

mdr55: is checkprep a super secret object squirreled away somewhere in core lich?   Its in a few locations in scripts, but, its never actually defined from what I can see.  Seems like old code left in?

https://github.com/search?q=repo%3Aelanthia-online%2Fdr-scripts%20checkprep&type=code

MahtraDR: https://github.com/elanthia-online/lich-5/blob/061e78cc73b3b10de902cae33af315a241bf7acc/lib/global_defs.rb#L1235

mdr55: oops my bad, committed!

PR #7100, opened 2025-02-09
[script][combat-trainer] Arbitrary condition checks for non-spell buffs

mdr55:
A direct extension of the custom checks idea to non-spell buffs, to facilitate, amongst other things, conditional release of spells.
The original custom condition checks affects the start of a spell; this PR is used to affect the end of a spell (or to affect other non-spell buffs)

Example of original syntax (before PR).  This is still valid and completely functional post PR. 
```
buff_nonspells:
  release usol: 300
```

What the PR allows for:
```
buff_nonspells:
  release usol: 60 && "DRSpells.active_spells['Universal Solvent'] && DRSkill.getxp('Targeted Magic') > 30 && DRSkill.getxp('Thanatology') > 30 && DRSkill.getxp('First Aid') > 30 && DRSkill.getxp('Skinning') > 30"
```

The fake  " && "  is used as a delimiter after the timer value, and before the conditional wrapped up in double quotes.  This variant will shut down USOL only after achieving the defined mindstate targets, rather than based solely on the timer.  In this scenario, the timer is more of a polling period, determining how frequently a check is made on mindstate progress.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2708,7 +2708,7 @@ class AbilityProcess
       timer = game_state.cooldown_timers[action]
 
       # break apart the blob into OG cooldown timer and the new optional conditional string using a fake && as a delimiter
-      blob_array = blob.to_s.split(/ && /,2)
+      blob_array = blob.to_s.split(/ && /, 2)
       cooldown = blob_array[0].to_i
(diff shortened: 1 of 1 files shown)


mdr55:
am reasonably sure from looking at the code and partial testing that CT won't break for thief and barbs.  But I don't have a thief or barb to actually test on.

PR #7108, opened 2025-02-14
Update yiamura.lic regex

Tirost:
Add regex logic to capture additional OBSERVE messaging like the following: You stare at your coralite yiamura.  Very quickly, your head aches as knowledge fills your mind!


MahtraDR:
Thanks!

PR #7114, opened 2025-02-15
Textsubs - Optional Guild magic, spells and spellbook weblinks

MikeLC-hub:
New TextSubs class method to create Elanthipedia weblinks.

Optional:  activated by including 'textsubs_spell_links: true' in your char-setup.yaml

- Hyperlinks the column-formatted SPELLS known spell display.
- Hyperlinks the 'Active Spells' window (linking spells on your person, known to your or not)

Can cause issue to buff-watcher if 'textsubs.lic' is called before 'spellmonitor.lic'. The simple fixes are noted on line 83. (This is another reason for the optional toggle).


MikeLC-hub:
So the Rubocop error is the #quiet on line 1. Turns out that the #quiet is what was used in the old textsubs to have it run 'silently'. The Rubocop correction is misreading the error as a comment, rather than as a lich command.

MahtraDR:
Could you please show a before/after set of screenshots? I can't test this on Profanity. Thanks!

MikeLC-hub:
weblinked_spells

The before is the left, the linked is on the right.

PR #7118, opened 2025-02-20
[script][hunting-buddy] allow multiple HB configs in same file

mdr55:
hunting_profiles_single_yaml: true  will allow multiple HB configs in the same file, by attaching the charname-suffix to hunting_info instead.  See example below

```
hunting_profiles_single_yaml: true

hunting_info:               # default HB profile for circle hunt (TM/Thana/FA/Skin/etc)
- :zone:
  - elder_blue_dappled_prereni
  args:
  :duration: 5
  avoid_buddies: true

hunting_info_weapon:        # HB profile for weapon hunt
- :zone:
  - the_other_elder_blue_dappled_prereni
  args:
  :duration: 6
  avoid_buddies: true
```
The 2nd profile is accessed when ";hunting-buddy weapon" is called, rather than loading char-weapon.yaml.  Obviously this won't have the benefit of actually loading char-weapon.yaml with all of its non-HB settings that could be paired with the 'weapon' HB profile.  The benefit is that it makes it much easier to edit, everything being located in the -setup yaml.

### DIFF
--- a/hunting-buddy.lic
+++ b/hunting-buddy.lic
@@ -58,7 +58,7 @@ class HuntingBuddy
     echo("  hunting_files: #{hunting_files}") if $debug_mode_hunting
 
     hunting_files.each do |character_file_suffix|
-      infos = get_settings([character_file_suffix]).hunting_info
+      infos = retrieve_settings(character_file_suffix)
       infos.each do |info|
         info['args'] ||= []
         # Hunting info from extra config needs the config name passed on in args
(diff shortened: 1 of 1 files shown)


mdr55:
rubocop failure is on existing code?

MahtraDR:
> rubocop failure is on existing code?

Yes. Honestly I never figured out how to remove the loop there. Someday I'll deal with it. 😀

The PR itself... have you ever used a different hunting info without changing any other settings? I'm struggling to conceptualize this one.

mdr55:
I'm starting to ease into split hunts now, and that means at least two different hunting_info sections.  I think with must_be_true & UserVars controlling the spells from T2/T3 and possibly d99 or r99 for the physical attacks, it may be possible to set up the splint hunt without a second file.  Then again my setups are atypical...

MahtraDR:
> I'm starting to ease into split hunts now, and that means at least two different hunting_info sections. I think with must_be_true & UserVars controlling the spells from T2/T3 and possibly d99 or r99 for the physical attacks, it may be possible to set up the splint hunt without a second file. Then again my setups are atypical...

As long as it doesn't bother existing use cases, all good. You've come up with some novel ways of doing things in ways I hadn't conceived, so I'm happy with this one too. I am just sharing my personal experience, since I always have multiple hunting files, and it's near never because of `hunting_info`, but rather because of all the other settings...

mdr55:
I found a solution to what I needed, and it requires more custom changes.  Given that and the typical use of HB is multiple yaml with additional non-HB data, lets park this change unless someone else asks for it.

PR #7119, opened 2025-02-22
chore: GitHub action workflow update

mrhoribu:
Remove redundant `bundle install` usage since ruby\setup-ruby action handles this with proper caching. Also updated all actions to latest v# supported.


MahtraDR:
Thank you. I appreciate it.

PR #7120, opened 2025-02-22
khri adaptation functionality

Thires:
updated base.yaml and combat-trainer.lic for it


mdr55:
its possible that whatever editor you used changed all the newlines CR/LF from whatever the github version had to something else.  Github is recognising all the lines as changed!!

Thires:
hmm, not sure what caused that

Thires:
okay the newest ones look like no conflict going on, not fully sure what was causing that but looking at them now don't see the CR/LF issue or replacing every single line

PR #7128, opened 2025-02-28
[script][burgle] Allow for real time change of mind to keep stolen items

mdr55:
The proposed code change, and the following alias

;alias add keep=;eq $BURGLE.keep if Script.running?('Burgle')

will allow a player to punch in the keep alias if an interesting/new item pops up, even if the yaml is set to dispose (pawn/bin/etc).  As long as the keep command is entered before 'go window' and as long as a wearable trashcan isn't used, burgle will now keep the loot.

### DIFF
--- a/burgle.lic
+++ b/burgle.lic
@@ -109,7 +109,6 @@ class Burgle
       pause 10
       end_burgle
     end
-
   end
 
   def valid_burgle_settings?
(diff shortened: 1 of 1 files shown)


mdr55:
most people have burgle running automatically via T2 (or something).  This allows a change of mind after its already launched, and you can see what item has actually been rolled by DR.  e.g. you have a whitelist, and something came up that wasn't on it that would have been disposed of.
  
Its not about launching it manually.

PR #7134, opened 2025-03-02
[script] [validate] Updated logic for missing setup file / bad extension

vtcifer:
* Moved the check for invalid extension (.yml) to before the presence check for `{character}-setup.yaml` so that it always runs
* Changed outcome of missing `{character}-setup.yaml` file to end checks early since the rest of them are meaningless without that file and gives users a false picture.

### DIFF
--- a/validate.lic
+++ b/validate.lic
@@ -18,15 +18,6 @@ class DRYamlValidator
     setup_data
 
     respond("")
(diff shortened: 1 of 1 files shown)


MahtraDR:
Thank you!

PR #7139, opened 2025-03-15
Update combat-trainer.lic

Algonem12:
Problem with manipulation for empaths after update. Fix is to remove the ordinal_string in the case DRC.bput

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -2878,7 +2878,6 @@ class ManipulateProcess
         game_state.construct(npc)
       else
         manipulate_count += 1
-        npc_occurrences[npc] += 1
       end
     end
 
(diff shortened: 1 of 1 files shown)


mdr55:
Given that the ordinals for "moving mobs" is now taken care of in core lich, the following lines could also be eliminated from the CT code as they're now orphans - they're all local variables used to generate the correct ordinal:

2871     npc_occurrences = Hash.new(0)
2877       index = npc_occurrences[npc]
2878      ordinal_string = $ORDINALS[index]
2885         npc_occurrences[npc] += 1

Algonem12:
I missed 2871, but it is out of there now.

Algonem12:
I removed npc_occurrences[npc] += 1 just wanted to ensure I was seeing that it was orphaned and not related to another portion first.

MahtraDR:
Thanks for this!

PR #7146, opened 2025-03-18
Added Learned Ranks tracking to exp-monitor.lic

MikeLC-hub:
With the switchover to lich5, Wrayth/Stormfront users lost the function to track EXP/RANKS gained over a session.

I have taken the portion of the script that handles tracking and displaying the learned ranks in the field experience window.

Also, I have updated the output to Field Experience Window display to not increase the number of characters, respacing the information to fit in the original space.

I have included a screenshot of the before and after.

expmoniter_learned_ranks

### DIFF
--- a/exp-monitor.lic
+++ b/exp-monitor.lic
@@ -1,5 +1,6 @@
 no_pause_all
 no_kill_all
+silence_me
 
 custom_require.call(['drinfomon'])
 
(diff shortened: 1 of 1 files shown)


MikeLC-hub:
Wow... Rubocop is stickler!

MahtraDR:
I'm going to close this one, for the sole reason that I actually had most of this in core lich already. The only thing that was missing was the exp tracking portion specifically.  You can see the PR here. I'd love it if you could test it! :)

https://github.com/elanthia-online/lich-5/pull/835

MahtraDR:
We'll do it here for now after all. :) Core lich doesn't have a clean way to inject code into the FE from the parser. Let's add this here, and re-add it to core-lich when that's changed to be able to do this cleanly.

PR #7148, opened 2025-03-20
[script] [attunement] enhancements

vtcifer:
- change room looping behavior - within the same login session, repeat calls to attunement will not start at beginning of the room list every time, and instead will pickup where it left off during the last call
- command line option to override mindstate gain goal
- command line option to use roomcount as a limiter (in addition to mindstate goal)
- command line option to reset and restart form the start of the room list


vtcifer:
Also heads up, this is the first use of SessionVars from 5.10.3 that I'm aware of.

MahtraDR:
Thanks. This looks cool!

PR #7149, opened 2025-03-23
Update locksmithing.lic - Update to worn lockboxes

MikeLC-hub:
Added an if_statement to check if the lockbox is a main_container. Will unlock and open the lockbox before wearing if so.

IE. This is an functional update for items such as https://elanthipedia.play.net/Item:Secure_bag_with_a_silversteel_lock .


MahtraDR:
This one I feel doesn't need to be a setting. We detect it's locked, and we unlock and open, and if we unlock and open, we should close and re-lock when done.

MikeLC-hub:
That is a much more straightforward approach.

PR #7150, opened 2025-03-25
[script] [stabbity] switch weapons if current is ineffective; avoid false-positives by not ignoring monsters

KatoakDR:
### Background

While running the Arena on my thief in Duskruin this year, identified a couple issues.

**Issue 1**
stabbity's yaml config lets you choose which weapon to use against different creatures to handle when one might be resistant to slice or impact, for example. However, this requires you to know the noun of the creature before you encounter it. When progressing through the arena, you might encounter new creatures that you haven't configured the correct weapon for. To resolve, you have to update your yaml and rerun the script.

**Issue 2**
stabbity identifies monsters in the room to attack based on NPCs and removing any ignored NPCs (e.g. 'warrior' to ignore an Empath's Alfar warrior). However, this can be a false-positive that ignores 'armored necrotic warrior' leaving the script to do nothing.

### Changes

- Swap between preferred and alternate weapon if the current weapon is not hitting effectively
- Don't ignore any NPCs so always try to watch and engage whoever is in the arena

### Usage

```
;stabbity arena
```

### DIFF
--- a/stabbity.lic
+++ b/stabbity.lic
@@ -216,12 +216,11 @@ class Stabbity
   end
 
   def get_npcs
-    # Not filtering by ignored NPCs because some nouns like "warrior"
-    # which generally refer to an Empath's Alfar warrior, are incorrectly
-    # ignoring enemy warriors like an armored necrotic warrior.
-    # And since this script is for solo thieves, there should be no other NPCs.
-    # DRRoom.npcs - @ignored_npcs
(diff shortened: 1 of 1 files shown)


KatoakDR:
@MahtraDR For your review, thanks

MahtraDR:
Sorry mate, I had seen, but something came up before I could click. Clicking now. Thank you! :)

PR #7154, opened 2025-03-28
[dev] update ruby version to 3.3

KatoakDR:
Per https://github.com/elanthia-online/dr-scripts/pull/7152, updates the `.ruby-version` file (read by `rbenv`) to 3.3. This is for developers working locally so that their ruby version manager and IDE use the appropriate version.

The `Gemfile.lock` file was auto-created when I ran `bundle` command. If you all don't want it then I will add it to the `.gitignore` instead. In Visual Studio Code with the [Ruby LSP](https://marketplace.visualstudio.com/items?itemName=Shopify.ruby-lsp) extension, I could not get the extension to activate without the lockfile.


MahtraDR:
I'd prefer to just have the version bump without the lock file but in honesty this won't impact a lot of people so let's merge it and see how we go! 😁

PR #7155, opened 2025-03-28
[script] [stabbity] handle bonded thrown weapons; check buffs and wands between fights

KatoakDR:
###

While running `;stabbity arena` in Duskruin this year, identified some room for improvements.

### Changes

- Improve handling for bonded thrown weapons by porting logic from `combat-trainer`
- Appropriately time checking buffs and wands to be between arena fights so that you don't interrupt avoiding traps or being hit by the enemy
- Add sample yaml config (also appears that trailing whitespace got trimmed when I saved the file)


MahtraDR:
Thank you!

PR #7157, opened 2025-03-29
[script] [vanity-pet] handle more pet messaging and when homes must be worn

KatoakDR:
I got my first puppy at this year's Duskruin. Noticed some new messaging that the script wasn't expecting, and scenarios where the pet home needed to be worn, not simply stowed in another container. Maybe they always had to be worn, not sure, but now the script handles it and warns user.

```
[vanity-pet]>get wolf-hound puppy

You carefully lift the puppy up by the scruff of its neck and then tuck it gently into the crook of your arm.

[vanity-pet]>put my wolf-hound puppy in my wooden crate

Your wolf-hound puppy nips at your fingers and doesn't seem to want to leave you.
>

... waiting ...

[vanity-pet: *** No match was found after 15 seconds, dumping info]

... text ...

[vanity-pet: for command put my wolf-hound puppy in my wooden crate]

Unable to stow pet. Ensure you are either holding or wearing your pet's home.

--- Lich: vanity-pet has exited.
```

### DIFF
--- a/vanity-pet.lic
+++ b/vanity-pet.lic
@@ -43,29 +43,30 @@ class VanityPet
 
   def stow_pet(pet_name, pet_container)
     echo "stowing pet" if $debug_mode_pet
-    unless DRCI.in_hands?(pet_name) || DRCI.get_item_if_not_held?(pet_name, pet_container)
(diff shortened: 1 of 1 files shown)


MahtraDR:
Happy to merge and will do so now. Note that these pets are uber ultra incredibly fragile. They will climb into boxes that you're about to trash for example. They will jump off you. Just be careful with your pets! 😁

PR #7171, opened 2025-04-29
[script] [bescort] add gondola bypass support for flying mounts

KatoakDR:
### Background

It would be nice if `bescort` would take into consideration if a user has a flying mount to bypass athletics checks when traveling. For my specific use case, when wanting to go under the gondola (instead of riding it or climbing around it) to go between Leth and Shard.

### Changes

- Add a `fly_under_gondola` function to use your flying mount to use the under gondola shortcut
- Validate you are in the correct starting room to ride (or fly under) the gondola

### Example Usage

```yaml
flying_mount: barkcloth rug

personal_wayto_overrides:
  fly_under_gondola_north:
    start_room: 19466 # under gondola at southside of the log
    end_room: 2245 # on road north of gondola at branch
    str_proc: "start_script('bescort', ['fly_under_gondola', 'north']); wait_while{running?('bescort')};"
    travel_time: "unless UserVars.flying_mount then nil else 0.2 end"
  fly_under_gondola_south:
    start_room: 2245 # on road north of gondola at branch
    end_room: 19466 # under gondola at southside of the log
    str_proc: "start_script('bescort', ['fly_under_gondola', 'south']); wait_while{running?('bescort')};"
    travel_time: "unless UserVars.flying_mount then nil else 0.2 end"
```

----

> [!IMPORTANT]
> Adds `fly_under_gondola` function to `bescort.lic` for bypassing gondola athletics checks using flying mounts, with new argument parsing for this feature.
> 
>   - **Behavior**:
>     - Adds `fly_under_gondola` function in `bescort.lic` to bypass gondola athletics checks using flying mounts.
>     - Validates starting room for `fly_under_gondola` and `ride_gondola` functions to ensure correct initiation.
>   - **Argument Parsing**:
>     - Adds `fly_under_gondola` option with `north` and `south` modes to argument definitions in `Bescort` class.
>   - **Misc**:
>     - Adjusts regex for `gondola` argument to ensure exact match.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=elanthia-online%2Fdr-scripts&utm_source=github&utm_medium=referral)<sup> for 91af13ee8ce20c4f9d337a9635c6a6bce1ad3761. You can [customize](https://app.ellipsis.dev/elanthia-online/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -152,6 +152,10 @@ class Bescort
         { name: 'mode', options: %w[north south], description: 'Which direction are you going?' }
       ],
       [
+        { name: 'fly_under_gondola', regex: /fly_under_gondola/i, description: 'Fly over the obstacles under the gondola between Shard and Leth Deriel' },
+        { name: 'mode', options: %w[north south], description: 'Which direction are you going?' }
+      ],
+      [
         { name: 'sandbarge', regex: /sandbarge/i, description: 'The sand barge between Hvaral, Oasis, and Muspari' },
         { name: 'start_location', options: %w[hvaral oasis muspari], description: 'Where are you coming from?' },
         { name: 'end_location', options: %w[hvaral oasis muspari], description: 'Where do you need to get to?' }
@@ -278,6 +282,8 @@ class Bescort
       take_theren_rope_bridge(args.mode)
     elsif args.gondola
       ride_gondola(args.mode)
+    elsif args.fly_under_gondola
+      fly_under_gondola(args.mode)
     elsif args.sandbarge
       take_sandbarge(args.start_location, args.end_location)
     elsif args.desert
@@ -1839,7 +1845,24 @@ class Bescort
     fput('rel mana')
   end
 
+  # Waits for and rides the gondola to the other platform.
   def ride_gondola(mode)
+    north_platform_room_id = 2249
+    south_platform_room_id = 2904
+
+    case mode.downcase
+    when /north/
+      unless Room.current.id == south_platform_room_id
+        echo("Must start at the south platform, room number #{south_platform_room_id}, to ride the gondola north")
+        exit
+      end
+    when /south/
+      unless Room.current.id == north_platform_room_id
+        echo("Must start at the north platform, room number #{north_platform_room_id}, to ride the gondola south")
+        exit
+      end
+    end
+
     case DRC.bput('go gondola', 'There is no wooden gondola here', 'Gondola, Cab')
     when /no wooden gondola here/i
       DRC.bput('look gondola', 'The wooden gondola')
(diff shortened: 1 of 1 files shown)


bescort.lic
```diff
@@ -1839,6 +1839,97 @@ class Bescort
...
+
+    # Set speed to slow so that we go one room at a time
+    # and don't mistakenly run down other paths.
+    use_flying_mount(@flying_mount, 'mount', 'slow') if @flying_mount
+
+    moveset.each { |dir| move(dir) }
+
+    use_flying_mount(@flying_mount, 'dismount') if @flying_mount
+
+    manual_go2(end_room)
```
KatoakDR, line 1923: 🤔 There is something off but I'm not sure what yet, but once you're under the gondola and trying to get to the south platform and you get to the room just after climbing the log, the pathing algorithm thinks the shortest route from that room to the south platform is to go back to the north platform. This causes the bescort script to backtrack and loop again, never getting to the final destination.

This quirk is only happening for me when I have Athletics >= 540, no flying_mount, and use the new `bypass_gondola` bescort logic.

If I have a flying mount then the workaround of setting `UserVars.athletics = nil` so that the pathing algorithm ignores skill-based `timeto` values then I can go between the north and south platforms perfectly.

I think in my next round of testing I'll adjust so that my bescort proposal does NOT try to climb through the rooms when taking the bypass since that's technically already built-in to the map.

image

KatoakDR: I think I know what's happening... after I've manually moved with the flying mount through the obstacles to about the midway point of the trip south, I then call `manual_go2(end_room)`

The function `manual_go2` calculates from that spot the shortest distance to the destination. Since there's many more rooms ahead of us to get to the south platform than from where we came from, the `timeto` values of the previous rooms sum up to be the faster route, going back to the north platform because my wayto override says that it's fastest from north-to-south platform starting up there.

What I want to try next is have the entire path calculated up front, then do the manual flying mount movements, and from that point on iterate through the precomputed path to the destination rather than recalculating a shortest path which will be wrong.

KatoakDR:
@ellipsis-dev hide your old comments and review again

MahtraDR:
All good and tested and such?

KatoakDR:
Yes, but that you ask, let me do more testing tonight. It's net new and
won't impact anyone without adding the personal wayto overrides. I've done
different tests over different days but I want to do all the tests in one
go for added confidence. Thanks

MahtraDR:
Your point that it's net new is good. Happy to merge. Let me know when. Now is ok too.

KatoakDR:
> Your point that it's net new is good. Happy to merge. Let me know when. Now is ok too.

We're good to merge, thanks!

PR #7178, opened 2025-05-29
[script] [vanity-pet] fix: no littering match string when try to drop pet in bank

KatoakDR:
```
[vanity-pet]>drop my snaggletoothed boar

A guard steps over to you and says, "No littering in the bank."

...

[vanity-pet: *** No match was found after 15 seconds, dumping info]
```


MahtraDR:
Thank you!

PR #7179, opened 2025-05-29
[script] [combat-trainer] fix: maneuver powershot no longer supports dual loaded bows

KatoakDR:
Since I've been out of the game a few years, I'm not certain when this change occurred, but I've noticed that attempting to perform `maneuver powershot` when you are dual loading a bow will now fail to perform rather than firing 1 of the 2 loaded arrows.

The previous logic would powershot then fire again to empty the second arrow.

Now, since you can't powershot with a dual loaded bow, the script will just always fire. To powershot then set `dual_load: false` in your character's yaml.

Attempting to powershot with a dual loaded bow will get you the message:
```
>maneuver powershot

You cannot make this shot with two <ammo-adjective> <ammo-noun-plural>.
```

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3999,9 +3999,13 @@ class AttackProcess
       waitrt?
 
       if game_state.loaded
-        # If we successfully perform a maneuver then count that as our action.
-        # Otherwise begin aiming for normal shot.
-        if charged_maneuver && game_state.use_charged_maneuver(charged_maneuver)
+        did_shoot = false
+        # If dual loading then cannot do powershot maneuver, just shoot.
+        if game_state.dual_load?
+          shoot_aimed('shoot', game_state)
+          did_shoot = true
+        # Otherwise, try to perform the maneuver.
+        elsif charged_maneuver && game_state.use_charged_maneuver(charged_maneuver)
           # Only react to the powershot flag if we just did a powershot maneuver.
           # This is an added protection in case the flag was tripped by an unrelated
           # action like throwing a weapon. This way we don't try to "stow my throwing hammer"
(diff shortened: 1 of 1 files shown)


vtcifer:
I don't think this needed a code change tbh.  The fix is for end users to pick either dual-load or power shot by setting their yaml appropriately (either don't turn on dual load if you want to use powershot with bows, or remove bows from charged_maneuvers if you want to use dual load).

```yaml
dual_load: false
charged_maneuvers:
  Brawling: palmstrike
  Bow: powershot
  Crossbow: powershot
  Slings: powershot
  Small Edged: cleave
  Large Edged: cleave
  Twohanded Edged: cleave
  Small Blunt: crash
  Large Blunt: crash
  Twohanded Blunt: crash
  Staves: twirl
  Polearms: impale
  Dual Wield: doublestrike
  Shield Usage: rush
```
or
```yaml
dual_load: true
charged_maneuvers:
  Brawling: palmstrike
  # Bow: powershot
  Crossbow: powershot
  Slings: powershot
  Small Edged: cleave
  Large Edged: cleave
  Twohanded Edged: cleave
  Small Blunt: crash
  Large Blunt: crash
  Twohanded Blunt: crash
  Staves: twirl
  Polearms: impale
  Dual Wield: doublestrike
  Shield Usage: rush
```

At best if we want to support preventing end users from using contradictory settings is to turn off dual load and message the user if they leave bows in combat maneuvers. 

There also seems there _might_ have been a defect introduced with this (ongoing discussion in yaml channel).

KatoakDR:
Thanks @vtcifer for the fix! Sorry for causing a hiccup for some folks

PR #7203, opened 2025-07-20
Update carve-lockpicks.lic

Neagley:
Added support for when 5 carves doesn't complete a master or grandmaster lockpick.  Matches on 6th carve error message


MahtraDR:
Thank you!

PR #7209, opened 2025-08-01
Update combat-trainer.lic

Algonem12:
Updates and fixes slivers. Will show how much time is left until next slivers are cast. Much more enjoyable now that is doesn't cast constantly, and casts as indicated.


MahtraDR:
How well tested is this? Have you had folks test who don't have the Moonblade spell? Who want slivers and donn't, and so on? Thanks!

PR #7217, opened 2025-08-31
[script][bescort]shard thief guild update

desertkaz:
Handle the case where password is not required due to a recent visit

----

> [!IMPORTANT]
> Update `enter_thief_guild` in `Bescort` class to conditionally say password based on door response, handling cases where it's not needed.
> 
>   - **Behavior**:
>     - Update `enter_thief_guild` in `Bescort` class to handle cases where password is not required.
>     - Only say password if door slit opens after knocking, indicating a password is needed.
>   - **Misc**:
>     - Modify `enter_thief_guild` to use regex for response matching.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=elanthia-online%2Fdr-scripts&utm_source=github&utm_medium=referral)<sup> for 02f207afc241f62adb56c3b88b554a2ea7cbdca4. You can [customize](https://app.ellipsis.dev/elanthia-online/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/bescort.lic
+++ b/bescort.lic
@@ -2123,7 +2123,7 @@ class Bescort
 
   def enter_thief_guild
     password = @settings.shard_thief_password
-    case DRC.bput('knock',/^You wait, but nothing happens/, /^A moment later, a slit opens in the door at eye level/)
+    case DRC.bput('knock', /^You wait, but nothing happens/, /^A moment later, a slit opens in the door at eye level/)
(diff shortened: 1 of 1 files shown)


MahtraDR:
Could you add us a case for when you get the password wrong? Maybe edit the script with a DRC.message saying "fix yer yaml ye grub! Yer password is wrong!" Or something like that?

PR #7220, opened 2025-09-06
Learned - update to add zero as argument

BinuDR:
Fixed this to not always display all skills you haven't gained any experience in and added the `zero` argument in to display all if you really want that.


MahtraDR:
Thank you!

PR #7300, opened 2026-02-18
Add Charge elemental charge script to charge elemental charge for Warrior Mages

Elysande1218:
Script allows a warrior mage to charge their elemental charge to a specific level in a specific room.

----

> [!IMPORTANT]
> Add `charge-elemental-charge.lic` script for Warrior Mages to automate charging elemental charge to a specified level in a specified room.
> 
>   - **Behavior**:
>     - New script `charge-elemental-charge.lic` for Warrior Mages to charge elemental charge to a specified level in a specified room.
>     - Uses `Summoning` class to automate charging process.
>   - **Classes & Methods**:
>     - `Summoning` class initializes with settings for `elemental_charge_minimum_level` and `elemental_charge_room`.
>     - `initialize` method walks to the charging room and starts charging.
>     - `done_charging?` method checks if the elemental charge is at or above the minimum level.
>     - `charge_elemental_charge` method performs the charging loop until the desired level is reached.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=elanthia-online%2Fdr-scripts&utm_source=github&utm_medium=referral)<sup> for e047e141c05b3a8799d7b8167f38818dc08122a2. You can [customize](https://app.ellipsis.dev/elanthia-online/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/charge-elemental-charge.lic
+++ b/charge-elemental-charge.lic
@@ -1,5 +1,5 @@
 =begin
-  Documentation: 
+  Documentation: https://elanthipedia.play.net/Lich_script_repository#charge-elemental-charge
 =end
 
 class Summoning
@@ -26,3 +26,4 @@ class Summoning
 end
 
 Summoning.new
+
--- a/charge-elemental-charge.lic
+++ b/charge-elemental-charge.lic
@@ -25,5 +25,3 @@ class Summoning
 end
 
 Summoning.new
-
-
--- a/charge-elemental-charge.lic
+++ b/charge-elemental-charge.lic
@@ -1,0 +1,28 @@
+=begin
+  Documentation: 
+=end
+
+class Summoning
+  def initialize
+    @settings = get_settings
+    @ec_minimum_level = @settings.elemental_charge_minimum_level || 6
+    @charging_room = @settings.elemental_charge_room
+    DRCT.walk_to(@charging_room)
+    charge_elemental_charge
+  end
+
+  def done_charging?
+    DRCA.check_elemental_charge >= @ec_minimum_level
+  end
+
+  def charge_elemental_charge
+    until done_charging?
+      #current_charge = DRCA.check_elemental_charge
+      fput('summon admittance')
+      waitrt?
+    end
+    fput('stand')
+  end
+end
+
+Summoning.new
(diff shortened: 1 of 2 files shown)


Elysande1218 (commented):
I removed the line that was causing the failure

MahtraDR:
## Updated description (post-review fixes applied)

This script automates charging elemental charge for Warrior Mages. It walks to a configured room and loops `summon admittance` until the charge reaches a target level.

### Fixes applied

The following issues were identified during review and have been fixed in the latest commit:

1. **Class renamed from `Summoning` to `ChargeElementalCharge`**: The repo already has a `class Summoning` in `summoning.lic`. Using the same name would overwrite that class at load time, breaking the existing summoning training script.

2. **Replaced raw `fput`/`waitrt?` with `DRCS.summon_admittance`**: The original loop used `fput('summon admittance')` followed by `waitrt?`. If the command failed (wrong room, no mana, etc.), there would be no roundtime, so `waitrt?` returned instantly and the loop would spam the game server. `DRCS.summon_admittance` handles errors properly -- it retreats on "too distracted", pauses between attempts, waits for roundtime, and calls `fix_standing`.

3. **Added nil guard for `elemental_charge_room`**: If the setting is not configured in the user's YAML, the script now prints a message and exits instead of silently calling `DRCT.walk_to(nil)` (which returns false and starts summoning in the wrong room).

4. **Removed `fput('stand')` at end**: `DRCS.summon_admittance` already calls `DRC.fix_standing` after each summon, making this redundant.

### YAML settings

```yaml
elemental_charge_minimum_level: 6    # Target charge level (0-11, default: 6)
elemental_charge_room: 1234          # Room ID to walk to for charging (required)
```

### Test coverage

7 specs added in `spec/charge_elemental_charge_spec.rb`:
- Charge level checks (below/at/above minimum)
- Summon loop count (calls `summon_admittance` correct number of times)
- Already-charged skip (no summon calls when charge is sufficient)
- Class naming (confirms `ChargeElementalCharge`, not `Summoning`)
- Nil room guard (returns early with message when room not set)

PR #7318, opened 2026-03-11
fix(athletics): allow rope training to continue after music stops

MahtraDR:
## Summary
- Remove false-positive "dead rope" detection when music stops during climbing rope training
- The "slacken" message was incorrectly treated as rope exhaustion, causing `athletics max` to exit after one session
- The rope isn't tired - it just needs music played again to continue

## Background
When using `;athletics max` with `have_climbing_rope: true`, the script would exit after one climbing session with only ~6 mindstates gained instead of training to 32.

The message "You finish practicing your climbing skill just as your rope begins to slacken" indicates the music stopped, not that the rope is tired for the day. This pattern was incorrectly in the `climbing-dead-rope` flag, causing premature exit.

## Test plan
- [ ] Run `;athletics max` with a climbing rope configured
- [ ] Verify it continues past one session and trains to 32 mindstates
- [ ] Verify it still exits properly when rope is actually exhausted ("You believe you can use it for X minutes per day")

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -140,6 +140,7 @@ class Athletics
     until done_training? || Flags['climbing-dead-rope']
       DRC.fix_standing
       Flags.reset('climbing-finished')
+      DRC.stop_playing
       break unless DRC.play_song?(@settings, @song_list, true, true, true)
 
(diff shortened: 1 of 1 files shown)


AndervolMD:
latest update works great. log posted of it working in the general thread.

PR #7376, opened 2026-04-05
[script][athletics] Low-rank Riverhaven climbing

urbaj-dr:
Since it's been discussed by the community, added support for lower level Riverhaven athletics training. Uses mostly 0-RT movements scattered around town (climbing steps, trees, etc.)

### DIFF
--- a/athletics.lic
+++ b/athletics.lic
@@ -53,7 +53,7 @@ class Athletics
     elsif athletics_location == 'Crossing'
       crossing_athletics
     elsif athletics_location == 'Riverhaven'
-      crossing_athletics if DRSkill.getrank('Athletics') > 140
+      riverhaven_athletics
     elsif athletics_location == 'Shard'
       shard_athletics
     elsif athletics_location == 'Hibarnhvidar'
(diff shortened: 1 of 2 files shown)


MahtraDR:
## Updated description (post-review fixes applied)

This PR adds a dedicated Riverhaven athletics training route for characters with `athletics_town: Riverhaven` (or Riverhaven as hometown). Previously, Riverhaven users at low ranks had **no athletics training** -- the old code only called `crossing_athletics` for rank > 140 and did nothing below that.

### How it works

The new `riverhaven_athletics` method provides a loop of 0-RT climbing movements (stairs, ladders, trees, banks) around Riverhaven. The route is rank-gated for progressive difficulty:

| Rank Range | Route |
|---|---|
| 0-9 | Base route: stairs, ladders, banks |
| 10-299 | Adds tree climbing (rooms 51158, 491) |
| 140-300 | Adds extended loop (rooms 11440, 7640) |
| 300+ | Falls back to `crossing_athletics` for harder obstacles (Segoltha bank, Arthelun rocks) that still teach at high ranks |

The `crossing_athletics` fallback above rank 300 is intentional -- the 0-RT Riverhaven climbs stop teaching at that level, so the script routes to progressively harder Crossing obstacles that still provide XP.

### Changes in this commit

- `athletics.lic`: New `riverhaven_athletics` method with rank-gated route. The dispatcher now calls `riverhaven_athletics` for all Riverhaven users (previously only called `crossing_athletics` for rank > 140).
- `spec/athletics_spec.rb`: 6 new specs covering low rank (skip tree), rank >= 10 (include tree), rank >= 140 (extended route), rank > 300 (crossing fallback). Added `getmodrank` mock support.

### YAML usage

No new settings required. Works with existing configuration:
```yaml
hometown: Riverhaven
# or
athletics_town: Riverhaven
```

PR #7387, opened 2026-04-25
Add consumable check before forging

Algonem12:
Checks consumable oil before starting the forging process to ensure no disruption at the anvil.


MahtraDR:
This PR modifies no files. What did you have in mind?

PR #7404, opened 2026-05-26
[scripts][appraisal].Add support for gem kits

kahlen-tech:
First time doing this so sorry if this is wrong. Added support in appraisal for using the gem kit stackers being sold at FF and AC before that. New yaml settings are:
```
gem_kit_name: gem kit
appraisal_training:
  - kit_pouches

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Gem kit pouch training: pouches from kits are inspected, counted, and automatically routed to appropriate containers with fallback placement.

* **Improvements**
  * Low-value threshold now defaults to 1 when unset.
  * Pouch labeling, counting, and routing logic refined for more consistent put-away into low-value or spare containers.
  * Validation updated to accept kit pouch training as a valid setting.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/elanthia-online/dr-scripts/pull/7404?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

**Comment by MahtraDR** [comment]
## Review: Code Analysis

Nice first PR -- the refactoring direction here is solid. Extracting `classify_pouch` and `pouch_name` out of the inline logic is the right call, and the kit-specific iteration (turn/pull, backwards traversal) is well thought out. The `pouch_low_value` bug fix is also a good catch.

A few items worth addressing before merge:

---

### Must Fix

**1. Add `kit_pouches` to `validate.lic` whitelist**

Line 137 of `validate.lic` has:
```ruby
valid_settings = %w[pouches zills art gear bundle]
```
This needs `kit_pouches` added. Without it, every user who sets `appraisal_training: kit_pouches` will get a warning on every single run, even though the feature works fine. CodeRabbit flagged this too -- it's in a separate file so it wasn't caught in their final pass.

**2. Add nil guard on `spare_gem_pouch_container` in `classify_pouch`**

The empty-pouch branch unconditionally sets `chosen_container = spare_gem_pouch_container`:
```ruby
when /There doesn.t appear to be anything in/
  DRC.message("Empty pouch found, putting it into spare_gem_pouch_container")
  chosen_container = spare_gem_pouch_container
```
If the user hasn't defined `spare_gem_pouch_container` in their YAML, this is nil, and `DRCI.put_away_item?(name, nil)` will likely error or behave unpredictably. The `low_value_gem_pouch_container` gets a `.to_s.empty?` check elsewhere, but `spare_gem_pouch_container` does not. Consider guarding it similarly or falling back to the default container when it's nil/empty.

---

### Should Fix

**3. `@count` side-effect hidden inside `classify_pouch`**

`classify_pouch` mutates the instance variable `@count` to `false` as a side effect when `low_value_gem_pouch_container` is empty:
```ruby
if @count && low_value_gem_pouch_container.to_s.empty?
  DRC.message(...)
  @count = false
end
```
This was already present in the old inline code, but now that it's in a reusable helper, a future maintainer reading `train_appraisal_with_kit_pouches` has no reason to suspect that calling `classify_pouch` permanently disables counting for all subsequent iterations. Consider either moving the `@count` guard to the callers, or at minimum adding a brief comment in `classify_pouch` noting the side effect.

**4. `'Roundtime'` match in `classify_pouch` is a silent no-op path**

In `classify_pouch`, the first `bput` matches against `'Roundtime'` among other patterns. If the appraise returns just `'Roundtime'` (without any of the more specific value/empty matches), the `case result` block falls through and `chosen_container` stays nil. The pouch gets appraised (XP gained) but not classified, so it silently goes to the default container. Not necessarily a bug, but worth being aware of -- if DR ever changes appraise output text, pouches could be silently mis-routed with no log message.

**5. `count_kit_pouches` uses a fragile manual polling loop**

```ruby
timer = Time.now
while Time.now - timer < 10
  line = get?
  if line.nil?
    pause 0.1
    next
  end
```
This raw `put`/`get?` loop with a hardcoded 10-second timeout is unusual compared to the rest of the file, which uses `DRC.bput` throughout. If the game is laggy, the timeout could expire mid-output and undercount pouches. The `pause 0.1` granularity means ~100 poll iterations per timeout window. It works, but it's more brittle than the Lich patterns used elsewhere. Consider whether there's a `DRC.bput`-based approach or at least bumping the timeout to be more forgiving.

---

### Nice to Have

**6. Add specs for the new helper methods**

Four new methods were introduced (`pouch_name`, `classify_pouch`, `count_kit_pouches`, `train_appraisal_with_kit_pouches`) and the existing `train_appraisal_with_pouches` was refactored. Even basic specs would catch regressions, especially for:
- `pouch_name` -- edge cases with nil adjective, empty string adjective, both present
- `classify_pouch` -- each branch (valued, low-value, empty, needs-opening-then-classified)
- `count_kit_pouches` -- parsing numbered rummage output

The refactored `train_appraisal_with_pouches` is especially important to cover since any regression there breaks existing users, not just new kit users.

**7. `gem_pouch_adjective` guard scope mismatch in `train_appraisal_with_pouches`**

The initial appraise-in-hand block is guarded by `if gem_pouch_adjective`, but the loop below always runs using `name` (built by `pouch_name`). If `gem_pouch_adjective` is nil, `pouch_name` strips it and returns just the noun, which works fine. But the guard above suggests the adjective is required for the first appraise while optional for the loop -- a minor inconsistency that predates this PR but is now more visible.

---

Overall the architecture is sound and the DRY refactor is well-motivated. Items 1 and 2 are the actionable blockers; the rest is incremental improvement. Thanks for contributing!

MahtraDR:
Hey mate, if you can resolve the conflicts, I'll merge this. Thanks!

PR #7418, opened 2026-06-01
perf(combat-trainer) [1/2]: reduce per-tick overhead in main loop

MahtraDR:
## Summary
Five low-risk optimizations targeting the 100ms main loop tick:

- **SafetyProcess 1x/tick** -- was running before each of the 8 processes (80 calls/sec down to 10)
- **SpellProcess early return when casting** -- skip 14 sub-check method dispatches that each independently bail on `game_state.casting`
- **Cache npcs per tick** -- `DRRoom.npcs - @ignored_npcs` was recomputed ~27x/tick; now cached in `update_room_npcs`
- **Init-time guild checks for stomp/pounce** -- `DRStats.barbarian?` / `DRStats.ranger?` resolved once at init instead of every tick; also fixes truthy-array bug (`game_state.npcs` -> `game_state.npcs.any?`)
- **Conditional ManipulateProcess** -- skipped entirely when `manipulate_threshold` is not configured; `get_process` converted from positional index to hash lookup

Net: +27 -26 lines, no behavioral changes.

## Test plan
- [ ] Run as Barbarian with war_stomp_on_cooldown -- verify stomp still fires
- [ ] Run as Ranger with pounce_on_cooldown -- verify pounce still fires
- [ ] Run as Empath with manipulate_threshold set -- verify manipulation works
- [ ] Run as any non-empath without manipulate_threshold -- verify no errors
- [ ] Run with offensive spells configured -- verify casting flow unchanged
- [ ] Run with debug flag -- verify process ordering unchanged
- [ ] Verify safety checks (bleeding, lodged items, item recovery) still fire

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unintended standing resets and skipped redundant spell checks while actively casting.

* **Performance Improvements**
  * Cache room NPC lists for faster, more consistent combat decisions.
  * Precompute ability readiness for quicker, reliable stun/pounce activations.

* **Refactor**
  * Reorganized combat process wiring into a keyed map; manipulate behavior only added when configured.


MahtraDR:
@coderabbitai review this PR again in full.

MahtraDR:
@coderabbitai Thanks for the thorough review. Addressing each finding:

**🔴 Critical: nil guard in `update_room_npcs`** -- Agreed, will fix. The merge resolution added the guard to the `npcs` fallback but missed `update_room_npcs`. Will add `(@ignored_npcs || [])` there too.

**🟢 Truthy-array fix** -- Confirmed, this is a real bug fix. `[]` is truthy in Ruby so stomp/pounce could fire with zero NPCs.

**🟡 `get_process('manipulate')` returns nil** -- Intentional, no change needed. The only caller in the entire codebase is `hunting-buddy.lic` calling `get_process('loot')`, which is always present (unconditional in `make_processes`). Nobody calls `get_process('manipulate')` externally. Returning nil from a hash lookup is standard Ruby -- adding a guard here would hide bugs rather than surface them.

**🟡 `@can_stomp`/`@can_pounce` precomputed at init** -- Not a real risk, no change needed. `make_processes` is only called once from `CombatTrainer.initialize` -- there is no reload path anywhere in the codebase. Guild does not change mid-session in DR.

MahtraDR:
@coderabbitai resolve all existing review comments and review this PR as it is now with all commits accounted for.

PR #7426, opened 2026-06-04
fix(script) - Enchant.lic has issues with a couple of hard coded MYs …

BinuDR:
…which interferes with room braziers

Added a new variable: `@brazier_ref` which adds a "my " in front of the brazier name if `@use_own_brazier` is true.

This is only used in calls when dealing with the brazier and the original `@brazier` is preserved and used for calls to things like `DRCC.get_crafting_item()`.

Removed the 2 hard coded "my"s and also the 1 hard coded "brazier".

Also fixed an issue where on resume if using a room brazier, it would still try to call `DRCC.get_crafting_item()` to get your brazier.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced enchanting operations to properly distinguish between external and personal brazier sources.
  * Improved routing of enchantment actions to correctly target the appropriate brazier.
  * Fixed spell casting and crafting actions to work reliably with different brazier types during resume flows.

### DIFF
--- a/enchant.lic
+++ b/enchant.lic
@@ -117,6 +117,8 @@ class Enchant
       end
     end
 
+    @brazier_ref = @use_own_brazier ? "my #{@brazier}" : @brazier
+
     args = parse_args(arg_definitions)
 
     @baseitem = args.base_noun || args.noun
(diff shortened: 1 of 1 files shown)


BinuDR:
This should be good but it could use testing.

PR #7428, opened 2026-06-07
[script][combat-trainer] include ranged check for blowguns in wield_weapon

desertkaz:
The hard return if brawling on 5150 prevented blowgun use.  Leveraging the !is_brawling_ranged? function allows both ranged and melee cases to function correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored weapon selection so ranged brawling attacks now follow the normal weapon-wielding flow, fixing cases where ranged brawling couldn’t select or equip weapons and improving combat reliability.

* **New Features / Configuration**
  * Added explicit aim-filler entries for brawling so ranged brawling uses the same aim-filler behavior as other ranged weapon types, improving consistency in aiming and attack preparation.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -5147,7 +5147,7 @@ class GameState
   # @returns name of the wielded weapon
   def wield_weapon(skill = weapon_skill, equip_main_hand = !offhand?)
     return unless skill
-    return if skill.eql?('Brawling')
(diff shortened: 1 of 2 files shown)


desertkaz:
added Brawling to aim_fillers in base.yaml.  No impact on the typical melee use case.

PR #7430, opened 2026-06-08
[script][combat-trainer] Implement 'Confidence' with restriction for Thieves

vtcifer:
Add 'Confidence' training_ability for thieves.  

Useful when capped on stealth or when stealth stays at high mind states so that they can continue to keep confidence high.

## Summary by CodeRabbit

* **New Features**
  * Added the **“Confidence”** training ability.
  * Availability is restricted to **Thief**-guild characters.

* **Bug Fixes / Behavior Changes**
  * If a non-Thief tries to select **Confidence**, it’s blocked after the first attempt for the rest of the session.
  * The hide/stalk/unhide training flow no longer stops based on Stealth skill rank.

* **Documentation / Validation**
  * Updated validation to allow **“Confidence”** in YAML training-ability configurations.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3351,6 +3351,13 @@ class TrainerProcess
       end
     when /^Stealth$/i
       hide_action(game_state)
+     when /^Confidence$/i
+      if DRStats.guild == 'Thief'
+        hide_action(game_state)
+      else
+        DRC.message("Confidence is only for thieves, removing from training_abilities.")
+        @training_abilities.delete('Confidence')
+      end
     when /^(Ret|Retreat) Stealth$/i
       DRC.retreat
       hide_action(game_state)
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3423,8 +3423,6 @@ class TrainerProcess
   private
 
   def hide_action(game_state)
-    return if DRSkill.getxp('Stealth') >= 34
-
     if DRC.hide?(@hide_type) && !@dont_stalk && !game_state.npcs.empty? && @hide_type == 'hide'
       stalked = DRC.bput('stalk', 'Stalk what',
                          'Try being out of sight',
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3351,7 +3351,7 @@ class TrainerProcess
       end
     when /^Stealth$/i
       hide_action(game_state)
-     when /^Confidence$/i
+    when /^Confidence$/i
       if DRStats.guild == 'Thief'
         hide_action(game_state)
       else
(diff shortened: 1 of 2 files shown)


MahtraDR:
Took a pass through this. Accepting that hiding/stealth raises Confidence, the validate.lic whitelist entry looks correct and well-placed. The combat-trainer.lic change has one significant problem though.

**1. (Major) The feature no-ops in exactly the scenario it targets.** The PR is described as useful "when capped on stealth or when stealth stays at high mind states," but the Confidence branch routes to `hide_action`, which opens with:

```ruby
def hide_action(game_state)
  return if DRSkill.getxp('Stealth') >= 34
```

`getxp >= 34` is mindlocked/capped Stealth, which is the precise condition described. At that point `hide_action` returns immediately, never hides, so no stealth attack/backstab fires and Confidence never gets trained. The branch does nothing in the one situation it was added for.

**2. (Major, consequence of #1) It is behaviorally identical to the existing `Stealth` ability.** Since the new branch just calls `hide_action` unchanged, it duplicates the existing `when /^Stealth$/i` case and inherits the same mindstate gate. Below the cap it adds nothing over `Stealth`; at/above the cap it does nothing. To serve its purpose it needs a hide path that bypasses the Stealth-mindstate early return, e.g. a `hide_action(game_state, ignore_stealth_mindstate: true)` flag or a dedicated method.

**3. (Minor) `Confidence` is missing from `@skill_map`.** Every other simple ability has a `@skill_map` entry. In `check_ability`, `check = @skill_map['Confidence']` is `nil`, so `check = [nil]` and `DRSkill.getxp(nil)` returns `0`, making `expcheck` always true. So Confidence is always eligible (subject only to its cooldown) and silently ignores `combat_training_abilities_target`. For a stat rather than a skill, "always run on cooldown" may be intended, but right now it works by accident. Worth an explicit entry/comment or confirming the intended gating.

**4. (Nit)** `DRStats.guild == 'Thief'` re-runs on every Confidence selection for a Thief. Harmless (the non-Thief path self-removes after one hit) and the guild string matches the convention used elsewhere. No change needed.

Bottom line: #1/#2 are the blocker. As written the ability cannot fulfill its stated purpose because `hide_action` bails out at capped/high Stealth mindstate. The fix is a hide path that skips the `getxp('Stealth') >= 34` guard, plus tidying up #3 while in there.

vtcifer:
Response - 
1/2: I simply removed the guard clause at the beginning of hide_actions, as that guard clause probably shouldn't have been there, as it blocks other times when you'd want to hide (backstab/ambush/poach).

3: Intentionally missing in skill map (not everything in training_abilities has a skill map, only ones that it makes sense for, this one it makes sense to not have anything in skill map)

4: as said, nitpick.

PR #7433, opened 2026-06-10
Adding tinkering into Workorders

belinciadr-ctrl:
This adds the tinkering option to the workorders script. Bronze mechanisms will be automatically crafted for the workorders. Other mechanisms are able to be selected, but the ingot must be provided.

Tinkering has had some updates that incorporate the recent update to DRCC's create_mechanisms.

It also corrects the naming convention in a few of the files that specify shaping_tools or carving_tools and updates to the same naming of the other Professions and uses engineering_tools. This happens in clerk-tools, carve, and taskmaster.

Base.yaml gets the update of mechanism_ingot_type: bronze in order to specify the type of mechanism that will be made for the tinkering workorders.

This change would require a yaml update of changing shaping_tools or carving_tools over to the newly named engineering_tools

## Summary by CodeRabbit

* **New Features**
  * Added support for tinkering workorder type with dedicated crafting workflow.

* **Changes**
  * Updated tool selection logic for crafting disciplines.
  * Enhanced mechanism crafting workflow with improved consolidation and ordering.
  * Added new configuration option for mechanism material type.
  * Modified tool repair behavior to allow excluding specific tools from repair operations.

### DIFF
--- a/carve.lic
+++ b/carve.lic
@@ -35,9 +35,9 @@ class Carve
     @training_spells = @settings.crafting_training_spells
 
     @main_tool = if @type == 'stack' || @type == 'bone'
-                   @settings.carving_tools.find { |item| /\bsaw/i =~ item }
+                   @settings.engineering_tools.find { |item| /\bsaw/i =~ item }
                  else
-                   @settings.carving_tools.find { |item| /\bchisel/i =~ item }
+                   @settings.engineering_tools.find { |item| /\bchisel/i =~ item }
                  end
 
     DRC.wait_for_script_to_complete('buff', ['carve'])
--- a/clerk-tools.lic
+++ b/clerk-tools.lic
@@ -32,7 +32,7 @@ class Clerk
       area = settings.custom_clerk_tools[custom_toolset][:area]
       @belt = nil
     elsif args.toolset == 'engineering'
-      tools = settings.shaping_tools
+      tools = settings.engineering_tools
       area = 'shaping'
       @belt = settings.engineering_belt
     elsif args.toolset == 'outfitting'
--- a/profiles/base.yaml
+++ b/profiles/base.yaml
@@ -1043,6 +1043,7 @@ workorders_materials:
   metal_type: bronze
   stone_type: alabaster
   wood_type: maple
+  mechanism_ingot_type: bronze # Currently the only automatic one for the Tinkering workorder.
 
 # Following settings, including herb_container:, are required to override workorders for remedies
 # container used for raw herbs foraged - used by alchemy.lic script
--- a/taskmaster.lic
+++ b/taskmaster.lic
@@ -250,7 +250,7 @@ class TaskMaster
     elsif base == 'lumber'
       buy_stackables(item, count, volume_per_item, base)
       shape_item(item, count, base, type)
-      repair(@settings.shaping_tools, @settings.engineering_belt)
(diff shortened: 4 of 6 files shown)


MahtraDR:
Thanks for this - the tinkering workorder support and the `cm_tinker` / `create_mechanisms` rework look like a solid, useful addition, and CI is green.

There's one blocking regression that needs to be resolved before this can merge, though: the `shaping_tools` / `carving_tools` -> `engineering_tools` rename.

The code now reads `@settings.engineering_tools` in `carve.lic`, `clerk-tools.lic`, `taskmaster.lic`, and `workorders.lic`, but `engineering_tools` isn't defined anywhere:

- `profiles/base.yaml` still ships `shaping_tools:` and `carving_tools:` (not renamed in this PR)
- `profiles/base-empty.yaml` still has `shaping_tools: []` / `carving_tools: []`
- none of the `profiles/Samples/*` set `engineering_tools`
- it isn't on `main` today either

So after this change `@settings.engineering_tools` is `nil` for every existing user **and** for the shipped defaults out of the box, which means `nil.find { ... }` / passing `nil` tools into `repair` crashes shaping, carving, lumber taskmaster, and the engineering clerk-tools path for everyone - not just people who skip the migration.

Because user setup YAMLs live outside this repo, a hard rename can't be made non-breaking just by updating the samples/base here - existing users would still break until they hand-edit their config. The safe path is a backward-compatible fallback rather than a rename, e.g.:

```ruby
tools = @settings.engineering_tools || @settings.shaping_tools   # shaping/clerk/taskmaster sites
tools = @settings.engineering_tools || @settings.carving_tools   # carve.lic / workorders carving
```

One caveat if you go this route: don't also add an `engineering_tools:` default to `base.yaml`, since a base-provided default would satisfy the `||` and shadow a user's customized `shaping_tools`/`carving_tools`. Keep the old keys as the defaults and let `engineering_tools` be an opt-in override.

Minor (already flagged by CodeRabbit): the `speed` arg (`/s\d+/i`) accepts `s0`/`s99`; worth clamping/validating to the documented s1-s12 before passing to `create_mechanisms`.

Happy to re-review once the tools lookup is made backward compatible.

simtel12:
Question: Why go the route of changing `shaping_tools` to `engineering_tools`; vs. adding a new `tinkering_tools`?

This seems to make the assumption that characters will only be doing one of the activities available - or keep all "Engineering" tools on hand, even if they won't be working that particular discipline in a given moment.

I'm new to the repo and the direction that we want to take here, but personally I'd prefer to split them - then I can use `clerk-tools` to store my shaping tools while I grab my tinkering tools, etc; and not pay the inventory space/weight for extra tools that I don't need at the moment, even though I will want them tomorrow.

Thoughts?

-- Edit --

FYI, `tinker.lic` already uses `tinkering_tools` and `carve.lic` uses `carving_tools`.  I think its worthwhile to decide "everything uses tools based on the skill (ex: engineering_tools), or everything uses tools based on the discipline (ex: tinkering_tools) - and then make it standard across the scripts.

belinciadr-ctrl:
My thoughts and intentions were to put engineering in line with the standard of all the other crafts currently implemented.

The extra weight of the tools you don't need isn't super relevant considering you are pulling them from the clerk, doing the workorder, and then putting them back.

PR #7444, opened 2026-06-23
fix(combat-trainer): audit fixes (12 bugs, 2 footguns, 2 perf) + consolidate specs to one file

MahtraDR:
## Summary

A combat-trainer audit pass with three parts:

1. Fix 12 bugs, 2 footguns, and 2 performance issues in `combat-trainer.lic`.
2. Consolidate the 8 combat-trainer spec files into a single canonical `spec/combat_trainer_spec.rb` (one spec per script).
3. Fix all 22 pre-existing spec failures (every one was test-side; none was a code defect).

Verification: `ruby -c` clean, rubocop clean on both touched files, and `spec/combat_trainer_spec.rb` runs **260 examples, 0 failures**.

## Bugs fixed

- **LootProcess** `@dump_item_count` was permanently decremented every tick a "junk" pile was in the room (instance mutation instead of a per-tick local), so the dump threshold drifted negative over a hunt.
- **SpellProcess#check_health_empath** operator precedence (`A && B || C`) defeated the 30s throttle when `HEAL` is set, issuing a blocking `PERCEIVE HEALTH SELF` every tick while wounded. Parenthesized to `A && (B || C)`.
- **SafetyProcess** `@untendable_counter` incremented on every bleeding tick and was never reset on success, so separate fully-tended bleeding episodes accumulated to a false "couldn't tend, stopping hunt". Now counts only genuine failures and resets when bleeding clears.
- **TrainerProcess#reset_ability** used `[(cooldown - offset), 0].min`, which is always 0, making the `offset` argument dead and forcing a full cooldown. Changed to `.max`.
- **PetProcess** zombie stance regex matched only `a` not `an`, so the default "an offensive stance" never registered and the command re-issued every tick.
- **TrainerProcess** Summon-Domain removal iterated the filtered Hash with `.each` (yielding `[k, v]`), so `delete([k, v])` was a no-op and the ability was never removed for sub-200 Summoning. Uses `.keys.select`.
- **TrainerProcess#hide_action** used `== ('x' || 'y')`, which collapses to the first literal, so the "already stalking" reply skipped the stop-stalk cleanup. Uses `include?`.
- **GameState** stance normalization ran via `Hash#keys` with a block (silently ignored), leaving the legacy-stance migration, missing-stance backfill, and invalid-stance validation all dead. Uses `each_key` (and stops shadowing the `settings` arg).
- **GameState** charged-maneuver cooldowns were seeded under `maneuver.downcase` but read and written with the raw value, so a capitalized maneuver name missed its timer. Keys consistently by downcase.
- **GameState#stomp** lacked a `when` for "What are you trying to attack" and fell into the success branch. Added it.
- **LootProcess#dissected?** discarded its retry recursion result. Returns it.
- **AbilityProcess** `warhorn_cooldown.to_i || 1200` made the default dead (`to_i` never returns nil). Uses `(settings.warhorn_cooldown || 1200).to_i`.

## Footguns

- Wrapped `eval()` of YAML `must_be_true` / `buff_nonspells` conditional strings in a rescue so a malformed config logs and is treated as false instead of aborting the combat loop.
- Bounded the recursion in `check_battle_cries` and `use_charged_maneuver` so a long face-next chain or persistent out-of-range cannot recurse unboundedly.

## Performance

- `check_buffs` issued a fresh `DRRoom.npcs` query (which also ignores `@ignored_npcs`) on the per-tick hot path. Uses the cached `game_state.npcs`.
- `GameState#dance` recomputed the dynamic dance skill every tick. Throttled to a few seconds, which also avoids weapon thrash.

## Spec consolidation

The 7 satellite spec files (`bug_fixes`, `safety`, `warhorn_egg`, `slivers`, `gempouch`, `force_cleanup`, `almanac`) are merged into `spec/combat_trainer_spec.rb`. The scaffolding that conflicted when those files were co-loaded is unified: one `load_lic_class`, one `RSpec.configure`, a reopened `Harness::DRSpells` (instead of a top-level shadow that dropped `active_spells`), a single `UserVars` store, the union of the command stubs, and all combat-trainer classes loaded.

## Spec failures fixed

All 22 were stale test scaffolding, not code defects: matcher arity drift (manipulate, gempouch), builders that bypassed `initialize` or omitted a key (check_slivers `tk_spell`, AbilityProcess `@can_stomp`/`@can_pounce`), and two tests asserting behavior the code had already corrected (the `.any?` stomp guard, and the empath shield-rush gate from #7355).

## Behavior change to note for reviewers

The stance-normalization fix re-enables a previously-dead validation block. Its `exit` on an invalid per-weapon `stances:` entry now actually fires (it never did before because `Hash#keys` ignored the block). This is the intended validation, but it means a malformed `stances:` config will now halt the script instead of being silently tolerated.

## Release Notes

* **Bug Fixes**
  * Made YAML-driven conditional checks fail safely (errors are logged and evaluation is treated as false).
  * Stabilized “junk” loot dumping with consistent per-tick thresholds.
  * Corrected safety/health edge cases, including tendme bleeding tracking and timer gating behavior.
  * Fixed multiple targeting/stance/cooldown normalization issues and hardened charged maneuver retry limits.
  * Ensured stomp correctly returns “not attacking” when the game refuses the action.
* **Performance**
  * Throttled dynamic dance skill recomputation (at most once every 5 seconds).
* **Tests**
  * Consolidated the RSpec suite by removing several older spec files and expanding shared coverage.


MahtraDR:
@coderabbitai review this pr in full taking all commits into account.

MahtraDR:
@coderabbitai review again in full

PR #7452, opened 2026-07-03
sell-loot: fail over between multiple gem-shop clerks

MahtraDR:
## Problem

Some towns staff their gem shop with more than one possible NPC, and there can be a gap where none of them is in the room. When the configured clerk was not present, `which_clerk` returned `nil` and the script fired malformed commands like `ask  for soft pouch` (`To whom are you speaking?`) and `sell my gem to `. This is the nil-clerk "sharp corner" the existing spec had pinned.

## Change

Add `wait_for_clerk`:

- A single named clerk (String) returns immediately -- no behavior change for towns with one clerk.
- A **list** of candidates is checked against `DRRoom.npcs` each attempt, failing over to whichever is present.
- If none is present (including the between-NPC gap), it pauses 60s and retries, up to 5 attempts, then emits a verbose give-up message (room title + id, who was tried, elapsed minutes, attempt count) and returns `nil`.

`sell_gems`, `sell_metals_and_stones`, and `check_spare_pouch` now use it and guard against `nil`: they skip gem-shop sales and let the rest of sell-loot continue, rather than issuing broken commands. `sell_gems` still closes the pouch on the way out.

This is general -- any town can opt into failover by listing multiple names under `gemshop.name`. Fang Cove's `gemshop.name` becomes a list (`Wickett`, `attendant`) as the first town to use it, matching the existing multi-clerk pattern already present elsewhere in `base-town.yaml`.

## Tests

- New `#wait_for_clerk` block: string passthrough, present-candidate selection, give-up-nil, verbose message content, retry/pause count.
- Nil-guard cases for all three call sites.
- Full suite: 75 examples, 0 failures. Rubocop clean on both touched Ruby files. `base-town.yaml` re-parses with the list.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gem-shop selling so it can handle missing or delayed shop clerks more reliably.
  * Selling now safely stops when no clerk is available, preventing failed sale attempts.
  * Added retry behavior when looking for a clerk, with clearer progress and failure messages.
  * Updated the gem-shop clerk setup to support multiple possible clerk names.

### DIFF
--- a/data/base-town.yaml
+++ b/data/base-town.yaml
@@ -1542,7 +1542,9 @@ Fang Cove:
     id: 8390
   gemshop:
     id: 8386
-    name: Wickett
+    name:
+    - Wickett
+    - attendant
   tannery:
     id: 8394
     name: Gorund
--- a/data/base-town.yaml
+++ b/data/base-town.yaml
@@ -1543,8 +1543,8 @@ Fang Cove:
   gemshop:
     id: 8386
     name:
-    - Wickett
-    - attendant
+      - Wickett
+      - attendant
   tannery:
     id: 8394
     name: Gorund
--- a/sell-loot.lic
+++ b/sell-loot.lic
@@ -386,6 +386,10 @@ class SellLoot
     return clerks if clerks.is_a?(String)
 
     candidates = Array(clerks)
+    # No configured clerk (nil or empty list) will never resolve -- bail
+    # immediately rather than burning CLERK_MAX_ATTEMPTS pauses on it.
+    return nil if candidates.empty?
+
     attempt = 0
     loop do
       attempt += 1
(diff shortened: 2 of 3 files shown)


MahtraDR:
@coderabbitai review

MahtraDR:
Thanks @coderabbitai - both nitpicks addressed in fc03727c:

1. **Empty/misconfigured candidate list** - `wait_for_clerk` now short-circuits with `return nil if candidates.empty?` (covers both `nil` and `[]` from `gemshop.name`), so it no longer burns the retry pauses on a condition that can't resolve. Kept the guard central in `wait_for_clerk` rather than duplicating checks in each caller.
2. **Hardcoded 60s minute math** - the give-up message now reports the elapsed wait in seconds, dropping the `waited / 60` assumption entirely.

Added a spec asserting nil/empty lists return immediately without pausing. Suite is 76 examples, 0 failures; rubocop clean.

MahtraDR:
@coderabbitai I've pushed fc03727c addressing both nitpicks - the empty/nil clerk-list short-circuit and the seconds-based give-up message. Could you re-review the latest changes?

MahtraDR:
@coderabbitai I've pushed 58923c2d, which fixes two regressions in the pre-flight loot gate:

1. **Stranded coins** - when `has_loot_to_sell?` was false the whole run bailed via an early return, so coins already on hand were never exchanged/deposited. Selling is now wrapped in `if has_loot_to_sell?` and banking always runs afterward, gated by a new `coins_to_bank?` check (local currency over the keep-on-hand amount, or any foreign currency to exchange).
2. **One missing loot type sinking the run** - all venues are now checked before deciding there is nothing to sell, and banking proceeds regardless.

Added five `#coins_to_bank?` unit tests plus orchestration cases for the no-loot/excess-coins and no-loot/no-coins paths. Suite is 82 examples, 0 failures; rubocop clean. Could you review the latest changes?

PR #7460, opened 2026-07-12
fix(healer): stop vit-transfer runaway that trips afk.lic logout

MahtraDR:
## Problem

Controlled stacking added in #7456 opens up to `MAX_VIT_STACKS` (3) concurrent `transfer <patient> vit quick` streams during vitality healing. Each quick transfer is a **continuous** drain, so three stacked streams pulled roughly 5%/sec of the empath's own vitality -- 100% down to 38% in about 12 seconds.

The VH refill couldn't keep pace (and stalled behind a blocking `waitrt?` in `:awaiting_drain`), and both of the healer's own safety lines sat *below* afk.lic's quit threshold (emergency fall at 30%, afk `health_threshold` typically ~50). Result: afk.lic detected low vitality and logged the empath out mid-heal, before the emergency fall could ever fire.

Observed timeline (real session): three `transfer vit quick` fired within one second while health read 100%, continuous "You feel your vitality transferring" drain, `[afk: approaching low vitality]`, then `[afk: Detected low vitality/spirit health ... Exiting]` at 38%.

## Fix

- **Disable stacking** (`MAX_VIT_STACKS = 1`). A single continuous stream already refills fine and stays controllable; stacking multiplies the drain past what the refill and the safety checks can handle. The headroom-scaling logic is retained (clamped to the cap) for a future safe stacking scheme.
- **Pin the vit floor and emergency-fall lines above afk.lic's `health_threshold`** at init: `emergency = afk_quit + 12`, `floor = afk_quit + 20`, each with the old constants (`30` / `60`) kept as `.max` baselines, and defaulting to afk's own effective `50` when unset. This guarantees the healer stops taking vit -- and falls to break links -- before afk quits, regardless of how the user tuned afk.
- **Don't block while critically low**: in `:awaiting_drain`, if health is already at/below the emergency line, abandon the cast and return so `check_emergency_fall` can break the link on the next unblocked tick (the blocking `waitrt?`/`waitcastrt?` is what let the drain outrun every check).

No behavior change for empaths whose vitality never approaches the floor; the thresholds only engage when a drain starts running away.

## Testing

- `ruby -c healer.lic` -> Syntax OK
- `rubocop healer.lic` -> no offenses

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved healer safety thresholds by adapting them to the configured health limit.
  * Prevented excessive vitality drain during healing.
  * Added emergency safeguards that stop healing and break links before logout risk.
  * Improved recovery behavior when vitality is low, allowing healing to resume only when safe.

### DIFF
--- a/healer.lic
+++ b/healer.lic
@@ -186,11 +186,18 @@ class Healer
 
     # Pin our vitality safety lines above afk.lic's quit threshold so the healer
     # stops taking vit (and falls to break links) before afk logs the character out.
+    # afk.lic reads health_threshold.to_i with no default, so an unset value is 0 and
+    # afk never quits on health -- 50 is the healer's own conservative fallback.
     afk_quit = @settings.health_threshold.to_i
-    afk_quit = 50 if afk_quit <= 0 # afk.lic's own effective default
-    @emergency_vit_threshold = [afk_quit + AFK_EMERGENCY_MARGIN, EMERGENCY_VIT_THRESHOLD].max
-    @self_vit_floor = [afk_quit + AFK_FLOOR_MARGIN, SELF_VIT_FLOOR].max
-    @vit_resume_threshold = [@self_vit_floor + 5, VIT_RESUME_THRESHOLD].max
+    afk_quit = 50 if afk_quit <= 0 # healer fallback (afk does no health quit when unset)
+
+    # Clamp the upper end too: DRStats.health caps at 100, so a high health_threshold
+    # could otherwise push a threshold past what health can reach (floor >= 100 =>
+    # never take vit; resume > 100 => :recovering never completes). Keep them strictly
+    # ordered and reachable: emergency < floor < resume <= 95.
+    @self_vit_floor = [[afk_quit + AFK_FLOOR_MARGIN, SELF_VIT_FLOOR].max, 90].min
+    @emergency_vit_threshold = [[afk_quit + AFK_EMERGENCY_MARGIN, EMERGENCY_VIT_THRESHOLD].max, @self_vit_floor - 5].min
+    @vit_resume_threshold = [[@self_vit_floor + 5, VIT_RESUME_THRESHOLD].max, 95].min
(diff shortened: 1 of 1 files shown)


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai Thanks for the review. Dispositions:

**1. Upper clamp (Major) — fixed** in a59464ae. You're right that a high `health_threshold` (>= 76) pushed `@self_vit_floor` to 100 (healer never takes vit) and `@vit_resume_threshold` past 100 (`:recovering` can soft-stall at full health). I went with sequential clamps that guarantee strict ordering and reachability rather than independent caps:

```ruby
@self_vit_floor          = [[afk_quit + AFK_FLOOR_MARGIN, SELF_VIT_FLOOR].max, 90].min
@emergency_vit_threshold = [[afk_quit + AFK_EMERGENCY_MARGIN, EMERGENCY_VIT_THRESHOLD].max, @self_vit_floor - 5].min
@vit_resume_threshold    = [[@self_vit_floor + 5, VIT_RESUME_THRESHOLD].max, 95].min
```

Verified across `health_threshold` 0..100: always `emergency < floor < resume <= 95`, and the default (afk_quit=50) is unchanged at 62/70/75.

**2. Fallback comment (Minor) — fixed** in a59464ae. Corrected to note afk.lic reads `health_threshold.to_i` with no default (unset => 0 => no health quit), and that 50 is the healer's own conservative fallback rather than afk's default.

**3. Emergency-abandon log throttle (Nitpick) — declining.** The "repeats every 0.5s" premise doesn't hold here: `check_emergency_fall` runs at the top of `process_queue`, before `process_spell_slot`, and unconditionally clears `@spell_task` when below the emergency line, so the `:awaiting_drain` branch can't re-enter that episode -- it logs at most once. And `log()` is debug-gated (`if debug?`), so there's no output at all for normal users. Adding an `@emergency_abandoned` flag would introduce state to reset across the `fall`/`release_spell_slot` lifecycle for no real benefit, so I'd rather not.

Issue #4201, opened 2020-03-30
;trade confused by contract case


GreyhallowDR:
When the script rummages your contract container it gets fooled by a contract case and thinks that you have an outstanding contract. Script stalls when it taps for contract but doesn't find one.

Workaround: Don't have a contract case in your designated contract container.
Possible fix: add a comma after contract on line 777 

    DRC.list_to_array(text).select { |item| item.include?('contract,') }

Not sure if that introduces other problems down the line but it seems to be running fine for me so far.

EDIT: Yes, it totally introduces problems down the road.

Tarhish:
I always stored mine in the pockets of my coat, so I don't think I would've noticed a contract-case related problem.  I think replacing that with
```
    DRC.list_to_array(text).reject { |item| item.include?('contract case') }
                                       .select { |item| item.include?('contract') }
```

might work?

MahtraDR:
Closing as nobody is doing trader-specific coding today.

Issue #3167, opened 2018-10-05
clean-lumber.lic isn't correctly checking if you are using a wood saw


Sarvatt:
just had someone report on lnet that it was trying to use a bone saw even though there is a check for wood/woodcutting saw in the script.

Tarhish:
In sheer coincidence I just saw one of the problems Maltaiche was talking about at the same time -- some of the fputs can fail to trigger right and everything piles up and collapses and end up trying to cut a deed or combine lumber and a branch.

MahtraDR:
Closing as stale.

Issue #1420, opened 2017-02-28
Crafting belts, need more categories


brycole:
I use a carpenters belt for shaping and an engineering belt for carving, but cannot put both belts under engineering. Instead I have to edit the YAML to swap belts when I'm doing carving/shaping.

Solution would be to break engineering belt into separate categories.

-carving_belt
-shaping_belt
-engineering_belt

etc.

MahtraDR:
Closing this. It's a bit awkward to close after so much time has passed, but having polled users on Discord, I couldn't find one who'd use this functionality today...

PR #7464, opened 2026-07-14
Fix contract-case + wood-saw bugs and add carving/shaping/tinkering belts (#4201, #3167, #1420)

MahtraDR:
Revives three issues that were closed as "not planned" years ago but are still valid, plus a lint cleanup on a touched file. Each change is grounded in the current code and covered by adversarial specs.

## #4201 - trade confused by a contract case
`find_contracts` filtered the contract container with `item.include?('contract')`, which also matches "a contract case"; the script then tried to `read` a non-existent contract from it and stalled. Now matches on the item noun (`DRC.get_noun(item) == 'contract'`), keeping "a Trading contract" and rejecting "a contract case". The reporter's original "contract," idea failed because a lone contract has no trailing comma.

## #3167 - clean-lumber used a bone saw
The saw check was dead code: `unless get_crafting_item('saw', ...) != /wood saw/` compared a String/nil to a Regexp (always true), so `unless true` never ran and the first item with noun "saw" (a bone saw qualifies) was used. Extracted `ensure_wood_saw`: fetch a saw, and if the one in hand is not a wood saw (`/wood\w* saw/i` matches woodcutting/wood/woodbutt, not bone), stow it and fetch "wood saw" by name. Per DR item matching, "wood saw" matches woodcutting/wood/woodbutt saws (adjective prefix) but never a bone saw; "woodcutting saw" would miss a plain "wood saw", so the broad noun is deliberate.

## #1420 - split engineering_belt into carving/shaping/tinkering
Carving, shaping, and tinkering use different tool sets (verified on Elanthipedia's Crafting tools page), and dr-scripts already splits `carving_tools`/`shaping_tools`/`tinkering_tools` while forcing a single `engineering_belt`. Adds `carving_belt` / `shaping_belt` / `tinkering_belt`, each falling back to `engineering_belt` so existing configs are untouched (`X_belt || engineering_belt`, the idiom already at restock-shop.lic). Mapping by what each script crafts/delegates to:

- carving: carve, carve-bead, carve-lockpicks
- shaping: shape, arrows (delegates to shape), clean-lumber (prep), fir
- tinkering: tinker, bolts (delegates to tinker)

Forging is intentionally left as one belt: weaponsmithing/armorsmithing/blacksmithing share the same hand tools and forge.lic handles all three.

Also fixes a latent crash this exposes: repair.lic built its toolbelt/toolset lists including `nil` for unset disciplines (carving/shaping/tinkering are usually unset), and `smart_get_gear`/`smart_stow_gear` then called `belt["items"]` / `tools.any?` on nil. Extracted `build_tool_lists`, which skips nils. base.yaml and base-help.yaml document the three new keys.

## trade.lic lint cleanup
Drove trade.lic to zero rubocop offenses. The one non-autocorrectable offense was `Lint/UnreachableLoop` in `turn_in_items` (a select-chain `.each` that unconditionally returned on its first iteration); replaced with `.first` + guard, preserving the single-iteration-then-recurse behaviour.

## Testing
New adversarial specs: `spec/trade_spec.rb`, `spec/clean_lumber_spec.rb`, `spec/repair_spec.rb` (22 examples). All green; touched files are rubocop-clean.

### DIFF
--- a/arrows.lic
+++ b/arrows.lic
@@ -7,7 +7,7 @@ class Arrows
     @settings = get_settings
     @bag = @settings.crafting_container
     @bag_items = @settings.crafting_items_in_container
-    @belt = @settings.engineering_belt
+    @belt = @settings.shaping_belt || @settings.engineering_belt # arrows are a shaping product; shaping_belt falls back to engineering_belt (#1420)
     @hometown = @settings.force_crafting_town || @settings.hometown
     @engineering_room = @settings.engineering_room
     @hometown_data = get_data('crafting')['shaping'][@hometown]
--- a/bolts.lic
+++ b/bolts.lic
@@ -9,7 +9,7 @@ class Bolts
     @worn_trashcan_verb = @settings.worn_trashcan_verb
     @bag = @settings.crafting_container
     @bag_items = @settings.crafting_items_in_container
-    @belt = @settings.engineering_belt
+    @belt = @settings.tinkering_belt || @settings.engineering_belt # bolts are a tinkering product; tinkering_belt falls back to engineering_belt (#1420)
     @hometown = @settings.force_crafting_town || @settings.hometown
     @engineering_room = @settings.engineering_room
     @hometown_data = get_data('crafting')['shaping'][@hometown]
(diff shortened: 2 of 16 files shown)


MahtraDR:
Splitting into one PR per topic (maintainer preference): #7465 (contract-case #4201), #7466 (wood-saw #3167), #7467 (sub-discipline belts #1420). Closing this combined PR.

PR #7465, opened 2026-07-14
trade: match contracts by noun so a contract case is not a contract (#4201)

### LINKED ISSUE #4201: ;trade confused by contract case

MahtraDR:
Resolves #4201

Closed as "not planned" but still valid.

`find_contracts` filtered the contract container with `item.include?('contract')`, which also matches "a contract case"; the script then tried to `read` a non-existent contract from it and stalled. Now matches on the item noun (`DRC.get_noun(item) == 'contract'`), keeping "a Trading contract" and rejecting "a contract case". The reporter's original "contract," idea failed because a lone contract has no trailing comma.

The second commit drives trade.lic to zero rubocop offenses (rubocop -A for indentation; the `Lint/UnreachableLoop` in `turn_in_items` -- a select-chain `.each` that always returned on its first iteration -- becomes `.first` + guard, preserving behaviour).

Testing: new `spec/trade_spec.rb` (6 adversarial examples: contract vs contract-case, case-only, multiple, plural "contracts", closed-container reopen). rubocop-clean.

### DIFF
--- a/spec/trade_spec.rb
+++ b/spec/trade_spec.rb
@@ -1,0 +1,106 @@
+require 'ostruct'
+
+# Load test harness which provides mock game objects
+load File.join(File.dirname(__FILE__), '..', 'test', 'test_harness.rb')
+include Harness
+
+# Extract and eval a class from a .lic file without executing top-level code
+def load_lic_class(filename, class_name)
+  return if Object.const_defined?(class_name)
+
+  filepath = File.join(File.dirname(__FILE__), '..', filename)
+  lines = File.readlines(filepath)
+
+  start_idx = lines.index { |l| l =~ /^class\s+#{class_name}\b/ }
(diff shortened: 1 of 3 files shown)


urbaj-dr:
Looks good, working as intended. No issues from my perspective. Thanks!

urbaj-dr (approved):

PR #7466, opened 2026-07-14
clean-lumber: verify a wood saw is used, not a bone saw (#3167)

### LINKED ISSUE #3167: clean-lumber.lic isn't correctly checking if you are using a wood saw

MahtraDR:
Resolves #3167

Closed as "not planned" but still valid.

The saw check was dead code: `unless get_crafting_item('saw', ...) != /wood saw/` compared a String/nil to a Regexp (always true), so `unless true` never ran and the first item whose noun is "saw" (a bone saw qualifies) was used to cut lumber. Extracted `ensure_wood_saw`: fetch a saw, and if the one in hand is not a wood saw (`/wood\w* saw/i` matches woodcutting/wood/woodbutt, not bone), stow it and fetch "wood saw" by name. Per DR item matching, "wood saw" matches woodcutting/wood/woodbutt saws (adjective prefix) but never a bone saw; "woodcutting saw" would miss a plain "wood saw", so the broad noun is deliberate.

Testing: new `spec/clean_lumber_spec.rb` (7 adversarial examples: each wood-saw variant incl. off-hand, bone-saw swap, exit-when-none, no-stow-on-empty-hands). rubocop-clean.


desertkaz:
Crafted saws unfortunately will not work with your `get wood saw` logic as the adjective is set by the crafted template, not material type it is used for.  
```
Bovi: a damite serrated wood saw
Venik: a damite serrated bone saw
```
are both "serrated saw"
```
>get ser saw
You get a damite serrated wood saw from inside your black backpack.
```

```
>;eq echo GameObj.right_hand.name
[exec1: serrated bone saw]
>;eq echo GameObj.left_hand.name
[exec1: serrated wood saw]
```

PR #7468, opened 2026-07-15
Batch steel forging by ingot capacity in restock-shop.lic

bmcneff:
Problem
`restock_steel `sized a single steel ingot for the entire list of missing steel items via `makesteel`. `makesteel` produces a raw ingot of `count × 10` volume (`makesteel 1 h` → 10, `makesteel 21 h` → 210), and 210 is the game's hard cap on ingot volume (count 21). `restock_steel `always refines the ingot, and refining trades volume for quality: a 210 ingot yields 168 with the steel-refining tech, or 105 without it. So the metal actually available to forge from is the refined ingot — at most 168 volume for a tech'd character. When the combined recipe volume of the missing items exceeds that refined capacity (or the derived count exceeds `makesteel`'s cap of 21), forging fails partway through the list.
Change
`restock_steel` now packs the missing steel items into batches whose total recipe volume stays within one steel ingot (168, matching `makesteel 21 h refine`), and forges each batch from its own freshly-made ingot until the list is exhausted.

Added `batch_by_ingot_volume `— greedy packing that closes a batch before any item would push it past the cap. An item whose own volume exceeds one ingot is skipped with a warning instead of failing mid-forge (currently none exist, highest is 120, but future-proofs).
Added `forge_steel_batch `— runs the original per-batch flow (`makesteel [count] refine` → smith + sell each item → dispose leftover), with count still derived by the existing `ceil(volume × 1.25 / 10)` formula.
Added `dispose_leftover_ingot `— moved leftover-ingot disposal to run per batch so ingots from separate batches don't accumulate in the crafting container. workorders still runs once at the end.
Added `MAX_STEEL_INGOT_VOLUME = 168` constant, overridable per-character via an optional `max_steel_ingot_volume` YAML setting. Note: `makesteel`'s count caps at 21, producing a raw 210-volume ingot; `restock_steel` always refines it, which raises quality but reduces usable volume to 168. Since this path always refines, 168 is the effective cap (it would be 210 if refine were dropped, or if further combining logic is added in the future).
Refactored the per-item volume lookup into a `volume_of` helper; `total_volume_of` now reuses it (logic unchanged).

Scope / impact

Steel only. The `restock_burlap`, `restock_linen`, `restock_silk`, and `restock_gryphon` handlers and the `restock_surfaces` dispatcher are unchanged; `total_volume_of` is behavior-preserving, so cloth/leather volume math is identical.
`ruby -c` passes.

Note for reviewers
Leftover-ingot disposal now runs once per batch rather than once per invocation. If a batch's volume lands on an exact multiple that leaves no remnant, `get_crafting_item('steel ingot', ...)` could come up empty — same call the original used, just exercised more often.

### DIFF
--- a/restock-shop.lic
+++ b/restock-shop.lic
@@ -3,6 +3,12 @@
 =end
 
 class ShopRestock
+  # Maximum recipe volume that fits within a single steel ingot.
+  # A `makesteel 21 h refine` ingot holds 168 volume, which is also the
+  # largest ingot makesteel can produce (its count arg is capped at 21).
+  # Override per-character with a `max_steel_ingot_volume` YAML setting.
+  MAX_STEEL_INGOT_VOLUME = 168
+
   def initialize
     @settings = get_settings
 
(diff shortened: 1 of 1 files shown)


Kiriawen:
Each ingot is 10 volume.  So a 21 is 210 volume, not 168.  168 is after refining.

210 volume is the hard limit in game

Also, 168 is with the tech.  without you get 105 volume

bmcneff:
> Each ingot is 10 volume. So a 21 is 210 volume, not 168. 168 is after refining.
> 
> 210 volume is the hard limit in game
> 
> Also, 168 is with the tech. without you get 105 volume

Correct. And restock-shop hardcodes refining. So one of the following solutions should be used:
1. A maximum upper limit of 105 volume should be hardcoded which will face an issue with falling short of the maximum volume needed for a single item (120)
2. A configurable upper limit defaulting to 168 should be used, or
3. Logic should be included to support multiple runs of makesteel until a 210 refined ingot is achieved, along with batching.

I chose to implement option 2, with option 3 something I may consider adding in the future.

bmcneff:
Updated the Problem description and the MAX_STEEL_INGOT_VOLUME constant description to more accurately reflect the ingot size cap and refining loss mechanics.

bmcneff:
Will hold off on submitting this for the time being.

PR #7471, opened 2026-07-18
spec: make the full RSpec suite run green in one process

MahtraDR:
## Summary

Running `rspec` with no args executed only ~44 of the ~2048 collected examples and was not reliably green. This PR makes the whole suite run green in a single process, consolidates duplicated test scaffolding, and fixes the pre-existing failures that the early abort had been hiding.

Before: `rspec` -> 44 examples executed, process exits mid-run.
After: `rspec` -> 2048 examples, 0 failures. Every spec also passes individually, and every touched file is rubocop-clean.

## Root causes fixed

1. **Hostile process-global hooks.** Nine specs registered `RSpec.configure { config.before(:each) { reset_data ... } }` at file-load time. RSpec runs a config-level `before` hook *after* a group's own `before` when the `configure` block is evaluated later than the group is defined (which is exactly what happens as spec files load alphabetically). A later spec's `reset_data` therefore ran last and clobbered the per-example world (guild, settings, hands, room) that an earlier spec had set in its own `before`; astrology's `#initialize` then hit its Moon-Mage guard and called a bare `exit`, killing the whole process. The single `reset_data` hook now lives in `spec_helper` (loaded first via a new `.rspec` `--require spec_helper`), so it always registers/runs before every group hook. Spec-specific world setup that lived in those global hooks moved into each spec's own describe-scoped `before`.

2. **Early aborts from leaked exits.** forge's `#swap_tool` reached `verify_tool_in_hand` -> `cleanup_and_exit` during real (non-dry) runs, leaking a `SystemExit` that terminated the process (only 78 of 108 forge examples ever ran, even alone). The swap_tool examples now neutralize the post-swap verification.

3. **Cross-spec stub pollution.** Specs each redefined the same top-level game stubs (`module DRC/DRCI/GameObj/Flags/DRSkill/Script/XMLData`, `def fput/echo/get/_respond/start_script/stop_script/sitting?/...`, and a copy of `load_lic_class` in 21 files) with conflicting behavior; whichever loaded last won for the whole process. The shared game-object mocks and lifecycle seams are consolidated into `test/test_harness.rb` (DRSkill xp/modrank, DRSpells known_spells, XMLData room_title, EquipmentManager, Script current/exists?, `_respond`/`start_script`/`stop_script`/`move`/`waitfor`/`checkname`/`sitting?`/`stunned?`/`bleeding?`), and `load_lic_class` now lives once in `spec_helper`. Specs rely on the shared harness plus per-example `allow()` stubs.

## Pre-existing genuine failures fixed (were hidden by the abort)

- **sew (3):** finish-message format drift; specs expected an old string.
- **forge (6):** logbook name (`forging` not `engineering`), sentry-block message text, and the recipe-book prep flow that now runs before ingot fetch.
- **enchant (3):** `@brazier_ref` (derived in `initialize`, which these specs bypass) is now set in the builder.
- **healer (1):** the "thin headroom" vit-transfer test now pins `@self_vit_floor` instead of assuming health 65 is above the computed floor.

## Testing

- `RBENV_VERSION=4.0.1 rspec` -> `2048 examples, 0 failures` (single process)
- Each `spec/*_spec.rb` run alone -> green
- `rubocop --no-server` on every touched file -> no offenses; ASCII only

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected forging completion logs and related confirmation messages.
  - Improved guidance shown when access or funds prevent entering a private forge.
  - Updated sewing completion messages for consistent formatting.
  - Preserved text normalization for curly apostrophes and em dashes.

- **Tests**
  - Improved test isolation and shared setup across gameplay features.
  - Added coverage for script cleanup, equipment handling, room state, and character state.
  - Strengthened edge-case coverage for puzzle, healing, work order, and item interactions.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,1 @@
+--require spec_helper
--- a/spec/astrology_spec.rb
+++ b/spec/astrology_spec.rb
@@ -2,32 +2,7 @@
 
 require 'ostruct'
 
-# Load test harness which provides mock game objects
-load File.join(File.dirname(__FILE__), '..', 'test', 'test_harness.rb')
-include Harness
-
-# Extract and eval a class from a .lic file without executing top-level code
-def load_lic_class(filename, class_name)
-  return if Object.const_defined?(class_name)
-
-  filepath = File.join(File.dirname(__FILE__), '..', filename)
-  lines = File.readlines(filepath)
-
-  start_idx = lines.index { |l| l =~ /^class\s+#{class_name}\b/ }
-  raise "Could not find 'class #{class_name}' in #{filename}" unless start_idx
-
-  end_idx = nil
-  (start_idx + 1...lines.size).each do |i|
-    if lines[i] =~ /^end\s*$/
-      end_idx = i
-      break
-    end
-  end
-  raise "Could not find matching end for 'class #{class_name}' in #{filename}" unless end_idx
-
-  class_source = lines[start_idx..end_idx].join
-  eval(class_source, TOPLEVEL_BINDING, filepath, start_idx + 1)
-end
+require_relative 'spec_helper'
 
 # Define stub modules only if not already defined
 module DRC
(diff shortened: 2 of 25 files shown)


MahtraDR:
@coderabbitai review in full

PR #7474, opened 2026-07-19
spec: centralize commons command-module stubs in the test harness

MahtraDR:
## Summary

Follow-up to #7471 addressing CodeRabbit's "centralize global module stubs" review comment.

Nearly every spec defined its own top-level stub modules for the commons command layer (`DRC`, `DRCI`, `DRCC`, `DRCM`, `DRCT`, `DRCH`, `DRCA`, `DRCS`, `DRCMM`, `DRCTH`) and `Lich`. Because the whole suite runs in one process, these duplicate definitions were order-dependent: modules guarded with `unless defined?` kept only whichever spec loaded first, and unguarded ones overwrote each other by load order — so a spec could silently run against another spec's game doubles.

This moves the union of those command-module methods into a single set of stubs in `test/test_harness.rb` with neutral, consistent defaults, and removes the per-spec definitions. There is now exactly **one** definition of each command module, so load order can no longer matter.

## Approach

- Unified stubs live in `test/test_harness.rb`, using consistent default conventions:
  - "is it present / did it happen" predicates (`in_hands?`, `exists?`, `has_tendable_bleeders?`, ...) default to `false`
  - "did the action succeed" predicates (`get_item?`, `put_away_item?`, `walk_to`, `cast_spell?`, ...) default to `true`
  - list/count accessors (`get_item_list`, `count_items_in_container`, `get_total_wealth`, ...) default to `[]` / `0` / `{}`
  - everything else is an inert `nil`-returning seam
- Specs keep overriding the handful of returns they assert on with per-example `allow(...).to receive(...)`.
- Two specs (enchant, plantheal) drove their doubles through spec-local `$mock_*` globals read by their now-removed base stubs; the one affected example is converted to a per-example `allow()` and the dead globals dropped.
- `UserVars` is intentionally **left per-spec**: it is per-script configuration (each script reads different keys with script-specific meanings), not a shared game double, so a single unified stub would not make sense.

Net **-708 lines** (union added once to the harness; ~19 per-spec copies removed).

## Testing

- `RBENV_VERSION=4.0.1 rspec` (full suite, one process) -> `2048 examples, 0 failures`
- Green under `--order random` (multiple seeds) -> load-order independent
- Every `spec/*_spec.rb` run alone -> green
- `rubocop --no-server` on all touched files -> no offenses

## Summary by CodeRabbit

* **Tests**
  * Centralized shared test support for common game interactions and messaging.
  * Removed duplicated test-only setup across numerous script test suites.
  * Improved test harness compatibility with equipment, room, user settings, health, and spell scenarios.
  * Updated tests to use shared helpers and more realistic runtime expectations.


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai Thanks, but I'm going to skip this one — I don't think it's a valid concern here.

The method under test, `PlantHeal#wait_for_passive_healing`, only ever reads `['score']`:

```ruby
until DRCH.check_health['score'].zero?
```

It never touches `bleeders`, `poisoned`, or `diseased`, so `and_return('score' => 5)` is a correct, minimal stub for exactly the path this example exercises (score stays 5 → timeout → healme fallback). The `NoMethodError` scenario can't occur for the code as written.

It's also consistent with the other two `check_health` stubs in the same `#wait_for_passive_healing` block, which use the identical score-only shape:

```ruby
{ 'score' => call_count >= 3 ? 0 : 5 }   # ~line 269
{ 'score' => call_count >= 2 ? 0 : 5 }   # ~line 291
```

Expanding just line 278 to the full health hash would make it inconsistent with its neighbors for no behavioral gain, so I'd rather leave the score-only convention as-is throughout this block.

PR #7484, opened 2026-07-25
Fix loot loss: transfer-items coins abort + droughtmans empty-only trash

MahtraDR:
## Problem

Cashing in a Droughtman's prize package could throw the loot away with the wrapper. Two independent defects combined:

1. **`transfer-items` aborts on coins.** `transfer_items` treats a `false` return from `DRCI.get_item` as fatal (`exit`). Coins auto-deposit into the coin purse on GET and never reach the hands, so `get_item` -> `in_hands?` reports **false** even though the pickup worked. The existing `if item =~ /coins?/i; next` guard sat *inside* the get-success branch, so it was dead code for coins. Result: the transfer exits the moment it hits coins (which nearly every package contains), abandoning every item after them.

   ```
   [transfer-items]>get my coins from my package
   You pick up 5 copper Lirums.
   Unable to get coins from package.
   --- Lich: transfer-items has exited.
   ```

2. **`droughtmans` trashes a non-empty package.** `package` ran the transfer then unconditionally `DRCI.dispose_trash('package', ...)`. The destination was hard-coded to `canvas sack`, which `get_sack` only fetches in the *not-already-in-maze* branch. Start the bot inside the maze (or lose the sack) and the loot is put back into the package, which is then trashed with everything still inside.

## Fix

**`transfer-items.lic`**
- Move the coins check *ahead of* the fatal get: coins are pulled into the purse and skipped without ever hitting the `exit`, so items after them still transfer. Genuine un-gettable items (hands full) still `exit` deliberately, unchanged.
- Full YARD docs added (the class had none).

**`droughtmans.lic`**
- `dispose_empty_package`: trash the wrapper **only once it is verifiably empty**; otherwise keep the package and warn. A `nil`/unreadable contents read is treated as not-empty, so a prize is never discarded.
- Loot destination is now configurable via a `droughtmans_loot_container` setting (default `canvas sack`); the dwarf sack is only fetched for the default container (`fetch_loot_container_if_needed`).
- `leave_winners_circle` extracted for a clean, testable `package` flow.

## Tests

- New `spec/transfer_items_spec.rb`: coins-abort regression (incl. the `get_item -> false` case), coins-then-item, destination-too-small fallback, non-coin fatal exit, trash routing, noun filter, "lot of other stuff" pagination, empty source, and `move_item`/`trash_item` units.
- Extended `spec/droughtmans_spec.rb`: empty/non-empty/nil-read trash guard, sack-gating, config default/override, winner's-circle exit.

## Test-infra: centralize `UserVars`

While adding the droughtmans config spec, four specs (astrology, combat_trainer, smoke, droughtmans) each hand-rolled a top-level `UserVars` class, and `researcher_spec` defined none -- it silently relied on `combat_trainer_spec`'s being loaded first (it could not run in isolation), the exact load-order fragility `spec_helper` warns about.

- Added one canonical `Harness::UserVars`: a dynamic store (unset getter -> nil, setter stores, `_reset` cleared before each example via `reset_data`).
- Removed the bespoke per-spec definitions; reduced `combat_trainer_spec` to a small reopen adding only its `moons` domain default on top of the shared store.
- Every spec now runs standalone (researcher: 177, smoke: 113, astrology: 179, ...).

## Verification

Full suite: **2225 examples, 0 failures**. `rubocop` clean on all touched files.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring where Droughtmans prize loot is placed.
  * Prize wrappers are now preserved when contents remain or cannot be verified.
  * Added safer handling when leaving the winner’s circle.

* **Bug Fixes**
  * Improved coin transfers so failed coin retrieval does not interrupt other item transfers.
  * Added safeguards for full hands, unavailable items, rejected destinations, and empty containers.

* **Documentation**
  * Expanded command and workflow documentation for item transfers.

* **Tests**
  * Added comprehensive coverage for prize packaging, loot containers, item transfers, and test configuration.

### DIFF
--- a/droughtmans.lic
+++ b/droughtmans.lic
@@ -499,7 +499,9 @@ class Droughtmans
   # Trash the prize wrapper only when the loot transfer emptied it. If any loot
   # remains (no/full loot container, an item that would not fit), the package is
   # kept instead so a prize is never thrown away. An unreadable package (nil
-  # contents) is treated as "not verified empty" and is likewise kept.
+  # contents) is treated as "not verified empty" and is likewise kept. The kept
+  # package is only re-stowed when still in hand -- by this point stow_hands has
+  # usually already put it away, so stowing again would just echo "Stow what?".
   # @return [void]
   def dispose_empty_package
     contents = DRCI.get_item_list('package', 'look')
@@ -507,7 +509,7 @@ class Droughtmans
       DRCI.dispose_trash('package', @worn_trashcan, @worn_trashcan_verb)
     else
       DRC.message("droughtmans: package still holds loot (check your #{@loot_container}) - keeping it instead of trashing")
-      DRCI.put_away_item?('package')
+      DRCI.put_away_item?('package') if DRCI.in_hands?('package')
     end
   end
 
--- a/droughtmans.lic
+++ b/droughtmans.lic
@@ -110,6 +110,10 @@ class Droughtmans
     @friends = settings.hunting_buddies
     @worn_trashcan = settings.worn_trashcan
     @worn_trashcan_verb = settings.worn_trashcan_verb
+    # Where won loot is deposited when cashing in a package. Defaults to the
+    # canvas sack fetched from the compound dwarf; override with any carried
+    # container (e.g. a worn backpack) via the droughtmans_loot_container setting.
+    @loot_container = settings.droughtmans_loot_container || 'canvas sack'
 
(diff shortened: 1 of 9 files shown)


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai review in full

PR #7486, opened 2026-07-25
droughtmans: pace movement, survive dark colored-door rooms, wand before rope

MahtraDR:
Follow-up to #7484 (merged). Three robustness fixes for the autonomous maze solver, each driven by an in-game report and covered by specs. Opt-in and non-breaking: default behavior is unchanged unless the new setting is used.

## 1. Movement pacing (stop falling every few rooms)

Maze steps carry near-zero roundtime, so `waitrt?` returns immediately and the solver fires the next `go` at once. It moves fast enough to stumble and fall every few rooms -- especially bad while carrying the key.

- New setting **`droughtmans_step_delay`** (seconds, default `0` = historical un-throttled behavior).
- New `throttle_movement` helper pauses after each *genuinely successful* step (failed/retried moves are not paced).

## 2. Dark colored-door rooms no longer hang 15s

`handle_doors` matched only `Obvious` / `You can't go there` / `You smash your nose`. A `go <door>` issued in a pitch-dark room returns an unmatched line, so `bput` blocked for its **full 15-second timeout** -- during which a rival waved a wand, froze the character, and stole the dropped golden key.

- Match `It's pitch dark and you can't see a thing`, `relight_room` (shake wand), and skip the tick; the door is retried once the room is lit.
- Extracts `relight_room`, now shared with `handle_dark_room` (DRY).

## 3. Wand rivals BEFORE pulling the rope

Pulling the rope drops the key on the floor. `handle_key_or_search` pulled it while room NPCs were still unfrozen, so a mage snatched the key before we could grab it:

```
You give a rope a sharp tug.  A golden key falls to the floor with a loud CLANK.
Cackling with glee, the Rakash mage snatches up the golden key! ... rushing away
get golden key
What were you referring to?
```

- Call `check_for_npcs` (freeze everyone present) immediately before `pull_rope`. No-ops when the room is empty.

## Tests

New specs cover: `throttle_movement` (positive/zero/negative delay); `do_move` end-to-end pacing (paces a success, not a failure); `relight_room` + `handle_dark_room`; `handle_doors` (obvious traversal, dark-room relight/skip, refused entry, and all three guard branches -- wandercounter threshold, no-door, same-door anti-repeat); and the wand-before-rope ordering (rivals frozen before the pull, no wand when there is no rope, and nemesis short-circuit).

Full suite: **2247 examples, 0 failures**. `rubocop` clean on touched files.

### DIFF
--- a/droughtmans.lic
+++ b/droughtmans.lic
@@ -114,6 +114,12 @@ class Droughtmans
     # canvas sack fetched from the compound dwarf; override with any carried
     # container (e.g. a worn backpack) via the droughtmans_loot_container setting.
     @loot_container = settings.droughtmans_loot_container || 'canvas sack'
+    # Optional pause (seconds) after each successful step. Movement is otherwise
+    # throttled only by roundtime, which in the maze is often near-zero, so the
+    # solver can move fast enough to stumble/fall every few rooms. Set
+    # droughtmans_step_delay to a small value (e.g. 1) to pace it; 0 keeps the
+    # historical un-throttled behavior.
+    @step_delay = (settings.droughtmans_step_delay || 0).to_f
 
     # Randomize the wall-follow hand so repeated runs explore different paths.
     @current_direction_map = [CLOCKWISE_MAP, COUNTER_CLOCKWISE_MAP].sample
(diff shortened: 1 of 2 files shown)


MahtraDR:
@coderabbitai review in full

Issue #1469, opened 2017-03-12
dispose() does not handle items too big to fit into buckets/bins/etc


skyleronken:
dispose() should appropriately drop items that dont fit into buckets/bins/etc rather than keep them in hand.

skyleronken:
[08:58] >put table in buck

[08:58] You're kidding, right?  It's a rough table!  It doesn't belong there!

[08:58] >drop tab

[08:58] You drop a rough maple table.

PR #7487, opened 2026-07-25
tendme: guard lodged-item disposal against trashing your own ammo

MahtraDR:
Script-side defense-in-depth for the lich-5 healing bug fixed in [elanthia-online/lich-5#1469](https://github.com/elanthia-online/lich-5/pull/1469).

## Problem

Tending out a lodged item (crossbow bolt, arrow, ...) can destroy the character's **own ammunition**. `DRCH.bind_wound` disposes the dislodged item via `dispose_trash`, whose fallback is a blind `get <item>` that reaches into worn containers. Normally the tended-out item is in a free hand, so the right item is trashed. But if it has left the hand before disposal runs, that GET matches a same-named item in a rucksack/quiver.

Real reproduction (Droughtman's Maze crossbow trap): the maze yanked the character out (`the floor grating underfoot swings open... refreshed and reattired at...`) between the tend removing the bolt and the disposal, clearing the hands:

```
tend my abdomen
You skillfully remove the crossbow bolt from your abdomen ...
(ejected to the Recovery Room)
get my crossbow bolt
You get some crossbow bolts from inside your fighter's rucksack.   <- own ammo
put my crossbow bolt in bucket
You drop some crossbow bolts in a bucket of viscous gloop.          <- destroyed
```

## Why the guard lives here

The destructive `get` -> `put in bucket` is **atomic inside `DRCH.bind_wound`**, and tendme delegates entirely to that method:

```ruby
.each { |body_part| tended_wounds = DRCH.bind_wound(body_part) || tended_wounds }
```

By the time control returns to tendme the ammo is already in the gloop (which destroys it). tendme cannot intercept mid-call, so to guard it must own the tend command.

## Fix

Add `tend_wound_safely` -- a guarded copy of `bind_wound` that disposes a dislodged item **only while it is in hand** -- and point `bind_open_wounds?` at it. When the item is in hand (normal), behavior is unchanged; when it is gone, disposal is skipped instead of grabbing a like-named item.

**This is a temporary shim.** It is marked in-code to be reverted to `DRCH.bind_wound` (and `tend_wound_safely` deleted) once #1469 ships to players -- at which point the commons method is the single source of truth.

## Tests

New `spec/tendme_spec.rb` (13 examples): terminal results (success / already-tended / fumble / too-injured), person passthrough, and the dislodge cases -- in-hand disposal + re-tend, the **not-in-hand regression** (no grab, no dispose), nil in-hands, exact-item in-hands check, `"some"`-quantified items, multi-item recursion, the no-removable-item (clay fragment) line, and person preservation across recursion. Adds `DRCH::TEND_*` pattern constants to the test harness (mirroring `common-healing-data.rb`).

Full suite: **2240 examples, 0 failures**. `rubocop` clean on touched files.

## Summary by CodeRabbit

* **New Features**
  * Improved wound tending to safely handle dislodged items and retry treatment automatically.
  * Supports tending on behalf of a specified person.
  * Correctly reports successful, already-treated, and failed tending outcomes.

* **Bug Fixes**
  * Prevents disposal of dislodged items unless they are confirmed to be in hand.

* **Tests**
  * Added coverage for repeated dislodges, failure states, item handling, and person forwarding.

### DIFF
--- a/spec/tendme_spec.rb
+++ b/spec/tendme_spec.rb
@@ -1,0 +1,128 @@
+require_relative 'spec_helper'
+
+# TendMe#initialize runs the full tend loop (pausing scripts, checking health,
+# looping), so we extract the class with load_lic_class and exercise the guarded
+# tend on a bare-allocated instance (TendMe.allocate). The game layer (DRC/DRCI/
+# DRCH/get_settings/waitrt?) comes from the shared test harness; individual
+# examples override return values with allow(...). Every example reads
+# top-to-bottom (DAMP).
+#
+# The focus is tend_wound_safely: a guarded copy of DRCH.bind_wound that only
+# disposes a dislodged item while it is in hand, so a maze eject (which clears
+# your hands the instant a bolt drops) can never make dispose_trash's blind GET
+# grab and trash the character's own like-named ammunition.
+load_lic_class('tendme.lic', 'TendMe')
+
+RSpec.describe TendMe do
+  let(:tender) { TendMe.allocate }
+
+  describe '#tend_wound_safely' do
+    let(:removed_abdomen) do
+      'You skillfully remove the crossbow bolt from your abdomen leaving the wound no worse than it was before.'
+    end
+    let(:tended) { 'That area is not bleeding.' }
+
+    # -- Terminal (non-dislodge) results ------------------------------------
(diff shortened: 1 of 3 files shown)


MahtraDR:
@coderabbitai review in full

PR #7492, opened 2026-07-27
droughtmans: rewrite as a dead-reckoning grid solver

MahtraDR:
## What

Rewrites `droughtmans.lic` from the wall-following solver into a **dead-reckoning grid solver**, merging the strongest ideas from two independent community rewrites while keeping the script upstream-clean (class stays `Droughtmans`, invocation stays `;droughtmans`).

## Why

The wall-follower discards the wand's dowse vision and can orbit the same square of rooms until the timer runs out. Two community forks solved this in very different ways; this PR takes the best of each:

- A **dead-reckoning grid** approach (visited-aware wandering, event-driven dowse-and-steer, quadrant rotation, white-door memory, keyholder-only wanding, robust looting).
- A **graph-solver race build** whose orthogonal robustness ideas are worth keeping, but whose core graph solver was deliberately **not** adopted: it builds edges on dead-reckoned coordinates that the maze's loops/drags/WHOOSH/wall-shifts corrupt, so its shortest-path advantage collapses exactly when the maze shuffles (its own logs show repeated timeouts). Dead-reckoning degrades gracefully instead.

## Highlights

**Navigation**
- Dead-reckons a grid position and steers toward unvisited rooms (no more orbiting).
- Rotates between the four quadrants via colored doors, preferring a different color than the one entered through; green doors used to get unstuck within a quadrant.
- Dowsing is now event-driven and actually **steers movement** toward the key/center instead of being thrown away; it also reads which quadrant we're in.
- Remembers the white door by grid position and returns straight to it with the key.

**Robustness**
- Recover a dropped key, including a drag-aware backtrack when a current sweeps us off it.
- Re-freeze maze mobs as the structural last act before every move (an unfrozen mob ambushes anyone leaving).
- Pace movement while carrying the key (a trip drops it — the single biggest cause of lost keys).
- Tend trap wounds by the specific bleeding body part, falling back to the `tendme` script.
- 60s no-move watchdog, active STAND self-thaw probe, stun waiting, dark-room light tracking.
- Robust package looting: coins + items with a **never-trash-a-non-empty-package** guarantee, and pass-is-money stow safety (passes are never routed through a droppable container).

**Player interaction**
- Strictly **keyholder-only** wanding (whitelist from announcements / dowse visions / LOOK). Never wands `hunting_buddies` or `droughtmans:no_attack_list`. White-door camping is an off-by-default opt-in — no automatic ambush.

## Settings

New options live under a nested `droughtmans:` block. The shared `hunting_buddies` / `worn_trashcan` / `worn_trashcan_verb` keys and the legacy flat `droughtmans_loot_container` / `droughtmans_step_delay` keys are still honored. Full documented list is in the script header.

## Tests

- `spec/droughtmans_spec.rb` rewritten to the new method surface (69 examples): dowse parsing, dead-reckoning move selection, `process_flags` ordering (prevail-then-found, drag-before-recovery), keyholder-only policy, the wave/reflection footguns, drag-aware key recovery, and the non-empty-package/pass-safety guards.
- Added `checkhidden` / `checkstunned` / `checkinvisible` stubs to `test/test_harness.rb`.
- Full suite green (2257 examples, 0 failures); rubocop clean on touched files.

## Note for reviewers

Automated tests pass, but the maze **entry flow changed** (start-location detection + `go door`/`go maze` and getting the wand in the Contestant's Box, rather than walking to a hard-coded compound room) and the auto canvas-sack fetch was dropped in favor of `loot_containers`/a plain stow. These paths can't be unit-tested and warrant an in-game smoke test before merge. Credit to the two community authors whose designs this merges.

## Summary by CodeRabbit

* **New Features**
  * Added more reliable grid-based maze navigation with section tracking, visit awareness, and risen-wall avoidance.
  * Improved event-driven dowsing and directional guidance, including center and colored-door tracking.
  * Wand actions now prioritize confirmed key holders and help manage frozen maze characters.
  * Added optional repeated runs and expanded configuration support.
* **Bug Fixes**
  * Improved recovery from drags, interruptions, dropped keys, freezing, and thawing.
  * Added safer prize and package looting, preventing disposal of non-empty packages.

### DIFF
--- a/droughtmans.lic
+++ b/droughtmans.lic
@@ -162,7 +162,7 @@ class Droughtmans
     @blackdoor      = user.fetch('blackdoor', false) == true # strict opt-in
     @camp           = user.fetch('camp', false) == true
     @multiplier     = user.fetch('multiplier', false) == true
-    @dowse          = user.fetch('dowse', true)
+    @dowse          = user.fetch('dowse', true) == true # strict; default on
     @section_moves  = user.fetch('section_moves', 25).to_i
     @section_moves  = 25 if @section_moves < 5
     @pass_adjective = user['pass_adjective']
@@ -459,8 +459,8 @@ class Droughtmans
 
   EVENT_FLAGS = %w[dm-key-found dm-prevailed dm-earned dm-npc-unfrozen dm-dark
                    dm-key-dropped dm-key-loose dm-wall-rises dm-self-thawed
-                   dm-self-frozen dm-other-thawed dm-repositioned dm-dragged
-                   dm-dowse-key dm-dowse-center].freeze
+                   dm-self-frozen dm-other-thawed dm-light-faded dm-repositioned
+                   dm-dragged dm-dowse-key dm-dowse-center].freeze
 
   # Reset run-scoped state. A previous run's unconsumed flags (e.g. the Recovery
   # Room "You have earned 8..." win line) must not trigger on this run.
@@ -622,7 +622,11 @@ class Droughtmans
   # GO DOOR until we are verifiably inside the maze. Fire-and-forget entry is how
   # a runner once ended up running maze logic in the Contestant's Box (laughing
   # fits leave you kneeling and the door refuses non-standers).
-  # @return [Boolean] whether we made it inside
+  #
+  # The boolean return is ADVISORY, not a hard stop: gear_up_and_enter ignores it
+  # and maze_loop re-enters the maze on its own whenever it finds itself still in
+  # the Contestant's Box, so a false here just means "not confirmed inside yet."
+  # @return [Boolean] whether we confirmed we made it inside (advisory)
   def enter_the_maze
     attempt = 0
     5.times do
@@ -701,13 +705,14 @@ class Droughtmans
       return :winner   if DRRoom.room_objs.any? { |o| o =~ /golden arch/ }
       return :finished if DRRoom.title =~ /Recovery Room/
 
-      # Optional early-out through the black door.
+      # Optional early-out through the black door. This ENDS the run without a
+      # win (no prize box), so finish as :finished, not :winner.
       if @blackdoor && room_has?(/black door/)
         fput('go black door')
         pause 0.2
         fput('go black door')
         pause 0.2
-        return :winner
+        return :finished
       end
 
       ensure_wand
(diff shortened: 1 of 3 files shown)


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai thanks for the review - went through all 11 points. Verdicts and fixes below.

### Actionable

1. **Ambush retry bound (white-door loop).** Fixed. Added an `ambush_attempts` counter (budget 3) around the freeze/drop branch; once exhausted it exits with the existing `:gone`/`:no_key` outcome instead of `next`-looping forever, and the counter resets whenever a non-ambush door response comes back.

2. **`go <color> door` instead of the full room-object phrase.** Fixed. The `bput` now issues `go #{color} door` using the already-extracted color adjective; color selection, section-rotation, and the failure/`room_objs.delete` handling are unchanged.

3. **`before_dying` should reuse `EVENT_FLAGS`, and `dm-light-faded` should be in it.** Fixed both. `dm-light-faded` was registered in `setup_flags` but present in neither list, so it was never torn down - added it to `EVENT_FLAGS`, and `before_dying` now iterates `Droughtmans::EVENT_FLAGS` instead of duplicating the literal.

4. **`stow_loose_hands` should skip the literal `Empty` from the hand accessors.** Not changed - I believe this one is a false positive. `DRC.left_hand`/`DRC.right_hand` return `nil` (not the string `"Empty"`) when a hand is empty:
   ```ruby
   def left_hand
     GameObj.left_hand.name == 'Empty' ? nil : fix_dr_bullshit(GameObj.left_hand.name)
   end
   ```
   so the existing `.compact` already drops empty hands and `stash_or_trash` never receives `"Empty"`. (The `_noun` variants have the buggy `== 'Empty'` object comparison, but those aren't used here.) Happy to add a belt-and-suspenders guard if you'd prefer.

### Nitpicks

5. **`@dowse` should strictly accept only `true`.** Done - `user.fetch('dowse', true) == true`, default-on preserved, matching the other boolean flags.

6. **Black-door early-out should return `:finished`, not `:winner`.** Done. It ends the run with no prize, so `:finished` avoids the winner-only oval/prize-box sequence in `finish_run`; movement and pauses unchanged.

7. **`/concealing all but/` should be treated as unknown, not pruned.** Done. Added a dedicated `when` branch that leaves `@whitelist` untouched and lets the per-player rate limiter re-check later, rather than falling into the pruning `else`.

8. **Only set `@current_quadrant` when both a N/S and E/W axis are present.** Done. A single-axis center bearing now stays a hint (kept in `@center_hint`) and is no longer stored as a quadrant, so it can't pollute `@searched_quadrants` or the quadrant comparisons/counts.

9. **Document that `enter_the_maze`'s return is advisory.** Done - updated the doc comment to note callers (`gear_up_and_enter`, `maze_loop`) don't treat `false` as a hard stop and will retry.

10. **Spec: make the `direction_toward` tie genuine.** Done. Switched to target `[0, 3]` with `northeast`/`northwest` exits (both cut Chebyshev distance to 2 - a real tie), with one destination marked visited; asserts the less-visited room is chosen.

11. **Spec: cover the non-fatal missing-pass branch of `#redeem_pass`.** Done. Added an example where the GET succeeds but REDEEM returns "What were you referring to?", asserting it echoes and returns normally without `SystemExit`; the fatal out-of-passes example is unchanged.

All specs green (70 examples, 0 failures) and rubocop clean on both files.

PR #7493, opened 2026-07-28
combat-trainer: summoned-weapon store/restore, guarded + throttled sliver creation

MahtraDR:
## Summary

Fixes moon-mage (and warrior-mage) combat-trainer bugs around a **summoned weapon** (moonblade/moonstaff or an elemental summoned weapon) and hardens sliver creation. All changes are contained to `combat-trainer.lic`; the commons helpers relied on already exist and are unchanged.

### Issue 1 - `EquipmentManager: Failed to match a weapon for moonblade:<skill>`

Several routines stow the training weapon and re-wield it around an interruption, via the raw `EquipmentManager` path:

```ruby
# ...
@equipment_manager.wield_weapon?(game_state.weapon_name, game_state.weapon_skill)
```

A summoned weapon's configured name is the spell noun (`moonblade` / `moonstaff` / an elemental weapon), which is **not a gear-list item**, so `item_by_desc` never matches:

- `wield_weapon?` prints `EquipmentManager: Failed to match a weapon for moonblade:<skill>` and re-equips nothing.
- `stow_weapon` also can't match, so the blade is **left stuck in hand** - e.g. astrology's `check_heavens` then can't free a hand and `center telescope` fails with "You must have both hands free", so the character only peers aimlessly.

This is exactly what `SetupProcess#check_weapon` already handles correctly on a weapon switch (moon mage -> `DRCMM.wear_moon_weapon?`, warrior mage -> `DRCS.break_summoned_weapon`). Extracted into two `GameState` helpers used on the raw-path sites:

- `GameState#stow_or_store_weapon` - summoned: wear (moon) / break (WM); else `stow_weapon`.
- `GameState#restore_weapon` - summoned: `prepare_summoned_weapon(false)` (re-hold + reshape); else `wield_weapon?`.

Applied in `TrainerProcess#check_heavens` (astrology) and the `casting_sorcery` re-wield in `SpellProcess#cast_spell`. For non-summoned gear weapons the helpers fall through to the identical `EquipmentManager` calls, so no behavior change for the common case.

### Issue 2 - `check_slivers` could destroy the wielded weapon

`check_slivers` snap-casts Moonblade at minimum prep and breaks it into TK ammo, but broke **unconditionally**. When the snap-cast fails (character lacks the skill to complete the pattern at minimum prep), no fresh blade exists and `break moonblade` shatters the moonblade **already in hand** - the wielded weapon of a moonblade melee trainer - leaving them unarmed. The break is now gated on the cast succeeding (`DRCA.cast_spell` returns the cast-success boolean).

### Issue 3 - `check_slivers` had no throttle (reported in review)

`check_slivers` recreated slivers the instant `DRSpells.slivers` went false, re-casting Moonblade every tick on targets that consume slivers quickly (and re-running a full 3-cast retry burst every tick for a character who cannot make slivers at all). It also ignored any configured cooldown. Added a configurable minimum interval between sliver re-creation attempts:

- `sliver_recast_delay` (seconds), **default 180**, `0` to disable.
- The cooldown starts whenever `check_slivers` commits to an attempt (after the moons-visible guard), so both successful and all-failed attempts respect it; a no-moons early return does **not** start it, so it retries once moons are up.

## Tests

`spec/combat_trainer_spec.rb`, all against the existing harness (no commons changes):

- **`stow_or_store_weapon` / `restore_weapon`**: moon (wear/re-hold), warrior mage (break/re-summon), normal gear (stow/wield exact args), summoned-membership keyed on skill not name, nil weapon_skill (no raise), non-moon/non-WM misconfig, a store->restore round trip asserting the gear list is never touched, and an integration case running the real `prepare_summoned_weapon` with zero `EquipmentManager` calls.
- **`check_heavens`**: store-before/restore-after ordering, telescope retrieved vs. unavailable vs. no-telescope, and a regression that it never calls `EquipmentManager` directly.
- **`check_slivers` break guard**: existing tests updated for the new "cast must succeed to break" contract, plus a failed-cast context (never breaks on failure, 3x retry, mixed fail/success ordering).
- **`check_slivers` throttle**: within-window block, elapsed-window cast, delay-1s boundary, custom shorter delay, `0` disables, cooldown-starts-on-attempt (success and all-fail), and no-cooldown-start when no moons are visible.

Full suite: **2443 examples, 0 failures**. `rubocop` clean on both touched files.

## Notes

- The same raw-path stow/re-wield pattern also exists in `meraud_commune`, `pray_mat`, `butcher_corpse`, and the regalia-shutdown sites. Those are guild paths that don't currently use summoned weapons (theurgy/necro), so they're left alone here to keep the diff focused; they could adopt the same helpers later.
- `break moonblade` targets whichever moonblade is in hand. The cast-success guard removes the "failed cast destroys the weapon" case; a residual ambiguity remains if a weapon blade and a freshly-summoned blade are both in hand, which is out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Added configurable throttling for telekinetic sliver recreation, with a default 180-second delay and an option to disable it.
  * Improved handling of summoned weapons during spell casting and astrology activities.

* **Bug Fixes**
  * Prevented moonblades from being broken when their replacement cast fails.
  * Fixed summoned moonblades and elemental weapons becoming stuck or failing to restore after casting or telescope use.
  * Ensured weapon restoration works reliably for both summoned and standard weapons.


MahtraDR:
@coderabbitai review in full

MahtraDR:
Thanks for the review. On the nitpick (route the remaining inline stow/wield sites through the new helpers), I verified each of the four:

- **`SpellProcess#cast_ritual`** - folded onto `stow_or_store_weapon` / `restore_weapon` (commit c22e820f). Its restore side was an exact duplicate; its stow side omitted the moon-wear branch (it always broke), so for a moon mage `cast_ritual` broke the moonblade and then couldn't re-hold it - centralizing also fixes that. Necromancer (nil summoned -> stow) and warrior-mage (break/re-summon) behavior is unchanged. Added cast_ritual specs.
- **`SetupProcess#check_weapon`** - left as-is: it disposes the *previous* weapon (`last_weapon_skill`/`last_weapon_name`), is guarded by `next_summoned`, and restores with `prepare_summoned_weapon(last_summoned)` not `(false)`. The helpers key off the *current* weapon and pass `false`, so it isn't a behavior-preserving substitution.
- **`LootProcess#check_skinning`** - left as-is: it checks `moon_mage? && wear_moon_weapon?` first and captures `stored_moon` / `summoned` locals that the later bundle + restore control flow depends on; the helper's summoned-first ordering and lack of captured state can't preserve that.
- **`SetupProcess#execute`** - left as-is: its non-summoned branch also calls `sheath_whirlwind_offhand`; folding would either drop that side effect or re-check `summoned_info` alongside the helper for no real simplification.

Full suite 2447 examples green, rubocop clean on both touched files.

PR #7483, opened 2026-07-25
[script][inventory-manager] Add ticket tracking and update ignores for blank lines

desertkaz:
Added a couple specific nodes to the inventory ignore after removing the `<` pattern to update after the backend model changes.  This prevents empty lines being saved (effected other segments as well but I missed before)

Tested on 3 characters.

image

## Summary by CodeRabbit

* **New Features**
  * Added support for collecting and storing ticket-related inventory using the new `--ticket` option.
  * Ticket entries are formatted consistently for clearer display.

* **Bug Fixes**
  * Improved recognition of ticket-tagged inventory entries.
  * Reduced unwanted recall and rendering text appearing in inventory results by expanding ignored output patterns.

PR #7498, opened 2026-07-29
inventory-manager: fix inventory parsing, parse the deed register, add a GUI browser

SekmehtDR:
Five commits. The first fixes the markup parsing bug this PR was opened for; the rest add a GUI browser, give the deed register a real parse, fix six crash paths found while bug-sweeping the script, put the save options behind a Scan button, and pick up the ticket save added in #7483.

Rebased onto current `main`, so this includes #7483. See [Overlap with #7483](#overlap-with-7483) below -- that PR filtered the same markup this one strips, and the two are compatible.

Credit for finding and fixing the original parsing bug goes to **Illiahanna**, who diagnosed the bad output, tracked it down, and wrote the first version of the fix.

---

## 1. Markup was written into inventory.yaml

`check_inventory` fed raw game lines into item text with only `line.lstrip` applied. Nothing removed XML, so bold toggles (`<pushBold/>` / `<popBold/>`), `<d cmd="...">` command tags, `<a exist="..." noun="...">` anchors and prompt/XML envelope lines all landed in `inventory.yaml` verbatim. Entities such as `&amp;` were stored undecoded.

On home saves this compounded in `clean_home_string`, which assumed every line was exactly `Category: item description.`:

```ruby
colon = string.index(":") + 2
string[colon..-2]
```

The `-2` is meant to drop the trailing period. With a tag on the end it drops the last character of the *tag* instead, and with no colon at all it raises. Actual behaviour of the old code:

| Input | Output |
|---|---|
| `Furniture: a wooden chair.` | `a wooden chair` ✅ |
| `<pushBold/>Furniture: a wooden chair.<popBold/>` | `a wooden chair.<popBold/` ❌ |
| `Furniture: <d cmd="look chair">a wooden chair</d>.` | `<d cmd="look chair">a wooden chair</d>` ❌ |
| `<popBold/>` | `NoMethodError: undefined method '+' for nil` ❌ |

Rows 2 and 3 are the "partial strings" and "stray formatting tags" seen in saved files. Row 4 aborts the save.

**Fixed** by stripping markup and decoding entities once, up front, before any parsing. Tags are stripped before entities are decoded, so a decoded `&lt;` cannot be re-read as a tag. Entities are decoded rather than deleted, so `a black &amp; white cloak` stays `a black & white cloak` instead of collapsing to `a black  white cloak`. Container-depth detection now reads the cleaned line — previously a leading `<pushBold/>` shifted the leading-whitespace count and misclassified the item. The generic container patterns no longer require an article, which had silently skipped items like `-two silver rings`; indentation thresholds are unchanged, since the old unanchored `\s{5}-` matched five *or more* spaces.

## 2. Deed register stored a raw ASCII table

`read my register` returns a table, not an item list, and all of it was being saved:

```
+-----+-------------+-----------------------------------+------------+   x3  borders
| Page| Type        | Deed                              | Notes      |   x1  header
|   1 | Wood        | some mistwood lumber              |            |   xN  deeds
Currently Stored: 38  Maximum: 100                                       x1  totals
```

A 38-deed register was stored as 43 "items", and the deeds themselves kept their pipe-delimited formatting — so `;inventory-manager search mistwood` returned `|   1 | Wood | some mistwood lumber |` rather than the item.

**Fixed** with `parse_deed_row`, which keeps the deed description as the item name so it searches like anything else, and folds the page number plus any player-written note into a single annotation. Borders, header and totals are dropped; the header falls out naturally because its page column reads `Page` rather than a number. A note containing a stray `|` is rejoined rather than truncated. Registers saved before this change are hidden by the viewer until re-saved.

## 3. Commands could retry forever

Six sites looped on `until capture != []`. `Lich::Util.issue_command` returns an empty array when its end pattern never arrives, so any command the game refused span forever. `;inventory-manager save scrolls` against a configured stacker you are not actually carrying repeated `flip my folio` indefinitely:

```
>flip my folio
You should be holding the folio.
>flip my folio
You should be holding the folio.
...
```

**Fixed** with `capture_command`, which tries three times, reports, and lets the caller carry on. `add_scrolls` also now honours the result of `DRCI.get_item_safe?`, which was being discarded — that is why a failed `get my scroll folio` still led to a `flip`.

## 4. Crash fixes from a bug sweep

Each of these was reproduced before being fixed:

- **`setup` raised on a brand-new database.** `YAML.load_file` returns `nil` for the empty file `setup` had just created, so indexing the missing version row failed. Every first run hit this.
- **Rummaging a container with one item, or none, crashed.** The response has no `" and "` to split on, so a `nil` was appended to the item list and the parse loop raised on it. Affected `vault_regular`, `family_vault`, `storage_box` and `pocket`. An empty container also stored `"nothing"` as an item.
- **`clean_home_string` returned nil** for a bare category such as `Furniture:`, crashing the caller's `concat`.
- **`;inventory-manager count` counted markup.** `get_inv_count` still did bare `line.lstrip`, so prompt, bold and xml lines were counted as items — 5 where there were 2. Tag stripping now lives in one `clean_item_line` method shared with `check_inventory`, so the two paths cannot drift apart again.
- **`add_scrolls` raised** when `scroll_stackers` was absent from the yaml.
- **`remove_character_data` reported success** for a name that was not present, and guarded the schema row case-sensitively. Character lookups are now case-insensitive, which would otherwise have allowed `remove Version` to delete the schema row.

## 5. New: `;inventory-manager gui`

A read-only browser over the saved database. Scanning stays entirely with the text commands; this only visualises what they have already written, so it needs no game interaction and is safe to leave open.

```
Agan                              549 items
  Carried & Worn                  365 items
    a leather backpack            container, worn
      a smooth rock               container
        a tiny vial
  Vault                           141 items
  Deed Register                     38 items
    some mistwood lumber          page 1
```

- Collapsible character > source > container > item tree. Nesting recorded in the database is reconstructed as real hierarchy, and an item that holds other items becomes a container node rather than appearing twice.
- Live search over item *and* container names, so a match keeps its container context.
- Character filter, a toggle for a flat list with `in <container>` details, expand/collapse, and reload for cross-instance sync.
- Elanthipedia lookup on double-click, the toolbar button, or right-click. Uses MediaWiki's plain `?search=`, which jumps straight to the article on an exact title match and falls back to search results otherwise. Articles, counts and `(closed)` are trimmed so `some iron ore` looks up `iron ore`.
- Window geometry persists via `CharSettings`.

Built against the GTK3 bindings Lich ships (`HAVE_GTK` / `Gtk.queue`), matching `kill-counter` and `status-monitor`. `HAVE_GTK` is referenced only on the `gui` path, so installs without ruby-gtk keep working normally. Closing the window used to raise `destroyed GLib::Object`; `delete_event` now clears the window handle before returning, since GTK owns the destroy from that point, and the `before_dying` cleanup still needed for `;kill` is guarded.

## 6. New: a Scan button in the GUI

Every `;inventory-manager save` option is available from the browser, so a scan can be started without dropping back to the command line. This adds no parsing of its own -- it queues the same save methods the text dispatch calls.

Scan opens a dialog listing all fifteen options as checkboxes, grouped and labelled with the command's own help text, so requirements like *"must be standing by your open vault"* are visible when choosing rather than buried in help output. Any combination can be selected; they run one at a time, in order, with a pause between each so consecutive saves do not stack commands on top of one another. Each reports whether it finished, was skipped or failed, and the tree is redrawn after every scan rather than only at the end.

**Threading.** The save methods block waiting on game output, which must never happen inside `Gtk.queue` or the UI deadlocks. The dialog only pushes selections onto a queue; the script thread drains it. Closing the window abandons whatever is still queued.

**Failure isolation.** Three ways a save can fail are contained so one bad option cannot stop the rest of the queue or take the window down:

- Several save methods call `exit` when a precondition fails -- wrong guild, book not found, no stacker configured. From the command line that simply ends the script; from a long-lived window it would close everything. Lich raises `SystemExit` for a bare `exit` (the script eval in `script.rb` is wrapped in `rescue SystemExit`), so it is caught and reported as skipped.
- An unexpected error is reported against the option that raised it.
- A save that will not finish -- one that keeps retrying something it can never do, such as fetching a book that is not there -- is abandoned after a timeout.

## Overlap with #7483

#7483 landed while this was open, and the two touched the same problem from different directions. It added `'<output class'` and `'<prompt time'` to `inventory_manager_ignores` in `profiles/base.yaml`, filtering the markup lines out by configuration; section 1 above strips markup in `check_inventory` instead.

They are compatible and both are kept. Markup lines are now skipped before the ignore list is consulted, so those two entries are redundant but harmless -- and they still protect anyone running an older copy of the script against a current `base.yaml`. Nothing from #7483 was reverted. Its `'You take a moment to recall'` entry is required by the ticket command and is untouched.

The ticket save itself is also now offered by the GUI: it has a source label in the tree and an entry in the Scan dialog, which it did not get automatically. The Scan regression harness derives its expected option set from the script's dispatch chain rather than a hardcoded count, so the next save option added upstream fails the check instead of quietly going unoffered.

## Compatibility

No data-format change; the `inventory.yaml` schema version stays at `2.00`. Existing saves keep whatever markup or raw table rows they already contain until re-saved — anyone affected should re-run the relevant `;inventory-manager save ...`. The viewer hides legacy register table rows in the meantime.

## Verification

`ruby -c` clean, and ASCII-only throughout for the new `Custom/AsciiOnlySource` cop.

Regression harnesses were run against each change:

- Stored-row parsing across all seven annotation formats, including the nested-container case where the parent's full description is embedded in the child's string.
- Tree building and de-duplication, asserting hierarchy, no duplicate items, and search filtering.
- The deed table end to end, from real `read my register` output through to what the viewer displays, including notes and a note containing a `|`.
- One reproduction per crash in section 4, each confirmed failing before the fix and passing after.
- The Scan queue: ordering, the pause landing between scans but not after the last, button state through a run, and each of the three failure modes proven not to halt the queue (the `exit` case genuinely calls `exit`; the stuck case genuinely runs past its timeout).

The parser was also run over a live 545-row `inventory.yaml`: no unparsed rows, no empty item names, no residual markup, no container text leaking into names, and 433 of 437 container references resolving to a real item (the remainder render under their source node, which is the intended fallback).

Widget behaviour — the tree, search, filters and Elanthipedia lookup — was exercised in game.

Two things worth flagging for review. The empty-container guard in section 4 drops an entry matching `/\Anothing\.?\z/i`; the crash fix itself is phrasing-independent, but that specific string is an assumption about DR's wording and I would welcome a correction. And the GUI is a substantial addition — happy to split it into its own PR if you would rather review it separately from the parsing fixes.

I am also glad to convert the harnesses above into a proper `spec/inventory_manager_spec.rb` for the new RSpec suite if that would help this land.

### DIFF
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -3,6 +3,17 @@
 =end
 
 class InventoryManager
+  # Entities the game emits inside item names. Decoded rather than deleted so
+  # "a black &amp; white cloak" stays "a black & white cloak".
+  XML_ENTITIES = {
+    'amp'  => '&',
+    'lt'   => '<',
+    'gt'   => '>',
+    'quot' => '"',
+    'apos' => "'",
+    '#39'  => "'"
+  }.freeze
+
   def initialize
     @active_character = Char.name.to_sym
     @item_db_path = "data/inventory.yaml"
@@ -223,8 +234,18 @@ class InventoryManager
     sub_container = ""
     lines.each do |line|
       line.sub!(/\(\d+\)/, '') if command == 'vault standard'
-      item = line.lstrip
+      # Game output carries XML noise (bold toggles, <d> command tags, prompts)
+      # that used to leak into inventory.yaml verbatim. Skip the structural
+      # lines outright, then strip tags and decode entities once, up front, so
+      # the item text and the indentation checks below both see clean text.
+      next if line =~ /<\/?(?:prompt|output)\b/ || line.include?('<?xml')
 
+      # Tags first, then entities -- otherwise a decoded &lt; could be mistaken
+      # for the start of a tag.
+      clean_line = line.gsub(/<[^>]*>/, '')
+                       .gsub(/&(amp|lt|gt|quot|apos|#39);/) { XML_ENTITIES[Regexp.last_match(1)] }
+
+      item = clean_line.strip
       next if item.empty?
       next if item.start_with?(*@inventory_ignores)
 
@@ -235,9 +256,9 @@ class InventoryManager
       # if the item starts with a -, it's in a container
       if item[0].eql? '-'
         item[0] = ''
-        if line =~ /\s{8,}-(?:a|an|some|the|several|clusters|one) /
+        if clean_line =~ /^\s{8,}-/
           item.concat(" (nested container - #{sub_container})")
-        elsif line =~ /\s{5}-(?:a|an|some|the|several|clusters|one) /
+        elsif clean_line =~ /^\s{5,}-/
           item.concat(" (in container - #{container})")
           sub_container = item.to_s
         else
@@ -249,9 +270,9 @@ class InventoryManager
         item.concat(' (worn)')
       elsif command == "read my storage book" || command == "read my vault book" || inventory_type == "shadow_servant"
         # Vault book, storage book and shadow servant via PG all have same output formatting
-        if line =~ /\s{8,}(?:a|an|some|the|several|clusters|one) /
+        if clean_line =~ /\s{8,}(?:a|an|some|the|several|clusters|one) /
           item.concat(" (nested container - #{sub_container})")
-        elsif line =~ /\s{6}(?:a|an|some|the|several|clusters|one) /
+        elsif clean_line =~ /\s{6}(?:a|an|some|the|several|clusters|one) /
           item.concat(" (in container - #{container})")
           sub_container = item.to_s
         else
@@ -259,9 +280,9 @@ class InventoryManager
         end
       elsif command == 'vault standard'
         # items in a container are prefixed by 11 spaces once the item count is removed
-        if line =~ /\s{16,}(?:a|an|some|the|several|clusters|one) /
+        if clean_line =~ /\s{16,}(?:a|an|some|the|several|clusters|one) /
           item.concat(" (nested container - #{sub_container})")
-        elsif line =~ /\s{11}(?:a|an|some|the|several|clusters|one) /
+        elsif clean_line =~ /\s{11}(?:a|an|some|the|several|clusters|one) /
           item.concat(" (in container - #{container})")
           sub_container = item.to_s
         else
@@ -296,8 +317,12 @@ class InventoryManager
 
   def clean_home_string(string)
     # We remove the category by finding the : in the string, plus drop the period at the end
-    colon = string.index(":") + 2
-    string[colon..-2]
+    colon = string.index(":")
+    # No category prefix means there is nothing to trim; slicing blind here used
+    # to raise on nil or lop the front off an unrelated line.
+    return string if colon.nil?
+
+    string[(colon + 2)..-2]
   end
 
   def add_ticket
--- a/inventory-manager.lic
+++ b/inventory-manager.lic
@@ -676,7 +676,9 @@ class InventoryManager
       return
     end
 
-    InventoryManagerGui.new(@item_db_path, @version_string).run
+    # Passing self lets the Scan button reuse the same save methods the text
+    # commands call, rather than duplicating any of that logic.
+    InventoryManagerGui.new(@item_db_path, @version_string, self).run
   end
 
   def remove_character_data(name_to_remove)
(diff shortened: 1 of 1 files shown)


SekmehtDR:
Piggybacking you here @desertkaz , funny enough I was working on a fix for what you did in base, but your code was merged before mine. I've incorporated your post pr version here, and added a GTK GUI and lets you Epedia stuff (was inspired by the Genie Plugin).

PR #7504, opened 2026-08-01
fix(crafting): reliability fixes + restock-shop features (forge, smith, workorders, restock-shop)

bmcneff:
# fix(crafting): reliability fixes + restock-shop features (forge, smith, workorders, restock-shop)

**Repo:** `elanthia-online/dr-scripts`
**Files:** `forge.lic`, `smith.lic`, `workorders.lic`, `restock-shop.lic`

## Summary

A set of related crafting/shop-automation changes discovered while running an
auto-restocking Trader shop. Three are standalone reliability fixes (forge,
smith, workorders); the fourth is a substantial feature+fix pass on the
restock-shop orchestrator.

The headline restock-shop change is **demand-sized steel batching**: instead of
making one fixed ingot per run and trashing the leftover, the script now keeps a
persistent refined ingot and tops it up only as much as the queued work needs.
It sizes each `makesteel` toward the remaining demand (bounded by the in-game
210-volume ingot cap and `makesteel`'s count-21 limit, adjusted for refine loss),
smelts the fresh ingot into the persistent one, skips any item too large to forge
from a single ingot, and carries the remainder to the next run (optionally as a
deed). That batching also underpins the low-carbon path and the persistent/deed
storage below.

No new dependencies; each script still reads its config from YAML.

## forge.lic

- **Temper in forge-less rooms.** Tempering issues `put/turn/analyze ... on forge`,
  but grinding happens at grindstone rooms that have no forge (e.g. the Smelting
  and Grinding Area), so tempering failed there with "What were you referring to?".
  Added `ensure_forge` / `forge_in_room?` (new `FORGE_MISSING_PATTERN`) and a guard
  in the work loop (`ensure_forge if @command.include?('forge')`) that walks to a
  free anvil room (forges are co-located with anvils) when the current room lacks one.
- **Tool/product noun collision on retrieval.** When crafting an item whose noun
  matches a forging tool (e.g. a crafted "wide shovel" vs. a "wide shovel" tool),
  the retrieval paths stowed only one hand and left the colliding tool in the other,
  causing the finished item to be tied to the belt. `prepare_item_in_left_hand` and
  `handle_oiling` now `stow_both_hands` before retrieving the item. Removed the now
  redundant `stow_hammer_and_tongs` helper (superseded by `stow_both_hands`).

## smith.lic

- **Idempotent stamping.** `forge.lic` already stamps the item after initial
  creation (when `mark_crafted_goods` is set), so `smith`'s subsequent stamp hit
  "There is already a maker's mark" and timed out. `stamp_item` now treats that
  response as success instead of retrying.

## workorders.lic

- **Repair insufficient-funds recovery.** When an NPC repair quote exceeded coin on
  hand, the confirm step only listened for "repair ticket" and 15s-timed-out on
  "You will need more coin". The quote line is now captured (cost included) via
  `REPAIR_QUOTE_PATTERN`; a new `confirm_repair` withdraws 10x the quoted cost from
  the bank, walks back, and retries (capped at 3 attempts, with clean bail-outs).

## restock-shop.lic

- **Demand-sized steel batching.** Previously each run made one fixed ingot and
  trashed whatever was left. Now steel is produced to fit the work: a single
  in-game ingot is capped at 210 raw volume (`RAW_STEEL_INGOT_CAP`), from at most
  a count-21 `makesteel` (`MAKESTEEL_MAX_COUNT`), and refining loses volume by a
  configurable factor (`steel_refine_factor`, default 1.25 with the tech / 2.0
  without). Before forging, `filter_forgeable` drops (with a warning) any item
  whose volume exceeds a single ingot, since it can't be made from one. Then for
  each item, `ensure_ingot_volume`/`add_ingot_volume` top the working ingot up
  only when it can't cover the next item, sizing each `makesteel` toward the
  *remaining* queued demand (capped at one ingot) to minimize crucible trips, and
  smelting (`combine_ingots`) each fresh refined ingot into the persistent one.
- **Word-based shop detection.** The game inserts the material adjective at an
  unpredictable spot in crafted names (e.g. "a weighted steel pickaxe"), so
  contiguous-substring `full_name` matching missed items and re-crafted duplicates.
  Detection now requires every word of `full_name` to appear in the listing.
- **sew material/noun arg order.** Tailoring calls passed `sew` arguments with
  material and noun swapped, so cloth items fell into the leather branch
  ("get my shirt leather"). Fixed the argument order for burlap/linen/silk/gryphon.
- **Persistent + deed steel ingot.** Carry the leftover refined ingot between runs
  (the batching above tops it up) rather than trashing it, optionally stored as a
  light deed (`deed_steel_ingot`); loose-ingot-first, and the deed is tapped only
  inside the forging society.
- **Low-carbon steel path.** Items with `material: low carbon steel` (e.g. tongs) are
  forged from a 1:1 iron:coal ingot on a separate, non-persisted path.
- **Carbon-grade guards.** Both grades are just "steel ingot" in-game; `analyze`'s
  "composed of: NN% <grade> carbon steel" line is used to keep grades from mixing.
- **Stamp tolerance.** Its own stamp step accepts "already a maker's mark".

## Tests

Existing specs updated for the changed behavior, and new coverage added for the
new methods (RSpec, using the repo's central game doubles):

- **`spec/forge_spec.rb`** — updated `#handle_oiling` (now stows both hands →
  `stow_crafting_item` twice). Added `#forge_in_room?` (item / empty / no-forge),
  `#ensure_forge` (skip when a forge is present; walk to a free anvil when absent;
  exit when none found), and a `FORGE_MISSING_PATTERN` check.
- **`spec/workorders_spec.rb`** — updated the repair test's `bput` stub to the new
  confirm flow. Added `REPAIR_QUOTE_PATTERN` / `REPAIR_NEED_COIN_PATTERN` checks,
  `#repair_cost_copper` (parse / nil), and `#confirm_repair` (accept → stow ticket;
  short on coin → withdraw 10x and retry; unknown cost → skip).
- **`spec/restock_shop_spec.rb`** (new) — `#already_stocked?` (word-presence
  matching), `#steel_ingot_carbon` (grade parse), `#makesteel_args` (high vs low),
  and the batching: `#filter_forgeable` (drops over-cap items), `#ensure_ingot_volume`
  (fills toward remaining demand, capped at one ingot), and `#add_ingot_volume`
  (count sizing toward `fill_to`, capped at `MAKESTEEL_MAX_COUNT`; combine when the
  ingot already has volume).

## Verification notes
- `ruby -c` clean on all four scripts and all three specs.
- Detection verified against real shop output; grade parsing verified against real
  `analyze` output for high/medium/low.
- forge/smith/workorders are pure additions over the current public versions
  (straight upgrades; nothing removed except a helper superseded by a broader one).
  restock-shop.lic is a large net-new feature set over the 266-line public base.

### DIFF
--- a/restock-shop.lic
+++ b/restock-shop.lic
@@ -291,7 +291,7 @@ class ShopRestock
   # Smelt the freshly-made refined ingot into the persistent one, yielding a
   # single combined ingot. Assumes both are refined (identical) and total <= 210.
   def combine_ingots
-    DRCC.find_empty_crucible(@hometown)
+    DRCC.find_empty_crucible(@town)
 
     added = 0
     2.times do
--- a/restock-shop.lic
+++ b/restock-shop.lic
@@ -3,6 +3,29 @@
 =end
 
 class ShopRestock
+  # Hard cap on a steel ingot's volume in-game (raw OR refined-and-combined).
+  # A raw ingot is (makesteel_count x 10) volume, and makesteel's count caps at
+  # 21, so a single make tops out at 210 raw. Ingots can also be smelted
+  # together, but the combined result can never exceed this same 210 cap.
+  RAW_STEEL_INGOT_CAP = 210
+
+  # makesteel's own arg cap on iron-nugget count -- a single make can't exceed
+  # this (21 x 10 = 210 raw volume).
+  MAKESTEEL_MAX_COUNT = 21
+
+  # Refine-loss factor: raw ingot volume / refined ingot volume.
+  #
+  # restock_steel always refines, which raises quality but reduces volume:
+  # a 210 raw ingot yields 168 WITH the steel-refining tech (210 / 168 = 1.25)
+  # or 105 WITHOUT it (210 / 105 = 2.0). So one makesteel+refine adds up to
+  # (count x 10 / factor) refined volume -- 168 max with the tech.
+  #
+  # Used to size each makesteel call before its fresh refined ingot is smelted
+  # into the persistent ingot (see restock_steel). Defaults to the with-tech
+  # value; override per-character with a `steel_refine_factor` YAML setting
+  # (use 2.0 if you lack the tech).
+  DEFAULT_STEEL_REFINE_FACTOR = 1.25
+
   def initialize
     @settings = get_settings
 
@@ -16,6 +39,19 @@ class ShopRestock
     @belt = @settings.forging_belt || @settings.engineering_belt
     @worn_trashcan = @settings.worn_trashcan
     @worn_trashcan_verb = @settings.worn_trashcan_verb
+
+    # Optional: store the persistent steel ingot as a (light) deed between runs.
+    @deed_ingot = @settings.deed_steel_ingot
+    if @deed_ingot
+      deeds_data = get_data('crafting').deeds[@town]
+      if deeds_data
+        @deeds_room = deeds_data['room']
+        @deeds_number = deeds_data['large_number'] # large packet = 250 claim forms
+      else
+        Lich::Messaging.msg('bold', "restock-shop: No deed data for #{@town}; disabling steel-ingot deed storage.")
+        @deed_ingot = false
+      end
+    end
 
     go_to_shop
     check_surfaces
@@ -33,7 +69,7 @@ class ShopRestock
                    DRC.list_to_array(raw_list)
                  end
       missing_items += @item_list.select { |item| surface == item['surface'] }
-                                 .reject { |item| raw_list.find { |raw_item| raw_item.match(item['full_name']) } }
+                                 .reject { |item| already_stocked?(item, raw_list) }
     end
 
     return if missing_items.empty?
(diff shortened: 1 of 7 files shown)


restock-shop.lic
```diff
@@ -51,34 +87,299 @@ class ShopRestock
...
+
+    result = DRC.bput('analyze my ingot', /composed of: [\d.]+% \w+ carbon steel/, 'I could not find')
+    DRCI.stow_item?('ingot')
+    result[/% (\w+) carbon steel/, 1]
+  end
+
+  # Smelt the freshly-made refined ingot into the persistent one, yielding a
+  # single combined ingot. Assumes both are refined (identical) and total <= 210.
+  def combine_ingots
+    DRCC.find_empty_crucible(@hometown)
```
MahtraDR: `@hometown` is never assigned in this file -- the rest of restock-shop uses `@town` (init, `dispose_leftover_ingot`, `ensure_deed_packet`). This is the only `@hometown` reference here, so it passes `nil`, and `find_empty_crucible(nil)` hits `get_data('crafting')['blacksmithing'][nil]['crucibles']` -> `NoMethodError` on the search path. The early `return` (tap succeeds + solo + empty crucible) hides it for a solo crafter standing at the just-used crucible, but it crashes as soon as another PC is in the room or the crucible isn't detected empty -- which is squarely the batching path this PR adds. The specs stub `combine_ingots`, so they don't catch it.

```suggestion
    DRCC.find_empty_crucible(@town)
```

workorders.lic
```diff
@@ -24,11 +24,19 @@ class WorkOrders
 
   NPC_NOT_FOUND_PATTERN = "What is it you're trying to give".freeze
 
+  # The clerk's repair quote, e.g. "That will cost 4,447 Kronars to repair.
+  # Just give it to me again..." -- captured (cost included) so we can read the
+  # price and top up from the bank if we are short before confirming.
+  REPAIR_QUOTE_PATTERN = /cost [\d,]+ \w+ to repair.*give it to me again/i
```
MahtraDR, line 33: Could you confirm this against real NPC output in-game? This regex can't span lines, so it assumes the cost and "give it to me again" land on a single game line. If the clerk splits them across two lines, the first `give` matches nothing, bput times out, and the repair silently no-ops -- a regression from the old bare `'Just give it to me again'` string, which matched its own line. The spec encodes a single combined line and the PR notes it's verified, so this is likely fine; just flagging it as the one thing that would silently break repair if the format differs.

MahtraDR (commented):
Reviewed the four scripts against the current source, ran the full suite (227 examples, 0 failures) and rubocop (clean). The fixes hold up and the sew material/noun swap + bput-return semantics all check out. Two things below: one confirmed bug with a suggested fix, and one assumption I couldn't verify without the game.

restock-shop.lic
bmcneff: fixed, that's a real crash in the batching path. Added a combine_ingots spec that asserts the crucible lookup uses @town so it can't regress.

workorders.lic
bmcneff: Confirmed against live output - the clerk puts the cost and 'give it to me again' on one line (That will cost 4,447 Kronars to repair.  Just give it to me again if you want...), which the spec encodes. To be safe against a split, I re-added the bare 'Just give it to me again' as a fallback so the prompt still matches and the repair proceeds (just without auto-funding) rather than timing out

PR #7510, opened 2026-08-02
docs: add standard documentation headers to 35 scripts

SekmehtDR:
## Goal

Standardise the basic documentation on scripts in this repository, so that anyone opening one can tell what it does from the first few lines, and find the Elanthipedia page for it where one exists.

Plenty of scripts already do this well. Others open straight into code with nothing at the top, so the only way to learn what they do is to read the implementation. This closes that gap for the 35 that were missing a header, using the same `=begin` convention already in use elsewhere in the repo.

## What changed

Each of the 35 scripts gains a header giving:

- a short description of what the script does and how it behaves
- a `Documentation:` link to its Elanthipedia page, where one exists

For example, `arenawatch.lic` went from opening on `def mainloop` to:

```ruby
=begin
  Watches arena NPCs for telegraphed attacks and responds with the matching
  evasive maneuver (pedal, bob, duck, jump, lean, cower). Pauses combat-trainer
  while a watch is in progress and unpauses it once the maneuver is set.
=end
```

And where a script already had a header but no link, only the link was added -- `cast.lic` is a one-line change.

## Scope

**Comments only.** No behaviour changes and no code touched: 221 added lines and zero deletions across the 35 files.

13 of the 35 gained an Elanthipedia link. The remainder have no wiki page to point at, so they carry a description alone; those links can be filled in later if pages are written. Every link was checked to confirm its anchor matches the script's filename, so none of them are dead on arrival.

## Verification

- `ruby -c` passes on all 35 files.
- `rubocop` passes on all 35, including `Custom/AsciiOnlySource` -- the headers are plain ASCII, with no arrows, em dashes or curly quotes.
- Every `Lich_script_repository#anchor` was checked against its filename.

The descriptions were written by reading each script. They describe what the code does today rather than what the name suggests, but a maintainer who knows a given script well may still spot a nuance worth correcting -- happy to adjust any of them.


SekmehtDR:
Note: This is just adding comments at the top of scripts that do not have some basic description of what they do, and or are missing their Elanthipedia link (if that's still being used). Thanks!

MahtraDR:
Thank you!

PR #7514, opened 2026-08-08
craft: check herbs before buy/forage; workorders: re-read logbook before turn-in

simtel12:
## Summary
- craft: check alchemy herbs before buy/forage and add `;craft alchemy debug` support, avoiding unnecessary bank/stock trips when ingredients are already on hand and skipping remedy when funds or herbs can't be acquired
- workorders: re-read the forging logbook's remaining count before turn-in instead of trusting the originally asked quantity, and stop claiming turn-in success on incomplete give responses

## Test plan
- [x] `rspec spec/workorders_spec.rb` passes (98 examples, 0 failures)
- [x] `ruby -c craft.lic` / `ruby -c workorders.lic` syntax check clean

PR #7518, opened 2026-08-09
workorders: add ;continue to resume an in-progress order from the logbook

simtel12:
## Summary
- Adds `;workorders <discipline> continue` to resume an in-progress work order by reading the discipline logbook, instead of canceling and re-requesting one (e.g. after a disconnect). Recovers the recipe and remaining quantity from the logbook text, and skips straight to turn-in when the logbook reports the order already complete.

Rebased on current `main` now that #7514 (which added `logbook_remaining`, relied on here) has merged, so the diff is just this feature.

## Test plan
- [x] `rspec spec/workorders_spec.rb` passes (106 examples, 0 failures)
- [x] `ruby -c workorders.lic` syntax check clean

### DIFF
--- a/craft.lic
+++ b/craft.lic
@@ -6,7 +6,8 @@ class Craft
   def initialize
     arg_definitions = [
       [
-        { name: 'craft', options: %w[forging outfitting engineering alchemy enchanting], description: 'What craft to train' }
+        { name: 'craft', options: %w[forging outfitting engineering alchemy enchanting], description: 'What craft to train' },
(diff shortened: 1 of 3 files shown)


simtel12:
Depends on #7514 — `resume_work_order` calls `logbook_remaining`, which is added there. This PR's diff currently includes #7514's commits too; it'll narrow to just this feature once #7514 merges.

MahtraDR:
@simtel12 I'm keen to merge this, but we have conflicts here too.

MahtraDR:
Please fix this one! :)

PR #7522, opened 2026-08-09
Add AI agent instructions (CLAUDE.md + AGENTS.md symlink)

MahtraDR:
## Problem

The repo has no shared instructions for AI agents. AI-assisted contributions and code review therefore lack repo-specific context and miss high-signal checks:

- Lich injects host globals (`fput`, `echo`, `Flags`, ...) and the `DRC*`/game-state modules at runtime, so naive review flags them as "undefined."
- The commons layer (`DRCC`, `Lich::Messaging`, `EquipmentManager`, ...) is a **cross-repo dependency** in **lich-5** (`lib/dragonrealms/commons/`), not here.
- The Ruby floor is `4.0.0` (no ceiling); autoloaded stdlib like `Set` needs no `require`.
- Rubocop and the two YAML CI gates have specific, non-obvious scopes.
- The rspec suite runs in one process on shared doubles, so one bad spec can break **unrelated** specs.

## Fix

Add the guidance once, in `CLAUDE.md`, and symlink `AGENTS.md` -> `CLAUDE.md`:

- `CLAUDE.md` is the real file — read directly by Claude Code and by code-review tooling that fetches `CLAUDE.md` by name.
- `AGENTS.md` is a git symlink (mode `120000`) so any agent that looks for `AGENTS.md` resolves to the same single source of truth. One file, no duplication or drift.

Documented: runtime environment, commons layer + its lich-5 location, rubocop/CI behavior, spec-suite integrity, testing conventions, YAML gates, and an adversarial code-review stance with a calibrated common-false-positives list.

## Tests

Docs-only change (Markdown + symlink). No code or specs: `CLAUDE.md` is not linted by rubocop (`ascii_only_source` is scoped to `*.lic`/`*.rb`) and triggers no CI job. Verified `upstream/main` has neither `CLAUDE.md` nor `AGENTS.md` (net-new, no conflict).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,140 @@
+# dr-scripts Project Instructions
+
+DragonRealms automation scripts for the Lich engine. Each `*.lic` file is a
+standalone Ruby script executed inside the Lich runtime, not a normal gem. These
+scripts depend on the separate **lich-5** framework repo (typically checked out
+alongside this one), which provides the runtime, game-state objects, and the
+commons library the scripts call into — a cross-repository dependency to keep in
+mind when a symbol appears "undefined" here.
+
+## Runtime environment (read before flagging "bugs")
+
+- **Ruby 4.0.0 floor, no ceiling.** Code must run on Ruby `>= 4.0.0`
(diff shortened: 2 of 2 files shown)


simtel12:
Might be a good idea to add this as `AGENTS.md` and then import it in `CLAUDE.md` (ie. just do `@AGENTS.md` in the claude file) so that other agents pick it up.

Vice versa also works.

PR #7530, opened 2026-08-12
[plugin][combat-trainer] Create combat-trainer-plugin-spit.rb

desertkaz:
#### Requires Dragon's Breath properly set as a buff.

Plugin that tracks the Heavy TM cooldown state for Dragon's Breath and spits the next round when cooldown is complete.


desertkaz:
main repo not right spot, closing

PR #7535, opened 2026-08-15
Update base-hunting.yaml

Elysande1218:
Adding
-  dryanoxie elanthipedia page
- Hunting area for sovamedo cub
- Hunting area for sovamedo adult
- Hunting area for snarlvine fox/vixen


MahtraDR:
Thank you!

PR #7536, opened 2026-08-15
Fix empathylink pause deadlock that could freeze the character

MahtraDR:
## Problem

empathylink's `touch()` paused combat-trainer directly with `pause_script('combat-trainer')` (`p_script`) and unpaused it afterward (`u_script`), operating entirely **outside** the shared `$safe_pause_lock` that the combat scripts coordinate through via `DRC.safe_pause_list`.

Lich pausing is cooperative — a script flagged paused keeps running until its next checkpoint. When combat-trainer's almanac routine acquired `$safe_pause_lock` and paused empathylink, empathylink (in that grace window) ran on into `touch()` → `p_script` and paused combat-trainer, then suspended at its next `bput` **before** reaching `u_script`. Result:

- combat-trainer was left paused (by empathylink) while still holding `$safe_pause_lock` and having paused the other scripts, so it never reached `safe_unpause_list`.
- empathylink was left paused (by combat-trainer), so it never reached `u_script`.

The two pinned each other, `$safe_pause_lock` was never released, every later script that called `safe_pause_list` spun forever on "Cannot pause, trying again", and the character froze in place.

Observed in a live log: `DRC: Pausing [...] to run combat-trainer` (combat-trainer's almanac study, running in its thread) immediately followed by `--- Lich: combat-trainer paused.` (empathylink's `p_script`) with no matching unpause, then endless `Cannot pause` from other scripts.

## Fix

Replace `p_script`/`u_script` with the `DRC.safe_pause_list` / `DRC.safe_unpause_list` protocol already used by `tarantula` and `tome`. Sharing `$safe_pause_lock` makes the two pause requests mutually exclusive, so neither script can pause the other mid-routine. The heal is wrapped in `begin/ensure` so the unpause and lock release always run, even if a heal command raises.

Behavior note: like tarantula/tome, empathylink now briefly pauses all other safe scripts (not just combat-trainer) for the duration of a heal touch. That window is short and matches the established idiom.

## Tests

Adds `spec/empathylink_spec.rb` covering:
- lock acquisition and retry-until-acquired
- the paused set is always unpaused, including the link-failed branch
- the unpause/lock-release still happens when a heal command raises mid-touch (the core deadlock regression)
- the deadlock-prone `p_script`/`u_script` helpers are gone

Full suite green locally (2907 examples, 0 failures); `rubocop -A` clean on `empathylink.lic`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Healing now safely pauses conflicting activities before starting and reliably resumes them afterward.
  * Improved recovery when linking fails or a command encounters an error.
  * Added retries when the safe pause is temporarily unavailable.
* **Tests**
  * Expanded coverage for healing, pause handling, retries, and error cleanup.

### DIFF
--- a/empathylink.lic
+++ b/empathylink.lic
@@ -47,25 +47,40 @@ class Empathylink
   end
 
   def touch(target)
-    p_script
-    waitrt?
-    case DRC.bput("touch #{target}",
-                  "You touch",
-                  "You lay your hand on",
-                  "Touch what",
-                  "I could not find",
-                  "You rest your hand on")
-    when 'You touch', 'You lay your hand on', 'You rest your hand on'
-      if @use_hodierna == true
-        hodierna_heal_patient(target)
-      else
-        heal_patient(target)
-      end
-    when 'Touch what', 'I could not find'
-      DRC.message("Failed to establish link with #{target}")
-      UserVars.empathylink[target] = Time.now.to_i + 1800 # Put them on hold for 30m
+    # Pause other scripts through the shared safe-pause lock ($safe_pause_lock)
+    # so a combat script cannot pause us mid-heal while we have it paused. The
+    # old approach called pause_script('combat-trainer') directly, outside that
+    # lock: if combat-trainer's own safe_pause_list paused empathylink between
+    # p_script and u_script, each ended up pausing the other and neither could
+    # unpause, deadlocking every script and freezing the character. safe_pause_list
+    # makes the two pause requests mutually exclusive, and the ensure guarantees
+    # the unpause (and lock release) even if a heal command raises.
+    until (scripts_to_unpause = DRC.safe_pause_list)
+      echo('Cannot pause, trying again in 5 seconds.')
+      pause 5
     end
-    u_script
+
+    begin
+      waitrt?
+      case DRC.bput("touch #{target}",
+                    "You touch",
+                    "You lay your hand on",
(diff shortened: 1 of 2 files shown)


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai thanks for the review — addressing both nitpicks:

**1. Visibility-inclusive `respond_to?` in the spec — done** (commit f7d0dce4).
The removed-helper assertions now use `respond_to?(:p_script, true)` / `respond_to?(:u_script, true)` so a re-added *private/protected* `p_script`/`u_script` is still caught, not just a public one. That's a real strengthening of the regression guard.

**2. Bounding the `until DRC.safe_pause_list` retry loop — intentionally skipped.**
The unbounded retry (echo + `pause` + loop) is the established idiom in this repo — `tarantula` (`turn_tarantula?`) and `tome` (`pause_scripts`) both loop forever on `DRC.safe_pause_list` with the same shape. The only way this loop spins indefinitely is if `$safe_pause_lock` is held and never released, and that permanent-hold scenario is exactly the deadlock this PR removes: post-fix the lock is only ever held for the brief duration of another script's exclusive work, so the retry drains within a cycle or two. Bounding empathylink alone would diverge from its siblings without addressing the (systemic, not empathylink-specific) hang case, so I've kept it consistent with the existing pattern.

PR #7552, opened 2026-08-20
enchant: study sigils out of sigil books

simtel12:
## Summary

Two changes to `enchant.lic`, one commit each.

**1. Study sigils out of sigil books (new `sigil_books` setting).**

Sigils had to be loose in the crafting container — `trace_sigil` issued a bare `get my <type> sigil`, which cannot reach a scroll filed in a sigil book. Now, when `sigil_books` is configured, each sigil is studied straight out of a book: turn to the type's section, read it for a page of that type, turn to that page, confirm the page, `STUDY` the book. That produces the same memorization as the loose path, so the `trace` afterwards is unchanged.

Books are searched in the listed order and picked up one at a time (stowed again before the next), since sigil books commonly share the noun `book`. A book only holds 20 scrolls, so a run usually spans several. When no book holds the requested type it falls back to the loose scroll exactly as before — with no books configured, behavior is identical to today.

```yaml
sigil_books:
  - small sigil book
  - large sigil book
```

**2. Continue scribing after an imbue resume.**

`handle_imbue_resume` ended at `imbue` and returned, so `run_enchant` finished and the script exited. The sigil prompt only arrives after the imbue roundtime, by which point nothing was listening, leaving the enchant half-finished until it was resumed a second time. `handle_new_enchant` has always followed `imbue` with get-burin-then-`scribe`; the resume path never did. Both now end in a shared `scribe_with_burin`.

## Game text this relies on

Captured in game rather than inferred; the spec fixtures are verbatim book output.

- The item reference for `TURN` must not contain the word "sigil", or the parser reads it as the sigil-type argument (`turn my sigil book to sigil congruence` → `Which sigil did you wish to turn the a small sigil book to?`). Commands use the bare noun.
- Turning to a section holding nothing still succeeds (`You turn the book to the rarefaction section.`) and simply reads empty, so presence is decided by the `READ`, not the `TURN`.
- `STUDY` with the book at the contents appraises the book (`You study the book and determine it is well-suited for holding sigil-scrolls.`) instead of memorizing anything. That reply is matched explicitly and treated as a failure, so a mis-sequenced study can't be mistaken for a memorized sigil.
- Studying blanks the page and renumbers the remaining pages, so no page map is cached — every sigil re-reads.
- The book must be held; both turns use a short timeout with `suppress_no_match`, since the subsequent `READ` is the real verification.

## Testing

- [x] `rspec` — 3054 examples, 0 failures (39 new in `spec/enchant_spec.rb`)
- [x] `rubocop enchant.lic` — clean
- [x] In game: enchant with `sigil_books` set, sigils pulled from the book across multiple sigils (pages renumber between studies)
- [x] In game: book exhausted of a type falls through to the next book, then to a loose sigil-scroll
- [ ] In game: `;enchant resume <item>` on an item awaiting an imbue now proceeds into scribing instead of exiting
- [ ] In game: unset `sigil_books` still retrieves loose sigil-scrolls as before

### DIFF
--- a/enchant.lic
+++ b/enchant.lic
@@ -172,8 +172,7 @@ class Enchant
     result = DRC.bput("analyze #{@item} on #{@brazier_ref}", *analyze_patterns)
 
     if ANALYZE_READY_PATTERNS.any? { |pat| pat.match?(result) }
-      DRCC.get_crafting_item(@burin, @bag, @bag_items, @belt)
-      scribe
+      scribe_with_burin
     elsif ANALYZE_IMBUE_PATTERN.match?(result)
       handle_imbue_resume
     else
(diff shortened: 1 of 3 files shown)


MahtraDR:
What's the plan with this one?

PR #7585, opened 2026-09-08
combat-trainer: appraise all available mobs each pass, gated by xp + cooldown

MahtraDR:
## What

Reworks `combat-trainer.lic`'s `appraise` (the `App` / `App Quick` / `App Careful` training actions) now that every creature carries a stable `<crtrStatus>` id.

### Before
- Appraised only the **first** non-blacklisted mob per pass.
- Kept a permanent `@no_app` blacklist that was populated **only** on `'Perhaps that ...'` (un-appraisable). A successfully-appraised mob was never recorded, so it got re-appraised every App tick; un-appraisable mobs were skipped for the rest of the session.
- Gate was rank-only (`Appraisal < 76`).

### After
- **XP gate:** skip entirely once Appraisal mindstate is at/over `@combat_training_abilities_target` — nothing left to learn this pulse. The hard rank floor (`< 76`) stays.
- **Appraise every mob:** loop `Creature.targets` and appraise each by id on each pass, instead of one.
- **Cooldown *is* the per-mob timer:** the App ability's configured cooldown (e.g. `App Quick: 120`) already spaces passes via `select_ability`, so it doubles as the "don't re-appraise a mob within the timer" gate. No id bookkeeping needed, and `@no_app` is removed.
- Un-appraisable mobs are simply retried next pass rather than permanently skipped.
- Added `break if game_state.retreating?` inside the loop and `waitrt?` between appraises so a mob's roundtime can't hang the next `app` on an unmatched "still recovering" line.

## Test plan
- `rubocop -A combat-trainer.lic` — no offenses.
- `bundle exec rspec spec/combat-trainer_spec.rb` — 5 examples, 0 failures.
- No spec previously covered `appraise`/`@no_app`; behavior change is confined to the `appraise` method.

### DIFF
--- a/combat-trainer.lic
+++ b/combat-trainer.lic
@@ -3659,8 +3659,6 @@ class TrainerProcess
         end
       end
 
-    @no_app = []
-
     @dont_stalk = settings.dont_stalk
     echo("  @dont_stalk: #{@dont_stalk}") if $debug_mode_ct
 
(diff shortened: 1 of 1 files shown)


MahtraDR:
Closing without merging.

The premise of this PR — that appraising every mob per cycle teaches Appraisal faster than appraising one — is disproven by an in-game log. Appraisal learning is gated by a **skill-wide ~34s pulse, not per-creature**:

- 86 `app #<id>` commands across 44 distinct creature ids over ~5 min → only **9** creature-appraisal XP pulses (excluding the almanac `+16`).
- Pulse gaps were flat at ~34s avg (27–44s) regardless of how many mobs were appraised between them.
- Decisive slice: in the first 27s the script appraised **11 distinct fresh ids and got exactly one `+1`**. A per-creature cooldown would have produced ~11 gains. Mindstate was not capped (7/34 → 27/34), so the ~34s pulse was the limiter.

So looping every target is ~9.6 appraisals per unit of learning — the extra appraisals only burn roundtime (and 8 were pure "cannot appraise … in combat" no-ops). `main` already appraises one target per cycle by id, which is all the pulse will absorb, so no change is needed.

(This PR was also red on CI — it removed the `@no_app`/single-target behavior that its own `spec/combat_trainer_spec.rb:5266-5301` still asserts.)

PR #7596, opened 2026-09-14
test: add load_lic_method extractor for scripts built from top-level defs (#7548)

### LINKED ISSUE #7548: spec: add a load_lic_method extractor for scripts built from top-level defs
`spec/spec_helper.rb` can extract a `class` (`load_lic_class`) or a `module` (`load_lic_module`) out of a `.lic` without running the script's top-level code. There is no equivalent for a plain top-level `def`, and that leaves a whole shape of script untestable.

## Why it matters

Not every script is built around a class. 16 of the 222 scripts hold 55 top-level `def`s between them, and four have **no** class or module at all:

| script | lines | classes/modules | top-level defs | has spec? |
| --- | --- | --- | --- | --- |
| circlecheck.lic | 838 | 0 | 8 | no |
| trigger-watcher.lic | 436 | 0 | 11 | no |
| heal-remedy.lic | 249 | 0 | 9 | no |
| jail-buddy.lic | 107 | 0 | 5 | no |

For these there is simply nothing for the current extractors to grab. Writing a spec means either restructuring the script into a class first, or hand-rolling an extractor in the spec file.

`dependency_spec.rb` already hit this and took the second option — it carries its own private `extract_method` and drives it over a list of eight method names. That is the same duplication `load_lic_module` had before it moved into spec_helper (it was defined twice, in `automap_spec` and `moonwatch_spec`), and `spec_helper.rb`'s own header calls out why duplicated top-level defs across spec files are a hazard: the whole suite runs in one process, so a duplicate definition wins for the entire run by load order.

## What it unlocks

- The four class-less scripts above become spec-able without restructuring them first — `circlecheck.lic` at 838 lines is the obvious candidate.
- `dependency_spec`'s private `extract_method` collapses into the shared helper, removing the duplicate-top-level-def hazard.
- Top-level helpers in scripts that *do* have classes become reachable — e.g. `combat-trainer.lic`'s `ct_safe_eval`, which wraps user-supplied conditional expressions from profile YAML in a `rescue` and is exactly the kind of thing worth a test.

## Sketch

`dependency_spec`'s existing implementation is most of the answer — it already handles de-indenting a nested `def` and finding the matching `end` at the right indent level:

```ruby
def load_lic_method(filename, method_name)
  return if Object.respond_to?(method_name, true)

  filepath = lic_path(filename)
  lines = File.readlines(filepath)
  # ... find `def <name>`, take through the matching `end` at the same indent,
  # strip the common indent, eval with the real start line
end
```

Worth matching the existing helpers' conventions: the `lic_path` expansion, the idempotence guard so repeated calls across co-running specs are safe, and (if #7547 lands) recording the eval'd range in `LIC_EVAL_RANGES` so coverage attribution stays correct.

Happy to put up a PR for this if there is no objection to the approach.

MahtraDR:
Resolves #7548.

## What & why

`spec/spec_helper.rb` can extract a `class`, `module`, or `constant` from a `.lic` without running the script's top-level code, but there was no equivalent for a plain top-level `def`. That left a whole shape of script untestable: the class-less scripts (`circlecheck.lic`, `trigger-watcher.lic`, `heal-remedy.lic`, `jail-buddy.lic`) and top-level helpers stranded in class-based scripts. `dependency_spec` had already hit this and hand-rolled a private `extract_method` — the same duplication `load_lic_module` was consolidated to avoid.

## Changes

- **`spec_helper.rb`: new `load_lic_method(filename, method_name)`.** Takes `def <name>` through the matching `end` at the same indent, strips the common indent, and evals the slice at its real start line so backtraces and coverage attribute correctly. Records the eval'd range in `LIC_EVAL_RANGES` like the other extractors.
- **Source-aware collision guard.** Top-level defs land as private methods on the shared `Object` namespace for the whole single-process suite, so two scripts each defining e.g. `main` would collide. The guard is idempotent for repeat calls on the *same* file but **raises** on a same-name-*different*-file collision instead of silently binding to whichever spec loaded first — the "a duplicate top-level def wins by load order" hazard the `spec_helper` header already warns about. (A plain `return`-style guard would hide exactly that bug.)
- **Migrate `dependency_spec.rb`** off its private `extract_method` onto the shared helper — proving the interface on its first consumer and removing the duplicated-top-level-def hazard. (The `DR_OBSOLETE_SCRIPTS` `.freeze` constant is not a `def`, so it stays on its existing constant-style extraction.)
- **Add `circlecheck_spec.rb`** covering the pure `clamp_progress` / `progress_bar` helpers of the class-less `circlecheck.lic`, without running its runtime-dependent `main`.

## Verification

- Full suite: **3216 examples, 0 failures** (before and after rubocop).
- `rubocop -A` clean on all three changed files under Ruby 4.0.
- Guard behavior proven by execution (temporary spec, not committed): idempotent same-file call; raises on same-name-different-file collision; raises a clear error on an absent def; the name anchor rejects a pure prefix (`clamp_progres` does not bind to `clamp_progress`).

## Notes / limitation

The `^<indent>end` scan is defeated by a heredoc or string literal containing `end` at the def's indent — a limitation shared with the existing `load_lic_class`/`load_lic_module` extractors, documented in a comment. Not a regression.


spec/spec_helper.rb
```diff
@@ -80,6 +80,16 @@
...
 
+# Which .lic file each top-level method was eval'd out of, keyed by method name.
+# Top-level defs land as PRIVATE INSTANCE METHODS ON Object and share one
+# namespace for the whole single-process suite, so two scripts that each define a
+# generically-named helper (e.g. circlecheck.lic's `main`) collide. A plain
+# idempotence guard would let whichever spec loaded first win *silently* -- the
+# exact "a duplicate top-level definition wins by load order" hazard the header
+# above warns about. This registry lets load_lic_method stay idempotent for the
+# SAME file but raise loudly on a same-name-different-file collision.
+LIC_METHOD_SOURCES = {}
```
simtel12: Would this be better if the extracted defs didn't land on `Object` at all? Something like evaluating each script's defs into its own module:

```ruby
mod = Module.new { extend self }
mod.module_eval(source, filepath, start_idx + 1)
```

and having the spec call `CircleCheck.progress_bar(25)` (or `include` it in the describe). I tried a quick version against `circlecheck.lic`: calls between the script's own helpers still resolve, harness methods like `echo` are still reachable, `allow(mod).to receive(...)` stubs cleanly, and `Object` stays untouched.

If that holds up, does the registry still need to exist? It seems like the collision it guards against couldn't happen anymore. The tradeoff I can see is that it reads a little differently from `load_lic_class`/`load_lic_module` — do you think that's worth it, or is there a reason to keep everything at top level that I'm missing?

spec/spec_helper.rb
```diff
@@ -233,6 +243,51 @@ def load_lic_constant(filename, const_name)
...
+# so a second spec extracting a same-named method from a different script fails
+# loudly instead of silently binding to whichever loaded first.
+#
+# Limitation shared with the class/module extractors: the `^<indent>end` scan is
+# defeated by a heredoc or string literal that contains `end` at the def's indent.
+def load_lic_method(filename, method_name)
+  filepath = lic_path(filename)
+
+  if (prior = LIC_METHOD_SOURCES[method_name])
+    return if prior == filepath
```
simtel12: What happens if the name already exists on `Object` but *wasn't* loaded through this helper?

`dependency.lic` defines `register_slackbot` and `send_slackbot_message`, and so does `test/test_harness.rb` (L1040–1042). After `dependency_spec` loads, `method(:register_slackbot).owner` is `Object`, so the harness stubs are shadowed for the whole run — and the guard doesn't notice, because `LIC_METHOD_SOURCES` has never heard of the harness. Same thing for a helper defined in a spec file: I defined one, then loaded a same-named def from a `.lic`, and the `.lic` version silently won.

Nothing depends on those two harness stubs today, so nothing breaks — but isn't that the exact "wins by load order" case the comment above says the guard exists to catch? Would checking `Object.private_method_defined?(name) || Object.method_defined?(name)` (for names not already in the registry) close that gap?

spec/spec_helper.rb
```diff
@@ -233,6 +243,51 @@ def load_lic_constant(filename, const_name)
...
+
+    raise "Top-level method '#{method_name}' already loaded from #{prior}; " \
+          "#{filepath} would overwrite it in the shared Object namespace. " \
+          'Rename one, or wrap the script in a class/module and use load_lic_class.'
```
simtel12: Should a test helper be nudging people to rename or restructure a production script? I checked, and no top-level def name is currently shared between two scripts (`circlecheck.lic` is the only one with a top-level `main`), so this path is hypothetical for now. If two scripts both legitimately have a `main`, it feels like the spec tooling, not the script, is the thing that should adapt — which is part of why the per-script module idea above appeals to me.

spec/spec_helper.rb
```diff
@@ -233,6 +243,51 @@ def load_lic_constant(filename, const_name)
...
+  lines = File.readlines(filepath)
+  prime_lic_coverage(filepath, lines.size)
+
+  # Anchor the end of the name so `foo` matches neither `foobar` nor `foo?`: the
+  # name must be followed by whitespace, `(`, `;`, or end-of-line.
+  start_idx = lines.index { |l| l =~ /^(\s*)def\s+#{Regexp.escape(method_name)}(?=[\s(;]|$)/ }
+  raise "Could not find 'def #{method_name}' in #{filename}" unless start_idx
+
+  indent = lines[start_idx][/^\s*/]
+  rel_end = lines[(start_idx + 1)..].index { |l| l =~ /^#{indent}end\s*$/ }
```
simtel12: How should this behave on a one-line def? With `def one; 1; end`, there's no bare `end` line, so the scan runs on to the *next* def's `end`. I tried it: loading only `one` also defined `two`, and `two` never made it into `LIC_METHOD_SOURCES`, so it's invisible to the guard.

No top-level def is written that way today, so this is more "what do we want to happen" than a live bug. Would it be worth raising if the start line already contains its own `end` (or if the slice defines more than one method)?

spec/spec_helper.rb
```diff
@@ -233,6 +243,51 @@ def load_lic_constant(filename, const_name)
...
+          "#{filepath} would overwrite it in the shared Object namespace. " \
+          'Rename one, or wrap the script in a class/module and use load_lic_class.'
+  end
+
+  lines = File.readlines(filepath)
+  prime_lic_coverage(filepath, lines.size)
+
+  # Anchor the end of the name so `foo` matches neither `foobar` nor `foo?`: the
+  # name must be followed by whitespace, `(`, `;`, or end-of-line.
+  start_idx = lines.index { |l| l =~ /^(\s*)def\s+#{Regexp.escape(method_name)}(?=[\s(;]|$)/ }
```
simtel12: nit: since this matches a `def` at any indent and takes the first hit, could a same-named instance method inside a class earlier in the file get picked up instead of the top-level one? Not the case in any current script — I checked all 16 — but is "any indent" something a consumer actually needs, given the helper is described as extracting top-level defs? There's a cross-file version of this too: `carve-bead`, `crossing-training`, `dusk-labyrinth`, `feed-cloak`, and `hunting-buddy` all have an indented `def main` inside a class. What would stop `load_lic_method('hunting-buddy.lic', 'main')` from lifting that instance method onto `Object`, and then tripping the guard against `circlecheck.lic`'s top-level `main`? Also the `(\s*)` capture isn't used (`indent` is recomputed on the next line).

simtel12 (commented):
I really like this — the class-less scripts have been a blind spot for the suite, and folding `dependency_spec`'s private `extract_method` into a shared helper is exactly the right cleanup. Nice catch on the old `[\s(]?` pattern letting `foo` match `def foobar`, too.

I left a few questions inline, mostly around the collision guard and what it can and can't see. The rest are nits.

nit (outside the diff, so leaving it here): in `spec/spec_helper.rb` L19, now that there's a shared helper, should this point at `load_lic_method` instead? Same for L61–62, which still lists only `load_lic_class / load_lic_constant` as the helpers spec_helper provides.

spec/dependency_spec.rb
```diff
@@ -49,7 +40,7 @@ def extract_method(lines, path, method_name)
   clear_hometown
   obsolete_script_dirs
   warn_obsolete_scripts
-].each { |fn| extract_method(dep_lines, dep_path, fn) }
+].each { |fn| load_lic_method('dependency.lic', fn) }
 
 # Extract the frozen DR_OBSOLETE_SCRIPTS constant (assignment line through the
 # terminating ".freeze"); extract_method only handles def bodies.
```
simtel12: nit: `extract_method` is gone after this PR — `load_lic_method only handles def bodies`?

spec/circlecheck_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      expect(progress_bar(0)).to eq("[#{'-' * 10}]")
+    end
+
+    it 'renders a full bar at 100 percent' do
+      expect(progress_bar(100)).to eq("[#{'#' * 10}]")
+    end
+
+    it 'rounds partial progress to the nearest cell' do
+      # 25% of 10 cells => 2.5 => rounds to 3 (Float#round is round-half-up)
+      expect(progress_bar(25)).to eq("[####{'-' * 7}]")
```
simtel12: nit: did it take you a second look too? It reads as four `#`, but the fourth starts the interpolation, so the expected value is `[###-------]`. The test is right, but would a plain `'[###-------]'` be easier on the next reader?

spec/spec_helper.rb
MahtraDR: Adopted, in `ba133aea`. `load_lic_methods` now evals the extracted defs into a fresh anonymous `Module.new` that `extend self`s, and returns it — internal bare cross-calls resolve, host/harness methods fall through to `Object`, and `Object` is left untouched. The `LIC_METHOD_SOURCES` registry and its raise-on-collision guard are gone entirely, since there is no shared namespace left to police. You were right that it reads more like `load_lic_class`, not less. Thanks for the nudge (and for actually prototyping it).

spec/spec_helper.rb
MahtraDR: Good catch, and it drove the redesign — but I went with the module approach above rather than the `Object.method_defined?` check, because that check would have broken this very spec. `dependency_spec` deliberately loads dependency.lic's real `register_slackbot`/`send_slackbot_message` over the harness stubs at test_harness.rb:1040-1042 to test them, so raising when the name already exists on `Object` would reject that on the first call. With the defs now living in a module, the harness stubs are no longer shadowed at all — so the load-order gap you found is closed at the root instead of guarded against. Fixed in `ba133aea`.

spec/spec_helper.rb
MahtraDR: Resolved by the module refactor in `ba133aea` — there's no shared namespace, so no collision, so the raise (and its "rename your script" nudge) is gone entirely. Agreed the tooling should adapt, not the script; now it just does.

spec/spec_helper.rb
MahtraDR: Added a guard in `ba133aea`: after slicing, it counts column-0 `def`s in the slice and raises `"...spans N top-level defs; rewrite it as a multi-line def"` if it is not exactly one. That catches both the one-line (`def x; 1; end`) and endless (`def x = 1`) cases, since either makes the `^end` scan run into a later def. Proven with a throwaway spec (loading `alpha` from a two-def fixture now raises; a well-formed multi-line def still extracts).

spec/spec_helper.rb
MahtraDR: Both fixed in `ba133aea`. The scan is now anchored to column 0 (`/^def\s+.../`), so `load_lic_methods('hunting-buddy.lic', 'main')` no longer lifts the indented instance `def main` — it raises `"Could not find top-level 'def main'"` (verified with a throwaway spec). Since it is column-0-only now, the `(\s*)` capture and the indent-stripping are gone too.

spec/dependency_spec.rb
MahtraDR: Fixed in `ba133aea` — the comment now reads that `load_lic_methods` only handles def bodies and `load_lic_constant` only a single-line assignment, which is why this multi-line `.freeze` constant is carved out by hand.

spec/circlecheck_spec.rb
MahtraDR: Yep — changed to a plain `'[###-------]'` in `ba133aea`. (CodeRabbit flagged the same thing independently, so clearly not just us.)

MahtraDR:
Thanks for the review — pushed `ba133aea` responding to all of it.

The big change: `load_lic_method` → `load_lic_methods`, which extracts a script's top-level defs into a fresh module (`extend self`) instead of onto `Object`, per your first inline suggestion. That retires the `LIC_METHOD_SOURCES` registry and the whole same-name-collision guard, and stops shadowing the harness stubs — so the two guard threads are moot now rather than patched.

On your header-doc nit: fixed both. spec_helper.rb L15-20 and the "Shared setup" block now list all four `load_lic_*` helpers (the method list was also missing `load_lic_module` before), and the stale "see dependency_spec.rb for the method-level extraction pattern" pointer is gone since it's a shared helper now.

Individual replies inline. Full suite still 3216 examples / 0 failures, rubocop clean; the column-0 anchor, the one-line-def guard, and Object-cleanliness were each proven by a throwaway spec (not committed).

MahtraDR:
@simtel12 should be all good now if you wouldn't mind looking again? Thanks!

spec/spec_helper.rb
simtel12: Following up on my own suggestion: the module wrapper has a side effect I should have thought about when I proposed it. The module is one long-lived object, so an `@ivar` an extracted def sets now survives from one example into the next — with top-level defs that state died with the RSpec example instance.

It bites sooner than I'd have guessed. Of the 55 top-level defs in the repo, 10 use ivars: `heal-remedy.lic` (7 defs — `@settings`, `@remedy_container`, `@wounds_applied`), `trigger-watcher.lic` (`@trigger_data`) and `jail-buddy.lic` (`@settings`, `@crime_plea`) — three of the four class-less scripts this PR exists to unlock. Several *write*: `@wounds_applied = true` in four heal-remedy defs, `@trigger_data` reassigned in `execute_reload`. So the first spec after `circlecheck` runs into it. Note too that those ivars are initialised by top-level code the extractor doesn't run (`heal-remedy.lic:56-67`), so a spec has to inject them regardless.

Three options:

1. **Central reset:** `load_lic_methods` registers each module, and the global hook clears their ivars next to `reset_data`. Automatic, but it also wipes state set in a `before(:context)`.
2. **`include HealRemedy` in the describe block:** calls run on the example instance, so injection is just `before { @remedy_container = 'sack' }` and isolation is free. Costs the module-style call sites you've written, and a script's `def echo` would shadow the harness one for that group.
3. **Fresh receiver per example:** `let(:s) { Object.new.extend(HealRemedy) }`.

I'd lean (2) for the ivar-heavy scripts and (1) if you want to keep module-style calls everywhere — but you know the suite's conventions better than I do, so I'm happy either way. Worth deciding before the next class-less spec lands rather than after.

MahtraDR: Confirmed — and nice catch on the consequence of your own suggestion. I reproduced it before deciding what to do rather than reasoning about it: because the returned module is one process-lived object, `Mod.foo` runs with `self == Mod`, so an `@ivar` a def assigns does survive into the next example, and `before(:each) { reset_data }` doesn't touch it. Throwaway spec against a heal-remedy-shaped fixture (top-level `@wounds_applied = false` init that isn't extracted; one def sets it `true`, another reads it): example A sets it, example B sees `true` with no setup of its own. The sharper half you flagged holds too — setting `@wounds_applied` in the example body doesn't reach the method, since the hook writes the *example instance's* ivar while the method reads the *module's*.

Two notes:

1. **It doesn't touch anything in this PR.** The leak needs an ivar-carrying extracted def, and neither spec here has one — `circlecheck`'s `clamp_progress`/`progress_bar` are pure, and all eight `dependency` defs are `$global`-only (verified). So this is the "decide before the next class-less spec" call you framed it as, not a defect in the diff.

2. **The fix needs no code change.** `mod.extend(mod)` leaves the defs as plain instance methods, so a future ivar-spec can just `include` the returned module in its `describe` block (your option 2): the defs then run on the per-example instance, `before { @settings = ... }` reaches them, and isolation is free. I went with documenting that over a central-reset hook — option 1 would also wipe `before(:context)` state, as you noted. Caveat added to the `load_lic_methods` doc comment in a208c6cd.

Net guidance: `Mod.foo` for pure/`$global` scripts (circlecheck, dependency), `include Mod` for the ivar ones (heal-remedy, jail-buddy, trigger-watcher). The `include` cost you named — a script's top-level `def echo` shadowing the harness `echo` for that group — is real, but none of the three define one today and it'd surface immediately if one did.

Thanks for chasing down the consequence of your own idea; the cross-example state axis is exactly where this kind of thing hides.

MahtraDR: @simtel12 We all good on this PR now? WHat do you think?

simtel12 (approved):
Looks good to me!

mrhoribu:
# PR Review: test: add load_lic_method extractor for scripts built from top-level defs (#7548) (elanthia-online/dr-scripts#7596)

**Reviewed:** 3 files, +138/-73 (real scope; see Coverage notes on the diff-base correction) at `375ed61e` on 2026-09-21. Based on: full clone, changed files at head, full commit history, and a live worktree checkout.
**Verdict:** Approve
**Summary:** Adds `load_lic_methods` to `spec/spec_helper.rb` — a fourth extraction helper (alongside `load_lic_class`/`load_lic_module`/`load_lic_constant`) that pulls one or more top-level `def`s out of a class-less `.lic` script and returns them wrapped in a fresh, self-extended anonymous module, so bare top-level helpers can be unit-tested without running the script's runtime-dependent entry point and without polluting the shared `Object` namespace. `dependency_spec.rb` is migrated off its hand-rolled private `extract_method` onto the new helper, and a new `circlecheck_spec.rb` exercises the pattern on a genuinely class-less script. This PR already merged; the review confirms the merged state.

## What this PR gets right

- **Solves its own worst version first.** The PR's commit history (visible via `git log`) shows the author's *first* cut bound extracted defs directly onto `Object` with a raise-on-collision registry — then, in response to review feedback, replaced it with the current self-extended anonymous module. That's a strictly better design (two scripts defining `main` now just get independent modules instead of one failing loudly) and the collision-registry code was cleanly removed rather than left dead.
- **Column-0 anchor is correct and tested.** `^def\s+#{Regexp.escape(method_name)}(?=[\s(;]|$)` only matches a def at the start of a line, so an indented instance method of the same name inside an earlier class (the `hunting-buddy.lic` `def main` case cited in the commit message) is never mis-extracted. Verified by hand-tracing the regex against a synthetic file with a name-prefix collision (`clamp_progres` vs `clamp_progress`) — it correctly skips the prefix and binds the real name at the correct line.
- **The one-line/endless-def guard is real, not aspirational.** Reproduced the exact failure mode outside the guard (a `def foo; 1; end` line has no `^end` of its own, so the `end`-scan runs on to a sibling def's `end` and would silently swallow it) and confirmed the `slice.count { |l| l =~ /^def\s/ } == 1` check catches it and raises instead of returning a corrupted slice.
- **Sibling top-level calls resolve correctly.** `dependency.lic`'s `warn_obsolete_scripts` calls the bare (unqualified) `obsolete_script_dirs`, and both are extracted into the same module. `mod.extend(mod)` puts `mod` in its own singleton-class ancestor chain, so `Dependency.warn_obsolete_scripts` resolves the bare `obsolete_script_dirs` call correctly — proven by the pre-existing `expect(Dependency.warn_obsolete_scripts).to eq(['roomnumbers'])` spec, which passes.
- **Removes a real, cited hazard.** The prior `extract_method` in `dependency_spec.rb` eval'd straight onto `Object`, meaning `register_slackbot` et al. were live global methods for the whole single-process suite — capable of silently shadowing (or being shadowed by) `test/test_harness.rb`'s own `register_slackbot` stub depending on load order. The new module-wrapped version can't touch `Object` at all, closing that off structurally rather than by convention.
- **Honest, forward-looking documentation of a real limitation.** The final commit adds (without any code change) a clear warning that the returned module is a long-lived singleton, so `@ivar` state set by one extracted def persists across examples — harmless for the two pure/`$global`-only scripts extracted here (confirmed: neither `circlecheck.lic`'s `clamp_progress`/`progress_bar` nor the extracted `dependency.lic` helpers touch `@ivar`s), but a real trap for the next class-less spec (`heal-remedy.lic`, `jail-buddy.lic`, `trigger-watcher.lic`, none of which have specs yet). Flagging this before it bites anyone is good citizenship.

## Nits

None worth listing — the extractor's docstring is long, but matches this file's existing convention (the SimpleCov block above it is equally dense), and every claim in it is verified accurate against the code.

## Coverage notes

- **Diff-base correction:** the PR's true parent is `23811b65` (single squash-merge commit `375ed61e` landed directly on `main`). My first diff attempt used a stale local `main` ref as the base and incorrectly attributed a large, unrelated SimpleCov/coverage block (actually from an earlier, separate PR #7547) and `load_lic_module`/`lic_path`/`prime_lic_coverage` (pre-existing) to this PR. Caught and corrected before drawing any conclusions from it — flagging this here only so the correction is on record, since it would otherwise have produced a confident-but-wrong "scope creep" finding.
- Ran the full RSpec suite against a worktree checkout of the merged commit: **3222 examples, 0 failures** (small delta from the PR's self-reported 3216 is `main` moving on since merge, unrelated to this change).
- Ran `rubocop` (Ruby 4.0.5 via rbenv) on all three changed files: **no offenses**.
- `gh pr checks 7596`: both CI jobs (`Run tests on Ruby 4.0`, CodeRabbit) report `pass`.
- Ran a fresh local CodeRabbit pass (`coderabbit review --agent --base-commit 23811b65 --committed`, authenticated as the repo org): **0 findings**.
- Verified the extraction logic against the actual `circlecheck.lic` (lines 485–504) and `dependency.lic` (lines 74–151) source — every extracted method's real source matches what the specs assert.
- Confirmed no leftover references anywhere in `spec/` or `test/` to the removed private `extract_method` or to the old bare (unqualified) method-name call style.
- Confirmed no other `.lic` script currently has a spec that would be exposed to the documented `@ivar`-persistence caveat (`heal-remedy.lic`/`jail-buddy.lic`/`trigger-watcher.lic` have no specs yet), so it is pure forward documentation today, not a live bug.
- Did not review `lich-5` for any interaction, since this PR touches only `dr-scripts` test infrastructure with no runtime/production code changes.

PR #7605, opened 2026-09-18
Update base-capture.yaml

Elysande1218:
Memory Garden was recreated, so remapped and a new route established


MahtraDR:
Thank you!

PR #7607, opened 2026-09-19
[data][base-spells] Add Prosperity in Absentia

kahlen-tech:
Define the Trader spell Prosperity in Absentia (PRIA) so dependency can fill in its spell data, and add the cast message for it.

Casting PRIA where no kiosk stands yet reports "Your tessera grows cold, ... a crystalline kiosk unfolds from nothingness". That line was absent from cast_messages, so DRCA.cast? never matched it and every cast that raises a new kiosk blocked for the full bput timeout before the calling script continued.

Recasting where a kiosk already stands instead reports "You gesture at a <adj> crystalline kiosk ... the spell pattern is refreshed", which the existing "^You gesture" entry already matches -- so only the new-kiosk case was affected, which is why it went unnoticed.


MahtraDR:
@urbaj-dr check for us?

urbaj-dr (changes requested):
I think that tessera nouns can be altered, so recommend replacing that noun with a wildcard ([see PR](https://github.com/kahlen-tech/dr-scripts/pull/1)). Everything else looks good. Thanks!

data/base-spells.yaml
```diff
@@ -278,6 +278,8 @@ cast_messages:
 - Sutures of reddish-black energy delve deep
 - Tears of hot regret flow down your eyes
 - Your eyes flare with a whorl
+# Prosperity in Absentia raising a new kiosk
+- ^Your tessera grows cold
```
MahtraDR:
```suggestion
- ^Your \w+ grows cold
```

MahtraDR (changes requested):

MahtraDR (approved):

MahtraDR:
Thank you!🙏🏾

